PR #1, opened 2020-09-23
Add first version of naive solver.

twalth3r:
This PR adds a first version of a naive order matching algorithm for a single token pair:

**Given:**
* a vector of orders selling t0 for t1
* a vector of orders selling t1 for t0

**Result:**
The naive solver simply matches the two orders with the best limit price (if possible). It returns two vectors containing the matched sell amounts of all orders.

**Not (yet) included:**
* anything related to the Uniswap AMM
* fees

Added several test cases.

Notice: Even though a lot of the code was borrowed from @bh2smith's naive solver for GPv1, this is my first ever non-trivial endeavor in rust, so please forgive any nonsense that I may have produced :upside_down_face: Happy to receive any suggestions for improvement obviously!

### DIFF
--- a/src/batcher/solve_pair.rs
+++ b/src/batcher/solve_pair.rs
@@ -79,8 +79,8 @@ pub fn solve_pair(
     let nr_orders_token1 = sell_orders_token1.len();
 
     // Init vectors of sell amounts.
-    let mut sell_volumes_token0 = vec![U256::zero(); nr_orders_token0];
-    let mut sell_volumes_token1 = vec![U256::zero(); nr_orders_token1];
+    let mut executed_sell_orders_token0: Vec<Order> = vec![];
+    let mut executed_sell_orders_token1: Vec<Order> = vec![];
 
     // Match orders with best limit prices, if possible.
     if !(sell_orders_token0.is_empty() || sell_orders_token1.is_empty()) {
@@ -98,16 +98,14 @@ pub fn solve_pair(
                     orders: [best_sell_order_token0, best_sell_order_token1],
                 }
             );
-            sell_volumes_token0[nr_orders_token0 - 1] = matched_amounts[0];
-            sell_volumes_token1[nr_orders_token1 - 1] = matched_amounts[1];
+            executed_sell_orders_token0.push(best_sell_order_token0.clone());
+            executed_sell_orders_token1.push(best_sell_order_token1.clone());
         };
     };
 
     return Ok(Solution {
-        sell_orders_token0: sell_orders_token0.clone(),
-        sell_volumes_token0: sell_volumes_token0,
-        sell_orders_token1: sell_orders_token1.clone(),
-        sell_volumes_token1: sell_volumes_token1,
+        sell_orders_token0: executed_sell_orders_token0,
+        sell_orders_token1: executed_sell_orders_token1,
     });
 }
 
@@ -162,8 +160,8 @@ pub mod test_util {
         let orders0: Vec<Order> = vec![];
         let orders1: Vec<Order> = vec![];
         let solution = solve_pair(&orders0, &orders1).unwrap();
-        assert_eq!(solution.sell_volumes_token0, vec![]);
-        assert_eq!(solution.sell_volumes_token1, vec![]);
+        assert_eq!(solution.sell_orders_token0, vec![]);
+        assert_eq!(solution.sell_orders_token1, vec![]);
     }
 
 
@@ -207,8 +205,8 @@ pub mod test_util {
             }
         ];
         let solution = solve_pair(&orders0, &orders1).unwrap();
-        assert_eq!(solution.sell_volumes_token0, vec![U256::zero()]);
-        assert_eq!(solution.sell_volumes_token1, vec![U256::zero()]);
+        assert_eq!(solution.sell_orders_token0, vec![]);
+        assert_eq!(solution.sell_orders_token1, vec![]);
     }
 
     #[test]
@@ -242,8 +240,8 @@ pub mod test_util {
             }
         ];
         let solution = solve_pair(&orders0, &orders1).unwrap();
-        assert_eq!(solution.sell_volumes_token0, vec![U256::from_dec_str("100").unwrap()]);
-        assert_eq!(solution.sell_volumes_token1, vec![U256::from_dec_str("80").unwrap()]);
+        assert_eq!(solution.sell_orders_token0, orders0);
+        assert_eq!(solution.sell_orders_token1, orders1);
     }
 
     #[test]
@@ -277,8 +275,8 @@ pub mod test_util {
             }
         ];
         let solution = solve_pair(&orders0, &orders1).unwrap();
-        assert_eq!(solution.sell_volumes_token0, vec![U256::from_dec_str("80").unwrap()]);
-        assert_eq!(solution.sell_volumes_token1, vec![U256::from_dec_str("100").unwrap()]);
+        assert_eq!(solution.sell_orders_token0, orders0);
+        assert_eq!(solution.sell_orders_token1, orders1);
     }
 
     #[test]
@@ -312,8 +310,8 @@ pub mod test_util {
             }
         ];
         let solution = solve_pair(&orders0, &orders1).unwrap();
-        assert_eq!(solution.sell_volumes_token0, vec![U256::from_dec_str("100").unwrap()]);
-        assert_eq!(solution.sell_volumes_token1, vec![U256::from_dec_str("100").unwrap()]);
+        assert_eq!(solution.sell_orders_token0, orders0);
+        assert_eq!(solution.sell_orders_token1, orders1);
     }
 
     #[test]
@@ -371,8 +369,8 @@ pub mod test_util {
             }
         ];
         let solution = solve_pair(&orders0, &orders1).unwrap();
-        assert_eq!(solution.sell_volumes_token0, vec![U256::zero(), U256::from_dec_str("100").unwrap()]);
-        assert_eq!(solution.sell_volumes_token1, vec![U256::zero(), U256::from_dec_str("80").unwrap()]);
+        assert_eq!(solution.sell_orders_token0, vec![orders0[1].clone()]);
+        assert_eq!(solution.sell_orders_token1, vec![orders1[1].clone()]);
     }
 
 
--- a/src/batcher/solve_pair.rs
+++ b/src/batcher/solve_pair.rs
@@ -1,16 +1,438 @@
 use crate::models::order::Order;
 use crate::models::solution::Solution;
 use anyhow::Result;
-use web3::types::U256;
+use web3::types::{U256, H256, Address};
+
+
+pub enum OrderPairType {
+    LhsFullyFilled,
+    RhsFullyFilled,
+    BothFullyFilled,
+}
+
+
+trait Matchable {
+    /// Returns which of the two orders can be fully matched, or both, or none.
+    fn match_compare(&self, other: &Order) -> Option<OrderPairType>;
+    /// Returns whether the orders can be matched. For this the tokens need
+    /// to match, there must be a price that satisfies both orders.
+    fn attracts(&self, other: &Order) -> bool;
+    /// Returns whether this order's sell token is the other order's buy token
+    /// and vice versa.
+    fn opposite_tokens(&self, other: &Order) -> bool;
+    /// Returns whether there is a price that satisfies both orders.
+    fn have_price_overlap(&self, other: &Order) -> bool;
+}
+
+
+impl Matchable for Order {
+    fn match_compare(&self, other: &Order) -> Option<OrderPairType> {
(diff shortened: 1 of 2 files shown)


src/batcher/solve_pair.rs
```diff
@@ -1,16 +1,438 @@
...
 pub fn solve_pair(
     sell_orders_token0: &Vec<Order>,
     sell_orders_token1: &Vec<Order>,
 ) -> Result<Solution> {
+    // Can we assume that the orders are sorted already?
+    assert!(check_orders_sorted_by_limit_price(&sell_orders_token0));
+    assert!(check_orders_sorted_by_limit_price(&sell_orders_token1));
+    // Or do we need to sort them first here?
+    //sell_orders_token0.sort();
+    //sell_orders_token1.sort();
```
bh2smith, line 77: Was this here before? I don't think we need this commented code if its not gonna be used... or was this a question?

src/batcher/solve_pair.rs
```diff
@@ -1,16 +1,438 @@
...
+    // Init vectors of sell amounts.
+    let mut sell_volumes_token0 = vec![U256::zero(); nr_orders_token0];
+    let mut sell_volumes_token1 = vec![U256::zero(); nr_orders_token1];
+
+    // Match orders with best limit prices, if possible.
+    if !(sell_orders_token0.is_empty() || sell_orders_token1.is_empty()) {
+
+        // The best orders are the last elements in their vectors.
+        let best_sell_order_token0 = &sell_orders_token0[nr_orders_token0 - 1];
+        let best_sell_order_token1 = &sell_orders_token1[nr_orders_token1 - 1];
```
marcovc, line 92: Should this be assumed, or should it be part of this function's job?

marcovc (approved):
I don't understand much of what is going on here, but I'm sure you do :)

src/batcher/solve_pair.rs
```diff
@@ -1,16 +1,438 @@
...
+    BothFullyFilled,
+}
+
+
+trait Matchable {
+    /// Returns which of the two orders can be fully matched, or both, or none.
+    fn match_compare(&self, other: &Order) -> Option<OrderPairType>;
+    /// Returns whether the orders can be matched. For this the tokens need
+    /// to match, there must be a price that satisfies both orders.
+    fn attracts(&self, other: &Order) -> bool;
```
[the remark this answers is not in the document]
twalth3r: Borrowed from https://github.com/gnosis/dex-services/blob/56d192b558f0e66e6c18c269e5bd1549dd990f20/services-core/src/price_finding/naive_solver.rs#L24

src/batcher/solve_pair.rs
twalth3r: This was meant as a question since I don't know about the surroundings :shrug:

src/batcher/solve_pair.rs
twalth3r: I wasn't sure. My assumption was that orders are sorted somewhere else, so I only put some asserts at the beginning of `solve_pair()`.

marcovc: What I meant is that even if the orders come up sorted, there's no guarantees that the last elements of both vectors will be matchable, right? I'm imagining the staircases [here](https://docs.google.com/document/d/1R2xyvY0KVcLeuDP_qV9FQWOUsozK7mjh-3IO6sMXtnc/edit#heading=h.rdmscaqzw5ke) could cross and continue going up in opposite directions.

src/batcher/solve_pair.rs
```diff
@@ -1,16 +1,438 @@
...
+        let best_sell_order_token0 = &sell_orders_token0[nr_orders_token0 - 1];
+        let best_sell_order_token1 = &sell_orders_token1[nr_orders_token1 - 1];
+
+        if let Some(order_pair_type) = best_sell_order_token0.match_compare(
+            &best_sell_order_token1
+        ) {
+            let matched_amounts = get_matched_amounts(
+                &Match {
+                    order_pair_type,
+                    orders: [best_sell_order_token0.clone(), best_sell_order_token1.clone()],
```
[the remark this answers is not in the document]
twalth3r: Right - I pushed an update doesn't require the `clone()` (but other weird things I don't really understand :see_no_evil:)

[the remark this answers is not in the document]
twalth3r: The original version was borrowed again from https://github.com/gnosis/dex-services/blob/56d192b558f0e66e6c18c269e5bd1549dd990f20/services-core/src/price_finding/naive_solver.rs#L166-L169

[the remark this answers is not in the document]
marcovc: Wow! I'm sorry I asked, it looks like the answer is even more intriguing. I'll see if I can find out what's going on ;)

src/batcher/solve_pair.rs
twalth3r: > I'm imagining the staircases here could cross and continue going up in opposite directions.

If they cross, there's also a match, right? Only if they don't cross at all there's no match. By "best orders" I mean the ones offering the lowest limit price, i.e., wanting to receive the least amount of tokens, so if these can't be matched, then nothing can matched. This is checked in the if-statement right below.

src/batcher/solve_pair.rs
```diff
@@ -1,16 +1,439 @@
...
+
+type MatchedAmounts = [U256; 2];
+
+
 
 pub fn solve_pair(
     sell_orders_token0: &Vec<Order>,
     sell_orders_token1: &Vec<Order>,
 ) -> Result<Solution> {
+    // Can we assume that the orders are sorted already?
```
josojo, line 74: Yes. I think we should make sure that they are always sorted, so that we can be even quicker with our solution provision

twalth3r: Ok, I removed the commented-out code in `bb0270f`.

src/batcher/solve_pair.rs
```diff
@@ -1,16 +1,439 @@
...
+                    orders: [best_sell_order_token0, best_sell_order_token1],
+                }
+            );
+            sell_volumes_token0[nr_orders_token0 - 1] = matched_amounts[0];
+            sell_volumes_token1[nr_orders_token1 - 1] = matched_amounts[1];
+        };
+    };
+
     return Ok(Solution {
         sell_orders_token0: sell_orders_token0.clone(),
```
josojo, line 107: As we just discussed in person, lets put only the filled orders into the solution vector. Since all orders will be fully filled, our solution struct should not have the sell_volumes_token0 and buy_volumes_token0. I think it would be cool, if you could do it in this PR as well

src/batcher/solve_pair.rs
marcovc: Ah.. I thought "best" order meant the ones with "better" price for the order owners. Everything is cool then, thanks!

src/batcher/solve_pair.rs
```diff
@@ -1,16 +1,435 @@
 use crate::models::order::Order;
 use crate::models::solution::Solution;
 use anyhow::Result;
-use web3::types::U256;
+use web3::types::{U256, H256, Address};
+
+
+pub enum OrderPairType {
```
josojo, line 7: I am wondering whether we need this particular enum: 
Since we will fill both orders fully, it is probably not really needed. But on the other hand, it would be good if the: sell_orders_token0 would always have the orders in the batch, where the sellvolume exceeds the other sellvolume, in our formula 'A(T0)>pA(T1)`, so that the smart contract does not need too invert the orders beforehand.

I am not sure, whether we want to considered this already in this PR. As you think, @twalth3r

src/batcher/solve_pair.rs
```diff
@@ -1,16 +1,435 @@
...
+        } else {
+            Some(OrderPairType::BothFullyFilled)
+        }
+    }
+
+    fn attracts(&self, other: &Order) -> bool {
+        self.opposite_tokens(other) && self.have_price_overlap(other)
+    }
+
+    fn opposite_tokens(&self, other: &Order) -> bool {
```
josojo, line 47: I think we can also assume that all orders are opposite_tokens if they are passed to the solver

josojo:
Cool Tom!
Sorry, you took the hard path with using these "complicated" traits and matchable checkers. I think one would not have to use them in our case: Simply popping the order vectors and comparing the limit prices within one function would have been sufficient. 

But it is cool that you are getting experience with it!

src/batcher/solve_pair.rs
twalth3r: Ok, changed that in 3efc3f4

src/batcher/solve_pair.rs
twalth3r: Removed quite some code that doesn't appear to be required for this first version in 096afc3 - also related to your previous comment:
> I am wondering whether we need this particular enum:
Since we will fill both orders fully, it is probably not really needed

src/batcher/solve_pair.rs
```diff
@@ -1,16 +1,359 @@
...
+
+    // Match orders with best limit prices, if possible.
+    if !(sell_orders_token0.is_empty() || sell_orders_token1.is_empty()) {
+
+        // The best orders are the last elements in their vectors.
+        let best_sell_order_token0 = &sell_orders_token0[nr_orders_token0 - 1];
+        let best_sell_order_token1 = &sell_orders_token1[nr_orders_token1 - 1];
+
+        if best_sell_order_token0.have_price_overlap(&best_sell_order_token1) {
+            executed_sell_orders_token0.push(best_sell_order_token0.clone());
```
josojo, line 45: You could write here:

Ok(Solution {
        sell_orders_token0: vec![best_sell_order_token0.clone()],
        sell_orders_token1: vec![best_sell_order_token1.clone()],
    });

Then you don't have to define these mutable executed_sell_orders_token objects.

src/batcher/solve_pair.rs
```diff
@@ -1,16 +1,359 @@
...
+
+    // Init vectors of sell amounts.
+    let mut executed_sell_orders_token0: Vec<Order> = vec![];
+    let mut executed_sell_orders_token1: Vec<Order> = vec![];
+
+    // Match orders with best limit prices, if possible.
+    if !(sell_orders_token0.is_empty() || sell_orders_token1.is_empty()) {
+
+        // The best orders are the last elements in their vectors.
+        let best_sell_order_token0 = &sell_orders_token0[nr_orders_token0 - 1];
```
josojo, line 41: You can also use the pop function:
See here:
https://doc.rust-lang.org/std/vec/struct.Vec.html
in order to avoid the nr_order_token declarations

josojo (approved):
Nice work Tom :) it rocks 👍

src/batcher/solve_pair.rs
twalth3r: It's gone.

src/batcher/solve_pair.rs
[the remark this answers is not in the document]
twalth3r: It's gone now.

src/batcher/solve_pair.rs
twalth3r: I wasn't sure if it's good practice to have multiple exit points in a function, but changed in 5cfd6fd now!

src/batcher/solve_pair.rs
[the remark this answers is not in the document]
nlordell: > Just curious: why does rust need clone here?

This is because of Rust ownership semantics. In layman's terms, only one binding (aka variable) can "own" any value at any time. If the `clone()` were removed, then the value would be "moved", i.e. the ownership would have been transferred from `sell_orders_token0` vec to the `Match {...}` instance. This is not allowed here, since `sell_orders_token0` is itself a "borrowed" value, so "moving" out of it is prohibited.

I would say that this is the concept with the steepest learning curve in Rust. I suggest looking at the chapter in the book about ownership and borrowing which probably explains it much better than I do:
https://doc.rust-lang.org/book/ch04-00-understanding-ownership.html

src/batcher/solve_pair.rs
nlordell: `pop` requires a mutable vec, but there is also [`last`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.last). Then the if statement can be re-written as:

```rust
if let (Some(best_sell_order_token0), Some(best_sell_order_token1)) = (sell_orders_token0.last(), sell_orders_token1.last()) {

}
```

nlordell (approved):
Looks good! Nice Rust-ing @twalth3r

src/batcher/solve_pair.rs
twalth3r: That's nice, thanks! Changed it in ac676c0

src/batcher/solve_pair.rs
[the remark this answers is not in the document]
marcovc: Thanks Nick! I was reading about it yesterday. It's (almost) clear now ;)

PR #2, opened 2020-09-23
order model + associated functions

josojo:
closes https://github.com/gnosis/dex-services/issues/1434

testplan:
unit tests only

I was not aware to properly use the re-exported types from ethcontract-rs {U25}, as I need to have the ethabi::encode function using another Token definition. I am not sure how to deal with it. Maybe @nlordell knows

```
expected enum `models::orders::ethabi::Token`, found enum `ethabi::token::token::Token`
note: perhaps two different versions of crate `ethabi` are being used?rustc(E0308)
expected enum `models::orders::ethabi::Token`, found enum `ethabi::token::token::Token`
```


nlordell:
> I was not aware to properly use the re-exported types from ethcontract-rs {U25}, as I need to have the ethabi::encode function using another Token definition.

The issue is that we are using two versions of `ethabi`. We have updated the version in `ethcontract` (gnosis/ethcontract-rs#340) but haven't released recently.

nlordell:
I created a PR for a version `ethcontract` version bump here: gnosis/ethcontract-rs#380

src/models/orders.rs
```diff
@@ -0,0 +1,142 @@
+extern crate anyhow;
+extern crate ethabi;
+extern crate ethsign;
+extern crate rustc_hex;
+extern crate serde_json;
+extern crate web3;
+
```
nlordell, line 7: `extern crate *` is no longer needed in Rust 2018 edition.

Cargo.toml
```diff
@@ -7,3 +7,12 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+ethcontract = "0.7.2"
+ethabi = "12.0.0"
```
nlordell, line 11: To not run into issues with crate version mismatches, I think it would be better to use `ethcontract::common::abi` which is a re-export of `ethabi`.

Cargo.toml
```diff
@@ -7,3 +7,12 @@ edition = "2018"
...
+serde = { version = "1.0", features = ["derive"] }
+multihash = "0.11.4"
```
nlordell, line 13: I don't think this is being used.

Cargo.toml
```diff
@@ -7,3 +7,12 @@ edition = "2018"
...
+rustc-hex = "2.1.0"
+anyhow = "1.0.32"
+web3 = "0.13.0"
```
nlordell, line 16: As for `ethabi`, I think it would be better to use `ethcontract::web3` to avoid version conflicts.

Cargo.toml
```diff
@@ -7,3 +7,12 @@ edition = "2018"
...
+serde_json = "1.0.57"
+ethsign = "0.7.3"
```
nlordell, line 18: I would use `ethcontract::web3::signing::Signature` instead of adding an additional crate dependency for the Signature data-type.

src/models/orders.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub valid_until: U256,
+}
+
+impl Order {
+    #[allow(dead_code)]
+    pub fn get_digest(&self) -> Result<[u8; 32]> {
+        let domain_separator: H256 = serde_json::from_str(
+            r#""0x24a654ed47680d6a76f087ec92b3a0f0fe4c9c82c26bff3bb22dffe0f120c7f0""#,
+        )
+        .unwrap();
```
nlordell, line 42: Just curious, where does this domain separator come from? Also, I would prefer the H256 `FromStr` implementation directly (that way you avoid the need to escape the `"` with the `r##` here-doc:
```suggestion
        let domain_separator: H256 = "0x24a654ed47680d6a76f087ec92b3a0f0fe4c9c82c26bff3bb22dffe0f120c7f0"
            .parse()
            .unwrap();
```

src/models/orders.rs
```diff
@@ -0,0 +1,142 @@
...
+    }
+    #[allow(dead_code)]
+    pub fn validate_order(&self) -> Result<bool> {
+        let message = self.get_digest()?;
+        let signature = Signature {
+            v: self.signature_v - 27 as u8,
+            r: *self.signature_r.as_fixed_bytes(),
+            s: *self.signature_s.as_fixed_bytes(),
+        };
+        let owner = signature.recover(&message)?;
```
nlordell, line 61: You could use `ethcontract::web3::signing::recover` instead of adding a `ethsign` dependency.

src/models/orders.rs
```diff
@@ -0,0 +1,142 @@
...
+impl Ord for Order {
+    fn cmp(&self, other: &Self) -> Ordering {
+        (self.sell_amount.mul(other.buy_amount)).cmp(&self.buy_amount.mul(other.sell_amount))
+    }
+}
+
+impl PartialOrd for Order {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        (self.sell_amount.mul(other.buy_amount))
+            .partial_cmp(&self.buy_amount.mul(other.sell_amount))
```
nlordell, line 100: ubernit: I would avoid the duplicate ordering implementation:
```suggestion
        Some(self.cmp(other))
```

src/models/orders.rs
```diff
@@ -0,0 +1,142 @@
...
+            nonce: 1,
+            signature_v: 27 as u8,
+            signature_r: serde_json::from_str(
+                r#""0x07cf23fa6f588cc3a91de8444b589e5afbf91c5d486c512a353d45d02fa58700""#,
+            )
+            .unwrap(),
+            signature_s: serde_json::from_str(
+                r#""0x53671e75b62b5bd64f91c80430aafb002040c35d1fcf25d0dc55d978946d5c11""#,
+            )
+            .unwrap(),
```
nlordell, line 85: Same comment as before, I think it makes more sense to use the `FromStr` implementation with `"...".parse().unwrap()` directly instead of parsing JSON for `Address`es and `H256`s.

src/models/orders.rs
```diff
@@ -0,0 +1,142 @@
...
+extern crate web3;
+
+use anyhow::Result;
+use ethabi::encode;
+use ethsign::Signature;
+use serde::{Deserialize, Serialize};
+use std::cmp::Ord;
+use std::cmp::Ordering;
+use std::cmp::PartialOrd;
+use std::ops::Mul;
```
nlordell, line 15: All `std::ops::*` traits are for operator overloading. So `Mul::mul(a, b)` (or `a.mul(b)`) is exactly equivalent to `a * b`. In fact, the `*` operator looks for the `Mul` implementation so that it knows what to call, so in general, you shouldn't need to `use` `std::ops::*` traits unless you intend on implementing them or using them as type constraints for generic methods.

src/models/orders.rs
```diff
@@ -0,0 +1,142 @@
...
+            )
+            .unwrap(),
+            valid_until: U256::from("0"),
+        }
+    }
+}
+
+impl Ord for Order {
+    fn cmp(&self, other: &Self) -> Ordering {
+        (self.sell_amount.mul(other.buy_amount)).cmp(&self.buy_amount.mul(other.sell_amount))
```
nlordell, line 93: ubernit: I would use [`full_mul`](https://docs.rs/web3/0.13.0/web3/types/struct.U256.html#method.full_mul) as it multiplies `U256 * U256 -> U512` and therefore can't overflow or panic.

src/models/orders.rs
josojo: Its coming from here: https://github.com/josojo/oba-contracts/blob/master/src/contracts/PreAMMBatcher.sol#L123
It's good practice to use them so that the orders valid for one domain/contract are not valid somewhere else.

nlordell: Cool, makes perfect sense. Learned something new :smile:

src/models/orders.rs
```diff
@@ -53,13 +45,9 @@ impl Order {
...
     pub fn validate_order(&self) -> Result<bool> {
         let message = self.get_digest()?;
-        let signature = Signature {
-            v: self.signature_v - 27 as u8,
-            r: *self.signature_r.as_fixed_bytes(),
-            s: *self.signature_s.as_fixed_bytes(),
-        };
-        let owner = signature.recover(&message)?;
-        return Ok(H160::from(owner.address()).eq(&self.owner));
+        let signature = encode(&[self.signature_r.into_token(), self.signature_s.into_token()]);
```
nlordell, line 48: `web3` provides a cool tool for building the signature for you <https://docs.rs/web3/0.13.0/web3/types/struct.Recovery.html>. You should be able to do:
```rust
let signature = Recovery::new(&message, self.signature_v, self.signature_r, self.signature_s).as_signature();
```

This way you don't have to worry about the details of encoding the r and s words, or converting the v-value into a recovery ID (from Electrum notation, or dealing with chain relay protection).

nlordell (approved):
Looks great!

PR #3, opened 2020-09-24
Adding mergify rules to repo

josojo:
This PR adds a Mergify rules to the oba-contracts so that we can use merge when green feature. It also automatically merges successful dependabot PRs.

Test Plan
 I will test it once dev-ops will have enabled it 😄

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,22 @@
+pull_request_rules:
+  - name: Merge approved and green PRs with `merge when green` label
+    conditions:
+      - "#approved-reviews-by>=1"
+      - status-success=test
+      - base=main
+      - label=merge when green
+    actions:
+      merge:
+        method: squash
+        strict: smart+fasttrack
+        commit_message: title+body
+  - name: Automatic merge for Dependabot pull requests
+    conditions:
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -0,0 +1,22 @@
+pull_request_rules:
+  - name: Merge approved and green PRs with `merge when green` label
+    conditions:
+      - "#approved-reviews-by>=1"
+      - status-success=test
```
nlordell, line 5: I checked the recent status checks and it looks like for Rust, it is called "build" for whatever arbitrary reason:

image

```suggestion
      - status-success=build
```

.mergify.yml
```diff
@@ -0,0 +1,22 @@
...
+      - label=merge when green
+    actions:
+      merge:
+        method: squash
+        strict: smart+fasttrack
+        commit_message: title+body
+  - name: Automatic merge for Dependabot pull requests
+    conditions:
+      - author~=^dependabot(|-preview)\[bot\]$
+      - status-success=test
```
nlordell, line 16:
```suggestion
      - status-success=build
```

nlordell (approved):

PR #4, opened 2020-09-24
Adding the orderbook model

josojo:
This PR prepares a new type orderbook model. It is wrapped into Arc and RwLock already, since eventually, we will need this orderbook to be shared between many async tasks (API and solution crafters).

testplan:
unit tests

### DIFF
--- a/src/models/orderbook.rs
+++ b/src/models/orderbook.rs
@@ -65,7 +65,7 @@ impl OrderBook {
         let mut current_orderbook = self.orders.write();
         let layer_hash_map = current_orderbook.entry(order.sell_token).or_default();
         let orders = layer_hash_map.entry(order.buy_token).or_default();
-        let search_result = orders.binary_search(&order.clone());
+        let search_result = orders.binary_search(&order);
         let pos = match search_result {
             Err(_) => return false, // order is not in orderbook
             Ok(e) => e,
--- a/src/models/orderbook.rs
+++ b/src/models/orderbook.rs
@@ -1,0 +1,128 @@
+use crate::models::Order;
+use anyhow::Result;
+use ethcontract::web3::types::Address;
+use parking_lot::RwLock;
+use serde::Deserialize;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+pub type OrderBookHashMap = HashMap<Address, HashMap<Address, Vec<Order>>>;
+
+#[derive(Clone, Deserialize)]
+pub struct OrderBook {
+    #[serde(with = "arc_rwlock_serde")]
+    pub orders: Arc<RwLock<OrderBookHashMap>>,
+}
+
+mod arc_rwlock_serde {
+    use parking_lot::RwLock;
+    use serde::de::Deserializer;
+    use serde::Deserialize;
+    use std::sync::Arc;
+
+    pub fn deserialize<'de, D, T>(d: D) -> Result<Arc<RwLock<T>>, D::Error>
+    where
+        D: Deserializer<'de>,
+        T: Deserialize<'de>,
+    {
+        Ok(Arc::new(RwLock::new(T::deserialize(d)?)))
+    }
+}
+
+impl OrderBook {
+    #[allow(dead_code)]
+    pub fn new() -> Self {
+        OrderBook {
+            orders: Arc::new(RwLock::new(HashMap::new())),
+        }
+    }
+    #[allow(dead_code)]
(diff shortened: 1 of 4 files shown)


src/models/orderbook.rs
```diff
@@ -0,0 +1,128 @@
+use crate::models::Order;
+use anyhow::Result;
+use ethcontract::web3::types::Address;
+use parking_lot::RwLock;
```
nlordell, line 4: This is actually a very performant RwLock implementation! Note that there is also one in `std::sync::RwLock`, but the `parking_lot` one is much better from what I've heard.

src/models/orderbook.rs
```diff
@@ -0,0 +1,128 @@
...
+    #[allow(dead_code)]
+    pub fn get_orders_for_tokens(self, token_1: Address, token_2: Address) -> Result<Vec<Order>> {
+        let current_orderbook = self.orders.read();
+        let empty_hash_vec: Vec<Order> = Vec::new();
+        let empty_hash_map: HashMap<Address, Vec<Order>> = HashMap::new();
+        let new_hash_map = current_orderbook.get(&token_1).unwrap_or(&empty_hash_map);
+        Ok(new_hash_map
+            .get(&token_2)
+            .unwrap_or(&empty_hash_vec)
+            .clone())
```
nlordell, line 61: Just a cool little tidbit (although it effectively makes no difference in this case since empty `Vec`s don't perform any allocations):
```suggestion
            .cloned()
            .unwrap_or(empty_hash_vec))
```

This clones the inner value if `get(...)` returns `Some(vec)`. This way `empty_hash_vec` isn't created, cloned, then discarded.

src/models/orderbook.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+    #[allow(dead_code)]
+    pub fn remove_order(&mut self, order: Order) -> bool {
+        let mut current_orderbook = self.orders.write();
+        let layer_hash_map = current_orderbook.entry(order.sell_token).or_default();
+        let orders = layer_hash_map.entry(order.buy_token).or_default();
+        let search_result = orders.binary_search(&order.clone());
```
nlordell, line 68: I think this `clone` might not be needed (since, from what I remember, `Address` implements `copy`).

nlordell (approved):
Looks good!

fleupold (commented):
Can we enable clippy on every PR to catch things like unnecessary clones and unused_mut? 

Regarding the use or `parking_lot::RwLock` over the std one: While it's cool to use a "better" implementation, it requires downloading and installing this dependency (e.g. on every CI build). I don't think the access speed to our Lock is going to be a bottleneck so I'd prefer using std lib unless there is a good reason for extra dependencies.

src/models/orderbook.rs
```diff
@@ -0,0 +1,128 @@
...
+
+impl OrderBook {
+    #[allow(dead_code)]
+    pub fn new() -> Self {
+        OrderBook {
+            orders: Arc::new(RwLock::new(HashMap::new())),
+        }
+    }
+    #[allow(dead_code)]
+    pub fn add_order(&mut self, order: Order) -> bool {
```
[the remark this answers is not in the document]
nlordell: Good catch!

src/models/orderbook.rs
```diff
@@ -0,0 +1,128 @@
...
+        let mut current_orderbook = self.orders.write();
+        let layer_hash_map = current_orderbook.entry(order.sell_token).or_default();
+        let orders = layer_hash_map.entry(order.buy_token).or_default();
+        let search_result = orders.binary_search(&order.clone());
```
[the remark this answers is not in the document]
nlordell: Also great catch! I don't even look for these anymore as I expect `clippy` would fail the build :rofl:

nlordell:
> Regarding the use or parking_lot::RwLock over the std one: While it's cool to use a "better" implementation, it requires downloading and installing this dependency (e.g. on every CI build). I don't think the access speed to our Lock is going to be a bottleneck so I'd prefer using std lib unless there is a good reason for extra dependencies.

While I agree with this statement, I also noticed that the `Cargo.lock` didn't add a new `parking_lot` lock entry as it was already included as a transitive dependency for another crate (`ethcontract` -> `web3`).

Now that I think about it, this `RWLock` will likely have to change to a `tokio::sync::RwLock` to avoid deadlock issues :see_no_evil:

PR #5, opened 2020-09-26
API for posting and viewing orders

josojo:
Closes: https://github.com/gnosis/dex-services/issues/1433

testplan:
```
cargo run
```
and in a new terminal
```
 curl --location --request POST 'localhost:3030/v1/orders' \
--header 'Content-Type: application/json' \
--header 'Content-Type: text/plain' \
--data-raw '{
    "sell_amount": "0xDE0B6B3A7640000",
    "buy_amount": "0xC7D713B49DA0000",
    "sell_token": "0xA193E42526F1FEA8C99AF609dcEabf30C1c29fAA",
    "buy_token": "0xFDFEF9D10d929cB3905C71400ce6be1990EA0F34",
    "owner": "0x63FC2aD3d021a4D7e64323529a55a9442C444dA0",
    "nonce": 1,
    "signature_v": 27,
    "signature_r": "0x07cf23fa6f588cc3a91de8444b589e5afbf91c5d486c512a353d45d02fa58700",
    "signature_s": "0x53671e75b62b5bd64f91c80430aafb002040c35d1fcf25d0dc55d978946d5c11",
    "valid_until": "0x0"
}'


```


src/api/filter.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+pub fn get(
+    orderbook: OrderBook,
+) -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
+    warp::get()
+        .and(warp::path("v1"))
+        .and(warp::path("orders"))
+        .and(warp::path::end())
+        .and(warp::path::end())
```
fleupold, line 37: Why double end?

src/api/api.rs
```diff
@@ -0,0 +1,10 @@
+use super::filter::get;
+use super::filter::post_order;
+use crate::models::OrderBook;
+use core::future::Future;
+use warp::Filter;
+
+pub fn run_api(orderbook: OrderBook) -> impl Future<Output = ()> + 'static {
```
fleupold, line 7: Could this live in src/api.rs directly?

src/api/api.rs
```diff
@@ -0,0 +1,10 @@
...
+    let routes = post_order(orderbook.clone()).or(get(orderbook.clone()));
```
fleupold, line 8: nit: last one is an unnecessary clone

src/api/filter.rs
```diff
@@ -0,0 +1,122 @@
...
+
+#[cfg(test)]
+pub mod test_util {
+    use super::*;
+    use crate::models::Order;
+    use ethcontract::web3::types::U256;
+    use warp::http::StatusCode;
+    use warp::test::request;
+
+    #[tokio::test]
```
fleupold, line 50: Instead of tokio tests, can we use `now_or_never` like in our price estimator [tests](https://github.com/gnosis/dex-services/blob/7c7237b52d1d5614d73935d84fb706fd40f4c3f6/price-estimator/src/filter.rs#L443-L450)?

src/api/filter.rs
```diff
@@ -0,0 +1,122 @@
+use super::handler::{add_order, get_orders};
+use crate::models::Order;
+use crate::models::OrderBook;
+use warp::Filter;
+
+fn json_body() -> impl Filter<Extract = (Order,), Error = warp::Rejection> + Clone {
+    // When accepting a body, we want a JSON body
+    // (and to reject huge payloads)...
+    warp::body::content_length_limit(1024 * 16).and(warp::body::json())
```
fleupold, line 9: nit: make this a constant somwhere?

src/api/filter.rs
```diff
@@ -0,0 +1,122 @@
+use super::handler::{add_order, get_orders};
+use crate::models::Order;
+use crate::models::OrderBook;
+use warp::Filter;
+
+fn json_body() -> impl Filter<Extract = (Order,), Error = warp::Rejection> + Clone {
+    // When accepting a body, we want a JSON body
```
fleupold, line 7: Just a personal opinion, I find this comment not really useful.

src/models/orders.rs
```diff
@@ -13,8 +13,8 @@ use std::cmp::PartialOrd;
 pub struct Order {
     pub sell_amount: U256,
     pub buy_amount: U256,
-    pub current_sell_amount: U256,
-    pub current_buy_amount: U256,
+    pub current_sell_amount: Option<U256>,
```
fleupold, line 16: Why is this optional now? Also is this remaining sell amount, or amount filled so far (maybe we can rename for clarification)?

fleupold (approved):
Looks good! I don't think we should need tokio for this though.

src/models/orders.rs
josojo: It is intended to be the sell_amount reduce by min(allowance(), balanceOf()). We need to store this information somewhere.

It is optional so that in the warp API call, we can leave these values out. They will only later be processed by our allowance and balanceOf readers.

josojo:
Reposting my personal message to @fleupold  for more transparency: 

I was trying to switch tokioo against async_std. Here is [my code:](https://github.com/gnosis/oba-services/compare/api...api-async-std?expand=1)

It compiles fine, but when I am running cargo run, I am getting the following error:
```
thread 'main' panicked at 'there is no reactor running, must be called from the context of Tokio runtime', /Users/alexherrmann/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-0.2.22/src/io/driver/mod.rs:201:9
```
somehow the warp lib wants to get started with tokio and not with async_std.
I found the following [hint here as well:](https://www.steadylearner.com/blog/read/How-to-use-Rust-Warp)
```
2. Warp uses tokio as its async task runner behind the scenes. You won't have to care for it much to make the example work except "its alternative is async std and they are not that compatible currently". 
```
Do you think that it is reasonable to stick to tokio under these given assumptions? I don’t know the libraries too well and can’t do a proper judgment

Also, I have seen that the [join!](https://docs.rs/async-std/1.6.5/async_std/future/trait.Future.html#method.join) commands are only available in the unstable variant.

src/models/orders.rs
fleupold: > It is optional so that in the warp API call, we can leave these values out

I personally think that this is a sign that we shouldn't use the exact same Order model for posting orders and for processing them (we are shoehorning one model to work in two places).

We could keep the raw Order struct for posting (without `current_sell_amount`) and have a MatchableOrder for solving, which consists of the raw Order and an extra field (available_sell_balance), which can be the min of balanceOf and allowance.

I understand why we need to look at the "available sell balance". However, why do we need to look at current_buy_amount?

josojo: > I personally think that this is a sign that we shouldn't use the exact same Order model for posting orders and for processing them (we are shoehorning one model to work in two places).

This is how I had it implemented before. I have also seen it like that in dex-services. But it introduces redundant code and types. It feels strange that if I am modifying the order model to introduce a new field, that I need to do it on several structs. Isn't it much easier to work in Options, in order to reduce the redundancy?

What is the advantage of introducing two order models?

>However, why do we need to look at current_buy_amount?
Yeah, current_buy_amount can always be calculated from current_sell_amount, sell_amount and buy_amount. I thought it would be easier to store this value than recalculating it. But, now that you are pointing it out, I will change it.

Though, most likely, we should also remove current_sell_amount, as a sell balance not easily be assigned directly to one order. If one user has two orders selling the same asset and the available balance is less then the sum of the sell_amounts, we need to deal with that. But I will do it in a later PR.

fleupold: > What is the advantage of introducing two order models?

The advantage is normalisation and scoping a model to each use case. It can become very hard to understand a data model where a lot of fields are optional, but implicitly they have a semantic (e.g. if current_sell_amount is set so is current_buy_amount). SAP's data model is a great example (if you have ever worked with it), where tables have up to 1000 columns (most of them nullable) and it requires expert knowledge to know which columns go with what others in what state of the system.

Such invariants can be enforced at compile time by using tightly scoped models. Converting between them is trivial code (which might seem redundant but has very little chance of error). Keeping track of different field semantic in different parts of the codebase 6 months down the line is much more prone to error.

I'm also not saying we should copy all fields, I think we can use composition where

```rust
struct MatchableOrder {
  raw_order: Order,
  available_sell_amount: U256
}
```

fleupold:
> Do you think that it is reasonable to stick to tokio under these given assumptions?

It looks like we are actually using the tokio runtime in the pricegraph (https://github.com/gnosis/dex-services/blob/7c7237b52d1d5614d73935d84fb706fd40f4c3f6/price-estimator/Cargo.toml#L21)

Maybe @nlordell recalls why we preferred async_std where possible. For the tests, I don't think you need to use await but can instead assert that everything returns `now_or_never` synchronously (since you shouldn't need to actually await on things in a unit test).

src/api/filter.rs
nlordell: Personally, I prefer `tokio::test`. It allows test code to use `async` directly which is very nice.

src/api/filter.rs
nlordell: To me, this is not something that we should a) hard code, or b) set in the driver code. To me this kind of setting belongs in our deployment ingress configuration (Nginx settings for example).

src/models/orders.rs
nlordell: I agree with @fleupold, The smaller the scope of a type, the better. If we use `Option<U256>` then we might need to introduce `None` handling code in places where it is impossible to be `None`, or introduce `unwrap`s which are brittle to refactoring. By encoding the exact "shape" of the data in its type, and using different types where the data has different "shapes" is a little nicer.

Note that `serde` has a `#[serde(flatten)]` feature, so the JSON could even look the same!

nlordell:
Warp requires `tokio`, which is why we use it in the `price-estimator` service.

> I don't think you need to use await but can instead assert that everything returns now_or_never synchronously 

This gets tricky with tokio timers, where, from what I recall, always yield. Personally, I really like using `tokio::test`. Asserting with `now_or_never` as @fleupold suggests can be useful when you want to assert that the future is expected to complete immediately, but this is not always what the test cares about.

> Maybe @nlordell recalls why we preferred async_std where possible.

You can take a look at the short discussion @e00E and I had about why we use it in the driver gnosis/dex-services#843. To sum up my understanding:
- `async-std` futures run on any executor (this is because a runtime is spawned automagically). At the time, we were using the `futures::executor` quite extensively, so `async-std` seemed like an easier choice to simplify integration.
- `async-std` is about providing an out of the box `async` version of the `std` library, (that being said, `tokio` is very complete and provides a very nice `async`-focused API, which in many cases is very similar to the `std` library).

They're both great projects with huge communities, that being said, crates like `web3` (and now `ethcontract` since updating to the latest `web3`) use `tokio` internally, so it makes more sense now to go with `tokio`.

josojo:
I am facing a weird issue:

Due to the fact that the new tokio::sync::rwlock is async, I can not implement a serialize function for the OrderBook struct for the [serde trait Serialize.](https://docs.serde.rs/serde/trait.Serialize.html), as the serialize function is required to be sync code. The trait needs a non-async serialize function.

This challenge is also explained [here:](https://users.rust-lang.org/t/how-to-serialize-deserialize-an-async-std-rwlock-t-where-t-serialize-deserialize/37407)
Unfortunately, there is no real solution proposed in this thread besides this very hacky way of just creating a new struct, which is not "underneath" a rwlock. ( as I pushed it in my latest commit).

Do you guys know about a better solution? @nlordell @fleupold

fleupold:
I don't think the proposed solution is very hacky. I found it a little weird to begin with that the Orderbook with the lock itself was serializable. In the future this class might grow and take more unserializable dependencies (e.g. a way to fetch balances or the like), thus having a raw unlocked plain old data model that is serializable seems good to me.

nlordell:
I also agree. I think de/serializable data should be plain data objects. Then types that control concurrent access would have `RwLock`s in order to share this data.

PR #6, opened 2020-09-27
Infinite loop for driver

josojo:
This PR implements the infintie loop for the driver.

There are many options on how to handle errors. Right now, I think it is the best to just print errors from the batch_process, and not interrupting the infinite loop. But maybe we have to overthink this option later.

testplan:

unit tests


src/main.rs
```diff
@@ -1,16 +1,41 @@
...
+        spawn(async move {
+            let res = batch_process(orderbook_for_iteration)
+                .await
+                .map_err(|e| format!(" {:?} while async call batch_process", e));
+            match res {
+                Err(e) => println!("{:}", e),
+                Ok(_) => (),
+            };
+        });
+        delay_for(Duration::from_secs(1)).await;
```
fleupold, line 39: Can we make this a constant please?

src/main.rs
```diff
@@ -1,16 +1,41 @@
...
+        e = handler_driver => {
+            println!("handler_driver returned {:?}", e);
         }
     }
 }
+
+async fn run_driver(orderbook: OrderBook) {
+    loop {
+        let orderbook_for_iteration = orderbook.clone();
+        spawn(async move {
```
fleupold, line 30: Why do you want to run this in a separate thread? Would you expect multiple run loops to be ongoing in parallel? Shouldn't one loop finish before we attempt another one?

Cargo.toml
```diff
@@ -14,4 +14,4 @@ anyhow = "1.0.32"
 serde_json = "1.0.57"
 parking_lot = "0.11.0"
 warp = "0.2"
-tokio = { version = "0.2.4", features = ["macros", "test-util"] }
\ No newline at end of file
+tokio = { version = "0.2.4", features = ["macros", "test-util"] }
```
fleupold, line 17: Do we really want to depend on tokio? We used `async-std` in dex-services and it worked fine. Tokio is a big hammer if I recall correctly. How come we already depend on it?

src/batcher/batcher.rs
```diff
@@ -0,0 +1,85 @@
...
+pub async fn batch_process(orderbook: OrderBook) -> Result<()> {
+    let tokens: Vec<Address> = orderbook
+        .orders
+        .read()
+        .keys()
+        .into_iter()
+        .map(|t| Address::from(t.as_fixed_bytes()))
+        .collect();
+
+    let token_pairs = get_token_pairs(&tokens);
```
fleupold, line 17: Would this method be better suited on the orderbook class directly?

src/batcher/batcher.rs
```diff
@@ -0,0 +1,85 @@
...
+    for token_pair in token_pairs {
+        let best_match = solve_pair(
+            orderbook
```
fleupold, line 20: ditto. Would be nice if we can just get all sell orders for a given token pair from the model directly.

src/main.rs
```diff
@@ -1,16 +1,41 @@
...
+    let handler_driver = run_driver(orderbook);
+
     select! {
-        err = handler_api => {
-            println!("run_api returned the following error {:?}", err);
+        e = handler_api => {
+            println!("run_api returned  {:?}", e);
+        },
+        e = handler_driver => {
+            println!("handler_driver returned {:?}", e);
```
fleupold, line 22: How can run_driver ever return?

src/main.rs
```diff
@@ -1,16 +1,41 @@
...
         }
     }
 }
+
+async fn run_driver(orderbook: OrderBook) {
```
fleupold, line 27: Can we mark this as `-> !` ?

src/main.rs
josojo: Would you expect multiple run loops to be ongoing in parallel?
Loops might not finish quickly, as we are waiting for transactions to be confirmed. The waiting for the mining will most likely block them quite a time. Hence, I think it makes sense to start them over, before they are officially ended.

I am using tokio::spawn and not thread::spawn, so it is really just another future.  But you are right that we might not even need tokio, I will try to rewrite this PR with it.

fleupold: You are right that tx confirmation might take some time. How do you envision the "starting over"? Would we try to settle the same orders as in the currently pending settlement (outbidding the old tx with the same nonce). 
Or would we optimistically assume the previous tx went through and resend a new settlement with a higher nonce?

josojo: I think that settling the same orders does not make too much sense. Here, I would prefer a good "tx retry mechanism", as we have for the submit_solution already.

Yes, I would propose that the old/pending async task marks the orders used in their submission and gives the new async task the chance to recognize already used orders, and hence preventing double submission.

src/batcher/batcher.rs
```diff
@@ -0,0 +1,85 @@
...
+                .unwrap_or(&HashMap::new())
+                .get(&token_pair.0)
+                .unwrap_or(&Vec::new()),
+        )?;
+        submit_solution(best_match)?;
+    }
+    Ok(())
+}
+
+fn get_token_pairs(tokens: &Vec<Address>) -> Vec<(Address, Address)> {
```
nlordell, line 40: A slightly more idiomatic way would be to use iterators for this. Something like:
```rust
let len = tokens.len()
(0..len).flat_map(|i| (i..len).map(move |j| (tokens[i], tokens[j])))
```

This way, you can return an iterator (instead of collecting into a `Vec`):
```suggestion
fn get_token_pairs(tokens: &Vec<Address>) -> impl Iterator<Item = (Address, Address)> + '_ {
```

Also, in general, passing `Vec` references as function parameters [is considered unidomatic](https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg):

```suggestion
fn get_token_pairs(tokens: &[Address]) -> impl Iterator<Item = (Address, Address)> + '_ {
```

josojo:
@e00E . It would be great if you could review this PR. Maybe you will not find too many issues, and then we can merge it. 
The other outstanding PRs need some more adaptions, I guess. If you find the time, feel free to get them ready. I will not have the chance to work on them this week.

src/batcher/batcher.rs
josojo: I would say no: Maybe we don't want to search on all tokens from the orderbook, but rather a subset. Hence, I see it as an independent method.

e00E (approved):
Having this code as scaffolding looks good to me. I expect it will change a lot when we get this ready to actually run (and I created some issues for discussion) but it's still nice to have some more structure right now.

PR #7, opened 2020-09-28
Naive solver

josojo:
Implements the naive solver for e2e tests. Most of this code was actually ported from the old naive solver and rewritten by @twalth3r  here: https://github.com/gnosis/oba-services/pull/1

Testplan:
unit tests


src/solver/naive_solver.rs
```diff
@@ -0,0 +1,208 @@
...
+
+pub fn solve_pair(sell_orders_token0: &Vec<Order>, sell_orders_token1: &Vec<Order>) -> Solution {
+    assert!(check_orders_sorted_by_limit_price(&sell_orders_token0));
+    assert!(check_orders_sorted_by_limit_price(&sell_orders_token1));
+
+    // Match orders with best limit prices, if possible.
+    if let (Some(best_sell_order_token0), Some(best_sell_order_token1)) =
+        (sell_orders_token0.last(), sell_orders_token1.last())
+    {
+        if best_sell_order_token0.have_price_overlap(&best_sell_order_token1) {
```
nlordell, line 27: @josojo, I think I remember you asking how to avoid having this nested `if` statement. There is now a new cool way that was introduced in more recent Rust versions:
```rust
if matches!(
    (sell_orders_token0.last(), sell_orders_token1.last()),
    (Some(best_sell_order_token0), Some(best_sell_order_token1))
    if best_sell_order_token0.have_price_overlap(&best_sell_order_token1)
) {
    // ...
}
```

Although its not very useful here (since you need the values wrapped by the `Option`), just letting you know that this really cool feature was added :smile:.

One other possibility would be to use a `match` statement, this way the compiler guarantees that all paths return the same type:
```rust
match (sell_orders_token0.last(), sell_orders_token1.last()) {
    (Some(best_sell_order_token0), Some(best_sell_order_token1))
        if best_sell_order_token0.have_price_overlap(&best_sell_order_token1) =>
    {
        Solution {
            sell_orders_token0: vec![best_sell_order_token0.clone()],
            sell_orders_token1: vec![best_sell_order_token1.clone()],
        }
    }
    _ => Solution::trivial(),
}
```

src/solver/naive_solver.rs
```diff
@@ -0,0 +1,208 @@
...
+            valid_until: U256::from("0"),
+        }
+    }
+    fn address(address_str: &str) -> Address {
+        return (address_str).parse().unwrap();
+    }
+
+    fn signature(signature_str: &str) -> H256 {
+        return (signature_str).parse().unwrap();
+    }
```
nlordell, line 111: Maybe these could go in a top-level `test` module that only gets included for tests and can be shared across the project. It can even contain a "prelude" for commonly used types during tests:

- `lib.rs`
```rust
#[cfg(test)]
mod test;
```

- `test.rs`
```rust
pub fn address(address_str: &str) -> Address {
    return (address_str).parse().unwrap();
}

pub fn signature(signature_str: &str) -> H256 {
    return (signature_str).parse().unwrap();
}

pub mod prelude {
    pub use super::*;
    pub use ethcontract::{Address, H256};
}
```

Then for unit tests:
```rust
#[cfg(test)]
mod tests {
    use crate::test::prelude::*;

    #[test]
    fn some_test() {
        // ...
    }
}
```

src/models/orders.rs
```diff
@@ -26,6 +26,20 @@ pub struct Order {
 }
 
 impl Order {
+    pub fn get_current_sell_amount(&self) -> U256 {
+        if let Some(current_sell_amount) = self.current_sell_amount {
+            current_sell_amount
+        } else {
+            self.sell_amount
+        }
+    }
```
nlordell, line 35: ubernit: Could you add some spacing to the methods, I find its harder to read without it:
```suggestion
    }

```

nlordell (approved):
Very nice!

josojo:
Valentin implemented a better naive solver ;)

PR #9, opened 2020-09-28
Moving parts into workspaces

josojo:
Starting to work with workspaces, since I want to create contracts, ci, e2e workspaces, similar to dex-services

testplan:
unit tests


fleupold (approved):
Midterm, we should probably also separate the warp and the solving part of the system into different binaries, so that they can run on different hosts with different availability guarantees.

nlordell:
> Midterm, we should probably also separate the warp and the solving part of the system into different binaries, so that they can run on different hosts with different availability guarantees.

Also agree - I think the service that collects orders should be its own crate/process/container.

Cargo.toml
```diff
@@ -1,17 +1,7 @@
...
-parking_lot = "0.11.0"
-warp = "0.2"
-tokio = { version = "0.2.4", features = ["macros", "test-util"] }
+[workspace]
+members = [
+    "driver",
+]
+default-members = [
+    "driver",
+]
```
nlordell, line 7: `default-members` is not needed here, by default all `members` are included :smile:

nlordell (approved):

josojo:
Closing, as it is no longer relevant

PR #10, opened 2020-09-29
Feedback from PR Adding the orderbook model #4

josojo:
@fleupold had additional feedback in the #4 PR.

This PR implements the feedback.
I am unwrapping the read() and write() functions of the new sync:Rwlock, as there can only be errors, once a panic occurred in any other write-operation, and in this case, we will interrupt our program anyways ( the other panic would have stopped the code anyways). Let me know whether this is good practice.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -2,21 +2,23 @@ name: Rust
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
-    branches: [ main ]
+    branches: [main]
 
 env:
   CARGO_TERM_COLOR: always
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
-    - name: Build
-      run: cargo build --verbose
-    - name: Run tests
-      run: cargo test --verbose
+      - uses: actions/checkout@v2
+      - name: Formatting checks
+        run: cargo fmt --all -- --check
+          - cargo clippy --all-targets --all-features -- -D warnings
+      - name: Build
+        run: cargo build --verbose
+      - name: Run tests
+        run: cargo test --verbose
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,9 +15,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: Formatting checks
+      - name: Fmt checks
         run: cargo fmt --all -- --check
-          - cargo clippy --all-targets --all-features -- -D warnings
+      - name: Clippy checks
+        run: cargo clippy --all-targets --all-features -- -D warnings
       - name: Build
         run: cargo build --verbose
       - name: Run tests
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Fmt checks
         run: cargo fmt --all -- --check
       - name: Clippy checks
-        run: cargo clippy --all-targets --all-features -- -D warnings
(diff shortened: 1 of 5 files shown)


src/models/orderbook.rs
```diff
@@ -38,31 +38,27 @@ impl OrderBook {
     }
     #[allow(dead_code)]
     pub fn add_order(&mut self, order: Order) -> bool {
-        let mut current_orderbook = self.orders.write();
+        let mut current_orderbook = self.orders.write().unwrap();
```
fleupold, line 41: Maybe short comment why unwrapping here (and in the callsites below) is fine.

src/models/orderbook.rs
```diff
@@ -38,31 +38,27 @@ impl OrderBook {
     }
     #[allow(dead_code)]
     pub fn add_order(&mut self, order: Order) -> bool {
```
fleupold, line 40: I don't think `add_order` requires `&mut` either.

fleupold (approved):

.github/workflows/rust.yml
```diff
@@ -2,21 +2,24 @@ name: Rust
...
-    - uses: actions/checkout@v2
-    - name: Build
-      run: cargo build --verbose
-    - name: Run tests
-      run: cargo test --verbose
+      - uses: actions/checkout@v2
+      - name: Fmt checks
+        run: cargo fmt --all -- --check
+      - name: Clippy checks
+        run: cargo clippy --all-targets --all-features -- -D warnings
```
nlordell, line 21:
```suggestion
        run: cargo clippy --workspace --all-targets --all-features -- -D warnings
```

Cargo.toml
```diff
@@ -11,5 +11,4 @@ ethcontract = "0.8.0"
 serde = { version = "1.0", features = ["derive"] }
 rustc-hex = "2.1.0"
 anyhow = "1.0.32"
-serde_json = "1.0.57"
-parking_lot = "0.11.0"
\ No newline at end of file
+serde_json = "1.0.57"
```
nlordell, line 14: nit: missing newline at the end of file.

src/models/orderbook.rs
```diff
@@ -1,10 +1,10 @@
 use crate::models::Order;
 use anyhow::Result;
 use ethcontract::web3::types::Address;
-use parking_lot::RwLock;
 use serde::Deserialize;
 use std::collections::HashMap;
 use std::sync::Arc;
+use std::sync::RwLock;
```
nlordell, line 7: I think this needs to be `tokio::sync::RwLock` so that it can be used in an `async` context. Sorry I missed that in the first review.

src/models/orderbook.rs
nlordell: In general, prefer `expect` over `unwrap` in these cases:
```suggestion
        let mut current_orderbook = self.orders.write().expect("rw lock can never be poisoned");
```

src/models/orderbook.rs
```diff
@@ -38,31 +38,27 @@ impl OrderBook {
...
-            .get(&token_2)
-            .cloned()
-            .unwrap_or(empty_hash_vec))
+        Ok(new_hash_map.get(&token_2).cloned().unwrap_or_default())
     }
     #[allow(dead_code)]
-    pub fn remove_order(&mut self, order: Order) -> bool {
-        let mut current_orderbook = self.orders.write();
+    pub fn remove_order(&self, order: Order) -> bool {
+        let mut current_orderbook = self.orders.write().unwrap();
```
nlordell, line 61: Same here

nlordell (approved):

src/models/orderbook.rs
josojo: Thanks for catching it now. I would have had to correct just in the next PR 🙄

PR #13, opened 2020-10-08
Use hex macro instead of FromHex

e00E:
This is better because it happens at compile time and is easier to read.

### Test Plan
The changed test passes.


josojo (approved):
nice :)

PR #16, opened 2020-10-08
Used grouped imports

e00E:
Because we generally prefer this style.

### Test Plan
CI

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -1,9 +1,9 @@
 pub mod filter;
 pub mod handler;
+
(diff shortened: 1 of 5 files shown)


josojo (approved):
thanks. Why doesn't fmt or clippy complains about them in the first place?

Can we enforce it somehow?

e00E:
There is probably not enough consensus about which format to prefer so there is no lint against. I don't think we need to enforce it either.

PR #17, opened 2020-10-08
Remove Result return type from function that cannot fail

e00E:
I didn't see a reason for `Result` here so I removed it.

### Test Plan
CI


josojo (approved):
thanks

PR #19, opened 2020-10-08
Mark handler error type as infallible

e00E:
to indiciate that it cannot actually fail. We still want a Result
because that is what warp expects for `and_then`.

### Test Plan
CI


josojo (approved):
cleaner for sure!

PR #29, opened 2020-10-20
Split into separate crates and binaries

e00E:
We know we want to have this eventually because the orderbook will run
indendent of the solver. This changes the structure of the code a lot so
it makes sense to do it early when it is less work.
The plan is that solver communicates with the orderbook through an http
api.
In the orderbook crate we continue to have the `struct Orderbook` that
manages the orderbook while in the solver crate we will probably have a
`trait Orderbook` that allows reading the orderbook.

Some solver code got removed because it has to be rewritten without using the orderbook crate's orderbook.

### Test Plan
CI


bh2smith (approved):
Looks good!

josojo:
Personally, I would do the split of orderbook server and driver server much later to keep the complexity low at the beginning. it is just getting us so much overhead for the whole setup - all the communication, deployment setup etc -, if we split the two servers from the start.

But, yeah. I am curious about the opinion of others...

e00E:
If we are going to split eventually then we should do it now. Otherwise we will make too many choices that depend on not being split and gloss over things that will be hard to change later when we do split. I'm afraid we would end up not noticing problems with the current oba approach early enough.

josojo:
>If we are going to split eventually then we should do it now.
Eventually, yes. But do we need to do it even before the launch of the first MVP, where we kinda assume 1-operator setup anyways? I am not sure. Do you see any necessity for it, assuming there is only one operator?

Do you think it would be wise, due to load distribution etc?

e00E:
Even with only 1 operator there is still some value into splitting the binaries. Often it is a good idea to separate concerns. In this case the http orderbook server for the frontend is different enough from the solver that it intuitively it feels like the correct choice to separate their binaries too. For example if there is a memory leak in the solver code it would not bring down the orderbook. Or once the orderbook is feature complete we can run it in production but still have the solver in staging because we are updating it often. Or we could have an emergency solver running that checks if the main solver has been down for 2 minutes and then takes over.

josojo (approved):
alright, then let's go for it.

josojo:
Currently, the oba spec becomes more complicated each day. It really gives me creeps, I think it does not need to be like that. 
I agree to all the arguments for a split of solver and order-collector, but complexity also has a huge cost. 

@fleupold could you also give your opinion here?

fleupold:
I agree that we are fighting with scope creep on a functional level, but for the operational side of the solver, I don't think we can have a production ready setup where solver and order receiving endpoint run in the same process (wouldn't allow us to run multiple solvers e.g., seems easy to DOS).

Therefore I agree that splitting the two services is important. Maybe we can make a way to run it in the same process (but communication still happens via some well defined potentially http interface). But this could make local testing easier.

PR #30, opened 2020-10-21
Add openapi sketch for orderbook

e00E:
This is the public api that is used by the frontend, potential other
operators, the solver.

I picked human readable encodings. token amounts are encoded as strings
because of common json library shortcomings for large integers.

You can preview the generated api documentation at https://editor.swagger.io/ .

### DIFF
--- a/openapi.yml
+++ b/openapi.yml
@@ -18,7 +18,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/Order"
+              $ref: "#/components/schemas/OrderNew"
   /api/v1/orders:
     get:
       summary: Get existing orders.
@@ -31,7 +31,7 @@ paths:
               schema:
                 type: array
                 items:
-                  $ref: "#/components/schemas/Order"
+                  $ref: "#/components/schemas/OrderFull"
   /api/v1/nonce:
     get:
       summary: The nonce to pick for a new order.
@@ -65,7 +65,8 @@ components:
     Nonce:
       description: Nonce of the batch in which the order is valid. uint32.
       type: integer
-    Order:
+    OrderNew:
+      description: Data a user provides when creating a new order.
       type: object
       properties:
         buyToken:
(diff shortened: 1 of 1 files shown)


openapi.yml
```diff
@@ -0,0 +1,105 @@
...
+      type: string
+      example: "0x6810e776880c02933d47db1b9fc05908e5386b96"
+    TokenAmount:
+      description: Amount of a token. uint112 encoded in decimal.
+      type: string
+      example: "1234567890"
+    Nonce:
+      description: Nonce of the batch in which the order is valid. uint32.
+      type: integer
+    Order:
```
e00E, line 68: An open question to me is whether the order should also contain a `createdAt` field. The spec says we should settle orders the latest when they are 1 minute old so we need to have this information somewhere.

1. We could let the user (really the frontend) supply this value.
2. We could make the orderbook choose the value and return it when getting existing orders.
3. We could ignore this in the orderbook and make the solver keep track of it.

1 has the disadvantage that the value can be tempered with like setting it in the past so that the order appears older than it really is.
3 makes the orderbook slightly simpler but seems very finicky in the solver code because we would need to keep track of the first time we see a specific order from the orderbook.
2 seems like the best choice.

josojo: Yes, I fully agree. 2 is the way to go

openapi.yml
```diff
@@ -0,0 +1,105 @@
...
+      # TODO: Later we could add filters through the url query to only get orders matching specific tokens etc.
+      responses:
+        200:
+          description: existing orders
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Order"
```
josojo, line 34: Should we leave this out for now, as we don't know where it will be stored?

josojo (approved):
Looks good. Great start.

openapi.yml
e00E: Do you mean how we will persist it on disk? I don't think that's relevant for the api because the orderbook service will be an abstraction over the underlying storage. So the solver only needs to know how to communicate with the orderbook hopefully.

openapi.yml
```diff
@@ -0,0 +1,105 @@
...
+          type: string
+          example: "0x00"
+        signature_r:
+          description: 32 bytes encoded as hex.
+          type: string
+          example: "0x0000000000000000000000000000000000000000000000000000000000000000"
+        signature_s:
+          description: 32 bytes encoded as hex.
+          type: string
+          example: "0x0000000000000000000000000000000000000000000000000000000000000000"
```
fleupold, line 105: Might be simpler to combine this to one value, or do you see a use case where I might be interested in one of the three values?

One thing to note is that this model right now doesn't carry the owner (which is fine as they can be derived from the signature). For convenience it might be nice though to expose a owner address at least for people querying orders. Wdyt?

openapi.yml
```diff
@@ -0,0 +1,105 @@
...
+        description: The order to create.
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Order"
+  /api/v1/orders:
+    get:
+      summary: Get existing orders.
+      # TODO: Later we could add filters through the url query to only get orders matching specific tokens etc.
```
fleupold, line 25: Where should the "validation logic" (e.g. checking sufficient approval etc) live? Would we allow to filter this as well?

fleupold (approved):
Looks good to me...

openapi.yml
e00E: I wasn't a hundred percent sure how the signature is used so I took the same names as the spec. If it is always used as the combined bytes then we should combine this into one 65 byte value.

We can treat the owner like the createdAt timestamp. When creating the order it isn't specified but the orderbook will automatically populate it.

openapi.yml
e00E: I was thinking that the orderbook is going to have some automatic cleanup of stale orders. For example, it can always remove (and reject) orders with a nonce or timestamp in the past so there is no need for a filter there.

For the approval balance it is more difficult because they can change quickly and can start out as invalid but become valid. That looks like something better suited for the solver but not sure yet.

PR #32, opened 2020-10-23
Implement serializable order type

e00E:
Implements the order type as described by the specification with
serialization as described by the openapi documentation.
This is in its own crate because we want to share this module between the
orderbook and the solver.

### Test Plan
Unit test for the serialization.


openapi.yml
```diff
@@ -55,9 +55,9 @@ paths:
 components:
   schemas:
     Address:
-      description: Ethereum 40 byte address encoded as a hex.
+      description: Ethereum 40 byte address encoded as a hex without `0x` prefix.
```
e00E, line 58: I've changed the api here because the 0x seems more annoying than useful. Most hex libraries do not use so I imagine this is easier for the frontend and the backend.

order/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+    pub fn serialize<S>(value: &H160, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut bytes = [0u8; 20 * 2];
+        // Can only fail if the buffer size does not match but we know it is correct.
+        hex::encode_to_slice(value, &mut bytes).unwrap();
+        // Hex encoding is always valid utf8.
+        let s = std::str::from_utf8(&bytes).unwrap();
+        serializer.serialize_str(s)
```
fleupold, line 129: nit: it might be slightly simpler to use the [LowerHex implementation of H160](https://tomusdrw.github.io/rust-web3/web3/types/struct.H160.html#impl-LowerHex)

```rust
serializer.serialize_str(format!("{:x}", value))
```

order/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct Full {
+    pub creation_time: DateTime<Utc>,
+    #[serde(with = "h160_hex")]
+    pub owner: H160,
+    #[serde(flatten)]
+    pub user_provided: UserProvided,
+}
+
+impl Serialize for Signature {
```
fleupold, line 65: Maybe signature could be its own model?

fleupold (approved):
Rather than an "order" crate, should this be more a "model" or "core" crate where more of the things that will be shared between the web service and driver can go? Or are we planning to have a lot of small crates (I remember dealing with many different crates had some challenges).

order/src/lib.rs
```diff
@@ -0,0 +1,214 @@
+//! Contains the order type as described by the specification with serialization as described by the openapi documentation.
+//!
+//! This is in its own crate because we want to share this module between the orderbook and the solver.
+
+use chrono::{offset::Utc, DateTime};
+use primitive_types::H160;
```
josojo, line 6: Should we use the H160 from ethcontracts-rs? I think this would be better, as later on we will heavily rely on eth-contract and converting between  primitive_types::H160 and ethcontract::H160 might become a hussle.

order/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+use serde::{de, Deserialize, Serialize};
+use std::fmt;
+
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub enum FillType {
+    FillKill,
```
josojo, line 13: No strong opinion, but I would call it `FillOrKill`

josojo (approved):
I agree with Felix, a crate `model` sounds better.

order/src/lib.rs
e00E: We want `hex` anyway for the u8 slices because those implement LowerHex and then I like using it for the address too.  The current implementation also doesn't allocate (but same can be achieved with `write!` instead of `format!`).

order/src/lib.rs
e00E: It's the same type so it won't be a problem.

order/src/lib.rs
e00E: What do you mean by it's own model? It is already it's own struct.

order/src/lib.rs
e00E: Will rename it here and also in the openapi (where it was `fillkill`).

e00E:
I've renamed the crate to models.

I'm not sure yet if we will have a `core` crate like in dex-services. We will have to see how it feels during development.

order/src/lib.rs
fleupold: sorry, I meant module and this its own file.

e00E: We will see how the file grows. Not yet needed.

PR #33, opened 2020-10-27
Update orderbook to openapi

e00E:
Implements order creation and querying from the openapi specification.
Moves or deletes come code that became obsolete or will be reimplemented in the future. While we're moving fast I don't think we need to do small separate PRs for everything.

### Test Plan
Updated unit tests but haven't tested with openapi yet. Will add more testing later.


orderbook/Cargo.toml
```diff
@@ -4,10 +4,12 @@ version = "0.1.0"
 edition = "2018"
 
 [dependencies]
-anyhow = "1.0.33"
-ethcontract = { version = "0.8.0", default-features = false }
-hex-literal = "0.3.1"
+anyhow = "1.0"
```
e00E, line 7: Here I've unified the version number to always be major and minor without patch.

orderbook/src/api/filter.rs
```diff
@@ -1,116 +1,81 @@
...
+fn extract_user_order() -> impl Filter<Extract = (UserOrder,), Error = warp::Rejection> + Clone {
+    // (rejecting huge payloads)...
+    warp::body::content_length_limit(MAX_JSON_BODY_PAYLOAD).and(warp::body::json())
+}
+
+pub fn create_order(
+    orderbook: Arc<OrderBook>,
 ) -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
-    warp::path!("v1" / "orders")
+    warp::path!("api" / "v1" / "orders")
```
fleupold, line 23: ubernit: Maybe the comon part of the path can be filtered out

orderbook/src/orderbook.rs
```diff
@@ -0,0 +1,89 @@
...
+
+#[derive(Debug)]
+pub enum RemoveOrderError {
+    DoesNotExist,
+}
+
+#[derive(Debug, Default)]
+pub struct OrderBook {
+    // TODO: Store more efficiently (for example HashMap) depending on functionality we need.
+    pub orders: RwLock<Vec<Order>>,
```
fleupold, line 23: nit: Even for the current functionality (duplicate check, remove) a HashSet might be more appropriate

orderbook/src/orderbook.rs
```diff
@@ -0,0 +1,89 @@
...
+            orders.swap_remove(index);
+            Ok(())
+        } else {
+            Err(RemoveOrderError::DoesNotExist)
+        }
+    }
+}
+
+/// None if signature doesn't match.
+fn user_order_to_full_order(user_order: UserOrder) -> Option<Order> {
```
fleupold, line 55: nit: I find having this return None when the signature doesn't match a bit implicit (it requires reading the comment). Maybe we can recover the signtature in a separate method (where it's clear that None means invalid singature) and then have this method take the owner as an argument?

fleupold (approved):
Nice! Would be great to test this in the wild before merging...

e00E:
I changed user_order_to_full_order to make the error clear.
For the orders Vector/HashSet we will definitely change this in the future but in this PR I don't want too think too hard about the correct type and vector is the simplest.

orderbook/src/api/handler.rs
```diff
@@ -1,31 +1,23 @@
...
-            ))
-        }
-    }
+pub async fn add_order(
+    orderbook: Arc<OrderBook>,
+    order: UserOrder,
+) -> Result<impl warp::Reply, Infallible> {
+    let (body, status_code) = match orderbook.add_order(order).await {
+        Ok(()) => ("ok", StatusCode::CREATED),
+        Err(AddOrderError::AlreadyExists) => ("already exists", StatusCode::BAD_REQUEST),
```
josojo, line 12: much cleaner 👍

josojo (approved):
Very clean PR. nice.

I am also okay to go with bigger PRs at the beginning ;)

PR #35, opened 2020-10-28
Query orderbook api from solver

e00E:
This is a first simple implementation of querying the http orderbook api from the solver.
I am not sure yet if this should be a mockable trait so this commit is just the struct.
Later we are going to test this with the orderbook api implementation but I want more structure in the code first.

### DIFF
--- a/solver/Cargo.toml
+++ b/solver/Cargo.toml
@@ -4,6 +4,8 @@ version = "0.1.0"
 edition = "2018"
 
 [dependencies]
-anyhow = "1.0.33"
-ethcontract = { version = "0.8.0", default-features = false }
-tokio = { version = "0.2.4", features =[ "macros"] }
+anyhow = "1.0"
+model = { path = "../model" }
+primitive-types = "0.7"
+reqwest = { version = "0.10", features = ["json"] }
+tokio = { version = "0.2", features =[ "macros"] }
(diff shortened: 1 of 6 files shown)


solver/Cargo.toml
```diff
@@ -4,6 +4,8 @@ version = "0.1.0"
 edition = "2018"
 
 [dependencies]
-anyhow = "1.0.33"
-ethcontract = { version = "0.8.0", default-features = false }
```
e00E, line 8: ethcontract removed until we need it. I've changed the Address/H160 to primitive-types because I find it cleaner to use it directly.

fleupold (approved):
I'm pretty sure this struct will have to be mockable for testing the inner works of the solving/filtering logic.

solver/src/batcher.rs
```diff
@@ -1,6 +1,5 @@
-use ethcontract::web3::types::Address;
```
josojo, line 1: I think that it is not good to have a mixture of `ethcontract::web3::types` and `primitive_types::H160 as Address;` types, as this add the complexity of conversion.

I would stick to `ethcontract::web3::types` since we will need ethcontract anyways for all the smart contract integrations.
Wdyt?

e00E: As far as I know there is no conversion. They are the same types just exported on different paths. If we had to do a conversion I would agree with you. But without it it is cleaner to use the minimal dependency. For example the `model` crate uses `H160` but it does not need to know about web3 or ethcontract.

PR #39, opened 2020-11-03
Outline settlement code

e00E:
This is the code structure I came up with for the settlement code. Mostly about how we know when we should settle and how to deal with the nonces.
I was trying to keep find_settlement free of any IO related traits but not sure how to best deal with getting the AMM price.

I like manually calling `update` in 5-10 seconds intervals instead of more complicated scheduling and allowing another layer on top of the driver to decide how to pass orders into it. driver is maybe not the best name, but not sure what to call it.

### DIFF
--- a/solver/src/ethereum.rs
+++ b/solver/src/ethereum.rs
@@ -1,0 +1,20 @@
+use crate::settlement::Settlement;
+use anyhow::Result;
+use async_trait::async_trait;
+use model::TokenPair;
+use primitive_types::H160;
+
+// Not final at all. Just want some prototypes available to use in other code.
+
+#[async_trait]
+pub trait SettlementContract {
+    async fn get_nonce(&self, token_pair: TokenPair) -> Result<u32>;
+    async fn settle(&self, settlement: Settlement) -> Result<()>;
+}
+
+#[async_trait]
+pub trait Ethereum {
+    async fn get_erc20_balance(&self, user: H160, token: H160) -> Result<u128>;
+    async fn is_spending_approved(&self, user: H160, token: H160, spender: H160) -> Result<bool>;
+    async fn get_uniswap_price(&self, token_pair: TokenPair) -> Result<f64>;
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,6 +13,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cff77d8686867eceff3105329d4698d96c2391c176d5d03adc90c7389162b5b8"
 
 [[package]]
+name = "async-trait"
+version = "0.1.41"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b246867b8b3b6ae56035f1eb1ed557c1d8eae97f0d53696138a50fa0e3a3b8c0"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "autocfg"
 version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1275,6 +1286,7 @@ name = "solver"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "async-trait",
  "model",
  "primitive-types",
  "reqwest",
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -61,6 +61,12 @@ pub struct UserOrder {
     pub signature: Signature,
 }
 
+impl UserOrder {
+    pub fn token_pair(&self) -> Option<TokenPair> {
+        TokenPair::new(self.buy_token, self.sell_token)
+    }
+}
+
 /// An order that is returned when querying the orderbook.
 ///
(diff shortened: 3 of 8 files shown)


solver/src/ethereum.rs
```diff
@@ -0,0 +1,20 @@
...
+// Not final at all. Just want some prototypes available to use in other code.
+
+#[async_trait]
+pub trait SettlementContract {
+    async fn get_nonce(&self, token_pair: TokenPair) -> Result<u32>;
+    async fn settle(&self, settlement: Settlement) -> Result<()>;
+}
+
+#[async_trait]
+pub trait Ethereum {
```
josojo, line 16: one of the very difficult tasks will be to decide whether the driver should just query regularly an etherum-node about the erc20 balances and allowances or whether we need to subscribe to events about erc20 updates.

This decision is important for sure, no matter which solution nonce solution we will implement. Maybe it makes sense to investigate the best solution for this, assuming we have 20, 100, 300 orders per batch, before writing nonce-related code. Just and idea :)

josojo:
For this PR, we need to make 2 important decisions:

1. Should we implement the code under the assumption of having only 1 operator?

I think you are assuming that we have only one operator, as not all nonces are updated, just the missing one. But what happens, if another solver settles in the meantime with the current nonce? Then we should also check for updates of the already existing nonces. 
My approach would be to select all tokenpairs with valid orders and then make one node request to ask for all nonces at the same time.

2. Should we `update` nonces, erc20 balances, erc20 allowances every 5 secs or should we get updated via listening to events?

I think it would be really nice to do some investigations about the performance with 50, 100, 500 orders in the system and then make a good decision. I think that updating is sufficiently good, but I doon't know. 
I think this is a nice task, that can be done independently from all the outstanding design decisions :)

e00E:
> I think you are assuming that we have only one operator

Mostly correct. I thought for the initial proof of concept we would make this assumption. In this code the none is only queried again if one of your settlements fails which it would if someone else increased the nonce. If we assume this happens often then yes we should have a smarter approach.

> My approach would be to select all tokenpairs with valid orders and then make one node request to ask for all nonces at the same time.

I'm assuming there is a function like "get_current_nonce(token_pair)" on the smart contract. Then yes we can set up a batch request like that. Feasibility probably depends on how many token pairs there are.

> Should we update nonces, erc20 balances, erc20 allowances every 5 secs or should we get updated via listening to events?

For the nonces we can do that (while handling optimistically increasing the nonce while waiting for your settlement transaction). For erc20 I am not sure how feasible that is. The big problem is that not all tokens implement the erc20 standard correctly so we can't rely on the events being emitted or even the internal logic making sense (like that the balance actually matches what we get from events). If we whitelisted only well behaving tokens that would be ok (and I think we should whitelist anyway). But maybe even common tokens do not implement the standard well enough.
The other problem is that there are many erc20 tokens and i imagine some of them have a very large number of events. Is it feasible to keep track of all events for all of the. And what if we a see an order for a new token for the first time? We would have to go through all historical events within 1 minute. I don't think it is feasible.
And even if we assume this does work then an order can still become invalid because of approval or balance after we submitted our settlement transaction but before it has been mined. For both approaches it is impossible to be sure that our settlement goes through but querying and caching validity is easier to implement.

josojo:
>If we assume this happens often then yes we should have a smarter approach.

My best guess is that we will have several solvers - most due to redundancy reasons. I think the redundancy is needed, as some "solving" logics might fail and we want to try out new methods.


>I'm assuming there is a function like "get_current_nonce(token_pair)" on the smart contract. Then yes we can set up a batch request like that. Feasibility probably depends on how many token pairs there are.

Let's keep that in mind. But maybe we will go for one globale nonce anways.

>For the nonces we can do that (while handling optimistically increasing the nonce while waiting for your settlement transaction). For erc20 I am not sure how feasible that is. The big problem is that not all tokens implement the erc20 standard correctly so we can't rely on the events being emitted or even the internal logic making sense (like that the balance actually matches what we get from events). If we whitelisted only well behaving tokens that would be ok (and I think we should whitelist anyway). But maybe even common tokens do not implement the standard well enough.
The other problem is that there are many erc20 tokens and i imagine some of them have a very large number of events. Is it feasible to keep track of all events for all of the. And what if we a see an order for a new token for the first time? We would have to go through all historical events within 1 minute. I don't think it is feasible.
And even if we assume this does work then an order can still become invalid because of approval or balance after we submitted our settlement transaction but before it has been mined. For both approaches it is impossible to be sure that our settlement goes through but querying and caching validity is easier to implement.

Yeah, I agree with many points. But I think the main criteria should be the speed of getting the information before starting the calculation. Probably, we will not have batches with more than 50 offline-orders and probably querying 50 balances and allowances is pretty quick. Maybe it is getting problematic if we also consider ~100 smart contract orders, but they will only be considered later anyways, I guess. So yeah, if querying 50 balances and allowances is quick, I am okay with going that route, sure.

bh2smith (approved):
Looks like a descent outline. Would have expected to see some `unimplemented!` but very cool.

e00E:
I'm merging this so I can make more progress but ofc can revert and change in the future again.

PR #47, opened 2020-11-23
Use travis based on Felix' template

e00E:
instead of github actions. For long term I'm still not sure which one is
better but we use travis in our other projects so it makes sense to keep
this configuration where we already know how to build docker images, run
ganache, etc.

### Test Plan
Check that travis picks this up and works.


fleupold (approved):
Looks good to me.

nlordell (approved):

josojo (approved):

bh2smith (approved):

PR #48, opened 2020-11-23
Add contracts module

e00E:
This is mostly copied from dex-services with some parts removed that we
don't (yet) need. Contains the vendor script and IERC20.

### Test Plan
CI builds. And tested locally that I see generated docs for IERC20.


contracts/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+    "ethcontract-generate",
+    "log",
+    "serde_json",
+]
+
+[dependencies]
+ethcontract = { version = "0.9",  default-features = false }
+serde = "1.0"
+
+# [bin-dependencies]
```
fleupold, line 25: Should this be uncommented?

contracts/README.md
```diff
@@ -0,0 +1,13 @@
+# contracts
+
+The library part of this crate contains `ethcontract` generated bindings to the smart contracts we
+use.
+
+# `vendor` Script
```
fleupold, line 6: Can you add the steps required to add a new contract (adding it in build.rs to expose it and vendor.rs to actually build it). I had some issues with it when I added the SolutionSubmission contract to dex-services (should have probably added it there myself).

fleupold (approved):

contracts/Cargo.toml
nlordell: Its not an official `Cargo.toml` as far as I'm aware. I *think* its just for delimiting the dependencies that are only added when enabling the `bin` feature for building the `vendor` script.

contracts/build.rs
```diff
@@ -0,0 +1,41 @@
...
+    let mut builder = Builder::new(artifact)
+        .with_visibility_modifier(Some("pub"))
+        .add_event_derive("serde::Deserialize")
+        .add_event_derive("serde::Serialize");
+
+    if let Ok(address) = fs::read_to_string(&address_file) {
+        println!("cargo:rerun-if-changed={}", address_file.display());
+        builder = builder.add_deployment_str(5777, address.trim());
+    }
+
```
nlordell, line 35: this might not be needed since we don't deploy in Rust. But it might be worth leaving it in if the eventually happens.

contracts/src/paths.rs
```diff
@@ -0,0 +1,17 @@
...
+use std::path::{Path, PathBuf};
+
+/// Path to file containing address of a contract deployed to a test network.
+pub fn contract_address_file(name: &str) -> PathBuf {
+    Path::new(env!("CARGO_MANIFEST_DIR"))
+        .join("..")
+        .join("target")
+        .join("deploy")
+        .join(format!("{}.addr", name))
+}
```
nlordell, line 12: Same as :point_up:

nlordell (approved):

contracts/Cargo.toml
e00E: This crate has a lib target and a bin target. Cargo currently does not support specifying dependencies for those individually so the workaround is to make the dependencies optional and collect them in a feature.

contracts/build.rs
e00E: True. I will remove it until we need it at which point we can copy it back from dex-services.

contracts/README.md
e00E: added

PR #49, opened 2020-11-24
1st draft of the api specification

josojo:
Laying out a specification for the api between front-end and backend.

Any input is welcomed. This is not an implementation PR, but rather a specification PR.

Let's find a common aggreement


openapi.yml
```diff
@@ -12,6 +12,14 @@ paths:
...
         201:
           description: Order has been accepted.
+        400:
+          description: Gas-tip is not sufficient
+        401:
+          description: Unauthorized, your account is deny-listed
+        409:
+          description: Order exists already
+        429:
+          description: Too many order placements
```
e00E, line 22: These status codes have fairly specific meaning for the http standard but I don't think they make sense to distinguish application level errors. Maybe 409 and 429 do make sense but I don't think this has to be part of the api guarantees.

We should specify in this in the api only for errors the frontend wants to be able to distinguish and react to. Instead of status codes it is probably better to introduce something like an error enum or json which we return as the body for 400 errors.

josojo: >These status codes have fairly specific meaning for the http standard but I don't think they make sense to distinguish application level errors. Maybe 409 and 429 do make sense but I don't think this has to be part of the api guarantees.

you are right, I read more about it and agree!

> json which we return as the body for 400 errors.

Yeah, I like it.

>the frontend wants to be able to distinguish and react to.
I also agree! I thought that the might want to react to all of them. Let's wait for Anxo's response.

openapi.yml
```diff
@@ -45,26 +65,43 @@ paths:
...
       type: string
       example: "1234567890"
-    Nonce:
-      description: Nonce of the batch in which the order is valid. uint32.
+    Tip:
+      description: The tip that is used to pay solvers. uint256.
       type: integer
+    Flags:
+          description: Integer encodes the information for buy/sell order and fillOrKill/standard order uint8 encoded in decimal.
+          type: integer
```
e00E, line 104: Why store the raw flags over the previous enum? To me this is internal to the encoding of the smart contract but for this api it makes more sense to keep the enums.

josojo: I thought the parsing into enums is one additional "unrequited" step before verifying the signature. I order to create the digest, we need the flag-int anyways.

But probably, you are right that the interface is cleaner with orderType and fillType. Let me change it back.

openapi.yml
nlordell: I would be careful not to conflate HTTP errors with API errors. That being said, I think the error codes are in general well chosen. Some proposed changes:
- 
```suggestion
        400:
          description: Order is malformed
        403:
          description: Forbidden, your account is deny-listed
        429:
          description: Too many order placements
        500:
          description: Error adding an order
```

So, the reason I would suggest a `403` instead of a `401` is that the former is intended to indicate that a certain authenticated user (authenticated via signature in this case) does not have sufficient access (i.e. is not authorized), while the latter indicates that authentication is required but missing (which is not the case since a signature is a valid authentication). From the [MDN page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403):
> This status is similar to 401, but in this case, re-authenticating will make no difference. The access is permanently forbidden and tied to the application logic, such as insufficient rights to a resource.

> Maybe 409 and 429 do make sense but I don't think this has to be part of the api guarantees.

I like 429 as this is generally used for rate limiting and is a pretty standard code for this scenario.

> return as the body for 400 errors.

I would reserve 400 for malformed input (such as the order JSON is missing properties or the signature is invalid) and use 500 errors for general errors adding orders (duplicate entries, missing balance, etc.)

In fact, to make it easy for the FE to differentiate errors, we can always return a
```rust
struct Error {
  code: ErrorKind,
  message: String
}
#[repr(i32)]
enum ErrorKind {
  MissingOrderData,
  InvalidSignature,
  DuplicateOrder,
  InsufficientFunds,
}
```

Which would serialize to something like:
```json
{
  "code": 2, // ErrorKind::DuplicateOrder
  "message": "Order exists already"  
}
```

openapi.yml
nlordell: > but for this api it makes more sense to keep the enums

I also agree, especially considering how the signed data uses an `enum` and `bool` and the `u8` flags is just used for encoding for gas efficiency.

openapi.yml
e00E: Not wanting to argue over which http error fits better is one of the reasons I don't want to use http errors and prefer the return body enum. We know which errors we have so let's represent those in the most obvious way possible like in your json struct and not think about the http status code as much. For me it is enough to say: 400 = couldn't understand (deserialize) request, 500 = any other error. And then for 500 we return this json response to describe the error.

I would make the error enum be represented as a struct with an openapi enum (which is json string) and and optional message that has more detail. No need to convert the enum to a number first.

nlordell: > And then for 500 we return this json response to describe the error.

I would return it for both `500` and `400` errors. There are things like invalid signatures that are orthogonal to deserialization but would, IMO constitute a "Bad Request" which would be nice to get a more descriptive error on.

> I would make the error enum be represented as a struct with an openapi enum (which is json string)

Yeah, that is much better.

anxolin: I agree with most of everything said in this thead

I like the proposal of error codes of Nick, I think is not only useful for the front end, you'll find it useful for Prometheus stats for example. 

Anyways, the way I would define this errors is, by setting an generic structure for errors. So everything that is not a `200` should have at least a "description" (string) and a "type" (enum). 

Normally, most errors would suffice with that data. But in some cases, you might want to add additional data.
Not required for us, but maybe convenient for you, to include the `stacktrace` in the error `500`. Although not sure strictly pretty, just for saving you the time to go and check the logs and match it with a certain time.

nlordell: > Normally, most errors would suffice with that data

Structured data for certain error types is a really nice idea! 

> Not required for us, but maybe convenient for you, to include the stacktrace in the error 500

My only concern with that might accidentally leaking private information (for example something part of the configuration).

nlordell: > Not wanting to argue over which http error fits better is one of the reasons

One thing worth mentioning is that certain HTTP errors do affect browser behaviour. For example a 401 can trigger a login prompt.

openapi.yml
```diff
@@ -32,10 +52,10 @@ paths:
...
                   $ref: "#/components/schemas/OrderFull"
-  /api/v1/nonce:
+  /api/v1/tip:
     get:
-      summary: The nonce to pick for a new order.
-      description: The best nonce to pick depends on how much time there is left in the current settlement according to the operator. This time is token-pair specific which is why the sell and buy tokens must be supplied.
+      summary: The tip that is used to pay solvers.
+      description: The tip is calculated by the order service and depends on the current gas price and the expected gas costs to settle the trade
       parameters:
         - name: sellToken
```
anxolin, line 60: Why does the tip requires a buy token? In my thinking, It should be only be relative to the sell token. Basically a client would ask "If sell GNO how much GNO you need as a TIP"? 

If there's a decision to take, like there's a minimal sell amount, we could model in the return data. "The GNO TIP is 0.1 GNO, but only for trades selling less than 10 GNO". Or any other logic.
So I wouldn't pass even the amounts.

I would suggest:
`/api/v1/tokens/`{sellToken}`/required-tip`

anxolin: Also, I think the operator should return also a date together with the tip (`expirationDate`). So the clients know until when the operator would allow to receive orders for that sell token with that minimal tip.

anxolin (commented):
Sorry for so many comments @josojo , but this is the first review I do from the API. 

I also need to do another small closer round with our specs in hand.
Overall looks great, good job there!! mostly I included nitpicks.

openapi.yml
```diff
@@ -32,10 +52,10 @@ paths:
                 type: array
                 items:
                   $ref: "#/components/schemas/OrderFull"
```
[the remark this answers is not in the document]
josojo: We removed the nonce, we will only have an appData, which can be used as a salt/nonce.
But usually, signed orders should not be the same from the front-end as their validTo will differ unless they are magically signed at the same time. Hence, we don't need a nonce.

openapi.yml
anxolin: > My only concern is that it might accidentally leaking private information (for example something from the configuration).

Fair enough. I've seen many times settling a flag just for this, in dev environments. There could be also a leak but not so important.

> One thing worth mentioning is that certain HTTP errors do affect browser behaviour. For example a 401 can trigger a login prompt.

mm, I've never saw it just by sending a 401, I think normally it's triggered by a special header the serve sends, signaling they want a basic authentication. Since the browsers knows how to do it, they just go for it. 
Anyway, it was a good call you changing the 401 for a 403. One is authentication, the other authorization.

openapi.yml
```diff
@@ -22,7 +30,19 @@ paths:
   /api/v1/orders:
     get:
       summary: Get existing orders.
-      # TODO: Later we could add filters through the url query to only get orders matching specific tokens etc.
+      parameters:
+        - name: owner
+          in: query
+          schema:
+            $ref: "#/components/schemas/Address"
+        - name: sellToken
```
anxolin, line 38: For anything that is in the URL, I wouldn't use `camelCase`, I would use kebab case

openapi.yml
josojo: Yeah I was thinking that we might want to subsidize differently, depending on the tokenpair and ordertype. But you are right, it is much easier in your suggestion and we should go for it.

openapi.yml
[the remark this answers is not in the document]
anxolin: Sorry, this part of the review I did it looking at another branch 🙈

Anyways. Interesting topic regardless. I thought the nonce was also for the use case:

- I sell 1M ether for a ridiculously cheap price (I messed up)
- I realise my mistake, I'll cancel my tx onchain, I don't trust the operator
- I'll use a nonce bigger than my last order, so the contract refuses it

Case two:
- same as case one, but I mess up with my expiration date. I don't want to keep my order active for a long time, because it can be exploited

I see how this might be just a trust issue. But is just that we usually try to provide onchain guaranteess

openapi.yml
```diff
@@ -12,6 +12,14 @@ paths:
       responses:
         201:
```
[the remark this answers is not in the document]
josojo: Yeah, if this would help the front-end, we could return the orderdigest as UUID. Sure

openapi.yml
```diff
@@ -45,26 +65,44 @@ paths:
...
-    Nonce:
-      description: Nonce of the batch in which the order is valid. uint32.
+    Tip:
+      description: The tip that is used to pay solvers. uint256.
       type: integer
+    OrderType:
+      description: Is this a buy order or sell order?
+      type: string
+      enum: [buy, sell]
     OrderNew:
```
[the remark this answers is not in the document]
josojo: Yeah, or another alternative would be `OrderCore` + `OrderEnrichement` = `Order`?

openapi.yml
```diff
@@ -45,26 +65,44 @@ paths:
...
       description: Data a user provides when creating a new order.
       type: object
```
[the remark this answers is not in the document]
josojo: I would leave that one out since we are not sure that we really want this, as far as I know...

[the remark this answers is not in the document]
anxolin: Ohh, I thought that after the research from BD we agreed this was in fact important
We can at least make it into the API, even if it's not the first implemented feature

openapi.yml
anxolin: Okay cool! one big reason would be, that if we use the amounts, we would need to ask on every amount change, and the UX could feel a bit sluggish also we would be making many queries to the backend that could be avoided.

openapi.yml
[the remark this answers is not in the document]
Velenir: If we get a UUID per order|orders, it would be super helpful to also have an endpoint to check on the status of an order.
Like `api/v1/order/<uuid> or even api/v1/order?uuid=[...ids]`
Or do you feel `/api/v1/orders?owner=..&sellToken=..&buyToken=..` already serves that purpose

Easily checking for individual order status by uuid would be nicer to use when a user has many orders historically but only a few pending.

openapi.yml
```diff
@@ -77,18 +115,17 @@ components:
...
+          $ref: "#/components/schemas/OrderType"
         fillType:
           description: Is this a fill-or-kill order a partially fillable order?
           type: string
           enum: [fillorkill, partial]
-        nonce:
-          $ref: "#/components/schemas/Nonce"
+        appData:
+          description: Identifier to make order unique or pass in arbitrary data. uint32 encoded in decimal.
+          type: integer
```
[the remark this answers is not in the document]
Velenir: @anxolin 
`validTo` has to be JSON stringifiable, so either number (for epoch) or string
Do you mean something like `JSON.stringify(new Date()) > ""2020-11-25T07:01:35.939Z""`?

openapi.yml
```diff
@@ -32,65 +51,74 @@ paths:
...
-      description: Nonce of the batch in which the order is valid. uint32.
-      type: integer
+    Tip:
+      description: The tip that is used to pay solvers. uint256.
+      type: object
+      properties:
+        expirationDate:
+          description: Expiration date of the offered tip. Encoded as ISO 8601 UTC.
+          type: string
+          example: "2020-10-22T07:57:18Z"
```
Velenir, line 84: I think it would be better to have `Tip::expirationDate` here and `OrderNew::validTo` be the same type. Either both epoch number or Date stringified

openapi.yml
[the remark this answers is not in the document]
e00E: I don't think orders have a specific uuid. It's just the hash and you already know this in the frontend because the user had to sign the order before submitting.

For the status it's not something i've thought about before. It depends on how the contract works and what the orderbook service is responsible for. Initially I thought it was really only there to collect and distribute orders and wouldn't even need to talk to an ethereum node, in which case there is no status: an order either exists or doesn't. I'm guessing an additional status would be whether it got matched but I'm not sure whether this is something that the orderbook service should be responsible for.

openapi.yml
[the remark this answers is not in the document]
e00E: There must be two types because only some information is needed when creating an order but more is returned when querying. I don't have a strong preference on what they are called.

openapi.yml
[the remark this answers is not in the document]
e00E: It could be a date too, but this is very specific to how it is interpreted in the smart contract which is the number of seconds since some epoch and it is part of the signed data so we have to make sure that any date conversions don't lose information.

openapi.yml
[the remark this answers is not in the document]
anxolin: I think it's a good idea, eventually, to allow the API be as close as possible as other exchanges, and be the only requirement for a CLI/UI to trade in GPv2

Usually all exchanges allow this, and I think would make integrations way easier. I believe this could help a lot adoption, and integrations with tools. 

So, regarding the state, sure they can review all settlements and try to identify the orders there and guess the state. It will get complicated when the settlement failed to execute the order specially if here's re-submissions.
So I find incredibly useful for a client to get an ID back and being able to ask for just the state of it. in summary I would focus in this product in making it look SIMPLER first. 
In this direction, one initial proposal for the state transitions (focusing on the most important use cases):


I understand a random UUID cannot be used, because this would limit in the future having a pool of operators you could use. It needs to be deterministic, but I would make the clients agnostic to how you generate the identity of the order. A hash of it's content sounds good to me.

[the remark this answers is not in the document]
e00E: What you describe sounds useful but should be a separate service because it does many more things than this orderbook part of the system. I would like this to really only be the exact functionality it was intended for which is receiving raw orders and distributing them to solvers.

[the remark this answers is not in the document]
anxolin: I think the service could be different, if you believe is fundamentally different responsibilities. I can also see that. Maybe even an implementation of this service, could just be either a custom indexer or the graph. 

However, I think that regardless of the logical separations in the modules, we should provide a consistent simple API, just for making it simpler for others.

openapi.yml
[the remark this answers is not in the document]
anxolin: For me was a nit, so go with a name you feel comfortable. This ones won't affect the endpoint itself :) 
My reasoning is, when they are just parameters for creating an entity, I just use a convention of "Creation" or "CreationParams", because "New" means "recent" (it could even be mistaken for order modification). With creation or creationParams I would like to suggest is the data required to create an order.

That's my reasoning, but go with what you feel comfortable.

openapi.yml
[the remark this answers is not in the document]
anxolin: Yes, I understand. But why a ISO date would lose information?
It would make it easier for user to submit using their time zone, or UTC if they prefer so. 

Otherwise, I feel we expose implementation details.

openapi.yml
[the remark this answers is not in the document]
josojo: The order service will be a single point of failure - at least it will not be decentralized in the beginning - and hence, I would prefer to keep it as simple as possible. 
Yes, spinning up another service for the order status, will definitively be needed. Let's discuss this server later in another PR.

Yes, the UUID would be the hash=orderdigest. I agree with Valentin that the front-end should be already aware of it.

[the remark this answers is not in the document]
anxolin: > The order service will be a single point of failure - at least it will not be decentralized in the beginning - and hence, I would prefer to keep it as simple as possible.

Sure, I think I'm happy with the separation. I think it's fine to be independent and under the same URL

> Yes, the UUID would be the hash=orderdigest. I agree with Valentin that the front-end should be already aware of it.

What do you mean with this? You referring to aware as generating himself, or aware that he receives the ID from the creation endpoint. I think for the base use case, it's better to just return it so u don't need to explain anything else to someone integrating. On top of that, yes, it's nice that our JS library also has a function that returns the ID for an order, but a basic integration wouldn't use that imo.

openapi.yml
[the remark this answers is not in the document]
josojo: >which is the number of seconds since some epoch

It will even be the seconds since the unix time counting. Federico checked that even in this case a uint32 would be sufficient for many upcoming years. ( I updated that descriptions)

Hence, I think it is fine to transfer the validTo information by a commonly used unix timestamp and it is not to specific to our implementation

openapi.yml
```diff
@@ -77,18 +115,17 @@ components:
...
           $ref: "#/components/schemas/TokenAmount"
         orderType:
-          description: Is this a buy order or sell order?
-          type: string
-          enum: [buy, sell]
+          $ref: "#/components/schemas/OrderType"
         fillType:
           description: Is this a fill-or-kill order a partially fillable order?
           type: string
           enum: [fillorkill, partial]
```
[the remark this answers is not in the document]
josojo: In this official [documentation](https://swagger.io/docs/specification/data-models/enums/), they also used lower case letters. I don't have a strong opinion...

openapi.yml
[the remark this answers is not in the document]
nlordell: Personally, I also like @anxolin's suggestion from an API perspective - it makes sense for the service to convert this timestamp to whatever representation it needs.

That being said, because we already require a user signature for the order, it already needs to be converted to the appropriate integer format (Unix timestamp, or seconds from Unix epoch). So with that in mind maybe it makes most sense to just use the `validTo` as an integer so that the data being signed is the same as the data being submitted with the order?

openapi.yml
josojo: The official [documentation](https://swagger.io/docs/specification/describing-parameters/) uses also camelCase: usderId instead of user_id

openapi.yml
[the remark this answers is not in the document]
nlordell: Currently the plan is for orders to be cancelled by order digest. That means, you can submit a `cancelOrder(orderDigest, owner, validTo)` call to the settlement contract which will flag the order as cancelled. In that regard, there is no more `nonce` and it is not needed to cancel orders on-chain.

Note that using the `orderDigest` means that if you have two identical orders, there is no way for GP to distinguish between the two. This is what @josojo was referring to WRT to using the new `appData` as salt, so that identical orders can be differentiated.

@anxolin I hope this clarifies things a bit.

openapi.yml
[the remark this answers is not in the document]
nlordell: In my experience URL parameters are typically lowercase as @josojo pointed out. I also prefer them as they they currently are (`fillorkill` and `partial`)

openapi.yml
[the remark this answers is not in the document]
josojo: One will still be able to cancel orders by their orderdigest, if required.

We will have a function that allows users to cancel their orders by[ modifying this data](https://github.com/gnosis/oba-contracts/blob/ndimSettlement/src/contracts/GPv2Settlement.sol#L37)

openapi.yml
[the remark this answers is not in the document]
nlordell: > What do you mean by this?

https://github.com/gnosis/oba-contracts/blob/7adfd42f6a5e0735b318cb2dd3928198656bd384/src/ts/index.ts#L311-L322

Orders have a well-defined hash (EIP-712) that uniquely identifies them. I think what @e00E and @josojo are saying is that this needs to be computed in the FE anyway as this hash is used for signing the order.

One small detail, however, is that this hashing is typically done by the signing wallet with EIP-712. However, I do think `web3` libraries typically implement EIP-712 struct hashes.

> the UUID would be the hash=orderdigest

One issue is that this is not unique for all orders, but rather unique for all orders of a given user. In that sense an order UUID is: `(ownerAddress, orderDigest)`. Alternatively we can use the `orderKey` as the UUID which is `keccak256(orderDitest, ownerAddress, validTo)`.

openapi.yml
```diff
@@ -32,65 +51,74 @@ paths:
...
+          description: Amount of tip in specified sellToken
+          type: integer
+    OrderType:
+      description: Is this a buy order or sell order?
+      type: string
+      enum: [buy, sell]
     OrderNew:
       description: Data a user provides when creating a new order.
       type: object
       properties:
```
nlordell, line 95: ubernit: We should probably order the order properties to match the order in which they are hashed, that is:
```
export const ORDER_TYPE_FIELDS = [
  { name: "sellToken", type: "address" },
  { name: "buyToken", type: "address" },
  { name: "sellAmount", type: "uint256" },
  { name: "buyAmount", type: "uint256" },
  { name: "validTo", type: "uint32" },
  { name: "appData", type: "uint32" },
  { name: "tip", type: "uint256" },
  { name: "kind", type: "uint8" },
  { name: "partiallyFillable", type: "bool" },
];
```

openapi.yml
nlordell: > mm, I've never saw it just by sending a 401, I think normally it's triggered by a special header the serve

:see_no_evil: Yes, it needs a `WWW-Authenticate` header.

openapi.yml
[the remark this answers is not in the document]
nlordell: I agree that this should be called `Order` to match the smart contract and spec. I think the order with additional signature/user data should be called `SignedOrder`.

openapi.yml
e00E: This is totally subjective. I don't *mind* either but I *prefer* kebab_case.

nlordell: Regarding `kebab-case` for URL query paramaters, I feel that is awkward as they are treated as separate words even though they are a single identifier (although I agree that URL **paths** should use `kebab-case`).

openapi.yml
nlordell, line 84: How is this intended to work with the settlement contract? The tip is signed with the order parameters and currently doesn't have a separate expiration date from the actual order.

josojo: No, it is not meant to have any impact on smart contracts.

Anxo asked for it with the following intend:
>Also, I think the operator should return also a date together with the tip (expirationDate). So the clients know until when the operator would allow receiving orders for that sell token with that minimal tip.

If I understand @anxolin correctly, then they would need it in the following scenario: The user asks to sends an order and the FE asks for the tip. Then we would have one "confirm module" and the signing process. If the user then takes too long, longer than the specified expiration time to confirm the order, the order service will not accept the order. The front-end can prevent this case if it is aware of an expiration time.

Alternatively, we could also assume that users don't need longer than 3 minutes for the confirmation module and the signing process and the order service just allows the tip that it has proposed within the last 3 minutes for this particular order.  Then, in the same scenario: "the user takes longer than 3 minutes", the front-end could also check the tip again right at the moment the end of the signing process and warn the user, if the new required tip is higher than the old one. THen the FE would have to restart with a new confirm module and a higher tip.

nlordell: Oh, its not part of the `Order` type :see_no_evil:.

Makes perfect sense! Sorry!

openapi.yml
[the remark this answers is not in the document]
josojo: Yes, right. Using the `orderKey` as UUID would be better than orderDigest. If the endpoint should return it, we can do it. If FE wants to compute it itself, this is also fine for me. Both ways are reasonable IMO.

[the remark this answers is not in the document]
anxolin: Cool, thanks for clarifying. Yep, I think then we should aim for both :) 

> One small detail, however, is that this hashing is typically done by the signing wallet with EIP-712. However, I do think web3 libraries typically implement EIP-712 struct hashes.

Also, there would be potentially sth weird with if you send the same order, with no expiration date in two different moments in time

I think, it would be good to include a `validFrom`, right? (not just for this, but also)

openapi.yml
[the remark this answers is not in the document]
anxolin: I'm happy anyways, the objects won't affect the endpoint signature. It sounded good to me what u first said of calling what the user `POST` a `SignedOrder`

openapi.yml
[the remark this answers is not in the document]
anxolin: okok, yes, makes sense, since for GET endpoints you won't want to use lower
the underscore part? I just felt was more readable, but also not super relevant

openapi.yml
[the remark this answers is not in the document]
anxolin: 👌 That makes total sense. it's a very low level endpoint, so it's good to send what you sign

openapi.yml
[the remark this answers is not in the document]
anxolin: > Currently the plan is for orders to be cancelled by order digest. That means, you can submit a cancelOrder(orderDigest, owner, validTo) call to the settlement contract which will flag the order as cancelled. In that regard, there is no more nonce and it is not needed to cancel orders on-chain.

That's way better than a nonce! u can cancel only what u need.

Will we be able to cancel multiple orders at once?

openapi.yml
anxolin: Yep, you convinced me. the resource path `kebap` and the params is ok in `camel`

openapi.yml
anxolin: Yes, we won't be querying the tip all the time. Only when you first select the token.
then depending on your commitment (`expireDate`) we won't be asking again in a while. 

`IMPORTANT`: for when you implement this, if u have some kind of dynamic calculation, please keep in mind that we should do roundings in the backend for the tip, so we keep it somewhat stable. This is because, if it changes while the user is confirming the order, it will ask him confirmation if he accepts the new price. For this reason, I think we shouldn't be changing on tiny price fluctuations. Rounding up to 1$?

openapi.yml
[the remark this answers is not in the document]
josojo: We don't have a `validFrom` in the spec anymore. To our understanding, this feature was not used much besides the IDO usecase and it would consume further gas.

>Yep, I think then we should aim for both :)

Okay, then I will change the API to return the orderKey

openapi.yml
```diff
@@ -32,71 +55,85 @@ paths:
...
+      type: object
+      properties:
+        expirationDate:
+          description: |
+            Expiration date of the offered tip. Order service might not accept 
+            the tip after this expiration date. Encoded as unix timestamp.
+          type: integer
+        tipAmount:
+          description: Amount of tip in specified sellToken
+          type: integer
```
josojo, line 92: This should be a string

openapi.yml
```diff
@@ -32,71 +55,85 @@ paths:
...
+            the tip after this expiration date. Encoded as unix timestamp.
+          type: integer
+        tipAmount:
+          description: Amount of tip in specified sellToken
+          $ref: "#/components/schemas/TokenAmount"
+    OrderType:
+      description: Is this a buy order or sell order?
+      type: string
+      enum: [buy, sell]
+    OrderCreation:
```
Velenir, line 97: There's no `owner` in `OrderCreation`. Why and can we add it?

openapi.yml
```diff
@@ -32,71 +55,85 @@ paths:
...
+        orderType:
+          description: "The type is either a buy or sell order"
+          $ref: "#/components/schemas/OrderType"
+        partiallyFillable:
+          description: Is this a fill-or-kill order or a partially fillable order?
+          type: boolean
         signature:
           description: 65 bytes encoded as hex without `0x` prefix. v + r + s from the spec.
           example: "0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
     OrderExtra:
```
Velenir, line 133: Can `OrderExtra` also include `UUID`, so that we get complete orders it when querying `/api/v1/orders`

openapi.yml
```diff
@@ -107,5 +144,18 @@ components:
           $ref: "#/components/schemas/Address"
     OrderFull:
       allOf:
-        - $ref: "#/components/schemas/OrderNew"
+        - $ref: "#/components/schemas/OrderCreation"
         - $ref: "#/components/schemas/OrderExtra"
+    UUID:
+     description: |
+      Unique identifier for the order, calculated by keccak256(orderDigest, ownerAddress, validTo),
```
fleupold, line 151: In an ideal world, I think UUIDs should be opaque (as in it has no semantics to the user) and as such the clients wouldn't care how it gets calculated.

If we want to overload it with semantics, can we instead of hashing orderDigest and ownerAddress together, concatenate them with`_` or the like so that a person looking at the UUID can still recover the individual parts? Or is there value in making it 32 bytes exactly? 

Given that the orderDigest is needed for cancellation on the smart contract, it would be nice to have a way to request it without needing to do local js keccak magicery.

openapi.yml
fleupold: The owner is implicitly provided by (and extracted from) the signature. It is part of OrderFull, which is the struct that should be used for querying information.

openapi.yml
```diff
@@ -107,5 +144,18 @@ components:
           $ref: "#/components/schemas/Address"
     OrderFull:
       allOf:
-        - $ref: "#/components/schemas/OrderNew"
+        - $ref: "#/components/schemas/OrderCreation"
```
fleupold, line 147: I know naming was discussed previously.

I'd slightly prefer the union type to be called Order (as I assume this is what most API clients will interact with by default and having the `Full` postfix seems a bit unspecific and verbose) and OrderMetadata over OrderExtra (which to me seems like an extra order type). Not sure if those names will actually show up in the API or codebase if not it's probably not worth fighting over them. 

I really like OrderCreation for the part that contains all submission information.

openapi.yml
Velenir: Signature is good to have,but it would be really nice to be able to immediately recognize and owner from the Order. Unless the signature itself is `<owner>_<uuid>` or something

nlordell: Since `OrderCreation` is just used as POST data for adding a new orders, I'm not sure I understand the value in requiring API clients to also post the `owner` if they already have to compute a ECDSA signature with a wallet that surely can also compute its public address.

> Signature is good to have,but it would be really nice to be able to immediately recognize the owner from the Order.

As @fleupold mentioned queried orders do contain an `owner` field representing the public address of the creator of the order. It is computed by recovering the public key of the private key used to sompute the ECDSA signature :point_up: 

> Unless the signature itself is <owner>_<uuid> or something

I think this is the ECDSA signature, so 65 bytes (`v || r || s`).

Velenir: > the value in requiring API clients to also post the owner
> queried orders do contain an owner field

Not for posting to the endpoint, but more for querying for all orders from all owners, to later be able to filter by owner address or some such

fleupold: > but more for querying for all orders from all owners

When querying, you will always get an instance of `OrderFull` which contains owner information, creation timestamp and other things that the `OrderCreation` payload doesn't expose.

nlordell: In that case, it should already be there. The type of a queried order is `OrderFull` which has `owner: Address` field:

https://github.com/gnosis/oba-services/blob/deacab85a1e9f797ada0e579e3f2fb82e8953a6d/openapi.yml#L133-L144

Velenir: Cool, somehow I got confused which `Order*` is returned where

openapi.yml
josojo: Yeah, orderDigest and ownerAddress is also fine with me. I don't think there is a value in having it 32bytes exactly. If there is no resistance, I am happy to change it.

nlordell: Maybe it makes sense for it to be `orderDigest || ownerAddress || validTo` as 56-byte array as discussed https://github.com/gnosis/oba-contracts/pull/195#discussion_r533639378 so that It can be pasted directly into Etherscan?

openapi.yml
```diff
@@ -32,71 +55,98 @@ paths:
...
     TokenAmount:
-      description: Amount of a token. uint112 encoded in decimal.
+      description: Amount of a token. uint256 encoded in decimal.
       type: string
       example: "1234567890"
-    Nonce:
-      description: Nonce of the batch in which the order is valid. uint32.
-      type: integer
-    OrderNew:
+    FeeInformation:
```
josojo, line 88: @anxolin @W3stside This is our newest proposal: we provide you with a feeRatio and a minimalFee.

Then the logic for calculating the fee will be Round(max(feeRatio*sellAmount, minimalFee)). Wdyt?

openapi.yml
```diff
@@ -32,71 +55,98 @@ paths:
...
+        Provides the information to calculate the fees. 
+      type: object
+      properties:
+        expirationDate:
+          description: |
+            Expiration date of the offered fee. Order service might not accept 
+            the fee after this expiration date. Encoded as unix timestamp.
+          type: integer
+        minimalFee:
+          description: Amount of minimal fee charged per order in specified sellToken
```
fleupold, line 99: Nit: maybe emphasize (either in the name or description) that this is an absolute amount.

openapi.yml
```diff
@@ -32,71 +55,98 @@ paths:
...
+          $ref: "#/components/schemas/TokenAmount"
+        feeRatio:
```
fleupold, line 101: Nit: Just an idea that we could denote this in basis points (which I think is pretty standard in finance)? This way it would a integer instead of a float on and avoid confusion around weather 0.1 means 0.1% or 10%.

openapi.yml
anxolin: Is this fee ration independent from the protocol fee? Or is meant to be used to replace the protocol fee?

We used to have 0.1% protocol fee, and the tip was for paying gas. My question is, we don't have any hard requirement anymore is just up to the operator to tell his clients to pay what he says?

openapi.yml
```diff
@@ -32,71 +55,98 @@ paths:
...
+          description: The fee ratio charged on a sellAmount.
+          example: 0.001
+          type: number
+          format: float
```
anxolin, line 105: I'm not super fan of FLOATS for precise thing. 

If you use them directly, then you'll get this kind of issues:
image

openapi.yml
anxolin: I agree with your last comments. We should use ISO dates. It's a more natural format for dates, and easier to read by anyone

This case is different than for the order. In the order the expirationDates is an epoch in the contract, therefore is what the user signs. So the creation of the orders endpoints has the original data (so the epoch) and the signature of it. So it was justified

openapi.yml
```diff
@@ -6,23 +6,46 @@ servers:
 - url: http://localhost:8080
   description: Local
 paths:
-  /api/v1/order:
+  /api/v1/orders:
```
anxolin, line 9: Will we be adding a `GET   /api/v1/orders/{ID}` endpoint?

openapi.yml
josojo: WE want to go with [this](https://gnosisinc.slack.com/archives/C8YGC05NG/p1606824110024700).  Yes, there will only be one fee. This fee is signed off by the user and will be Round(max(feeRatio*sellAmount, minimalFee))

openapi.yml
josojo: I added the GET endpoint. The Delete is more complicated and we should add it later at some point. I am not sure how the authentication should work. Probably, one must have a specific signed message with the order digest.

nlordell: I imagine something like:
```
signTypedData({
    operation: "delete",
    orderUid: "0x...",
})
```

The service could then verify that
```
ecrecover(message) == order.owner
```

And return a `403: Forbidden` if the check fails.

openapi.yml
josojo: I agree that here we have a different situation. Though I see value in keeping it always to unix time, as we have chosen it once.

But I really don't have a strong opinion.

openapi.yml
josojo: I would only go for `orderDigest || ownerAddress`, as this makes the UID unique already. But if you guys prefer nicks suggestion, please give him a thumbs up and I will change it.

nlordell: The idea behind using `orderDigest || ownerAddress || validTo` is that is allows this UID to be used to interact with the contract directly for reading fill amount and cancellation (on Etherscan for example). The hope was to reduce the amount of "magic numbers" used to identify orders to a single canonical form that can be used in multiple places.

openapi.yml
anxolin: Yes is not a critical decision, my point is that, in doubt I always prefer readability. Specially if this is meant to be use for integrations with 3rd parties.

Some pros/cons to make up our your minds.
https://stackoverflow.com/questions/47426786/epoch-or-iso8601-date-format

openapi.yml
anxolin: Ok 👌
I guess for being more specific, we'll round up to the next wei in the sell token currency.

openapi.yml
anxolin: Nice, also we avoid float issues

anxolin (approved):
🏅Good job @josojo !

openapi.yml
```diff
@@ -31,81 +54,149 @@ paths:
...
       allOf:
-        - $ref: "#/components/schemas/OrderNew"
-        - $ref: "#/components/schemas/OrderExtra"
+        - $ref: "#/components/schemas/OrderCreation"
+        - $ref: "#/components/schemas/OrderMetadata"
+    UID:
+     description: |
+      Unique identifier for the order: orderDigest_ownerAddress_validTo,
+      where orderDigest = keccak256(orderStruct). bytes32.
+     type: string
```
e00E, line 194: This isn't bytes32. Should probably say something like "65 bytes encoded as hex without `0x`. Bytes 0 to 30 are the order digest, bytes 30 to 42 the owner address and bytes 42..66 valid to. (i made up the sizes).

nlordell: `s/65/56/`. (`0..32 = orderDigest`, `32..52 = ownerAddress` and `52..56 = validTo`) but otherwise agree.

openapi.yml
```diff
@@ -31,81 +54,162 @@ paths:
...
-          description: Time offset in seconds from epoch until which the order is valid. uint32.
+          description: Unix timestamp until the order is valid. uint32.
+          type: integer
+        appData:
+          description: |
+            Arbitrary identifier sent along with the order.
+            Could be used to track the interface or other meta-aspects of the order. uint32 encoded
           type: integer
+        feeAmount:
+          description: "Fees: feeRatio + minimal_fee in atoms"
```
anxolin, line 173: @josojo I understand it should be

```suggestion
          description: "Fees: feeRatio * sellAmount + minimal_fee in atoms"
```

josojo:
Given that Felix very nicely deployed the swagger UIs, should we merge this PR for easy checking? Or should each PR implementing the API keep no updating the openapi.yml according to the current state?

e00E:
I like doing it in the individual PRs so it's clear what is and isn't implemented. But probably easier for frontend to see everything. I'm okay with either.

e00E (approved):

PR #50, opened 2020-11-24
Update order type for current smart contract

e00E:
and reorder to fields to match smart contract https://github.com/gnosis/oba-contracts/blob/main/src/contracts/libraries/GPv2Encoding.sol#L11 .

There is still discussion going on in https://github.com/gnosis/oba-services/pull/49 but we can already merge the order type and u256 serialization and come back to the error handling later.

### Test Plan
Tested with openapi and curl that order creation and retrieval works.

### DIFF
--- a/model/src/h160_hexadecimal.rs
+++ b/model/src/h160_hexadecimal.rs
@@ -1,0 +1,42 @@
+use primitive_types::H160;
+use serde::{de, Deserializer, Serializer};
+use std::fmt;
+
+pub fn serialize<S>(value: &H160, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    let mut bytes = [0u8; 20 * 2];
+    // Can only fail if the buffer size does not match but we know it is correct.
+    hex::encode_to_slice(value, &mut bytes).unwrap();
+    // Hex encoding is always valid utf8.
+    let s = std::str::from_utf8(&bytes).unwrap();
+    serializer.serialize_str(s)
+}
+
+pub fn deserialize<'de, D>(deserializer: D) -> Result<H160, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    struct Visitor {}
+    impl<'de> de::Visitor<'de> for Visitor {
+        type Value = H160;
+
+        fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+            write!(formatter, "an ethereum address as a hex encoded string")
+        }
+
+        fn visit_str<E>(self, s: &str) -> Result<Self::Value, E>
+        where
+            E: de::Error,
+        {
+            let mut value = H160::zero();
+            hex::decode_to_slice(s, value.as_mut()).map_err(|err| {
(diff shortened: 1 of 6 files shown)


model/src/h160_hexadecimal.rs
```diff
@@ -0,0 +1,42 @@
+use primitive_types::H160;
```
e00E, line 1: This has moved from lib.rs.

model/src/lib.rs
```diff
@@ -44,20 +34,20 @@ pub struct Signature {
 #[derive(Eq, PartialEq, Clone, Copy, Debug, Default, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct UserOrder {
```
e00E, line 36: reorderd fields, changed to u256, changed fill_type from enum to bool partially_fillable because that's the same as in the smart contract

openapi.yml
```diff
@@ -32,66 +32,43 @@ paths:
...
+        appData:
+          description: TODO description. uint32.
+          type: integer
+        tip:
+          $ref: "#/components/schemas/TokenAmount"
+        orderKind:
+          description: Is this a buy order or sell order?
+          type: string
+          enum: [buy, sell]
+        partiallyFillable:
```
josojo, line 69: Yeah, I like this better than in it is in my proposal!

josojo (approved):

openapi.yml
```diff
@@ -32,66 +32,43 @@ paths:
...
-          description: Is this a fill-or-kill order a partially fillable order?
-          type: string
-          enum: [fillorkill, partial]
-        nonce:
-          $ref: "#/components/schemas/Nonce"
         validTo:
           description: Time offset in seconds from epoch until which the order is valid. uint32.
           type: integer
+        appData:
+          description: TODO description. uint32.
```
fleupold, line 61: @josojo can you provide a description please? How about "Arbitrary identifier sent along with the order. Could be used to track the interface or other meta-aspects of the order"

model/src/u256_decimal.rs
```diff
@@ -0,0 +1,35 @@
+use primitive_types::U256;
+use serde::{de, Deserializer, Serializer};
+use std::fmt;
+
+pub fn serialize<S>(value: &U256, serializer: S) -> Result<S::Ok, S::Error>
```
fleupold, line 5: Why can we no longer use `serde_with::rust::display_fromstr`? Is it just to get rid of the dependency? 

Ubernit: Should the two serialization mods maybe be combined in a submod (to keep `/src` more easily comprehensible)

fleupold (approved):
Looks good!

openapi.yml
josojo: I proposed in the other PR:
`Identifier to make order unique or pass in arbitrary data. uint32 encoded in decimal.` 
But I like yours better Felix. I will adopt it.

model/src/u256_decimal.rs
e00E: U256's FromString implementation expects hex but in the api says these are decimal encoded. That's the only reason.

PR #51, opened 2020-11-26
Add trade encoding function

e00E:
Which creates the data that the smart contract's `decodeTrade` expects.

### Test Plan
New unit test.

### DIFF
--- a/solver/src/encoding.rs
+++ b/solver/src/encoding.rs
@@ -1,0 +1,84 @@
+use model::{OrderKind, UserOrder};
+use primitive_types::U256;
+
+const TRADE_STRIDE: usize = 204;
+
+/// Creates the data which the smart contract's `decodeTrade` expects.
+pub fn encode_trade(
+    order: &UserOrder,
+    sell_token_index: u8,
+    buy_token_index: u8,
+    executed_amount: U256,
+) -> [u8; TRADE_STRIDE] {
+    let mut result = [0u8; TRADE_STRIDE];
+    result[0] = sell_token_index;
+    result[1] = buy_token_index;
+    order.sell_amount.to_big_endian(&mut result[2..34]);
+    order.buy_amount.to_big_endian(&mut result[34..66]);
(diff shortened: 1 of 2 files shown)


solver/src/encoding.rs
```diff
@@ -0,0 +1,84 @@
...
+    order.sell_amount.to_big_endian(&mut result[2..34]);
+    order.buy_amount.to_big_endian(&mut result[34..66]);
+    result[66..70].copy_from_slice(&order.valid_to.to_be_bytes());
+    result[70..74].copy_from_slice(&order.app_data.to_be_bytes());
+    order.tip.to_big_endian(&mut result[74..106]);
+    result[106] = encode_order_flags(order);
+    executed_amount.to_big_endian(&mut result[107..139]);
+    result[139] = order.signature.v;
+    result[140..172].copy_from_slice(&order.signature.r.to_fixed_bytes());
+    result[172..204].copy_from_slice(&order.signature.s.to_fixed_bytes());
```
nlordell, line 25: Why not?
```suggestion
    result[140..172].copy_from_slice(order.signature.r.as_bytes());
    result[172..204].copy_from_slice(order.signature.s.as_bytes());
```

nlordell (approved):
Wonderful!

solver/src/encoding.rs
```diff
@@ -0,0 +1,84 @@
+use model::{OrderKind, UserOrder};
+use primitive_types::U256;
+
+const TRADE_SIZE: usize = 204;
+
+/// Creates the data which the smart contract's `decodeTrade` expects.
+pub fn encode_trade(
```
fleupold, line 7: probably doesn't make a difference, but curious why not have this as an extension on the Order struct directly?

fleupold (approved):

solver/src/encoding.rs
e00E: Not sure yet where it should live and a function is easier to move. Maybe we'll have a Trade struct later on which it can be implemented. And this function is only relevant to the solver and not the orderbook.

PR #52, opened 2020-11-30
Add settlement struct and remove old nonce code

e00E:
Making this a separate PR before I add my uniswap naive solver.
I've removed driver code related to how the nonce previously worked to avoid confusion when more people touch the code base.

For the interactions I'm not sure how they are going to look like. I've included two basic uniswap functions so that I can return them in my solver.

### Test Plan
CI

### DIFF
--- a/solver/src/settlement.rs
+++ b/solver/src/settlement.rs
@@ -1,27 +1,36 @@
-use model::{Order, TokenPair, UserOrder};
-use std::{
-    collections::HashMap,
-    time::{Duration, Instant},
-};
+use model::UserOrder;
+use primitive_types::{H160, U256};
+use std::collections::HashMap;
 
-pub struct Settlement {
-    pub token_pair: TokenPair,
-    pub orders: Vec<UserOrder>,
-    // TODO: sell amounts, AMM trades
+#[derive(Debug)]
+pub struct Trade {
+    pub order: UserOrder,
+    pub executed_amount: U256,
 }
 
-// Assumes all orders have spending approved and enough balance.
-pub fn find_settlement(
-    _orders: &[Order],
-    _nonces: &HashMap<TokenPair, u32>,
-    _now: Instant,
-    _max_order_age: Duration,
-    // TODO: AMM price
-) -> Option<Settlement> {
-    // TODO: Find oldest order where nonce is correct. If it is older than max_order_age settle that pair.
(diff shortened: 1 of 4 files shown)


nlordell (approved):

solver/src/settlement.rs
```diff
@@ -1,27 +1,36 @@
...
-    // unreasonably priced orders then we cannot settle it. But there could be a counter
-    // unreasonable order that does allow settling.
-    todo!()
+#[derive(Debug, Default)]
+pub struct Settlement {
+    pub clearing_prices: HashMap<H160, U256>,
+    pub fee_factor: U256,
+    pub trades: Vec<Trade>,
+    pub interactions: Vec<Interaction>,
+    pub order_refunds: Vec<()>,
```
fleupold, line 35: what is this for?

solver/src/settlement.rs
```diff
@@ -1,27 +1,36 @@
...
-    pub orders: Vec<UserOrder>,
-    // TODO: sell amounts, AMM trades
+#[derive(Debug)]
+pub struct Trade {
+    pub order: UserOrder,
+    pub executed_amount: U256,
+}
+
+#[derive(Debug)]
+pub enum Interaction {
```
fleupold, line 12: Maybe not needed immediately, but I think in the long run it might be better to model Interactions as a trait object and have different implementations of e.g. Uniswap, Curve, etc that can just return the calldata needed for "executing" the interaction.

With an enum I'm afraid this will become a large centralised data-structure.

fleupold (approved):

solver/src/settlement.rs
e00E: I took it from the smart contract interface:
>     /// @param encodedOrderRefunds Encoded order refunds for clearing storage

solver/src/settlement.rs
e00E: Let's do that when we see that it would be beneficial but I think you are probably right.

solver/src/settlement.rs
fleupold: I see. @nlordell I was hoping we could model storage freeings as an interaction as well (rather than bloating up the public interface).

nlordell: I think that is a really good idea.

PR #55, opened 2020-12-01
Add uniswap simulator

e00E:
Not yet used but will be to verify that interaction from naive amm
settlement works.

### Test Plan
new unit tests

### DIFF
--- a/solver/src/main.rs
+++ b/solver/src/main.rs
@@ -4,6 +4,7 @@ mod batcher;
 mod driver;
 mod encoding;
 mod ethereum;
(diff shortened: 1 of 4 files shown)


fleupold:
After discussing that we could just "try" to submit the solution on chain is this still needed? If so what are your thoughts around using the actual smart contract implementation instead of copying it locally?

e00E:
I agree that we won't need this immediately. Will keep the PR open as draft in case we want it later.

fleupold:
What's the latest status on this? Can we close this PR?

PR #68, opened 2020-12-02
Adding Dockerfile

fleupold:
As a first step to get continuous development running. I went for using the same container for both binaries (as it's less code to maintain, faster to build on CI and I don't expect the binary to be that big in size). Looking forward to your thoughts if this is a bad idea. This decision will also affect if I have to request 1 or multiple dockerhub images

The .dockerignore/Dockerfile is mostly taken from https://github.com/gnosis/rust-template however due to having multiple workspaces the trick of compiling dependencies before copying the code doesn't work.

### Test Plan

```sh
docker build --tag gnosispm:oba-services -f docker/Dockerfile .
docker run gnosispm:oba-services solver # panics due to "not implemented"
docker run -p 8080:8080 gnosispm:oba-services orderbook
``` 

and then fetching http://localhost:8080/api/v1/orders in the browser.

### DIFF
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -8,6 +8,6 @@ use std::sync::Arc;
 async fn main() {
     let orderbook = Arc::new(OrderBook::default());
     let filter = api::handle_all_routes(orderbook);
-    let result = warp::serve(filter).bind(([127, 0, 0, 1], 8080)).await;
+    let result = warp::serve(filter).bind(([0, 0, 0, 0], 8080)).await;
     println!("warp exited: {:?}", result);
 }
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+target/
+Dockerfile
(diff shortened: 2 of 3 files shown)


orderbook/src/main.rs
```diff
@@ -8,6 +8,6 @@ use std::sync::Arc;
 async fn main() {
     let orderbook = Arc::new(OrderBook::default());
     let filter = api::handle_all_routes(orderbook);
-    let result = warp::serve(filter).bind(([127, 0, 0, 1], 8080)).await;
+    let result = warp::serve(filter).bind(([0, 0, 0, 0], 8080)).await;
```
fleupold, line 11: This was needed since loopback (127.0.0.1) is not accessible from the host system (your browser). This is also the IP at which we bind the price estimator.

e00E (approved):
Seems like a good idea with the single docker image. And we can still easily change it later if we want to.

josojo (approved):
Nice, testplan works.

setup works for me as well.

PR #70, opened 2020-12-03
Add uniswap contract

e00E:
For this I had to allow us to specify the full npm path to contracts
because the uniswap package has a different layout.

Also updated to the latest version of the openzeppelin repo.

### Test Plan
`cargo open --doc` and see that both contracts show up.


fleupold (approved):
Nice, I'm looking at depending on a preliminary version of gp-v2-contracts next.

PR #71, opened 2020-12-03
Encode uniswap interaction call

e00E:
Does not yet have the target address of the contract because I didn't
know how it is supposed to be encoded.

### Test Plan
unit test for encoding.

### DIFF
--- a/solver/src/encoding.rs
+++ b/solver/src/encoding.rs
@@ -1,7 +1,9 @@
+use contracts::IUniswapV2Router02;
 use model::{OrderCreation, OrderKind};
-use primitive_types::U256;
+use primitive_types::{H160, U256};
 
 const TRADE_STRIDE: usize = 204;
+const UNISWAP_INTERACTION_SIZE: usize = 260;
 
 /// Creates the data which the smart contract's `decodeTrade` expects.
 pub fn encode_trade(
@@ -35,6 +37,51 @@ fn encode_order_flags(order: &OrderCreation) -> u8 {
         result |= 0b00000010;
     }
     result
+}
+
+fn encode_uniswap_call(
+    amount_in: U256,
+    amount_out_min: U256,
+    token_in: H160,
+    token_out: H160,
+    payout_to: H160,
+) -> [u8; UNISWAP_INTERACTION_SIZE] {
+    let uniswap = IUniswapV2Router02::at(&dummy::dummy_web3(), H160::zero());
+    let method = uniswap.swap_exact_tokens_for_tokens(
+        amount_in,
+        amount_out_min,
+        vec![token_in, token_out],
+        payout_to,
+        U256::MAX,
+    );
+    let data = method.tx.data.unwrap().0;
+    let mut result = [0u8; UNISWAP_INTERACTION_SIZE];
+    result.copy_from_slice(data.as_slice());
+    result
+}
+
+// To create an ethcontract instance we need to provide a web3 even though we never use it. This
+// module provides a dummy transport and web3.
+mod dummy {
+    use jsonrpc_core::Call as RpcCall;
+    use serde_json::Value;
+    use web3::{api::Web3, Transport};
+
+    #[derive(Clone, Debug)]
+    pub struct DummyTransport;
+    impl Transport for DummyTransport {
+        type Out = futures::future::Pending<web3::Result<Value>>;
+        fn prepare(&self, _method: &str, _params: Vec<Value>) -> (web3::RequestId, RpcCall) {
+            unimplemented!()
+        }
+        fn send(&self, _id: web3::RequestId, _request: RpcCall) -> Self::Out {
+            unimplemented!()
+        }
+    }
+
+    pub fn dummy_web3() -> Web3<DummyTransport> {
+        Web3::new(DummyTransport)
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


solver/src/encoding.rs
```diff
@@ -37,6 +39,51 @@ fn encode_order_flags(order: &UserOrder) -> u8 {
     result
 }
 
+fn encode_uniswap_call(
+    amount_in: U256,
+    amount_out_min: U256,
+    token_in: H160,
+    token_out: H160,
+    payout_to: H160,
+) -> [u8; UNISWAP_INTERACTION_SIZE] {
```
fleupold, line 48: Wouldn't it be more convenient if this method returned Bytes (or Vec<u8>). Then our code wouldn't have to know the stride of the encoded call which seems more flexible. Also with multiple interactions we likely have to do a lot of concatenation and eventually things will be submitted as Bytes to the smart contract.

solver/src/encoding.rs
```diff
@@ -81,4 +128,33 @@ mod tests {
         assert_eq!(encoded[140..172], u8_as_32_bytes_be(10));
         assert_eq!(encoded[172..204], u8_as_32_bytes_be(11));
     }
+
+    #[test]
+    fn encode_uniswap_call_() {
```
fleupold, line 133: I always like unit tests, but here I'm not sure if we are not actually more testing ethcontracts rather than out code.

fleupold (approved):

solver/src/encoding.rs
e00E: Yeah I agree. Probably won't have a test like this for all the interactions. I was checking manually whether the bytes make sense and decided to keep the test.

solver/src/encoding.rs
e00E: The caller can always call to_vec. I'm not very opposed to making this return a vec directly just conceptually this has a fixed size so that's what this low level method should return.

nlordell: I also prefer it as `[u8; UNISWAP_INTERACTION_SIZE]`. The  caller can always `Vec::extend_from_slice` to build many of these or even `.to_vec()` to convert it to a vector. My main argument is that the using an array instead of a vec encodes the length information in the type instead of just losing it.

solver/src/encoding.rs
```diff
@@ -37,6 +39,51 @@ fn encode_order_flags(order: &UserOrder) -> u8 {
...
+        U256::MAX,
+    );
+    let data = method.tx.data.unwrap().0;
+    let mut result = [0u8; UNISWAP_INTERACTION_SIZE];
+    result.copy_from_slice(data.as_slice());
+    result
+}
+
+// To create an ethcontract instance we need to provide a web3 even though we never use it. This
+// module provides a dummy transport and web3.
```
nlordell, line 64: This is unfortunate. Maybe we can do some refactoring in `ethcontract` so that this isn't true. I ran into this issue just the other day and was extremely happy to find out that `ethers.js` can create ABI interface types just for encoding and decoding without a provider (Ethers.js lingo for a `Transport`).

I created https://github.com/gnosis/ethcontract-rs/issues/413 to capture this.

solver/src/encoding.rs
nlordell: I'm fine with this test, as if we decide to not use `ethcontract` in the `encode_uniswap_call` implementation, we have a unit test that ensures it works :smile:.

nlordell (approved):
Looks good!

solver/src/encoding.rs
fleupold: Sure, but why do we care about the length? If the Uniswap signature changed (which is impossible but let's assume this was another "upgradable" interaction), we would have to change the size constant as well. And right now it wouldn't even give us a compilation error but likely just panic at runtime when it tries to `copy_from_slice` from a differently sized slice. 

My point is that I don't see what part of the solver or even contract for that matter actually cares about how many bytes an interaction has - isn't it just an arbitrarily sized byte vector?

nlordell: > Sure, but why do we care about the length?

It just enriches the type with more information.

> If the Uniswap signature changed (which is impossible but let's assume this was another "upgradable" interaction), we would have to change the size constant as well.

The idea is that then the implementation of `encode_uniswap_call` would be incorrect, but not the outside code that compiles with the `[u8; UNISWAP_INTERACTION_SIZE]` type. Then, if the size would change, the compiler would help catch errors that would cause panics at runtime (like indexing out of bounds):

```rust
    let array = [0u8; 20];
    println!("{}", array[21]);
```

Produces an error
```
error: this operation will panic at runtime
 --> src/main.rs:3:20
  |
3 |     println!("{}", array[21]);
  |                    ^^^^^^^^^ index out of bounds: the length is 20 but the index is 21
```

By giving more information to the compiler just enables it to do more checking. This would not be the case with `Vec`.

Additionally, calling code would have to make less assumptions about the length of the encoded bytes (for example, does indexing the `encode_uniswap_call` result need to be checked everytime for errors?) and in unit tests in calling code passing a `[u8; UNISWAP_INTERACTION_SIZE]` make it impossible to create some bytes with the wrong length, this means if the size of `UNISWAP_INTERACTION_SIZE` would change because it is an upgradable contract, then these tests would have to manually be found and adjusted and the compiler would not be able to help, making it possible for calling code to have bugs that won't get caught by the unit tests and cause runtime errors.

> just panic at runtime when it tries to copy_from_slice from a differently sized slice.

It would break the `encode_uniswap_call` unit test.

nlordell: > My point is that I don't see what part of the solver or even contract for that matter actually cares about how many bytes an interaction has - isn't it just an arbitrarily sized byte vector?

In that case:
```rust
type UniswapInteraction = [u8; UNISWAP_INTERACTION_SIZE];

trait ContractInteraction: AsRef<[u8]> {}
impl ContractInteraction for UniswapInteraction {}
```

solver/src/encoding.rs
```diff
@@ -37,6 +39,51 @@ fn encode_order_flags(order: &UserOrder) -> u8 {
...
+        amount_in,
+        amount_out_min,
+        vec![token_in, token_out],
+        payout_to,
+        U256::MAX,
+    );
+    let data = method.tx.data.unwrap().0;
+    let mut result = [0u8; UNISWAP_INTERACTION_SIZE];
+    result.copy_from_slice(data.as_slice());
+    result
```
nlordell, line 60: nit: We can use a cool new Rust feature: https://doc.rust-lang.org/std/convert/trait.TryFrom.html#impl-TryFrom%3CVec%3CT%3E%3E
```suggestion
    result.try_into().expect("encoded Uniswap interaction unexpected byte length")
```

solver/src/encoding.rs
```diff
@@ -37,6 +39,51 @@ fn encode_order_flags(order: &UserOrder) -> u8 {
...
+) -> [u8; UNISWAP_INTERACTION_SIZE] {
+    let uniswap = IUniswapV2Router02::at(&dummy::dummy_web3(), H160::zero());
+    let method = uniswap.swap_exact_tokens_for_tokens(
+        amount_in,
+        amount_out_min,
+        vec![token_in, token_out],
+        payout_to,
+        U256::MAX,
+    );
+    let data = method.tx.data.unwrap().0;
```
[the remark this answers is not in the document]
e00E: turned it into expect with message

josojo (approved):
LGTM

solver/src/encoding.rs
fleupold: > Then, if the size would change, the compiler would help catch errors that would cause panics at runtime (like indexing out of bounds)

Believe me that I generally love type safety. My point is given that we treat contract interactions as arbitrary bytes, they **should** not be indexed into or otherwise inspected by the caller (they should only concatenate them to larger calldata). Your point about the unit test failing and thus catching potential changes is good. 

I don't think this debate is blocking the PR, feel free to merge as is.

nlordell: I'm also fine with using a `Vec` for the return type, it's really a very small detail that I don't think it's important. I just like using a fixed size array slightly better as I feel the type more accurately describes what is being returned (a Uniswap contract interaction vs a general contract interaction) and it can always be converted to a `Vec` with `to_vec`. This allows the caller to essentially decide whether or not it wants the type erasure.

Issue #63, opened 2020-12-02
Implement tip api


e00E:
We need to communicate what the expected tip for an order is to the frontend. A proper implementation likely needs to figure out erc20 token prices and make sure the orderbook and the solver agree on the tip.
For the MVP we might want to simplify the tip to some fixed amount or 0.

PR #72, opened 2020-12-03
fee api endpoint

josojo:
First step for #63, the main objective is to bring the API endpoints live to unblock the front-end.

### Test Plan
unit tests only

### DIFF
--- a/orderbook/src/api/filter.rs
+++ b/orderbook/src/api/filter.rs
@@ -18,7 +18,8 @@ fn extract_user_order() -> impl Filter<Extract = (UserOrder,), Error = warp::Rej
     warp::body::content_length_limit(MAX_JSON_BODY_PAYLOAD).and(warp::body::json())
 }
 
-fn extract_sell_token() -> impl Filter<Extract = (H160,), Error = warp::Rejection> + Clone {
+fn extract_sell_token(
+) -> impl Filter<Extract = (handler::FeeRequestBody,), Error = warp::Rejection> + Clone {
     // (rejecting huge payloads)...
     warp::body::content_length_limit(MAX_JSON_BODY_PAYLOAD).and(warp::body::json())
 }
@@ -54,6 +55,7 @@ pub mod test_util {
     use super::*;
     use model::Order;
     use primitive_types::U256;
+    use serde_json::json;
     use warp::{http::StatusCode, test::request};
 
     #[tokio::test]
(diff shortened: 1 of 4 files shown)


orderbook/src/api/filter.rs
```diff
@@ -17,6 +18,11 @@ fn extract_user_order() -> impl Filter<Extract = (UserOrder,), Error = warp::Rej
     warp::body::content_length_limit(MAX_JSON_BODY_PAYLOAD).and(warp::body::json())
 }
 
+fn extract_sell_token() -> impl Filter<Extract = (H160,), Error = warp::Rejection> + Clone {
+    // (rejecting huge payloads)...
+    warp::body::content_length_limit(MAX_JSON_BODY_PAYLOAD).and(warp::body::json())
```
e00E, line 23: This extracts the sell token from the body but I think it is supposed to be a url parameter.

e00E: Maybe we should update the corresponding openapi change in this PR too.

josojo: This was actually on purpose. We wanted to have it that way, in order to add easily more parameters later on. This might be needed, if we want to subsidize particular trades, eg. above certain sellAmounts, between certain tokens, only fill/Kill orders etc.

orderbook/src/api/filter.rs
```diff
@@ -59,6 +73,26 @@ pub mod test_util {
         assert_eq!(response_orders, orderbook_orders);
     }
     #[tokio::test]
+    async fn get_fee_info_() {
+        let filter = get_fee_info();
+        let sell_token: H160 = "6b175474e89094c44da98b954eedeac495271d0f".parse().unwrap();
```
e00E, line 78: For the test it would be better to actually create the json body exactly as it is supposed to sent, probably with the serde_json::json macro. That way we see exactly what the expected encoding. The way the test is written now it is not clear how H160 will be serialized.

e00E (approved):

orderbook/src/api/filter.rs
fleupold: I must have missed this specification detail in the API proposal. It makes using the API e.g. in a normal browser harder (to test something locally you'd always have to use curl). Maybe we could specify the sell token in the URL to get the default fee and have any extra parameters that we might want to consider later in the body?

Also I don't see how adding parameters to the body is simpler than to the URL? They could also be optional in the URL in order to be backwards compatible.

fleupold (commented):
This was merged quite fast (sorry I didn't get to do a pass over all outstanding PRs before this afternoon and it was posted quite late yesterday).

PR #74, opened 2020-12-04
Naming convention from API-PR with uid

josojo:
Naming convention from API-PR with uid variable.

### Test Plan
checking against the API PR

### DIFF
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -45,7 +45,7 @@ pub struct UserOrder {
     pub valid_to: u32,
     pub app_data: u32,
     #[serde(with = "u256_decimal")]
-    pub tip: U256,
+    pub fee_amount: U256,
     pub order_kind: OrderKind,
     pub partially_fillable: bool,
     pub signature: Signature,
@@ -170,7 +170,7 @@ mod tests {
           "buyAmount": "0",
           "validTo": 4294967295u32,
           "appData": 0,
-          "tip": "115792089237316195423570985008687907853269984665640564039457584007913129639935",
+          "feeAmount": "115792089237316195423570985008687907853269984665640564039457584007913129639935",
           "orderKind": "buy",
           "partiallyFillable": false,
           "signature": "0102000000000000000000000000000000000000000000000000000000000000030400000000000000000000000000000000000000000000000000000000000005",
@@ -185,7 +185,7 @@ mod tests {
                 buy_amount: 0.into(),
                 valid_to: u32::MAX,
                 app_data: 0,
-                tip: U256::MAX,
+                fee_amount: U256::MAX,
                 order_kind: OrderKind::Buy,
                 partially_fillable: false,
                 signature: Signature {
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -62,18 +62,11 @@ impl OrderCreation {
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub struct OrderUid(pub [u8; 56]);
 
-impl Default for OrderUid {
-    fn default() -> Self {
-        OrderUid([0 as u8; 56])
-    }
-}
-
 impl Serialize for OrderUid {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: Serializer,
     {
-        // Todo: build in underscores
         serializer.serialize_str(&hex::encode(self.0.iter()))
     }
 }
@@ -95,7 +88,6 @@ impl<'de> Deserialize<'de> for OrderUid {
             where
                 E: de::Error,
             {
-                // Todo parse digest, owner and validTo individually and consider underscores
                 let mut value = [0 as u8; 56];
                 hex::decode_to_slice(s, value.as_mut()).map_err(|err| {
                     de::Error::custom(format!("failed to decode {:?} as hex: {}", s, err))
@@ -123,7 +115,7 @@ impl Default for OrderMetaData {
         Self {
             creation_date: DateTime::from_utc(NaiveDateTime::from_timestamp(0, 0), Utc),
             owner: Default::default(),
-            uid: Default::default(),
+            uid: OrderUid([0 as u8; 56]),
         }
     }
 }
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -134,7 +134,9 @@ impl Default for OrderMetaData {
 #[derive(Eq, PartialEq, Clone, Default, Debug, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct Order {
+    #[serde(flatten)]
     pub order_meta_data: OrderMetaData,
+    #[serde(flatten)]
     pub order_creation: OrderCreation,
 }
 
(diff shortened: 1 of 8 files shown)


model/src/lib.rs
```diff
@@ -45,41 +45,49 @@ pub struct UserOrder {
...
 
+/// An order that is returned when querying the orderbook.
+///
+/// Contains extra fields thats are populated by the orderbook.
+#[derive(Eq, PartialEq, Clone, Default, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Order {
+    pub order_meta_data: OrderMetaData,
+    pub order_creation: OrderCreation,
+}
```
e00E, line 89: I don't think this matches the openapi spec because there is no `#[serde(flatten)]` anymore which is why you had to change the test to use individual structs.

josojo: Do you think it should be returned flattened?

The openapi  PR does not mention the fattening anymore.

e00E: I think it does because it uses

```
Order:
      allOf:
        - $ref: "#/components/schemas/OrderMetadata"
```

we should load it in https://editor.swagger.io/ to confirm. I just tried this but it looks like there are some errors in the spec that the editor complains about.

model/src/lib.rs
```diff
@@ -45,41 +45,49 @@ pub struct UserOrder {
...
 #[serde(rename_all = "camelCase")]
-pub struct Order {
-    pub creation_time: DateTime<Utc>,
+pub struct OrderMetaData {
+    pub creation_date: DateTime<Utc>,
     #[serde(with = "h160_hexadecimal")]
     pub owner: H160,
-    #[serde(flatten)]
-    pub user_provided: UserOrder,
+    pub uid: String,
```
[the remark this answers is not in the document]
e00E: I also think this should internally store fixed size bytes instead of string. A wrapping struct like what Nic suggested is nice because we can implement methods and serialization on it.
If the serialization is complicated for this PR feel free to do it in another.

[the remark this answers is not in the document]
josojo: Unfortunately, I needed to reimplement all traits, because arrays of size bigger than 32 suck in rust, according to my limited knowledge :) 
If you guys know better, please! help.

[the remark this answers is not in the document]
e00E: I don't think so. ```rust
#[derive(Eq, PartialEq, Clone, Copy, Debug)]
pub struct Foo([u8; 56]);
```
compiles.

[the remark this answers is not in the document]
Comment (nlordell) [reply model/src/lib.rs]: You might need to update your Rust version. Recently they've been rolling out "const generics" so most of the std library traits are all implemented `for<const N: usize> [u; N]`.

File: model/src/lib.rs [hunk 3]
```diff
@@ -51,35 +52,94 @@ pub struct UserOrder {
...
+        OrderUid([0 as u8; 56])
+    }
+}
+
+impl Serialize for OrderUid {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        // Todo: build in underscores
```
nlordell, line 76: Maybe it makes sense not to use underscores in the end so the UID can be pasted into Etherscan. WDYT? Maybe prefix with `0x`?

model/src/lib.rs
```diff
@@ -51,35 +52,94 @@ pub struct UserOrder {
...
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct OrderUid(pub [u8; 56]);
+
+impl Default for OrderUid {
+    fn default() -> Self {
+        OrderUid([0 as u8; 56])
+    }
+}
+
+impl Serialize for OrderUid {
```
nlordell, line 71: I used this crate the other day and was very happy with it: https://docs.rs/serde_with/1.6.0/serde_with/hex/struct.Hex.html. It has a bunch of helpers for implementing serialization and deserialization.

Alternatively, you could implement `Display` and `FromStr` and then use: https://docs.rs/serde_with/1.6.0/serde_with/struct.DisplayFromStr.html. This way you can customize how the hex is serialized but don't have to add the complex `Visitor` code.

orderbook/src/orderbook.rs
```diff
@@ -52,12 +53,21 @@ impl OrderBook {
...
 struct InvalidSignatureError {}
-fn user_order_to_full_order(user_order: UserOrder) -> Result<Order, InvalidSignatureError> {
-    // TODO: verify signature and extract owner
+fn user_order_to_full_order(user_order: OrderCreation) -> Result<Order, InvalidSignatureError> {
+    // TODO: verify signature and extract owner, get orderDigest
+    let owner = H160::zero();
+    let digest = H256::zero();
+    let valid_to = user_order.valid_to.to_be_bytes();
+    let mut uid = [0 as u8; 56];
+    uid.copy_from_slice(&[digest.as_bytes(), owner.as_bytes(), &valid_to].concat());
```
nlordell, line 62: It should be possible to use the standard library `TryFrom` implementation to convert directly to an array:
```suggestion
    let uid = OrderUid([digest.as_bytes(), owner.as_bytes(), &valid_to].concat().try_into());
```

nlordell (approved):
Nice!

model/src/lib.rs
e00E: According to the openapi spec there are no underscores in the string, just hex encoded bytes without 0x. No 0x is also what I specified for other fields in the openapi spec.

orderbook/src/orderbook.rs
josojo: Oh right, it was not working for me, just because of my old rust version. Nice thanks

model/src/lib.rs
josojo: Okay, let's keep the underscores out. 
I see that `0x` prefix could be convenient, though we don't use them for any API interface, and hence it would be little bit weird. I guess users can add 0x easily themselves. Hopefully not too many users will do that anyways.

model/src/lib.rs
```diff
@@ -51,35 +52,94 @@ pub struct UserOrder {
...
 
-/// An order that is returned when querying the orderbook.
-///
-/// Contains extra fields thats are populated by the orderbook.
-#[derive(Eq, PartialEq, Clone, Copy, Debug, Deserialize, Serialize)]
+// uid as 56 bytes: 32 for orderDigest, 20 for ownerAddress and 4 for validTo
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct OrderUid(pub [u8; 56]);
+
+impl Default for OrderUid {
```
fleupold, line 65: nit: Do we need Default on OrderUid? Maybe we can just implement default on UserOrder which could create the correct ID on an empty order. It could be weird to use an id that doesn't confirm to the derived format.

fleupold (approved):
Looks good to me. One comment about Defaults

model/src/lib.rs
josojo: I tried your first suggestion, but it does only work for vectors or arrays below the size of 32 :( Even with the new compiler.

I also tried your second suggestion. But here I would have to implement the `Display` trait for [u8;56] and not for OrderUid. Otherwise, I get:
```
`[u8; 56]` doesn't implement `std::fmt::Display`
```
And I am not allowed to implement Display for [0;56], as I am not in the "defining" crate.

Somehow my knowledge is not sufficient to jump over these hurdles.

PR #75, opened 2020-12-04
Find orders to settle from whole orderbook

e00E:
by looking at the best prices for both sides of each token pair.

### Test Plan
new unit tests

### DIFF
--- a/solver/src/naive_solver.rs
+++ b/solver/src/naive_solver.rs
@@ -1,0 +1,144 @@
+mod two_order_settlement;
+
+use crate::settlement::Settlement;
+use model::{OrderCreation, OrderKind, TokenPair};
+use primitive_types::U512;
+use std::{cmp::Ordering, collections::HashMap};
+
+pub fn settle(orders: impl Iterator<Item = OrderCreation>) -> Option<Settlement> {
+    let orders = organize_orders_by_token_pair(orders);
+    // TODO: Settle multiple token pairs in one settlement.
+    orders
+        .into_iter()
+        .find_map(|(_, orders)| settle_pair(orders))
+}
+
+fn settle_pair(orders: TokenPairOrders) -> Option<Settlement> {
+    let most_lenient_a = orders.sell_token_0.into_iter().min_by(order_by_price)?;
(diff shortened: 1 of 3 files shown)


solver/src/naive_solver.rs
```diff
@@ -0,0 +1,144 @@
...
+        }
+    }
+    result
+}
+
+fn usable_order(order: &UserOrder) -> bool {
+    matches!(order.order_kind, OrderKind::Sell)
+        && !order.sell_amount.is_zero()
+        && !order.buy_amount.is_zero()
+        && !order.partially_fillable
```
fleupold, line 50: Why do we exclude partially fillable orders? They can also be completely settled. 

Generally, I feel like order filtering (e.g. for 0 amount orders) should happen outside the core solvers (but instead in the order fetching parts of the system).

solver/src/naive_solver.rs
```diff
@@ -0,0 +1,144 @@
+mod two_order_settlement;
+
+use crate::settlement::Settlement;
+use model::{OrderKind, TokenPair, UserOrder};
+use primitive_types::U512;
+use std::{cmp::Ordering, collections::HashMap};
+
+pub fn settle(orders: impl Iterator<Item = UserOrder>) -> Option<Settlement> {
```
fleupold, line 8: Nice, I assume this will be the public interface that eventually becomes a trait so we can have multiple solvers?

fleupold (approved):
Looks great and very nice to read...

solver/src/naive_solver.rs
e00E: I guess partially fillable would work too but for that we need to fetch the already filled amount first.
With the 0 amount it's not clear whether this could even happen in the orderbook or whether all solvers need this prefiltered. Another solver might have no problem settling orders that have 0 buy amount.

solver/src/naive_solver.rs
e00E: Something like this, don't know exactly yet. We will see when we integrate a second solver.

josojo (approved):
Very nice!

PR #76, opened 2020-12-04
Autodeploy

fleupold:
This PR makes it so that we can automatically deploy and restart services that run in staging when merging into master. Scrip is mostly copied from [gas-token-mining](https://github.com/gnosis/gas-token-mining/blob/d196ffd6b7a439e62e3f339a1e4b210cc03b2915/docker/deploy.sh#L1) while using the same autodeploy push tag as gp-v1-services

When ready, we can easily restart more services by adding them to the URL here: https://travis-ci.com/github/gnosis/gp-v2-services/settings

### Test Plan
Once merged see dev-dfusion-v2-api-rinkeby pod restarting.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,3 +35,13 @@ script:
 
 before_cache:
   - rm -rf "$TRAVIS_HOME/.cargo/registry/src"
+
+deploy:
+  - provider: script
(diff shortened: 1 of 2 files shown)


e00E (approved):

josojo (approved):
Nice!

fleupold:
🤞

PR #82, opened 2020-12-07
Add minimal driver run loop

e00E:
in order to add more structure to the solver code. This outlines a very
simple run loop that attempts to settle every 30 seconds. No
transactions run in parallel which means we won't need code to keep track of multiple running transactions for the demo.
For now uses `println` but want to use the tracing crate in the future instead of logging. Will make another PR for that.

### Test Plan
not yet testable. can test this once we have the settlement contract and maybe ganache.

### DIFF
--- a/solver/Cargo.toml
+++ b/solver/Cargo.toml
@@ -14,5 +14,5 @@ model = { path = "../model" }
 primitive-types = "0.7"
 reqwest = { version = "0.10", features = ["json"] }
 serde_json = "1.0"
-tokio = { version = "0.2", features =[ "macros"] }
-web3 = { version = "0.13", default-features = false, features = [] }
+tokio = { version = "0.2", features =[ "macros", "time"] }
+web3 = { version = "0.13", default-features = false }
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -1,7 +1,47 @@
-use crate::ethereum::SettlementContract;
+use crate::{ethereum::SettlementContract, naive_solver, orderbook::OrderBookApi};
+use anyhow::{Context, Result};
 use std::{sync::Arc, time::Duration};
+
+const SETTLE_INTERVAL: Duration = Duration::from_secs(30);
 
 pub struct Driver {
     contract: Arc<dyn SettlementContract>,
-    max_order_age: Duration,
+    orderbook: OrderBookApi,
+}
+
+impl Driver {
+    async fn run_forever(&mut self) {
+        loop {
+            println!("starting settle attempt");
+            match self.single_run().await {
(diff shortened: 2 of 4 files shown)


josojo (approved):
Yeah, starting with non-parallel runs seems very reasonable for the start. LGTM

solver/src/ethereum.rs
```diff
@@ -9,7 +9,10 @@ use primitive_types::H160;
 #[async_trait]
 pub trait SettlementContract {
     async fn get_nonce(&self, token_pair: TokenPair) -> Result<u32>;
-    async fn settle(&self, settlement: Settlement) -> Result<()>;
+    // Call simulates the transaction.
+    async fn settle_call(&self, settlement: &Settlement) -> Result<()>;
+    // Send really executes the transaction.
+    async fn settle_send(&self, settlement: &Settlement) -> Result<()>;
```
fleupold, line 15: ubernit: simulate_settle, execute_settle are maybe slightly more explicit what these methods do (for the caller it might be less important how they do it).

solver/src/driver.rs
```diff
@@ -1,7 +1,47 @@
...
+    async fn run_forever(&mut self) {
+        loop {
+            println!("starting settle attempt");
+            match self.single_run().await {
+                Ok(()) => println!("ok"),
+                Err(err) => println!("error: {:?}", err),
+            }
+            tokio::time::delay_for(SETTLE_INTERVAL).await;
+        }
+    }
```
fleupold, line 22: Here as well I'd slightly prefer a scheduler, but less important than in the orderbook (this class here is still small and relatively focused on actually doing the runloop) so it can also wait.

fleupold (approved):

solver/src/ethereum.rs
e00E: I won't change the name now because it's just a placeholder trait until we have the settlement ethcontract contract.

solver/Cargo.toml
```diff
@@ -10,4 +10,4 @@ maplit = "1.0"
 model = { path = "../model" }
 primitive-types = "0.7"
 reqwest = { version = "0.10", features = ["json"] }
-tokio = { version = "0.2", features =[ "macros"] }
+tokio = { version = "0.2", features =[ "macros", "time"] }
```
bh2smith, line 13: I think I saw this dependency introduced in another recent PR. You might find conflicts with this at some point... but no big deal

bh2smith (approved):
Very exciting!

bh2smith:
I suppose the most generic way to go is to make a class for `Interaction` and have a pre-made `genericFailing` and `genericPassing` constructor.

PR #83, opened 2020-12-07
Use tracing for logging

e00E:
Iirc Nic recommend the tracing crate before and I would like to use it instead of the previous `log`. `tracing` is able to log those messages too so we won't lose them.
Currently this is using all defaults but the output format, log filter etc can be adjusted.

### Test Plan
`env RUST_LOG=debug cargo run --bin orderbook`
> Dec 07 13:22:33.692  INFO orderbook: serving order book address=0.0.0.0:8080

### DIFF
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -2,12 +2,15 @@ mod api;
 mod orderbook;
 
 use crate::orderbook::OrderBook;
-use std::sync::Arc;
+use std::{net::SocketAddr, sync::Arc};
 
 #[tokio::main]
 async fn main() {
+    tracing_subscriber::fmt::init();
     let orderbook = Arc::new(OrderBook::default());
     let filter = api::handle_all_routes(orderbook);
-    let result = warp::serve(filter).bind(([0, 0, 0, 0], 8080)).await;
-    println!("warp exited: {:?}", result);
+    let address = SocketAddr::new([0, 0, 0, 0].into(), 8080);
+    tracing::info!(%address, "serving order book");
+    warp::serve(filter).bind(address).await;
+    tracing::error!("warp exited");
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,6 +20,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
+dependencies = [
+ "winapi 0.3.9",
+]
+
+[[package]]
 name = "anyhow"
 version = "1.0.35"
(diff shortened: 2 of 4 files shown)


josojo:
Happy to use it, though, I don't know it at all.

orderbook/src/main.rs
```diff
@@ -2,12 +2,15 @@ mod api;
...
 async fn main() {
+    tracing_subscriber::fmt::init();
     let orderbook = Arc::new(OrderBook::default());
     let filter = api::handle_all_routes(orderbook);
-    let result = warp::serve(filter).bind(([0, 0, 0, 0], 8080)).await;
-    println!("warp exited: {:?}", result);
+    let address = SocketAddr::new([0, 0, 0, 0].into(), 8080);
+    tracing::info!(%address, "serving order book");
+    warp::serve(filter).bind(address).await;
+    tracing::error!("warp excited");
```
bh2smith, line 15: Is this excited or exited?
```suggestion
    tracing::error!("warp exited");
```

bh2smith (approved):
Just one typo

orderbook/src/main.rs
e00E: Wait, is that how you spell the past tense of `to exit`? I thought it was `exited` and `excited` would be the adjective.

fleupold (commented):
For v1 we did a bunch of customisations for the logging (e.g. forwarding error to stderr, combining multiline log messages into a single line for kibana, panic hook to get those messages logged), which were needed for operations.

Are they easy to add with the tracing module? If so, can we set them up in this PR? Or is it worthwhile to reuse the logging module we have from v1?

e00E:
For some of these I'm not sure if they are the best solution to the problem they're trying to solve so I don't want to blindly copy them. Although I do think we can implement all of those. On the off chance we cannot it would not be hard to get back to `log`.

The panic one definitely makes sense so I can add it to this PR.

fleupold:
> Although I do think we can implement all of those.

For the ones you don't implement here, could you please create a task so we can track these unpleasant but important work items.

nlordell:
:tada: I think `tracing` crate is more appropriate than simple `log`-ing for this kind of project. For the unconvinced:
> In asynchronous systems like Tokio, interpreting traditional log messages can often be quite challenging. Since individual tasks are multiplexed on the same thread, associated events and log lines are intermixed making it difficult to trace the logic flow. tracing expands upon logging-style diagnostics by allowing libraries and applications to record structured events with additional information about temporality and causality — unlike a log message, a span in tracing has a beginning and end time, may be entered and exited by the flow of execution, and may exist within a nested tree of similar spans. In addition, tracing spans are structured, with the ability to record typed data as well as textual messages.

Its basically structured logging well-suited to asynchronous applications. Its by the Tokio-rs team and has become widely used.

nlordell (approved):

e00E:
Considering we don't have logging at all yet, I am going to merge this as is and as Felix said create an issue to track further features.

PR #85, opened 2020-12-07
Remove expired orders in order book

e00E:
### Test Plan
New unit test

### DIFF
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -1,3 +1,5 @@
+use std::time::SystemTime;
+
 use model::{Order, OrderCreation, OrderMetaData, OrderUid};
 use tokio::sync::RwLock;
 
@@ -28,7 +30,9 @@ pub struct OrderBook {
 
 impl OrderBook {
     pub async fn add_order(&self, order: OrderCreation) -> Result<OrderUid, AddOrderError> {
-        // TODO: Check order signature, nonce, valid_to.
+        if !has_future_valid_to(now_in_epoch_seconds(), &order) {
+            return Err(AddOrderError::PastValidTo);
+        }
         let mut orders = self.orders.write().await;
         if orders.iter().any(|x| x.order_creation == order) {
             return Err(AddOrderError::DuplicatedOrder);
@@ -54,6 +58,28 @@ impl OrderBook {
             Err(RemoveOrderError::DoesNotExist)
         }
     }
+
+    // Run maintenance tasks like removing expired orders.
+    pub async fn run_maintenance(&self) {
+        self.remove_expired_orders(now_in_epoch_seconds()).await;
+    }
+
+    async fn remove_expired_orders(&self, now_in_epoch_seconds: u64) {
+        // TODO: use the timestamp from the most recent block instead?
+        let mut orders = self.orders.write().await;
+        orders.retain(|order| has_future_valid_to(now_in_epoch_seconds, &order.order_creation));
+    }
+}
+
+fn now_in_epoch_seconds() -> u64 {
+    SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .expect("now earlier than epoch")
+        .as_secs()
+}
+
+fn has_future_valid_to(now_in_epoch_seconds: u64, order: &OrderCreation) -> bool {
+    order.valid_to as u64 > now_in_epoch_seconds
 }
 
 struct InvalidSignatureError;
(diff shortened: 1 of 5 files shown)


orderbook/src/orderbook.rs
```diff
@@ -51,6 +53,25 @@ impl OrderBook {
...
+    pub async fn run_maintenance_forever(&self) -> ! {
+        loop {
+            tracing::debug!("running order book maintenance");
+            self.remove_expired_orders(SystemTime::now()).await;
+            tokio::time::delay_for(MAINTENANCE_INTERVAL).await;
+        }
+    }
+
+    async fn remove_expired_orders(&self, now: SystemTime) {
+        // TODO: use the timestamp from the most recent block instead?
```
e00E, line 67: This is a discussion point. If we prefer using the block then I still want to merge this PR as is and do that change in another PR where I would also add web3 as a dependency.l

orderbook/src/orderbook.rs
```diff
@@ -51,6 +53,25 @@ impl OrderBook {
             Err(RemoveOrderError::DoesNotExist)
         }
     }
+
+    // Clean up operations like removing expired orders.
+    pub async fn run_maintenance_forever(&self) -> ! {
+        loop {
+            tracing::debug!("running order book maintenance");
+            self.remove_expired_orders(SystemTime::now()).await;
```
fleupold, line 61: I think it would be good to separate out the way we decide when to do maintenance (loop with sleep vs. reacting on newBlockEvents stream) from the actual maintenance we are doing (removing expired orders, checking for updated allowances, etc).

Could there be a maintenance scheduler that just invokes orderbook.do_maintenance?

fleupold (commented):
Unfortunately the PR doesn't have a meaningful description.

Is this supposed to be the only place where we filter expired orders or are you planning to do it synchronously on the read path as well?

What other things are supposed to be added to the maintenance? Is this where allowance and fillment status checks are supposed to happen?

orderbook/src/orderbook.rs
e00E: That's definitely something we could do but I don't want to over engineer now. At this point it just a sleep now so introducing a maintenance scheduler is overkill.

e00E:
> Is this supposed to be the only place where we filter expired orders or are you planning to do it synchronously on the read path as well?

Should filter there directly too, will add that to this PR.

> What other things are supposed to be added to the maintenance? Is this where allowance and fillment status checks are supposed to happen?

Probably, that is why I called it maintenance. But we'll see what the best place for them is when we actually implement them.

orderbook/src/orderbook.rs
josojo: I think the local time is perfectly fine.

orderbook/src/orderbook.rs
```diff
@@ -51,6 +53,25 @@ impl OrderBook {
...
+    }
+
+    async fn remove_expired_orders(&self, now: SystemTime) {
+        // TODO: use the timestamp from the most recent block instead?
+        let now = now
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("now earlier than epoch")
+            .as_secs();
+        let mut orders = self.orders.write().await;
+        orders.retain(|order| order.order_creation.valid_to as u64 > now);
```
josojo, line 73: I would add here some buffer. If the tx needs to be settled within the next second, we should also remove it, as there is no chance that it will be settled.

Maybe this buffer should be: MAINTENANCE_INTERVAL

josojo (approved):
Yeah, cool that you introduced a maintainance thread. Definitively needed.

I am fine with going as you suggested for now. But later one I would prefer to store the "active orders" and "historic orders" both within the orderbook system. The historic order will be needed for the gas-refunding etc..

orderbook/src/orderbook.rs
e00E: Makes sense from the solver's point of view but I'm not sure about the orderbook. I feel the orderbook shouldnt' delete too early and let the solver's decide what they think their minimal handling time is.

orderbook/src/orderbook.rs
fleupold: Can we move the `run_maintenance_forever` outside of orderbook into a separate place (I don't think it uses self atm). The orderbook will do a lot of things related to the orderbook data itself so scheduling the maintenance should really live somewhere else (I'm fine with leaving it a function for now).

e00E: Makes sense. I moved it into main since that is the only place where we use tokio directly to create tasks. If that function gets more complicated it can also go into it's own module.

bh2smith (approved):

orderbook/src/orderbook.rs
josojo: yeah, makes sense!

PR #86, opened 2020-12-07
Remove allocation from uid creation

e00E:
The previous version was creating an intermediate vector which isn't
necessary.

### Test Plan
The function isn't tested yet because we do not decode and check the signature.


fleupold (approved):
nice!

bh2smith (approved):

PR #89, opened 2020-12-08
order api for posting orders

josojo:
Order api for posting: returns the values according to spec PR. I just amended that we are returning everytime an errorType, as we have it anyways.

### Test Plan
unit tests only

### DIFF
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -4,11 +4,11 @@
 
 pub mod h160_hexadecimal;
 pub mod u256_decimal;
-
 use chrono::{offset::Utc, DateTime, NaiveDateTime};
 use primitive_types::{H160, H256, U256};
 use serde::{de, Deserialize, Serialize};
 use serde::{Deserializer, Serializer};
+use std::convert::TryInto;
 use std::fmt;
 
 #[derive(Eq, PartialEq, Clone, Copy, Debug, Deserialize, Serialize)]
@@ -55,6 +55,25 @@ pub struct OrderCreation {
 impl OrderCreation {
     pub fn token_pair(&self) -> Option<TokenPair> {
         TokenPair::new(self.buy_token, self.sell_token)
+    }
+    pub fn order_owner(&self) -> H160 {
+        // dummy implementation, waiting for Valentins PR
+        H160::zero()
+    }
+    pub fn order_digest(&self) -> H256 {
+        // dummy implementation, waiting for Valentins PR
+        H256::zero()
+    }
+    pub fn order_uid(&self) -> OrderUid {
+        let owner = self.order_owner();
+        let digest = self.order_digest();
+        let valid_to = self.valid_to.to_be_bytes();
+        OrderUid(
+            [digest.as_bytes(), owner.as_bytes(), &valid_to]
+                .concat()
+                .try_into()
+                .unwrap(),
+        )
     }
 }
 
--- a/orderbook/src/api/filter.rs
+++ b/orderbook/src/api/filter.rs
@@ -113,16 +113,16 @@ pub mod test_util {
         };
         let response = post().await;
         assert_eq!(response.status(), StatusCode::CREATED);
-        let body: handler::UidResponse = serde_json::from_slice(response.body()).unwrap();
+        let body: serde_json::Value = serde_json::from_slice(response.body()).unwrap();
 
-        assert_eq!(body, serde_json::from_value(expected_uid).unwrap());
+        assert_eq!(body, expected_uid);
         // Posting again should fail because order already exists.
         let response = post().await;
         assert_eq!(response.status(), StatusCode::BAD_REQUEST);
 
-        let body: handler::OrderPostError = serde_json::from_slice(response.body()).unwrap();
+        let body: serde_json::Value = serde_json::from_slice(response.body()).unwrap();
         let expected_error =
             json!({"errorType": "DuplicatedOrder", "description": "order already exists"});
-        assert_eq!(body, serde_json::from_value(expected_error).unwrap());
+        assert_eq!(body, expected_error);
     }
 }
--- a/orderbook/src/api/filter.rs
+++ b/orderbook/src/api/filter.rs
@@ -101,6 +101,7 @@ pub mod test_util {
         let orderbook = Arc::new(OrderBook::default());
         let filter = create_order(orderbook.clone());
         let order = OrderCreation::default();
+        let expected_uid = json!({"UID": "0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"});
         let post = || async {
             request()
                 .path("/api/v1/orders")
@@ -112,8 +113,16 @@ pub mod test_util {
         };
         let response = post().await;
         assert_eq!(response.status(), StatusCode::CREATED);
+        let body: handler::UidResponse = serde_json::from_slice(response.body()).unwrap();
+
+        assert_eq!(body, serde_json::from_value(expected_uid).unwrap());
         // Posting again should fail because order already exists.
         let response = post().await;
         assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+
+        let body: handler::OrderPostError = serde_json::from_slice(response.body()).unwrap();
+        let expected_error =
+            json!({"errorType": "DuplicatedOrder", "description": "order already exists"});
+        assert_eq!(body, serde_json::from_value(expected_error).unwrap());
     }
 }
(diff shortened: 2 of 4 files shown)


orderbook/src/orderbook.rs
```diff
@@ -1,16 +1,21 @@
...
 
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Debug, Eq, PartialEq, Serialize, Deserialize)]
 pub enum AddOrderError {
-    AlreadyExists,
+    DuplicatedOrder,
     InvalidSignature,
     #[allow(dead_code)]
-    PastNonce,
+    ForBidden,
```
bh2smith, line 10:
```suggestion
    Forbidden,
```

model/src/lib.rs
```diff
@@ -56,6 +56,25 @@ impl OrderCreation {
...
+    }
+    pub fn order_digest(&self) -> H256 {
+        // dummy implementation, waiting for Valentins PR
+        H256::zero()
+    }
+    pub fn order_uid(&self) -> OrderUid {
+        let owner = self.order_owner();
+        let digest = self.order_digest();
+        let valid_to = self.valid_to.to_be_bytes();
+        OrderUid(
```
bh2smith, line 71: Ubernit: I know we are also doing this in contracts but ID should suffice as meaning unique ID. Since that what an ID is. Explicitly stating uniqueness is redundant in my opinion.

orderbook/src/api/handler.rs
```diff
@@ -27,16 +27,81 @@ pub struct FeeRequestBody {
...
+            StatusCode::BAD_REQUEST,
+        ),
+        Err(AddOrderError::InvalidSignature) => (
+            warp::reply::json(&OrderPostError {
+                error_type: AddOrderError::InvalidSignature,
+                description: String::from("invalid signature"),
+            }),
+            StatusCode::BAD_REQUEST,
+        ),
+        Err(AddOrderError::ForBidden) => (
```
bh2smith, line 68:
```suggestion
        Err(AddOrderError::Forbidden) => (
```

bh2smith (approved):
Cool Error handling! Just a couple comments and a nit.

model/src/lib.rs
e00E: I don't have a strong opinion on the name but I do strongly feel that it should be the same name as in contracts. In contracts it is called uid (as you also mentioned) so we should keep the same name.

orderbook/src/api/filter.rs
```diff
@@ -112,8 +113,16 @@ pub mod test_util {
...
         let response = post().await;
         assert_eq!(response.status(), StatusCode::CREATED);
+        let body: handler::UidResponse = serde_json::from_slice(response.body()).unwrap();
+
+        assert_eq!(body, serde_json::from_value(expected_uid).unwrap());
         // Posting again should fail because order already exists.
         let response = post().await;
         assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+
+        let body: handler::OrderPostError = serde_json::from_slice(response.body()).unwrap();
```
e00E, line 123: I think the type of body should be `serde_json::Value` so that it can be directly compared to `expected_error` without first converting `expected_error` to OrderPostError.

orderbook/src/orderbook.rs
```diff
@@ -1,16 +1,21 @@
-use model::{Order, OrderCreation, OrderMetaData, OrderUid};
-use primitive_types::{H160, H256};
-use std::convert::TryInto;
+use model::{Order, OrderCreation, OrderMetaData};
+use serde::{Deserialize, Serialize};
 use tokio::sync::RwLock;
 
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Debug, Eq, PartialEq, Serialize, Deserialize)]
 pub enum AddOrderError {
```
e00E, line 6: It might be cleaner to remove Serialize, Deserialize here and turn `OrderPostError`'s `error_type: AddOrderError,` into `error_type: String`.  This is because the actual serialized error string is only relevant to the api (filter) and also because the orderbook might have slightly different errors than what appears in the api.
For example if we wanted to handle TooManyRequests on the filter layer, we would still need it in the enum even though the OrderBook struct never creates it.

josojo: Yeah, I agree. I also took out the TooManyRequests error out of the struct.

model/src/lib.rs
josojo: I will not change it, unless it is changed in the openAPI documentation PR. If you guys feel strongly about it, please raise it there.

orderbook/src/api/handler.rs
```diff
@@ -27,16 +27,74 @@ pub struct FeeRequestBody {
...
+        ),
+        Err(AddOrderError::InsufficientFunds) => (
+            warp::reply::json(&OrderPostError {
+                error_type: String::from("InsufficientFunds"),
+                description: String::from(
+                    "order owner must have funds worth at least x in his account",
+                ),
+            }),
+            StatusCode::BAD_REQUEST,
+        ),
```
e00E, line 97: Suggestion for the code to repeat itself less:

```suggestion
        Err(err) => {
            let (error_type, description, status_code) = match err {
                AddOrderError::DuplicatedOrder => (
                    "DuplicatedOrder",
                    "order already exists",
                    StatusCode::BAD_REQUEST,
                ),
                AddOrderError::InvalidSignature => (
                    "InvalidSignature",
                    "invalid signature",
                    StatusCode::BAD_REQUEST,
                ),
                AddOrderError::Forbidden => (
                    "Forbidden",
                    "Forbidden, your account is deny-listed",
                    StatusCode::FORBIDDEN,
                ),
                AddOrderError::PastValidTo => (
                    "PastValidTo",
                    "validTo is in the past",
                    StatusCode::BAD_REQUEST,
                ),
                AddOrderError::MissingOrderData => (
                    "MissingOrderData",
                    "at least 1 field of orderCreation is missing",
                    StatusCode::BAD_REQUEST,
                ),
                AddOrderError::InsufficientFunds => (
                    "InsufficientFunds",
                    "order owner must have funds worth at least x in his account",
                    StatusCode::BAD_REQUEST,
                ),
            };
            let error = OrderPostError {
                error_type: error_type.to_string(),
                description: description.to_string(),
            };
            (warp::reply::json(&error), status_code)
        }
```

e00E (approved):

orderbook/src/api/handler.rs
josojo: Awesome thanks!

orderbook/src/api/handler.rs
```diff
@@ -27,16 +27,69 @@ pub struct FeeRequestBody {
...
     order: OrderCreation,
 ) -> Result<impl warp::Reply, Infallible> {
     let (body, status_code) = match orderbook.add_order(order).await {
-        Ok(()) => ("ok", StatusCode::CREATED),
-        Err(AddOrderError::AlreadyExists) => ("already exists", StatusCode::BAD_REQUEST),
-        Err(AddOrderError::InvalidSignature) => ("invalid signature", StatusCode::BAD_REQUEST),
-        Err(AddOrderError::PastNonce) => ("nonce is in the past", StatusCode::BAD_REQUEST),
-        Err(AddOrderError::PastValidTo) => ("validTo is in the past", StatusCode::BAD_REQUEST),
+        Ok(()) => (
+            warp::reply::json(&UidResponse {
```
fleupold, line 49: nit: I think currently the spec says order placement should return a string and not a dictionary

orderbook/src/orderbook.rs
```diff
@@ -1,15 +1,18 @@
...
 #[derive(Debug, Eq, PartialEq)]
 pub enum AddOrderError {
-    AlreadyExists,
+    DuplicatedOrder,
     InvalidSignature,
     #[allow(dead_code)]
-    PastNonce,
+    Forbidden,
+    #[allow(dead_code)]
+    MissingOrderData,
```
fleupold, line 11: Would be nice if this could take an argument with the (first) field name that is missing.

orderbook/src/api/handler.rs
josojo: dang it: it says:
```
application/json
```
https://github.com/gnosis/gp-v2-services/pull/49/files#diff-5016bdc926b6c0f649f363739d09159a1db664aae19a08fa6b38b03406f040dcR16

Hence, I thought that spec says we should return a json. but you are right that the type of UID is just a String. I will change it, sure.

nlordell: In that case shouldn't it return a JSON encoded string, so `"0x..."`?

orderbook/src/orderbook.rs
josojo: Yeah, I agree. I tried it for over 1 hour, but my rust knowledge is too limited. Hence, I am giving up and focusing on unblocking the front-end.

PR #90, opened 2020-12-08
Trivial, changing feeRatio to 0 for demo

josojo:


bh2smith (approved):
Nice!

e00E (approved):
This looks like something we will make configurable with command line arguments but until then this is fine.

PR #91, opened 2020-12-08
[Easy] Add GPv2Contract artefact

fleupold:
This PR adds the dependencies for GPv2 Settlement. Note, that part of the change accidentally slipped in to a hot fix I applied last week when trying to get autodeploy running (https://github.com/gnosis/gp-v2-services/commit/72138783bb76eccfb7662132ea19d236ab08a50e)

Given the setup of hardhat (which has separate jsons for each network) we are currently specifically importing the rinkeby artefact. We also cannot automatically load the deployed address as hardhat's .json is slightly different from truffle's. Given that this is the same for the Uniswap contract it might be ok to simply define the contract addresses in code somewhere (there is also https://github.com/gnosis/ethcontract-rs/issues/382 tracking ethers support in ethcontracts).

### Test Plan
Depending on the new contract in main works (if we add contracts to the solver's Cargo.toml)


bh2smith (approved):

bh2smith:
Seems legit!

e00E (approved):
Another way to test this and also useful for development is checking the docs:

```
cargo doc --no-deps --open
```

then navigate to or search for `g_pv_2_settlement::Contract`.

Importing the rinkeby version and specifying the address through command line arguments is fine for the demo.

PR #93, opened 2020-12-08
Verify order signatures

e00E:
and allow creation of signed orders for tests.
Resides in the model crate because this will be useful for the order
book and the solver.
Also moved uid creation to model for the same reason.

### Test Plan
new unit tests

### DIFF
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -58,58 +58,15 @@ pub struct OrderCreation {
 }
 
 impl OrderCreation {
-    pub const ORDER_TYPE_HASH: [u8; 32] =
-        hex!("b71968fcf5e55b9c3370f2809d4078a4695be79dfa43e5aa1f2baa0a9b84f186");
-
     pub fn token_pair(&self) -> Option<TokenPair> {
         TokenPair::new(self.buy_token, self.sell_token)
     }
 
-    pub fn order_digest(&self) -> [u8; 32] {
-        let mut hash_data = [0u8; 320];
-        hash_data[0..32].copy_from_slice(&Self::ORDER_TYPE_HASH);
-        // Some slots are not assigned (stay 0) because all values are extended to 256 bits.
-        hash_data[44..64].copy_from_slice(self.sell_token.as_fixed_bytes());
-        hash_data[76..96].copy_from_slice(self.buy_token.as_fixed_bytes());
-        self.sell_amount.to_big_endian(&mut hash_data[96..128]);
-        self.buy_amount.to_big_endian(&mut hash_data[128..160]);
-        hash_data[188..192].copy_from_slice(&self.valid_to.to_be_bytes());
-        hash_data[220..224].copy_from_slice(&self.app_data.to_be_bytes());
-        self.fee_amount.to_big_endian(&mut hash_data[224..256]);
-        hash_data[287] = match self.order_kind {
-            OrderKind::Sell => 0,
-            OrderKind::Buy => 1,
-        };
-        hash_data[319] = self.partially_fillable as u8;
-        signing::keccak256(&hash_data)
-    }
-
-    pub fn signing_digest_typed_data(&self, domain_separator: &[u8; 32]) -> [u8; 32] {
-        let mut hash_data = [0u8; 66];
-        hash_data[0..2].copy_from_slice(&[0x19, 0x01]);
-        hash_data[2..34].copy_from_slice(domain_separator);
-        hash_data[34..66].copy_from_slice(&self.order_digest());
-        signing::keccak256(&hash_data)
-    }
-
-    pub fn signing_digest_message(&self, domain_separator: &[u8; 32]) -> [u8; 32] {
-        let mut hash_data = [0u8; 92];
-        hash_data[0..28].copy_from_slice(b"\x19Ethereum Signed Message:\n64");
-        hash_data[28..60].copy_from_slice(domain_separator);
-        hash_data[60..92].copy_from_slice(&self.order_digest());
-        signing::keccak256(&hash_data)
-    }
-
-    pub fn signing_digest(&self, domain_separator: &[u8; 32]) -> [u8; 32] {
-        if self.signature.v & 0x80 == 0 {
-            self.signing_digest_typed_data(domain_separator)
-        } else {
-            self.signing_digest_message(domain_separator)
-        }
-    }
-
     // If signature is valid returns the owner.
     pub fn validate_signature(&self, domain_separator: &[u8; 32]) -> Option<H160> {
+        // The signature related functionality is defined by the smart contract:
+        // https://github.com/gnosis/gp-v2-contracts/blob/main/src/contracts/libraries/GPv2Encoding.sol
+
         let v = self.signature.v & 0x1f;
         let message = self.signing_digest(domain_separator);
         let recovery = Recovery::new(message, v as u64, self.signature.r, self.signature.s);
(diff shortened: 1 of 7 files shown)


josojo (approved):
Very nice, this is valuable for the next steps. Thanks a lot for doing it so quickly.

Also, like that you are offering both from the start, signing typed and signing message.

Let's merge it quickly, I think it is important for the front-end guys

model/src/lib.rs
```diff
@@ -52,26 +58,97 @@ pub struct OrderCreation {
...
+            self.signing_digest_message(domain_separator)
+        }
     }
-    pub fn order_digest(&self) -> H256 {
-        // dummy implementation, waiting for Valentins PR
-        H256::zero()
+
+    // If signature is valid returns the owner.
+    pub fn validate_signature(&self, domain_separator: &[u8; 32]) -> Option<H160> {
+        let v = self.signature.v & 0x1f;
```
fleupold, line 113: why this? Could you please add a comment?

model/src/lib.rs
```diff
@@ -52,26 +58,97 @@ pub struct OrderCreation {
...
+        self.fee_amount.to_big_endian(&mut hash_data[224..256]);
+        hash_data[287] = match self.order_kind {
+            OrderKind::Sell => 0,
+            OrderKind::Buy => 1,
+        };
+        hash_data[319] = self.partially_fillable as u8;
+        signing::keccak256(&hash_data)
+    }
+
+    pub fn signing_digest_typed_data(&self, domain_separator: &[u8; 32]) -> [u8; 32] {
```
fleupold, line 87: can these be private? I would think only `validate_signature` and `uid` need to be public?

model/src/lib.rs
```diff
@@ -52,26 +58,97 @@ pub struct OrderCreation {
...
-        uid.0[32..52].copy_from_slice(self.order_owner().as_fixed_bytes());
+        uid.0[0..32].copy_from_slice(&self.order_digest());
+        uid.0[32..52].copy_from_slice(owner.as_fixed_bytes());
         uid.0[52..56].copy_from_slice(&self.valid_to.to_be_bytes());
         uid
     }
 }
 
+// Intended to be used by tests that need signed orders.
+impl OrderCreation {
```
fleupold, line 130: nit: mark it to be only compiled under under test?

model/src/lib.rs
```diff
@@ -293,4 +370,69 @@ mod tests {
         let token = H160::from_low_u64_be(1);
         assert_eq!(TokenPair::new(token, token), None);
     }
+
+    #[test]
+    fn signature_typed_data() {
+        // copied from a gp-v2-contracts test
```
fleupold, line 376: ubernit: Could you add the source link?

model/src/lib.rs
```diff
@@ -293,4 +370,69 @@ mod tests {
...
+            },
+        };
+        let expected_owner = hex!("70997970C51812dc3A010C7d01b50e0d17dc79C8");
+        let owner = order.validate_signature(&domain_separator).unwrap();
+        assert_eq!(owner, expected_owner.into());
+    }
+
+    #[test]
+    fn sign_self() {
+        let mut order = OrderCreation::default();
```
fleupold, line 431: I'd slightly prefer if we used UserOrder::default and in there create a real signature that is valid. Creating an invalid order creation and then mutating it into something allows to create semantically invalid data (not so important for the tests but also alluding to my general preference of checking signature at order creation time).

fleupold (approved):
I think it would be nice if signature verification could happen on order creation so we disallow invalid orders to be created at all.

nlordell (approved):
Implementation looks correct! My only comments are:
- Maybe things like computing the domain separator, type hash and struct hash could maybe be under a `eip712` module just so its SUPER clear the signing standard that is being used?
- Maybe some functionality from `ethabi` can be reused for encoding the order for hashing to make the purpose clearer (although, maybe its just impractical, I don't remember the API too well).

model/src/lib.rs
nlordell: Its because the signing scheme is encoded in the signature. Maybe this can be turned into an abstraction so that its clearer? Something like:
```
enum SigningScheme {
    Eip712TypedData,
    Message,
}
```

bh2smith (approved):
This is really cool. I agree with all the suggestions already made. In particular it would really clean up the code of the decoding was done with some existing library rather than manually here.

model/src/lib.rs
```diff
@@ -52,26 +58,97 @@ pub struct OrderCreation {
 }
 
 impl OrderCreation {
+    pub const ORDER_TYPE_HASH: [u8; 32] =
```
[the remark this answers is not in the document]
e00E: Don't see how lazy_static helps but I'm adding a comment here:

>  // This constant and the signature related functions come from
    // https://github.com/gnosis/gp-v2-contracts/blob/main/src/contracts/libraries/GPv2Encoding.sol .

model/src/lib.rs
e00E: With a comment linking GPv@Encoding.sol I don't think we have to explain this in more detail here. This code doesn't make any design decisions. It just copies the implementation in the smart contract.

model/src/lib.rs
e00E: Made private

model/src/lib.rs
e00E: No because it wouldn't exist for the tests of dependent crates.

model/src/lib.rs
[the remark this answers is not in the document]
nlordell: As in you can compute this value, it's the EIP-712 type hash for the Order (the `keccak256` expression from the comment above the link)

model/src/lib.rs
nlordell: Personally I would prefer if the implementation referenced the EIP-712 which is the signing standard used instead of the SC (which just adds the detail of also supporting wallets that don't support signing typed data) maybe using the JS code from the same repo as reference (as that is what the solidity code is tested against). My main argument being that this would make the Rust code more expressive and self-documenting than if it just ported the Solidity implementation, which is a bit low level to get around Solidity/EVM limitations and not necessarily because it is the cleanest implementation possible. In fact, it could be implemented as `keccack256(abi.encode(...))` if we weren't going for gas optimizations.

model/src/lib.rs
e00E: Addressing this and 

> I think it would be nice if signature verification could happen on order creation so we disallow invalid orders to be created at all.

This type was intended to implement the serialization format. I can see some merit in having a checked `CorrectlySignedOrder` but I don't think enforcing that for this type meshes well with the serialization aspect. Enforcing that isn't as simple as it seems so I don't want to include it in this PR. I made an issue to discuss this in https://github.com/gnosis/gp-v2-services/issues/96 .

model/src/lib.rs
```diff
@@ -52,26 +58,97 @@ pub struct OrderCreation {
...
     pub fn token_pair(&self) -> Option<TokenPair> {
         TokenPair::new(self.buy_token, self.sell_token)
     }
-    pub fn order_owner(&self) -> H160 {
-        // dummy implementation, waiting for Valentins PR
-        H160::zero()
+
+    pub fn order_digest(&self) -> [u8; 32] {
+        let mut hash_data = [0u8; 320];
+        hash_data[0..32].copy_from_slice(&Self::ORDER_TYPE_HASH);
```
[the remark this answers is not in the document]
e00E: Made an issue for this because I don't want to bog down this PR with it. When I implemented this I found it easier to set the bytes directly but agree we can see if there is an easy way to reuse those crates.

model/src/lib.rs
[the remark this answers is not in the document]
e00E: I see. Coming from the idea that signing is explained in some textual spec that we implement that makes more sense. On the other hand I approached this whole signature algorithm from the smart contract side in which case it makes sense to have the constant here directly. Not sure what is better.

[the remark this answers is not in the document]
e00E: I'm going to add this and your other comment below to the issue of reusing web3 or ethabi. Then we have one issue to track converting this to a more high level implementation but we can still merge the PR now so that we have this functionality. It's more of a cleanup issue then.

model/src/lib.rs
nlordell: I don't want to block merging this PR. I imagine this is also in some ways related to refactoring and #96 and #97 so we can merge as is IMO.

e00E:
I am going to merge this now so we get this in quickly for the frontend. There are two follow up issues for potential future changes.

PR #94, opened 2020-12-09
On fee request: use sellToken from path

josojo:
Correcting feedback from PR: https://github.com/gnosis/gp-v2-services/pull/72

### Test Plan
Unit tests

### DIFF
--- a/orderbook/src/api.rs
+++ b/orderbook/src/api.rs
@@ -8,7 +8,17 @@ use warp::Filter;
 pub fn handle_all_routes(
     orderbook: Arc<OrderBook>,
 ) -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
-    filter::create_order(orderbook.clone())
-        .or(filter::get_orders(orderbook))
-        .or(filter::get_fee_info())
+    let order_creation = filter::create_order(orderbook.clone());
+    let order_getter = filter::get_orders(orderbook);
+    let fee_info = filter::get_fee_info();
+
+    let label = |label: &'static str| warp::any().map(move || label);
+    let routes_with_labels = warp::path!("api" / "v1" / ..).and(
+        (label("order_creation").and(order_creation))
+            .or(label("order_getter").and(order_getter))
+            .unify()
+            .or(label("fee_info").and(fee_info))
+            .unify(),
+    );
+    warp::any().and(routes_with_labels.map(|_, result| result))
 }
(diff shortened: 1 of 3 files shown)


bh2smith (approved):
From what I can tell this is correcting the path used in the previous PR to the suggestion made in https://github.com/gnosis/gp-v2-services/pull/72#discussion_r536242551. Seems fine, but what happened to the `extract_sell_token` method that was "rejecting large payloads"?

josojo:
>but what happened to the extract_sell_token method that was "rejecting large payloads"?

well, we don't have payloads anymore :) its all part of the path and protecting against too long parts is not worth it IMO

nlordell:
> its all part of the path and protecting against too long parts is not worth it IMO

In general I would also have configured this as part of the pod's ingress NxinX configuration instead of having it in code (which can still be done).

orderbook/src/api.rs
```diff
@@ -8,7 +8,17 @@ use warp::Filter;
...
+
+    let label = |label: &'static str| warp::any().map(move || label);
+    let routes_with_labels = warp::path!("api" / "v1" / ..).and(
+        (label("order_creation").and(order_creation))
+            .or(label("order_getter").and(order_getter))
+            .unify()
+            .or(label("fee_info").and(fee_info))
+            .unify(),
+    );
+    warp::any().and(routes_with_labels.map(|_, result| result))
```
e00E, line 23: The whole label and unify thing was only for the prometheus metrics. Since we don't have those yet I wouldn't make this more complicated than it needs to be now.

```suggestion
    warp::path!("api" / "v1" / ..).and(order_creation.or(order_getter).or(fee_info))
```

orderbook/src/api/filter.rs
```diff
@@ -18,16 +20,10 @@ fn extract_user_order() -> impl Filter<Extract = (OrderCreation,), Error = warp:
...
-    // (rejecting huge payloads)...
-    warp::body::content_length_limit(MAX_JSON_BODY_PAYLOAD).and(warp::body::json())
-}
-
 pub fn create_order(
     orderbook: Arc<OrderBook>,
-) -> impl Filter<Extract = impl warp::Reply, Error = warp::Rejection> + Clone {
-    warp::path!("api" / "v1" / "orders")
+) -> impl Filter<Extract = (WithStatus<Json>,), Error = warp::Rejection> + Clone {
+    warp::path!("orders")
```
e00E, line 26: All of the `Extract = impl warp::Reply` -> `Extract = (WithStatus<Json>,)` aren't necessary`.

orderbook/src/api.rs
josojo: THanks, yes. I should have studied it better. thanks

e00E (approved):

PR #95, opened 2020-12-09
Returning UID as string

josojo:
... and not as a json object. Good catch from Felix.

Nick mentioned that we might even should use the "0x" prefix as it as specified like that, but I think we should not do it, as we don't do it in any other place.

### DIFF
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -3,8 +3,10 @@ info:
   version: 0.0.1
   title: Order Book API
 servers:
-- url: http://localhost:8080
-  description: Local
+  - url: https://protocol-rinkeby.dev.gnosisdev.com
+    description: Rinkeby (Staging)
+  - url: http://localhost:8080
+    description: Local
 paths:
   /api/v1/orders:
     post:
--- a/orderbook/src/api/filter.rs
+++ b/orderbook/src/api/filter.rs
@@ -101,7 +101,7 @@ pub mod test_util {
         let orderbook = Arc::new(OrderBook::default());
         let filter = create_order(orderbook.clone());
         let order = OrderCreation::default();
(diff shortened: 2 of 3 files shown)


nlordell:
> "0x" prefix as it as specified like that, but I think we should not do it, as we don't do it in any other place.

I think we should use this in general (as in here and in other places as well) as it is the notation used in the [Ethereum JSON RPC specification](https://eth.wiki/json-rpc/API#hex-value-encoding). Specifically it states:

> When encoding UNFORMATTED DATA (byte arrays, account addresses, hashes, bytecode arrays): encode as hex, prefix with “0x”, two hex digits per byte. Examples:
> [...]
> - WRONG: 004200 (must be prefixed 0x)

e00E:
Maybe not the best place for this discussion but I disagree. The api is for developers, not users. The frontend can still let users enter things with the 0x prefix. But for developers it is simpler to leave it and this is what many libraries expect. Iirc for the previous api the frontend was manually adding the prefix and we were manually stripping it.

nlordell:
> The api is for developers, not users

I completely agree, My point is that if **developers** use these responses for `web3` interactions, then they will need to add the prefix (since the Web3 JSON RPC spec requires it). IMO the ecosystem has come to expect bytes to be encoded as `"0x..."` JSON strings since most other APIs use this. My comment has nothing to do with how the FE chooses to present this data.

> this is what many libraries expect

I think this is true for Rust crates, but not for Ethereum JSON APIs in general. From what I've seen, the `0x` prefix is ubiquitously used in JSON for hex-encoded bytes in the Ethereum ecosystem (the `rust-web3` crate also uses this encoding for example for its `serde::(Des|S)erialize` implementations).

nlordell (approved):

Issue #98, opened 2020-12-09
Logging Features


e00E:
1. [ ] create logging crate for shared setup between binaries
2. [ ] error log panics
3. [ ] log to stderr based on severity
4. [ ] combining multiline log messages into a single line

I don't think 4 is needed anymore but we need to make sure that the log separator matches the logs we emit.
For 3 I am wondering whether this is the correct approach and if we shouldn't stay on stdout but make sure that warn and error severity is properly detected. Iirc in v1-services we introduced this to detect errors for the docker-compose e2e tests but maybe there is a better solution.

fleupold:
>  if we shouldn't stay on stdout but make sure that warn and error severity is properly detected

For kibana it was a practical way to sort logs by severity (otherwise it would have required a logging format dependent regex). Do you have a concrete concern with this approach? If I understand unix file descriptors correctly stderr is intended precisely for that purpose.

e00E:
It's more that in traditional linux you often chain commands together. So the data should stay in stdout and any diagnostic information like "retrying connection to server" would go into stderr. For our purposes this is irrelevant so we can use whatever.
My feeling is that parsing the log string is a little more correct than using two streams which feels more like a workaround because we're changing our binary to get around limitations of the stagaing setup.
`log:"ERROR"` had the problem that it detected the word error anywhere in the message. I'll see if there is a fix for that.

e00E:
This is all implemented except for logging to stderr. Instead of  doing that, I set up a better log filter in staging that makes sure the `ERROR` part is in the correct place after the timestamp. This prevents false positives if `ERROR` shows up some place else in the log message.

PR #100, opened 2020-12-09
Setup tracing in shared crate

e00E:
with the proper log separator timestamp and panic handler.

part of https://github.com/gnosis/gp-v2-services/issues/98

### Test Plan
add panic to orderbook/main.rs and see that log messages is generated.

### DIFF
--- a/solver/Cargo.toml
+++ b/solver/Cargo.toml
@@ -15,4 +15,6 @@ primitive-types = "0.7"
 reqwest = { version = "0.10", features = ["json"] }
 serde_json = "1.0"
 tokio = { version = "0.2", features =[ "macros", "time"] }
+tracing = "0.1"
+tracing-setup = { path = "../tracing-setup" }
 web3 = { version = "0.13", default-features = false }
--- a/tracing-setup/src/lib.rs
+++ b/tracing-setup/src/lib.rs
@@ -25,7 +25,7 @@ fn set_panic_hook() {
         let thread_name = thread.name().unwrap_or("<unnamed>");
         // It is not possible for our custom hook to print a full backtrace on stable rust. To not
         // lose this information we call the default panic handler which prints the full backtrace.
-        // The lpreceding log makes kibana consider this a multi line log message.
+        // The preceding log makes kibana consider this a multi line log message.
         tracing::error!("thread '{}' {}:", thread_name, info);
         default_hook(info);
     };
--- a/tracing-setup/src/lib.rs
+++ b/tracing-setup/src/lib.rs
@@ -1,0 +1,33 @@
+use std::{
+    panic::{self, PanicInfo},
+    thread,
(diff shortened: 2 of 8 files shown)


solver/Cargo.toml
```diff
@@ -15,4 +15,6 @@ primitive-types = "0.7"
 reqwest = { version = "0.10", features = ["json"] }
 serde_json = "1.0"
 tokio = { version = "0.2", features =[ "macros", "time"] }
+tracing = "0.1"
+tracing-setup = { path = "../tracing-setup" }
```
fleupold, line 19: Does it make sense to re-export the tracing macros and just depend our own logging crate (this way the concrete logging library would be abstracted away). Or is there too much tracing specific things that would make this abstraction too leaky?

tracing-setup/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+
+// Sets a panic hook so panic information is logged in addition to the default panic printer.
+fn set_panic_hook() {
+    let default_hook = panic::take_hook();
+    let hook = move |info: &PanicInfo| {
+        let thread = thread::current();
+        let thread_name = thread.name().unwrap_or("<unnamed>");
+        // It is not possible for our custom hook to print a full backtrace on stable rust. To not
+        // lose this information we call the default panic handler which prints the full backtrace.
+        // The lpreceding log makes kibana consider this a multi line log message.
```
fleupold, line 28:
```suggestion
        // The preceding log makes kibana consider this a multi line log message.
```

fleupold (approved):

solver/Cargo.toml
e00E: I thought about this too but I think it makes sense for the other crates to use tracing directly. Tracing has a couple of improvements in the macros over log so to use those we wouldn't be compatible anyway.

nlordell: One possibility is to add to `tracing-setup/src/lib.rs`
```rust
pub use tracing;
```

Then the dependencies would use it like:
```rust
use tracing_setup::tracing;

tracing::info!("");
```

This just ensures all versions of `tracing` stay the same throughout the workspace.

(I would probably also `s/tracing-setup/something-else-i-have-no-idea-how-to-name-things/` in this case as well).

nlordell: /cc @e00E as this PR is already marked for merging.

e00E: True that it trivially makes the version stay the same but has the disadvantage that the other crates have to `use` it first instead of doing `tracing::info!` directly which I like.

nlordell: > instead of doing tracing::info! directly which I like

I think that's a really good point especially for logging, as removing any barriers to being able to add logs is a good thing.

PR #103, opened 2020-12-09
Correcting the Domain separator for user_order_to_full_order

josojo:
Found a tiny bug, wrong domain separator was used.

### DIFF
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -105,12 +105,28 @@ impl OrderCreation {
         address
     }
 }
+pub struct InvalidSignatureError;
 
 // See https://github.com/gnosis/gp-v2-contracts/blob/main/src/contracts/libraries/GPv2Encoding.sol
 impl OrderCreation {
     const ORDER_TYPE_HASH: [u8; 32] =
         hex!("b71968fcf5e55b9c3370f2809d4078a4695be79dfa43e5aa1f2baa0a9b84f186");
 
+    pub fn order_creation_to_full_order(
+        self: OrderCreation,
+    ) -> Result<Order, InvalidSignatureError> {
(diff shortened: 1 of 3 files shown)


model/src/lib.rs
```diff
@@ -105,12 +105,28 @@ impl OrderCreation {
...
+pub struct InvalidSignatureError;
 
 // See https://github.com/gnosis/gp-v2-contracts/blob/main/src/contracts/libraries/GPv2Encoding.sol
 impl OrderCreation {
     const ORDER_TYPE_HASH: [u8; 32] =
         hex!("b71968fcf5e55b9c3370f2809d4078a4695be79dfa43e5aa1f2baa0a9b84f186");
 
+    pub fn order_creation_to_full_order(
+        self: OrderCreation,
+    ) -> Result<Order, InvalidSignatureError> {
```
e00E, line 117: I'd like to keep this in the orderbook crate because that's the only place where we do this operation.

model/src/lib.rs
```diff
@@ -105,12 +105,28 @@ impl OrderCreation {
...
+        let owner = self
+            .validate_signature(&Self::ORDER_TYPE_HASH)
```
e00E, line 119: You've changed this to order type hash but I don' think it's correct. The other function just has a todo comment about passing in the domain separator becauase it's something we will pass as a command line argument.

josojo: Are you sure?
I think Nick was using the order type hash [here]( https://github.com/gnosis/gp-v2-contracts/blob/main/src/contracts/libraries/GPv2Encoding.sol#L240) to calculate the order digest.

And the owner we are interested in is received from ec-recover of the message with this order digest.

PR #106, opened 2020-12-10
Set up command line argument parsing

e00E:
Using a helper crate to share common arguments between both binaries. Having to copy paste them was a gripe we had with v1-services.

### Test Plan
`cargo run --bin orderbook -- --help`
`cargo run --bin solver -- --help`

### DIFF
--- a/orderbook/Cargo.toml
+++ b/orderbook/Cargo.toml
@@ -10,6 +10,8 @@ model = { path = "../model" }
 primitive-types = "0.7"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+shared-arguments = { path = "../shared-arguments" }
+structopt = { version = "0.3" }
 tokio = { version = "0.2", features =[ "macros", "time"] }
 tracing = "0.1"
 tracing-setup = { path = "../tracing-setup" }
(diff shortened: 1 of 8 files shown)


fleupold (approved):

josojo (approved):
NIce

orderbook/Cargo.toml
```diff
@@ -10,6 +10,8 @@ model = { path = "../model" }
 primitive-types = "0.7"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+shared-arguments = { path = "../shared-arguments" }
+structopt = { version = "0.3" }
```
nlordell, line 14: Maybe arguments should
```
pub use structopt::StructOpt;
```

To guarantee the versions are compatible?

nlordell: @e00E In case this gets merged before you see this comment.

e00E: In this case I think that is better because we only use structopt in main.rs. Will make this change.

e00E: Sadly it doesn't work because of how the structopt derive macro is set up. It expects users to directly depend on structopt.

PR #107, opened 2020-12-10
Remove dependabot PR automerge

e00E:
We can discuss this in this PR.
Imo it is better if we have full control over what gets merged because
* it prevents the staging deployment from randomly restarting when we might not want it to
* there are some crate updates that even though they pass CI we might not want to apply. I feel more comfortable reading the changelog first.

### Test Plan
next time dependabot makes a mergable PR it shouldn't be automerged

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -10,13 +10,3 @@ pull_request_rules:
         method: squash
         strict: smart+fasttrack
         commit_message: title+body
-  - name: Automatic merge for Dependabot pull requests
-    conditions:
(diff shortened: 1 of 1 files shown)


fleupold (approved):
I think the reasoning makes sense. I personally found myself not checking the change logs and blindly merging version bumps (which felt tedious and as such I wanted to automate it).

But I'm happy if we want to apply more due-diligence on our dependency mgmt.

bh2smith (approved):

PR #109, opened 2020-12-10
Automatically detect whether logs should be colored

e00E:
In Kibana ansi color codes are not handled so they introduce random
characters to the logging output.

### Test Plan
`cargo run --bin orderbook` has colors
`cargo run --bin orderbook | less` does not

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2245,6 +2245,7 @@ dependencies = [
 name = "tracing-setup"
 version = "0.1.0"
 dependencies = [
+ "atty",
  "tracing",
  "tracing-subscriber",
 ]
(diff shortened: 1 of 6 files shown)


nlordell (approved):
I suspect that you can use https://docs.rs/atty/0.2.14/atty/ to determine the default value. If the the pod starts the estimator container without a TTY (which I expect to be true) then it will just automatically disable ANSI colours for the ouput.

Either way, I think being able to disable colours with a command line flag is very valuable, just a suggestion on how to improve the default value.

bh2smith (approved):

e00E:
Thanks for linking atty. I think I actually like removing the option and always doing it automatically more so that we have less command line arguments.

fleupold (approved):
nice!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2Fl41lPv6RuNMoRshnW%2Fgiphy.gif&token=chameleon" data-gitmeme-token="chameleon"></a>

PR #110, opened 2020-12-10
Pass domain separator as command line argument

e00E:
This has to be configurable for different deployments and us using the
correct value early helps the frontend.

Alex is working on a better version of  that uses the contract address and ethabi's abi encoding. When we have that we can remove the command line argument.

### Test Plan
`cargo run --bin orderbook` prints the domain separator

### DIFF
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -1,6 +1,6 @@
 use std::time::SystemTime;
 
-use model::{Order, OrderCreation, OrderMetaData, OrderUid};
+use model::{DomainSeparator, Order, OrderCreation, OrderMetaData, OrderUid};
 use tokio::sync::RwLock;
 
 #[derive(Debug, Eq, PartialEq)]
@@ -24,11 +24,19 @@ pub enum RemoveOrderError {
 
 #[derive(Debug, Default)]
 pub struct OrderBook {
+    domain_separator: DomainSeparator,
     // TODO: Store more efficiently (for example HashMap) depending on functionality we need.
-    pub orders: RwLock<Vec<Order>>,
+    orders: RwLock<Vec<Order>>,
 }
 
 impl OrderBook {
+    pub fn new(domain_separator: DomainSeparator) -> Self {
+        Self {
+            domain_separator,
+            orders: RwLock::new(Vec::new()),
+        }
+    }
+
     pub async fn add_order(&self, order: OrderCreation) -> Result<OrderUid, AddOrderError> {
         if !has_future_valid_to(now_in_epoch_seconds(), &order) {
             return Err(AddOrderError::PastValidTo);
@@ -37,7 +45,7 @@ impl OrderBook {
         if orders.iter().any(|x| x.order_creation == order) {
             return Err(AddOrderError::DuplicatedOrder);
         }
-        let order = user_order_to_full_order(order).map_err(|_| AddOrderError::InvalidSignature)?;
+        let order = self.order_creation_to_order(order)?;
         let uid = order.order_meta_data.uid;
         tracing::debug!(?order, "adding order");
         orders.push(order);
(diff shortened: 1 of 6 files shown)


josojo (approved):
Yes, I think it is the right approach to unblock front-end.

My rust is so rusty that it might take a little bit longer. Though, actually it is really not complicated and in js I would have coded it within 2 hours for sure :D 

I will rebase on this Pr...

orderbook/src/orderbook.rs
```diff
@@ -69,6 +77,20 @@ impl OrderBook {
         let mut orders = self.orders.write().await;
         orders.retain(|order| has_future_valid_to(now_in_epoch_seconds, &order.order_creation));
     }
+
+    fn user_order_to_full_order(&self, user_order: OrderCreation) -> Result<Order, AddOrderError> {
```
josojo, line 81: nit: you might wanna change the name reflecting that we don't have "user_orders" anymore

e00E:
So if this does depend on the contract address then maybe my approach doesn't make much sense (but we can still keep the DomainSeparator type). We would have to update the value in staging every time we redeploy contracts.
Maybe we could read the value from the contract? But I see it is internal at the moment.

fleupold:
> Maybe we could read the value from the contract? But I see it is internal at the moment.

Let's make it public. I think there are a few private fields which should in fact be publicly accessible for ease of use.

nlordell:
> Let's make it public.

So, while we can make this public, I think we should embrace EIP-712 and compute it in the service. From what I understand from the EIP, this is meant to be computed by the signer or verifyer and has a very well-defined definition.

fleupold:
> I think we should embrace EIP-712 and compute it in the service

In my view the motivation for EIP-712 is to make it more clear to the users what data they are signing. As such any frontend creating a digest to sign (e.g. Metamask), should make sure that when they prompt users with a preview of a structured message, the definition of the format matches with the contract that it is intended for (by manually recomputing the hash or).

Here, we are trying to recover the address from an existing signature. Since this very signature is later to be used with the settlement contract, I'd say its fine to read what domain separator the contract expects rather than computing it ourselves (I don't see an advantage in doing that and it requires more code).

nlordell:
> is to make it more clear to the users what data they are signing

To me the other aspect is intent - the signing Rust code would also explicitly define its expectations on the signing domain and type of the data that was signed and is being recovered.

> I don't see an advantage in doing that and it requires more code

Maybe it is a little overkill, but in my opinion this sort of thing *should* be something that is part of the Rust `web3` library (as it is in the JavaScript `web3` and `ethers` libraries), in which case I would find the code slightly more descriptive to have some call to `eip712::domain_separator(eip712::Domain { ... })` as it would more clearly link the code to the actual signing scheme being used.

The other thing I had thought about before is that computing the domain separator instead of requesting it means one less call that can fail on startup. However, this isn't true as the chain ID needs to be requested anyway.

I guess, I had just envisioned a `eip712` module that would compute all these things and make the calling code very clear on its intent and eventually possibly be added upstream to `web3` crate for others to use, but this is probably quite a bit of overkill.

josojo:
IMO, the domain separator should be public and each project building on gpv2 should decide themselves, whether they want to use it or compute it.

I, personally, would use the domain separator directly from the contract, as it is questionable whether the benefit is worth the additional effort.

nlordell:
> IMO, the domain separator should be public and each project building on gpv2 should decide themselves

Yes I completely agree.

> it is questionable whether the benefit is worth the additional effort

Since the service is only verifying these signatures and not signing, I think you guys are right and there is no real advantage to computing it vs querying it. I think I would feel more strongly about this if there was already a good Rust crate for EIP-712 operations as then I imagine the code to be extremely simple and, at least in my opinion, more clearly describe what the domain separator is exactly and how it should be used (as it would be tied to a technical specification - EIP-712).

fleupold:
> if there was already a good Rust crate for EIP-712

So much ecosystem, so few Fridays...

bh2smith (approved):
Looks good although I am still not entirely sure what the purpose of the domain separator is? Is this for evaluating anayltics like where the orders are coming from? Is it possible to use a domain separator that would cause the order to be invalid?

e00E:
This PR has been obsoleted as we aret calculating it automatically at start up now.

PR #111, opened 2020-12-10
Create a shared OrderbookReading trait between API and Driver

fleupold:
This is a suggestion that I'd like to get feedback on. By introducing a shared OrderbookReading trait we could use the in-memory Orderbook directly inside the driver (rather than going via the API using HTTP).

While I'm not sure the symmetry between the two concepts can be guaranteed in the long run (have not yet come up with a good example for why not) it would make e2e-testing the system much more direct: We could create a driver that is using the in-memory book directly into which we can insert an order. We can then call `driver.single_run()` inside the test and when the method returns we know we can check the smart contract if our orders where matched as expected.

Otherwise (if everything was run as separate processes) we would somehow have to "drive" the actual driver from the outside and might run into some timing/waiting issues since we don't tightly control the runloop. While this would likely be more e2e in that it actually tests everything it might also be more brittle and less deterministic.

Let me know what you think...

### Test Plan
CI for now...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1120,6 +1120,8 @@ dependencies = [
 name = "model"
 version = "0.1.0"
 dependencies = [
+ "anyhow",
+ "async-trait",
  "chrono",
  "hex",
  "hex-literal",
@@ -1252,6 +1254,7 @@ name = "orderbook"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "async-trait",
  "chrono",
  "model",
  "primitive-types",
--- a/model/Cargo.toml
+++ b/model/Cargo.toml
@@ -4,6 +4,8 @@ version = "0.1.0"
 edition = "2018"
 
 [dependencies]
+anyhow = "1.0"
+async-trait = "0.1"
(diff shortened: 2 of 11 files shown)


e00E:
I see 3 styles of e2e tests:

1. Complete separation: Start the orderbook and driver binaries as they run in the real deployment.
2. Same process with http: Test creates orderbook and solver instances and runs them both itself. This allows the test full control like calling `single_run`.
3. In memory orderbook from this PR

What I like about 2 over 3 is that it still tests most of the code running the real orderbook. That way we could also manually call `maintenance` and check that the orderbook updates the status of a matched order or removes it.

fleupold:
Agree that 2 & 3 make the e2e tests easier to control, run and debug. 2 gives us slightly more test surface (it tests the most outward facing API of the orderbook) which I think is nice. It's slightly more effort (tests have to deserialize orders and do POST requests vs invoking methods with the pre-made model) and I find it hard to say yet if there will be pieces between the Orderbook and the outward facing API that we won't care about testing.

Assuming you also prefer 2) I will go for this one.

e00E:
Yes, 2 sounds good.

PR #112, opened 2020-12-10
Implement Display for OrderUid

e00E:
This impliclity also provided `to_string` and can be reused in the
Serialize impl.

This came up in a conversation with Alex because he needs to convert the uid to a string for an api route.

### Test Plan
New unit test

### DIFF
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -4,13 +4,14 @@
 
 pub mod h160_hexadecimal;
 pub mod u256_decimal;
+
 use chrono::{offset::Utc, DateTime, NaiveDateTime};
 use hex_literal::hex;
 use primitive_types::{H160, H256, U256};
 use secp256k1::{constants::SECRET_KEY_SIZE, SecretKey};
 use serde::{de, Deserialize, Serialize};
 use serde::{Deserializer, Serializer};
-use std::fmt;
+use std::fmt::{self, Display};
 use web3::{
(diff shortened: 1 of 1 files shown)


josojo (approved):
Thank you!

-- just Felix, if you wonder --
I asked Valentin first, and then you answered and then Valentin coded this here.
Anyways. Thank you both a lot.

model/src/lib.rs
```diff
@@ -159,12 +160,23 @@ impl OrderCreation {
...
 pub struct OrderUid(pub [u8; 56]);
 
+impl Display for OrderUid {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let mut bytes = [0u8; 56 * 2];
+        // Unwrap because the length is always correct.
+        hex::encode_to_slice(&self.0, &mut bytes).unwrap();
+        // Unwrap because the string is always valid utf8.
+        let str = std::str::from_utf8(&bytes).unwrap();
+        f.write_str(str)
```
fleupold, line 170: Why not simply?
```suggestion
        f.write_str(&hex::encode(self.0.iter()))
```

fleupold (approved):

model/src/lib.rs
e00E: That version always allocates a string while encode_to_slice doesn't . Not a big deal but I've been trying to avoid this in several places.

PR #114, opened 2020-12-11
Specific order getter

josojo:
API end for getting an order with UID, as suggested in the API-defining PR: https://github.com/gnosis/gp-v2-services/pull/49/files

### Test Plan
unit tests


orderbook/src/api/handler.rs
```diff
@@ -83,6 +84,20 @@ pub async fn get_orders(orderbook: Arc<OrderBook>) -> Result<impl warp::Reply, I
...
 
+pub async fn get_order_by_uid(
+    uid: OrderUid,
+    orderbook: Arc<OrderBook>,
+) -> Result<impl warp::Reply, Infallible> {
+    if let Some(order) = orderbook.get_order_by_uid(uid).await {
+        Ok(with_status(json(&order), StatusCode::OK))
+    } else {
+        Ok(with_status(
+            json(&json!({ "description": "Order was not found"})),
```
fleupold, line 95: ubernit: I have no particular preference, but let's try to unify the response type when we are just returning a single error message (either as just a string as [here](https://github.com/gnosis/gp-v2-services/pull/111/files#diff-9ebbf7bf2a6a172779327578c2fc9dee475881fdbb9d4e326bf78c0bbaa7a522R86), or an object with one description field).

orderbook/src/orderbook.rs
```diff
@@ -48,6 +48,14 @@ impl OrderBook {
         self.orders.read().await.clone()
     }
 
+    pub async fn get_order_by_uid(&self, uid: OrderUid) -> Option<Order> {
+        let orders = self.get_orders().await;
+        orders
+            .into_iter()
+            .filter(|order| order.order_meta_data.uid == uid)
+            .last()
```
fleupold, line 56: Why last instead of first? I assume we want to guarantee that an orderUid can only be in the list once. Can we maybe enforce this at insertion time (and/or even using a more duplicate resistant datastructure like a HashSet or Map?) and then (debug) assert that there is only one match here.

orderbook/src/api/filter.rs
```diff
@@ -68,6 +78,41 @@ pub mod test_util {
...
+            .method("GET")
+            .reply(&filter)
+            .await;
+        assert_eq!(response.status(), StatusCode::OK);
+        let response_orders: Order = serde_json::from_slice(response.body()).unwrap();
+        let orderbook_orders = orderbook.get_orders().await;
+        assert_eq!(response_orders, orderbook_orders[0]);
+    }
+    #[tokio::test]
+    async fn get_order_by_uid_invalid_request() {
```
fleupold, line 101: nit: This test is not an invalid request though.

fleupold (approved):

orderbook/src/api/handler.rs
fleupold: Reading https://github.com/gnosis/gp-v2-services/issues/115 I guess this format here is better since consistent for the frontend.

orderbook/src/orderbook.rs
e00E: Yes now that we have the uid we can turn the vector into a hash map.

josojo: I rebased on the latest main commit with the code containing the new hashmap structure.

model/src/lib.rs
```diff
@@ -170,6 +171,18 @@ impl OrderCreation {
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub struct OrderUid(pub [u8; 56]);
 
+impl FromStr for OrderUid {
+    type Err = hex::FromHexError;
+    fn from_str(s: &str) -> Result<OrderUid, hex::FromHexError> {
+        let mut value = [0 as u8; 56];
+        let s_without_prefix = s
+            .strip_prefix("0x")
+            .ok_or(hex::FromHexError::InvalidStringLength)?;
```
josojo, line 180: Here, I would like to throw a custom error stating the error more clearly.

Which error type would you guys recommend? I think I want to use a very custom error type. Should I implement my own or is there a good one to use out of the box.

fleupold: Why not allow both? `s.strip_prefix("0x").unwrap_or(s)` given that fromHex seems to work with and without the prefix?

fleupold (approved):
Looks good to me

model/src/lib.rs
josojo: Yeah, this is definitively the most practical solution :) I love it.

It's just that we never did it that way and Valentin advocated for a well defined interface, and hence, we are doing:
```
let s = s.strip_prefix("0x").ok_or_else(|| {
                    de::Error::custom(format!(
                        "{:?} can't be decoded as hex uid because it does not start with '0x'",
                        s
                    ))
                 ```
                 In all the deserialize functions.

Comment (e00E) [reply model/src/lib.rs:180]: I'm okay with allowing both too although I slightly prefer being strict. I have taken the same shortcut for allowing both for the `H160Wrapper`:
```rust
impl FromStr for H160Wrapper {
    type Err = FromHexError;
    fn from_str(s: &str) -> Result<Self, Self::Err> {
        let s = s.strip_prefix("0x").unwrap_or(s);
        Ok(H160Wrapper(H160(FromHex::from_hex(s)?)))
    }
}
```
The correct way do not allow it would be to create a new error enum which is either MissingPrefix or FromHexError probably using https://github.com/dtolnay/thiserror .

orderbook/src/api/filter.rs
```diff
@@ -80,6 +88,41 @@ pub mod test_util {
...
+    #[tokio::test]
+    async fn get_order_by_uid_for_non_existent_order_() {
+        let orderbook = Arc::new(OrderBook::default());
+        let filter = get_order_by_uid(orderbook.clone());
+        let mut order_creation = OrderCreation::default();
+        order_creation.valid_to = u32::MAX;
+        order_creation.sign_self();
+        let order = orderbook.order_creation_to_order(order_creation).unwrap();
+        let response = request()
+            .path(&format!("/orders/{:}", order.order_meta_data.uid))
```
e00E, line 119: No need to create the full order here. You can create the uid directly like `let uid = OrderUid([u8; 56])` or add a derive for Default to OrderUid and use that. But if you prefer having the order here that's fine too.

orderbook/src/api/handler.rs
e00E: It's probably a good idea to make what is now "OrderPostError" just "ApiError" and return this as json for every error. That ensures that any error is always the same json format. Don't have to do it in this PR.

orderbook/src/api/filter.rs
```diff
@@ -80,6 +88,41 @@ pub mod test_util {
...
 
+    #[tokio::test]
+    async fn get_order_by_uid_() {
+        let orderbook = Arc::new(OrderBook::default());
+        let filter = get_order_by_uid(orderbook.clone());
+        let mut order_creation = OrderCreation::default();
+        order_creation.valid_to = u32::MAX;
+        order_creation.sign_self();
+        let order = orderbook.order_creation_to_order(order_creation).unwrap();
+        orderbook.add_order(order_creation).await.unwrap();
```
e00E, line 99: This function already returns the uid so there is no need to manually to `order_creation_to_order` here and that function can be made private again (this PR makes it `pub` so that you can use it here).

e00E (approved):

PR #116, opened 2020-12-11
Make all hex serialization use `0x` prefix

e00E:
This is what most people prefer. I do not want to accept both encodings
so this change requires the 0x always. Better to define the format more
exactly and not have amibguity.

Note to self: tag Anxo when this is merged

### Test Plan
adjusted unit tests


bh2smith (approved):

josojo (approved):
Solid!

fleupold:
Did you tag @anxolin or reach out to him some other way?

e00E:
We had a slack thread.

PR #118, opened 2020-12-11
Call settle in driver with fully encoded settlement

e00E:
instead of the placeholder trait we had before. Also changed some H160
and U256 to be passed by reference. Doesn't make a big difference it
just felt slightly more right to me as they are big types internally.

A small annoyance here is the `expect` on `encode_trades`. I don't think it's worth refactoring something here because we will have to adapt the settlement struct to the real solvers later anyway.

### Test Plan
uniswap interaction test has been adjusted.


solver/src/encoding.rs
```diff
@@ -41,23 +43,37 @@ fn encode_order_flags(order: &OrderCreation) -> u8 {
...
+    result
+}
+
+fn encode_interaction_data_length(length: usize) -> Option<[u8; 3]> {
+    let bytes = length.to_be_bytes();
+    let (left, right) = bytes.split_at(bytes.len() - 3);
+    if left.iter().any(|byte| *byte != 0) {
+        return None;
+    }
+    Some(right.try_into().expect("usize smaller than 3 bytes"))
```
fleupold, line 76: I didn't fully understand the expect message at first. It cannot fair because we split_at length - 3, right? Maybe saying 'right has exactly 3 bytes' ?

ubernit: Maybe just if length > 2**12 then instead of iterating (or adding a comment that this checks we don't overflow)

solver/src/encoding.rs
```diff
@@ -41,23 +43,37 @@ fn encode_order_flags(order: &OrderCreation) -> u8 {
...
+        *amount_out_min,
+        vec![*token_in, *token_out],
+        *payout_to,
         U256::MAX,
     );
     let data = method.tx.data.expect("call doesn't have calldata").0;
-    data.try_into().expect("unexpected data size")
+    let mut result = [0u8; UNISWAP_TOTAL_SIZE];
+    result[..20].copy_from_slice(target.as_fixed_bytes());
+    result[20..23].copy_from_slice(&encode_interaction_data_length(UNISWAP_DATA_SIZE).unwrap());
```
fleupold, line 65: Can we jsutify the unwrap in an expect?

solver/src/driver.rs
```diff
@@ -32,16 +34,25 @@ impl Driver {
...
+                    .expect("naive solver created invalid settlement"),
+                settlement.encode_interactions(
+                    &self.uniswap_router.address(),
+                    &self.settlement.address(),
+                ),
+                Vec::new(),
+            )
+        };
+        settle().call().await.context("settle call failed")?;
+        settle().send().await.context("settle send failed")?;
```
fleupold, line 55: nit can we in the context call it simulating settle vs executing sett (the difference between send and call might not be imediately obvious when reading this in a larger log context)

solver/src/driver.rs
```diff
@@ -32,16 +34,25 @@ impl Driver {
...
+            self.settlement.settle(
+                settlement.tokens(),
+                settlement.clearing_prices(),
+                settlement.fee_factor,
+                settlement
+                    .encode_trades()
+                    .expect("naive solver created invalid settlement"),
+                settlement.encode_interactions(
+                    &self.uniswap_router.address(),
+                    &self.settlement.address(),
```
fleupold, line 49: Instead of passing in all dependencies for all possible interactions into this function, I'd like to see each interaction to somehow know intrinsically (via a Factory or the like) how it can encode itself and what dependencies it needs. 
While this is fine for the demo with just one interaction, I think making sure we can easily add new interactions into the codebase is one of the first refactorings we should look at before accruing more debt.

solver/src/driver.rs
```diff
@@ -32,16 +34,25 @@ impl Driver {
...
-        self.contract
-            .settle_call(&settlement)
-            .await
-            .context("settle call failed")?;
-        self.contract
-            .settle_send(&settlement)
-            .await
-            .context("settle send failed")?;
+        let settle = || {
+            self.settlement.settle(
```
fleupold, line 40: nit: a lot of `settle` here. Maybe we can call it `self.settlement_contract` (eventually there could maybe also be another abstraction between the actual contract and the driver that e.g. handles the retry/gas pricing).

fleupold (approved):
Is this the last step we need for the e2e demo from our end or am I missing something?

josojo (approved):
really nice !

solver/src/encoding.rs
e00E: Added a comment to explain this more.

solver/src/encoding.rs
e00E: Added comment.

PR #120, opened 2020-12-11
Domain separator calculation

josojo:
Calculates the domain separator with the chainId and contract address.

### Test Plan
Run ganache,
Run contract deployment from readme in contracts crate
Run the orderbook and see that it is working.

### DIFF
--- a/model/Cargo.toml
+++ b/model/Cargo.toml
@@ -11,6 +11,9 @@ primitive-types = { version = "0.7" }
 secp256k1 = "0.17"
 serde = { version = "1.0", features = ["derive"] }
 web3 = { version = "0.13", default-features = false }
+lazy_static = "1.4"
+ethabi = "12.0"
+
 
 [dev-dependencies]
 serde_json = "1.0"
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -6,7 +6,10 @@ pub mod h160_hexadecimal;
 pub mod u256_decimal;
 
 use chrono::{offset::Utc, DateTime, NaiveDateTime};
+use ethabi::{encode, Token};
+use hex::{FromHex, FromHexError};
 use hex_literal::hex;
+use lazy_static::lazy_static;
 use primitive_types::{H160, H256, U256};
 use secp256k1::{constants::SECRET_KEY_SIZE, SecretKey};
 use serde::{de, Deserialize, Serialize};
@@ -64,7 +67,7 @@ impl OrderCreation {
     }
 
     // If signature is valid returns the owner.
-    pub fn validate_signature(&self, domain_separator: &[u8; 32]) -> Option<H160> {
+    pub fn validate_signature(&self, domain_separator: &DomainSeparator) -> Option<H160> {
         // The signature related functionality is defined by the smart contract:
         // https://github.com/gnosis/gp-v2-contracts/blob/main/src/contracts/libraries/GPv2Encoding.sol
 
@@ -86,9 +89,9 @@ impl OrderCreation {
 
 // Intended to be used by tests that need signed orders.
 impl OrderCreation {
-    pub const TEST_DOMAIN_SEPARATOR: [u8; 32] = [0u8; 32];
+    pub const TEST_DOMAIN_SEPARATOR: DomainSeparator = DomainSeparator([0u8; 32]);
 
-    pub fn sign_self_with(&mut self, domain_separator: &[u8; 32], key: SecretKeyRef) {
+    pub fn sign_self_with(&mut self, domain_separator: &DomainSeparator, key: SecretKeyRef) {
         let message = self.signing_digest_message(domain_separator);
         // Unwrap because the only error is for invalid messages which we don't create.
         let signature = Key::sign(&key, &message, None).unwrap();
@@ -131,23 +134,23 @@ impl OrderCreation {
         signing::keccak256(&hash_data)
     }
 
-    fn signing_digest_typed_data(&self, domain_separator: &[u8; 32]) -> [u8; 32] {
+    fn signing_digest_typed_data(&self, domain_separator: &DomainSeparator) -> [u8; 32] {
         let mut hash_data = [0u8; 66];
         hash_data[0..2].copy_from_slice(&[0x19, 0x01]);
-        hash_data[2..34].copy_from_slice(domain_separator);
+        hash_data[2..34].copy_from_slice(&domain_separator.0);
         hash_data[34..66].copy_from_slice(&self.order_digest());
         signing::keccak256(&hash_data)
     }
 
-    fn signing_digest_message(&self, domain_separator: &[u8; 32]) -> [u8; 32] {
+    fn signing_digest_message(&self, domain_separator: &DomainSeparator) -> [u8; 32] {
         let mut hash_data = [0u8; 92];
         hash_data[0..28].copy_from_slice(b"\x19Ethereum Signed Message:\n64");
-        hash_data[28..60].copy_from_slice(domain_separator);
+        hash_data[28..60].copy_from_slice(&domain_separator.0);
         hash_data[60..92].copy_from_slice(&self.order_digest());
         signing::keccak256(&hash_data)
     }
 
-    fn signing_digest(&self, domain_separator: &[u8; 32]) -> [u8; 32] {
+    fn signing_digest(&self, domain_separator: &DomainSeparator) -> [u8; 32] {
         if self.signature.v & 0x80 == 0 {
             self.signing_digest_typed_data(domain_separator)
         } else {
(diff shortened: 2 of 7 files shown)


orderbook/src/main.rs
```diff
@@ -2,11 +2,18 @@ mod api;
...
+use ethcontract::{Http, Web3};
+use model::DomainSeparator;
+use primitive_types::H160;
+use std::fs;
 use std::{net::SocketAddr, sync::Arc, time::Duration};
 use structopt::StructOpt;
 use tokio::task;
 use warp::Filter;
 
+#[path = "../../contracts/src/paths.rs"]
```
josojo, line 14: I did not know how to do it correctly. 

Should we share this code in the lib of contracts?

Gernerally, I am not sure whether we really wanna read the contract addresses from here. I will wait for Valentins implementation to see "a proper solution"

orderbook/src/main.rs
```diff
@@ -27,10 +34,24 @@ pub async fn orderbook_maintenance(orderbook: Arc<OrderBook>) -> ! {
 }
 #[tokio::main]
 async fn main() {
+    // Todo: merge valentins PR to get it from env
+    const NODE_URL: &str = "http://localhost:8545";
+
+    let http = Http::new(NODE_URL).expect("Node url invalid, or service not available");
+    let web3 = Web3::new(http);
+
+    let chain_id = web3.eth().chain_id().await.expect("Could not get chainId");
```
josojo, line 43: Probably, we want to run here:
```
wait_for_node(&web3).await?;
```

But I was not able to export from the other crate in a proper way.

model/src/lib.rs
```diff
@@ -6,7 +6,10 @@ pub mod h160_hexadecimal;
 pub mod u256_decimal;
 
 use chrono::{offset::Utc, DateTime, NaiveDateTime};
+use ethcontract::common::abi::{encode, Token};
```
fleupold, line 9: Does model really have to depend on ethcontracts? Are encode and Token not reexported from some more basic type crate?

orderbook/src/main.rs
fleupold: You should be able to get the address from calling `GPv2Settlement::deployed().await`

orderbook/src/main.rs
fleupold: I think it's ok to assume the RPC endpoint is running when we start the API.

model/src/lib.rs
```diff
@@ -329,13 +332,76 @@ impl TokenPair {
...
+            Token::Uint((*DOMAIN_TYPE_HASH).into()),
+            Token::Uint((*DOMAIN_NAME).into()),
+            Token::Uint((*DOMAIN_VERSION).into()),
+            Token::Uint(chain_id.into()),
+            Token::Address(contract_address),
+        ]);
+
+        DomainSeparator(signing::keccak256(abi_encode_string.as_slice()))
+    }
+}
```
fleupold, line 381: This module is getting quite large. How about splitting it up into multiple smaller ones (in a separate PR)?

model/src/lib.rs
```diff
@@ -329,13 +332,76 @@ impl TokenPair {
     }
 }
 
+#[derive(Copy, Eq, PartialEq, Clone, Default)]
+pub struct DomainSeparator(pub [u8; 32]);
+
+// For command line argument parsing.
```
fleupold, line 338: Where do we plant o use this?

model/src/lib.rs
```diff
@@ -86,9 +89,9 @@ impl OrderCreation {
 
 // Intended to be used by tests that need signed orders.
 impl OrderCreation {
-    pub const TEST_DOMAIN_SEPARATOR: [u8; 32] = [0u8; 32];
+    pub const TEST_DOMAIN_SEPARATOR: DomainSeparator = DomainSeparator([0u8; 32]);
```
fleupold, line 92: nit: isn't this just DomainSeparator::Default?

fleupold (approved):
Nice, I can see how implementing this manual is quite a lot of work. I'm wondering what was your argument against not just reading it from the smart contract?

model/src/lib.rs
josojo: ups, yeah I will remove the comment. But the fromstr trait is very handy. It is for example used here:
```
let expected_domain_separator: DomainSeparator =
            "9d7e07ef92761aa9453ae5ff25083a2b19764131b15295d3c7e89f1f1b8c67d9"
                .parse()
                .unwrap();
```
in the tests.

model/src/lib.rs
josojo: The compiler yells at me::
```
calls in constants are limited to constant functions, tuple structs and tuple variants
```
So I have to use lazy_static, and I guess the current solution is shorter and easier to read.

model/src/lib.rs
josojo: https://github.com/gnosis/gp-v2-services/issues/121

josojo:
Thanks for the grad feedback.

No, there is no particular reason for calculating it beyond the ones nick mentioned.  I just wanted to get it done, as I spend already so much fookin time on it.

model/src/lib.rs
```diff
@@ -329,13 +332,75 @@ impl TokenPair {
...
+
 #[cfg(test)]
 mod tests {
     use super::*;
     use chrono::NaiveDateTime;
     use serde_json::json;
     use std::str::FromStr;
 
+    #[test]
+
```
e00E, line 390:
```suggestion
```

model/src/lib.rs
```diff
@@ -329,13 +332,75 @@ impl TokenPair {
...
+    fn domain_separator_rinkeby() {
+        let contract_address: H160 = "91D6387ffbB74621625F39200d91a50386C9Ab15".parse().unwrap();
```
e00E, line 392: With the hex macro it's slightly cleaner:
```suggestion
    let contract_address: H160 = hex!("91D6387ffbB74621625F39200d91a50386C9Ab15").into();
```

model/src/lib.rs
```diff
@@ -329,13 +332,75 @@ impl TokenPair {
...
+        let chain_id: u64 = 4;
+        let domain_separator_rinkeby =
+            DomainSeparator::get_domain_separator(chain_id, contract_address);
+        // domain separator is taken from rinkeby deployment at address 91D6387ffbB74621625F39200d91a50386C9Ab15
+        let expected_domain_separator: DomainSeparator =
+            "9d7e07ef92761aa9453ae5ff25083a2b19764131b15295d3c7e89f1f1b8c67d9"
+                .parse()
+                .unwrap();
```
e00E, line 400: Here too. And then I think we don't need the DomainSeparator FromStr implementation anymore.
```suggestion
        let expected_domain_separator: DomainSeparator = DomainSeparator(hex!(
            "9d7e07ef92761aa9453ae5ff25083a2b19764131b15295d3c7e89f1f1b8c67d9"
        ));
```

orderbook/Cargo.toml
```diff
@@ -16,3 +17,5 @@ tokio = { version = "0.2", features =[ "macros", "time"] }
 tracing = "0.1"
 tracing-setup = { path = "../tracing-setup" }
 warp = "0.2"
+ethcontract = "0.9.1"
```
e00E, line 20:
```suggestion
ethcontract = { version = "0.9", default-features = false, features = ["http-tls"] }
```

model/Cargo.toml
```diff
@@ -11,6 +11,9 @@ primitive-types = { version = "0.7" }
 secp256k1 = "0.17"
 serde = { version = "1.0", features = ["derive"] }
 web3 = { version = "0.13", default-features = false }
+lazy_static = "1.4.0"
+ethabi = "12.0.0"
```
e00E, line 15: Keeping the version number consistent with the rest of the file
```suggestion
lazy_static = "1.4"
ethabi = "12.0"
```

e00E (approved):

.travis.yml
```diff
@@ -35,6 +35,7 @@ jobs:
         - $SCCACHE --version
       script:
         - cargo fmt -- --check
+        - cargo build
```
e00E, line 38: Don't think this is needed. Build failure is because of the thing I mentioned in the call that ethcontract doesn't see the address. But Felix said he would push a fix for that.

orderbook/Cargo.toml
```diff
@@ -17,3 +18,5 @@ tokio = { version = "0.2", features =[ "macros", "time"] }
 tracing = "0.1"
 tracing-setup = { path = "../tracing-setup" }
 warp = "0.2"
+ethcontract = { version = "0.9", default-features = false, features = ["http-tls"] }
```
e00E, line 21:
```suggestion
web3 = { version = "0.13", default-features = false, features = ["http-tls"] }
```

orderbook/src/main.rs
```diff
@@ -30,7 +33,18 @@ async fn main() {
...
     tracing_setup::initialize(args.shared.log_filter.as_str());
     tracing::info!("running order book with {:#?}", args);
-    let orderbook = Arc::new(OrderBook::default());
+
+    let http = Http::new(args.shared.node_url.as_str()).expect("Couldn't connect to HTTP");
+    let web3 = Web3::new(http);
+
+    let settlement_contract = GPv2Settlement::deployed(&web3)
+        .await
+        .expect("Couldn't load deployed settlement");
```
e00E, line 42:
```suggestion
    let transport = web3::transports::Http::new(args.shared.node_url.as_str())
        .expect("transport creation failed");
    let web3 = web3::Web3::new(transport);
    let settlement_contract = contracts::GPv2Settlement::deployed(&web3)
```

e00E (approved):
One last suggestion to make the web3 creation the same in orderbook/main.rs and solver/main.rs

PR #123, opened 2020-12-14
Set up web3 with node url argument

e00E:
We need a Web3 instance to make calls to the contract.
Next step is to set up the contracts instances with this web3.

### Test Plan
`cargo run --bin solver -- --node-url "http://localhost:8080"` to see that web3 is created. Doesn't actually need to have a node running at that address because no call is made.

### DIFF
--- a/shared-arguments/src/lib.rs
+++ b/shared-arguments/src/lib.rs
@@ -1,6 +1,7 @@
 //! Contains command line arguments and related helpers that are shared between the binaries.
 
 use std::{num::ParseFloatError, time::Duration};
+use url::Url;
 
 #[derive(Debug, structopt::StructOpt)]
 pub struct Arguments {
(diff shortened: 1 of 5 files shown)


shared-arguments/src/lib.rs
```diff
@@ -10,6 +11,19 @@ pub struct Arguments {
         default_value = "warn,orderbook=debug,solver=debug"
     )]
     pub log_filter: String,
+
+    /// The Ethereum node URL to connect to.
+    #[structopt(long, env = "NODE_URL")]
```
josojo, line 16: The local host as default would be nice for testing

josojo (approved):
nice, was just about to prepare the same

shared-arguments/src/lib.rs
e00E: Changed default to `http://localhost:8545`.

fleupold: Not a big deal but personally I find the node_url shouldn't have a default as it's the one thing you should deliberately think about and specify when running this thing (what network am I on).

PR #125, opened 2020-12-14
Allow 0x prefix in fee info route

e00E:
I forgot about this when I changed the other routes to use the prefix.

### Test Plan
adapted unit test


josojo (approved):
Sweet. thanks a lot for jumping in!

PR #126, opened 2020-12-14
[Trivial], changing api path for fees

josojo:
according to anxo's feedback

### DIFF
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -98,7 +98,7 @@ paths:
           description: Order deleted
         404:
           description: Order was not found
-  /api/v1/token/{sellToken}/fee:
+  /api/v1/tokens/{sellToken}/fee:
     get:
       description: |
         The fee that is charged for placing an order.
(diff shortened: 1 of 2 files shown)


orderbook/openapi.yml
```diff
@@ -98,7 +98,7 @@ paths:
           description: Order deleted
         404:
           description: Order was not found
-  /api/v1/fee/{sellToken}:
+  /api/v1/token/{sellToken}/fee:
```
e00E, line 101:
```suggestion
  /api/v1/tokens/{sellToken}/fee:
```

e00E (approved):

anxolin:
Thanks for this :) 

> according to anxo's feedback

Just for the record, I don't invent this rules 🙃, you can find a lot about REST API design, is not just I decided was not a good name.  

For example, in this link they have a similar example to yours:
https://scotch.io/bar-talk/designing-a-restful-web-api

just change `stores` for `tokens`, and `report` for `fee`

image

PR #128, opened 2020-12-14
Make interactions a trait instead of an enum

e00E:
This is much more flexible for the future because it allows interactions
to locally encapsulate the data they need instead of having everything
in one big enum.

The meat of this change is that now  there is a trait that Interactions implement to encode themself:
```rust
    fn encode(&self, writer: &mut dyn Write) -> Result<()>;
```
Following from that the uniswap interaction has been moved into it's own module away from encoding.rs.

### Test Plan
the uniswap interaction unit test has been adapted

### DIFF
--- a/solver/src/encoding.rs
+++ b/solver/src/encoding.rs
@@ -1,13 +1,10 @@
 use std::convert::TryInto;
 
-use contracts::UniswapV2Router02;
 use model::{OrderCreation, OrderKind};
-use primitive_types::{H160, U256};
+use primitive_types::U256;
 
 pub const TRADE_STRIDE: usize = 206;
-const INTERACTION_BASE_SIZE: usize = 20 + 3; // target address + data length
-const UNISWAP_DATA_SIZE: usize = 260;
-const UNISWAP_TOTAL_SIZE: usize = INTERACTION_BASE_SIZE + UNISWAP_DATA_SIZE;
+pub const INTERACTION_BASE_SIZE: usize = 20 + 3; // target address + data length
 
 /// Creates the data which the smart contract's `decodeTrade` expects.
 pub fn encode_trade(
@@ -45,33 +42,8 @@ fn encode_order_flags(order: &OrderCreation) -> u8 {
     result
 }
 
-pub fn encode_uniswap_call(
-    target: &H160,
-    amount_in: &U256,
-    amount_out_min: &U256,
-    token_in: &H160,
-    token_out: &H160,
-    payout_to: &H160,
-) -> [u8; UNISWAP_TOTAL_SIZE] {
-    let uniswap = UniswapV2Router02::at(&dummy::dummy_web3(), H160::zero());
-    let method = uniswap.swap_exact_tokens_for_tokens(
-        *amount_in,
-        *amount_out_min,
-        vec![*token_in, *token_out],
-        *payout_to,
-        U256::MAX,
-    );
-    let data = method.tx.data.expect("call doesn't have calldata").0;
-    let mut result = [0u8; UNISWAP_TOTAL_SIZE];
-    result[..20].copy_from_slice(target.as_fixed_bytes());
-    // Unwrap because we know uniswap data size can be stored in 3 bytes.
-    result[20..23].copy_from_slice(&encode_interaction_data_length(UNISWAP_DATA_SIZE).unwrap());
-    result[23..].copy_from_slice(data.as_slice());
-    result
-}
-
 // None if length doesn't fit in 3 bytes.
-fn encode_interaction_data_length(length: usize) -> Option<[u8; 3]> {
+pub fn encode_interaction_data_length(length: usize) -> Option<[u8; 3]> {
     let bytes = length.to_be_bytes();
     let (left, right) = bytes.split_at(bytes.len() - 3);
     if left.iter().any(|byte| *byte != 0) {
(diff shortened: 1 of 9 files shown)


solver/src/encoding.rs
```diff
@@ -43,33 +40,8 @@ fn encode_order_flags(order: &OrderCreation) -> u8 {
     result
 }
 
-pub fn encode_uniswap_call(
-    target: &H160,
```
e00E, line 47: moved to `interacations/uniswap.rs`

solver/src/encoding.rs
```diff
@@ -80,37 +52,13 @@ fn encode_interaction_data_length(length: usize) -> Option<[u8; 3]> {
     Some(right.try_into().unwrap())
 }
 
-// To create an ethcontract instance we need to provide a web3 even though we never use it. This
-// module provides a dummy transport and web3.
-mod dummy {
```
e00E, line 85: moved to `interactions.rs`

solver/src/encoding.rs
```diff
@@ -148,38 +96,4 @@ mod tests {
         assert_eq!(encoded[140..172], u8_as_32_bytes_be(10));
         assert_eq!(encoded[172..204], u8_as_32_bytes_be(11));
     }
-
-    #[test]
-    fn encode_uniswap_call_() {
```
e00E, line 153: moved to `interactions/uniswap.rs`

solver/src/naive_solver/two_order_settlement.rs
```diff
@@ -1,14 +1,55 @@
...
-use primitive_types::U256;
+use primitive_types::{H160, U256};
+use std::collections::HashMap;
+
+#[derive(Debug)]
+pub struct TwoOrderSettlement {
+    clearing_prices: HashMap<H160, U256>,
+    trades: [Trade; 2],
+    interaction: Option<AmmSwapExactTokensForTokens>,
+}
```
e00E, line 14: A new type that only the naive solver uses. This struct is easier to work with locally than the full `Settlement` but can be converted into that.

solver/src/settlement.rs
```diff
@@ -1,31 +1,32 @@
...
-        token_in: H160,
-        token_out: H160,
-    },
+pub trait Interaction: std::fmt::Debug {
+    // TODO: not sure if this should return a result.
+    // Write::write returns a result but we know we write to a vector in memory so we know it will
+    // never fail. Then the question becomes whether interactions should be allowed to fail encoding
+    // for other reasons.
+    fn encode(&self, writer: &mut dyn Write) -> Result<()>;
 }
```
e00E, line 22: We can discuss this or leave as is for now.

solver/src/interactions/uniswap.rs
```diff
@@ -0,0 +1,83 @@
...
+            self.amount_in,
+            self.amount_out_min,
+            vec![self.token_in, self.token_out],
+            self.payout_to,
+            U256::MAX,
+        );
+        let data = method.tx.data.expect("no calldata").0;
+        writer.write_all(self.uniswap_router_address.as_fixed_bytes())?;
+        // Unwrap because we know uniswap data size can be stored in 3 bytes.
+        writer.write_all(&encoding::encode_interaction_data_length(UNISWAP_DATA_SIZE).unwrap())?;
```
fleupold, line 35: Should we use `data.len()` instead of `UNISWAP_DATA_SIZE`? We could also assert they are equal but I don't really think there is value in knowing the exact length of every interaction (other than for the purpose of testing).

solver/src/interactions/uniswap.rs
```diff
@@ -0,0 +1,83 @@
...
+use anyhow::Result;
+use contracts::UniswapV2Router02;
+use primitive_types::{H160, U256};
+
+const UNISWAP_DATA_SIZE: usize = 260;
+const UNISWAP_TOTAL_SIZE: usize = INTERACTION_BASE_SIZE + UNISWAP_DATA_SIZE;
+
+#[derive(Debug)]
+pub struct UniswapInteraction {
+    pub uniswap_router_address: H160,
```
fleupold, line 14: How about using the actual instance here? Then we wouldn't need any dummy_web3. I think in the mid-term we also want to use the actual instance to check e.g. if we have already approved the token or we need to that as part of the interaction list.

solver/src/settlement.rs
```diff
@@ -58,26 +59,12 @@ impl Settlement {
         Some(bytes)
     }
 
-    pub fn encode_interactions(&self, uniswap_router_address: &H160, payout_to: &H160) -> Vec<u8> {
-        let mut bytes = Vec::new();
+    pub fn encode_interactions(&self) -> Result<Vec<u8>> {
+        let mut cursor = Cursor::new(Vec::new());
```
fleupold, line 63: Very neat!

solver/src/interactions/uniswap.rs
```diff
@@ -0,0 +1,83 @@
+use crate::{
+    encoding::{self, INTERACTION_BASE_SIZE},
+    settlement::Interaction,
+};
+use anyhow::Result;
+use contracts::UniswapV2Router02;
+use primitive_types::{H160, U256};
+
+const UNISWAP_DATA_SIZE: usize = 260;
+const UNISWAP_TOTAL_SIZE: usize = INTERACTION_BASE_SIZE + UNISWAP_DATA_SIZE;
```
fleupold, line 10: Is this still used?

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2Fit6W8D4FfvaPC%2Fgiphy.gif&token=beautfiul" data-gitmeme-token="beautfiul"></a>

solver/src/interactions/uniswap.rs
e00E: switched to data.len()

solver/src/interactions/uniswap.rs
e00E: removed

solver/src/interactions/uniswap.rs
e00E: made this change

josojo (approved):
This one is indeed beautiful. 

 Felix found already all points for improvements :) sparkling clean

PR #129, opened 2020-12-14
Upgrade settlement contract

e00E:
So that we have the newest address for order signature verification.

### Test Plan
CI


josojo (approved):
Thanks

PR #133, opened 2020-12-15
Change encoding order of signature fields

e00E:
The smart contract internally uses v + r + s but the frontend uses r + s + v which seems overall more common. So change our api to use that same
ordering.

### Test Plan
adjusted unit test


fleupold (approved):
nanonit: maybe also reorder the fields on the struct (not that it matters, it's just slightly more coherent)?

PR #134, opened 2020-12-15
Update order signature and encoding to smart contract

e00E:
For the encoding of trades there is a new field `fee_discount` and for the signature the way the order type (buy or sell) is computed has changed.

### Test Plan
adjusted unit tests

### DIFF
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -113,7 +113,14 @@ impl OrderCreation {
 // See https://github.com/gnosis/gp-v2-contracts/blob/main/src/contracts/libraries/GPv2Encoding.sol
 impl OrderCreation {
     const ORDER_TYPE_HASH: [u8; 32] =
-        hex!("b71968fcf5e55b9c3370f2809d4078a4695be79dfa43e5aa1f2baa0a9b84f186");
+        hex!("b2b38b9dcbdeb41f7ad71dea9aed79fb47f7bbc3436576fe994b43d5b16ecdec");
+
+    // keccak256("sell")
+    const ORDER_KIND_SELL: [u8; 32] =
+        hex!("f3b277728b3fee749481eb3e0b3b48980dbbab78658fc419025cb16eee346775");
+    // keccak256("buy")
+    const ORDER_KIND_BUY: [u8; 32] =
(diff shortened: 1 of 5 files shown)


model/src/lib.rs
```diff
@@ -113,7 +113,14 @@ impl OrderCreation {
 // See https://github.com/gnosis/gp-v2-contracts/blob/main/src/contracts/libraries/GPv2Encoding.sol
 impl OrderCreation {
     const ORDER_TYPE_HASH: [u8; 32] =
-        hex!("b71968fcf5e55b9c3370f2809d4078a4695be79dfa43e5aa1f2baa0a9b84f186");
+        hex!("b2b38b9dcbdeb41f7ad71dea9aed79fb47f7bbc3436576fe994b43d5b16ecdec");
+
+    // keccak256("sell")
+    const ORDER_KIND_SELL: [u8; 32] =
+        hex!("f3b277728b3fee749481eb3e0b3b48980dbbab78658fc419025cb16eee346775");
```
e00E, line 120: As for the order type hash, I am open to computing this in a more high level way by actually doing the keccak but for now I  am keeping the exact hex string as it is in the contract to make it obviously correct.

josojo (approved):
Hash updates are correct. I double checked them.

I would also just copy-paste the ORDER_KIND_SELL and ORDER_KIND_BUY, as this is more obvious on whether an error was made.

model/src/lib.rs
```diff
@@ -113,7 +113,14 @@ impl OrderCreation {
...
+    // keccak256("buy")
+    const ORDER_KIND_BUY: [u8; 32] =
+        hex!("6ed88e868af0a1983e3886d5f3e95a2fafbd6c3450bc229e27342283dc429ccc");
```
fleupold, line 123: As a followup, should we add a check on startup to make sure the smart contract agrees about that value?

fleupold (approved):
Looks good! Once that is merge I can send the e2e test our for review (it's working modulo the Uniswap approval)

model/src/lib.rs
e00E: That is probably a good sanity check, yes.

PR #135, opened 2020-12-15
Set up CORS in orderbook api

e00E:
It is cleaner to use warp's cors module than to manually set the header.
We also need to allow methods and headers specifically for the frontend.

I don't fully understand CORS so it would be good if this could be reviewed by a frontend developer.

### Test Plan
Tested with openapi that I can make a post request.

### DIFF
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -41,10 +41,12 @@ async fn main() {
     let chain_id = web3.eth().chain_id().await.expect("Could not get chainId");
     let domain_separator =
         DomainSeparator::get_domain_separator(chain_id.as_u64(), settlement_contract.address());
-
     let orderbook = Arc::new(OrderBook::new(domain_separator));
-    let filter = api::handle_all_routes(orderbook.clone())
-        .map(|reply| warp::reply::with_header(reply, "Access-Control-Allow-Origin", "*"));
+    let cors = warp::cors()
+        .allow_any_origin()
+        .allow_methods(vec!["GET", "POST", "DELETE", "OPTIONS", "PUT", "PATCH"])
+        .allow_headers(vec!["Origin", "Content-Type", "X-Auth-Token", "X-AppId"]);
+    let filter = api::handle_all_routes(orderbook.clone()).with(cors.build());
     tracing::info!("serving order book");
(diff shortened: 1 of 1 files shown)


fleupold (approved):
LGTM but maybe @anxolin can sign off as well. To my understanding the current approach only adds the CORS header when we have a successful reply (which might not be the case for preflight OPTION requests). I'm not sure, why we need to add the other headers though.

orderbook/src/main.rs
```diff
@@ -41,10 +41,12 @@ async fn main() {
     let chain_id = web3.eth().chain_id().await.expect("Could not get chainId");
     let domain_separator =
         DomainSeparator::get_domain_separator(chain_id.as_u64(), settlement_contract.address());
-
     let orderbook = Arc::new(OrderBook::new(domain_separator));
-    let filter = api::handle_all_routes(orderbook.clone())
-        .map(|reply| warp::reply::with_header(reply, "Access-Control-Allow-Origin", "*"));
+    let cors = warp::cors()
+        .allow_any_origin()
+        .allow_methods(vec!["GET", "POST", "DELETE", "OPTIONS"])
```
anxolin, line 47: since u are adding the list, I would just in case add PUT and PATCH, since we might allow in the future endpoints using those

orderbook/src/main.rs
```diff
@@ -41,10 +41,12 @@ async fn main() {
...
+        .allow_headers(vec!["Origin", "Content-Type", "X-Auth-Token", "X-AppId"]);
```
anxolin, line 48: Do we want need to be this explicit with the header? 

I'm not expert either, is it more secure than `Access-Control-Allow-Headers: *` ?

anxolin (approved):
Looks good to me, let me know when u need me to test this in the web with the X-AppId header

orderbook/src/main.rs
fleupold: It doesn't seem to have allow_any_headers, but maybe we can try allow_header("*")

e00E: Looks like warp doesn't support this currently. Maybe something we could implement for warp.

PR #137, opened 2020-12-15
Store orders in a hash map based on uid in orderbook

e00E:
This will be needed for the not yet implemented route to get an order based
on uid.

### Test Plan
existing tests

### DIFF
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -160,7 +160,7 @@ impl OrderCreation {
 }
 
 // uid as 56 bytes: 32 for orderDigest, 20 for ownerAddress and 4 for validTo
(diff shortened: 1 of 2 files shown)


josojo (approved):
The route for getting an order by orderId is already implemented as one of my PRs. But I was just filtering all orders. 

Having this hashmap data-structure helps for sure, also for the maintenance of "kicking out" settled orders.

Though for later, I think the most efficient structure would be:
`Hashmap< tokenPair, Hashmap< uid, Order>>>`

Then, we can easily get the orders per pair and then them over to the solvers.

fleupold (approved):
Nice, I think for more advance data structuring (to allow quick filtering by token pair, etc) we should move to our DB storage layout as this in memory thing is just meant for the first demo.

PR #138, opened 2020-12-15
E2e test

fleupold:
This PR adds an e2e test for the services component. At the moment we are placing two perfectly matching EOA orders, which get matched on the test network deployed contracts. It should also work with Uniswap as soon as we approve the settlement contract for transfer (atm it's failing due to insufficient approval). I can implement this but am waiting for #128 to merge.

I created libs for the two bin crates (orderbook, solver) so that their main component can be loaded directly in the process of the unit test. This was already very helpful for debugging when trying to pin point some of the encoding issues (being able to breakpoint driver and orderbook in the same process directly in the editor was nice). The test is also running very fast since we control the runloop.

### Test Plan
Start ganache
```
(cd contracts; cargo run --bin deploy --features bin)
cargo test
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 members = [
     "contracts",
+    "e2e",
     "model",
     "orderbook",
     "shared-arguments",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,3 +8,12 @@ members = [
     "solver",
     "tracing-setup",
 ]
+
+default-members = [
+    "contracts",
+    "model",
+    "orderbook",
+    "shared-arguments",
+    "solver",
+    "tracing-setup",
+]
--- a/e2e/tests/uniswap_trade_test.rs
+++ b/e2e/tests/uniswap_trade_test.rs
@@ -1,5 +1,6 @@
 use contracts::{ERC20Mintable, GPv2Settlement, UniswapV2Factory, UniswapV2Router02};
 use ethcontract::prelude::{Account, Address, Http, PrivateKey, Web3, U256};
+use hex_literal::hex;
 use model::{DomainSeparator, OrderCreationBuilder, OrderKind};
 use orderbook::orderbook::OrderBook;
 use secp256k1::SecretKey;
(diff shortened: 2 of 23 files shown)


e2e/tests/uniswap_trade_test.rs
```diff
@@ -0,0 +1,186 @@
...
+const TRADER_B_PK: [u8; 32] = [
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2,
+];
+
+const NODE_HOST: &str = "http://localhost:8545";
+const API_HOST: &str = "http://localhost:8080";
+const ORDER_PLACEMENT_ENDPOINT: &str = "/api/v1/orders/";
+
+#[tokio::test]
+async fn test_with_ganache() {
```
e00E, line 22: Because the test requires ganache I prefer if it didn't ran on `cargo test` in the crate root. Could achieve this by using  default-members in the workspace Cargo.toml or we could make the test a real binary target instead of a test. Or we could create an "integration test" as https://doc.rust-lang.org/stable/book/ch11-03-test-organization.html#integration-tests described. I have never tried. I also don't mind having whatever is the simplest solution first and doing more work in a later PR.

Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [workspace]
 members = [
     "contracts",
+    "e2e",
```
e00E, line 4: I guess we will set this test up in travis too. It's fine if that's another PR.

e2e/tests/uniswap_trade_test.rs
```diff
@@ -0,0 +1,186 @@
...
+use serde_json::json;
+use std::{str::FromStr, sync::Arc};
+use web3::signing::SecretKeyRef;
+
+const TRADER_A_PK: [u8; 32] = [
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1,
+];
+const TRADER_B_PK: [u8; 32] = [
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 2,
+];
```
e00E, line 15: If you prefer you can use the `hex!` macro here that we have in other parts of the code.

e2e/tests/uniswap_trade_test.rs
e00E, line 22: We should set up tracing in this test to get log messages. Same way as in the binaries with `tracing_setup::initialize`.

e00E (approved):
I like how making the binary crates also have a library and then using that library in the e2e test works out.

e2e/tests/uniswap_trade_test.rs
fleupold: Good point. I made everything but e2e a default-member. This achieves that it isn't run on `cargo test`. On travis we run `cargo test --workspace` so it's covered.

Cargo.toml
fleupold: I added it to this PR

model/src/lib.rs
```diff
@@ -61,6 +61,68 @@ pub struct OrderCreation {
     pub signature: Signature,
 }
 
+#[derive(Default)]
+pub struct OrderCreationBuilder {
+    order_creation: OrderCreation,
+}
+
+#[allow(dead_code)]
+impl OrderCreationBuilder {
```
josojo, line 70: This makes it really nice to read. Really like it

josojo (approved):
Really neat!

I love the coexistence of main and lib. Sooo useful for this cool test-steering

PR #141, opened 2020-12-16
Approve in Uniswap Interaction

fleupold:
This PR adds functionality to encode the approval tx of the sell token to Uniswap (which is required for the e2e test). In order to keep encoding synchronous, this is atm just passed in from above (in the future we should probably keep track of which approvals we have already given and only add them for newly traded tokens).

### Test Plan
Fully working e2e test (in two PRs). For now CI

### DIFF
--- a/solver/Cargo.toml
+++ b/solver/Cargo.toml
@@ -14,7 +14,7 @@ path = "src/main.rs"
 [dependencies]
 anyhow = "1.0"
 contracts = { path = "../contracts" }
-ethcontract = "0.9"
+ethcontract = { version = "0.9", default-features = false }
 futures = "0.3"
 hex-literal = "0.3"
 jsonrpc-core = "14.0"
--- a/solver/Cargo.toml
+++ b/solver/Cargo.toml
@@ -13,9 +13,10 @@ path = "src/main.rs"
 
 [dependencies]
 anyhow = "1.0"
-async-trait = "0.1"
 contracts = { path = "../contracts" }
+ethcontract = "0.9"
 futures = "0.3"
+hex-literal = "0.3"
 jsonrpc-core = "14.0"
 maplit = "1.0"
 model = { path = "../model" }
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -2,6 +2,7 @@ use crate::{naive_solver, orderbook::OrderBookApi};
 use anyhow::{anyhow, Context, Result};
 use contracts::{GPv2Settlement, UniswapV2Router02};
 use std::time::Duration;
+use tracing::info;
 
 const SETTLE_INTERVAL: Duration = Duration::from_secs(30);
 
@@ -49,12 +50,12 @@ impl Driver {
         let settlement = match naive_solver::settle(
             orders.into_iter().map(|order| order.order_creation),
             &self.uniswap_contract,
-            &self.settlement_contract.address(),
+            &self.settlement_contract,
         ) {
             None => return Ok(()),
             Some(settlement) => settlement,
         };
-        tracing::debug!("got settlement");
+        info!("Computed {:?}", settlement);
         // TODO: check if we need to approve spending to uniswap
         // TODO: use retry transaction sending crate for updating gas prices
         let encoded_interactions = settlement
@@ -74,7 +75,7 @@ impl Driver {
                 )
                 .gas(8_000_000u32.into())
         };
-        settle().call().await.context("settle simulation failed")?;
+        //settle().call().await.context("settle simulation failed")?;
         settle().send().await.context("settle execution failed")?;
         Ok(())
     }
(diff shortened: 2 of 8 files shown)


solver/Cargo.toml
```diff
@@ -13,9 +13,10 @@ path = "src/main.rs"
 
 [dependencies]
 anyhow = "1.0"
-async-trait = "0.1"
 contracts = { path = "../contracts" }
+ethcontract = "0.9"
```
e00E, line 17: Remove default features. We already get the http transport from web3 directly. This is why there are so many changes in Cargo.lock, because there are a lot of new transitive dependencies added.
```suggestion
ethcontract = { version = "0.9", default-feature = false }
```

solver/src/driver.rs
```diff
@@ -74,7 +75,7 @@ impl Driver {
                 )
                 .gas(8_000_000u32.into())
         };
-        settle().call().await.context("settle simulation failed")?;
+        //settle().call().await.context("settle simulation failed")?;
         settle().send().await.context("settle execution failed")?;
```
e00E, line 79: Should the call be commented out?

solver/src/interactions.rs
```diff
@@ -2,4 +2,23 @@
...
+use crate::encoding;
+use anyhow::anyhow;
+use anyhow::Result;
+use primitive_types::H160;
 pub use uniswap::UniswapInteraction;
+use web3::types::Bytes;
+
+fn encode_interaction(
+    target: H160,
+    calldata: Bytes,
```
e00E, line 14: I prefer Vec<u8> directly instead of web3::Bytes because I don't feel the wrapper adds anything.

e00E (approved):

solver/src/interactions/uniswap.rs
```diff
@@ -36,42 +57,61 @@ mod tests {
...
-        let encoded = cursor.into_inner();
-        assert_eq!(&encoded[0..20], contract.address().as_fixed_bytes());
-        assert_eq!(encoded[20..23], [0, 1, 4]);
-        let call = &encoded[23..];
-        let signature = [0x38u8, 0xed, 0x17, 0x39];
+
+        // Verify Approve
+        let mut approve_call = cursor.into_inner();
+        assert_eq!(&approve_call[0..20], token_in.as_fixed_bytes());
+        assert_eq!(approve_call[20..23], [0, 0, 68]);
```
josojo, line 94: Maybe a tiny comment that this is about the length of the calldata can help to understand it.

solver/src/driver.rs
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2F3o6ZsSRkXifoEJfpMA%2Fgiphy-downsized-medium.gif&token=great_catch" data-gitmeme-token="great_catch"></a>

solver/src/interactions/uniswap.rs
```diff
@@ -1,33 +1,54 @@
...
     pub amount_out_min: U256,
     pub token_in: H160,
     pub token_out: H160,
-    pub payout_to: H160,
 }
 
 impl Interaction for UniswapInteraction {
     fn encode(&self, writer: &mut dyn std::io::Write) -> Result<()> {
+        self.encode_approve(writer)?;
+        self.encode_swap(writer)
```
josojo, line 21: On the contract level the unsiwap router interaction and the approve interaction are different interaction. Here they are kind of in a common interaction struct.

Your choice makes sense, as these two are tightly coupled. But I would not have expected them under a common UniswaapInteraction just because they are different ones on the smart contract level.

josojo (approved):
You are rocking it.


Maybe I would decouple the approval interaction from the uniswap interaction, once you are doing this refactoring:
```
// TODO(fleupold) Only set allowance if we need to
```

This would be nice, as we could reuse the code for a token approval against curve etc.

solver/src/interactions/uniswap.rs
fleupold: I do think they are very related. To me an interaction in the context of the solver is a recipe of how to use some on-chain liquidity (settle against a smart contract order). Each component should be responsible for deciding what it needs to realize the desired trade. In this case if we have never traded against that token on Uniswap it means encoding two interactions, otherwise one.

Similarly I'd see some other implementation of that trade dealing with Curve, Balancer or whatsoever. We can share the logic of how to encode an approval somewhere more general.

We could rename the method to `encode_interactions` as in the smart contract an interaction is one call and here one order can result in multiple interactions.

solver/src/interactions/uniswap.rs
fleupold: I'd like to refactor these tests a bit to make them slightly easier to read and change later on. Will add a comment for now.

PR #142, opened 2020-12-16
Fix price logic in two_order_settlement

### LINKED ISSUE #66: Change naive amm algorithm price vector calculation
When I created this I made it match the master version of the contract but it looks like that will change slightly in https://github.com/gnosis/oba-contracts/pull/203 .

fleupold:
I believe the clearing prices in the two order settlement are inverted. The logic should be if I need to sell a lot of some token (large sell amount) the value of this token should be small (small sell token price) and vice versa.

E.g. if I'm willing to sell 20k USD for 1 BTC the price of USD (sell token) is 1 (buy amount) and of BTC (buy token) is 20k (sell amount).

I noticed this when trying to settle the e2e test with uniswap.

### Test Plan
Adjusted unit tests

### DIFF
--- a/solver/src/naive_solver/two_order_settlement.rs
+++ b/solver/src/naive_solver/two_order_settlement.rs
@@ -141,8 +141,8 @@ fn amm_match(sell_a: &OrderCreation, sell_b: &OrderCreation) -> Option<TwoOrderS
 
     Some(TwoOrderSettlement {
         clearing_prices: maplit::hashmap! {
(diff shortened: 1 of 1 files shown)


josojo (approved):
In the contracts, we require:
```
        // amount_x * price_x = amount_y * price_y
```

hence, we have a good catch!

e00E:
I made #66 about this but forgot about it.

e00E (approved):

PR #143, opened 2020-12-16
E2E test with Uniswap interaction

fleupold:
And here we are! This tests adjusts the existing e2e logic to make the 2nd order smaller than the first one, requiring for us to hit Uniswap and trade the excess via there. This means we should be good for the demo once I configure the solver to run in staging.

### Test Plan
`cargo test -p e2e`

### DIFF
--- a/e2e/tests/uniswap_trade_test.rs
+++ b/e2e/tests/uniswap_trade_test.rs
@@ -19,7 +19,7 @@ const ORDER_PLACEMENT_ENDPOINT: &str = "/api/v1/orders/";
 
 #[tokio::test]
 async fn test_with_ganache() {
-    tracing_setup::initialize("debug");
+    tracing_setup::initialize("warn,orderbook=debug,solver=debug");
     let http = Http::new(NODE_HOST).expect("transport failure");
     let web3 = Web3::new(http);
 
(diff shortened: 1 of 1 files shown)


josojo (approved):
NOICE.

No other changes required? You guys rock!

e2e/tests/uniswap_trade_test.rs
```diff
@@ -80,19 +80,19 @@ async fn test_with_ganache() {
...
     tx!(
         solver,
-        token_a.approve(uniswap_router.address(), to_wei(100))
+        token_a.approve(uniswap_router.address(), to_wei(100_000))
     );
     tx!(
         solver,
-        token_b.approve(uniswap_router.address(), to_wei(100))
+        token_b.approve(uniswap_router.address(), to_wei(100_000))
     );
```
e00E, line 88: Maybe it would be cleaner to have the uniswap liquidity come from a different account that is unrelated to the solver/the settlement. Just so it's clear that with the other PR the solver will make sure the token is approved in uniswap on it's own.

e2e/tests/uniswap_trade_test.rs
```diff
@@ -170,14 +170,14 @@ async fn test_with_ganache() {
...
-    assert_eq!(balance, to_wei(100));
+    assert_eq!(balance, to_wei(80));
 
     let balance = token_a
         .balance_of(trader_b.address())
         .call()
         .await
         .expect("Couldn't fetch TokenA's balance");
-    assert_eq!(balance, to_wei(100));
+    assert_eq!(balance, 62500000000000000000u128.into());
```
e00E, line 180: As a sanity check we could also check that the respective sell token has decreased by the expected amount.

e00E (approved):

bh2smith (approved):
Very nice!

PR #144, opened 2020-12-16
Pass in Private Key for settlement

fleupold:
This PR makes it so that we can specify a specific account when interacting with the settlement contract (requiring for any non local demo). We will have to make sure that any account used here is whitelisted on the smart contract level.

### Test Plan
See that without specifying a private key we can no longer start the solver.

cargo run --bin solver

After exporting `PRIVATE_KEY` it works.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1967,6 +1967,7 @@ dependencies = [
  "anyhow",
  "async-trait",
  "contracts",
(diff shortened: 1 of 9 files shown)


e00E (approved):

josojo (approved):
Nice!

PR #146, opened 2020-12-17
Allow order builder to create full orders

e00E:
I would like to use it in more places for which this will be useful.

### Test Plan
still compiles

### DIFF
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -63,64 +63,69 @@ pub struct OrderCreation {
 }
 
 #[derive(Default)]
-pub struct OrderCreationBuilder {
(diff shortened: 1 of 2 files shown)


josojo (approved):

model/src/lib.rs
```diff
@@ -63,64 +63,69 @@ pub struct OrderCreation {
 }
 
 #[derive(Default)]
-pub struct OrderCreationBuilder {
-    order_creation: OrderCreation,
-}
+pub struct OrderBuilder(Order);
```
fleupold, line 66: 👍

fleupold (approved):

PR #147, opened 2020-12-17
Getting ready for mainnet

fleupold:
This PR prepares the backend for a mainnet demo. As far as I can tell we only need the updated addresses (URLs can be configured in kubernetes) and provide a hacky way of ensuring we use enough gas. Given that ethcontracts ships with web3 gas estimation, I thought the easiest way is to just hardcode a factor for now.

### Test Plan

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1957,6 +1957,7 @@ dependencies = [
  "contracts",
  "ethcontract",
  "futures 0.3.8",
+ "hex",
  "hex-literal",
  "jsonrpc-core",
  "maplit",
--- a/contracts/build.rs
+++ b/contracts/build.rs
@@ -19,12 +19,16 @@ fn main() {
     generate_contract("ERC20Mintable", hashmap! {});
     generate_contract(
         "UniswapV2Router02",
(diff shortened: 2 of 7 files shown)


e00E:
Why mainnet for the demo? I feel we're wasting money on real transactions and making things more complicated with gas prices.
We also aren't ready to run the solver on mainnet unsupervised because we could end up wasting gas on failing transactions over and over again.

e00E (approved):

fleupold:
The demo on rinkeby is quite clumsy (the prices are way off and there is basically only one pair DAI/WETH which can be used to demo a trade). Also the tooling around (e.g. etherscan) doesn't show as nicely what is happening and a demo with real value is just more "impressive" in general.

I agree that we will be "waisting" real money on gas, but I'm not worried regarding failing txs (in my local tests so far we never executed a failing tx). Since we are not enforcing a fee I would think the biggest risk is someone doing a lot of miniscule trades (but I also think it would have to be someone from the team as we didn't share any links publicly yet).
Maybe we can test it and then only turn it on for the demo (outside of it we either don't run the solver or use a very small gas fraction).

josojo (approved):

fleupold:
I guess regardless of the decision on which network to do the demo, it's ok to merge this (we can control if we want to use it in kubernetes).

PR #153, opened 2020-12-17
Run pull requests on github actions

e00E:
I'm not yet sure what the best way to do deployments is so they continue
to be run by travis.
ganache is run without docker because while github actions supports
docker services I could not figure out how to pass arguments to them.

### Test Plan
Github actions should run for this PR.
After merging travis should run for the deployment.
Also check that merge when green works with this PR.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -1,0 +1,29 @@
+name: pull request
+on: [pull_request]
+jobs:
+  rust:
+    runs-on: ubuntu-latest
+    env:
+      CARGO_INCREMENTAL: 0
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
+          npm install ganache-cli
+          node_modules/.bin/ganache-cli --networkId 5777 --gasLimit 10000000 --gasPrice 0&
+      - run: |
+          rustup --version
+          rustup show
+          cargo --version
+          cargo fmt --version
+          cargo clippy --version
+      - run: cargo fmt --all -- --check
+      - run: cargo run --bin deploy --features bin
+        working-directory: contracts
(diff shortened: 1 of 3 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  rust:
+    runs-on: ubuntu-latest
+    env:
+      CARGO_INCREMENTAL: 0
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
+          npm install ganache-cli
+          node_modules/.bin/ganache-cli --networkId 5777 --gasLimit 10000000 --gasPrice 0&
```
fleupold, line 12: nit: for debugging purposes, is it possible to pipe the output into a file and cat it at the end?

.travis.yml
```diff
@@ -7,44 +7,9 @@ os: linux
 dist: focal
 
 jobs:
-  fast_finish: true
   include:
-    - name: "Test Rust"
-      language: rust
-      rust: stable
-      env:
-        - SCCACHE_VERSION=0.2.13
```
fleupold, line 16: How does caching work on github actions?

fleupold (approved):
Nice!

josojo (approved):

.github/workflows/pull-request.yaml
e00E: should be possible but when would we print it? the job stops when something fails so just printing at the end wouldn't help in that case. if I see an easy to achieve this i'll do it but if not it isn't worth it to make the build script more complicated.

.travis.yml
e00E: caching looked a little more complicated so i didn't add it to this pr. will measure if sccache speeds up build and look more into caching in another pr.

PR #154, opened 2020-12-17
Filter settled order

josojo:
As the proper solution for updating the settled orders should only be implemented after we have the db solution, this PR delivers an interim solution: Each order is checked for the filledAmount in the smart contract directly via an RPC call.

Improvement options:
- batched request: Here, I would have to investigate more on how to do it. Currently, the rpc calls are just happening in parallel. I think it is good enough for an implementation that is anyway just temporary.
- proper testing: Here, I need to investigate how to mock the smart contract calls. 
- better readability: Actually I tried to implement a sequential order filtering that is much easier to read:
```
async fn remove_settled_orders(&self, settlement_contract: &GPv2Settlement) {
        let mut orders = self.orders.write().await;
         orders.retain(async move |uid, _| self.has_not_yet_been_settled(*uid).await);
    }
  ```
  But I was not smart enough to get this async operation into the iter, even though I tried to work with stream::iter...
  
  
I think we should do the proper testing. The other options are not so important to me.

### Test Plan
Unit tests.

### DIFF
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -1,10 +1,14 @@
+use contracts::GPv2Settlement;
+use futures::future::join_all;
+use futures::join;
+use model::{DomainSeparator, Order, OrderCreation, OrderMetaData, OrderUid};
+use primitive_types::U256;
 use std::{
     collections::{hash_map::Entry, HashMap},
     time::SystemTime,
 };
-
-use model::{DomainSeparator, Order, OrderCreation, OrderMetaData, OrderUid};
 use tokio::sync::RwLock;
+use tracing::info;
 
 #[derive(Debug, Eq, PartialEq)]
 pub enum AddOrderError {
@@ -75,14 +79,62 @@ impl OrderBook {
     }
 
     // Run maintenance tasks like removing expired orders.
-    pub async fn run_maintenance(&self) {
-        self.remove_expired_orders(now_in_epoch_seconds()).await;
+    pub async fn run_maintenance(&self, settlement_contract: &GPv2Settlement) {
+        let remove_order_future = self.remove_expired_orders(now_in_epoch_seconds());
+        let remove_settled_orders_future = self.remove_settled_orders(settlement_contract);
+        join!(remove_order_future, remove_settled_orders_future);
     }
 
     async fn remove_expired_orders(&self, now_in_epoch_seconds: u64) {
         // TODO: use the timestamp from the most recent block instead?
         let mut orders = self.orders.write().await;
         orders.retain(|_, order| has_future_valid_to(now_in_epoch_seconds, &order.order_creation));
+    }
+
+    async fn remove_settled_orders(&self, settlement_contract: &GPv2Settlement) {
+        let uids_to_be_removed = self.get_uids_of_settled_orders(settlement_contract).await;
+        info!(
+            "removing following settled orders from orderbook: {:?}",
+            uids_to_be_removed
+        );
+        let mut orders = self.orders.write().await;
+        orders.retain(|uid, _| uids_to_be_removed.contains(uid));
+    }
+
+    async fn get_uids_of_settled_orders(
+        &self,
+        settlement_contract: &GPv2Settlement,
+    ) -> Vec<OrderUid> {
+        let orders = self.orders.read().await;
+        let uid_futures = orders.iter().map(|(uid, _)| async move {
+            (
+                *uid,
+                self.has_not_yet_been_settled(*uid, settlement_contract)
+                    .await,
+            )
+        });
+        let uid_pairs: Vec<(OrderUid, bool)> = join_all(uid_futures).await;
+        uid_pairs
+            .iter()
+            .filter(|(_, not_yet_settled)| *not_yet_settled)
+            .map(|(uid, _)| *uid)
+            .collect()
+    }
+
+    async fn has_not_yet_been_settled(
+        &self,
+        uid: OrderUid,
+        settlement_contract: &GPv2Settlement,
+    ) -> bool {
+        let filled_amount = settlement_contract
+            .filled_amount(uid.0.to_vec())
+            .call()
+            .await
+            .unwrap_or(U256::MAX);
+        // As a simplification the function is returning false,
+        // if the order was already partially settled
+        // or if it was canceled.
+        !filled_amount.gt(&U256::zero())
     }
 
     fn order_creation_to_order(&self, user_order: OrderCreation) -> Result<Order, AddOrderError> {
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -1,6 +1,5 @@
 use contracts::GPv2Settlement;
-use futures::future::join_all;
-use futures::join;
+use futures::{future::join_all, join};
 use model::{DomainSeparator, Order, OrderCreation, OrderMetaData, OrderUid};
 use primitive_types::U256;
 use std::{
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -121,7 +121,7 @@ impl OrderBook {
             .filled_amount(uid.0.to_vec())
             .call()
             .await
-            .unwrap_or(U256::MAX);
+            .unwrap_or(U256::zero());
         // As a simplification the function is returning the uid,
         // if the order was already partially settled
         // or if it was canceled.
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -107,34 +107,30 @@ impl OrderBook {
     ) -> Vec<OrderUid> {
         let orders = self.orders.read().await;
         let uid_futures = orders.iter().map(|(uid, _)| async move {
-            (
-                *uid,
-                self.has_not_yet_been_settled(*uid, settlement_contract)
-                    .await,
-            )
+            self.has_uid_been_settled(*uid, settlement_contract).await
         });
-        let uid_pairs: Vec<(OrderUid, bool)> = join_all(uid_futures).await;
-        uid_pairs
-            .iter()
-            .filter(|(_, not_yet_settled)| *not_yet_settled)
-            .map(|(uid, _)| *uid)
-            .collect()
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: orderbook/src/orderbook.rs [hunk 1]
```diff
@@ -75,8 +78,9 @@ impl OrderBook {
     }
 
     // Run maintenance tasks like removing expired orders.
-    pub async fn run_maintenance(&self) {
+    pub async fn run_maintenance(&self, settlement_contract: &GPv2Settlement) {
         self.remove_expired_orders(now_in_epoch_seconds()).await;
+        self.remove_settled_orders(settlement_contract).await;
```
fleupold, line 83: nit: these can also run in parallel right?

orderbook/src/orderbook.rs
```diff
@@ -85,6 +89,52 @@ impl OrderBook {
...
+    async fn has_not_yet_been_settled(
+        &self,
+        uid: OrderUid,
+        settlement_contract: &GPv2Settlement,
+    ) -> bool {
+        let filled_amount = settlement_contract
+            .filled_amount(uid.0.to_vec())
+            .call()
+            .await
+            .unwrap_or(U256::MAX);
```
fleupold, line 131: In order to test this, this call would have to be decoupled from the orderbook class. You could either achieve this by instead of passing GPv2Settlement in the constructor you pass a `OrderFillmentQuerying` trait which exposes `filled_amount(order_id)`. In tests you can mock this to your liking, in prod you implement a struct that forwards this call to an inner `GPv2Settlement`.

Instead of passing a trait (which I think is more object oriented, but functionally no different as long as there is just one call you intend to mock) you could also pass ing a `filled_amount` function into the constructor (@e00E usually prefers this approach). In prod this function captures GPv2Settlement and forwards the call, in test you create a local closure just returning the amounts you want.

orderbook/src/orderbook.rs
```diff
@@ -85,6 +89,52 @@ impl OrderBook {
...
+        // As a simplification the function is returning false,
```
fleupold, line 132: This requires adding a filled amount to the order user data. Could you create a task for it (probably a nice basic addition to add soon)?

orderbook/src/orderbook.rs
```diff
@@ -85,6 +89,52 @@ impl OrderBook {
...
+
+    async fn get_uids_of_settled_orders(
+        &self,
+        settlement_contract: &GPv2Settlement,
+    ) -> Vec<OrderUid> {
+        let orders = self.orders.read().await;
+        let uid_futures = orders.iter().map(|(uid, _)| async move {
+            (
+                *uid,
+                self.has_not_yet_been_settled(*uid, settlement_contract)
```
fleupold, line 110: ubernit: this could just return `Option<OrderUid>`, Some signalling this order has been settled.

orderbook/src/orderbook.rs
```diff
@@ -85,6 +89,52 @@ impl OrderBook {
...
+                    .await,
+            )
+        });
+        let uid_pairs: Vec<(OrderUid, bool)> = join_all(uid_futures).await;
+        uid_pairs
+            .iter()
+            .filter(|(_, not_yet_settled)| *not_yet_settled)
+            .map(|(uid, _)| *uid)
```
fleupold, line 118: Ubernit: consider using [filter_map](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map) directly?

fleupold (approved):
Looks very clean! I think this is sufficient for the demo, although given that the maintenance job only runs periodically I think we might still be running in some errors (at least on the first subsequent solver run).

Would checking order validity synchronously on the get_orders path be a temporary option (I realize that as we get more orders this will easily become too expensive)?

Generally, I think it would be nice to have a "new block" scheduler which triggers the on-chain specific maintenance. Internally this can start with polling frequently but eventually I think it would be nice to "react" directly to new blocks as the full node confirms them (potentially using websockets). 
This way we would hopefully get notified at the same time as the driver realizes the tx was mined and as such would have the orderbook cleaned by the time it is queried it again. Alternatively we can use blocknative to get a subscription on certain events.

Maybe before we add more maintenance tasks (approval, balance checks, etc) we should figure out the right architecture going forward.

orderbook/src/orderbook.rs
e00E: Since we have easy to run integration tests now you could add a check for filled orders being removed to our existing test. I like how when doing it like that we don't need to adapt the code under test.

e00E:
> Would checking order validity synchronously on the get_orders path be a temporary option (I realize that as we get more orders this will easily become too expensive)?

Another temporary alternative I've thought of to this PR now, is that we could just have an in memoy set of orders we have attempted to settle in the solver and filter incoming orders based on that. We would add orders to that set regardless of whether the settlement succeeded or failed.

e00E:
> better readability: Actually I tried to implement a sequential order filtering that is much easier to read:

Something like this should work:

```rust
        let all_order_uids = self
            .orders
            .read()
            .await
            .keys()
            .into_iter()
            .copied()
            .collect::<Vec<_>>();
        for order_uid in all_order_uids {
            if has_been_settled(uid, settlement_contract).await {
                self.orders.write().await.remove(order_uid);
            }
        }
```

orderbook/src/orderbook.rs
josojo: https://github.com/gnosis/gp-v2-services/issues/155

josojo:
>Another temporary alternative I've thought of to this PR now, is that we could just have an in memoy set of orders we have attempted to settle in the solver and filter incoming orders based on that. We would add orders to that set regardless of whether the settlement succeeded or failed.

This could generate smooth UX for the demo. Though I don't like it too much, as it only works with one solver setup, and I expect soon several solvers. (Solver team is marching on... )

>Would checking order validity synchronously on the get_orders path be a temporary option (I realize that as we get more orders this will easily become too expensive)?

Yes! Sounds not too bad. I will not include it in this PR, but we could do it in another PR.

orderbook/src/orderbook.rs
josojo: I like both suggestions. THanks guys!

I will pick one of Felix's suggestion to close the PR and then I will try to add Valentin's option additional with a new PR to the e2e test, if this is okay for you guys.

orderbook/src/orderbook.rs
josojo: I looked into it, but I did not find an example on how to filter map a touple on the first argument only...

orderbook/src/orderbook.rs
e00E: That's okay but the way I imagined it is that the e2e obsoletes the task Felix suggested so then implementing that first would be wasted.

josojo: I kinda like to test functions without e2e tests, as this gives us quicker feedback during development processes. I always think of the F>I>R>S>t principle.  Though, yeah, in this particular instance, it might make sense to just have the e2e test.

I coded the e2e test in the newest commit.

orderbook/src/orderbook.rs
josojo: I finally understood your suggestion. Yeah. that is cool. And then the filter_map is also easier. I will try to do it.

orderbook/src/orderbook.rs
```diff
@@ -1,10 +1,14 @@
+use contracts::GPv2Settlement;
+use futures::future::join_all;
+use futures::join;
```
e00E, line 3:
```suggestion
use futures::{future::join_all, join};
```
(might need cargo fmt after this)

orderbook/src/orderbook.rs
```diff
@@ -85,6 +91,48 @@ impl OrderBook {
...
+
+    async fn get_uids_of_settled_orders(
+        &self,
+        settlement_contract: &GPv2Settlement,
+    ) -> Vec<OrderUid> {
+        let orders = self.orders.read().await;
+        let uid_futures = orders.iter().map(|(uid, _)| async move {
+            self.has_uid_been_settled(*uid, settlement_contract).await
+        });
+        let uid_pairs: Vec<Option<OrderUid>> = join_all(uid_futures).await;
```
e00E, line 112: Have to be a little careful here since all futures run in parallel and there might be a number of connections limit on the node. Doing sequentially doesn't have this problem but is slower. I feel like sequentially might be safer until we use a batched transport.

orderbook/src/orderbook.rs
```diff
@@ -85,6 +91,48 @@ impl OrderBook {
...
+    async fn has_uid_been_settled(
+        &self,
+        uid: OrderUid,
+        settlement_contract: &GPv2Settlement,
+    ) -> Option<OrderUid> {
+        let filled_amount = settlement_contract
+            .filled_amount(uid.0.to_vec())
+            .call()
+            .await
+            .unwrap_or(U256::MAX);
```
e00E, line 125: This unwrap_or means that if there is a connection error we will consider all orders as filled. Maybe on error it should consider as not filled?

josojo: >Maybe on error it should consider as not filled?

My thought was: if the whole price calculation fails due to an already filled order, the damage is much bigger than not settling one particular order

e00E: The problem is that if the node is unreachable this will trigger for all orders and remove all of them. And if the node is unreachable then the solver can't do it's work anyway.

orderbook/src/orderbook.rs
josojo: Yeah, maybe it is safer. Though, I don't think this is really a risk for our own nodes. They are setup with the following config:
```
            "--tx-queue-size=8192",
```
Assuming one call does not take longer than 1 second to work on, this setting allows us to easily have more than 8000 requests per second

e00E: If that does indeed control the number of connections then it's probably fine.

orderbook/src/orderbook.rs
josojo: Yeah, I changed it. If there is an already filled order, then the computation is for nothing, but at least in the next run the orders can be settled again.

e00E (approved):

PR #156, opened 2020-12-18
Cache rust files in CI

e00E:
I tested manually caching, using sccache and finally this pre made
action. The action works very well and is a clean one line solution.
Internally it caches the cargo registry and the target folder but not
crates from this workspace. The cache gets invalidated if Cargo.lock
changes.
sccache makes the build more complicated and doesn't give a speedup when
we already cache like this.
A PR were no dependencies change causes only the top level crates from this workspace to be rebuilt in which case the whole ci run completes in 1 minute.

### Test Plan
After this PR we should have faster CI times.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -3,10 +3,9 @@ on: [pull_request]
 jobs:
   rust:
     runs-on: ubuntu-latest
-    env:
-      CARGO_INCREMENTAL: 0
     steps:
       - uses: actions/checkout@v2
+      - uses: Swatinem/rust-cache@v1
       - run: |
           npm install ganache-cli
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -3,10 +3,9 @@ on: [pull_request]
 jobs:
   rust:
     runs-on: ubuntu-latest
-    env:
-      CARGO_INCREMENTAL: 0
```
e00E, line 7: Removed this because it's managed by the cache action already.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia4.giphy.com%2Fmedia%2F3o72FcJmLzIdYJdmDe%2Fgiphy.gif&token=applause" data-gitmeme-token="applause"></a>

josojo (approved):

PR #164, opened 2020-12-22
Makind Order::default() return a valid order

### LINKED ISSUE #96: Enforce valid order signature at type level?
https://github.com/gnosis/gp-v2-services/pull/93#discussion_r538619943

Felix:
> I'd slightly prefer if we used UserOrder::default and in there create a real signature that is valid. Creating an invalid order creation and then mutating it into something allows to create semantically invalid data (not so important for the tests but also alluding to my general preference of checking signature at order creation time).

fleupold:
Fixes #96

While this does not ensure a valid order at type level (I could still create an order and set the signature to something bad), this makes our "default" instance coherent (having a valid signature, valid expiry date and having a metadata that corresponds to the underlying creation date).

It simplifies the tests slightly (no manual mutation needed in order to create valid orders e.g. for the handler tests) and overall should make it simpler to debug issues when tests are failing (as weird data inconsistencies should not be the case anymore).

### Test Plan
cargo test

### DIFF
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -4,7 +4,7 @@ use crate::{h160_hexadecimal, u256_decimal, DomainSeparator, Signature, TokenPai
 use chrono::{offset::Utc, DateTime, NaiveDateTime};
 use hex_literal::hex;
 use primitive_types::{H160, U256};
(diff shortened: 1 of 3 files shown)


e00E (approved):

josojo (approved):
nice

PR #166, opened 2020-12-23
Only encode uniswap approval when needed

e00E:
to save gas. This is coupled with a change to the Interaction trait to make the encoding function async. An alternative to this is to move this work into the constructor of the interaction itself or add another function to the trait like prepare_interaction. I don't like the latter because you would always call it before encode anyway and could make the internal state of the interaction more complicated. Doing it in the constructor has the advantage that the self encode stays simpler but it makes it more complicated to create an interaction in the first place.
For the uniswap interaction specifically we could later add a cache to remember that we have already given approval.

### Test Plan
e2e test

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -50,6 +50,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
 
 [[package]]
+name = "async-trait"
+version = "0.1.42"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8d3a45e77e34375a7923b1e8febb049bb011f064714a8e17a1a616fef01da13d"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "atty"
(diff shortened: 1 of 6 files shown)


fleupold (commented):
I thought about this challenge when adding the approve call and eventually decided against making `encode` async because to me it felt like something that should be non-blocking. My thinking was that after the solution has been found we want to settle it as soon as possible and therefore should encourage to do all potentially long running computation beforehand (or in parallel to the solution finding).

Curious to hear if there are maybe other "interactions" we can think of now that would absolutely require async encoding.

That being said, the logic of the change looks good (I like how you worked around the unit test which for me was the final trigger to pass in this flag from above).

e00E:
> encourage to do all potentially long running computation beforehand
How can we do this when don't know what tokens will be in the solution beforehand? The only way I see is to check this for all tokens that we have orders for (and update the list for new orders) and then pass that info to the uniswap interaction.

fleupold:
Yes, basically my thinking was we could do it at the same time when we query the reserve sizes (which also needs to happen optimistically for all possible pairs). At that point we can also remember that in case we want to use a given pool, it might need to be approved first.

e00E:
I see. I agree we should do it at that point if we have to do perform such an action anyway.

PR #169, opened 2020-12-24
Sightly less naive solver

fleupold:
This PR introduces a solver that can match an arbitrary number of buy/sell orders on both sides of the same token pair in a single settlement with a uniswap AMM.

The mathematical logic behind the approach has been described here: https://docs.google.com/document/d/1jS22wxbCqo88fGsqEMZgRQgiAcHlPqxoMw3CJTHst6c

For now we assume some simplifications just to get the basic algorithm reviewed before the logic is getting too large.

Follow ups:
- If there is no uniswap trade required (perfect balanced of both sides), we should not report a price of 0 but take the existing reserves
- We need to verify limit prices and iteratively kick out the least favorable excess order in case we don't match and try solving again
- We should better handle the case if the desired trade amount exceeds the reserve in the Uniswap pool (right now this would result in a bigint -> u256 conversion error.

### Test Plan
Added unit tests

### DIFF
--- a/solver/src/naive_solver/multi_order_solver.rs
+++ b/solver/src/naive_solver/multi_order_solver.rs
@@ -74,7 +74,7 @@ fn split_into_contexts(
             reserve: u256_to_bigint(
                 reserves
                     .get(&order.buy_token)
-                    .ok_or(anyhow!("No reserve for token {}", &order.buy_token))?,
+                    .ok_or_else(|| anyhow!("No reserve for token {}", &order.buy_token))?,
             ),
             ..Default::default()
         });
@@ -83,7 +83,7 @@ fn split_into_contexts(
             reserve: u256_to_bigint(
                 reserves
                     .get(&order.sell_token)
-                    .ok_or(anyhow!("No reserve for token {}", &order.sell_token))?,
+                    .ok_or_else(|| anyhow!("No reserve for token {}", &order.sell_token))?,
             ),
             ..Default::default()
         });
--- a/solver/src/naive_solver/multi_order_solver.rs
+++ b/solver/src/naive_solver/multi_order_solver.rs
@@ -1,0 +1,369 @@
+#![allow(dead_code)]
+
+use super::single_pair_settlement::{AmmSwapExactTokensForTokens, SinglePairSettlement};
+use crate::settlement::Trade;
+use anyhow::{anyhow, Result};
+use model::order::{OrderCreation, OrderKind};
+use num::{bigint::Sign, BigInt};
+use std::collections::HashMap;
+use web3::types::{Address, U256};
+
+#[derive(Default)]
+struct TokenContext {
+    address: Address,
+    reserve: BigInt,
+    buy_volume: BigInt,
+    sell_volume: BigInt,
+}
+
+impl TokenContext {
+    pub fn is_excess(&self, other: &TokenContext) -> bool {
+        &self.reserve * (&other.sell_volume - &other.buy_volume)
+            < &other.reserve * (&self.sell_volume - &self.buy_volume)
+    }
+}
+
+/**
+ * Computes a settlement using orders of a single pair and the direct AMM between those tokens.
+ * Returns an error if computation fails, orders are not already filtered for a specific token pair, or the reserve information for that
+ * pair is not available.
+ */
+pub fn solve(
+    orders: impl Iterator<Item = OrderCreation> + Clone,
+    reserves: &HashMap<Address, U256>,
+) -> Result<SinglePairSettlement> {
+    let (context_a, context_b) = split_into_contexts(orders.clone(), reserves)?;
+    let (shortage, excess) = if context_a.is_excess(&context_b) {
+        (context_b, context_a)
+    } else {
+        (context_a, context_b)
+    };
+    let uniswap_out = compute_uniswap_out(&shortage, &excess);
+    let uniswap_in = bigint_to_u256(&compute_uniswap_in(&uniswap_out, &shortage, &excess))?;
+    let uniswap_out = bigint_to_u256(&uniswap_out)?;
+    let interaction = if uniswap_out > U256::zero() {
+        Some(AmmSwapExactTokensForTokens {
+            amount_in: uniswap_in,
+            amount_out_min: uniswap_out,
+            token_in: excess.address,
+            token_out: shortage.address,
+        })
+    } else {
+        // TODO(fleupold) set correct clearing prices when supply/demand already match current uniswap price
+        None
+    };
+    // TODO(fleupold) check that all orders comply with the computed price. Otherwise, remove the least favorable excess order and try again.
+    Ok(SinglePairSettlement {
+        clearing_prices: maplit::hashmap! {
+            shortage.address => uniswap_in,
+            excess.address => uniswap_out,
+        },
+        trades: orders.into_iter().map(Trade::fully_matched).collect(),
+        interaction,
+    })
+}
+
+fn split_into_contexts(
+    orders: impl Iterator<Item = OrderCreation>,
+    reserves: &HashMap<Address, U256>,
+) -> Result<(TokenContext, TokenContext)> {
+    let mut contexts = HashMap::new();
+    for order in orders {
+        contexts.entry(order.buy_token).or_insert(TokenContext {
+            address: order.buy_token,
+            reserve: u256_to_bigint(
+                reserves
+                    .get(&order.buy_token)
+                    .ok_or(anyhow!("No reserve for token {}", &order.buy_token))?,
+            ),
+            ..Default::default()
+        });
+        contexts.entry(order.sell_token).or_insert(TokenContext {
+            address: order.sell_token,
+            reserve: u256_to_bigint(
+                reserves
+                    .get(&order.sell_token)
+                    .ok_or(anyhow!("No reserve for token {}", &order.sell_token))?,
+            ),
+            ..Default::default()
+        });
+        match order.kind {
+            OrderKind::Buy => {
+                contexts.get_mut(&order.buy_token).unwrap().buy_volume +=
+                    u256_to_bigint(&order.buy_amount)
+            }
+            OrderKind::Sell => {
+                contexts.get_mut(&order.sell_token).unwrap().sell_volume +=
+                    u256_to_bigint(&order.sell_amount)
+            }
+        }
+    }
+    if contexts.len() != 2 {
+        return Err(anyhow!("Orders contain more than two tokens"));
+    }
+    let mut contexts = contexts.drain().map(|(_, v)| v);
+    Ok((contexts.next().unwrap(), contexts.next().unwrap()))
+}
+
+/**
+ * Given information about the shortage token (the one we need to take from Uniswap) and the excess token (the one we gie to Uniswap), this function
+ * computes the exact out_amount required from Uniswap to perfectly match demand and supply at the effective Uniswap price (the one used for that in/out swap).
+ *
+ * The derivation of this formula is described in https://docs.google.com/document/d/1jS22wxbCqo88fGsqEMZgRQgiAcHlPqxoMw3CJTHst6c/edit
+ * It assumes GP fee (φ) to be 1 and Uniswap fee (Φ) to be 0.997
+ */
+fn compute_uniswap_out(shortage: &TokenContext, excess: &TokenContext) -> BigInt {
+    let numerator_minuend = 997 * (&excess.sell_volume - &excess.buy_volume) * &shortage.reserve;
+    let numerator_subtrahend =
+        1000 * (&shortage.sell_volume - &shortage.buy_volume) * &excess.reserve;
+    let denominator = (1000 * &excess.reserve) + (997 * (&excess.sell_volume - &excess.buy_volume));
+
+    (numerator_minuend - numerator_subtrahend) / denominator
+}
+
+/**
+ * Given the desired amount to receive and the state of the pool, this computes the required amount
+ * of tokens to be sent to the pool.
+ * Taken from: https://github.com/Uniswap/uniswap-v2-periphery/blob/4123f93278b60bcf617130629c69d4016f9e7584/contracts/libraries/UniswapV2Library.sol#L53
+ */
+fn compute_uniswap_in(out: &BigInt, shortage: &TokenContext, excess: &TokenContext) -> BigInt {
+    1000 * out * &excess.reserve / (997 * (&shortage.reserve - out)) + 1
+}
+
+fn u256_to_bigint(input: &U256) -> BigInt {
+    let mut bytes = [0; 32];
+    input.to_big_endian(&mut bytes);
+    BigInt::from_bytes_be(Sign::Plus, &bytes)
+}
+
+fn bigint_to_u256(input: &BigInt) -> Result<U256> {
+    let (sign, bytes) = input.to_bytes_be();
+    if sign == Sign::Minus {
+        return Err(anyhow!("Negative BigInt to U256 conversion"));
+    }
+    if bytes.len() > 32 {
+        return Err(anyhow!("BigInt too big for U256 conversion"));
+    }
+    Ok(U256::from_big_endian(&bytes))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn to_wei(base: u32) -> U256 {
+        U256::from(base) * U256::from(10).pow(18.into())
+    }
+
+    #[test]
+    fn finds_clearing_price_with_sell_orders_on_both_sides() {
+        let token_a = Address::from_low_u64_be(0);
+        let token_b = Address::from_low_u64_be(1);
+        let orders = vec![
+            OrderCreation {
+                sell_token: token_a,
+                buy_token: token_b,
+                sell_amount: to_wei(40),
+                buy_amount: to_wei(30),
+                kind: OrderKind::Sell,
+                partially_fillable: false,
+                ..Default::default()
+            },
+            OrderCreation {
+                sell_token: token_b,
+                buy_token: token_a,
+                sell_amount: to_wei(100),
+                buy_amount: to_wei(90),
+                kind: OrderKind::Sell,
+                partially_fillable: false,
+                ..Default::default()
+            },
+        ];
+
+        let reserves = maplit::hashmap! {
+            token_a => to_wei(1000),
+            token_b => to_wei(1000)
+        };
+        let result = solve(orders.clone().into_iter(), &reserves).unwrap();
+
+        // Make sure the uniswap interaction is using the correct direction
+        let interaction = result.interaction.unwrap();
+        assert_eq!(interaction.token_in, token_b);
+        assert_eq!(interaction.token_out, token_a);
+
+        // Make sure the sell amounts +/- uniswap interaction satisfy min_buy amounts
+        assert!(orders[0].sell_amount + interaction.amount_out_min >= orders[1].buy_amount);
+        assert!(orders[1].sell_amount - interaction.amount_in > orders[0].buy_amount);
+
+        // Make sure the sell amounts +/- uniswap interaction satisfy expected buy amounts given clearing price
+        let price_a = result.clearing_prices.get(&token_a).unwrap();
+        let price_b = result.clearing_prices.get(&token_b).unwrap();
+
+        // Multiplying sellAmount with priceA, gives us sell value in "$", divided by priceB gives us value in buy token
+        // We should have at least as much to give (sell amount +/- uniswap) as is expected by the buyer
+        let expected_buy = orders[0].sell_amount * price_a / price_b;
+        assert!(orders[1].sell_amount - interaction.amount_in >= expected_buy);
+
+        let expected_buy = orders[1].sell_amount * price_b / price_a;
+        assert!(orders[0].sell_amount + interaction.amount_out_min >= expected_buy);
+    }
+
+    #[test]
+    fn finds_clearing_price_with_sell_orders_on_one_side() {
+        let token_a = Address::from_low_u64_be(0);
+        let token_b = Address::from_low_u64_be(1);
+        let orders = vec![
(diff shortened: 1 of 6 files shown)


solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -0,0 +1,369 @@
...
+
+/**
+ * Computes a settlement using orders of a single pair and the direct AMM between those tokens.
+ * Returns an error if computation fails, orders are not already filtered for a specific token pair, or the reserve information for that
+ * pair is not available.
+ */
+pub fn solve(
+    orders: impl Iterator<Item = OrderCreation> + Clone,
+    reserves: &HashMap<Address, U256>,
+) -> Result<SinglePairSettlement> {
```
e00E, line 34: There should be a high level solve function that can handle all orders and splits them by pair internally like for the other naive solver. If you want to add that later that's fine.

e00E, line 34: I'd rather have this not return a result and instead panic on error because the errors (more than one pair in the orders, internal bigint conversion fails, reserve missing) are programmer errors that should never happen.
Could change the types to encode that directly. Instead of Iterator Order we could take already split
sell_orders, buy_orders, sell_reserve, buy_reserve which removes the reserve hash map too.
The one error that we cannot prevent is if we get a bigint that is larger than u256 max but in that case we should just return None instead of an error.

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -0,0 +1,369 @@
...
+    let (context_a, context_b) = split_into_contexts(orders.clone(), reserves)?;
+    let (shortage, excess) = if context_a.is_excess(&context_b) {
+        (context_b, context_a)
+    } else {
+        (context_a, context_b)
+    };
+    let uniswap_out = compute_uniswap_out(&shortage, &excess);
+    let uniswap_in = bigint_to_u256(&compute_uniswap_in(&uniswap_out, &shortage, &excess))?;
+    let uniswap_out = bigint_to_u256(&uniswap_out)?;
+    let interaction = if uniswap_out > U256::zero() {
```
e00E, line 44: I'm ok with either version just pointing out this alternative
```suggestion
    let interaction = if !uniswap_out.is_zero() {
```

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -0,0 +1,369 @@
...
+    for order in orders {
+        contexts.entry(order.buy_token).or_insert(TokenContext {
+            address: order.buy_token,
+            reserve: u256_to_bigint(
+                reserves
+                    .get(&order.buy_token)
+                    .ok_or_else(|| anyhow!("No reserve for token {}", &order.buy_token))?,
+            ),
+            ..Default::default()
+        });
```
e00E, line 80: Alternatively, to avoid querying the hashmap again:
```rust
        let buy_context = contexts.entry(order.buy_token).or_insert(TokenContext {
            address: order.buy_token,
            reserve: u256_to_bigint(
                reserves
                    .get(&order.buy_token)
                    .ok_or_else(|| anyhow!("No reserve for token {}", &order.buy_token))?,
            ),
            ..Default::default()
        });
        if matches!(order.kind, OrderKind::Buy) {
            buy_context.buy_volume += u256_to_bigint(&order.buy_amount)
        }
```
and similar for sell_context

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -0,0 +1,369 @@
...
+    }
+    if contexts.len() != 2 {
+        return Err(anyhow!("Orders contain more than two tokens"));
+    }
+    let mut contexts = contexts.drain().map(|(_, v)| v);
+    Ok((contexts.next().unwrap(), contexts.next().unwrap()))
+}
+
+/**
+ * Given information about the shortage token (the one we need to take from Uniswap) and the excess token (the one we gie to Uniswap), this function
```
e00E, line 109:
```suggestion
 * Given information about the shortage token (the one we need to take from Uniswap) and the excess token (the one we give to Uniswap), this function
```

solver/src/naive_solver/multi_order_solver.rs
e00E, line 80: Use insert_with to avoid doing u256 conversion even entry already exists:
```suggestion
        contexts
            .entry(order.buy_token)
            .or_insert_with(|| TokenContext {
                address: order.buy_token,
                reserve: u256_to_bigint(reserves.get(&order.buy_token).expect("no reserve token")),
                ..Default::default()
            });
```
but that way can't return a result which I think we shouldn't do anyway.

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -0,0 +1,369 @@
+#![allow(dead_code)]
+
+use super::single_pair_settlement::{AmmSwapExactTokensForTokens, SinglePairSettlement};
+use crate::settlement::Trade;
+use anyhow::{anyhow, Result};
+use model::order::{OrderCreation, OrderKind};
+use num::{bigint::Sign, BigInt};
```
e00E, line 7: There is a BigUInt type which is probably more appropriate in most places.

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -0,0 +1,369 @@
...
+ * The derivation of this formula is described in https://docs.google.com/document/d/1jS22wxbCqo88fGsqEMZgRQgiAcHlPqxoMw3CJTHst6c/edit
+ * It assumes GP fee (φ) to be 1 and Uniswap fee (Φ) to be 0.997
+ */
+fn compute_uniswap_out(shortage: &TokenContext, excess: &TokenContext) -> BigInt {
+    let numerator_minuend = 997 * (&excess.sell_volume - &excess.buy_volume) * &shortage.reserve;
+    let numerator_subtrahend =
+        1000 * (&shortage.sell_volume - &shortage.buy_volume) * &excess.reserve;
+    let denominator = (1000 * &excess.reserve) + (997 * (&excess.sell_volume - &excess.buy_volume));
+
+    (numerator_minuend - numerator_subtrahend) / denominator
```
e00E, line 121: The two BigInt divisions (here and in compute_uniswap_in) will round down. Just pointing that out in case it isn't intentional as I don't fully understand the algorithm from the doc yet so it's not clear to me if that is intended.
There is also a BigRational type if that's helpful.

solver/src/naive_solver/multi_order_solver.rs
fleupold: The reason I used BigInt internally (rather than e.g. U512 which would probably be sufficient width wise), is that some of the arithmetic can underflow (e.g. when computing excess.sell_volume - excess.buy_volume in compute_uniswap_out).

Do you think it's better to have the internal structs use the more strict type and only convert in the methods directly?

solver/src/naive_solver/multi_order_solver.rs
fleupold: I was going to plug this in to the naive solver's settle method instead of `settle_pair` so I can reuse the `organize_orders_by_token_pair` logic.

fleupold: Agree that the programming errors should be turned into assertions. One logic error that I'm not sure how to handle is if our shortage is larger than the reserve inside one of the Uniswap pools. In this case you suggest to also return None?

solver/src/naive_solver/multi_order_solver.rs
fleupold: Nice, I was struggling with the borrow checker, but splitting up the match into two ifs works nicely. Thanks!

solver/src/naive_solver/multi_order_solver.rs
fleupold: We need to get a full number of out tokens for the uniswap payload. My plan was to deal with rounding errors by introducing a very small arbitrary fee (oba fee in the doc). This way we should be able to retain some small surplus in the contract and defend a bit against race conditions. If we simply round up here, my worry is that the in amount will also be larger and we might not have enough sell amounts on the other side of the trade.

solver/src/naive_solver/single_pair_settlement.rs
```diff
@@ -9,9 +9,9 @@ use std::collections::HashMap;
 
 #[derive(Debug)]
 pub struct SinglePairSettlement {
-    clearing_prices: HashMap<H160, U256>,
-    trades: Vec<Trade>,
-    interaction: Option<AmmSwapExactTokensForTokens>,
+    pub clearing_prices: HashMap<H160, U256>,
+    pub trades: Vec<Trade>,
+    pub interaction: Option<AmmSwapExactTokensForTokens>,
```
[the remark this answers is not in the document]
fleupold: Yes, this was a bit hacky just to have access to it. I think eventually we can remove the existing solver and rearrange the structs accordingly. Just wanted to leave as much as possible in place as this PR is already quite large. I'm happy to move the two structs up a level.

fleupold (commented):
Thanks for the review. I don't think I will be able to address all feedback today, but will update the PR again before I leave on vacation.

solver/src/naive_solver/multi_order_solver.rs
e00E: I think you're referring to

> We should better handle the case if the desired trade amount exceeds the reserve in the Uniswap pool (right now this would result in a bigint -> u256 conversion error.

but I don't understand this. This function doesn't know about the real state of the uniswap pool, it just creates a generic amm interaction. So where does that conversion error happen?

solver/src/naive_solver/single_pair_settlement.rs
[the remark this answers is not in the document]
e00E: I'm fine doing this later when we remove the other solver. No need to make this nice during the temporary phase where we have both.

solver/src/naive_solver/multi_order_solver.rs
e00E: If that underflow is temporary in that function and it always receives and returns positive numbers then I would use the BigInteger only inside there and keep the interface clean.

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -0,0 +1,369 @@
...
+    sell_volume: U256,
+}
+
+impl TokenContext {
+    pub fn is_excess(&self, other: &TokenContext) -> bool {
+        u256_to_bigint(&self.reserve)
+            * (u256_to_bigint(&other.sell_volume) - u256_to_bigint(&other.buy_volume))
+            < u256_to_bigint(&other.reserve)
+                * (u256_to_bigint(&self.sell_volume) - u256_to_bigint(&self.buy_volume))
+    }
```
e00E, line 25: I think this is (5) from the doc but in case isn't the fee missing?

josojo:
@e00E will you take the command for this and the other PRs of Felix? 

I think it would be beneficial to have them all merged, such that the PR dependencies are not getting too long. What do you think?

e00E:
I am a little worried that I don't understand the math behind the new solver well enough so I felt it would be better to wait for Felix. This PR will have merge conflicts with my solver-trait and http-solver PRs too so I plan on merging those first and then fixing the conflicts here but still waiting for Felix to be back for the final merge of this.

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -0,0 +1,369 @@
...
+}
+
+/**
```
nlordell, line 28: ubernit: Maybe use the rustdoc comments here? `///`

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -0,0 +1,369 @@
...
+    let mut contexts = contexts.drain().map(|(_, v)| v);
+    (contexts.next().unwrap(), contexts.next().unwrap())
+}
+
+/**
+ * Given information about the shortage token (the one we need to take from Uniswap) and the excess token (the one we give to Uniswap), this function
+ * computes the exact out_amount required from Uniswap to perfectly match demand and supply at the effective Uniswap price (the one used for that in/out swap).
+ *
+ * The derivation of this formula is described in https://docs.google.com/document/d/1jS22wxbCqo88fGsqEMZgRQgiAcHlPqxoMw3CJTHst6c/edit
+ * It assumes GP fee (φ) to be 1 and Uniswap fee (Φ) to be 0.997
```
nlordell, line 109: I'm not sure this is right... or does this assume that a fee discount will be computed to adjust for the Uniswap trade?

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -0,0 +1,369 @@
...
+    let numerator_minuend = 997
+        * (u256_to_bigint(&excess.sell_volume) - u256_to_bigint(&excess.buy_volume))
+        * u256_to_bigint(&shortage.reserve);
+    let numerator_subtrahend = 1000
+        * (u256_to_bigint(&shortage.sell_volume) - u256_to_bigint(&shortage.buy_volume))
+        * u256_to_bigint(&excess.reserve);
+    let denominator = (1000 * u256_to_bigint(&excess.reserve))
+        + (997 * (u256_to_bigint(&excess.sell_volume) - u256_to_bigint(&excess.buy_volume)));
+    bigint_to_u256(&((numerator_minuend - numerator_subtrahend) / denominator))
+        .expect("uniswap_out should always be U256 compatible if excess is chosen correctly")
```
nlordell, line 121: This looks correct to me :+1:

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -0,0 +1,369 @@
...
+}
+
+/**
+ * Given the desired amount to receive and the state of the pool, this computes the required amount
+ * of tokens to be sent to the pool.
+ * Taken from: https://github.com/Uniswap/uniswap-v2-periphery/blob/4123f93278b60bcf617130629c69d4016f9e7584/contracts/libraries/UniswapV2Library.sol#L53
+ */
+fn compute_uniswap_in(out: U256, shortage: &TokenContext, excess: &TokenContext) -> U256 {
+    U256::from(1000) * out * excess.reserve / (U256::from(997) * (shortage.reserve - out)) + 1
```
nlordell, line 130: I like the Uniswap approach to rounding issues: `+ 1` :stuck_out_tongue:.

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -0,0 +1,369 @@
...
+struct TokenContext {
+    address: Address,
+    reserve: U256,
+    buy_volume: U256,
+    sell_volume: U256,
+}
+
+impl TokenContext {
+    pub fn is_excess(&self, other: &TokenContext) -> bool {
+        u256_to_bigint(&self.reserve)
```
nlordell, line 21: I think this is missing the Uniswap fees, (assuming this returns true if `self` is in excess):
```
1000*self.reserve*(other.sell_volume-other.buy_volume) < 997*other.reserve*(self.sell_volume-self.buy_volume) ```

File: solver/src/naive_solver/multi_order_solver.rs [hunk 14]
```diff
@@ -0,0 +1,369 @@
...
+#[derive(Default)]
+struct TokenContext {
+    address: Address,
+    reserve: U256,
+    buy_volume: U256,
+    sell_volume: U256,
+}
+
+impl TokenContext {
+    pub fn is_excess(&self, other: &TokenContext) -> bool {
```
nlordell, line 20: naming-nit: When I read this I wasn't immediately sure of the direction, maybe this can be renamed to:
```suggestion
    pub fn is_excess(&self, deficit: &TokenContext) -> bool {
```

So its really clear that `self` is the one is excess?

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -0,0 +1,369 @@
...
+        })
+    } else {
+        // TODO(fleupold) set correct clearing prices when supply/demand already match current uniswap price
+        None
+    };
+    // TODO(fleupold) check that all orders comply with the computed price. Otherwise, remove the least favorable excess order and try again.
+    SinglePairSettlement {
+        clearing_prices: maplit::hashmap! {
+            shortage.address => uniswap_in,
+            excess.address => uniswap_out,
```
nlordell, line 60: Can't this theoretically have 0-prices? Is that what you meant by the `TODO` on line 53?

nlordell (approved):
The math looks mostly good to me (one comment about the excess/deficit token computation). The one concern I have is around the `φ == 1` assumption with the fees not reflected in the total buy and sell amounts for the token. I think this can lead to the order's limit price **with** fees being worse than the Uniswap price (although I'm not 100% sure about this). I imagine you intend on using the `feeDiscount` mechanism with this solver (in that case, my concern is likely a moot point)?

solver/src/naive_solver/multi_order_solver.rs
bh2smith: Is this `u256_to_bigint` something that could be upstreamed to primitive types? Looks a bit ugly here.

solver/src/naive_solver/multi_order_solver.rs
vkgnosis: Yes I would also expect None in that case since this set of orders cannot be solved.

solver/src/naive_solver/multi_order_solver.rs
fleupold: Yes.

solver/src/naive_solver/multi_order_solver.rs
fleupold: This is basically assuming there is no GP related fee (only the 0.997 uniswap fee). As for the real solver, I think the fee can be deducted from the sell/buy amounts before the orders are passed into the solver engine. The solvers themselves can run on the assumption that there is no protocol fee.

solver/src/naive_solver/multi_order_solver.rs
fleupold: For the fee, see below. I'm assuming 0 protocol fee for now (even when introduced my assumption is we can deduct the fee before we pass in the orders). 

> Is this u256_to_bigint something that could be upstreamed to primitive types?

I'm not sure bigint is common enough to justify such a conversion (before we were doing it with std types but here we are dealing with two separate crates).

solver/src/naive_solver/multi_order_solver.rs
fleupold: For the excess I thought it was not necessary to consider the uniswap fee. Wouldn't we otherwise have a potentially asymmetric situation where neither token is `in_excess`? I'm basically comparing the ratio of buy and sell tokens to the current spot price of the AMM (which doesn't consider fees)

nlordell: > Wouldn't we otherwise have a potentially asymmetric situation where neither token is in_excess

Correct, this happens when the price is in the 60bps spread from the Uniswap spot price. In that regions, there should be no Uniswap trade.

In that case, the price can be computed from the total buy and sell amounts

vkgnosis (approved):

solver/src/naive_solver/multi_order_solver.rs
nlordell: Ah OK, I remember the decision now, we would pass the instance without the fees (deducted from the balances and not amounts I think right?) and then the driver would adjust the `feeDiscount` to account for any Uniswap trades. In this way, the solver and the naive solver can share the same code for computing fee discounts.

solver/src/naive_solver/multi_order_solver.rs
fleupold: I see, yet isn't it slightly more intuitive to still consider the excess token based on spot price without accounting for fees? The outcome should be the same if we don't touch the AMM anyways, right?

nlordell: > isn't it slightly more intuitive to still consider the excess token based on spot price without accounting for fees?

Good point, I also agree with this.

I just did a once over on the code, and I think it may end up panicking in the case that the price ends up in the Uniswap spread when computing the Uniswap out amount (as I imagine the result would be negative in that case):
```rust
.expect("uniswap_out should always be U256 compatible if excess is chosen correctly")
```

fleupold: I see your point now. It does indeed panic in some cases (e.g. with two sell orders 1001 for 1000 and 1000 for 1000). 

I have changed the `is_excess` function to consider the fees and in case there is no overlap in either way create a "direct match" using the current spot price. This is likely suboptimal in case two orders are overlapping but both are on one side of the current spot price (e.g. spot price at $1 I'm willing to sell for $1.01, you are willing to buy for $1.03) but it might be good enough for now.

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -0,0 +1,437 @@
...
+fn solve_without_uniswap(
+    orders: impl Iterator<Item = OrderCreation> + Clone,
+    context_a: &TokenContext,
+    context_b: &TokenContext,
+) -> SinglePairSettlement {
+    SinglePairSettlement {
+        clearing_prices: maplit::hashmap! {
+            context_a.address => context_b.reserve,
+            context_b.address => context_a.reserve,
+        },
```
nlordell, line 60: In the future, this should probably not be so naive as a single +/- 1 atom can cause a 0.3% jump in price. Absolutely not important for this PR though.

nlordell (approved):
Looks great!

As a side note, I mentioned on Slack another way to solve these batches by solving for prices = final reserves. I personally think that algorithm is better suited for a naive solver as there:
- is no special case when the price is somewhere in the 0.6% Uniswap spread (oe even exactly on the spot price)
- is continuous over input buy and sell amounts.
- difficult to compute on-chain (as it requires a square root) but not an issue off-chain here (and rounding should be able to be trivially done benefiting orders selling the deficit token).

Just to keep in mind. That being said, this not so naive solver is AWESOME and can already do a bunch of work settling batches as is!

PR #170, opened 2020-12-27
Use multi order solver

fleupold:
This PR makes use of the slightly less naive solver. For this we need to fetch the current "reserves" of the direct AMM between the tokens we are trying to match and pass it down into the logic.

To get the reserves, we need to query the `UniswapPair` (I added this contract artefact) from the UniswapFactory instance and then call `get_reserves` on it.

Overall the architecture is a little bit hacky as a lot of components are passed through large parts of the stack. I'd like to abstract "smart contract orders" such as Uniswap or Sushiswap pools (but later also Balancer, Curve, Prediction Market outcome tokens, etc) away from the solver interface so that we only hand an abstract set of orders which can be UserOrders or some form of smart contract order trait. Each solver can then work with the subset of orders they can handle.

However, in order to make sure the less naive solver works and to make the demo slightly more usable for external people it might be worthwhile merging like this and then refactor.

### Test Plan
I needed to remove the naive solver unit tests as we now rely on a smart contract call which is not yet easily mockable. However, the logic should still be covered by the multi order solver specific unit tests and the e2e test.

I can now trade e.g. STAKE for WXDAI on our xdai Demo interface (https://pr91--gpv2swap.review.gnosisdev.com/#/swap) directly without requiring a counter trade. This makes the demo more usable for external people who want to try it out and get a feeling for the e2e flow (in order to demonstrate coincedence of wants better we might have to prolong the solver sleep interval a bit to have more opportunity for overlap).

https://blockscout.com/poa/xdai/tx/0x1aedffc0c9a51bc1ace77303a8e5b622f4624a506a55d4ab9327abe15d3452a8/token-transfers is the first settlement using this solver.

### DIFF
--- a/e2e/tests/uniswap_trade_test.rs
+++ b/e2e/tests/uniswap_trade_test.rs
@@ -168,7 +168,8 @@ async fn test_with_ganache() {
         std::time::Duration::from_secs(10),
     );
     let solver = solver::naive_solver::NaiveSolver {
-        uniswap: uniswap_router,
+        uniswap_router,
+        uniswap_factory,
         gpv2_settlement: gp_settlement.clone(),
     };
     let mut driver =
@@ -181,14 +182,14 @@ async fn test_with_ganache() {
         .call()
         .await
         .expect("Couldn't fetch TokenB's balance");
-    assert_eq!(balance, to_wei(80));
+    assert_eq!(balance, U256::from(99650498453042316810u128));
 
     let balance = token_a
         .balance_of(trader_b.address())
         .call()
         .await
         .expect("Couldn't fetch TokenA's balance");
-    assert_eq!(balance, 62500000000000000000u128.into());
+    assert_eq!(balance, U256::from(50175363672226073522u128));
 
     // Drive orderbook in order to check the removal of settled order_b
     orderbook.run_maintenance(&gp_settlement).await.unwrap();
--- a/e2e/tests/uniswap_trade_test.rs
+++ b/e2e/tests/uniswap_trade_test.rs
@@ -182,14 +182,14 @@ async fn test_with_ganache() {
         .call()
         .await
         .expect("Couldn't fetch TokenB's balance");
-    assert_eq!(balance, U256::from(99650498453042316810u128));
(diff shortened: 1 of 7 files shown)


e00E (approved):
This makes sense short term and agree we should refactor long term.

In the long term we might want to separate the solver part from the driver part more. Different solvers care about different things like one solver only trying to do direct matches, another one doing uniswap matches and another one using different amms.
From that point of view we could have a minimal solver trait with something like `async fn solve(orders)` and then solvers are responsible themselves for querying for example the uniswap pool.

e00E:
In https://github.com/gnosis/gp-v2-services/pull/176 I added a module for uniswap pool information. Should probably merge that PR first.

e2e/tests/uniswap_trade_test.rs
```diff
@@ -181,14 +182,14 @@ async fn test_with_ganache() {
         .call()
         .await
         .expect("Couldn't fetch TokenB's balance");
-    assert_eq!(balance, to_wei(80));
+    assert_eq!(balance, U256::from(99650498453042316810u128));
```
nlordell, line 185: ubernit: Maybe include `_` separators in the numerical literal so its easy to read:
```suggestion
    assert_eq!(balance, U256::from(99_650_498_453_042_316_810u128));
```

You can even enforce this with clippy with the `#![deny(clippy::unreadable_literal)]` https://rust-lang.github.io/rust-clippy/master/#unreadable_literal

model/src/lib.rs
```diff
@@ -100,6 +100,12 @@ impl TokenPair {
     }
 }
 
+impl Default for TokenPair {
```
nlordell, line 103: ubernit: Conceptually, it doesn't really make sense to me for `TokenPair` to implement `Default` since this default isn't really valid and is really only useful for tests (maybe the implementation should be `#[cfg(test)]`?). However, this is very nit-picky of me so feel free to leave as is.

nlordell (approved):
Look great! Just some tiny nit-picks.

PR #174, opened 2020-12-29
Make solver a trait

e00E:
In order to disentangle solvers from the driver we introduce a Solver
trait that is implemented for the naive solver.
Similarily we will implement this trait for the python solver.
Solvers are responsible for gathering the information they need
themself. So the python solver implementation would in the solve
implementation  fetch the state of the uniswap pools and then call the
python solver with that information.

### Test Plan
CI, no logic change

### DIFF
--- a/solver/src/main.rs
+++ b/solver/src/main.rs
@@ -1,6 +1,6 @@
 use ethcontract::PrivateKey;
 use reqwest::Url;
-use solver::driver::Driver;
+use solver::{driver::Driver, naive_solver::NaiveSolver};
 use std::time::Duration;
 use structopt::StructOpt;
 #[derive(Debug, StructOpt)]
@@ -54,6 +54,10 @@ async fn main() {
             .expect("couldn't load deployed settlement");
     let orderbook =
         solver::orderbook::OrderBookApi::new(args.orderbook_url, args.orderbook_timeout);
-    let mut driver = Driver::new(settlement_contract, uniswap_contract, orderbook);
+    let solver = NaiveSolver {
+        gpv2_settlement: settlement_contract.clone(),
+        uniswap: uniswap_contract,
+    };
+    let mut driver = Driver::new(settlement_contract, orderbook, Box::new(solver));
     driver.run_forever().await;
 }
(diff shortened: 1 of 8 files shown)


solver/src/naive_solver.rs
```diff
@@ -10,6 +11,22 @@ use model::{
 use primitive_types::U512;
 use std::{cmp::Ordering, collections::HashMap};
 
+pub struct NaiveSolver {
+    pub uniswap: UniswapV2Router02,
+    pub gpv2_settlement: GPv2Settlement,
+}
+
+#[async_trait::async_trait]
```
josojo, line 19: So much less painful than it used to be :)

josojo (approved):
Nice. Thanks for making this small step before implementing the python solver

solver/src/main.rs
```diff
@@ -54,6 +54,10 @@ async fn main() {
             .expect("couldn't load deployed settlement");
     let orderbook =
         solver::orderbook::OrderBookApi::new(args.orderbook_url, args.orderbook_timeout);
-    let mut driver = Driver::new(settlement_contract, uniswap_contract, orderbook);
+    let solver = NaiveSolver {
+        gpv2_settlement: settlement_contract.clone(),
+        uniswap: uniswap_contract,
+    };
```
bh2smith, line 60: ubernit: in the test file instantiation, the arguments are swapped. Just sayin'

solver/src/main.rs
```diff
@@ -54,6 +54,10 @@ async fn main() {
...
+    let mut driver = Driver::new(settlement_contract, orderbook, Box::new(solver));
```
bh2smith, line 61: nit: Should we call this `orderbook_api` here as well (like in the testfile)

bh2smith (approved):

solver/src/solver.rs
```diff
@@ -0,0 +1,8 @@
+use crate::settlement::Settlement;
+use anyhow::Result;
+use model::order::Order;
+
+#[async_trait::async_trait]
+pub trait Solver {
+    async fn solve(&self, orders: Vec<Order>) -> Result<Option<Settlement>>;
+}
```
fleupold, line 8: nit: docs for this public trait (in particular the semantics of the return type) would be nice.

PR #176, opened 2020-12-31
Add http interface to python solver

e00E:
using the new solver trait. The code can succesfully send an instance to
the solver but not yet decode the result.

This also contains a function to get the uniswap pool for a token pair like in https://github.com/gnosis/gp-v2-services/pull/170 .

### Test Plan
Start the solver and run `cargo test real_solver -- --ignored --nocapture` .


contracts/build.rs
```diff
@@ -25,7 +25,15 @@ fn main() {
             100 => Address::from_str("1C232F01118CB8B424793ae03F870aa7D0ac7f77").unwrap(),
         },
     );
-    generate_contract("UniswapV2Factory", hashmap! {});
+    generate_contract(
+        "UniswapV2Factory",
+        hashmap! {
+            1 => Address::from_str("5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6f").unwrap(),
+            4 => Address::from_str("5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6f").unwrap(),
+            100 => Address::from_str("A818b4F111Ccac7AA31D0BCc0806d64F2E0737D7").unwrap(),
```
bh2smith, line 33: Given that we are prepared for xDAI here, is the `UniswapV2Factory` analogous to the honeyswap contracts on xDAI?

solver/src/http_solver.rs
```diff
@@ -0,0 +1,270 @@
...
+            request = request.header("X-API-KEY", api_key);
+        }
+        let request = request.json(&body);
+
+        let response = request
+            .send()
+            .await
+            .context("failed to send solver request")?
+            .error_for_status()
+            .context("solver response was not success")?;
```
bh2smith, line 164:
```suggestion
            .context("solver response was unsuccessful")?;
```

bh2smith (approved):
Very nice work here!

contracts/build.rs
e00E: I am not sure. I took this part from Felix's PR where he also had to add this contract. It should be this exact contract https://github.com/Uniswap/uniswap-v2-core/blob/master/contracts/UniswapV2Factory.sol just on xdai. Whether honeyswap also uses this or they have their own version I don't know.

solver/src/http_solver.rs
```diff
@@ -0,0 +1,270 @@
...
+                (index.to_string(), order)
+            })
+            .collect()
+    }
+
+    async fn uniswaps(&self, orders: &[Order]) -> Result<HashMap<String, UniswapModel>> {
+        // TODO: use a cache
+        let mut uniswaps = HashMap::new();
+        for order in orders {
+            let pair = order.order_creation.token_pair().expect("invalid order");
```
e00E, line 106: I'm having doubts about this. This code is getting the uniswap pool for every token pair in the orders. But what if there is a third pool for which there is no order which nonetheless connects the token better. We probably want the direct pools AND every individual token with ETH since that is usually the biggest pool. Then the question is whether the solver is aware of this too.

josojo: >We probably want the direct pools AND every individual token with ETH since that is usually the biggest pool.
Yeah agreed.

>Then the question is whether the solver is aware of this too.
Sure, the solvers will be able to trade "rings" through many pools

josojo (approved):
Very clean PR. Awesome.

Let's finish the API spec discussions, then modify it here quickly and we have it soon running 👍

e00E:
I'd still like to merge this now even if the spec changes. We will just make a PR with required changes then. That keeps them smaller.

PR #177, opened 2020-12-31
Document database

### LINKED ISSUE #175: Database design
We want to store a database to persistently store for example order data in so that we do not lose it when the service restarts.

For the number of orders we will have a simple postgresql database should be fine. There we can have a table for storing all orders indexed by their uuid. There can be an index on valid_to to efficiently retrieve only valid orders or we can delete orders as they expire or store them in a different place.

In another table we can keep track of all `Trade` events with which we can calculate the current filled amount for partially fillable orders. This would work like the event based orderbook from v1 except that it is a lot simpler here as we are only adding up the filled amount for each uid. This table would replace the on disk event cache. Once an order is no longer valid we can delete the trade events to free up space or keep them for historical analysis.

e00E:
This is the database setup for the orderbook service.

closes #175 

### Test Plan
n/a


database/README.md
```diff
@@ -0,0 +1,14 @@
+This document explains the database that the orderbook api uses.
+
+We need to store orders persistently even if the orderbook service goes. For this we use a PostgreSQL database. The schema can be found in `schema.sql`.
```
josojo, line 3: PostgreSQL database allows us these easy resets of the data base, that we need to ensure that our event listening is robust against reorgs, right?

josojo (approved):
LGTM

database/README.md
e00E: The way I designed the schema it is easy to remove all events newer than some block number. This is exactly what our event based code needs.

database/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+In addition to the orders, we also store trade events. This allows us to keep track of filled amounts.
+
+To prevent users from spamming the database we should when receiving an order check that
+* the erc20 spending is approved
+* valid_to is not in the past and not more than <1 year> in the future
```
bh2smith, line 9: Why not more than 1 year? This would prevent us from having CMM brackets (which amount to unlimited orders with infinite validity)...

database/README.md
```diff
@@ -0,0 +1,14 @@
...
+* the order owner does not already have more than <1000> valid orders
```
bh2smith, line 10: Seems fair. Do we plan to have a cleanup functionality which archives or purges invalid/expired orders?

database/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+This ensures that the amount of orders we have to keep track of per user is limited and that there is a real cost the user incurs before being able to submit orders because they need to have sent an ethereum transaction.
+
+Technically, this could still be attacked with a fake token that claims that approval has always been granted. This is more evidence ro my belief that we should whitelist erc20 tokens that we will handle.
```
bh2smith, line 14:
```suggestion
Technically, this could still be attacked with a fake token that claims that approval has always been granted. This is more evidence to my belief that we should whitelist erc20 tokens that we will handle.
```

bh2smith (approved):
Looks amazing, just a few comments and questions inline also it appears that `setup.txt` is an empty file. Was this intended?

database/README.md
```diff
@@ -0,0 +1,14 @@
+This document explains the database that the orderbook api uses.
+
+We need to store orders persistently even if the orderbook service goes. For this we use a PostgreSQL database. The schema can be found in `schema.sql`.
+
+In addition to the orders, we also store trade events. This allows us to keep track of filled amounts.
+
+To prevent users from spamming the database we should when receiving an order check that
+* the erc20 spending is approved
```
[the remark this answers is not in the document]
e00E: For the database it won't matter if the user places the same order again because the primary key is the uid which will be the same. So there can only be that one entry. We will probably detect this when in our create-order endpoint and return an appropriate error message.

database/README.md
e00E: It depends. If we feel the need to clean up space in the database we can delete expired orders or move them to a more long term storage place. Or it might take up so little space that it doesn't matter. I just want to make sure a user cannot place an unlimited amount of orders here.

database/README.md
e00E: I still had in mind that one of the design decisions for v2 is that orders should have a "short" validity period. From the perspective of the database that is nice because it creates a natural limit for how long we *have to* keep track of orders. But there is no technical reason for that. We could totally allow orders valid forever.

PR #178, opened 2021-01-05
Add more network endpoints to openapi

e00E:
It should include all our endpoints.

### Test Plan
CI


josojo (approved):
Thanks for cleaning and completing the list

PR #183, opened 2021-01-06
Implement get_orders order filtering

### LINKED ISSUE #179: Change get_orders endpoint
In openapi this takes 3 required parameters for the user, sell token, buy token.
* This isn't actually implemented. We ignore the parameters and and always return all orders.
* If it were implemented the api would lose the ability to query all orders.

We probably want the ability to query both, all orders and user orders and all user orders based on tokens. We can accomplish this by making the three parameters optional.

e00E:
So that this works as expected for the frontend even when we do not yet
have the order database.

### Test Plan
new unit test


josojo (approved):
Another clean PR :) great.

warp::query is a great tool.

orderbook/src/api/handler.rs
```diff
@@ -82,8 +82,24 @@ pub async fn add_order(
...
+        owner
+            .map(|owner| owner == order.order_meta_data.owner)
+            .unwrap_or(true)
+            && sell_token
+                .map(|token| token == order.order_creation.sell_token)
+                .unwrap_or(true)
+            && buy_token
+                .map(|token| token == order.order_creation.buy_token)
+                .unwrap_or(true)
+    });
```
fleupold, line 102: Very slick!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FcInRB3UQMkkouSUQ3a%2Fgiphy.gif&token=slick" data-gitmeme-token="slick"></a>

fleupold (approved):

Issue #115, opened 2020-12-11
All http errors should return a json struct


e00E:
Like when we can't deserialize an order. This is helpful for the frontend.
This probably involves installing a catch all rejection handler in warp.

PR #184, opened 2021-01-06
Make the orderbook storage a trait as preparation for database

vkgnosis:
I am not sure if we are going to keep the in memory orderbook around but
for the short term it is useful to have this as a trait so that we can
start implementing the postgresql storage.
Long term the in memory orderbook could still be useful for tests so
that no running postgres instance is needed for unit tests.

### Test Plan
tests still pass

### DIFF
--- a/orderbook/src/api/handler.rs
+++ b/orderbook/src/api/handler.rs
@@ -1,5 +1,5 @@
-use crate::orderbook::{AddOrderError, OrderBook};
-
+use crate::storage::{AddOrderResult, Storage};
+use anyhow::Result;
 use chrono::prelude::{DateTime, FixedOffset, Utc};
 use model::{
     order::{OrderCreation, OrderUid},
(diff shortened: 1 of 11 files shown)


orderbook/src/api/handler.rs
```diff
@@ -26,83 +26,100 @@ pub struct FeeInfo {
...
-#[derive(PartialEq, Serialize, Deserialize, Debug)]
+#[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
-pub struct OrderPostError {
-    error_type: String,
-    description: String,
+struct Error<'a> {
+    error_type: &'a str,
+    description: &'a str,
+}
```
vkgnosis, line 34: This is related to https://github.com/gnosis/gp-v2-services/issues/115 .

orderbook/src/storage.rs
```diff
@@ -0,0 +1,34 @@
...
+use contracts::GPv2Settlement;
+use model::order::{Order, OrderCreation, OrderUid};
+
+pub use memory::OrderBook as InMemoryOrderBook;
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum AddOrderResult {
+    Added(OrderUid),
+    DuplicatedOrder,
+    InvalidSignature,
```
vkgnosis, line 14: It works out better to have `Added` be part of this enum instead of being the Ok side of the `add_order`. This allows makes the error for `add_order` always an internal error (like not being able to reach the database).

josojo (approved):
Nice.

I would have directly implemented the database, as I think that the performance loss will be negligible and we don't need the memory one in the future. But having this trait makes the transition easier to test and implement. cool.

PR #185, opened 2021-01-07
HttpSolver login refactor and clippy corrections

josojo:
Allows connecting to pwd protected hosted solver.

PS: Needed to do quite some clippy corrections, due to rust update.

### Test Plan
```
cargo test real_solver -- --ignored --nocapture  
```
with
```
let solver = HttpSolver::new(
            "https://optimizer.dev.gnosisdev.com:8000".parse().unwrap(),
            Some("password from 1password: Look for vault: Dev Optimizer API>),
            SolverConfig {
                max_nr_exec_orders: 100,
                time_limit: 100,
            },
            uniswap,
        );
```

### DIFF
--- a/solver/src/http_solver.rs
+++ b/solver/src/http_solver.rs
@@ -32,7 +32,7 @@ impl SolverConfig {
 pub struct HttpSolver {
     base: Url,
     client: Client,
-    api_key: Option<String>,
+    api_pwd: Option<&'static str>,
     config: SolverConfig,
     uniswap: UniswapV2Factory,
 }
@@ -40,7 +40,7 @@ pub struct HttpSolver {
 impl HttpSolver {
     pub fn new(
         base: Url,
-        api_key: Option<String>,
+        api_pwd: Option<&'static str>,
         config: SolverConfig,
         uniswap: UniswapV2Factory,
     ) -> Self {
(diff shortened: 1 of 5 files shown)


solver/src/http_solver.rs
```diff
@@ -32,15 +32,15 @@ impl SolverConfig {
 pub struct HttpSolver {
     base: Url,
     client: Client,
-    api_key: Option<String>,
+    api_pwd: Option<&'static str>,
```
vkgnosis, line 35: I think we are going to get the base url as a command line argument. Then we can directly encode the username and password there (https://serverfault.com/questions/371907/can-you-pass-user-pass-for-http-basic-authentication-in-url-parameters) so there is no need to manually set username and password here.

josojo: I would prefer my method, as the api_pwd will come from a secret env, instead of a normal parameter visible to anyone. But since this is security critical, I don't have a strong opinion

vkgnosis: Maybe we can setup the solver so that it is only accessible from within our kubernetes cluster. Even if not I don't feel this is really secret like the private keys. It's not a problem if we as the devs know the url with username and password.

josojo: No, I think it is very convenient to test the solver also from our home computers. Then we don't have to setup the solver code.

But okay, then we will display the pwd in the kubernetes config file. That is not really an issue.

PR #186, opened 2021-01-07
Fix clippy warnings for Rust 1.49

vkgnosis:
Extracted for Alex' PR so we can merge this immediately.

### Test Plan
CI, but github is still on 1.48 which is why we didn't notice this


vkgnosis:
Since these are the same changes as Alex did in his PR and I approve of those changes I am going to merge now even though this PR technically didn't get approved yet.

PR #188, opened 2021-01-08
Implement storing and updating of trade events in database

vkgnosis:
which uses a new Database trait that abstracts over the actual sql
queries that still need to be implemented. I am using a trait there to
encapsulate which postgresql library we're using and the specific sql
queries.

### Test Plan
This module cannot easily be tested because it consists entirely of connecting the contract to the database trait. For the same reason I feel unit tests are less important because the actual logic is simple. When the database is fully implemented I plan on switching the e2e to it.

### DIFF
--- a/orderbook/src/trade_events.rs
+++ b/orderbook/src/trade_events.rs
@@ -1,0 +1,147 @@
+use crate::database::{Database, Trade as DbTrade};
+use anyhow::{anyhow, Context, Error, Result};
+use contracts::{
+    g_pv_2_settlement::{event_data::Trade as ContractTrade, Event as ContractEvent},
+    GPv2Settlement,
+};
+use ethcontract::{errors::ExecutionError, Event, EventMetadata};
+use futures::{Stream, StreamExt, TryStreamExt};
+use model::order::OrderUid;
+use std::{convert::TryInto, ops::RangeInclusive, sync::Arc};
+use web3::{Transport, Web3};
+
+// We expect that there is never a reorg that changes more than the last n blocks.
+const MAX_REORG_BLOCK_COUNT: u64 = 25;
+// When we insert new trade events into the database we will insert at most this many in one
+// transaction.
+const INSERT_TRADE_BATCH_SIZE: usize = 250;
+
+pub struct TradeEvents {
+    contract: GPv2Settlement,
+    db: Arc<dyn Database>,
+}
+
+impl TradeEvents {
+    /// Get new events from the contract and insert them into the database.
+    pub async fn update_events(&mut self) -> Result<()> {
+        let range = self.event_block_range().await?;
+        let events = self
+            .past_events(&range)
+            .await
+            .context("failed to get past events")?
+            .ready_chunks(INSERT_TRADE_BATCH_SIZE)
+            .map(|chunk| chunk.into_iter().collect::<Result<Vec<_>, _>>());
+        futures::pin_mut!(events);
+        // We intentionally do not go with the obvious approach of deleting old events first and
+        // then inserting new ones. Instead, we make sure that the deletion and the insertion of the
+        // first batch of events happen in one transaction.
+        // This is important for two reasons:
+        // 1. It ensures that we only delete if we really have new events. Otherwise if fetching new
+        //    events from the node fails for whatever reason we might keep deleting events over and
+        //    over without inserting new ones resulting in the database table getting cleared.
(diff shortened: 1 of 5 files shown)


orderbook/src/trade_events.rs
```diff
@@ -0,0 +1,147 @@
...
+    async fn past_events(
+        &self,
+        block_range: &RangeInclusive<u64>,
+    ) -> Result<impl Stream<Item = Result<DbTrade>>, ExecutionError> {
+        Ok(self
+            .contract
+            .all_events()
+            .from_block((*block_range.start()).into())
+            .to_block((*block_range.end()).into())
+            .block_page_size(100)
```
josojo, line 112: Nit: you can use a constant here and I would also choose the value bigger.

orderbook/src/trade_events.rs
```diff
@@ -0,0 +1,147 @@
...
+
+// We expect that there is never a reorg that changes more than the last n blocks.
+const MAX_REORG_BLOCK_COUNT: u64 = 25;
+// When we insert new trade events into the database we will insert at most this many in one
+// transaction.
+const INSERT_TRADE_BATCH_SIZE: usize = 250;
+
+pub struct TradeEvents {
+    contract: GPv2Settlement,
+    db: Arc<dyn Database>,
```
josojo, line 21: I would have stored the last_handled_blocked as well in here and did not get it from the db just for performance reasons and saving one query. But this is very unimportant, I guess.

josojo (approved):
Nice! 

I have to admit that this is "latently" smarter and cleaner compared to how events are indexed in ido-service. :D Nice.

orderbook/src/trade_events.rs
vkgnosis: I didn't feel that extracting this out to a constant helped document it. I will increase it to 1000.

orderbook/src/trade_events.rs
vkgnosis: I don't think it makes a noticable difference and as you said I prefer this way because it's slightly cleaner.

orderbook/src/trade_events.rs
```diff
@@ -0,0 +1,147 @@
...
+        // 1. It ensures that we only delete if we really have new events. Otherwise if fetching new
+        //    events from the node fails for whatever reason we might keep deleting events over and
+        //    over without inserting new ones resulting in the database table getting cleared.
+        // 2. It ensures that other users of the database are unlikely to see an inconsistent state
+        //    some events have been deleted but new ones not yet inserted. This is important in case
+        //    for example another part of the code calculates the total executed amount of an order.
+        //    If this happened right after deletion but before insertion, then the result would be
+        //    wrong. In theory this could still happen if the last MAX_REORG_BLOCK_COUNT blocks had
+        //    more than INSERT_TRADE_BATCH_SIZE trade events but this is unlikely.
+        // There alternative solutions for 2. but this one is the most practical. For example, we
```
fleupold, line 48: I'm fine with this solution. I also wouldn't mind an opaque Database trait chain where the tip has the in memory version of unfinalized trades and it reads through to a "persistent" Database trait implementation.

However, if consistency is so important, shouldn't we make sure we write all events in a single transaction to the database (rather than page by page)?

fleupold (approved):

orderbook/src/trade_events.rs
vkgnosis: The single transaction is not always going to be possible like when we have been offline for a long time and have accumulated many events. But that situation is very rare and inconsistency isn't a problem during that initial event update anyway. As long as insert_trade_batch_size is large enough it will almost always be in one page.

The tip in memory, persistent in database I have also considered (below this line). It has the advantage that we only ever append to the database but the disadvantage that reads have to go through an extra struct. With this version we could decide to connect the solver directly to the database to avoid the extra path through http and the orderbook api.

fleupold: Maybe a compromise (that addresses both the inconsistency issue and works for full re-syncs) would be to buffer events in a Vec until `block_number_of_most_recent_trade` is exceeded and then commit the first page. After long offline times or full resync `block_number_of_most_recent_trade` should be long in the past as well so the first vector should not be too large.

It might not actually be that important (I was just judging from the code comment that it was) and we might actually never notice the inconsistency, (only if we more than 10 trades per block on avg).

PR #193, opened 2021-01-11
Implement the trade events in postgres

vkgnosis:
Based on the previous place holder trait.

### Test Plan
New test used by CI with real postgres db

### DIFF
--- a/orderbook/Cargo.toml
+++ b/orderbook/Cargo.toml
@@ -24,6 +24,7 @@ primitive-types = "0.7"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 shared-arguments = { path = "../shared-arguments" }
+sqlx = { version = "0.4", default-features = false, features = ["runtime-tokio-native-tls", "postgres"] }
 structopt = { version = "0.3" }
 tokio = { version = "0.2", features =[ "macros", "time"] }
 tracing = "0.1"
--- a/orderbook/src/database.rs
+++ b/orderbook/src/database.rs
@@ -1,7 +1,11 @@
-use anyhow::Result;
+use anyhow::{Context, Result};
+use futures::FutureExt;
 use model::order::OrderUid;
 use primitive_types::U256;
+use sqlx::{Connection, Executor, PgPool, Postgres, Transaction};
+use std::convert::TryInto;
 
+#[derive(Debug, Default)]
 pub struct Trade {
     pub block_number: u64,
     pub log_index: u64,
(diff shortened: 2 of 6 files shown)


orderbook/Cargo.toml
```diff
@@ -24,6 +24,7 @@ primitive-types = "0.7"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 shared-arguments = { path = "../shared-arguments" }
+sqlx = { version = "0.4", default-features = false, features = ["runtime-tokio-native-tls", "postgres"] }
```
vkgnosis, line 27: I decided on sqlx because compared to tokio-postgres it has connection pooling and automatic caching of queries built in. Since this is only used in the database module it wouldn't be too hard to switch later anyway.

orderbook/src/database.rs
```diff
@@ -11,16 +15,148 @@ pub struct Trade {
     pub fee_amount: U256,
 }
 
-/// Abstraction over the database we use to store persistent data like orders and trade events.
-#[async_trait::async_trait]
-pub trait Database {
-    async fn block_number_of_most_recent_trade(&self) -> Result<u64>;
```
vkgnosis, line 17: Changed from trait to struct for simplicity. We can still turn it back into a trait if we need to but I don't want to do it proactively.

orderbook/src/database.rs
```diff
@@ -11,16 +15,148 @@ pub struct Trade {
...
+    pub async fn replace_trades(
         &self,
         delete_from_block_number: u64,
-        new_trades: &[Trade],
-    ) -> Result<()>;
+        trades: Vec<Trade>,
+    ) -> Result<()> {
+        let mut connection = self.pool.acquire().await?;
+        connection
+            .transaction(move |transaction| {
```
vkgnosis, line 66: The way sqlx specifies the lifetimes for the transaction closure means we have to take ownership over the trades.

orderbook/src/database.rs
```diff
@@ -11,16 +15,145 @@ pub struct Trade {
...
+}
+
+impl Database {
+    pub fn new(uri: &str) -> Result<Self> {
+        Ok(Self {
+            pool: PgPool::connect_lazy(uri)?,
+        })
+    }
+
+    pub async fn block_number_of_most_recent_trade(&self) -> Result<Option<u64>> {
```
fleupold, line 31: to account for reorgs (while we are offline), do we also need to store the block_hash with each trade?

orderbook/src/trade_events.rs
```diff
@@ -57,13 +57,13 @@ impl TradeEvents {
             let trades = trades.context("failed to get event")?;
             if !have_deleted_old_events {
                 self.db
-                    .replace_trades(*range.start(), &[])
+                    .replace_trades(*range.start(), trades)
```
fleupold, line 60: Out of curiosity is there a reason you want the deletion and first insertion atomically if the other insertions are not done atomically?

fleupold (approved):
Looks good to me.

orderbook/src/trade_events.rs
vkgnosis: This line was a bug in the previous PR. It should have inserted the trades all along. There is a long comment a couple of lines above that explains this reasoing.

orderbook/src/database.rs
vkgnosis: No. The trade event updating code will always delete events starting at  `block_number_of_most_recent_trade - MAX_REORG_BLOCKS`. If any hash changes that isn't a problem because those blocks are deleted based on number.

PR #198, opened 2021-01-12
Use github native dependabot

vkgnosis:
As https://github.com/marketplace/dependabot-preview says this now a
built in github feature so we can move to that instead.
I like how it has a configuration file that is in the repository.

Before we merge this we should remove the old dependabot application. I don't know how to do that. @fleupold maybe?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+# https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-dependencies-updated-automatically
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"


.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+# https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-dependencies-updated-automatically
+version: 2
+updates:
+  - package-ecosystem: "cargo"
```
fleupold, line 4: Should we set a reviewer as well or does it chose us automatically?

fleupold (approved):
Nice, I deleted dependabot from the gp-v1-services repo.

.github/dependabot.yml
vkgnosis: I think it should still use our team because of the code owneship thing. We will see.

PR #205, opened 2021-01-13
Implement order insertion and querying in postgres

vkgnosis:
The querying part will be extended with filtering in a future PR.

### Test Plan
New tests for order insertion

### DIFF
--- a/orderbook/src/database/orders.rs
+++ b/orderbook/src/database/orders.rs
@@ -1,0 +1,212 @@
+use super::Database;
+use crate::u256_conversions::*;
+use anyhow::{anyhow, Context, Result};
+use bigdecimal::BigDecimal;
+use chrono::{DateTime, Utc};
+use futures::{stream::TryStreamExt, Stream};
+use model::{
+    order::{Order, OrderCreation, OrderKind, OrderMetaData, OrderUid},
+    Signature,
+};
+use primitive_types::H160;
+use std::convert::TryInto;
+
+#[derive(sqlx::Type)]
+#[sqlx(rename = "OrderKind")]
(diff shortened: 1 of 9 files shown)


bh2smith (approved):
Making our own ORM! Sweet deal.

orderbook/src/database/orders.rs
```diff
@@ -0,0 +1,208 @@
...
+                COALESCE(SUM(t.fee_amount), 0) AS sum_fee \
+            FROM orders o LEFT OUTER JOIN trades t ON o.uid = t.order_uid
+            WHERE o.valid_to >= $1 \
+            GROUP BY o.uid;";
+        // To get only not fully executed orders we probably want to use something like
+        // `HAVING sum_sell < orders.sell_amount` but still need to find the best way to make this
+        // pick the correct column based on order type.
+
+        #[derive(sqlx::FromRow)]
+        struct Row {
```
fleupold, line 87: nit: Maybe moving row outside of this function and implementing TryInto<Order> makes the second half of this function smaller and easier to read (assuming it will become more complex when we add filtering).

orderbook/src/u256_conversions.rs
```diff
@@ -0,0 +1,30 @@
+use bigdecimal::BigDecimal;
+use num_bigint::{BigInt, Sign, ToBigInt};
+use primitive_types::U256;
+
+pub fn u256_to_big_int(input: &U256) -> BigInt {
```
fleupold, line 5: I have used this also in the less naive solver. Maybe we can create a util crate to collect common helper functions like this.

fleupold (approved):

PR #206, opened 2021-01-13
Update ethcontract, web3

vkgnosis:
and transitive dependencies.
Context: We just had an ethcontract release with the newest web3 version.

### Test Plan
CI

### DIFF
--- a/e2e/Cargo.toml
+++ b/e2e/Cargo.toml
@@ -5,15 +5,14 @@ edition = "2018"
 
 [dev-dependencies]
 contracts = { path = "../contracts" }
-ethcontract = { version = "0.9",  default-features = false, features = ["http"] }
+ethcontract = { version = "0.10",  default-features = false, features = ["http"] }
 hex-literal = "0.3"
 model = { path = "../model" }
 orderbook = { path = "../orderbook" }
-secp256k1 = "0.17"
+secp256k1 = "0.20"
 reqwest = "0.10"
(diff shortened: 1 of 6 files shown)


bh2smith (approved):

fleupold (approved):

e2e/Cargo.toml
```diff
@@ -5,15 +5,14 @@ edition = "2018"
...
+ethcontract = { version = "0.10",  default-features = false, features = ["http"] }
 hex-literal = "0.3"
 model = { path = "../model" }
 orderbook = { path = "../orderbook" }
-secp256k1 = "0.17"
+secp256k1 = "0.20"
 reqwest = "0.10"
 tokio = { version = "0.2", features =[ "macros"] }
-web3 = { version = "0.13", default-features = false }
+web3 = { version = "0.15", default-features = false }
```
nlordell, line 15: Instead of maintaining a separate `web3` dependency here, why not use the re-exported `web3` crate from `ethcontract::web3`?

**edit**: Idem for the `ethabi`, `jsonrpc-core` and `primitive_types` (unless it makes use of a non-exported type which I didn't see).

vkgnosis: I find it clearer to depends on crates we use directly instead of through a re-export but as you point that has the disadvantage of changing more versions.
Also for example models uses primitive types but it doesn't need all of ethcontract.

PR #207, opened 2021-01-13
Split CI up into {fmt, clippy} and {test}

vkgnosis:
Because they build separate dependencies anyway (clippy compiles like
`check`) so this can speed up CI a little.

Switched around the order of operations for tests so that it runs the fast unit tests first, then the postgres and tests and finally the e2e ganache test.

### Test Plan
CI has separate jobs

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -6,14 +6,11 @@ on:
   push:
     branches: [main]
 jobs:
-  rust:
+  rust-misc:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: Swatinem/rust-cache@v1
-      - run: |
-          npm install ganache-cli
-          node_modules/.bin/ganache-cli --networkId 5777 --gasLimit 10000000 --gasPrice 0&
       - run: |
           rustup --version
           rustup show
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -6,32 +6,38 @@ on:
...
       - run: cargo fmt --all -- --check
-      - run: cargo run --bin deploy --features bin
-        working-directory: contracts
       - run: cargo clippy --locked --workspace --all-features --all-targets -- -D warnings
-      - run: cargo test --locked --workspace --all-features
+  rust-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: Swatinem/rust-cache@v1
```
fleupold, line 26: I would have thought that having to download the cache twice would take longer than the time it takes to run clippy itself on a warm build but it looks like overall the CI time is down from 3m 37s [here](https://github.com/gnosis/gp-v2-services/actions/runs/483087532) to 2:42 on this PR.

But maybe something to keep in mind as our dependencies and thus the cache grows.

fleupold (approved):

.github/workflows/pull-request.yaml
vkgnosis: Jobs run in parallel and have a unique cache (i think). So it's not going to be slower.

fleupold: Right, since we don't own/pay for the infrastructure we don't care about the global efficiency. My point was more that if two tasks have a large combined setup cost (downloading cache + build) and one runs its independent action in a few seconds it might not be optimal to run them in parallel and incur twice the setup cost globally.

vkgnosis: Agree. In this case the shared setup cost is only downloading the dependencies from crates io but not any compilation because of the different compilation targets, so it's quite low.

PR #208, opened 2021-01-13
[LessNaiveSolver] Prune orders that don't satisfy solutions limit price

fleupold:
This PR implements verification of limit prices of the orders we add to our solution (a remaining TODO). 

After computing a solution (initially with all orders for the given token pair regardless of limit price), we check that all trades satisfy their order's limit. This is normally true, if orders have been submitted via the UI since we account for potential slippage and make sure the spread is sufficient. It is not true however, if the AMM's price changed significantly or users created a manual order via the API. 
If a limit price is violated, we remove one order from the problem set and restart computation. 

We remove the order with the most strict limit price of the ones selling the token we are in excess of (the token we are providing to the AMM). By doing so, we reduce the overall excess and as such the AMM price should move less (which might in the next run satisfy other orders that are also in violation with the computed price now but still close enough to the previous AMM price). 

We continue to iterate until eventually there are no more orders in the Set. In this case we return an empty/trivial solution (might be worthwhile to return None instead).

The intuition of the approach is that if we keep removing the "outliers" (worst limit prices) from the problem set, at some point we should end up with a set of orders that can be matched partially with one another and the AMM's liquidity.

### Test Plan

Unit tests added (most of the line count)

### DIFF
--- a/solver/src/naive_solver/multi_order_solver.rs
+++ b/solver/src/naive_solver/multi_order_solver.rs
@@ -36,7 +36,7 @@ pub fn solve(
     pool: &Pool,
 ) -> SinglePairSettlement {
     let mut orders: Vec<OrderCreation> = orders.collect();
-    while orders.len() > 0 {
+    while !orders.is_empty() {
         let (context_a, context_b) = split_into_contexts(orders.clone().into_iter(), pool);
         let solution = solve_orders(orders.clone().into_iter(), &context_a, &context_b);
         if is_valid_solution(&solution) {
@@ -199,7 +199,7 @@ fn compute_uniswap_in(out: U256, shortage: &TokenContext, excess: &TokenContext)
 /// Returns true if for each trade the executed price is not smaller than the limit price
 /// Thus we ensure that `buy_token_price / sell_token_price >= limit_buy_amount / limit_sell_amount`
 ///
-fn is_valid_solution<'a>(solution: &SinglePairSettlement) -> bool {
+fn is_valid_solution(solution: &SinglePairSettlement) -> bool {
     for trade in solution.trades.iter() {
         let order = trade.order;
         let buy_token_price = match solution.clearing_prices.get(&order.buy_token) {
--- a/solver/src/naive_solver/multi_order_solver.rs
+++ b/solver/src/naive_solver/multi_order_solver.rs
@@ -22,6 +22,43 @@ impl TokenContext {
                 * u256_to_bigint(&deficit.reserve)
                 * (u256_to_bigint(&self.sell_volume) - u256_to_bigint(&self.buy_volume))
     }
+
+    pub fn is_excess_before_fees(&self, deficit: &TokenContext) -> bool {
+        u256_to_bigint(&self.reserve)
+            * (u256_to_bigint(&deficit.sell_volume) - u256_to_bigint(&deficit.buy_volume))
+            < u256_to_bigint(&deficit.reserve)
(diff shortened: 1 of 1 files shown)


solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -22,18 +22,55 @@ impl TokenContext {
...
+            return solution;
+        } else {
+            // remove order with worst limit price that is selling excess token (to make it less excessive) and try again
+            let excess_token = if context_a.is_excess_before_fees(&context_b) {
+                context_a.address
+            } else {
+                context_b.address
+            };
+            orders.sort_by(price_comparator_for_selling_excess_token(excess_token));
+            orders.pop();
```
vkgnosis, line 52: Could do something like this so it's not sorted every time:
```rust
let comparator = price_comparator_for_selling_excess_token(excess_token);
let index = orders.iter().enumerate().max_by(|left, right| comparator(left.1, right.1).unwrap().0);
orders.swap_remove(index);
```
And we can add a `filter(order_sells_excess_token)` instead of treating orders that don't sell the excess token differently in the the comparator. With the filter we wouldn't `unwrap` in the snippet above and instead break out of the loop early if there is no longer any such order.

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -160,6 +195,54 @@ fn compute_uniswap_in(out: U256, shortage: &TokenContext, excess: &TokenContext)
...
+///
+/// Returns true if for each trade the executed price is not smaller than the limit price
+/// Thus we ensure that `buy_token_price / sell_token_price >= limit_buy_amount / limit_sell_amount`
+///
+fn is_valid_solution(solution: &SinglePairSettlement) -> bool {
+    for trade in solution.trades.iter() {
+        let order = trade.order;
+        let buy_token_price = match solution.clearing_prices.get(&order.buy_token) {
+            Some(price) => price,
+            None => return false,
```
vkgnosis, line 207: Shouldn't this panic because it should be guaranteed at this point that every token has a price?

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -160,6 +195,54 @@ fn compute_uniswap_in(out: U256, shortage: &TokenContext, excess: &TokenContext)
...
+        };
+        let sell_token_price = match solution.clearing_prices.get(&order.sell_token) {
+            Some(price) => price,
+            None => return false,
+        };
+
+        if order.sell_amount * buy_token_price < order.buy_amount * sell_token_price {
+            return false;
+        }
```
vkgnosis, line 216: In the smart contract this is
```rust
assert!(sell_amount * sell_price >= buy_amount * buy_price);
```
which seems different.

nlordell: Correct, the prices are inverted here.

The example I like to use to get the order right is:
> Beer costs 4€ and chocolate 1.50€
> So 3 beers are equivalent to 8 chocolates because and `3 * 4€ = 8 * 1.50€`

fleupold: Really good catch! Will adjust the unit test to also catch this.

solver/src/naive_solver/multi_order_solver.rs
fleupold: Very nice suggestion. This made the whole comparator easy enough to inline it.

vkgnosis (approved):

nlordell (approved):
Looks good!

PR #209, opened 2021-01-14
Fix deploy on tag

vkgnosis:
We haven't created any tag yet so we didn't observe this but we wouldn't
have extracted the correct version number.

### Test Plan
Once we create a tag this should work. Don't think we need to create one just to test.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -15,4 +15,4 @@ jobs:
       DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
     steps:
       - uses: actions/checkout@v2
-      - run: ./docker/deploy.sh ${GITHUB_REF#refs/heads/}
+      - run: docker/deploy.sh ${GITHUB_REF#refs/(tags|heads)/}


.github/workflows/deploy.yaml
```diff
@@ -15,4 +15,11 @@ jobs:
...
     steps:
       - uses: actions/checkout@v2
-      - run: ./docker/deploy.sh ${GITHUB_REF#refs/heads/}
+      - run: |
+          case $GITHUB_REF in
+            refs/heads/*)
+              version=${GITHUB_REF#refs/heads/};;
+            refs/tags/*)
+              version=${GITHUB_REF#refs/tags/};;
+          esac
```
nlordell, line 24: nit: This can be written more succinctly as `${GITHUB_REF#refs/(tags|heads)/}`:

```
$ GITHUB_REF=refs/tags/v10    
$ echo ${GITHUB_REF#refs/(tags|heads)/}
v10
$ GITHUB_REF=refs/heads/foo/bar
$ echo ${GITHUB_REF#refs/(tags|heads)/}
foo/bar
$ GITHUB_REF=refs/unknown/stuff           
$ echo ${GITHUB_REF#refs/(tags|heads)/}
refs/unknown/stuff
```

If you don't care that the second segment is `tags` or `heads`, you can also write this as `${GITHUB_REF#refs/*/}`
```
$ GITHUB_REF=refs/tags/v10    
$ echo ${GITHUB_REF#refs/*/}
v10
$ GITHUB_REF=refs/heads/foo/bar
$ echo ${GITHUB_REF#refs/*/}
foo/bar
$ GITHUB_REF=refs/unknown/stuff           
$ echo ${GITHUB_REF#refs/*/}
refs/unknown/stuff
```

nlordell (approved):

PR #211, opened 2021-01-14
Fix CI deploy

vkgnosis:
The previous syntax was a zsh speciality that isn't part of normal bash.


nlordell:
It might be the `bash` version. I would try `${TEST_ENV#refs/*/}` as well to see if it works.

nlordell (approved):

PR #212, opened 2021-01-14
Filter orders by owner, sell, buy in Postgres

vkgnosis:
### Test Plan
new test

### DIFF
--- a/orderbook/src/database/orders.rs
+++ b/orderbook/src/database/orders.rs
@@ -11,6 +11,15 @@ use model::{
 use primitive_types::H160;
 use std::convert::TryInto;
 
+/// Any default value means that this field is unfiltered.
+#[derive(Default)]
+pub struct OrderFilter<'a> {
+    pub min_valid_to: u32,
+    pub owner: Option<&'a H160>,
+    pub sell_token: Option<&'a H160>,
+    pub buy_token: Option<&'a H160>,
+}
+
 #[derive(sqlx::Type)]
 #[sqlx(rename = "OrderKind")]
 #[sqlx(rename_all = "lowercase")]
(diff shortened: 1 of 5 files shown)


orderbook/src/database/orders.rs
```diff
@@ -11,6 +11,15 @@ use model::{
 use primitive_types::H160;
 use std::convert::TryInto;
 
+/// Any default value means that this field is unfiltered.
+#[derive(Default)]
+pub struct OrderFilter<'a> {
+    pub min_valid_to: u32,
```
bh2smith, line 17: Why min valid to and not just valid to?

bh2smith (approved):
A lot to digest, but looks really good!

orderbook/src/database/orders.rs
vkgnosis: Because the check is `min_valid_to <= order.valid_to` so this makes it the clearest.

orderbook/src/database/orders.rs
```diff
@@ -77,13 +88,28 @@ impl Database {
...
         // pick the correct column based on order type.
+
+        fn as_bytes_or_default(h160: Option<&H160>) -> &[u8] {
+            h160.map(|h160| h160.as_bytes()).unwrap_or_default()
+        }
+
         sqlx::query_as(QUERY)
-            .bind(min_valid_to)
+            .bind(filter.min_valid_to)
+            .bind(filter.owner.is_none())
```
fleupold, line 107: nit: I think you should just be able to pass in the option directly and check e.g. for `$3 is NULL OR o.owner = $3` (https://docs.rs/sqlx/0.4.2/sqlx/macro.query.html#nullability-bind-parameters)

fleupold (approved):
Looks great!

orderbook/src/database/orders.rs
vkgnosis: Nice suggestion, implemented

PR #213, opened 2021-01-15
Include executed amounts in order meta data

vkgnosis:
The solver needs at least the filled amount (depending on order type)
and I figured this could be useful information in general so we might as
well put them all in the api.
On the other hand we could decide that this shouldn't be part of the
public api because we do not want to guarantee that we provide it and
find another way to get this information to the solver.
The type is BigUint and not U256 because every amount except the filled
amount can technically sum up to an amount larger than U256::MAX over
multiple trades. The filled amount cannot because it is stored as a U256
in the smart contract.

### Test Plan
Serialization unit test has been adjusted

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1318,10 +1318,12 @@ dependencies = [
  "hex",
  "hex-literal",
  "lazy_static",
+ "num-bigint",
  "primitive-types",
  "secp256k1",
  "serde",
  "serde_json",
+ "serde_with",
  "web3",
 ]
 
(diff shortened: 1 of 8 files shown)


fleupold (approved):
I think it's ok to offer this on the public API since it's required to judge about the current state of the orderbook (which is eventually something we might want the DAO to decide on based on running an API).

> The type is BigUint and not U256 because every amount except the filled amount can technically sum up to an amount larger than U256::MAX over multiple trades.

I don't understand this. Isn't the max_sell/buy amount the maximum amount that can be reached and that fits into U256? Or are you worried that while e.g. the max_sell_amount is U256, if split over multiple trades, its buy_amount could sum up to something larger than U256?

In the json output does Biguint become the same format as sellAmount (string) or does it get treated differently? It would be nice if the API didn't make a type difference between "amount" values as they are conceptually not really different.

vkgnosis:
> Or are you worried that while e.g. the max_sell_amount is U256, if split over multiple trades, its buy_amount could sum up to something larger than U256?

Exactly.

> In the json output does Biguint become the same format as sellAmount (string) or does it get treated differently? It would be nice if the API didn't make a type difference between "amount" values as they are conceptually not really different.

It is a string containing decimal digits (like other amounts) as I have documented in the openapi file. The reason I give it a new type name in openapi is to make it clear that this can be larger than a U256.

PR #214, opened 2021-01-15
Filter fully executed orders in postgres

vkgnosis:
While doing this I realized that there is another piece missing: handling order invalidated events. These make an order unexecutable without a trade so we need to track them too. Will happen in another PR.

### Test Plan
new unit tests


fleupold (approved):
Sorry, I missed this PR. Looks good to me. 

I feel we have a lot of "ignored" tests. Should some of these maybe still run on Github CI?

vkgnosis:
All of the `postgres_` ignored tests are run by CI.

PR #220, opened 2021-01-18
Filter for invalidated orders

vkgnosis:
Users can manually invalidate orders in the smart contract which we need
to handle in the database.

In future PRs: query and insert invalidation events into the db, allow this filter to be used in the http api

### Test Plan
New unit test

### DIFF
--- a/database/schema.sql
+++ b/database/schema.sql
@@ -35,6 +35,13 @@ CREATE TABLE trades (
     PRIMARY KEY (block_number, log_index)
 );
 
+CREATE TABLE invalidations (
+    block_number bigint NOT NULL,
+    log_index bigint NOT NULL,
+    order_uid bytea NOT NULL,
+    PRIMARY KEY (block_number, log_index)
+);
+
 -- Indexes for common operations that should be efficient.
 
 -- Get a specific user's orders.
@@ -45,3 +52,6 @@ CREATE INDEX order_valid_to ON orders USING BTREE (valid_to);
 
 -- Get all trades belonging to an order.
 CREATE INDEX trade_order_uid on trades USING BTREE (order_uid, block_number, log_index);
+
+-- Get all trades belonging to an order.
+CREATE INDEX invalidations_order_uid on invalidations USING BTREE (order_uid, block_number, log_index);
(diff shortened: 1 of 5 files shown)


model/src/order.rs
```diff
@@ -266,6 +266,7 @@ pub struct OrderMetaData {
     pub executed_sell_amount: BigUint,
     #[serde(with = "serde_with::rust::display_fromstr")]
     pub executed_fee_amount: BigUint,
+    pub invalidated: bool,
```
fleupold, line 269: Would it make sense to expose the invalidation time as an Option? Might be useful for debugging purposes to see when an order was invalidated.

database/schema.sql
```diff
@@ -35,6 +35,13 @@ CREATE TABLE trades (
     PRIMARY KEY (block_number, log_index)
 );
 
+CREATE TABLE invalidations (
```
fleupold, line 38: ubernit: In the future, we might want to support offchain-invalidations as well. Do you think those would be stored in the same database? Otherwise we can maybe already call this table invalidation_events or something else hinting at the fact that this is onchain?

fleupold (approved):
Looks good!

model/src/order.rs
vkgnosis: We don't exactly have the time, just the block number and I don't want to make that table more complicated for potential future debugging. I could see that we would later have some internal routes for getting more debug info.

database/schema.sql
vkgnosis: Offchain invalidation would be the delete orders route, right? In that case I thought we would delete the order directly from the database. I don't want to store those in this table because if it's not an event it has a different schema. I'll add a comment in the schema to explain what the table is.

fleupold: I think we will at some point need "time-travel" capabilities to derive the state of the orderbook at a certain point in time (maybe block number) so that we can reason about the "fairness" of solutions (and/or benchmark changes based on historic data). Therefore, I'd not remove anything from the table but just add at what point in time an order became invalid and maybe also store a timestamp with each event/block number in order to be able to join off-chain and on-chain events.

But it's not something we need to tackle right now.

Issue #40, opened 2020-11-10
Checking order validity


e00E:
For an order to be settleable the user must have enough balance and approved the contract for spending. This can change any time so we can never be sure that an order is valid before settling but we can still decide how often and in what way to check this.

One approach is to cache this information per user and token and update it in some intervals based by making more view function calls. Here we can batch those requests but eventually we hit some throughput limit of the node.

Another idea is to listen to erc20 events to create this information. This would make each individually query faster because we have the information locally. However getting the events is a lot of work similar to the event based orderbook but for every erc20 token we are interested in. It also assumes it is possible to recreate the balance of user based on events which might not be the case in some erc20 implementations. Or maybe events are not emitted at all and so on.

fleupold:
Maybe we can start off by just explicitly checking the balance in fixed time intervals (or whenever there is a `new_block` event). If node throughput becomes a bottleneck we could have a more sophisticated cache invalidation strategy that is based on internal calls (or events).

fleupold:
One service we should evaluate is https://www.blocknative.com/explorer They seem to allow monitoring wallets for any kind of interactions in the mempool

e00E:
A related question is whether the orderbook service should check order validity or whether only the solver should. The orderbook service would still check `validTo` so it can clear orders eventually but it could skip all the blockchain related checks.
Having the check in the orderbook service would allow us to not have the check in the solver and could make running multiple solvers easier since they can just trust the orderbook.
On the other hand this puts more work into the orderbook service that it conceptually does not need to do and when/if there are other order sources we most likely would want to check in the solver anyway.

fleupold:
I was also thinking about this. Eventually we want some solver independent (DAO governed) service to have consensus on the list of "valid" user orders as they might be protected by some envy freeness criterion. Therefore I slightly tend towards putting it in the orderbook service (not in the solver for that matter).

Another argument might be that clients are likely interested in querying an orderbook that is actually valid so the orderbook service might as well do that for all clients.

I'm also hoping that the filtering logic can be somewhat service agnostic (in that it operates on some `Orderbook` struct which is shared between the solver and API crate) and as such could be easily™️ moved around if need be.

fleupold:
I was playing around with blockscout a bit and I think we should be able to use their [WS subscription](https://docs.blocknative.com/websocket) with a filter like this:

```
[...]

PR #221, opened 2021-01-19
Store order invalidation events

vkgnosis:
The trades module has been renamed to events and it's methods have been changed to take an enum of trade event, invalidation event as we are handling more than just one event type now. Git shows this as deletion and new file because too many lines changed.
Likewise I renamed trade_events to event_updater.
Initially I thought it would be better to have separate functions for each event (replace events, insert new events etc) but after trying that I think this works better.

### Test Plan
adjusted event unit test


fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2F3o7abB06u9bNzA8lu8%2Fgiphy.gif&token=lgtm" data-gitmeme-token="lgtm"></a>

Have you already thought about how Approvals and Balances will interact with this? Would we try to capture these via events as well or would we "call" allowance/getBalance on the ERC20s? Would we want to persist these values? Would the ordering of events within a block matter?

vkgnosis:
There was some discussion about this in https://github.com/gnosis/gp-v2-services/issues/40 . My opinion is that we should call. But we should discuss more in that issue.

PR #222, opened 2021-01-19
Implement order book trait with postgres

vkgnosis:
as a replacement for the in memory order book which is being used at the
moment.

Before we switch we still need a database running in staging and I need to think about how to organize the warp route tests that currently require a functioning orderbook. I am tending towards making them e2e tests.

### Test Plan
the only visible logic change is moving the order filtering from the warp filter into the in memory order book. this is still being tested in the existing unit test.

### DIFF
--- a/orderbook/src/storage/postgresql.rs
+++ b/orderbook/src/storage/postgresql.rs
@@ -1,1 +1,50 @@
-// TODO: implement Storage for PostgreSQL database.
+use super::*;
+use crate::database::{Database, OrderFilter};
+use anyhow::Result;
+use contracts::GPv2Settlement;
+use futures::TryStreamExt;
+use model::{
+    order::{Order, OrderCreation, OrderUid},
+    DomainSeparator,
+};
+
+pub struct OrderBook {
+    domain_separator: DomainSeparator,
+    database: Database,
+}
+
+impl OrderBook {
+    pub fn _new(domain_separator: DomainSeparator, database: Database) -> Self {
+        Self {
+            domain_separator,
+            database,
+        }
+    }
+}
+
+#[async_trait::async_trait]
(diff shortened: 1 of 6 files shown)


orderbook/src/storage/postgresql.rs
```diff
@@ -1 +1,50 @@
...
+#[async_trait::async_trait]
+impl Storage for OrderBook {
+    async fn add_order(&self, order: OrderCreation) -> Result<AddOrderResult> {
+        if !has_future_valid_to(now_in_epoch_seconds(), &order) {
+            return Ok(AddOrderResult::PastValidTo);
+        }
+        let order = match Order::from_order_creation(order, &self.domain_separator) {
+            Some(order) => order,
+            None => return Ok(AddOrderResult::InvalidSignature),
+        };
```
fleupold, line 34: ubernit: this part is duplicated between the two implementations.

fleupold (approved):
> I need to think about how to organize the warp route tests that currently require a functioning orderbook. I am tending towards making them e2e tests.

I think some interactions them can become e2e tests (like adding and fetching an order, making sure it's not returned after it's invalid). However, I'd still like to have a bunch of unit tests for the routes that test error cases and other encoding related things. Can we maybe maintain a limited functionality in-memory storage that we can drop in for the tests?

What is preventing us from using a temporary database on CI other than the principle that unit tests shouldn't do I/O?

vkgnosis:
> What is preventing us from using a temporary database on CI other than the principle that unit tests shouldn't do I/O?

Nothing. We are already doing that. So we could make the route tests #ignore and run them like the other postgres tests on CI. But for local development it would be nice to not have to run postgres so I think I will go with mockall on the order book.

Issue #194, opened 2021-01-11
Separate order fetching and solver friendly order serialization into separate components


fleupold:
This task is a more long term thought on how to make the system scalable and allow easy addition of other onchain liquidity sources.

#176 introduced the HTTP based solver and fetches Uniswap orders directly in the solver code. I understand this was done for pragmatic reason. In the longer term I think the fetching of onchain liquidity sources should happen independent of the specific solver used to allow it to not necessarily touch any solver code.

E.g. we might want to add SushiSwap and Balancer pools sooner rather than later. The former would still be compatible with the slightly less naive solver whereas the latter would not. I think it would be nice to have an orthogonal "onchain liquidity fetching" chain that returns a list of abstract Order types. Different solvers can decide to handle or not handle specific order types (each order could provide a way of how it can be approximated with the most basic ordertype that all solvers support - limit or linear orders).

When adding a new liquidity source, we only need to add a new "fetcher" to the chain and provide a way how it can be approximated by limit orders. Additionally we can later add custom handling for more customized order types in each solver individually. This separates concerns of order fetching and solver friendly order serialization.

vkgnosis:
After seeing PR I understand this more. Commenting here instead of the PR for more high level stuff.

I agree with separating liquidity source fetching from the solver code is good. This gives the driver more control over how this is done and makes the data more easily sharable between solvers and also disentangles the solver from networking related code as it can work with the liquidity sources as data.

I am doubtful that we can represent liquidity sources to the solvers as limit orders without letting the solver know the exact details of the liquidity. Yes, everything that mechanically works like uniswap can be represented by the same data and we can do some things with traits there to reuse code (likewise for different RFQ systems) but I feel that is as far as we can go:

All liquidity have in common that they should only be used if they are useful for user orders. A liquidity source does not have tokens but it can be used to convert tokens from users. If liquidity sources were not distinguishable from user orders then solvers would not be able to work properly. For example it could try to match the order representing wrapping ETH with the order representing unwrapping ETH. Or it would use the wrapping order directly without giving it a user order as input.

A uniswap pool isn't equivalent to a limit order because it has dependencies between price and the pool size and maximum tradable amounts.

fleupold:
[...]

PR #223, opened 2021-01-19
[Refactoring 1] Introducing abstract liquidity object

fleupold:
Part of #194

*This is a proposal for how we can separate liquidity sources and solver handling logic to allow the two to grow more independently. Read in the context of the three PRs in this stack*

This PR defines an abstract Liquidity type, which can be interpreted by the solver(s). For now this datatype is shared between the two solvers (naive & http). If we see more solvers evolve in the future (with fundamentally different order types) it might make sense to have different liquidity types, one for each solver. This way, we could e.g. still model a Curve pool as limit orders for more naive solvers, but e.g. provide native support in the http_solver.

On the other hand, we have many liquidity sources (e.g. Offchain Orderbook, SCP RFQ system, Uniswap, SushiSwap. Balancer, Curve, Compound cToken converter, etc...). Each liquidity source yields a lot of abstract liquidity and provides a callback handler into how a specific "match" of this liquidity translates into trades and interactions.

As an example, if we have a user selling GNO for ETH, our source would provide a LimitOrder object selling GNO for WETH. In the settlement handler we would return a GNO/WETH trade **and** an interaction to unwrap the amount of WETH bought into ETH (we might build additional logic towards how to order/merge interactions in the end to avoid e.g. doing multiple unwrap calls).

Similarly an AMM liquidity source (e.g. Uniswap or Sushiswap) would yield an AMM liquidity object. In the settlement handler it could perform things such as approving the settlement contract if necessary.

An external market maker's RFQ system can be implemented as a liquidity source with a list of partially matchable limit orders that get matched via a 0x bridge (opposed to native GPv2 EOA orders).

PredictionMarket outcome tokens or other "split and mergeable" tokens (e.g. LP tokens) would become a new liquidity subtype (once the solvers support it) with potentially many sources.

### Test Plan
Unit tests at the end of the PR stack

### DIFF
--- a/solver/src/liquidity.rs
+++ b/solver/src/liquidity.rs
@@ -1,4 +1,4 @@
-use model::order::OrderKind;
+use model::{order::OrderKind, TokenPair};
 use num_rational::Rational;
 use primitive_types::{H160, U256};
 use settlement::{Interaction, Trade};
@@ -9,16 +9,12 @@ use mockall::automock;
 
 use crate::settlement;
 
-///
 /// Defines the different types of liquidity our solvers support
-///
 pub enum Liquidity {
     Limit(LimitOrder),
     Amm(AmmOrder),
 }
-///
 /// Basic limit sell and buy orders
-///
 #[derive(Clone)]
 pub struct LimitOrder {
     pub sell_token: H160,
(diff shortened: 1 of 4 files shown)


solver/src/liquidity.rs
```diff
@@ -0,0 +1,75 @@
...
+///
+#[async_trait::async_trait]
+pub trait LiquiditySource {
+    ///
+    /// Given all liquidity sourced so far, returns the additional liquidity that can be sourced from
+    /// the concrete implementation.
+    ///
+    async fn get_liquidity(
+        &self,
+        liquidity_so_far: impl Iterator<Item = &Liquidity> + Send + Sync + 'async_trait,
```
vkgnosis, line 32: Shouldn't need Sync

solver/src/liquidity.rs
```diff
@@ -0,0 +1,75 @@
...
+use std::sync::Arc;
+
+#[cfg(test)]
+use mockall::automock;
+
+use crate::settlement;
+
+///
+/// Defines the different types of liquidity our solvers support
+///
```
vkgnosis, line 15: Why surround doc comment with empty lines? Not important but I prefer just having the comment.

solver/src/liquidity.rs
```diff
@@ -0,0 +1,75 @@
...
+pub trait LimitOrderSettlementHandling: Send + Sync {
+    fn settle(&self, executed_amount: U256) -> (Option<Trade>, Vec<Box<dyn Interaction>>);
+}
+
+///
+/// 2 sided constant product automated market maker with equal reserve value and a trading fee (e.g. Uniswap, Sushiswap)
+///
+#[derive(Clone)]
+pub struct AmmOrder {
+    pub tokens: (H160, H160),
```
vkgnosis, line 63: Should be TokenPair

josojo:
Cool. Yes, your suggestion makes a ton of sense to me. Like it.

Nit:
>If we see more solvers evolve in the future (with fundamentally different order types) it might make sense to have different liquidity types, one for each solver.

Since we don't wanna pull and store the information of one liquidity source several times for each solver, I would introduce another trait: `representLiquidityForSolver`, that would transform one Liquidity object of the Liquidity-Enum and destruct it into a good representation for the solver. For each liquidity, we can have several `representLiquidityForSolver` traits implemented.

E.g:
```
Liquidity: Curve - AMM
```
```
Implementation 1: representLiquidityForSolver -> destructs the AMM into linear orders
```
```
Implementation 1: representLiquidityForSolver -> just takes the reserve values and sends it to the solver.
```

But I think the introduction of this trait can also happen later.

josojo (approved):

fleupold:
I addressed the comments and following up on the discussion in #194 I removed the LiquiditySource trait but am leaving the liquidity type as an enum for now.

vkgnosis (approved):

PR #224, opened 2021-01-19
[Refactoring 2] Introducing concrete liquidity implementations

fleupold:
Part of #194

This PR implements the two liquidity sources we support at the moment (offchain orderbook and uniswap) in the new architecture.

In the future I'd expect additional liquidity sources (e.g. SCP's RFQ system) to be separate implementations of the LiquiditySource trait.

### Test Plan
Unit tests in next PR

### DIFF
--- a/solver/src/settlement.rs
+++ b/solver/src/settlement.rs
@@ -36,7 +36,7 @@ impl Trade {
     }
 }
 
-pub trait Interaction: std::fmt::Debug + Send + Sync {
+pub trait Interaction: std::fmt::Debug + Send {
     // TODO: not sure if this should return a result.
     // Write::write returns a result but we know we write to a vector in memory so we know it will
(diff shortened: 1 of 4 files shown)


josojo (approved):
Noice!

solver/src/settlement.rs
```diff
@@ -26,9 +26,17 @@ impl Trade {
...
+        Self {
+            order,
+            executed_amount,
+            fee_discount: 0,
+        }
+    }
 }
 
-pub trait Interaction: std::fmt::Debug {
+pub trait Interaction: std::fmt::Debug + Send + Sync {
```
vkgnosis, line 39: Is that Sync needed? No big deal if it is just looks like it wouldn't.

vkgnosis (approved):

PR #226, opened 2021-01-21
Refactor warp test so that it does not need order book

vkgnosis:
We accomplish this by splitting the filter into three parts:
1. convert request into the arguments for order book method
2. convert result of order book method into a response
3. combine the other parts trivially together

Now we can test 1. 2. individually, neither needing access to an order
book.

This has the advantage that we neither need to mock, nor run a real postgres instance.

I would like to make similar changes to the other filters if this gets positively reviewed. 

### Test Plan
Converted existing tests

### DIFF
--- a/orderbook/src/api.rs
+++ b/orderbook/src/api.rs
@@ -2,12 +2,16 @@ mod filter;
 mod handler;
 
 use crate::storage::Storage;
+use serde::Serialize;
 use std::sync::Arc;
-use warp::Filter;
+use warp::{
+    reply::{json, Json},
+    Filter, Reply,
+};
 
 pub fn handle_all_routes(
     orderbook: Arc<dyn Storage>,
-) -> impl Filter<Extract = (impl warp::Reply,), Error = warp::Rejection> + Clone {
+) -> impl Filter<Extract = (impl Reply,), Error = warp::Rejection> + Clone {
     let order_creation = filter::create_order(orderbook.clone());
     let order_getter = filter::get_orders(orderbook.clone());
     let fee_info = filter::get_fee_info();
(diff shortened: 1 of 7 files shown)


orderbook/src/api.rs
```diff
@@ -19,3 +23,24 @@ pub fn handle_all_routes(
             .or(order_by_uid),
     )
 }
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct Error<'a> {
```
vkgnosis, line 29: This has been moved to here from handler.rs  because both filter.rs and handler.rs need it.

orderbook/src/database/orders.rs
```diff
@@ -13,14 +13,14 @@ use std::convert::TryInto;
 
 /// Any default value means that this field is unfiltered.
 #[derive(Default)]
-pub struct OrderFilter<'a> {
+pub struct OrderFilter {
```
vkgnosis, line 16: While working on the PR I realized that forcing this struct to contain references makes it annoying to use (because we construct the filter in warp) so I changed it to owned values.

fleupold: It's a bit unfortunate that this change causes a lot of touched lines that are unrelated to the test refactoring. Maybe next time it can be broken up into a separate PR?

fleupold (approved):
Looks good! As far as I can tell the real change is quite straight-forward but hidden in a lot of touched lines.

orderbook/src/database/orders.rs
vkgnosis: Sure it could. I thought this was small enough and it being motivated by the other change that it wasn't worth it to make a separate PR but I'll calibrate my threshold for this down.

PR #227, opened 2021-01-21
Enable tokio threaded runtime

### LINKED ISSUE #15: Separate order collection and solver part into different binaries
https://github.com/gnosis/oba-services/pull/9#pullrequestreview-498146729:
> Midterm, we should probably also separate the warp and the solving part of the system into different binaries, so that they can run on different hosts with different availability guarantees.

vkgnosis:
Based on Cargo.lock not changing this feature was transitively enabled
anyway but we should still explicitly specify that we want the threaded
runtime.

Closes #15 

### Test Plan
CI


fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2F3o751Yxe9UjX26BZbG%2Fgiphy.gif&token=ok" data-gitmeme-token="ok"></a>

PR #229, opened 2021-01-22
Convert get_fee_info for testability

vkgnosis:
### Test Plan
adapted unit test

### DIFF
--- a/orderbook/src/api/filter.rs
+++ b/orderbook/src/api/filter.rs
@@ -1,13 +1,15 @@
 use super::handler;
 use crate::{database::OrderFilter, storage::Storage};
 use anyhow::Result;
+use chrono::{DateTime, FixedOffset, Utc};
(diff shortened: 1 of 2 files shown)


orderbook/src/api/filter.rs
```diff
@@ -235,20 +262,21 @@ pub mod test_util {
...
-                .path(&path_string)
-                .method("GET")
-                .reply(&filter)
-                .await
-        };
-        let response = post().await;
-        let body: handler::FeeInfo = serde_json::from_slice(response.body()).unwrap();
+    async fn get_fee_info_request_ok() {
+        let filter = get_fee_info_request();
+        let token = String::from("0x0000000000000000000000000000000000000000");
```
fleupold, line 267: nanonit: I found the test-case slightly nicer when it was not testing the Default value (which might be returned for other reasons).

fleupold (approved):

PR #232, opened 2021-01-22
Adding an orderbook component between API handler and storage

fleupold:
Part of #40 

This PR adds another layer in between API handler and storage (happy for better naming suggestion) which for now simply combines the common logic between in-memory and postgres storage when it comes to adding orders.

In the next PR I'm planning to also make it aware of the AccountBalanceFetching component. It will then create the appropriate subscriptions to watch allowance/balance changes as well as filter orders for which there is not enough balance.

### Test Plan
CI

### DIFF
--- a/orderbook/src/storage/memory.rs
+++ b/orderbook/src/storage/memory.rs
@@ -98,6 +98,10 @@ impl Storage for OrderBook {
                         .buy_token
                         .map(|token| token == order.order_creation.buy_token)
                         .unwrap_or(true)
+                    && filter
+                        .uid
+                        .map(|uid| uid == order.order_meta_data.uid)
+                        .unwrap_or(true)
             })
             .cloned()
             .collect())
@@ -117,6 +121,7 @@ impl Storage for OrderBook {
         let remove_order_future = self.remove_expired_orders(now_in_epoch_seconds());
         let remove_settled_orders_future = self.remove_settled_orders(settlement_contract);
         futures::join!(remove_order_future, remove_settled_orders_future);
+        println!("maintained");
         Ok(())
     }
 }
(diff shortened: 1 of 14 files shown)


vkgnosis:
Should this be another layer in between (wrapping other stuff)? The way I imagined it was that we have Postgres for the db part, AccountBalanceFetching for the on chain part and then `struct OrderBook` has both as members and uses them together in its methods. Also related is that I would like to get rid of the the Storage trait again now that I have changed warp tests to not need a real order book.

fleupold:
I am envisioning the struct that is created in this PR to contain the DB and AccountBalanceFetching as members (the next PR will be introduction of AccountBalanceFetching and afterwards using it inside this struct). 
AccountBalanceFetching needs the user of a created order, which is why I moved the OrderCreation -> FullOrder logic into this component.

An alternative approach would be to store all balances inside the database as well and use SQL to query and filter. The advantage would be that at some point we likely also have to persist historic balances (so that the DAO can get a hollistic view of the orderbook at some point in the past), but I'm not a fan of further complicating the SQL logic for querying orders (since balances might update frequently it might be nice to just keep the latests balances and memory and write a history log to disk).

orderbook/src/storage/memory.rs
```diff
@@ -115,6 +98,10 @@ impl Storage for OrderBook {
                         .buy_token
                         .map(|token| token == order.order_creation.buy_token)
                         .unwrap_or(true)
+                    && filter
+                        .uid
+                        .map(|uid| uid == order.order_meta_data.uid)
+                        .unwrap_or(true)
```
fleupold, line 104: This seems like a bug that existed before but is only now caught by the e2e test.

vkgnosis (approved):

PR #242, opened 2021-01-26
Use gas price estimation

vkgnosis:
like in gpv1 we have several estimators that are selected in the command
line argument.

Does not yet retry with higher gas prices, just uses the gas price for the one transaction.

`gas_price_estimation.rs` has mostly been copied from gpv1. While adapting it I noticed that we do not yet have an http client abstraction and don't define a crate wide `web3` type that uses that http client so I resused the exact types we already had. Will add those in a separate PR.

Note to self: Should update staging so that as in gpv1 we use better gas estimators than just web3.

### Test Plan
e2e test still works

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -1,7 +1,12 @@
-use crate::liquidity::{uniswap::UniswapLiquidity, Liquidity};
-use crate::{orderbook::OrderBookApi, solver::Solver};
-use anyhow::{anyhow, Context, Result};
+use crate::{
+    liquidity::{uniswap::UniswapLiquidity, Liquidity},
+    orderbook::OrderBookApi,
+    settlement_submission,
+    solver::Solver,
+};
+use anyhow::{Context, Result};
 use contracts::GPv2Settlement;
+use gas_estimation::GasPriceEstimating;
 use std::time::Duration;
 use tracing::info;
 
@@ -12,6 +17,7 @@ pub struct Driver {
     orderbook: OrderBookApi,
     uniswap_liquidity: UniswapLiquidity,
     solver: Box<dyn Solver>,
+    gas_price_estimator: Box<dyn GasPriceEstimating>,
 }
 
 impl Driver {
@@ -20,12 +26,14 @@ impl Driver {
         uniswap_liquidity: UniswapLiquidity,
         orderbook: OrderBookApi,
         solver: Box<dyn Solver>,
+        gas_price_estimator: Box<dyn GasPriceEstimating>,
     ) -> Self {
         Self {
             settlement_contract,
-            solver,
             orderbook,
             uniswap_liquidity,
+            solver,
+            gas_price_estimator,
         }
     }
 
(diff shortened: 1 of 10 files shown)


solver/src/driver.rs
```diff
@@ -53,30 +57,13 @@ impl Driver {
         };
         info!("Computed {:?}", settlement);
         // TODO: check if we need to approve spending to uniswap
-        // TODO: use retry transaction sending crate for updating gas prices
```
vkgnosis, line 56: moved to  settlement_submission.rs

solver/src/gas_price_estimation.rs
```diff
@@ -0,0 +1,72 @@
...
+        GnosisSafe,
+        Web3,
+    }
+}
+
+#[derive(Clone)]
+struct Client(reqwest::Client);
+
+#[async_trait::async_trait]
+impl Transport for Client {
```
vkgnosis, line 24: Can't implement foreign trait for foreign type so we need a wrapper type.

solver/src/settlement_submission.rs
```diff
@@ -0,0 +1,50 @@
...
+                Vec::new(),
+            )
+            .gas(MAX_GAS.into())
+    };
+    tracing::info!(
+        "Settlement call: {}",
+        hex::encode(settle().tx.data.expect("data").0),
+    );
+    let gas_price = gas
+        .estimate_with_limits(MAX_GAS.into(), Duration::from_secs(60))
```
fleupold, line 40: Can we make the 60s configurable? I would think a reasonable default should be shorter than that (e.g. 30s)

fleupold (approved):
Very nice! Glad we are part of the small slice:

image

solver/src/gas_price_estimation.rs
```diff
@@ -0,0 +1,72 @@
...
+    EthGasStation, GasNowGasStation, GasPriceEstimating, GnosisSafeGasStation,
+    PriorityGasPriceEstimating, Transport,
+};
+use serde::de::DeserializeOwned;
+use structopt::clap::arg_enum;
+use web3::Web3;
+
+arg_enum! {
+    #[derive(Debug)]
+    pub enum GasEstimatorType {
```
bh2smith, line 12: Wow! So many estimation services! I love it!

solver/src/main.rs
```diff
@@ -25,9 +26,21 @@ struct Arguments {
...
+    /// `Web3`: supports every network.
+    #[structopt(
+        long,
+        env = "GAS_ESTIMATORS",
+        default_value = "Web3",
+        possible_values = &GasEstimatorType::variants(),
+        case_insensitive = true,
+        use_delimiter = true
+    )]
+    gas_estimators: Vec<GasEstimatorType>,
```
bh2smith, line 43: Is there any way to have mixed gas price averaging or is this a silly idea?

solver/src/settlement_submission.rs
```diff
@@ -0,0 +1,50 @@
+use std::time::Duration;
+
+use crate::settlement::Settlement;
+use anyhow::{anyhow, Context, Result};
+use contracts::GPv2Settlement;
+use ethcontract::GasPrice;
+use gas_estimation::GasPriceEstimating;
+use primitive_types::U256;
+
+const MAX_GAS: u32 = 8_000_000;
```
bh2smith, line 10: This is a v1 thing right? I guess it applies anywhere.

bh2smith (approved):
Awesome!

solver/src/settlement_submission.rs
vkgnosis: It was a constant in the gpv2 code somewhere already.

solver/src/main.rs
vkgnosis: On a technical level it would be easy to average sources together but I don't think we need a complex set up like that. The reason for having several sources is that some might be down so it can use the next one in the list.

PR #243, opened 2021-01-27
Introduce BalanceFetcher

fleupold:
Part of #40

This component defines an interface for balance/allowance fetching and a concrete implementation that registers `(account, token)` pairs by fetching their balance and allowance for the Allowance manager contract. 
Updates can be triggered periodically, which refetch all subscribed balances and allowances using the new Batch Call feature of ethcontracts (will likely happen inside orderbook maintenance).

In the future we probably also want a way to unsubscribe token pairs that are no longer part of the active orderbook. 

### Test Plan
Unfortunately testing ethcontract related code at the moment is painful. I added an e2e test that require ganache, but would much prefer if we could make it easier to test these type of interactions as a unit test. There is https://github.com/gnosis/ethcontract-rs/issues/424 to track part of this effort. Maybe for now the easiest solution would be to expose [TestTransport](https://github.com/gnosis/ethcontract-rs/blob/5c677d10bfa37da073c1c80bb4f3c562bb2dc9c8/ethcontract/src/test/transport.rs#L24) behind a feature so that we can use it to mock web3 responses.

### DIFF
--- a/orderbook/src/account_balances.rs
+++ b/orderbook/src/account_balances.rs
@@ -7,14 +7,18 @@ use std::sync::Mutex;
 use contracts::IERC20;
 use primitive_types::{H160, U256};
 
+#[cfg_attr(test, mockall::automock)]
 #[async_trait::async_trait]
 pub trait BalanceFetching: Send + Sync {
     // Register owner and address for balance updates in the background
     async fn register(&self, owner: H160, token: H160);
 
-    // Returns the current balance available to the allowance manager for the given owner and token.
-    // Should be non-blocking. Returns 0 if balance is not available for whatever reason.
-    fn get_balance(&self, owner: H160, token: H160) -> U256;
+    // Register multiple owner and addresses for balance updates in the background
+    async fn register_many(&self, owner_token_list: Vec<(H160, H160)>);
+
+    // Returns the latest balance available to the allowance manager for the given owner and token.
+    // Should be non-blocking. Returns None if balance has never been fetched.
+    fn get_balance(&self, owner: H160, token: H160) -> Option<U256>;
 }
 
 pub struct Web3BalanceFetcher {
@@ -32,8 +36,8 @@ struct SubscriptionKey {
 
 #[derive(Clone, Default)]
 struct SubscriptionValue {
-    balance: U256,
-    allowance: U256,
+    balance: Option<U256>,
+    allowance: Option<U256>,
 }
 
 impl Web3BalanceFetcher {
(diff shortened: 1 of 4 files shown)


orderbook/src/account_balances.rs
```diff
@@ -0,0 +1,195 @@
...
+                        .view()
+                        .batch_call(&mut batch),
+                    instance
+                        .allowance(subscription.owner, self.allowance_manager)
+                        .view()
+                        .batch_call(&mut batch),
+                )
+                .then(move |(balance, allowance)| {
+                    self.update_with(subscription, balance, allowance);
+                    ok::<_, MethodError>(())
```
vkgnosis, line 76: This means the mutex is locked for every individual request. We already get all the responses at the same time into a vector so it would be better to only lock the mutex once to apply all of them.

orderbook/src/account_balances.rs
```diff
@@ -0,0 +1,195 @@
...
+        let _ = self.register_many(subscriptions.iter()).await;
+    }
+
+    async fn register_many(
+        &self,
+        subscriptions: impl Iterator<Item = &SubscriptionKey>,
+    ) -> Result<()> {
+        let mut batch = CallBatch::new(self.web3.transport());
+        let calls = subscriptions
+            .map(|subscription| {
```
vkgnosis, line 62: Eventually we will want to limit the number of requests per batch.

orderbook/src/account_balances.rs
```diff
@@ -0,0 +1,195 @@
...
+use ethcontract::{batch::CallBatch, errors::MethodError, Http, Web3};
+use futures::future::{join, join_all, ok, FutureExt as _};
+use std::collections::HashMap;
+use std::sync::Mutex;
+
+use contracts::IERC20;
+use primitive_types::{H160, U256};
+
+#[async_trait::async_trait]
+pub trait BalanceFetching: Send + Sync {
```
vkgnosis, line 11: Why a trait? Are we expecting to mock it or have alternative implementations? Don't have a problem with that just feel that if the reason is maybe having an alternative implementation in the future then it would be better to only create the trait then.

vkgnosis, line 11: Shouldn't the update function which is pub on the struct be part of the trait too?

vkgnosis (approved):
All of things I pointed out could be changed in future PRs too if you prefer.

fleupold (commented):
I also made register_many part of the public interface after noticing it's needed by the fetching component.

PR #246, opened 2021-01-29
Use balance fetcher

fleupold:
Part of #40 using the component created in #243.

We register tokens whenever an order is added and check balances when fetching orders. There, we also need to potentially create subscriptions (in case the order is coming from disk after a cold start).

If the cache is becoming too big we could potentially create a cleanup job/LRU strategy or the like to reduce the number of active subscriptions.

In order to pass the balance information to the solver and frontend, I created a new field on metadata. In the next PR I'm planning to add a filter to allow only fetching orders with valid balances.

### Test Plan
Added unit tests.

### DIFF
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -1,6 +1,10 @@
 //! Contains the order type as described by the specification with serialization as described by the openapi documentation.
 
-use crate::{h160_hexadecimal, u256_decimal, DomainSeparator, Signature, TokenPair};
+use crate::{
+    h160_hexadecimal,
+    u256_decimal::{self, DecimalU256},
+    DomainSeparator, Signature, TokenPair,
+};
 use chrono::{offset::Utc, DateTime, NaiveDateTime};
 use hex_literal::hex;
 use num_bigint::BigUint;
@@ -8,6 +12,7 @@ use primitive_types::{H160, U256};
 use secp256k1::key::ONE_KEY;
 use serde::{de, Deserialize, Serialize};
 use serde::{Deserializer, Serializer};
+use serde_with::serde_as;
 use std::fmt::{self, Display};
 use std::str::FromStr;
 use web3::{
@@ -118,6 +123,7 @@ impl OrderBuilder {
 }
 
 /// An order as provided to the orderbook by the frontend.
+#[serde_as]
 #[derive(Eq, PartialEq, Clone, Copy, Debug, Deserialize, Serialize, Hash)]
 #[serde(rename_all = "camelCase")]
 pub struct OrderCreation {
@@ -253,6 +259,7 @@ impl OrderCreation {
 }
 
 /// An order as provided to the orderbook by the frontend.
+#[serde_as]
 #[derive(Eq, PartialEq, Clone, Debug, Deserialize, Serialize, Hash)]
 #[serde(rename_all = "camelCase")]
 pub struct OrderMetaData {
@@ -260,7 +267,7 @@ pub struct OrderMetaData {
     #[serde(with = "h160_hexadecimal")]
     pub owner: H160,
     pub uid: OrderUid,
-    // Minimum of balance and allowance for GPv2. None if information is not available
+    #[serde_as(as = "Option<DecimalU256>")]
     pub available_balance: Option<U256>,
     #[serde(with = "serde_with::rust::display_fromstr")]
     pub executed_buy_amount: BigUint,
@@ -390,7 +397,7 @@ mod tests {
             "creationDate": "1970-01-01T00:00:03Z",
             "owner": "0x0000000000000000000000000000000000000001",
             "uid": "0x1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111",
-            "availableBalance": None::<U256>,
+            "availableBalance": "100",
             "executedBuyAmount": "3",
             "executedSellAmount": "4",
             "executedFeeAmount": "5",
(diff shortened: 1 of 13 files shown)


orderbook/src/orderbook.rs
```diff
@@ -39,11 +49,42 @@ impl Orderbook {
     }
 
     pub async fn get_orders(&self, filter: &OrderFilter) -> Result<Vec<Order>> {
-        self.storage.get_orders(filter).await
+        let orders_without_balance = self.storage.get_orders(filter).await?;
```
vkgnosis, line 52: Both of the steps in this function could be done more efficiently by moving some functionality into the balance fetcher.
Currently we do:
1. for every order: get_balance, if none put into vec
2. register_many
3. for every order: get_balance

Instead if there was a function like `make_sure_subscription_exits_and_create_and_update_if_not(impl Iteratior<(H160, H160)>` we could avoid locking the mutex for every order in the first step.
Or taking it a step further there could be a function `update_orders_with_balances_and_subscribe(&mut [Order])` that combines everything.
I know that specializes the balance fetcher a lot but I think it's worth it.

This can happen in a future PR and I don't mind implementing it myself. Let me know if I should.

orderbook/src/orderbook.rs
```diff
@@ -39,11 +49,42 @@ impl Orderbook {
...
+        // Enrich orders with balance information
+        Ok(orders_without_balance
+            .into_iter()
+            .map(|mut order| {
+                order.order_meta_data.available_balance = self
+                    .balance_fetcher
+                    .get_balance(order.order_meta_data.owner, order.order_creation.sell_token);
+                order
+            })
+            .collect())
```
vkgnosis, line 79: I would do this in a non functional manner and update the orders in place instead of creating a new vector
```rust
for order in orders_without_balance.iter_mut() {
    order.order_meta_data.available_balance = ...
}

**Review by vkgnosis — APPROVED** [review]

File: model/src/order.rs [hunk 3]
```diff
@@ -260,6 +260,8 @@ pub struct OrderMetaData {
     #[serde(with = "h160_hexadecimal")]
     pub owner: H160,
     pub uid: OrderUid,
+    // Minimum of balance and allowance for GPv2. None if information is not available
+    pub available_balance: Option<U256>,
```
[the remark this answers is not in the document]
fleupold: Good catch. It seems to be working with `#[serde_as(as = "Option<DisplayFromStr>")]` but only for serialization (deserialization seems to be interpret it radix encoding). Technically we only need serialization, but it seems too hacky to rely on that.

Will implement a wrapper type...

orderbook/src/orderbook.rs
fleupold: I'm personally not really worried that acquiring a mutex will be the bottleneck in this operation. I agree it's nicer to write performant code (e.g. in the case of the balance fetcher itself where the change was a detail of them implementation), but here I feel it's a tradeoff against making a component very specialised for one use case.

Even if we are not going to use this component elsewhere, I personally prefer a simpler interface (here I can register a bunch of subscriptions, potentially unregister them, and query a single balance in a non-blocking form) over a highly specialised one. 
My somewhat subjective argument is that it's cognitively easier to work with/maintain such components in the long run (when we have many different components).

orderbook/src/orderbook.rs
fleupold: In the next PR I'm using the iterator to also filter orders that don't have enough balance. Is there a syntactically nice way to do it non-functionally?

model/src/order.rs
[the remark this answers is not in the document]
vkgnosis: We want deserialization too because we use this type to decode the response for the order book in the driver.

orderbook/src/orderbook.rs
vkgnosis: Maybe Vec::retain

orderbook/src/orderbook.rs
vkgnosis: My pov is that there is a specific operation that we want to accomplish here: for every order in a vector we want to set the balance which either involves getting it from the cache or from the node.
Based on the specific design of the cache there is a most efficient way to do this which is to calculate the set of uncached balances, fetch them into the cache, update all orders from cache. Doing it this way locks the mutex twice no matter how many orders there are.
But now we are artificially limiting us to a much less efficient implementation because of the way we picked the abstraction boundary interface as it only supports getting the balance of single orders (and register_many). "much less efficient" because for every order we lock the mutex twice and there can be a lot of orders.
To me this looks like a problem with the interface. Why wouldn't it support the exact operation we have created it for? We have generalized prematurely. I agree that it is a more complex interface but the total complexity of the component is still low.
We have this operation that updates the balances in a Vec either way. It can either be here and inefficient or directly in the component and efficient.

fleupold: Maybe my understanding is incorrect, but according to [this benchmark](https://github.com/matklad/lock-bench) acquiring a mutex in rust takes 1.5 nanoseconds in light contention (still less than 10 ns in heavy contention). 

So maybe this question comes down to “premature generalisation” vs “premature optimisation”. Personally, I find generalisation still valuable even if the component is not reused since it allows the maintainer to think in more common/generic concepts (similar to why people tend to use design patterns over and over - others have seen similar components before and thus understand them more quickly).

I originally designed `get_balance` to be non-blocking in order to ensure the read path is fast (as in _"no network request has to be made"_ fast, thus thinking in milliseconds). However, seeing that orders can come from disk and might not yet be registered this might have been unreasonable.
Therefore, I’d be fine to change the interface to a single `async get_balances(impl Iter<(Address, Token)>)` (maybe even with the fact that it creates subscriptions being implicit unless we actively unsubscribe from the orderbook). This way it would still be simple yet more natural with the underlying implementation. Let's do this in a separate PR though.

PR #249, opened 2021-01-29
Create `shared` crate for other funtionality shared between binaries

### LINKED ISSUE #117: Do we want one or multiple crates for shared functionality between orderbook and solver?
We have the crates model, shared-arguments, tracing-setup but we could also have one crate that has those as modules like "services-core" in v1-services.
Having them as modules is a little easier to manage because there is only one dependency and creates less top level folders.
Having them as crates allows more fine grained reusability in case there is ever a third binary that only needs subset of functionality. But then we could still have features for individual modules or not care that we are importing more than we need.

vkgnosis:
closes #117

### Test Plan
CI, no logic change


fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FkigfYxdEa5s1ziA2h1%2Fgiphy.gif&token=like" data-gitmeme-token="like"></a>

PR #254, opened 2021-02-01
Remove old naive solver

vkgnosis:
This code is no longer being used.

### Test Plan
CI


fleupold (approved):
Thanks for the cleanup!

PR #255, opened 2021-02-01
Connect to postgres when url is supplied as parameter

vkgnosis:
We now have a database in staging. To test that it works we need this PR. If it does work I will make another PR that removes  the in memory order book.

Edit: I am testing this now by manually running this branch on kubernets so maybe we don't have to merge this.\
Edit: Turns out compiling on one of the existing pods is extremely slow so will probably have to merge this after all.

### Test Plan
CI


bh2smith (approved):
Sweet. Nice simple connection test.

orderbook/src/main.rs
```diff
@@ -54,8 +61,15 @@ async fn main() {
         .expect("Deployed contract constants don't match the ones in this binary");
     let domain_separator =
         DomainSeparator::get_domain_separator(chain_id, settlement_contract.address());
-    let storage = InMemoryOrderBook::default();
-    let orderbook = Arc::new(Orderbook::new(domain_separator, Box::new(storage)));
+    let storage: Box<dyn Storage> = match args.db_url {
```
fleupold, line 64: nanonit: maybe log what kind of database we are using for convenience.

fleupold (approved):
Nice

PR #257, opened 2021-02-02
Remove in memory orderbook

vkgnosis:
Staging is working with postgres now so the in memory order book can be
removed.

I would also like to remove the whole storage module in the next PR, moving those functions directly to `struct OrderBook`.

### Test Plan
e2e test works on CI

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,12 +33,41 @@ Several pieces of functionality are shared between the order book and the solver
 
 ## Testing
 
-Run unit tests with `cargo test`.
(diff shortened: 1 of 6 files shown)


fleupold (approved):
It took me a while to figure out how to run the e2e tests now. Can you add to the readme your setup of how you run postgres locally?

My setup is:

```bash
docker run -d -e POSTGRES_HOST_AUTH_METHOD=trust -e POSTGRES_USER=`whoami` -p 5432:5432 postgres

psql -f database/schema.sql

cargo test --locked --package e2e
```

But maybe there is a better one?

PR #259, opened 2021-02-02
Fetch Uniswap liquidity via batched call

fleupold:
This PR refactors the way we currently fetch Uniswap liquidity sources from using individual requests to a single RPC batch call. This is more efficient and will be needed if we want to include intermittent pools via the native token (GNO<>OWL should also add GNO<>ETH, ETH<>OWL)

On the way I noticed that the pool struct is no longer really necessary (it's kind of 1:1 replaced by the AMM liquidity subtype) so I removed it to simplify the logic in `solver/src/liquidity/uniswap.rs`. I also changed the way we compute the Uniswap pair addresses from using a RPC call to actually implementing the deterministic `CREATE2` address derivation in Rust. This saves us a dependent RPC call (the result is needed to create the `get_reserves` call). This reduces the amount of batches we need from 2 to 1.

### Test Plan
Unit test for create2 
E2E test to make sure we are still fetching the uniswap liquidity sources correctly.


vkgnosis (approved):

fleupold:
Turns out honeyswap seems to be using a slightly different imlementation of the token pair, thus their init hash is different (3f88503e8580ab941773b59034fb4b2a63e86dbc031b3633a925533ad3ed2b93). Not sure if this should be seen as a fundamentally different liquidity source (like sushi swap) or if we should special case it inside uniswap.

PR #260, opened 2021-02-02
Include native token pools as an intermediate hop when fetching Unisw…

fleupold:
This PR implements a TODO in the code "intermediate hops via ETH" (aka the native token).  I added the native token contract to the codebase (will likely be needed anyways when we want to wrap/unwrap to support ETH as a buy token).

### Test Plan
Run against xDAI, trying to trade HNY for WETH. See that we are fetching 3 pools (HNY<>WETH, HNY<>WXDAI, WXDAI<>WETH)


solver/src/liquidity/uniswap.rs
```diff
@@ -68,26 +70,33 @@ impl UniswapLiquidity {
...
+        };
+
+        for order in offchain_orders {
+            get_reserves_if_not_yet_tracked(
+                TokenPair::new(order.buy_token, order.sell_token).expect("buy token = sell token"),
+            );
+
+            // Include every token with native token pair in the pools
+            TokenPair::new(order.buy_token, self.native_token_wrapper)
+                .map(&mut get_reserves_if_not_yet_tracked);
```
vkgnosis, line 82: I find it a little cleaner to use `if let` here instead of `map` because we do not actually want to map, we just call it for the side effect. This works for `Option` but for example for `Iterator` it wouldn't because iterators are lazy.

vkgnosis (approved):

PR #261, opened 2021-02-02
Retry transaction sending when gas prices increase

vkgnosis:
Using the retry crate from gpv1.

A lot of the code is similar to solution_submission.rs in gp-v1 but adapted to the rest of the gpv2 code. `is_transaction_error` and `GasPriceStream` are exactly the same. It wasn't worth to factor those into individual crates because they are small and the stream relies on a sleep trait which is project specific.

### Test Plan
CI e2e test.

### DIFF
--- a/e2e/tests/uniswap_trade_test.rs
+++ b/e2e/tests/uniswap_trade_test.rs
@@ -1,4 +1,4 @@
-use contracts::{ERC20Mintable, GPv2Settlement, UniswapV2Factory, UniswapV2Router02};
+use contracts::{ERC20Mintable, UniswapV2Factory, UniswapV2Router02};
 use ethcontract::{
     prelude::{Account, Address, Http, PrivateKey, Web3, U256},
     H160,
@@ -72,7 +72,8 @@ async fn test_with_ganache() {
     let uniswap_router = UniswapV2Router02::deployed(&web3)
         .await
         .expect("Failed to load deployed UniswapFactory");
-    let gp_settlement = GPv2Settlement::deployed(&web3)
+
+    let gp_settlement = solver::get_settlement_contract(&web3, solver.clone())
         .await
         .expect("Failed to load deployed GPv2Settlement");
     let gp_allowance = gp_settlement
--- a/solver/Cargo.toml
+++ b/solver/Cargo.toml
@@ -17,6 +17,7 @@ async-trait = "0.1"
 contracts = { path = "../contracts" }
 ethcontract = { version = "0.11", default-features = false }
 futures = "0.3"
+transaction-retry = { git = "https://github.com/gnosis/gp-transaction-retry.git", tag = "v0.1.0" }
 gas-estimation = { git = "https://github.com/gnosis/gp-gas-estimation.git", tag = "v0.1.0", features = ["web3_"] }
 hex = "0.4"
 hex-literal = "0.3"
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -10,6 +10,10 @@ use gas_estimation::GasPriceEstimating;
 use std::time::Duration;
 use tracing::info;
 
+// There is no economic viability calculation yet so we're using an arbitrary very high cap to
+// protect against a gas estimator giving bogus results that would drain all our funds.
+const GAS_PRICE_CAP: f64 = 500e9;
+
 pub struct Driver {
     settlement_contract: GPv2Settlement,
     orderbook: OrderBookApi,
@@ -87,10 +91,11 @@ impl Driver {
         } else {
             // TODO: check if we need to approve spending to uniswap
             settlement_submission::submit(
-                settlement,
                 &self.settlement_contract,
                 self.gas_price_estimator.as_ref(),
                 self.target_confirm_time,
+                GAS_PRICE_CAP,
+                settlement,
             )
             .await
             .context("failed to submit settlement")?;
--- a/solver/src/settlement_submission/retry.rs
+++ b/solver/src/settlement_submission/retry.rs
@@ -1,0 +1,109 @@
+use super::EncodedSettlement;
+use anyhow::Result;
+use contracts::GPv2Settlement;
+use ethcontract::{
+    dyns::DynMethodBuilder,
+    errors::{ExecutionError, MethodError},
+    jsonrpc::types::Error as RpcError,
+    transaction::{confirm::ConfirmParams, ResolveCondition},
+    web3::error::Error as Web3Error,
+    GasPrice, Void,
+};
+use primitive_types::U256;
+use transaction_retry::{TransactionResult, TransactionSending};
+
+fn is_transaction_error(error: &ExecutionError) -> bool {
+    // This is the error as we've seen it on openethereum nodes. The code and error messages can
+    // be found in openethereum's source code in `rpc/src/v1/helpers/errors.rs`.
+    // TODO: check how this looks on geth and infura. Not recognizing the error is not a serious
+    // problem but it will make us sometimes log an error when there actually was no problem.
+    matches!(error, ExecutionError::Web3(Web3Error::Rpc(RpcError { code, .. })) if code.code() == -32010)
+}
+
+pub struct SettleResult(pub Result<(), MethodError>);
+impl TransactionResult for SettleResult {
+    fn was_mined(&self) -> bool {
+        if let Err(err) = &self.0 {
+            !is_transaction_error(&err.inner)
+        } else {
+            true
+        }
+    }
+}
+
+pub struct SettlementSender<'a> {
+    pub contract: &'a GPv2Settlement,
+    pub nonce: U256,
+    pub settlement: EncodedSettlement,
+}
+#[async_trait::async_trait]
+impl<'a> TransactionSending for SettlementSender<'a> {
+    type Output = SettleResult;
+    async fn send(&self, gas_price: f64) -> Self::Output {
+        tracing::info!("submitting solution transaction at gas price {}", gas_price);
+        let mut method = settle_method_builder(self.contract, self.settlement.clone())
+            .nonce(self.nonce)
+            .gas_price(GasPrice::Value(U256::from_f64_lossy(gas_price)));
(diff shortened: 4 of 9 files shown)


e2e/tests/uniswap_trade_test.rs
```diff
@@ -59,7 +59,13 @@ async fn test_with_ganache() {
     let uniswap_router = UniswapV2Router02::deployed(&web3)
         .await
         .expect("Failed to load deployed UniswapFactory");
-    let gp_settlement = GPv2Settlement::deployed(&web3)
+
+    // I took this private key from the gp-v1 e2e test. Not sure why this actually works.
+    // Does Ganache guarantee that it works?
+    let private_key = "4f3edf983ac636a65a842ce7c78d9aa706d3b113bce9c46f30d7d21715b23b1d"
```
fleupold, line 65: If ganache is started with the default seed phrase, this will be the 0th account in the list. I don't think we should rely on this though. The more general way to use accounts is via the `Accounts` type. On ganache accounts are "unlocked" meaning they can sign transactions. For xDAI/mainnet/Rinkeby we can create an offline account that is initialized with a private key.

I think `get_settlement_contract` should take `Account` as an argument instead of private key, so that we can pass account[0] in the e2e test.

solver/src/driver.rs
```diff
@@ -80,15 +84,29 @@ impl Driver {
...
             self.target_confirm_time,
+            GAS_PRICE_CAP,
+            settlement,
         )
         .await
         .context("failed to submit settlement")?;
         Ok(())
     }
+
+    async fn transaction_count(&self) -> Result<U256> {
```
fleupold, line 105: nit: why have this logic here and not in `submit`? Seems like it would fit more coherently over there.

solver/src/settlement_submission/retry.rs
```diff
@@ -0,0 +1,136 @@
...
+// Create a never ending stream of gas prices based on checking the estimator in fixed intervals
+// and enforcing the minimum increase. Errors are ignored.
+pub fn gas_price_stream(
+    target_confirm_time: Duration,
+    gas_price_cap: f64,
+    estimator: &dyn GasPriceEstimating,
+) -> impl Stream<Item = f64> + '_ {
+    let stream = stream::unfold(true, move |first_call| async move {
+        if !first_call {
+            tokio::time::delay_for(GAS_PRICE_REFRESH_INTERVAL).await;
```
fleupold, line 26: ubernit: move constant to this module?

solver/src/settlement_submission/retry.rs
```diff
@@ -0,0 +1,136 @@
...
+        }
+        let estimate = estimator
+            .estimate_with_limits(MAX_GAS as f64, target_confirm_time)
```
fleupold, line 29: Can we add a todo to use the real estimate gas amount instead of MAX_GAS?

solver/src/settlement_submission/retry.rs
```diff
@@ -0,0 +1,136 @@
...
+            Err(err) => {
+                tracing::error!("gas price estimation failed: {:?}", err);
+                None
+            }
+        }
+    });
+    transaction_retry::gas_price_increase::enforce_minimum_increase_and_cap(gas_price_cap, stream)
+}
+
+fn is_transaction_error(error: &ExecutionError) -> bool {
```
fleupold, line 48: It's a shame we are duplicating this logic. Would it not make sense to also expose this method in the retry crate (not blocking this PR but generally)?

solver/src/settlement_submission/retry.rs
```diff
@@ -0,0 +1,136 @@
+use super::{EncodedSettlement, GAS_PRICE_REFRESH_INTERVAL, MAX_GAS};
```
fleupold, line 1: It feels to me as if this module is collecting a bunch of parts that are not really related to one another. Maybe the gas_price_stream should be part of the `gas_price_estimation` module and SettleResult/SettlementSender also seem not strictly related to retry. Maybe this module could be called transaction?

solver/src/settlement_submission/retry.rs
```diff
@@ -0,0 +1,136 @@
...
+                tracing::debug!("estimated gas price {}", gas_price);
+                Some(gas_price)
+            }
+            Err(err) => {
+                tracing::error!("gas price estimation failed: {:?}", err);
+                None
+            }
+        }
+    });
+    transaction_retry::gas_price_increase::enforce_minimum_increase_and_cap(gas_price_cap, stream)
```
fleupold, line 45: I personally think it would be nice to expose a gas price stream factory that takes a target time, amount, price_cap, estimator and a "sleep" function from the retry or price estimating crate. It seems generic enough to me and the logic with the unfolded stream is a bit nifty (so nice not to maintain it twice).

solver/src/settlement_submission/retry.rs
```diff
@@ -0,0 +1,136 @@
...
+        tracing::info!("submitting solution transaction at gas price {}", gas_price);
+        let mut method = self
+            .contract
+            .settle(
+                self.settlement.tokens.clone(),
+                self.settlement.clearing_prices.clone(),
+                self.settlement.encoded_trades.clone(),
+                self.settlement.encoded_interactions.clone(),
+                Vec::new(),
+            )
```
fleupold, line 85: nit: This code is duplicated between simulate and execute. Maybe the settlement helper struct can expose a function that returns a method builder given?

fleupold (approved):
The only blocking thing is that I wouldn't rely on ganache being started with the correct seed phrase. Otherwise mostly nits, I think it would have been nice to reuse more of the logic, but ok.

solver/Cargo.toml
```diff
@@ -17,6 +17,7 @@ async-trait = "0.1"
 contracts = { path = "../contracts" }
 ethcontract = { version = "0.11", default-features = false }
 futures = "0.3"
+transaction-retry = { git = "https://github.com/gnosis/gp-transaction-retry.git", rev = "b12c37e0916854dbf046d14e17cc4a7765666610" }
```
bh2smith, line 20: Have we not settled on a version of this yet?

solver/src/driver.rs
```diff
@@ -7,10 +7,14 @@ use crate::{
...
 use contracts::GPv2Settlement;
 use gas_estimation::GasPriceEstimating;
+use primitive_types::U256;
 use std::time::Duration;
 use tracing::info;
 
 const SETTLE_INTERVAL: Duration = Duration::from_secs(30);
+// There is no economic viability calculation yet so we're using an arbitrary very high cap to
+// protect against a gas estimator giving bogus results that would drain all our funds.
+const GAS_PRICE_CAP: f64 = 500e9;
```
bh2smith, line 16: Seems reasonable... 500 Gwei happens sometimes.

solver/src/driver.rs
bh2smith: I am actually surprised how many lines are needed just to get the nonce. Also, not sure what the settlement contract has to do with an EOA's nonce.

bh2smith (commented):
Looks nice so far. Seems there are a few things to cleanup at this time.

bh2smith (approved):

solver/src/settlement_submission/retry.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use jsonrpc_core::ErrorCode;
+
+    #[test]
+    fn test_submission_result_was_mined() {
```
bh2smith, line 118: Does this test somehow include a retry? If so, I am unable to tell how.

solver/src/settlement_submission/retry.rs
vkgnosis: Moved the stream to it's own module. Don't like it in gas_price_estimation because that is just responsible for creating the estimator whereas the stream is only used for retry.

vkgnosis: > SettleResult/SettlementSender also seem not strictly related to retry. Maybe this module could be called transaction

Imo they are because these types are only used for retry.

solver/src/settlement_submission/retry.rs
vkgnosis: The problem is that there are too many implementation details like sleeping, logging and how to call the estimator and there is no good generic way to take them. To me duplicating makes more sense.

solver/src/settlement_submission/retry.rs
vkgnosis: I don't think it makes sense in the retry crate because it is ethcontract specific. Maybe behind an ethcontract feature but then it is annoying when the version of ethcontract doesn't match. Maybe it would make sense in ethcontract or web3 directly but I don't think so because it is so node specific.

solver/src/settlement_submission/retry.rs
vkgnosis: Used in several places now so should stay in parent.

solver/src/settlement_submission/retry.rs
vkgnosis: it does not and is not supposed to. just tests that detection of was_mined works.

solver/Cargo.toml
vkgnosis: I haven't pushed a tag for it yet so no. When we make the next change in that repo we can set a version.

solver/src/settlement_submission/retry.rs
fleupold: This discussion goes outside the scope of this PR, but what's the concern of making the retry library opinionated on e.g. logging/sleeping (not that we could pass it as abstract traits from above) similar to how other libraries (e.g. web3 are)? 

It find there is still quite a lot of manual code involved to integrate transaction retries and it feels to me that the use case of wanting to submit some MethodBuilder with a specific target time and a gas price estimator is quite common.

vkgnosis: I suspect that it wouldn't be useful because people want to customize too many parts of it. Like what if you want to do some extra work every time a resubmission happens other than what it provides by default. I did make an issue now on the retry repo to try implementing this with an optional `ethcontract` feature so I am going to try it out.
I have another idea for the gas price stream too.

PR #264, opened 2021-02-02
[Easy] Fix inverted logic in in-memory orderbook

fleupold:
Retain keeps the elements for which the condition returns true. Thus we want to retain the orders that are not to be removed.

### Test Plan
Added a step to the e2e test that would fail before


bh2smith (approved):
Prima!

fleupold:
I guess this is no longer needed as the in memory orderbook is being deleted in another PR. Will merge regardless in case we want to resurrect it one day (might be better to resurrect a version with less bug).

vkgnosis:
For one of my PRs I also ran into this CI failure, will make PR for it shortly.

PR #266, opened 2021-02-03
Fix musl image to prevent deploy stet from failing

fleupold:
The deploy step on master is [failing](https://github.com/gnosis/gp-v2-services/runs/1818274604) because of a compilation error in the musl target.

It looks like it's related to the latest musl image which was published yesterday (none of the failing dependencies changed during that time): https://hub.docker.com/r/clux/muslrust

It seems like the latest image is based on a daily nightly release. A simple fix for now is to fix the musl image we use to the last know working one.

### Test Plan
Tested locally that `clux/muslrust:nightly-2021-02-02` fails compilation, but `clux/muslrust:nightly-2021-02-01` passes.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM clux/muslrust:nightly-2021-02-01 as cargo-build
+FROM clux/muslrust:stable as cargo-build
 WORKDIR /usr/src/oba-services
 
 # Copy and Build Code
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM clux/muslrust as cargo-build
+FROM clux/muslrust:nightly-2021-02-01 as cargo-build
 WORKDIR /usr/src/oba-services
 
 # Copy and Build Code


docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM clux/muslrust as cargo-build
+FROM clux/muslrust:nightly-2021-02-01 as cargo-build
```
nlordell, line 1: Why not use the [stable tag](https://hub.docker.com/layers/clux/muslrust/stable/images/sha256-71f83effec7fac19ace2acba2de9a022bc0210db0809ec4ece6b43a72575b8cd?context=explore) for the latest stable Rust release:
```suggestion
FROM clux/muslrust:stable as cargo-build
```

This also makes it so the tests are run on the same compiler version as what is being deployed.

fleupold: Good point. I only saw their nightly builds.

nlordell (approved):

vkgnosis (approved):

PR #269, opened 2021-02-03
Fix check that order has been settled

vkgnosis:
The order book api is supposed to still return settled orders when
queried by uid but the e2e test was using this condition. This works
with the in memory order book because it cleans up the orders when they
are settled where as the postgres order book does not (it filters them on the
get_orders route depending on the filter).

### Test Plan
CI e2e test

### DIFF
--- a/e2e/tests/uniswap_trade_test.rs
+++ b/e2e/tests/uniswap_trade_test.rs
@@ -10,7 +10,7 @@ use orderbook::{
 };
 use secp256k1::SecretKey;
 use serde_json::json;
-use solver::liquidity::uniswap::UniswapLiquidity;
+use solver::{liquidity::uniswap::UniswapLiquidity, orderbook::OrderBookApi};
 use std::{str::FromStr, sync::Arc, time::Duration};
 use web3::signing::SecretKeyRef;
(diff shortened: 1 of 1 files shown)


e2e/tests/uniswap_trade_test.rs
```diff
@@ -211,18 +207,20 @@ async fn test_with_ganache() {
...
-            "{}{}{}",
-            API_HOST,
-            ORDER_PLACEMENT_ENDPOINT,
-            order_b.uid(&trader_b.address())
-        ))
-        .send()
-        .await;
-    assert_eq!(placement.unwrap().status(), 404);
+    let orders = create_orderbook_api().get_orders().await.unwrap();
+    assert!(orders
```
fleupold, line 211: Would it be safe to just check that it's empty, given that both orders should be matched?

fleupold (approved):

e2e/tests/uniswap_trade_test.rs
vkgnosis: I guess that should be fine too because we clear the order database in the test. Will change to that.

PR #271, opened 2021-02-03
Enable event updater

vkgnosis:
Silly mistake on my part. I wrote the whole EventUpdater code but never integrated it as part of the maintenance function.
I tested locally that this works with mainnet. Note that the first event update is going to take a really long time because it has to start at block 0.  This wasn't  the case in v1 because there we had information about the deployment transaction hash available in the artifact which ethcontract knows how to use in order to set a lower bound for the block of the first event. I am working on a PR to ethcontract to allow us to use that there too.

### Test Plan
CI

### DIFF
--- a/orderbook/src/event_updater.rs
+++ b/orderbook/src/event_updater.rs
@@ -4,11 +4,13 @@ use contracts::{
     g_pv_2_settlement::{event_data::Trade as ContractTrade, Event as ContractEvent},
     GPv2Settlement,
 };
-use ethcontract::{errors::ExecutionError, Event as EthcontractEvent, EventMetadata};
+use ethcontract::{
+    dyns::DynTransport, errors::ExecutionError, Event as EthcontractEvent, EventMetadata,
+};
 use futures::{Stream, StreamExt, TryStreamExt};
 use model::order::OrderUid;
 use std::{convert::TryInto, ops::RangeInclusive};
-use web3::{Transport, Web3};
+use web3::Web3;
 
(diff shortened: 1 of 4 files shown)


fleupold (approved):
> Note that the first event update is going to take a really long time because it has to start at block 0.

Given we don't really care about any trades happening before we start tracking order, would it make sense to start at the current block if the database is empty? 
In a year from now when I run things locally from an empty DB I might not want to resync blocks from today on.

vkgnosis:
It is something to think about, maybe as an optional command line argument. In general I think it is more likely that we do want all events and starting from scratch is rare, or you can import an existing event db like we sometimes did for v1.
The problem is someone could re-submit an order that has been filled a long time ago but is still valid_to and you would try to settle the order because you will never see the old filled event.

fleupold:
> The problem is someone could re-submit an order that has been filled a long time ago but is still valid_to and you would try to settle the order because you will never see the old filled event.

For this case I'd prefer a check in the add order handler that prevents people from resubmitting already known invalid orders.

> or you can import an existing event db like we sometimes did for v1.

How do you envision the local setup for development? I was not planning to keep a valid database with all orders locally, but would just spin up a fresh postgres docker container whenever I need to run the binary. For this I'd like to avoid syncing from scratch.

vkgnosis:
I thought we would keep a database like we had to in v1, but sure it would be better without for local development. After this is merged I am going to make another PR with some command line argument we can use for local testing. Could just be a flag that checks the current block on start up and doesn't handle events older than that.

PR #273, opened 2021-02-03
Do not submit empty settlement

vkgnosis:
Noticed in kibana that were are submitting empty solutions on xdai. Not sure if this is a bug in the naive solver but anyway it is probably a good idea to have this check.

### Test Plan
still compiles

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -80,6 +80,9 @@ impl Driver {
             Some(settlement) => settlement,
         };
         info!("Computed {:?}", settlement);
+        if settlement.trades.is_empty() {
+            info!("Skipping empty settlement");
+        }
         // TODO: check if we need to approve spending to uniswap
         settlement_submission::submit(
             settlement,


solver/src/driver.rs
```diff
@@ -80,6 +80,9 @@ impl Driver {
             Some(settlement) => settlement,
         };
         info!("Computed {:?}", settlement);
+        if settlement.trades.is_empty() {
```
fleupold, line 83: Should we also check that interactions are empty? I don't think we have defined yet why we would want to settle a trade with just interactions, but I think it's conceptually possible (maybe via some 0x limit orders that are not coming from market makers and would give us positive fees)

fleupold (approved):
Good catch!

solver/src/driver.rs
vkgnosis: I don't think we should check it now because right now that is more likely a bug then intentional. If we do have such a use case we can change it.

PR #274, opened 2021-02-04
Extend get_orders with more filters and exlclude invalid by default

vkgnosis:
This allows users of the api to make use of the full available filters.
It also fixes a bug where by default we were not setting min_valid_to
which would return old orders in the postgres orderbook.

### Test Plan
CI

### DIFF
--- a/shared/src/lib.rs
+++ b/shared/src/lib.rs
@@ -1,2 +1,3 @@
 pub mod arguments;
+pub mod time;
 pub mod tracing;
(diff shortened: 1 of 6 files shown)


shared/src/lib.rs
```diff
@@ -1,2 +1,3 @@
 pub mod arguments;
 pub mod tracing;
+pub mod valid_to;
```
fleupold, line 3: nanonit: `valid_to` is quite specific (the function can be used for more). Maybe `time` is a better name?

fleupold (approved):
Nice fix!

PR #275, opened 2021-02-04
[Trivial] Actually skip empty settlement

fleupold:
🤡 review of #273 didn't make me catch that we are not actually doing anything except for printing we will skip settlement. This PR makes it so we actually don't submit in this case

### Test Plan
Test we don't see `solver::settlement_submission: Settlement call` and `Skipping empty settlement` in the same runloop


vkgnosis (approved):
:|

PR #276, opened 2021-02-04
Do not use MockStorage in order book tests

vkgnosis:
Came up when I wanted to remove the Storage trait in the other PR. I
prefer not using mocks when possible like in this case. I refactored
get_orders so that it can be tested without Storage.

### Test Plan
CI


orderbook/src/orderbook.rs
```diff
@@ -89,74 +63,65 @@ impl Orderbook {
...
-        predicate::{always, eq},
-        Sequence,
-    };
-    use model::{
-        order::{OrderBuilder, OrderCreation},
-        DomainSeparator,
-    };
-
-    #[tokio::test]
-    async fn watches_owners_sell_token_balance_for_added_orders() {
```
vkgnosis, line 111: This test I removed because all it tests is that add_order calls balance_fetcher.register which we do unconditionally so I don't think there is a point in having a test for it.

fleupold: Even if the codepath is unconditional, I feel like it's a nice documentation and assurance of behaviour of the public contract of this component. If one day we refactored the orderbook and remove the line to register balances for on the add path I think it's nice when an automated test catches that.

That being said, if this is the only thing requiring the mock, I'm ok with letting it go.

fleupold (approved):
What I like about this change is that it makes the tests much more concise. What I don't like so much is that now the tests are very much dependent on how the logic is factored internally into private helper methods.

I think we had this discussion in some form or another before, but my general preference when it comes to testing is to think about what the component is supposed to do (in terms of a public/generic contract) and then write unit tests that assert that behaviour. What I like about this kinda "black-box" approach is that later on we can refactor the component internally as we like, if the tests pass the component is likely still working as expected.

I see that with `mockall` this approach is not always working that well, because we have to set all the expectations on the dependencies which in the end requires some form of white box understanding of the internal interactions of the component.

vkgnosis:
I agree that there is no clear winner here. The black boxness has a benefit but I don't think that we get this in our case because the test is not really black box as it has to set up assertions on private behavior too in the form of the mocked traits. Arguable our black box shouldn't need this either.

> If one day we refactored the orderbook and remove the line to register balances for on the add path I think it's nice when an automated test catches that.

This is true but also I feel that if you remove the line in the code then you might also accidentally remove the line in the test. I tend to see tests as making sure some complex operation really does what we want to, but not useful to repeat back the original function if it is so simple.

If we want true black box tests which I agree are useful, we could try writing more e2e tests. I think we can probably have more of them compared to v1 because here we can control the driver and orderbook in a fine grained way without long wait times. Will investigate that.

PR #278, opened 2021-02-05
Differentiate between sellAmount w & w/o fees

### LINKED ISSUE #262: Make order executed amounts easier to use
As noted in https://github.com/gnosis/gp-v2-contracts/issues/392 we should change the executed_ fields slightly to make them easier to use by frontends.

fleupold:
Fixes https://github.com/gnosis/gp-v2-contracts/issues/392

We currently display the executed sell amounts as they are emitted by the smart contracts which includes any fees paid in the settle token (thus it's not the gross exchange rate).

This PR make this explicit in the Trade model (without changing existing DB semantics) and adds redundancy to the order metadata, explicitly stating the net and gross sell amount (happy for other naming suggestions)

cc @alfetopito @anxolin 

### Test Plan
CI

### DIFF
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -272,7 +272,9 @@ pub struct OrderMetaData {
     #[serde(with = "serde_with::rust::display_fromstr")]
     pub executed_buy_amount: BigUint,
     #[serde(with = "serde_with::rust::display_fromstr")]
-    pub executed_sell_amount: BigUint,
+    pub executed_net_sell_amount: BigUint,
+    #[serde(with = "serde_with::rust::display_fromstr")]
+    pub executed_gross_sell_amount: BigUint,
     #[serde(with = "serde_with::rust::display_fromstr")]
     pub executed_fee_amount: BigUint,
     pub invalidated: bool,
@@ -286,7 +288,8 @@ impl Default for OrderMetaData {
             uid: Default::default(),
             available_balance: Default::default(),
             executed_buy_amount: Default::default(),
-            executed_sell_amount: Default::default(),
+            executed_net_sell_amount: Default::default(),
+            executed_gross_sell_amount: Default::default(),
             executed_fee_amount: Default::default(),
             invalidated: Default::default(),
         }
@@ -399,8 +402,9 @@ mod tests {
             "uid": "0x1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111",
             "availableBalance": "100",
             "executedBuyAmount": "3",
-            "executedSellAmount": "4",
-            "executedFeeAmount": "5",
+            "executedNetSellAmount": "4",
+            "executedGrossSellAmount": "5",
+            "executedFeeAmount": "6",
             "invalidated": true,
             "sellToken": "0x000000000000000000000000000000000000000a",
             "buyToken": "0x0000000000000000000000000000000000000009",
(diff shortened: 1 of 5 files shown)


orderbook/openapi.yml
```diff
@@ -246,8 +246,11 @@ components:
           description: "Amount of sellToken available for the settlement contract to spend on behalf of the owner. Null if API was unable to fetch balance."
           $ref: "#/components/schemas/TokenAmount"
           nullable: true
-        executedSellAmount:
-          description: "The total amount of sellToken that has been executed for this order."
+        executedNetSellAmount:
+          description: "The total amount of sellToken that has been executed for this order including fees."
+          $ref: "#/components/schemas/BigUint"
+        executedGrossSellAmount:
+          description: "The total amount of sellToken that has been executed for this order without fees."
```
alfetopito, line 253: I'm in favour of keeping the field named `executedSellAmount` with the contents being `value - fee`, and using `executedGrossSellAmount` for same as is today (equivalent to contract; `value + fee`)

fleupold: You are just lazy to update the frontend 😉 (I guess backwards compatibility is a reasonable argument, although we are still early in our development process).
I really don't have a strong feeling, just felt like if there is a gross there should also be a net...

alfetopito: Well, that is the case most of the time... :sweat_smile: but what I'm proposing will require changes since right now I'm [deducing `executedFeeAmount` from `executedSellAmount` to get the final value](https://github.com/gnosis/gp-ui/blob/develop/src/api/operator/utils.ts#L82).

My reasoning is to have the name and contents of `executed<Buy|Sell>Amount` equivalent, with the extra info for `sells` in another variable.
So the next dev working on this is not surprised that `executedBuyAmount` doesn't have the same "meaning" as `executedSellAmount`.
Although one could argue that the original meaning is correct already since it's what's added/removed from user's balance... :shrug:

anxolin: I don't mind too much about the name of this, but I have to say, gross/net confused me. 

Here the gross says `without fees`, which I would read as "not including fees". If they don't include fees, for me is they don't have the fees applied. But I think you mean the opposite. 

Maybe is easier just to say executedSellAmount, executedSellAmountBeforeFees

I'd be ok with net/gross, just the "without" in the description confused me.

model/src/order.rs
```diff
@@ -399,8 +402,9 @@ mod tests {
             "uid": "0x1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111",
             "availableBalance": "100",
             "executedBuyAmount": "3",
-            "executedSellAmount": "4",
-            "executedFeeAmount": "5",
+            "executedNetSellAmount": "4",
+            "executedGrossSellAmount": "5",
+            "executedFeeAmount": "6",
```
anxolin, line 407: Are these numbers alright? if the fee is 6, then shoudln't be the gross net+fee: `4 + 6 = 10` <> `5`

orderbook/openapi.yml
fleupold: I think I mixed up net/gross in the first version. Might be a sign that using executedSellAmount vs executedSellAmountBeforeFees is more intuitive for non-accountants. Will change

model/src/order.rs
fleupold: They are not correct. The issue is that the datatype doesn't enforce that the values are interdependent (and this unit test only tests the encoding), since I didn't find a good way to make serde (the serialisation framework) derive an extra value from a struct (other than using an inner struct which would then itself have the same issue).

vkgnosis (approved):

alfetopito (approved):

anxolin (approved):

PR #284, opened 2021-02-08
[Easy] convert get orders error to reply

bh2smith:
A simple PR that eliminates duplication of the following match statement over two different files:

```js
    match result {
        Ok(orders) => Ok(reply::with_status(reply::json(&orders), StatusCode::OK)),
        Err(err) => {
            tracing::error!(?err, "get_orders error");
            return Ok(reply::with_status(
                super::internal_error(),
                StatusCode::INTERNAL_SERVER_ERROR,
            ));
        }
    }
```

We introduce `common.rs` with shared method `convert_get_orders_error_to_reply` that is used in two places.

### Test Plan
Existing unit tests.

### DIFF
--- a/orderbook/src/api.rs
+++ b/orderbook/src/api.rs
@@ -54,7 +54,7 @@ fn internal_error() -> Json {
 
 pub fn convert_get_orders_error_to_reply(err: anyhowError) -> WithStatus<Json> {
     tracing::error!(?err, "get_orders error");
-    return with_status(internal_error(), StatusCode::INTERNAL_SERVER_ERROR);
+    with_status(internal_error(), StatusCode::INTERNAL_SERVER_ERROR)
 }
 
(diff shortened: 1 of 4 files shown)


orderbook/src/api.rs
```diff
@@ -1,3 +1,4 @@
+mod common;
```
vkgnosis, line 1: We have put other common code directly into `api.rs` so I think `convert_get_orders_error_to_reply` can also go there.

vkgnosis (approved):

orderbook/src/api.rs
bh2smith: Cool, didn't know this. Will do.

bh2smith: Done!

fleupold (approved):

PR #286, opened 2021-02-09
Parse http solver response

vkgnosis:
Edit: will split this up into smaller PRs

First step in updating the http solver. Previously we were not parsing the response.
Change is smaller than it looks because some structs moved to new place and test takes a lot of lines to set up.

### Test Plan
New unit test but cannot test with the staging deployed solver atm because it is running an older version with some bugs and needs a new release.

### DIFF
--- a/solver/src/http_solver.rs
+++ b/solver/src/http_solver.rs
@@ -1,10 +1,23 @@
-use crate::{liquidity::Liquidity, settlement::Settlement, solver::Solver};
-use anyhow::{Context, Result};
-use model::{order::OrderKind, u256_decimal};
-use primitive_types::{H160, U256};
+mod model;
+mod settlement;
+
+use self::model::*;
+use crate::{
+    liquidity::{AmmOrder, LimitOrder, Liquidity},
+    settlement::Settlement,
+    solver::Solver,
+};
+use ::model::order::OrderKind;
+use anyhow::{ensure, Context, Result};
+use primitive_types::H160;
 use reqwest::{Client, Url};
-use serde::{Deserialize, Serialize, Serializer};
 use std::collections::{HashMap, HashSet};
+
+// TODO: limit trading for tokens that don't have uniswap - fee pool
+// TODO: exclude partially fillable orders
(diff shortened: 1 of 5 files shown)


solver/src/http_solver.rs
```diff
@@ -114,140 +155,111 @@ impl HttpSolver {
...
             .context("failed to get response body")?;
-        let _decoded: Solution =
-            serde_json::from_slice(&body).with_context(|| match std::str::from_utf8(&body) {
-                Ok(body) => format!("failed to decode response body: {}", body),
-                Err(_) => format!("failed to decode response body: {:?}", body),
-            })?;
-        Ok(None)
+        let body_str =
+            std::str::from_utf8(body_bytes.as_ref()).context("failed to decode response body")?;
+        tracing::debug!("response body {}", body_str);
```
vkgnosis, line 184: added more logging here to ensure we always log the response even on non success status code

solver/src/http_solver.rs
```diff
@@ -114,140 +155,111 @@ impl HttpSolver {
...
+            status
+        );
+        serde_json::from_str(body_str).context("failed to decode response json")
     }
 }
 
-// types used in the solver http api
-
-#[derive(Debug, Default, Serialize)]
-struct BatchAuctionModel {
```
vkgnosis, line 168: These have moved to model.rs, along with new structs for the response

solver/src/http_solver.rs
```diff
@@ -114,140 +155,111 @@ impl HttpSolver {
...
+    use crate::liquidity::{
+        AmmOrder, LimitOrder, MockAmmSettlementHandling, MockLimitOrderSettlementHandling,
+    };
+    use ::model::TokenPair;
+    use num::Rational;
+    use std::sync::Arc;
 
     // cargo test real_solver -- --ignored --nocapture
+    // set the env variable GP_V2_OPTIMIZER_URL to use a non localhost optimizer
     #[tokio::test]
```
vkgnosis, line 229: I am using an env variable because it contains credentials which we don't want in version control.

PR #287, opened 2021-02-09
Move http solver types to own module and add response structs

vkgnosis:
Moved to new module because it got too big.

### Test Plan
Can't test response structs yet because solver has a bug that makes it return wrong type.


fleupold (approved):
👍

PR #288, opened 2021-02-09
Make http solver always log raw request and response

vkgnosis:
This is useful for debugging.

### Test Plan
Tested with local solver.

### DIFF
--- a/solver/src/http_solver.rs
+++ b/solver/src/http_solver.rs
@@ -3,7 +3,7 @@ mod model;
 use self::model::*;
 use crate::{liquidity::Liquidity, settlement::Settlement, solver::Solver};
 use ::model::order::OrderKind;
-use anyhow::{Context, Result};
+use anyhow::{ensure, Context, Result};
 use primitive_types::H160;
 use reqwest::{Client, Url};
 use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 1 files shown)


solver/src/http_solver.rs
```diff
@@ -130,36 +130,43 @@ impl HttpSolver {
...
-    async fn solve(&self, orders: Vec<Liquidity>) -> Result<Option<Settlement>> {
+    async fn send(&self, model: &BatchAuctionModel) -> Result<SettledBatchAuctionModel> {
         let mut url = self.base.clone();
         url.set_path("/solve");
         self.config.add_to_query(&mut url);
-        let body = self.create_body(orders.as_slice()).await?;
+        let query = url.query().map(ToString::to_string).unwrap_or_default();
         let mut request = self.client.post(url);
         if let Some(api_key) = &self.api_key {
             request = request.header("X-API-KEY", api_key);
```
fleupold, line 141: nit: maybe we can set this header to [sensitve](https://docs.rs/reqwest/0.11.0/reqwest/header/struct.HeaderValue.html#method.set_sensitive) to that it doesn't get leaked in the logs?

solver/src/http_solver.rs
```diff
@@ -130,36 +130,43 @@ impl HttpSolver {
...
+        let body_bytes = response
             .bytes()
             .await
             .context("failed to get response body")?;
-        let _decoded: SettledBatchAuctionModel =
-            serde_json::from_slice(&body).with_context(|| match std::str::from_utf8(&body) {
-                Ok(body) => format!("failed to decode response body: {}", body),
-                Err(_) => format!("failed to decode response body: {:?}", body),
-            })?;
+        let body_str =
```
fleupold, line 153: why not use `respone.text()` or even `response.json()` directly? (the latter I assume because you want to print the intermediate representation for debugging)

fleupold (approved):

solver/src/http_solver.rs
vkgnosis: done

solver/src/http_solver.rs
vkgnosis: fixed.

PR #289, opened 2021-02-09
Allow using remote http solver in test

vkgnosis:
This is an env variable because the url contains credentials which we
don't want in source control. Also adds an amm order to the test.

### Test Plan
It sends request to the staging solver (credentials in 1password). Can't decode response yet.

### DIFF
--- a/solver/src/http_solver.rs
+++ b/solver/src/http_solver.rs
@@ -173,35 +173,50 @@ impl Solver for HttpSolver {
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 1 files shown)


fleupold (approved):
I assume the API_KEY is not yet supported? I'd prefer using that over keeping the URL hidden (which might be easier to guess/reverse-engineer than an API key)

vkgnosis:
> I assume the API_KEY is not yet supported? I'd prefer using that over keeping the URL hidden (which might be easier to guess/reverse-engineer than an API key)

Afaik it using basic http authentication only atm.

PR #291, opened 2021-02-09
Price Estimator for Orderbook API

fleupold:
Part of computing reasonable fees.

This PR introduces a component that can compute a price estimate between two tokens based on the direct Uniswap pair. In the future, we can make the logic more sophisticated by considering multiple likely "base tokens" as intermediate hops. This is how the Uniswap SDK is implementing their price estimate (cf. [gp-v2-swap](https://github.com/gnosis/gp-v2-swap/blob/c53137bed99a9037800b67dd4468a4e4b2d4435b/src/constants/index.ts#L64))

### Test Plan
Added a unit test

### DIFF
--- a/orderbook/src/price_estimate.rs
+++ b/orderbook/src/price_estimate.rs
@@ -1,0 +1,88 @@
+use anyhow::{anyhow, Result};
+use ethcontract::H160;
+use model::TokenPair;
+use shared::uniswap_pool::PoolFetching;
+use std::iter::once;
+
+#[allow(dead_code)]
+struct UniswapPriceEstimator {
+    pool_fetcher: Box<dyn PoolFetching>,
+}
+
+impl UniswapPriceEstimator {
+    // Estimates the price using the direct pool between sell and buy token. Price is given in
+    // how much of sell_token needs to be sold for one buy_token.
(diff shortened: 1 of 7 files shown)


orderbook/src/price_estimate.rs
```diff
@@ -0,0 +1,88 @@
...
+            .fetch(once(pair).collect())
+            .await
+            .pop()
+            .ok_or_else(|| anyhow!("Uniswap pool does not exist"))?;
+        let (sell_reserve, buy_reserve) = if pool.tokens.get().0 == sell_token {
+            (pool.reserves.0, pool.reserves.1)
+        } else {
+            (pool.reserves.1, pool.reserves.0)
+        };
+        Ok((sell_reserve as f64) / (buy_reserve as f64))
```
vkgnosis, line 33: We might want to have a check for the case where one of the reserves is 0. Not strictly necessary I am just wondering if it could lead to problems with how this will be used in the future. If one of the reserves is 0 then the price could be 0 or +inf.

vkgnosis (approved):

orderbook/src/price_estimate.rs
fleupold: Good point! Will filter out those pools

PR #292, opened 2021-02-09
Component to estimate & validate min viable fee

fleupold:
This PR adds a component that allows estimating a min viable fee and validate if a fee is considered viable. For the latter, we store previous estimates in memory for now and upon validation check if we have given a fee estimate that satisfies the provided fee in the past.

We probably need to store these value in the database as otherwise we might run into issues as soon as we add HPA to our API or if fees are fetched before a cold start and the order is only submitted afterwards.

I will tackle this in a future PR.

Min fee is estimated by using a constant gas amount estimate (100k which is roughly what a Uniswap settlement costs), and dynamic gas as well as sell token price estimates.

### Test Plan
Unit tests

### DIFF
--- a/orderbook/src/fee.rs
+++ b/orderbook/src/fee.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 
 use anyhow::Result;
-use chrono::{DateTime, FixedOffset, Utc};
+use chrono::{DateTime, Duration, Utc};
 use primitive_types::{H160, U256};
 use std::sync::Mutex;
 
@@ -20,7 +20,7 @@ pub struct MinFeeCalculator {
 }
 
 const GAS_PER_ORDER: f64 = 100_000.0;
-const STANDARD_VALIDITY_FOR_FEE_IN_SEC: i32 = 60;
+const STANDARD_VALIDITY_FOR_FEE_IN_SEC: i64 = 60;
 
 impl MinFeeCalculator {
     pub fn new(
@@ -49,7 +49,7 @@ impl MinFeeCalculator {
             .await?;
 
         let min_fee = U256::from_f64_lossy(gas_price * token_price * GAS_PER_ORDER);
-        let valid_until = (self.now)() + FixedOffset::east(STANDARD_VALIDITY_FOR_FEE_IN_SEC);
+        let valid_until = (self.now)() + Duration::seconds(STANDARD_VALIDITY_FOR_FEE_IN_SEC);
         let result = (min_fee, valid_until);
 
         self.measurements
@@ -86,10 +86,10 @@ impl MinFeeCalculator {
 
 #[cfg(test)]
 mod tests {
+    use chrono::Duration;
     use std::cell::RefCell;
     use std::rc::Rc;
     use std::sync::Arc;
-    use std::time::Duration;
 
     use super::*;
 
@@ -104,7 +104,7 @@ mod tests {
     struct FakeGasEstimator(Arc<Mutex<f64>>);
     #[async_trait::async_trait]
     impl GasPriceEstimating for FakeGasEstimator {
-        async fn estimate_with_limits(&self, _: f64, _: Duration) -> Result<f64> {
+        async fn estimate_with_limits(&self, _: f64, _: std::time::Duration) -> Result<f64> {
             Ok(*self.0.lock().unwrap())
         }
     }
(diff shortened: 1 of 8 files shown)


orderbook/src/fee.rs
```diff
@@ -0,0 +1,180 @@
...
+            .measurements
+            .lock()
+            .expect("Thread holding mutex panicked")
+            .get(&token)
+        {
+            // Previous measurements are ordered ascending by expiry data
+            for (suggested_fee, expiry_date) in measurements.iter().rev() {
+                if expiry_date < &(self.now)() {
+                    break;
+                }
```
vkgnosis, line 76: Maybe this could also remove estimates that have expired? Otherwise we accumulate them forever. But if you're changing this component soon anyway then not important.

orderbook/src/fee.rs
```diff
@@ -0,0 +1,180 @@
...
+    // and an expiry date for the estimate.
+    pub async fn min_fee(&self, token: H160) -> Result<Measurement> {
+        let gas_price = self.gas_estimator.estimate().await?;
+        let token_price = self
+            .price_estimator
+            .estimate_price(token, self.native_token)
+            .await?;
+
+        let min_fee = U256::from_f64_lossy(gas_price * token_price * GAS_PER_ORDER);
+        let valid_until = (self.now)() + FixedOffset::east(STANDARD_VALIDITY_FOR_FEE_IN_SEC);
```
vkgnosis, line 52: I don't think the use of FixedOffset is correct in this file. It is meant for timezones but you are using as a way to change a point in time. Probably more appropriate is to to add and subtract Duration from DateTime<Utc>. Also since we shouldn't need timezones we can use SystemTime instead of DateTime.

vkgnosis (changes requested):

orderbook/src/fee.rs
fleupold: You are right, I borrowed the approach from [here](https://github.com/gnosis/gp-v2-services/blob/main/orderbook/src/api/get_fee_info.rs#L28). Will change it there, too.

orderbook/src/fee.rs
```diff
@@ -0,0 +1,178 @@
...
+    price_estimator: Box<dyn PriceEstimating>,
+    gas_estimator: Box<dyn GasPriceEstimating>,
+    native_token: H160,
+    // TODO persist past measurements to shared storage
+    measurements: Mutex<HashMap<H160, Vec<Measurement>>>,
+    now: Box<dyn Fn() -> DateTime<Utc>>,
+}
+
+const GAS_PER_ORDER: f64 = 100_000.0;
+const STANDARD_VALIDITY_FOR_FEE_IN_SEC: i64 = 60;
```
vkgnosis, line 23: Now you can make this a real `Duration` too.

orderbook/src/fee.rs
```diff
@@ -0,0 +1,178 @@
...
+
+use anyhow::Result;
+use chrono::{DateTime, Duration, Utc};
+use primitive_types::{H160, U256};
+use std::sync::Mutex;
+
+use crate::price_estimate::PriceEstimating;
+use gas_estimation::GasPriceEstimating;
+
+type Measurement = (U256, DateTime<Utc>);
```
vkgnosis, line 11: Like I said in the other comment this could be SystemTime instead of DateTime too. For the http api we would then convert it to DateTime in order to format it correctly but here it doesn't need to be. I don't know if this is better though, so  no need to change if you like DateTime.

vkgnosis (approved):

orderbook/src/fee.rs
fleupold: For writing it to the DB it might be nice to be denoted in DateTime<Utc> rather than system time (although system time seems to be UTC too). It's less conversion work to leave it as is.

orderbook/src/fee.rs
fleupold: Duration::seconds doesn't seem to be `const`. Probably not worth using a lazy_static here?

vkgnosis: Agree not worth using lazy static but I thought it was const https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_secs

fleupold: We are using `chrono::DateTime` and [`chrono::Duration::seconds`](https://docs.rs/chrono/0.4.19/chrono/struct.Duration.html#method.seconds) is not static

vkgnosis: Could create the chrono Duration from the std Duration when used. But anyway not important at all and not blocking the PR.

PR #293, opened 2021-02-09
Use min fee calculator in API

fleupold:
Wire up the min fee calculator with the API to return realistic price estimates to the UI. This is mostly gluing together code (moving gas estimator into the shared crate).

There are some more things to do, but this should already give us a good starting point for reasonable mainnet fees:
- reject orders that are not satisfying min viable fee on creation
- persist previous min fee estimates to disk
- use more common base pairs for price estimation

### Test Plan
Run the orderbook API for mainnet and see somewhat high fee estimates (~ 0.3GNO, 0.02ETH at the time of writing).


orderbook/src/api/get_fee_info.rs
```diff
@@ -22,25 +25,42 @@ pub fn get_fee_info_request() -> impl Filter<Extract = (H160,), Error = Rejectio
...
+        Ok((minimal_fee, expiration_date)) => {
+            let fee_info = FeeInfo {
+                expiration_date,
+                minimal_fee,
+                fee_ratio: 0u32,
+            };
+            Ok(reply::with_status(reply::json(&fee_info), StatusCode::OK))
+        }
+        Err(err) => {
+            tracing::error!(?err, "get_fee error");
```
vkgnosis, line 39: The way I used error logs like this for the other routes is for internal errors. If err just means that the token wasn't found then we shouldn't error log it imo because that is the expected result.
Might require some changes in the fee calculator api so that it can differentiate these cases.

vkgnosis (approved):

PR #294, opened 2021-02-10
Enforce fee

fleupold:
This PR makes it so that we reject orders for which the fee is too low.

### Test Plan

Change get_fee_info endpoint to return 0. Try to place order using frontend. Observe:

image

### DIFF
--- a/orderbook/src/api/create_order.rs
+++ b/orderbook/src/api/create_order.rs
@@ -51,6 +51,13 @@ pub fn create_order_response(result: Result<AddOrderResult>) -> impl Reply {
             ),
             StatusCode::BAD_REQUEST,
         ),
+        Ok(AddOrderResult::InsufficientFee) => (
+            super::error(
+                "InsufficientFee",
+                "consult /token/<address>/fee for minimum fee required",
+            ),
+            StatusCode::BAD_REQUEST,
+        ),
         Err(_) => (super::internal_error(), StatusCode::INTERNAL_SERVER_ERROR),
     };
     warp::reply::with_status(body, status_code)
--- a/orderbook/src/api/create_order.rs
+++ b/orderbook/src/api/create_order.rs
@@ -52,10 +52,7 @@ pub fn create_order_response(result: Result<AddOrderResult>) -> impl Reply {
             StatusCode::BAD_REQUEST,
         ),
         Ok(AddOrderResult::InsufficientFee) => (
-            super::error(
-                "InsufficientFee",
-                "consult /token/<address>/fee for minimum fee required",
-            ),
+            super::error("InsufficientFee", "Order does not include sufficient fee"),
             StatusCode::BAD_REQUEST,
         ),
         Err(_) => (super::internal_error(), StatusCode::INTERNAL_SERVER_ERROR),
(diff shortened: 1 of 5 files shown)


orderbook/Cargo.toml
```diff
@@ -14,20 +14,23 @@ path = "src/main.rs"
...
 chrono = { version = "0.4", default-features = false, features = ["clock"] }
 contracts = { path = "../contracts" }
 ethcontract = { version = "0.11", default-features = false }
 futures = "0.3.12"
+gas-estimation = { git = "https://github.com/gnosis/gp-gas-estimation.git", tag = "v0.1.0", features = ["web3_"] }
 hex = { version = "0.4", default-features = false }
 model = { path = "../model" }
 num-bigint = "0.3"
-primitive-types = "0.8"
+primitive-types = { version = "0.8", features = ["fp-conversion"] }
```
bh2smith, line 27: Love to see this in use! Its my baby

orderbook/src/api/create_order.rs
```diff
@@ -51,6 +51,13 @@ pub fn create_order_response(result: Result<AddOrderResult>) -> impl Reply {
             ),
             StatusCode::BAD_REQUEST,
         ),
+        Ok(AddOrderResult::InsufficientFee) => (
+            super::error(
+                "InsufficientFee",
+                "consult /token/<address>/fee for minimum fee required",
```
bh2smith, line 57: Could we provide the actual token address here given that we surely know it at the time of displaying this message?

orderbook/src/fee.rs
```diff
@@ -0,0 +1,178 @@
...
+pub struct MinFeeCalculator {
+    price_estimator: Box<dyn PriceEstimating>,
+    gas_estimator: Box<dyn GasPriceEstimating>,
+    native_token: H160,
+    // TODO persist past measurements to shared storage
+    measurements: Mutex<HashMap<H160, Vec<Measurement>>>,
+    now: Box<dyn Fn() -> DateTime<Utc> + Send + Sync>,
+}
+
+const GAS_PER_ORDER: f64 = 100_000.0;
```
bh2smith, line 22: Is this where we hard code the per trade gas cost average as benchmarked in the smart contract? Perhaps we can say a word about this here or include a link to the benchmarks?

solver/src/liquidity/uniswap.rs
```diff
@@ -4,7 +4,7 @@ use ethcontract::{batch::CallBatch, Http, Web3};
 use model::TokenPair;
 use num::rational::Rational;
 use primitive_types::{H160, U256};
-use shared::uniswap_pool::PoolFetcher;
+use shared::uniswap_pool::{PoolFetcher, PoolFetching as _};
```
bh2smith, line 7: Where is the  `as _` made use of here in this file?

bh2smith (approved):
Looks good, just a few comments and questions.

vkgnosis (approved):

solver/src/liquidity/uniswap.rs
fleupold: We are not actually passing or creating a PoolFetching around so we don't need to import the symbol. We do however use an implementer of PoolFetching and call methods that are part of the trait, which is why we need to import the trait (anonymously). Otherwise the compiler will complain that we cannot invoke a method that is part of the PoolFetching trait.

orderbook/src/fee.rs
fleupold: This was a rebase error, it's actually part of #292

orderbook/src/api/create_order.rs
fleupold: We don't have access to the sell token in this method. I changed the error text to be less specific and a format that also renders nicer in the UI (where you don't really want to call an API endpoint).

fleupold:
CI is blocked on #299

PR #300, opened 2021-02-10
Persistent MinFeeMeasurement Storage

fleupold:
Another flavor of our database struct which knows how to persist and load min fee measurements. Not that this requires updating the database schema when merging into staging.

### Test Plan
Unit test.

### DIFF
--- a/orderbook/src/database/fees.rs
+++ b/orderbook/src/database/fees.rs
@@ -5,75 +5,50 @@ use anyhow::{anyhow, Context, Result};
 use bigdecimal::BigDecimal;
 use chrono::{DateTime, Utc};
 use ethcontract::{H160, U256};
-use futures::StreamExt;
-
-#[derive(PartialEq, Debug, Clone, Copy)]
-pub struct MinFeeMeasurement {
-    token: H160,
-    expiry: DateTime<Utc>,
-    min_fee: U256,
-}
 
 impl Database {
-    pub async fn save_fee_measurement(&self, measurement: MinFeeMeasurement) -> Result<()> {
+    pub async fn save_fee_measurement(
+        &self,
+        token: H160,
+        expiry: DateTime<Utc>,
+        min_fee: U256,
+    ) -> Result<()> {
         const QUERY: &str =
             "INSERT INTO min_fee_measurements (token, expiration_timestamp, min_fee) VALUES ($1, $2, $3);";
         sqlx::query(QUERY)
-            .bind(measurement.token.as_bytes())
-            .bind(measurement.expiry)
-            .bind(u256_to_big_decimal(&measurement.min_fee))
+            .bind(token.as_bytes())
+            .bind(expiry)
+            .bind(u256_to_big_decimal(&min_fee))
             .execute(&self.pool)
             .await
             .context("insert MinFeeMeasurement failed")
             .map(|_| ())
     }
 
-    pub async fn load_fee_measurements(
+    pub async fn get_min_fee(
         &self,
         token: H160,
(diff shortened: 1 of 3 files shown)


database/schema.sql
```diff
@@ -44,6 +44,13 @@ CREATE TABLE invalidations (
     PRIMARY KEY (block_number, log_index)
 );
 
+-- MinFee Measurements from the API
+CREATE TABLE min_fee_measurements (
+  token bytea NOT NULL,
+  expiration_timestamp timestamptz NOT NULL,
+  min_fee numeric(78,0) NOT NULL
+);
```
vkgnosis, line 52: One thing I realize now is that the current design creates a new row every time queries the fee, which isn't great.
Maybe we could avoid this by checking whether the new fee is the same as the previous fee and in that case extending expiration_timestamp instead.
Or setting the expiration timestamp in our response to the same that is already there up to some limit. So the total time would be reduced for the later request. If our normal fee is supposed to last 5 minutes then we could say we will reuse the for 1 minute so users would see expiration time's between 4 and 5 minutes. In this case we could do this only if the fee is the same or even if the fee has changed because it probably won't change drastically.
Would probably be good to talk about this before merging with the new schema, in case the discussion would make it change.

orderbook/src/database/fees.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    pub async fn load_fee_measurements(
+        &self,
+        token: H160,
+        min_expiry: DateTime<Utc>,
+    ) -> Vec<MinFeeMeasurement> {
+        const QUERY: &str = "\
+            SELECT expiration_timestamp, min_fee FROM min_fee_measurements \
+            WHERE token = $1 AND expiration_timestamp >= $2
+            ";
```
vkgnosis, line 39: I think the way this used it should get the token and current time and return a single fee which is the minimum over all fees that are currently valid. That way no vector is returned.

orderbook/src/database/fees.rs
```diff
@@ -0,0 +1,136 @@
...
+            .filter_map(
+                |result: Result<MinFeeMeasurementQueryRow, _>| match result {
+                    Ok(row) => row.into_measurement(token).ok(),
+                    Err(err) => {
+                        tracing::error!(?err, "Fetching min fee from db");
+                        None
+                    }
+                },
+            )
+            .collect()
```
vkgnosis, line 59: Imo the function should return Result<Vec> and here we would collect the Iterator<Result<T>> into Result<Vec<T>>

orderbook/src/database/fees.rs
```diff
@@ -0,0 +1,136 @@
...
+use futures::StreamExt;
+
+#[derive(PartialEq, Debug, Clone, Copy)]
+pub struct MinFeeMeasurement {
+    token: H160,
+    expiry: DateTime<Utc>,
+    min_fee: U256,
+}
+
+impl Database {
```
vkgnosis, line 17: Missing function to delete expired measurements.

database/schema.sql
fleupold: I'd suggest the following strategy:

The timestamp we report to the frontend is always 1 minute from now. Internally, we persist new estimates with a validity of 2 minutes. When querying a min fee we check if we have an estimate in the DB that is still valid 1 minute from now, if true return it. Otherwise create and save a new estimate. This way we should only have one record per minute.

This logic can live entirely in the minFeeCalculator logic.

vkgnosis: Sounds good.

vkgnosis (approved):

PR #301, opened 2021-02-10
Use persistent storage

fleupold:
Making use of the persisted fee storage. For testing purposes I turned the store/load functions into a trade which can be implemented in memory (like today) or via going to disk. For the unit tests, we are using the in-memory components.

### Test Plan
Load frontend pointed to local orderbook. Swap tokens until you see the metamask signature popup (don't sign yet). Restart orderbook. Sign order and see that it doesn't get rejected.


vkgnosis (approved):

bh2smith:
How does one load the front end pointed at local instance? Sounds fancy and useful.

fleupold:
@bh2smith by using `http://localhost:8080/api/v1` [here](https://github.com/gnosis/gp-v2-swap/blob/c53137bed99a9037800b67dd4468a4e4b2d4435b/src/custom/utils/operator.ts#L11-L13). Maybe we should add those instructions to their README.

PR #302, opened 2021-02-11
Http solver fixes after testing with updated real solver

vkgnosis:
Sorry for combining several small changes into one PR. This was all motivated by testing with the real solver and several of these changes are just a couple of lines so I didn't want to have the overhead of creating that many PRs.

* introduce a PreparedModel struct for combing the model sent to the
solver with the data it references.
* get rid of references in the model data to remove lifetimes
* use that struct to simplify `solve` so that the real solver test can
skip the `convert_settlement` step so that it does not have to set up
mocks on the order handling.
* fix two fields that were changed in the solver api
* re-add trace log because it was useful during testing
* use real amm fees

### Test Plan
Test with real solver works and prints response

### DIFF
--- a/solver/src/http_solver.rs
+++ b/solver/src/http_solver.rs
@@ -1,7 +1,7 @@
 mod model;
 mod settlement;
 
-use self::model::*;
+use self::{model::*, settlement::PreparedModel};
 use crate::{
     liquidity::{AmmOrder, LimitOrder, Liquidity},
     settlement::Settlement,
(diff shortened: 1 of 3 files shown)


solver/src/http_solver/settlement.rs
```diff
@@ -8,22 +8,29 @@ use model::order::OrderKind;
 use primitive_types::{H160, U256};
 use std::collections::HashMap;
 
+// To send an instance to the solver we need to identify tokens and orders through strings. This
+// struct combines the created model and a mapping of those identifiers to their original value.
+pub struct PreparedModel {
```
fleupold, line 13: nit: Does `model` need to be a part of this? It doesn't seem required for `convert_settlement`. Maybe `prepare_model` can return a pair of `BatchAuctionModel` and `SettlementContext`, the latter consisting only of the three mappings.

solver/src/http_solver.rs
```diff
@@ -127,39 +120,50 @@ impl HttpSolver {
...
     }
 
+    fn prepare_model(&self, liquidity: Vec<Liquidity>) -> PreparedModel {
+        let tokens = self.tokens(liquidity.as_slice());
+        let orders = split_liquidity(liquidity);
+        let limit_orders = self.orders(orders.0);
+        let amm_orders = self.amms(orders.1);
+        let model = BatchAuctionModel {
+            tokens: self.token_models(&tokens),
+            orders: self.order_models(&limit_orders),
```
fleupold, line 155: It's a bit confusing to me to have `self.orders`, `self.order_models`, etc. (their separation is not really clear). 

Could we make the function names more explicit (e.g. order_list_to_map) and do they even have to be on self? Maybe we can even use the Into trait to convert `HashMap<String, AmmOrder>` into `HashMap<String, UniswapModel>`?

fleupold (approved):
Less efficient than <String, Token/AmmOrder/LimitOrder> maps but maybe more readable could be if we allowed to find an element by String directly on `Vec<Liquidity>`

solver/src/http_solver.rs
vkgnosis: I've changed the function names from `self.orders` to `self.map_orders_for_solver` and added another explanatory comment to `prepare_model` where these functions are used. I feel that is to make sense of this.
They don't have to be on self now but this will change when we want to query (and cache) the token decimals and name. When it is clear how that works out with self we can move what still doesn't need self.

PR #304, opened 2021-02-11
Refactor solver result to settlement conversion

vkgnosis:
While working on finding the correct ordering of uniswap uses we get
from the solver I found that this code is a lot easier to work with if
we first convert it to something more convenient to use than the raw
data we get from the solver.
We put error handling and association of the string identifiers up front
before doing more work with the data.
I also changed the implementation of new_prices to
be more efficient by only going through all orders once instead of once
for every price.

Test Plan:
existing test. but could make more fine grained test in the future as this has become easier with this refactor.

### DIFF
--- a/solver/src/http_solver/settlement.rs
+++ b/solver/src/http_solver/settlement.rs
@@ -1,4 +1,4 @@
-use super::model::{Price, SettledBatchAuctionModel};
+use super::model::*;
 use crate::{
     liquidity::{AmmOrder, LimitOrder},
     settlement::Settlement,
@@ -6,7 +6,10 @@ use crate::{
 use anyhow::{anyhow, ensure, Result};
 use model::order::OrderKind;
 use primitive_types::{H160, U256};
-use std::collections::HashMap;
+use std::{
+    collections::{hash_map::Entry, HashMap},
+    iter,
+};
 
 // To send an instance to the solver we need to identify tokens and orders through strings. This
 // struct combines the created model and a mapping of those identifiers to their original value.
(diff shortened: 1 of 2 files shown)


solver/src/http_solver/settlement.rs
```diff
@@ -18,92 +21,172 @@ pub struct PreparedModel {
...
+            settled.prices,
+        )?;
+        Ok(Self {
+            executed_limit_orders,
+            executed_amms,
+            prices,
+        })
+    }
+
+    fn into_settlement(self) -> Settlement {
```
fleupold, line 77: Why not use Into trait?

solver/src/http_solver/settlement.rs
```diff
@@ -18,92 +21,172 @@ pub struct PreparedModel {
...
+        let mut settlement = Settlement::default();
```
fleupold, line 78: nit: I'd slightly prefer explicitly constructing the Settlement here instead of using default. This gives us more compiler safety if we ever add a field to the settlement.

solver/src/http_solver/settlement.rs
```diff
@@ -18,92 +21,172 @@ pub struct PreparedModel {
...
-    amms: &HashMap<String, AmmOrder>,
-    settlement: &mut Settlement,
-) -> Result<()> {
-    for (index, model) in model.uniswaps.iter() {
-        let amm = amms
-            .get(index.as_str())
-            .ok_or_else(|| anyhow!("invalid amm {}", index))?;
-        let (input, output) =
-            if model.balance_update1.is_positive() && model.balance_update2.is_negative() {
+fn new_orders(
```
fleupold, line 98: nit: I find this method name very vague. Maybe something like `match_prepared_and_settled_orders`?

fleupold (approved):

solver/src/http_solver/settlement.rs
vkgnosis: This has come up a couple of times. The reason for having Into and From is to allow other code to be generic about conversions. So instead of taking `Url` you can take `into Url` which is implemented for `String`. This makes sense for public types to make them more ergonomic to use in high level apis.
For a private type that has only one specific purpose like this imo it does not and implementing Into only adds boilerplate.

PR #306, opened 2021-02-12
Query smaller event range when possible

### LINKED ISSUE #305: Event Updater uses large block range
Cf. Kibana logs: https://logs.gnosisdev.com/goto/aa165cdbdba01c8750991118c26cba58

> orderbook::event_updater: updating events in block range 8051255..=8056117

Note that the range is roughly 5000 blocks. We never actually tighten it to e.g. only the last 25 blocks. It's possible that this is causing some strain on our nodes (we have seen an increased amount of alerts in staging recently and I noticed rinkeby alerts popping up when I just restarted the rinkeby API after a bit of downtime as the nodes seem to have rescaled). 

Same seems to be true for xDAI and mainnet.

I think this might be related to the fact that we only store the `last_handled_block` in storage if there is activity and thus in times of little activity accrue a very large range.

vkgnosis:
Instead of using only the most recent event block from the db we also store the last
handled block in self so that during long times of no events we do not query needlessly
large block ranges.

related #305 

### Test Plan
Run locally and observe block range is small after initial update

### DIFF
--- a/orderbook/src/event_updater.rs
+++ b/orderbook/src/event_updater.rs
@@ -21,15 +21,20 @@ const INSERT_EVENT_BATCH_SIZE: usize = 250;
 pub struct EventUpdater {
     contract: GPv2Settlement,
     db: Database,
+    last_handled_block: Option<u64>,
 }
 
 impl EventUpdater {
     pub fn new(contract: GPv2Settlement, db: Database) -> Self {
-        Self { contract, db }
+        Self {
+            contract,
+            db,
+            last_handled_block: None,
+        }
     }
 
     /// Get new events from the contract and insert them into the database.
-    pub async fn update_events(&self) -> Result<()> {
+    pub async fn update_events(&mut self) -> Result<()> {
         let range = self.event_block_range().await?;
         tracing::debug!("updating events in block range {:?}", range);
         let events = self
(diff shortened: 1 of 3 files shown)


orderbook/src/event_updater.rs
```diff
@@ -113,7 +125,7 @@ impl EventUpdater {
             .all_events()
             .from_block((*block_range.start()).into())
             .to_block((*block_range.end()).into())
-            .block_page_size(1000)
+            .block_page_size(500)
```
vkgnosis, line 128: Decreased this since Felix mentioned our event updating might be loading nodes a lot

orderbook/src/event_updater.rs
```diff
@@ -21,15 +21,20 @@ const INSERT_EVENT_BATCH_SIZE: usize = 250;
 pub struct EventUpdater {
     contract: GPv2Settlement,
     db: Database,
+    last_handled_block: Option<u64>,
```
fleupold, line 24: nit: could be an [AtomicU64](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html) to avoid the mutex on the whole EventUpdater.

fleupold (approved):

orderbook/src/event_updater.rs
vkgnosis: It could be, on the other hand I don't think it ever makes sense to call update events in parallel.

PR #308, opened 2021-02-12
Trade API Specification

bh2smith:
Introducing and inviting discussion to the Trade API.

So far I have only included the openapi documentation to start discussion on what we expect to return from a trade query. At this moment, the description states that trades can be queried by `owner` or `order_uid` which seems reasonable as a good starting point.

I noticed in the API doc that we also have OrderMetaData which contains a lot of information pertaining to trade data. Is this something we plan to keep along side the get_trade functionality?


orderbook/openapi.yml
```diff
@@ -153,8 +153,34 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/FeeInformation"
-        400:
+        404:
```
bh2smith, line 156: @fleupold - I wonder if the 400 here was intentional or was meant to be a 404 like the others.

orderbook/openapi.yml
```diff
@@ -153,8 +153,34 @@ paths:
...
+          required: false
+      responses:
+        200:
+          description: all trades
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Trade"
+        404:
+          description: No trades found.
```
vkgnosis, line 183: I think the return type should be array of Trade like for the orders route and we don't need 404 because if there are no trades then the result is an empty array.

orderbook/openapi.yml
```diff
@@ -265,6 +291,29 @@ components:
...
+          $ref: "#/components/schemas/UID"
+        sellAmount:
+          description: "Amount sold in this trade"
+          $ref: "#/components/schemas/BigUint"
+        buyAmount:
+          description: "Amount bought in this trade"
+          $ref: "#/components/schemas/BigUint"
+        feeAmount:
+          description: "Fee paid (in sell token)"
+          $ref: "#/components/schemas/BigUint"
```
vkgnosis, line 316: For these the type can be `TokenAmount` instead of `BigUint` because they are limited to `U256`.

orderbook/openapi.yml
```diff
@@ -265,6 +291,29 @@ components:
       allOf:
         - $ref: "#/components/schemas/OrderCreation"
         - $ref: "#/components/schemas/OrderMetaData"
+    Trade:
+      description: |
+        Trade data such as executed amounts, fees, order id and block number.
+      type: object
+      properties:
```
anxolin, line 298: Could we return also the hash of the
- tx (the settlement tx hash) --> people will be able to use it to check in etherscan
- Can we return batch Id: this batch id should be enough to identify a batch, and allow us to query info about it in future endpoints
- executionTime  ---> at what time this trade was executed
-

orderbook/openapi.yml
```diff
@@ -265,6 +291,29 @@ components:
...
+        blockNumber:
+          description: "Block in which trade occurred"
+          type: integer
+        logIndex:
+          description: "Index in which transaction was included in block"
+          type: integer
+        orderUid:
+          description: "Unique ID of the order matched by this trade"
+          $ref: "#/components/schemas/UID"
+        sellAmount:
```
anxolin, line 308: Do you plan to return also the sellAmountAtomsBeforeFees like in the orders?

orderbook/openapi.yml
```diff
@@ -265,6 +291,29 @@ components:
...
+        Trade data such as executed amounts, fees, order id and block number.
+      type: object
+      properties:
+        blockNumber:
+          description: "Block in which trade occurred"
+          type: integer
+        logIndex:
+          description: "Index in which transaction was included in block"
+          type: integer
+        orderUid:
```
anxolin, line 305: why orderUid and not order Id? 
people would expect ids to be unique anyways....
normally u see just id everywhere

orderbook/openapi.yml
```diff
@@ -265,6 +291,29 @@ components:
...
+          description: "Unique ID of the order matched by this trade"
+          $ref: "#/components/schemas/UID"
+        sellAmount:
+          description: "Amount sold in this trade"
+          $ref: "#/components/schemas/BigUint"
+        buyAmount:
```
anxolin, line 311: should we specify they are in atoms?

orderbook/openapi.yml
```diff
@@ -153,8 +153,34 @@ paths:
...
+      description: |
+        By default all trades are returned.
+        Queries can be refined by specifiying owner or order_uid.
+      parameters:
+        - name: owner
+          in: query
+          schema:
+            $ref: "#/components/schemas/Address"
+          required: false
+        - name: order_uid
```
anxolin, line 170: why snake case? also I wouldn't say `Uid` --> should it be `orderId`

orderbook/openapi.yml
```diff
@@ -265,6 +291,29 @@ components:
...
+          type: integer
+        orderUid:
+          description: "Unique ID of the order matched by this trade"
+          $ref: "#/components/schemas/UID"
+        sellAmount:
+          description: "Amount sold in this trade"
+          $ref: "#/components/schemas/BigUint"
+        buyAmount:
+          description: "Amount bought in this trade"
+          $ref: "#/components/schemas/BigUint"
```
alfetopito, line 313: I'm wondering, we might want to de-normalize the Trade and add `buyToken` and `sellToken` in the response.
This will be needed in case we want to display trades without loading the respective orders. For example, a user's summary of trades.

Is this easily available on your side?

orderbook/openapi.yml
```diff
@@ -153,8 +153,34 @@ paths:
...
-        400:
+        404:
           description: sellToken not existent
+  /api/v1/trades/:
+    get:
+      summary: Get existing Trades.
+      description: |
+        By default all trades are returned.
+        Queries can be refined by specifiying owner or order_uid.
+      parameters:
```
anxolin, line 164: not need to implemet now, but I would expect in the future:
- token: Either sell or buy token. Allows to see trades with one token
- I'm not adding the param for market, cause that one may be sth like `/markets/<market>/trades` 
- sellToken
- buyToken
- batchId: Trades included in a batch
- appId: Trades with a given appId
- fromDate
- toDate

anxolin, line 164: I think the pagination parameters are missing. 

there will be many trades, so we will need to define some kind of paging params

anxolin (commented):
Nice you are taking this endpoint!

I added many comments, I guess some of the filter we will need it later on, this is why I added it already here.

The first filter we will need is `orderId`, probably next will be `owner`.

orderbook/openapi.yml
fleupold: Yes, good catch this should be a 404

orderbook/openapi.yml
fleupold: orderUid is taken from the smart contracts. If it's only for the unnecessary u in the name, I'd prefer to stick to the smart contract notation (less ambiguity what this can be used for).

orderbook/openapi.yml
fleupold: If we denote that here we should also change it in the orders API (where we have the cummulative amount). Maybe we can add unit as a parameter type later (not sure we want atoms and base unit amounts in the same response).

orderbook/openapi.yml
fleupold: Good point. We don't have a notion of batch ID yet, but it makes sense. I was thinking batch_id will likely be something like tx_hash + tx_index of the Settlement transaction

orderbook/openapi.yml
fleupold: We don't yet have it for orders either. Not sure if we should add it here or later.

fleupold (commented):
Nice proposal. Looking pretty good already!

orderbook/openapi.yml
bh2smith: Yes, this is surely something I would think makes sense, but it is also my impression that you will have already loaded orders anyway... Given you you have orders already, in say, a mapping of some sort, would it be too much overhead to simply "join" them on the fly? Although it might sound as if I am fighting to not include them I do still think its a good idea... but then we shouldn't also include owner as well...

orderbook/openapi.yml
bh2smith: Which ever you prefer... Note that `feeAmount` it already part of "Trade" so you have access all the necessary information. But I don't intend to include all three (i.e. `a,  b,  a+b`).

orderbook/openapi.yml
bh2smith: Note also that in the `Orders` table this is `UID` as the Primary Key and in the `Trades` as `orderUid` as a Foreign Key (that is, assuming my understanding of PKs and FKs are correct).

orderbook/openapi.yml
bh2smith: good catch, not snake_case but also not `OrderId` but rather `orderUid`

orderbook/openapi.yml
bh2smith: Created an issue #311 for these points. 

Note that the last one seems to be a request for orders... I assume you want to see trades for open and not open orders.

orderbook/openapi.yml
bh2smith: I have included `owner`, `sellToken` and `buyToken`

orderbook/openapi.yml
anxolin: okok

orderbook/openapi.yml
anxolin: I think this is something we know we'll need, so I don't see why not to add it.
We cannot expect to return all orders or all trades given a filter criteria of:
- owner
- appId
- dates
- ...
there will be too many results

orderbook/openapi.yml
anxolin: It wouldn't be for me :) 
I don't need it, it's just applying the same thing we did in the orders, there was confusion if sellAmount was with or without fees, so we thought that adding the denormalized would make the API easier to use.

Personally, I don't care about this one. Just flag in it.

orderbook/openapi.yml
anxolin: okokok, never mind, yes I don't like that U, but I can live with it. Scratch this

orderbook/openapi.yml
bh2smith: Yes it will be included ASAP, but not as per of this PR.

orderbook/openapi.yml
bh2smith: Me neither, but it stays.

orderbook/openapi.yml
bh2smith: The implementation at the moment returns `buy_amount`, `sell_amount` and `fee_amount`

orderbook/openapi.yml
bh2smith: There is an additional train of thought that supports `Uid` here and that is the fact that an individual user might want to see their order ID as a sequence of integers (i.e. first order, second order, etc..) while these, user specific, order IDs are not universally unique over all, they are still IDs from the owners perspective. Hence the distinction between Unique ID and ID

anxolin: I'm not sure I got this point
> while these, user specific, order IDs are not universally unique over all

You say "Universally" because we considered using UUID, but we went for a unique ID that is not universal and we want to specify this kind of things in it's name. 

Fine with me, as I say, it feels strange to me, I never see people using this naming convention. UUIDs are a special data type with it's RFC spec. Anyways, is not that important to me, I just wanted to mention it

orderbook/openapi.yml
anxolin: okok 👌

orderbook/openapi.yml
fleupold: To make it align with the order endpoint, I would say we should return sell_amount_before_fees as well here (the redundancy was deemed ok in order to have more clarity for the consumer of the API without having to read the detailed documentation of the endpoint)

orderbook/openapi.yml
fleupold: As discussed in the meeting yesterday we can compute a TradeUID base on tx hash + log index of the trade event which we should also return here. Can you add these values?

fleupold (approved):
LGTM

orderbook/openapi.yml
bh2smith: Are you suggesting that the field `sellAmount` be renamed to `sellAmountBeforeFees` here? Or would it suffice to just include this as part of the field description? I would prefer to keep the name as small as possible if everyone is in favour.

fleupold: I would like to have both fields: sellAmount AND sellAmountBeforeFees.

Basically, the same way we have it right now in Order: https://github.com/gnosis/oba-services/blob/997a72beb01261b1d7c67b6ce40429b0b9d6e8d8/orderbook/openapi.yml#L249-L254

bh2smith: Gotcha, so then I shouldn't bother with fee amount.

orderbook/openapi.yml
bh2smith: I included `batchId` as `blockNumber * 10^4 + log_index` because the maximum number of transactions that could ever fit into a single block is about 594. This is computed as `blockGasLimit / minimumGasForTx`

```
12500000/21000 ~ 595
```

I hope this is alright for everyone. @fleupold

bh2smith: This also allows room for the block gas limit to go up to 21M

orderbook/openapi.yml
bh2smith: Done. Check it out!

orderbook/openapi.yml
vkgnosis: I don't think it has to be packed this way because the type doesn't have to be integer. If you use hex encoded bytes it could be block number concatenated with log index. But then as Felix said we probably want transaction hash instead of block number to make it easier to look up the tx later. I'm also fine with figuring out and discussing this in another issue/PR .

vkgnosis (approved):

orderbook/openapi.yml
bh2smith: Yes, we are certainly going to need transaction hash since log_index is not what I thought it was. This will be better for the explorer since they can provide a link to etherscan for the tx

bh2smith: I think this should be left out for now and included in a later feature branch. There is already enough going on here and it would be an easy fix afterwards.

bh2smith: For some reason I was confusing it with transaction index which would have given us a nice Integral `batchId`. Maybe we should call it `batchTx` when we implement it.

bh2smith: Note that I have included this as a TODO (after the fact) in the Trade API issue #308

orderbook/openapi.yml
fleupold: @bh2smith why did you remove fee amount? For orders where we have:

- executedSellAmount
- executedSellAmountBeforeFees
- executedFeeAmount

The redundancy was introduced on purpose to make it easier for the consumer of this API to understand what each value implies. Could you add the fee amount back please?

orderbook/openapi.yml
fleupold: Ok to do it later, and in general we don't have to inform in the API docs what the semantics of a unique ID is (we do it for orderUid because it can be used in the contracts directly). As far as the user of this API is concerned it's some opaque string they can use to uniquely identify a trade,

PR #313, opened 2021-02-15
Add init db script

fleupold:
This PR add an init_db script that can be called to set up the database either on an empty state or to incrementally update it from a previous state (e.g. when we add a new table/column etc.)

The script + all other required db information is packaged in the docker container we create on every build so that it can be called as an init container from within our kubernetes setting.

### Test Plan
Run `/usr/database/init_db.sh postgresql://localhost` both on a cold and warm database.


database/schema.sql
```diff
@@ -4,10 +4,17 @@
 -- fixed size.
 -- `u32`s are stored in `bigint` which is an 8 bytes signed integer because Postgre does not have
 -- unsigned integers.
+BEGIN;
 
-CREATE TYPE OrderKind AS ENUM ('buy', 'sell');
+DO $$
+BEGIN
```
vkgnosis, line 10: Why is this `DO $$` needed? The reference I found is https://www.postgresql.org/docs/current/sql-do.html but not sure what is going on here?

vkgnosis (approved):
Looks good for now although I expect this will change for incompatible schema changes. But we can tackle that when we get there.

database/schema.sql
fleupold: It's [dollar quoting](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING) to avoid conflict with any quotes that happen inside the query.

fleupold:
Given @anxolin's suggestion we can look into [flyway](https://flywaydb.org/) right away to make future migrations easier.

fleupold:
Re-requesting review as this PR changed quite a bit.

bh2smith (approved):
Cute!

anxolin (commented):
Yes, I would user a migration tool, so you can do the migration files, the first could be this create schema. It can facilitate deploying new versions without sweating too much. Flyway was pretty straight fw if I remember well, there was a CLI and a bunch of files, it works very similarly to truffle migrations.  

Unrelated to this PR, but some thoughts I had reading the schema:
- there's a few indexes missing: like sell token
- The FKs are missing, like from trade to its order
- Ideally we would save already the denormalized short id 
- I would consider adding a FK in the trades to the settlement, I would make sure there is a tx hash (that would be important for us in the FE), and I would consider to move those "block_number". Maybe you want to even use this FK+log_index as the PK.


I wouldn't do this one, until we see we have performance issues, but, one thought I had is, If we expect issues with having many orders, you can set a flag for the orders that are final:
- expired
- canceled
- fully executed

For these order you know u'll never need them again (for the ones considered by the solver).
If we have too many orders, an index in this field would basically allow us to quickly filter out most of the orders. As I said, this one, better to wait for the real need

vkgnosis (approved):
> Re-requesting review as this PR changed quite a bit.

I don't think anything changed.

fleupold:
> I don't think anything changed.

I forgot to push 🙈

fleupold:
Regarding @anxolin feedback (not related to this PR so maybe better to have on an Issue?)

> The FKs are missing, like from trade to its order

I believe the reason for no imposing a FK is that we could potentially see a trade for which we didn't record the order data (in case of data loss, as trades come from on-chain). Right now we still want to store those in the database.

> Ideally we would save already the denormalized short id

Can you elaborate on that? Shouldn't the database be usually be normalized if possible?

> and I would consider to move those "block_number".

I don't understand what you mean by this. Where do you want to move block_numbers to?

docker/Dockerfile.binary
```diff
@@ -9,9 +9,11 @@ RUN cargo build --target x86_64-unknown-linux-musl --release
 FROM alpine:latest
 
 # Handle signal handlers properly
-RUN apk add --no-cache tini
+RUN apk add --no-cache tini postgresql-client
 COPY --from=cargo-build /usr/src/oba-services/target/x86_64-unknown-linux-musl/release/orderbook /usr/local/bin/orderbook
 COPY --from=cargo-build /usr/src/oba-services/target/x86_64-unknown-linux-musl/release/solver /usr/local/bin/solver
 
+COPY database usr/database
```
vkgnosis, line 16: This is probably no longer needed with the move to flyway in a separate container.

README.md
```diff
@@ -65,11 +64,12 @@ postgres -D data
 createdb -h localhost $USER
 
 # Finally for all methods to test that the server is reachable and to set the schema for the tests.
-psql -h localhost -f database/schema.sql
+docker build --tag gp-v2-migrations -f docker/Dockerfile.migration .
+# If you are running postgres in locally, your URL is `localhost` instead of `host.docker.internal`
+docker run -ti -e FLYWAY_URL="jdbc:postgresql://host.docker.internal/?user="$USER"&password=" -v $PWD/database/sql:/flyway/sql gp-v2-migrations migrate
```
vkgnosis, line 69: I am a little sad that running postgres tests locally now requires building and running a docker container. It is probably worth it over making our own migration system though. One way to make this simpler would be to still have a full schema.sql but then we would have to duplicate the information. Or maybe we can make a script that manually applies all the migration sql files. Anyway, this is fine for this PR, just thinking out loud.

vkgnosis (approved):

anxolin:
@fleupold I'll remove the feedback from this PR since it's unrelated. I mentioned it since it was the first PR I see regarding the database structure and some things came to mind. I didn't review the schema in detail, those were just a few things that I notice.
Happy to open a discussion issue if that the prefered way.

.github/workflows/pull-request.yaml
```diff
@@ -29,11 +29,12 @@ jobs:
       - uses: actions/checkout@v2
       - uses: Swatinem/rust-cache@v1
       - run: cargo test --locked --all-features
+      - run: curl -L https://repo1.maven.org/maven2/org/flywaydb/flyway-commandline/7.3.1/flyway-commandline-7.3.1.tar.gz -o flyway-commandline-7.3.1.tar.gz && tar -xzf flyway-commandline-7.3.1.tar.gz --strip-components=1 && rm flyway-commandline-7.3.1.tar.gz
```
anxolin, line 32: nit: Maybe u want to use the cache action of GH?

.github/workflows/pull-request.yaml
```diff
@@ -29,11 +29,12 @@ jobs:
...
       - run: |
           sudo systemctl start postgresql.service
           sudo -u postgres createuser $USER
           sudo -u postgres createdb $USER
-          psql -f database/schema.sql
+          flyway-commandline-7.3.1/flyway -url="jdbc:postgresql://localhost/?user="$USER"&password=" -locations="filesystem:database/sql/" migrate
```
anxolin, line 37: ubernit: the flyway 7.3.1 is repeated in at least 5 places, maybe an env var could help u change versions easily

README.md
anxolin: I guess u can have alternative instructions for doing it without docker

https://flywaydb.org/documentation/usage/commandline/#download-and-installation

Like using:

```bash
wget -qO- https://repo1.maven.org/maven2/org/flywaydb/flyway-commandline/7.5.3/flyway-commandline-7.5.3-linux-x64.tar.gz | tar xvz && sudo ln -s `pwd`/flyway-7.5.3/flyway /usr/local/bin 
```

And then:
https://flywaydb.org/documentation/getstarted/firststeps/commandline

database/sql/V002__create_trades.sql
```diff
@@ -0,0 +1,14 @@
+-- Trade events from the smart contract.
```
anxolin, line 1: I guess you split it into files just to try out the migrations

I think is fine, but also there was nothing wrong with having all the initial database in one file, and u add other files when u need to do the actual migrations.

anxolin (approved):
Cool stuff! very productive for a day u were feeling sick 😷

bh2smith:
> I would make sure there is a tx hash

@anxolin and @fleupold - I came across the same issue yesterday when I realized that `log_index` was not `transaction_index` and concluded that txHash would be important to keep in the trades table. It was my plan for today to make a separate PR for this. However, I think it would be nice to merge the existing Trade API as is before hand.

Once we have `txHash`, we will have a notion of `batchTx` per trade. I will get started on this right away if it wasn't intended to include here.

bh2smith (approved):

fleupold:
Ok, this was quite a battle, but it seems to be working now ([tmate](https://github.com/marketplace/actions/debugging-with-tmate) for the win). The crux was that flydb connects via TCP/IP even if we only want to connect to localhost and by default the access mode in postgres for network connections is md5 (which requires to actually know the user's password). psql connects via Unix sockets and it's auth mode is peer (which means it just checks if the user is the current user without caring for a password). 
I fixed this by manually setting a password for our test user.

I also added caching of the downloaded file to reduce the time we need to build.

Issue #297, opened 2021-02-10
Better handle increased scope of the order book api


vkgnosis:
I am a little worried about the scope of the backend order book api. Imo there are are three broad categories of apis:

1. Api needed for the system to run. This is roughly what we have implemented right now. It allows users to submit orders, delete orders, get the minimum fee and solvers to get all solvable orders. It operates on low level data that matches how the smart contract works.
2. Api that makes the order book easier to use for people who call it directly (including the frontend). Things like getting all of the orders of a specific user including unsolvable orders (for example because filled), getting denormalized data like a more abstract order state, getting executed trades for an order.
3. Api that implements functionality specific only to our frontend. Things like top traded tokens or other statistics that we want to show in our frontend but that we don't expect other people to use directly. The border to 2. is fuzzy.

The reason I am making this distinction is because I feel that right now we are mixing all three categories together and before we implement further apis we should consider if we want to separate them in some way. The drawbacks I see:
* Individual routes become more awkward as they have to handle more use cases and creates dependencies between unrelated things. For example `get_orders` from the perspective of the solvers does not need a user filter and `OrderMetaData` does not need to contain a `executedSellAmountBeforeFees` field. But of those are useful for the frontend.
* Development is harder. When we change a route we have to consider how this affects all categories instead of just one.
* The deployment is less resilient. If we have a bug or a panic or in for example complicated web socket code for the frontend then that shouldn't bring down the category 1 api.

What I would like to do is at least separate category 1 from the rest on a route and code level. We could have a paths like (ignoring bad naming) `/api/minimal/v1`, `/api/user-friendly/v1` and implement them in different rust modules. This gets rid of the api level entanglement. It will make it easier to move to separate crates / binaries in the future which has more organizational overhead so we might not want to that right now.
I hope that from the perspective of the frontend this would increase our development velocity and potentially allow frontend team members to implement some apis directly in rust themself without probably waiting (probably frustratingly long) for us to discuss and implement the suggestions.

vkgnosis:
Looking for feedback on this @gnosis/gp-services  .

bh2smith:
Very well said. I can easily get behind this as I have experience DB concurrency issues in the past for similar reasons. Of course the separate crates is also a very good idea.

fleupold:
[...]

PR #314, opened 2021-02-15
Trade API Implementation

### LINKED ISSUE #317: Trade API
Would like an API endpoint for querying trades from the database. This task can be split into three parts:

1. API Specification (#308)
2. Database Preparation (#316)
3. Exposing the API endpoint (#314)

- [ ] Include transaction hash in the trade table - #334

bh2smith:
Built on top of #316, which is built on #308 this PR closes #317. 

First implementation of a trade API (searchable by `owner` and `order_id`). We introduce a Trade object consisting of a semi joined (enriched version) of `DbTrade` with `Orders` that includes all the columns of `DbTrade` along with `owner`, `sellToken` and `buyToken` from the `Orders` table.

Since it doesn't really make sense to query both owner and orderUid at the same time, we have introduced a TradeFilter Validation that occurs before the select query is executed.

### Test Plan
Tests introduced here are 

- get_trades_request_ok: All scenarios where TradeFilter is ok
- get_trades_request_err: When trade filter is "invalid"
- get_trades_response_ok: When Response is Ok (always)


fleupold:
Can you split this PR up into smaller chunks? Starting maybe with the DB part and then working your way up to the consumer facing API?

I think we also want to incorporate some of the thoughts of #297 and would have to split some of the existing logic into a core submodule before adding the trade filter & handler in the same module.

bh2smith:
Thanks for your review @fleupold. My plan now for PR separation is to split this into two, making a total of three as follows:


1. Define Trade API - #308
    orderbook/openapi.yml

2. Files relating to the Trade structs and Database (PR doesn't yet exist)
    model/src/lib.rs
    model/src/order.rs
    model/src/trade.rs

    orderbook/src/database.rs
    orderbook/src/database/events.rs
    orderbook/src/database/orders.rs
    orderbook/src/database/trades.rs

    orderbook/src/integer_conversions.rs → orderbook/src/conversions.rs

3. API Implementation (This PR would become smaller):
 introducing trade filter and http requests
 
    orderbook/src/api.rs
    orderbook/src/api/get_trades.rs
    
    
  As for the logical splitting as suggested in #297 - I would think that this would pertain mostly to this PR in which some of the files mind wind up in different places (although I might reach out for some advice regarding these new modules etc...)

vkgnosis:
> As for the logical splitting as suggested in #297 - I would think that this would pertain mostly to this PR in which some of the files mind wind up in different places (although I might reach out for some advice regarding these new modules etc...)

In order to not block you too much feel free to ignore #297 for this change. I will probably make some PRs for that issue to move code around including get_trades. Or if my change is merged first then you would see this when rebasing.

orderbook/src/api/get_trades.rs
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+}
+
+pub fn get_trades_request() -> impl Filter<Extract = (TradeFilter,), Error = Rejection> + Clone {
+    warp::path!("trades")
+        .and(warp::get())
+        .and(warp::query::<Query>())
+        .map(|query: Query| query.trade_filter())
+}
```
bh2smith, line 34: Include request validation here that not both `owner` and `orderUid` are set. @fleupold I think this would satisfy the conversation we're having (here)[https://github.com/gnosis/gp-v2-services/pull/316#discussion_r580190577]

e2e/tests/uniswap_trade_test.rs
```diff
@@ -130,7 +130,7 @@ async fn test_with_ganache() {
             .await
             .expect("Couldn't query domain separator"),
     );
-    let db = Database::new("postgresql://").unwrap();
+    let db = Arc::new(Database::new("postgresql://").unwrap());
```
vkgnosis, line 133: We can clone and pass around `Db` directly without wrapping it in `Arc` because it is using that internally already.

orderbook/src/api/get_trades.rs
```diff
@@ -1,9 +1,150 @@
...
+    }
+
+    fn validate(&self) -> Result<TradeFilter, TradeFilterError> {
+        // Ensure that not both owner and order_uid are specified
+        match (self.order_uid.as_ref(), self.owner.as_ref()) {
+            (Some(_), Some(_)) => Err(TradeFilterError::InvalidFilter(
+                "Cannot specify both owner and order_uid".to_owned(),
+            )),
+            _ => Ok(self.trade_filter()),
+        }
```
vkgnosis, line 42: Maybe slightly cleaner is something like
```rust
if self.order_uid.is_some() && self.owner.is_some() {
    return Err(...)
}
Ok(...)
```

vkgnosis (approved):

e2e/tests/uniswap_trade_test.rs
bh2smith: I will give this a try again, but the first time I tried this there was something the compiler didn't like.

orderbook/src/api/get_trades.rs
bh2smith: Right yes, I was initially thinking this would leave room for other possibilities but this can easily be adapted in the future. Will use your suggestion.

e2e/tests/uniswap_trade_test.rs
bh2smith: This is going to cause a chain of reverts... I had to update the Event Updater to take Arc<Database> and this is why this unit test changed. Are you sure about this?

bh2smith: Or do you mean that I should just change the unit test to use something like `Arc::from(db.clone())` everywhere?

vkgnosis: I mean that everything that you made store Arc<Db> could be just Db and we clone the Db directly. Don't see why it would cause an error because Db is Clone.

bh2smith: Its all good now.

fleupold (approved):
Looks good. 

I agree with @vkgnosis that searching for `database: Arc<Database>` in this change set should not yield any results (the update doesn't yet seem to be pushed).

README.md
```diff
@@ -38,7 +38,7 @@ Some unit tests and the e2e tests require a local postgres setup. For some ways
 
 Run postgres unit tests: `cargo test --jobs 1 postgres -- --ignored --test-threads 1`
 
-Run e2e tests: `cargo test -p e2e -- --test-threads 1`.
+Run e2e tests: `xcargo test -p e2e -- --test-threads 1`.
```
bh2smith, line 41:
```suggestion
Run e2e tests: `cargo test -p e2e -- --test-threads 1`.
```

orderbook/src/main.rs
```diff
@@ -100,14 +101,19 @@ async fn main() {
...
         Box::new(balance_fetcher),
         fee_calculator.clone(),
     ));
     check_database_connection(orderbook.as_ref()).await;
 
-    let serve_task = serve_task(orderbook.clone(), fee_calculator, args.bind_address);
+    let serve_task = serve_task(
+        database.clone(),
+        orderbook.clone(),
+        fee_calcuator,
```
bh2smith, line 114:
```suggestion
        fee_calculator,
```

fleupold:
> I agree with @vkgnosis that searching for database: Arc<Database> in this change set should not yield any results (the update doesn't yet seem to be pushed).

@bh2smith I saw you 👍  this suggestion, however if I search for `database: Arc<Database>` in this now merged changeset I still get 4 results. Do you know how this happened?

bh2smith:
> Do you know how this happened?

Apparently, I misunderstood what was expected here.

PR #315, opened 2021-02-16
Add new route for solvable orders

vkgnosis:
As discussed this is a route that is only intended to be used by
solvers. For now it returns the same order information as get_orders. In
the future we can change this without breaking the frontend.

### Test Plan
e2e test

### DIFF
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -181,6 +181,21 @@ paths:
                 type: array
                 items:
                   $ref: "#/components/schemas/Trade"
+  /api/v1/solvable_orders:
+    get:
+      summary: Get solvable orders.
+      description: |
+        The set of orders that solvers should be solving right now. These orders are determined to
+        be valid at the time of the request.
+      responses:
+        200:
+          description: the orders
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Order"
 components:
   schemas:
     Address:
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -9,6 +9,7 @@ use model::{
     order::{Order, OrderCreation, OrderUid},
     DomainSeparator,
 };
+use shared::time::now_in_epoch_seconds;
 use std::sync::Arc;
 use tokio::sync::Mutex;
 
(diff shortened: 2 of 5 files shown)


solver/src/orderbook.rs
```diff
@@ -16,7 +16,7 @@ impl OrderBookApi {
     }
 
     pub async fn get_orders(&self) -> reqwest::Result<Vec<Order>> {
-        const PATH: &str = "/api/v1/orders";
+        const PATH: &str = "/api/v1/solvable_orders";
```
bh2smith, line 19: Doesn't this change the request path for orders too? Shouldn't there just be a new method introduced here `get_solvable_orders` that introduces this new path and not overwrite the existing one?

bh2smith (approved):
This looks fine, but my understanding of get solvable orders is that it should return the equivalent of `get_orders` with `valid = true` filter. Is it really worth having another endpoint just for this?

orderbook/src/orderbook.rs
```diff
@@ -90,6 +91,17 @@ impl Orderbook {
         Ok(orders)
     }
 
+    pub async fn get_solvable_orders(&self) -> Result<Vec<Order>> {
+        let filter = OrderFilter {
+            min_valid_to: now_in_epoch_seconds(),
+            exclude_fully_executed: true,
+            exclude_invalidated: true,
+            exclude_insufficient_balance: true,
```
bh2smith, line 99: Right, I see its all three flags equal true. I guess this makes a bit more sense to have a new route for it.

solver/src/orderbook.rs
vkgnosis: This is the path we use to query orders in the solver so we ony need the new path.

orderbook/openapi.yml
```diff
@@ -155,6 +155,21 @@ paths:
                 $ref: "#/components/schemas/FeeInformation"
         400:
           description: sellToken not existent
+  /api/v1/solvable_orders:
```
fleupold, line 158: Should we somehow separate the core and extended api in the path? Maybe by `/core/v1/solvable_orders` or `api/v1/core/solvable_orders`?

fleupold (approved):
👍

orderbook/openapi.yml
vkgnosis: Not sure. If we keep everything in the same api docs then also feels like it doesn't need separate paths and these categories like "core" feel like something we do internally but doesn't have to show on the api.

Issue #311, opened 2021-02-13
[API] Exotic Trade Query Fiilters


bh2smith:
- token: Either sell or buy token. Allows to see trades with one token
- sellToken
- buyToken
- batchId: Trades included in a batch
- appId: Trades with a given appId
- fromDate
- toDate
- open: bool. Where
    - true gives trades for open orders. 
    - false: gives trades for orders that are EXPIRED, FILLED or CANCELED

anxolin:
Awesome! We can make nice things with these!

bh2smith:
When enriching the `TradeFilter` to include some of these exotic search parameters, it was proposed by @fleupold (in #316) to use the following pattern for `order_uid` and `owner` since we prohibit both from being specified in trade query:

```js
struct TradeFilter {
  from: DateTime<UTC>,
  to: DateTime<UTC>,
  inner: Option<Inner>
}

enum Inner{
  Owner(H160),
  OrderUid(OrderUid),
}
```

fleupold:
The same is true for our OrderFilter (there we can also filter by user and UID) so if we changed it, it would probably make sense to change in both places.

anxolin:
not sure i understand, but that `inner` thing sounds strange. I guess you tried to model is one filter or the other. U guys know best about Rust good patterns.

fleupold:
Ben didn't copy my comment asking to pardon the poor/lazy naming. 

The point is that part of the "filter" is mutually exclusive (it doesn't make sense to filter on an orderID and a user), so a good way to harden our type to enforce this is by using an enum. 
However, there are some fields that are additive to the mutually exclusive fields (e.g. "from" & "to"). So to model this we need a struct which contains an "inner" enum.

bh2smith:
The naming isn't quite settled yet, but the point is that only one of `owner` or `order_uid` can be specified since both wouldn't make sense, the "inner" is just an enum that is one or the other.

anxolin:
It's nice rust allows to do that. I can see this getting messier if you really want to capture the logic. At the end you'll have to validate the somewhere, and there'll probably be logics like, the date range cannot be longer than 1 year. If they specify this filter, it necessarily need this other, ...
It was curiosity only, if it comes to that, it's easy to go to another model. 

The part I don't get now, is why the date is not optional? If I filter by orderId, I shouldn't need dates, right?

bh2smith:
[...]

PR #316, opened 2021-02-16
Trade API: Database setup

bh2smith:
Partial fulfillment of #317 

This PR introduces `DbTrade` and `Trade` structs related to the Trade API as specified in #308.
We implement the trade query here and test that it works with both filter arguments.

### Test Plan
Apart from existing CI, postgres tests should also be run.


vkgnosis:
Since this change is multiple PRs you could try making the "target branch" here the previous PR branch that this refers. That way the diff looks clean and it's clear which commit comes from where because right now the opanpi change is repeated in here.

model/src/trade.rs
```diff
@@ -0,0 +1,62 @@
...
+use num_bigint::BigUint;
+use primitive_types::{H160, U256};
+use serde::{Deserialize, Serialize};
+// use serde_with::serde_as;
+use crate::order::OrderUid;
+
+#[derive(Eq, PartialEq, Clone, Debug, Deserialize, Serialize, Hash)]
+#[serde(rename_all = "camelCase")]
+pub struct DbTrade {
+    pub order_uid: OrderUid,
```
vkgnosis, line 12: I don't think this struct should live here because it is only relevant for the db module where it used to be defined.

bh2smith: Moved it back!

orderbook/src/database/trades.rs
```diff
@@ -0,0 +1,378 @@
...
+impl Database {
+    pub fn trades<'a>(&'a self, filter: &'a TradeFilter) -> impl Stream<Item = Result<Trade>> + 'a {
+        const QUERY: &str = "\
+            SELECT \
+                t.block_number, \
+                t.log_index, \
+                t.order_uid, \
+                t.buy_amount, \
+                t.sell_amount, \
+                t.sell_amount - t.fee_amount as sell_amount_before_fees,\
```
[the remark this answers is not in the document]
vkgnosis: Assuming that the smart contract is correct this can't happen though, right? Because sell_amount has already been calculated as fee_amount + traded_amount.

orderbook/src/database/trades.rs
```diff
@@ -0,0 +1,378 @@
...
+                o.owner, \
+                o.buy_token, \
+                o.sell_token \
+            FROM trades t \
+            JOIN orders o \
+            ON o.uid = t.order_uid \
+            WHERE \
+                ($1 IS NULL OR o.owner = $1) \
+            AND \
+                ($2 IS NULL OR o.uid = $2);";
```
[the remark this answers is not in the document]
vkgnosis: One order uid can still have several trades for partially fillable orders. From an api perspective it does allow you to filter for owner and uid even though that doesn't make sense. But if we had separate queries to eliminate the second where clause we would just end up duplicating a lot of the code.

orderbook/src/database/trades.rs
```diff
@@ -0,0 +1,378 @@
...
+            buy_token,
+            sell_token,
+        };
+        let order = Order {
+            order_meta_data: OrderMetaData {
+                owner,
+                uid,
+                executed_sell_amount_before_fees: BigUint::from(executed_sell_amount),
+                executed_fee_amount: BigUint::from(fee_amount),
+                executed_buy_amount: BigUint::from(executed_buy_amount),
```
vkgnosis, line 142: Don't think you need to set these because they are not part of the information that is stored in the db.

orderbook/src/conversions.rs
```diff
@@ -36,3 +36,10 @@ pub fn big_decimal_to_u256(big_decimal: &BigDecimal) -> Option<U256> {
     let big_int = big_decimal.to_bigint()?;
     bigint_to_u256(&big_int)
 }
+
+pub fn h160_from_vec(vec: Vec<u8>) -> Result<H160> {
```
fleupold, line 40: very nice!

orderbook/src/database/events.rs
```diff
@@ -15,18 +16,31 @@ pub struct EventIndex {
...
 #[derive(Debug)]
 pub enum Event {
-    Trade(Trade),
+    DbTrade(DbTrade),
     Invalidation(Invalidation),
 }
 
-#[derive(Debug, Default)]
-pub struct Trade {
+#[derive(Eq, PartialEq, Clone, Debug, Deserialize, Serialize, Hash)]
```
fleupold, line 23: Do we really need all these added derivations? It seems to me like this file should actually not be touched by this PR (execpt for the rename in the imports).

orderbook/src/database/trades.rs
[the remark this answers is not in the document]
fleupold: That's a good point. If we want to make the API tighter we could turn TradeFilter into an enum that can be either owner or order_uid and return BAD_REQUEST if the consumer of the API specifies both.

orderbook/src/database/trades.rs
```diff
@@ -0,0 +1,375 @@
...
+
+        let (owners, _order_ids, trades) = populate_dummy_trade_db(db.clone()).await;
+
+        assert_trades(
+            &db,
+            &TradeFilter {
+                owner: Some(owners[1]),
+                ..Default::default()
+            },
+            &[],
```
fleupold, line 313: I don't like about this test that the information how many order we should expect for owners[1] is located somewhere in the details of `populate_dummy_trade_db`. I think this test will be easier to maintain if the information that is required for the assert is "close" to it. 

I understand you don't want to copy the boilerplate code from the first test three times. I see two options:
1. Test the filter directly in the first test (using the actual owner and some random owner)
2. Split `populate_dummy_trade` in `add_trade(owner, orderId) -> Trade`. this way you have control over how often you call it with what inputs in the test case directly.

orderbook/src/orderbook.rs
```diff
@@ -90,6 +94,10 @@ impl Orderbook {
         Ok(orders)
     }
 
+    pub async fn get_trades(&self, filter: &TradeFilter) -> Result<Vec<Trade>> {
```
fleupold, line 97: Should this method be exposed on the orderbook? I feel like trades are conceptually different and would prefer a not to make this component more responsible than it needs to be? As of right now we could directly pass the db (probably some form of interface would make sense to limit the scope we pass around).

fleupold (commented):
Looking really good! I think we can remove some code from this PR, make the test case a little bit easier to maintain and should discuss what component will expose the trade fetching functionality to the API.

model/src/trade.rs
```diff
@@ -0,0 +1,40 @@
...
+    pub sell_amount: BigUint,
+    #[serde(with = "serde_with::rust::display_fromstr")]
+    pub sell_amount_before_fees: BigUint,
+    // ORDER DATA
+    pub owner: H160,
+    pub buy_token: H160,
+    pub sell_token: H160,
+}
+
+impl Default for Trade {
```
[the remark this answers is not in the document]
bh2smith: I'll give it a try! Good idea.

orderbook/src/database/events.rs
bh2smith: Yea possibly, this is remnants from having moved it and put it back. I'll check it out.

orderbook/src/database/trades.rs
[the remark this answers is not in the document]
bh2smith: I like this idea!

orderbook/src/database/trades.rs
bh2smith: Actually I'm only using populate dummy trade for the two tests that use the filter, but I could probably consolidate all three (or just the two for filters) tests into one and make the different assertions on the same data.

bh2smith:
@fleupold - thanks for your feedback, I will take your suggested changes into account ASAP. As for discussion on what to expose for the API, I think this might be more relevant to the other open PR that already implements this.

orderbook/src/database/trades.rs
bh2smith: I have consolidated all three tests into one that uses populate dummy trade db and makes all necessary assertions with and without filters. Looks a lot better.

orderbook/src/orderbook.rs
bh2smith: Sounds about right. Good eye. I will have to look a little closer at this to see what to do here.

orderbook/src/database/trades.rs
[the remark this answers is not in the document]
bh2smith: Ok, so I like this idea, but I can imagine that this Trade filter might be enriched at some later point (perhaps filtering for trades between dates) and at this point, the enum would become outdated. Is there any harm in leaving it as a struct for now and just including an assertion/constraint somewhere that ensures these two fields aren't both set?

orderbook/src/orderbook.rs
bh2smith: All done! Weird that I wasn't getting `warning(unused_code)` here before moving it out as a stand alone function...

bh2smith:
@fleupold and @vkgnosis - I believe all the comments and suggestions to have been accounted for and that this PR is in a ready state for final review and maybe even merge. The only outstanding suggestion I can see if that of `BAD_REQUEST` and using an enum for TradeFilter, which I have described above as "probably not the best idea considering we will add other parameters in the future". This outstanding suggestion has been forwarded to #314 where I propose we can do request validation inside the trade request for such a nonsense query.

orderbook/src/database/events.rs
```diff
@@ -15,12 +15,12 @@ pub struct EventIndex {
...
 #[derive(Debug)]
 pub enum Event {
-    Trade(Trade),
+    DbTrade(DbTrade),
     Invalidation(Invalidation),
 }
 
 #[derive(Debug, Default)]
-pub struct Trade {
+pub struct DbTrade {
```
vkgnosis, line 23: Nit: I feel this rename is unnecessary because we are  already in the database module. If the problem is that we already have `Trade` in `model` and you want to use both structs then you can solve that by importing with a different name like `use database::events::Trade as DbTrade`.

orderbook/src/database/trades.rs
```diff
@@ -0,0 +1,305 @@
...
+            &db,
+            &TradeFilter {
+                order_uid: Some(order_ids[2]),
+                ..Default::default()
+            },
+            &trades[1..],
+        )
+        .await;
+    }
+}
```
vkgnosis, line 305: I would like to see a short test for the case where for whatever reason we have a trade event but not matching order uid to make sure that the query still works and does not include the trade.

vkgnosis (approved):

model/src/trade.rs
```diff
@@ -0,0 +1,23 @@
+//! Contains the Trade type as described by the specification with serialization as described by the openapi documentation.
```
vkgnosis, line 1: Imo it would be good to have a real json serialization test like `deserialization_and_back()`  in `model/src/order.rs` so that we can be sure it matches what we say in openapi.

orderbook/src/database/events.rs
bh2smith: Yea cool, I can fix this. Great suggestion.

orderbook/src/database/trades.rs
bh2smith: I believe we have a test that explicitly returns nothing in the response a when filtering by valid order_uid if this is what you are suggesting.

vkgnosis: I mean that we have a trade event for uid X, but no corresponding order with the same uid. So what should happen is that during the join of trades and orders this trade gets skipped because there is no corresponding order.

model/src/trade.rs
bh2smith: Done! and nice catch, I had forgotten the switch to `camelCase`

orderbook/src/database/trades.rs
bh2smith: Interesting, we can test this but I would hope that this never happens!

bh2smith: Included!

orderbook/src/database/trades.rs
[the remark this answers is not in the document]
fleupold: In that case the filter could still become a struct containing the enum (pardon bad naming):

```rust 
struct TradeFilter {
  from: DateTime<UTC>,
  to: DateTime<UTC>,
  inner: Option<Inner>
}

enum Inner{
  Owner(H160),
  OrderUid(OrderUid),
}
```

fleupold (commented):
I'm fine not making the change to the enum (I just don't find the argument is very compelling). 

I'm a bit disappointed that you find the test case is factored better now (I might have not made my initial concern clear enough). Happy to discuss this in person if it helps but also not strictly blocking this PR.

orderbook/src/database/trades.rs
bh2smith: This is a very nice idea and I completely agree with the maintainability argument. Just before I rework this, I am assuming it is safe to say that the orders can be populated in advance via some method called `populate_dummy_orders`since this is really what I was trying to avoid doing over and over again.

bh2smith:
Thanks for your careful review @fleupold, your suggestion to go with this Filter struct in combination with an `enum` is quite nice and I would like to implement it at the same time we enrich the trade filter to include more juicy stuff. This is mostly because the following PR, already implements pre-validation for the filter as is and this initial Trade API is growing tiresome. I have included your proposed snippet in the Exotic Trade Filter Issue #311.

I will now fix the unit tests here.

orderbook/src/database/trades.rs
bh2smith: Actually never mind this populate dummy orders, I think we can also create and insert the order at the time of add trade in most of these tests.

fleupold (approved):
Thanks, I like the tests much better now!

PR #321, opened 2021-02-17
Order Cancellation: Specification

### LINKED ISSUE #320: Soft (offchain) Order Cancellation

bh2smith:
First part of #320

Specification for API endpoint for order cancellation


vkgnosis:
I don't have a strong opinion on POST vs DELETE or whether the the uid is in part of the body. Since we already specced out the route via DELETE and uid in path I would just keep it that way.

orderbook/openapi.yml
```diff
@@ -118,22 +118,55 @@ paths:
...
         Authentication must be provided by signing the following message:
-        signTypedData({operation: "delete", orderUid: "0x.."})
       parameters:
         - in: path
           name: UID
           schema:
             $ref: "#/components/schemas/UID"
           required: true
+      requestBody:
+        description: The order to cancel.
```
fleupold, line 132: The body is the EIP712 signature of `struct OrderCancellation { orderUid: bytes }` from the order's owner.


nit: no strong feelings, but maybe having the request body be an object with field signature instead of directly being a string might make it easier to add other parameters later if we decided to do so.

fleupold (commented):
Looks good modulo all the commented code.

orderbook/openapi.yml
bh2smith: I'm pretty sure the order body is the Signature struct... as declared a few lines below here... Is that not what you mean?

bh2smith: Or do you mean that the `#/components/schemas/Signature` is not well defined as is?

fleupold: What's the type of Signature? It's not defined as `type: object` so I'd assume it's a raw `string`. This means a valid body would be

```
"0xff...fff"
```

and I think it would be nicer to have the body be:

```
{ "signature": "0xff...fff" }
```

bh2smith: Gotcha!

bh2smith: I have updated this according to your input, but now have the same string in both `requestBody.description` and `OrderCancellation.description`. I would think that we might change the request body description to just say "Signed OrderCancellation" if we want to avoid redundancy.

fleupold: Sounds good to me. If you remove the commented code and move this PR from Draft I think this is good to go.

bh2smith: Whoops, I already started implementation. Will have to separate these.

orderbook/openapi.yml
```diff
@@ -118,20 +118,28 @@ paths:
         404:
           description: Order was not found
     delete:
-      summary: Deletes order from order service,
+      summary: Deletes order from order service
```
fleupold, line 121: ubernit: Maybe call it invalidate or cancel rather than delete here and in the description below (since we are not actually deleting it).

orderbook/openapi.yml
```diff
@@ -118,20 +118,28 @@ paths:
...
         Authentication must be provided by signing the following message:
-        signTypedData({operation: "delete", orderUid: "0x.."})
       parameters:
         - in: path
           name: UID
           schema:
             $ref: "#/components/schemas/UID"
           required: true
+      requestBody:
+        description: "EIP712 signature of struct OrderCancellation { orderUid: bytes } from the order's owner"
```
fleupold, line 132: I believe you wanted to change this to something more general?

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2F3KVcCFmm4vemxPcoAq%2Fgiphy.gif&token=soft" data-gitmeme-token="soft"></a>

PR #322, opened 2021-02-17
Http solver: filter orders without native token connection

vkgnosis:
For the solver to run correctly we need to be sure that there are no isolated islands of
tokens without connection between them. As a simple solution that works most of the time
we remove orders without a uniswap pool connection their sell token to the native token.
Our fee estimation code also assumes such a pool exists.

### Test Plan
new unit test

### DIFF
--- a/solver/src/http_solver.rs
+++ b/solver/src/http_solver.rs
@@ -45,10 +45,16 @@ pub struct HttpSolver {
     client: Client,
     api_key: Option<String>,
     config: SolverConfig,
+    native_token: H160,
 }
 
 impl HttpSolver {
-    pub fn new(base: Url, api_key: Option<String>, config: SolverConfig) -> Self {
+    pub fn new(
+        base: Url,
+        api_key: Option<String>,
+        config: SolverConfig,
+        native_token: H160,
+    ) -> Self {
         // Unwrap because we cannot handle client creation failing.
         let client = Client::builder().build().unwrap();
         Self {
@@ -56,6 +62,7 @@ impl HttpSolver {
             client,
             api_key,
             config,
+            native_token,
         }
     }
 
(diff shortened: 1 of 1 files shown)


fleupold (commented):
Would it be hard to implement an iterative or recursive search algorithm that adds in the first path all tokens directly connected to ETH and in the n-th path all tokens directly connected to any token previously found, until the set of connected tokens doesn't change? It would be O(n^2) but I'm not to worried about this performance.

My worry by implementing the simple strategy now is that we couple the limitation of the fee computation with another quite unrelated part (I was planning to extend the fee estimation to use the same algorithm the frontend uses which allows up to 2 intermediate hops after I merge the persistence).

If the more sophisticated search is hard to implement I'm ok merging this approach, but I'm worried we might forget about it when we extend the range of orders we allow in the frontend.

vkgnosis:
Updated with something like you suggest.

solver/src/http_solver.rs
```diff
@@ -212,6 +224,43 @@ fn split_liquidity(liquidity: Vec<Liquidity>) -> (Vec<LimitOrder>, Vec<AmmOrder>
...
+                false
+            } else if fee_connected_tokens.contains(&tokens.1) {
+                fee_connected_tokens.insert(tokens.0);
+                added_token = true;
+                false
+            } else {
+                true
+            }
+        });
+        if amms.is_empty() || !added_token {
```
fleupold, line 252: nanonit: curious why not a while loop? Not that it makes a big difference, I personally prefer while loops as they make the termination condition a bit more obvious than searching for the break.

fleupold (approved):
👏 great implementation

solver/src/http_solver.rs
vkgnosis: The is_empty could be moved to the while loop but the !added_token would still have to be checked inside and this way they are together. No preference either way.

PR #323, opened 2021-02-17
Pick solvable orders while taking balances into account

vkgnosis:
The order book has to make a choice for which orders to include when a user has multiple orders selling the same token but not enough balance for all of them. This is a first PR to fix that. We use a simple algorithm to pick which orders are solvable by ordering orders by uid and keeping track of how much available balance is left.
This could be improved for partially fillable orders (see comment in code) but I wanted to get this in first before tackling that.
There are definitely other criteria that could be used instead of order uid. The point of this PR is to get a simple solution in place. We can make it more optimal in future PRs.

### Test Plan
adapted unit test

### DIFF
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -296,7 +296,7 @@ impl Default for OrderMetaData {
 }
 
 // uid as 56 bytes: 32 for orderDigest, 20 for ownerAddress and 4 for validTo
-#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Ord, PartialOrd)]
 pub struct OrderUid(pub [u8; 56]);
 
 impl FromStr for OrderUid {
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -296,7 +296,7 @@ impl Default for OrderMetaData {
 }
 
 // uid as 56 bytes: 32 for orderDigest, 20 for ownerAddress and 4 for validTo
-#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Ord, PartialOrd)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub struct OrderUid(pub [u8; 56]);
 
 impl FromStr for OrderUid {
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -135,10 +135,9 @@ impl Orderbook {
         // prioritized when a user's balance is too small to cover all of their orders.
         // We can hopefully resolve this when we have a custom struct for orders in the
         // get_solver_orders route and a custom endpoint to query user balances for the frontend.
+        set_available_balances(orders.as_mut_slice(), &balances);
         if filter.exclude_insufficient_balance {
             orders = solvable_orders(orders, &balances);
-        } else {
-            set_available_balances(orders.as_mut_slice(), &balances);
         }
         Ok(orders)
     }
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -325,7 +325,7 @@ mod tests {
             .expect_get_balance()
             .return_const(Some(10.into()));
 
-        let orders = vec![
+        let mut orders = vec![
             Order {
                 order_creation: OrderCreation {
                     sell_amount: 3.into(),
@@ -333,24 +333,7 @@ mod tests {
                     ..Default::default()
                 },
                 order_meta_data: OrderMetaData {
-                    creation_date: DateTime::<Utc>::from_utc(
-                        NaiveDateTime::from_timestamp(2, 0),
-                        Utc,
-                    ),
-                    ..Default::default()
-                },
-            },
-            Order {
-                order_creation: OrderCreation {
-                    sell_amount: 1.into(),
-                    fee_amount: 0.into(),
-                    ..Default::default()
-                },
-                order_meta_data: OrderMetaData {
-                    creation_date: DateTime::<Utc>::from_utc(
-                        NaiveDateTime::from_timestamp(0, 0),
-                        Utc,
-                    ),
+                    creation_date: DateTime::from_utc(NaiveDateTime::from_timestamp(2, 0), Utc),
                     ..Default::default()
                 },
             },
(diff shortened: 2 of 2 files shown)


fleupold (changes requested):
Thanks for tackling this issue!

Can we instead of sorting by UID sort by timestamp for more stability and combine the balance fetching and balance filtering into one method?

orderbook/src/orderbook.rs
```diff
@@ -123,13 +124,54 @@ async fn set_order_balance(orders: &mut [Order], balance_fetcher: &dyn BalanceFe
...
-        let balance = order.order_meta_data.available_balance.unwrap_or_default();
-        !balance.is_zero()
-            && (order.order_creation.partially_fillable
-                || balance >= (order.order_creation.sell_amount + order.order_creation.fee_amount))
-    });
+// The order book has to make a choice for which orders to include when a user has multiple orders
+// selling the same token but not enough balance for all of them.
+// Assumes balance fetcher is already tracking all balances.
+fn solvable_orders(mut orders: Vec<Order>, balance_fetcher: &dyn BalanceFetching) -> Vec<Order> {
+    // Group orders by owner and sell token. For every group we get the balance that the orders
```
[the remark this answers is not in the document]
vkgnosis: Agree. Will change to creation time.

orderbook/src/orderbook.rs
```diff
@@ -123,13 +124,54 @@ async fn set_order_balance(orders: &mut [Order], balance_fetcher: &dyn BalanceFe
...
+        let key = Key {
+            owner: order.order_meta_data.owner,
+            sell_token: order.order_creation.sell_token,
+        };
+        orders_map.entry(key).or_default().push(order);
+    }
+
+    let mut result = Vec::new();
+    for (key, orders) in orders_map {
+        let mut remaining_balance = match balance_fetcher.get_balance(key.owner, key.sell_token) {
```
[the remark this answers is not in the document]
vkgnosis: I think I should make another PR first so that we can return a different `Order` struct in the solvable_orders route than the orders route for the frontend. There we can decide how we want to encode balance for partial orders without affecting the old route.
Then I will come back to here and see what the best way to refactor the balance fetching is.

orderbook/src/orderbook.rs
```diff
@@ -123,13 +124,54 @@ async fn set_order_balance(orders: &mut [Order], balance_fetcher: &dyn BalanceFe
...
+            Some(balance) => balance,
+            None => continue,
+        };
+        for order in orders {
+            // TODO: This is overly pessimistic for partially filled orders where the needed balance
```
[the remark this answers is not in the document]
vkgnosis: Made an issue for this.

orderbook/src/orderbook.rs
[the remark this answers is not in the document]
vkgnosis: I have refactored this code to get rid of the "warmed up" problem. Made an issue for the partially fillable orders because I think it's better to tackle that outside of this PR.

vkgnosis:
Please review again. We can merge this before https://github.com/gnosis/gp-v2-services/pull/324 so we get this fix in quickly.

model/src/order.rs
```diff
@@ -296,7 +296,7 @@ impl Default for OrderMetaData {
 }
 
 // uid as 56 bytes: 32 for orderDigest, 20 for ownerAddress and 4 for validTo
-#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, Ord, PartialOrd)]
```
fleupold, line 299: nit: Is this still needed?

orderbook/src/orderbook.rs
```diff
@@ -276,58 +308,71 @@ mod tests {
...
         ];
 
-        remove_orders_without_sufficient_balance(&mut orders);
-
-        // Only the partially fillable order is included
-        assert_eq!(orders.len(), 1);
-        assert_eq!(orders[0].order_creation.partially_fillable, true);
+        let balances = hashmap! {Default::default() => U256::from(10)};
+        let solvable_orders = solvable_orders(orders.clone(), &balances);
+        // First order has higher uid so it isn't picked.
```
fleupold, line 375: timestamp right?

orderbook/src/orderbook.rs
```diff
@@ -276,58 +308,71 @@ mod tests {
...
+                        Utc,
+                    ),
                     ..Default::default()
                 },
             },
             Order {
                 order_creation: OrderCreation {
-                    sell_amount: 200.into(),
-                    partially_fillable: false,
+                    sell_amount: 1.into(),
```
fleupold, line 345: nanonit: any reason we want three orders here? I feel it might be slightly easier to understand with 2 orders.

orderbook/src/orderbook.rs
```diff
@@ -123,9 +127,18 @@ impl Orderbook {
...
+        // orders for the solver or the frontend. For the frontend (else case) balances are always
+        // actual balances but for the solver there is custom logic to decide which orders get
+        // prioritized when a user's balance is too small to cover all of their orders.
+        // We can hopefully resolve this when we have a custom struct for orders in the
+        // get_solver_orders route and a custom endpoint to query user balances for the frontend.
         if filter.exclude_insufficient_balance {
-            remove_orders_without_sufficient_balance(&mut orders);
+            orders = solvable_orders(orders, &balances);
+        } else {
+            set_available_balances(orders.as_mut_slice(), &balances);
```
fleupold, line 141: Can we set the available balances regardless of whether or not we `exclude_insufficient_balance` orders? I frequently use the `/api/v1/orders/` endpoint and I think it would be nice to get non-0 values for available balance.

Or is there a reason we don't want to set that value?

fleupold (approved):

orderbook/src/orderbook.rs
vkgnosis: It should be set either way. If `exclude_insufficient_balance` is true then we go into the solvable_orders branch which sets it and if it is false we go into this branch which sets it.

fleupold: I'm probably blind, but I can't find where `solvable_orders` sets `available_balance`...

vkgnosis: You're right. I was convinced I did this in in solvable_orders but it's not in the diff. Fixed by moving `set_available_balances` out of the branch.

PR #324, opened 2021-02-18
Use specialized struct for solver orders

vkgnosis:
There is no need to carry around the extra information for the front end facing orders endpoint so we use a smaller struct for the solver facing endpoint. Likewise querying these orders from the db can be done with a specialized query.
This is good to have before #323 . In this PR the balance filtering for solver orders is temporarily removed. Will add it back in that PR.
Before writing more tests for the db query I want to see if this PR gets positive feedback. If so I will add more tests before merging into main.

### Test Plan
e2e test

### DIFF
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -17,6 +17,15 @@ use std::fmt::{self, Display};
 use std::str::FromStr;
 use web3::signing::{self, Key, SecretKeyRef};
 
+/// An order returned by the `solvable_orders` route.
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SolverOrder {
+    #[serde(flatten)]
+    pub order_creation: OrderCreation,
+    pub executable_amount: Option<U256>,
+}
+
 /// An order that is returned when querying the orderbook.
 ///
 /// Contains extra fields that are populated by the orderbook.
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -191,9 +191,9 @@ paths:
                 type: array
                 items:
                   $ref: "#/components/schemas/Trade"
-  /api/v1/solvable_orders:
+  /api/v1/solver_orders:
     get:
-      summary: Get solvable orders.
+      summary: Get orders that should be considered by solvers.
       description: |
         The set of orders that solvers should be solving right now. These orders are determined to
         be valid at the time of the request.
@@ -205,7 +205,7 @@ paths:
               schema:
                 type: array
                 items:
-                  $ref: "#/components/schemas/Order"
+                  $ref: "#/components/schemas/SolverOrder"
   /api/v1/fee:
     get:
       description: |
@@ -361,6 +361,16 @@ components:
         invalidated:
           description: Has this order been invalidated?
           type: boolean
+    PartiallyFillable:
+      description: Extra information about partially fillable orders.
+      type: object
+      properties:
+        executableAmount:
+          description: |
+            Amount of the order that can be executed based on user balance and previous executions.
+            Null for fill or kill orders.
+          $ref: "#/components/schemas/TokenAmount"
+          nullable: true
     Order:
       allOf:
         - $ref: "#/components/schemas/OrderCreation"
@@ -373,6 +383,10 @@ components:
         signature:
           description: "OrderCancellation signed by owner"
           $ref: "#/components/schemas/Signature"
+    SolverOrder:
+      allOf:
+        - $ref: "#/components/schemas/OrderCreation"
+        - $ref: "#/components/schemas/PartiallyFillable"
     Trade:
       description: |
         Trade data such as executed amounts, fees, order id and block number.
--- a/orderbook/src/api/get_solver_orders.rs
+++ b/orderbook/src/api/get_solver_orders.rs
@@ -1,29 +1,29 @@
 use crate::api::convert_get_orders_error_to_reply;
 use crate::orderbook::Orderbook;
 use anyhow::Result;
-use model::order::Order;
+use model::order::SolverOrder;
 use std::{convert::Infallible, sync::Arc};
 use warp::{hyper::StatusCode, reply, Filter, Rejection, Reply};
 
-fn get_solvable_orders_request() -> impl Filter<Extract = (), Error = Rejection> + Clone {
-    warp::path!("solvable_orders").and(warp::get())
+fn get_solver_orders_request() -> impl Filter<Extract = (), Error = Rejection> + Clone {
+    warp::path!("solver_orders").and(warp::get())
 }
 
-fn get_solvable_orders_response(result: Result<Vec<Order>>) -> impl Reply {
+fn get_solver_orders_response(result: Result<Vec<SolverOrder>>) -> impl Reply {
     match result {
         Ok(orders) => Ok(reply::with_status(reply::json(&orders), StatusCode::OK)),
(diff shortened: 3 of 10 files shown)


orderbook/openapi.yml
```diff
@@ -181,9 +181,9 @@ paths:
                 type: array
                 items:
                   $ref: "#/components/schemas/Trade"
-  /api/v1/solvable_orders:
+  /api/v1/solver_orders:
```
vkgnosis, line 184: changed because I find this name slightly better

orderbook/src/database.rs
```diff
@@ -37,3 +41,34 @@ impl Database {
         Ok(())
     }
 }
+
+#[derive(sqlx::Type)]
+#[sqlx(rename = "OrderKind")]
+#[sqlx(rename_all = "lowercase")]
+enum DbOrderKind {
+    Buy,
```
vkgnosis, line 49: Moved here from orders.rs because it's shared with solver_orders

orderbook/openapi.yml
fleupold: I actually like the other name better because it describes the orders where here we imply the use case (which I don't think we have to). Anyways, no big deal.

model/src/order.rs
```diff
@@ -19,6 +19,16 @@ use web3::{
...
     types::Recovery,
 };
+
+/// An order returned by the `solvable_orders` route.
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SolverOrder {
+    #[serde(flatten)]
+    pub order_creation: OrderCreation,
+    pub executable_amount: Option<U256>,
```
fleupold, line 29: I assume for partial orders we need to differentiate how much fee has already been paid (how much of sellAmount - executable amount is due to prior fills and how much is due to missing balance) in order to compute the correct fees/discounts.

orderbook/src/database/solver_orders.rs
```diff
@@ -0,0 +1,110 @@
...
+use futures::{stream::TryStreamExt, Stream};
+use model::{
+    order::{OrderCreation, OrderKind, SolverOrder},
+    Signature,
+};
+
+impl Database {
+    pub fn solver_orders(&self, min_valid_to: u32) -> impl Stream<Item = Result<SolverOrder>> + '_ {
+        const QUERY: &str = "\
+        SELECT * FROM ( \
```
fleupold, line 14: I think it's a shame to have this query logic duplicated only for shortcutting a few where filters (which I assume the query optimizer cuts out anyways).

fleupold (commented):
I see value for the specialised struct, however I'm not convinced we should duplicate the fetching logic (I assume for balance computation a lot of the logic will also be duplicated). 

Maybe we can internally convert a frontend Order into the specialized order but use the same path under the hood?

model/src/order.rs
vkgnosis: Whether the executable amount is limited due to balance or previous executoins doesn't matter to the solver right? That was my thinking for having only one field. For every execution the fee the contract can take is `order.feeAmount * executedSellAmount / order.sellAmount` so it doesn't depend on previous executon fees.

orderbook/src/database/solver_orders.rs
vkgnosis: I thought it was an improvement also considering that the frontend facing query might become more complicated in the future or diverge more in general. It is true that logic is duplicated but that also makes this code path simpler.

vkgnosis:
>  however I'm not convinced we should duplicate the fetching logic (I assume for balance computation a lot of the logic will also be duplicated).

This definitely deserves discussion. To me these routes feel different enough that I thought it would be better to split them up even though they could share code right now. I am worried that otherwise we end up making get_solver_orders more complicated as get_orders gets extended more.
It is true that balance computation will be somewhat duplicated too. On the other hand I am not sure if availableBalance even makes sense on the frontend facing orders anymore when the solver_orders is it's own struct. For the frontend a new route `user_balance <user address> <sell token>` probably makes more sense than attaching this to every order. So the logic would end up existing only once if we made that change too.

model/src/order.rs
fleupold: Yes you are right, I thought the feeDiscount depended on previously charged fees, but it only depends on the executed amount of the current trade.

orderbook/openapi.yml
bh2smith: I prefer this new name because everyone I saw solvable orders it gave me the impression that we already know they can be matched.

orderbook/src/database.rs
```diff
@@ -37,3 +41,34 @@ impl Database {
...
+
+    fn into(self) -> OrderKind {
+        match self {
+            Self::Buy => OrderKind::Buy,
+            Self::Sell => OrderKind::Sell,
+        }
+    }
+}
+
+fn h160_from_vec(vec: Vec<u8>) -> Result<H160> {
```
bh2smith, line 69: This has been moved to the conversion crate in my recent PR so we can expect conflicts at some point.

orderbook/src/api/get_solver_orders.rs
```diff
@@ -1,15 +1,15 @@
 use crate::api::convert_get_orders_error_to_reply;
 use crate::orderbook::Orderbook;
 use anyhow::Result;
-use model::order::Order;
+use model::order::SolverOrder;
 use std::{convert::Infallible, sync::Arc};
 use warp::{hyper::StatusCode, reply, Filter, Rejection, Reply};
 
 fn get_solvable_orders_request() -> impl Filter<Extract = (), Error = Rejection> + Clone {
```
bh2smith, line 8: Should this also be renamed to solver?

bh2smith (approved):
Nifty.

vkgnosis:
I've resolved the merge conflicts. Looking back at the discussion I think I was waiting for a response from you @fleupold .

orderbook/src/database/solver_orders.rs
fleupold: One thing that is missing in this query now is the soft cancellation check. I think this is an example of where it might be easy to forget to update one callsite.

fleupold (commented):
Sorry for not following up here (maybe we should have a stale bot pinging these kinds of PRs). I still think the frontend will want to know on a per order basis what is the fillable amount of a single order according to the users balance given all other orders. While a `user_balance <user address> <sell token>` endpoint can answer this on a global level (before considering any or all orders), we might want to flag individual orders as temporarily not fillable because their executable amount is lower than their sell amount.

fleupold (commented):
I do really like the concept of `executable_amount` (it's much clearer than availableBalance - executed amount).

Given that `available_balance` is not yet used by any component (and will likely be replaced by a separate endpoint) maybe we can expose this `executable_amount` on `OrderMetaData` and leave the decision if the OrderStruct and db query should be separated between frontend and solvers to a later date?

orderbook/src/database/solver_orders.rs
bh2smith: I think this should be ok now since the offline cancellation is represented as part of the `invalidated` field now.

vkgnosis:
Closing this for now because we started making use of order meta data in the solver too and the point about not duplicating the db query makes sense.
The executable amount change is useful but I don't like the code duplication just for that. Can attempt to do something about that again in the future if it becomes more clear if/how the frontend wants to react to some orders not being solvable.

PR #325, opened 2021-02-19
Add a stream that yields the current block whenever it changes

vkgnosis:
I think it would be useful to have this in several places where we want to react to the current block changing like updating user balances.
At first I implemented this with a filter but then realized this doesn't make sense as the comment in the code explains.

### Test Plan
new test with real node


shared/src/current_block.rs
```diff
@@ -0,0 +1,73 @@
...
+/// A stream that yields the current block whenever it changes.
+pub fn block_stream(web3: Web3<impl Transport>) -> impl Stream<Item = Block> + FusedStream {
+    // We do not use a block filter because filters have been unreliable in the past, do not work
+    // with load balancing over multiple nodes and would still need to be polled unless we used web
+    // sockets.
+    futures::stream::unfold(H256::zero(), move |previous| {
+        let web3 = web3.clone();
+        async move {
+            loop {
+                tokio::time::delay_for(POLL_INTERVAL).await;
```
fleupold, line 23: nit: shouldn't we delay at the end of the loop to yield the current block without delay?

shared/src/current_block.rs
```diff
@@ -0,0 +1,73 @@
...
+                    Some(hash) => hash,
+                    None => {
+                        tracing::warn!("missing hash");
+                        continue;
+                    }
+                };
+                if hash == previous {
+                    continue;
+                }
+                return Some((block, hash));
```
fleupold, line 41: I don't think this is handling the case well when fetching the current_block taskes longer than the time between blocks (which theoretically can be less than 1s). From the documentation I would assume the stream to yield all block changes without any gaps (and a consumer might rely on the hashes returned to e.g. populate a database. 

I'm not sure we have to provide a gapless experience (by inspecting block.parent and comparing it with a lit of previous hashes) but should maybe clarify that this is a "latest block stream" which might have gaps.

fleupold (approved):
I definitely like this initiative to have a more reactive architecture...

How do you envision this to work with multiple components which each depend on a block stream? Ideally there should only be one instance of "web3 polling" per process and multiple consumers can subscribe to updates without creating a new I/O dependent stream themselves. 
Maybe this could be abstracted away behind some factory that fans out new blocks via channels into lightweight sub-streams which are all backed by the same network roundtrip?

shared/src/current_block.rs
vkgnosis: Can't really have it just at the end because want to yield the current block from the stream immediately, but that returns from the function. We could keep it at the start but have a bool with which we can skip the very first sleep.  Adding the bool is only  3 lines but on the other hand I don't think it's wrong to sleep immediately either.

shared/src/current_block.rs
vkgnosis: Agree, will document.

vkgnosis:
> Maybe this could be abstracted away behind some factory that fans out new blocks via channels into lightweight sub-streams which are all backed by the same network roundtrip?

Totally agree with this. When we have merged this base functionality I will look into the best way to share one poller with multiple streams.

vkgnosis:
Since it wasn't that much code I added the "single poll instance' feature as another commit. Please review again.

shared/src/current_block.rs
```diff
@@ -0,0 +1,100 @@
...
+    impl Stream<Item = Block> + Clone + Unpin,
+    impl Future<Output = ()>,
+) {
+    let default_hash = H256::default();
+    let default_block = Block {
+        hash: Some(default_hash),
+        number: Some(0.into()),
+        ..Default::default()
+    };
+    let (sender, receiver) = watch::channel(default_block);
```
fleupold, line 37: I find yielding an "invalid" default block initially really hacky. Could we instead pass `Option<Block>` and wrap receiver in a `filter_map` that only forwards the `Some` items?

fleupold (commented):
Would it make sense to expose a `CurrentBlockStreamFactory` struct that takes care of lazily creating the future and driving it and itself just exposes a `current_block_stream` method (which under the hood clones from the channel).

I feel like right now the interface looks quite scary (raw) for someone that just wants to get a stream of block updates (they have to take care of driving the future somehow themselves). 
I also feel like it's leaking the way we accomplish the sharing by exposing the future and a clonable stream.

vkgnosis:
> I feel like right now the interface looks quite scary (raw) for someone that just wants to get a stream of block updates (they have to take care of driving the future somehow themselves).

The alternative to this is to implicitly `tokio::spawn` the updater future immediately which is fine by me too. But that would be the same for the CurrentBlockStreamFactory, right? I don't see how else the Factory could take of driving it.

For the Stream + Clone I thought that was the simplest solution. If I want to consume a block stream I can just take impl Stream<Item=Block> instead of CurrentBlockStreamFactory. This didn't feel like an implementation detail to me.

shared/src/current_block.rs
vkgnosis: Agree. Implemented this in the first new commit. Had to introduce a wrapper type because filter_map's Stream isn't clonable.

shared/src/current_block.rs
```diff
@@ -0,0 +1,118 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use futures::StreamExt;
+
+    // cargo test current_block -- --ignored --nocapture
+    #[tokio::test]
+    #[ignore]
+    async fn mainnet() {
+        let node = "https://dev-openethereum.mainnet.gnosisdev.com";
```
bh2smith, line 109: Is this a public node? Does it require an API key?

vkgnosis: Since the test works it apparently doesn't require any authentication.

fleupold:
> The alternative to this is to implicitly tokio::spawn the updater future immediately which is fine by me too. But that would be the same for the CurrentBlockStreamFactory, right? I don't see how else the Factory could take of driving it.

One minor difference I see that now the API is still slightly simpler to misuse. An unaware consumer might just call `current_block_stream` again instead of cloning the previous stream and thus creating a new task. This can happen with a factory too (if the consumer creates a new factory) but requires a bit more ignorance.

Anyways, I think this is good to go.

PR #331, opened 2021-02-23
Order Cancellation: Implementation

### LINKED ISSUE #320: Soft (offchain) Order Cancellation

bh2smith:
Partial fulfillment of #320

TODO

- [x] Invalidate order on successful signature validation
- [x] test the signature validation
- [x] Filter out cancelled orders from solvable orders (this occurs by default since we piggyback on the invalidated field)
- [x] ~Figure out how to use null in deserialization and back test~ Not needed until invalidation field is `Option<DateTime>`

Saved for the next PR:

- [ ] Write the cancel_order API (request, response, etc...)
- [ ] Probably other tests and things I am missing

### Test Plan

Unit and E2e tests.


model/src/order.rs
```diff
@@ -258,21 +258,70 @@ impl OrderCreation {
...
+    pub const ORDER_CANCELLATION_TYPE_HASH: [u8; 32] =
+        hex!("7b41b3a6e2b3cae020a3b2f9cdc997e0d420643957e7fea81747e984e47c88ec");
+
+    fn order_cancellation_digest(&self) -> [u8; 32] {
+        let mut hash_data = [0u8; 64];
+        hash_data[0..32].copy_from_slice(&Self::ORDER_CANCELLATION_TYPE_HASH);
+        hash_data[44..64].copy_from_slice(&signing::keccak256(&self.order_uid.0));
+        signing::keccak256(&hash_data)
+    }
+    // TODO - Many duplicated methods from OrderCreation, need EIP712 Signing Library
```
fleupold, line 281: I think these methods could become part of the Signature type and be shared between OrderCreation and OrderCancellation if it takes as argument the domain separator and `hash_data`. `hash_data` can then be composed individually inside `OrderCancellation` and `OrderCreation`

model/src/order.rs
```diff
@@ -258,21 +258,70 @@ impl OrderCreation {
...
+    }
+}
 
 /// An order as provided to the orderbook by the frontend.
 #[serde_as]
 #[derive(Eq, PartialEq, Clone, Debug, Deserialize, Serialize, Hash)]
 #[serde(rename_all = "camelCase")]
 pub struct OrderMetaData {
     pub creation_date: DateTime<Utc>,
+    pub cancellation_date: Option<DateTime<Utc>>,
```
fleupold, line 324: Does this need to become part of the OpenApi? I'm also wondering if we even need to differentiate between "on-chain" invalidated and "soft" cancelled field (we could reuse `invalidated` on this struct). If we do introduce a new field we should probably make sure the naming reflects these two differences in a more obvious way and the data returned is roughly in the same format (e.g. return a timestamp for both).

orderbook/openapi.yml
```diff
@@ -383,8 +383,10 @@ components:
           type: string
           enum:
             [
+                Cancelled,
```
fleupold, line 386: This is not an error is it? It's the success case on the Result enum and as such shouldn't be listed here.

orderbook/src/database/orders.rs
```diff
@@ -78,6 +78,20 @@ impl Database {
             .map(|_| ())
     }
 
+    pub async fn cancel_order(&self, order: &Order) -> Result<()> {
+        const QUERY: &str = "\
+            UPDATE orders
+            SET cancellation_timestamp = $1 \
```
fleupold, line 84: We should probably only set it if its not already set? Otherwise I could mess with the history by re-cancelling an already cancelled order.

orderbook/src/orderbook.rs
```diff
@@ -78,8 +81,35 @@ impl Orderbook {
...
+            Some(signer) => {
+                if signer == order.order_meta_data.owner {
+                    self.database.cancel_order(&order).await?;
+                    // Can't fail to remove here, since we know it exists already
+                    Ok(OrderCancellationResult::Cancelled)
+                } else {
+                    Ok(OrderCancellationResult::InvalidSignature)
+                }
+            }
+            None => Ok(OrderCancellationResult::MissingSignature),
```
fleupold, line 111: In this case it seems like the signature is invalid (it's missing if it's not even part of the request).

orderbook/src/orderbook.rs
```diff
@@ -78,8 +81,35 @@ impl Orderbook {
...
+        };
+
+        match cancellation.validate_signature(&self.domain_separator) {
+            Some(signer) => {
+                if signer == order.order_meta_data.owner {
+                    self.database.cancel_order(&order).await?;
+                    // Can't fail to remove here, since we know it exists already
+                    Ok(OrderCancellationResult::Cancelled)
+                } else {
+                    Ok(OrderCancellationResult::InvalidSignature)
```
fleupold, line 108: Technically the signature is valid (just for another user), but I'm fine not differentiating between the two cases (although it might help debugging if it logged the wrong address).

orderbook/src/orderbook.rs
```diff
@@ -78,8 +81,35 @@ impl Orderbook {
...
+        let order = match orders.first() {
+            Some(order) => order,
+            None => return Ok(OrderCancellationResult::OrderNotFound),
+        };
+
+        match cancellation.validate_signature(&self.domain_separator) {
+            Some(signer) => {
+                if signer == order.order_meta_data.owner {
+                    self.database.cancel_order(&order).await?;
+                    // Can't fail to remove here, since we know it exists already
```
fleupold, line 105: nit: I don't understand this comment

model/src/order.rs
```diff
@@ -258,21 +258,70 @@ impl OrderCreation {
...
+/// An order cancellation as provided to the orderbook by the frontend.
+#[serde_as]
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Deserialize, Serialize, Hash)]
+pub struct OrderCancellation {
+    // TODO - does it make sense to include the orderUid here?
+    pub order_uid: OrderUid,
+    pub signature: Signature,
+}
+
+impl OrderCancellation {
```
fleupold, line 270: A test case would be nice so that the frontend can verify how the signing logic needs to look like

fleupold (commented):
Wow, this is looking really good. Some comments inline. Great how you managed to pull this together so quickly.

model/src/order.rs
bh2smith: This is also part of the todo.

model/src/order.rs
bh2smith: Yes certainly. I guess this would be an intermediary step between a fully compatible EIP712 signing library.

model/src/order.rs
bh2smith: I think we can simply go with making invalidated an optional date time and use the same field for both.

orderbook/src/orderbook.rs
bh2smith: I guess this wording is old (I.e "remove") was referring to the fact that this orderID is already known to exist from above.

model/src/order.rs
fleupold: Right now we don't store the timestamp of the block that carried the invalidated event, but it's something we can add.

model/src/order.rs
bh2smith: Have made an EIP712Signing trait that is used by both `OrderCreation` and `OrderCancellation` which now only have to implement their unique "digest".

fleupold: I would prefer doing this in this PR (or even better throwing in a PR before that moves the existing order signature validation into the Signature struct) over introducing the redundancy here.

model/src/order.rs
```diff
@@ -42,7 +43,7 @@ impl Order {
         order_creation: OrderCreation,
         domain: &DomainSeparator,
     ) -> Option<Self> {
-        let owner = order_creation.validate_signature(domain)?;
+        let owner = order_creation.validate_signature(domain, order_creation.signature)?;
```
fleupold, line 46: nit: I don't like that we are passing a value from the object itself into a method on this object (it should have access to `self.signature` inside `validate_signature` and there is no reason to pass any other signature).

I think we can expose a `validate_signature(domain)` on both `OrderCreation` and `OrderCancellation` which internally calls into the shared helper methods.

model/src/signing.rs
```diff
@@ -0,0 +1,43 @@
...
+    fn signing_digest(&self, domain_separator: &DomainSeparator, signature: Signature) -> [u8; 32] {
+        // This is fallback for wallets that don't support EIP712
+        if signature.v & 0x80 == 0 {
+            self.signing_digest_typed_data(domain_separator)
+        } else {
+            self.signing_digest_message(domain_separator)
+        }
+    }
+
+    fn validate_signature(&self, domain_separator: &DomainSeparator, signature: Signature) -> Option<H160> {
```
fleupold, line 35: Instead of creating a trait that needs to be added to all structs that carry a signature, I would add these methods to `impl Signature` directly and have it take the digest and domain separator as an arguments. Then expose `validate_signature` on OrderCreation/Cancellation and internally call `self.signature.validate(domain_separator, digest)`

model/src/order.rs
bh2smith: Right, I could copy-paste the validate signature methods into these structs directly (but then we have duplicated code). Traits can't have fields and as a result I cannot access signature if the method is defined in the trait.

model/src/signing.rs
bh2smith: Very nice idea. This could clear up the problem with the trait I was having.

model/src/order.rs
```diff
@@ -202,19 +192,18 @@ impl OrderCreation {
...
-
     // keccak256("sell")
     const ORDER_KIND_SELL: [u8; 32] =
         hex!("f3b277728b3fee749481eb3e0b3b48980dbbab78658fc419025cb16eee346775");
     // keccak256("buy")
     const ORDER_KIND_BUY: [u8; 32] =
         hex!("6ed88e868af0a1983e3886d5f3e95a2fafbd6c3450bc229e27342283dc429ccc");
 
-    fn order_digest(&self) -> [u8; 32] {
+    pub fn digest(&self) -> [u8; 32] {
```
bh2smith, line 206: Given that digest is shared among any struct that has a Signature field, I would think we should still have a trait Signing which declares the digest function signature as a mandatory method that needs to be implemented.

orderbook/src/database/orders.rs
bh2smith: In this PR, since we are going to go  with the simple boolean for now, overwriting true with true shouldn't be a problem. However, I will create an issue about using timestamps that mentions this as well.

orderbook/src/database/orders.rs
```diff
@@ -78,6 +78,19 @@ impl Database {
             .map(|_| ())
     }
 
+    pub async fn cancel_order(&self, order: &Order) -> Result<()> {
+        const QUERY: &str = "\
+            UPDATE orders
+            SET invalidated = true \
```
fleupold, line 84: To reduce future db updates maybe we can still persist the timestamp but only return a boolean to the caller of the API.

Also I don't think orders has an invalidated field yet, so you would need to create a db migration for it anyways.

model/src/order.rs
```diff
@@ -42,7 +40,9 @@ impl Order {
         order_creation: OrderCreation,
         domain: &DomainSeparator,
     ) -> Option<Self> {
-        let owner = order_creation.validate_signature(domain)?;
+        let owner = order_creation
```
fleupold, line 43: I'd slightly prefer if order_creation would still expose valid_signature which does this under the hood (so that this caller here doesn't have to know too much about the internals of the struct). This way the trait would also become valuable as `validate_signature(domain)` could be implemented on the trait (using digest) and only `digest` has to be implemented by OrderCreation and OrderSignature.

bh2smith: Yes, this makes perfect sense, my approach here will be to add this snippet to both `OrderCancellation` and `OrderCreation` 

```rust
    pub fn validate_signature(&self, domain_separator: &DomainSeparator) -> Option<H160> {
        self.signature.validate(domain_separator, &self.digest())
    }
```

Which would result in minimal code duplication, but I would like to take this one step further so that the snippet only has to be written once. My only issue is that this method requires access to the signature field of each struct which I can't implement on the trait level... Am I missing something or is this really as generic as it can get?

bh2smith: This is the commit making the suggested change as proposed with the snippet. Just want to confirm that this subtle code duplication can't avoided:

https://github.com/gnosis/gp-v2-services/pull/331/commits/40005030af34371ac8e12c92bd8baf5edb694be3

fleupold: If the trait exposes a digest() method, you should be able to put your snippet from above as a default implementation into the trait directly (instead of implementing it on both OrderCancellation and OrderCreation)

bh2smith: I will try this out, but I was more concerned about the `struct.signature` field that the method accesses.

fleupold: I believe the trait would also require a signature() method, which could be implemented by returning `self.signature`.

bh2smith: Yes, this would do it! I'll give that a try.

model/src/order.rs
```diff
@@ -202,19 +190,45 @@ impl OrderCreation {
...
+            signature: Default::default(),
+        };
+        result.sign_self_with(&DomainSeparator::default(), &SecretKeyRef::new(&ONE_KEY));
+        result
+    }
+}
 
-    fn order_digest(&self) -> [u8; 32] {
+impl OrderCancellation {
+    fn sign_self_with(&mut self, domain_separator: &DomainSeparator, key: &SecretKeyRef) {
```
fleupold, line 218: nit: I feel like this logic could now also be implemented in the trait.

model/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2018"
 
 [dependencies]
-chrono = { version = "0.4", default-features = false, features = ["serde"] }
+chrono = { version = "0.4", default-features = false, features = ["serde", "clock"] }
```
fleupold, line 7: why is clock needed?

model/Cargo.toml
```diff
@@ -14,7 +14,7 @@ primitive-types = { version = "0.8" }
 secp256k1 = "0.20"
 serde = { version = "1.0", features = ["derive"] }
 serde_with = { version = "1.6", default-features = false, features = ["macros"] }
-web3 = { version = "0.15", default-features = false }
+web3 = { version = "0.15", default-features = false, features = ["signing"] }
```
fleupold, line 17: What are we using  from the signing feature that we didn't use before?

fleupold (commented):
Almost there. For some inspiration on how to do the migration you can check #364

Looking at the size of the PR, I think it would be fine to implement `api/cancel_order.rs` in a follow up PR.

model/Cargo.toml
bh2smith: The issue here is that when the entire project is compiled together, everything was fine but when you try to compile/build/run/test `model` stand alone with something like:

```sh
cargo build -p model --tests
```
then we run into the following errors:

```
➜  gp-v2-services git:(320/cancellation_implementation) ✗ cargo build -p model --tests
   Compiling model v0.1.0 (/Users/bensniff/Projects/gnosis/gp-v2-services/model)
error[E0432]: unresolved imports `web3::signing::Key`, `web3::signing::SecretKeyRef`
  --> model/src/order.rs:18:27
   |
18 | use web3::signing::{self, Key, SecretKeyRef};
   |                           ^^^  ^^^^^^^^^^^^ no `SecretKeyRef` in `signing`
   |                           |
   |                           no `Key` in `signing`

error[E0425]: cannot find function `recover` in module `signing`
  --> model/src/lib.rs:89:18
   |
89 |         signing::recover(&message, &signature, recovery_id).ok()
   |                  ^^^^^^^ not found in `signing`

error: aborting due to 2 previous errors

Some errors have detailed explanations: E0425, E0432.
For more information about an error, try `rustc --explain E0425`.
error: could not compile `model`
```

More concisely, we import it here and it is used several times in this file.

http://github.com/gnosis/gp-v2-services/blob/456859c8a1c0a76f27f986e5296e8602a4fdff63/model/src/order.rs#L18-L18

model/Cargo.toml
bh2smith: This is used here in `model/order.rs` at line 46.

http://github.com/gnosis/gp-v2-services/blob/456859c8a1c0a76f27f986e5296e8602a4fdff63/model/src/order.rs#L38-L54

The only reason we were getting away with it before is because we already had it when compiling the project as a whole but not when building models alone: ` cargo build -p model`

model/src/order.rs
bh2smith: Very clever! Excellent observation.

bh2smith: Although, I am not sure though because the last three lines are actually setting the signature... I would have to implement a set_signature method or something on both... or maybe change things so that this returns a signature to be set outside:

```rust
    fn sign_self_with(&mut self, domain_separator: &DomainSeparator, key: &SecretKeyRef) {
        let message = self
            .signature()
            .signing_digest_message(domain_separator, &self.digest());
        // Unwrap because the only error is for invalid messages which we don't create.
        let signature = Key::sign(key, &message, None).unwrap();
        self.signature.v = signature.v as u8 | 0x80;
        self.signature.r = signature.r;
        self.signature.s = signature.s;
    }
```

bh2smith: Ok, I think I have found something nice that works well and reduces redundancy much more.

bh2smith:
This PR is now ready for review. All suggestions and questions up till this point have been addressed and there are a couple of outstanding tasks to complete in a follow up PR. That is:

- Update the 200 response of the API declaration to return boolean representing whether or not the order was canceled. 
- Write the order cancellation API itself (i.e. request and response handlers with tests).

Thanks for your careful review @fleupold. Looking forward to getting this thing landed.

model/src/lib.rs
```diff
@@ -20,6 +22,29 @@ pub struct Signature {
...
+    fn signature(&self) -> Signature;
+    fn update_signature(&mut self, new_signature: Signature);
+
+    fn sign_self_with(&mut self, domain_separator: &DomainSeparator, key: &SecretKeyRef) {
+        let message = self
+            .signature()
+            .signing_digest_message(domain_separator, &self.digest());
+        // Unwrap because the only error is for invalid messages which we don't create.
+        let signature = Key::sign(key, &message, None).unwrap();
+        self.update_signature(Signature {
```
fleupold, line 36: Instead of adding `update_signature` on the trait, can this method instead return a `Signature` that you then set on self when you call it in the implementation of `default`?

fleupold (approved):
Looks good to me!

orderbook/src/database/orders.rs
```diff
@@ -78,6 +78,24 @@ impl Database {
             .map(|_| ())
     }
 
+    pub async fn cancel_order(&self, order: &Order) -> Result<()> {
```
vkgnosis, line 81: Seeing as this function only needs the uid I would rather just pass that in than the whole order.

model/src/lib.rs
```diff
@@ -37,6 +62,48 @@ impl Signature {
             v: bytes[64],
         }
     }
+
+    fn signing_digest_typed_data(
+        &self,
```
vkgnosis, line 67: This and the following function don't use `self` so could remove the `&self` here and then call them through `Signature::signing_...`.

model/src/lib.rs
bh2smith: Sure I could do that but it is also used elsewhere (but only in the unit tests I believe).

bh2smith: I suspect now that we don't even necessarily need to test `self_sign_with` anymore for both Cancellation and Creation structs since they are now redundant.

bh2smith: All fixed. Looks much better now.

orderbook/src/database/orders.rs
bh2smith: Good find!

model/src/lib.rs
bh2smith: Very nice catch!

bh2smith:
@fleupold - Not sure what is up with the e2e test here... Will have to look into it at a later date. Let me know if you have any thoughts on this one. Could it be the altered database?

model/src/lib.rs
```diff
@@ -20,6 +22,30 @@ pub struct Signature {
     pub v: u8,
 }
 
+pub trait EIP712Signing {
+    fn digest(&self) -> [u8; 32];
+    fn signature(&self) -> Signature;
+
+    fn sign_self_with(
+        &mut self,
```
vkgnosis, line 30:
```suggestion
    fn sign_self_with(
        &self,
```

vkgnosis (approved):
> I think it has something to do with this line in the query... wondering if there is a name collision naming a column with a column that exists in the table:

Maybe it is about operator precedence. You could try

```
 (COUNT(invalidations.*) > 0 OR invalidated IS NOT NULL) AS invalidated
```

bh2smith:
I tried this out as a raw query directly on the database and it was fine. Will look into it again soon.

bh2smith:
Ok, so after a bunch of digging and many attempts at running this test over and over in different ways, for some reason it is the `get_solvable_orders` filter which is causing no orders to be returned for this test. The weirdest part is that I can actually fetch the orders and find a solution by simply commenting this completely not relevant part of the solvable order filter:

http://github.com/gnosis/gp-v2-services/blob/320/cancellation_implementation/orderbook/src/orderbook.rs#L131-L131

Then of course, the test fails because the balances are insufficient... I wonder what is happening here? Has something changed in the test that isn't funding the accounts?

fleupold:
I think the issue is related to the signature recover from the signed order (which is plausible given that this PR refactors the signature code).

I debugged this by disabling [`simulate_settlement`](https://github.com/gnosis/oba-services/blob/9e32a8eedc9886c482f12b64f24eb04fec1be088/solver/src/settlement_submission.rs#L30) (so that we record an actual failed tx on ganache which we can [load into tenderly](https://github.com/Tenderly/tenderly-cli#export)). I then printed the trader addresses, which in my setup turned out to be 

- Trader A 0x7e5f…5bdf
- Trader B 0x2b5a…d6cf

I then loaded the failed tx into tenderly (using `tenderly export <tx hash>`) and looking at the failing instruction:

image

I can see it reverts in the first [transferFrom](https://www.4byte.directory/signatures/?bytes4_signature=0x23b872dd). Looking at the arguments, I see that receiver and amount look fine. However, the sender is *0x5f1f443c07bc4aad435c4a38c762398d01b07* which is none of the listed traders above (which also explains why they don't fail the balance check).

This means the smart contract recovers a different owner from the signature than our order is supposed to be created for.

Here is a more minimal unit test that demonstrates the issue;

<details>
<summary> Test </summary>

```rust
#[test]
fn test_ben() {
    const TRADER_A_PK: [u8; 32] =
        hex!("0000000000000000000000000000000000000000000000000000000000000001");
    let trader_a = Account::Offline(PrivateKey::from_raw(TRADER_A_PK).unwrap(), None);
    let order = OrderBuilder::default()
        .with_sell_token(H160::zero())
        .with_sell_amount(100.into())
        .with_buy_token(H160::zero())
        .with_buy_amount(80.into())
        .with_valid_to(u32::max_value())
        .with_kind(OrderKind::Sell)
        .sign_with(
            &DomainSeparator::default(),
            SecretKeyRef::from(&SecretKey::from_slice(&TRADER_A_PK).unwrap()),
        )
        .build();

    let owne = order
        .order_creation
        .validate_signature(&DomainSeparator::default())
        .unwrap();
    assert_eq!(owner, trader_a.address());
}
```
</details>

In hindsight it might have been helpful to refactor the signature code in a separate PR to avoid the confusion as to where the problem stems from.

model/src/order.rs
fleupold: From the discussion above, I thought we agreed to reuse the existing `invalidated` field instead of creating a new field on the OrderMetaData struct.

fleupold (changes requested):
Let's please get rid of the extra cancellation timestamp field and reuse invalidated

bh2smith:
> Let's please get rid of the extra cancellation timestamp field and reuse invalidated

Invalidated was never a field in the orders table. Are you suggesting that we use the invaldations table and just insert a new invalidation when a soft order cancellation comes in?

bh2smith:
Thanks @fleupold for helping out here. I guess we should really have some tests in place to prevent this. At the moment, this is quite a mess to clean up, but I will look into it and see what went wrong. I suppose the easiest way to get this out the door might be to leave the signature related changes on the order out of this PR and try to make the necessary fix in a separate one.

fleupold:
> Invalidated was never a field in the orders table. Are you suggesting that we use the invaldations table and just insert a new invalidation when a soft order cancellation comes in?

The database table and `OrderMetadata` struct are two different things. I suggest leaving the database table as you have it (introducing a new column that is a timestamp), but map that information into the same boolean field on the `OrderMetadata` struct.

bh2smith:
Ok I think I get it now. I introduced an unnecessary cancellation time stamp on the meta data that is probably not even being used for anything.

model/src/order.rs
bh2smith: Right. I only included this because I thought it might be useful to pass whatever metadata we have on the order when queried. Of course we can still use invalidated, but there is a subtle loss of information (that being the actual time when it was invalidated). However, I agree, its not all that necessary and the boolean should suffice.

bh2smith:
Ok @fleupold - I am going to remove the cancellation date from the meta data, but the reason it was introduced was to test that the value isn't overwritten if the order is canceled twice. I will have to remove this part from the test because we can't check this without directly querying for the date field from the DB.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FHw8vYF4DNRCKY%2Fgiphy.gif&token=shipit" data-gitmeme-token="shipit"></a>

PR #333, opened 2021-02-24
Use different internal fee validity to reduce DB writes

fleupold:
Currently, we write one record per fee request to storage. @vkgnosis pointed out that this could become a DOS vector. As a countermeasure, this PR introduce the concept of an "internal validity" (two minutes) which is used to store records in the database, and an "external validity" of 1 minute, which is the official value we propagate to the client.

If a fee request for a token comes in we first check if we have already made an estimate for which the internal validity is still larger than the official validity we plan to report (that is if we have done an estimate for the same token in the last minute). In this case we reuse this estimate instead of creating and persisting a new one.

### Test Plan
Unit tests still work

### DIFF
--- a/orderbook/src/fee.rs
+++ b/orderbook/src/fee.rs
@@ -62,10 +62,9 @@ impl MinFeeCalculator {
     // Returns an error if there is some estimation error and Ok(None) if no information about the given
     // token exists
     pub async fn min_fee(&self, token: H160) -> Result<Option<Measurement>> {
-        let official_valid_until =
-            (self.now)() + Duration::seconds(STANDARD_VALIDITY_FOR_FEE_IN_SEC);
-        let internal_valid_until =
-            (self.now)() + Duration::seconds(PERSISTED_VALIDITY_FOR_FEE_IN_SEC);
+        let now = (self.now)();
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):

orderbook/src/fee.rs
```diff
@@ -44,6 +62,31 @@ impl MinFeeCalculator {
...
+        let min_fee = match self.compute_min_fee(token).await? {
+            Some(fee) => fee,
+            None => return Ok(None),
+        };
+
+        let _ = self
+            .measurements
+            .save_fee_measurement(token, internal_valid_until, min_fee)
+            .await;
+        Ok(Some((min_fee, official_valid_until)))
```
bh2smith, line 86: So I guess the meat of this logical change is right here, where we check the DB for a non-expired min fee and use that if available. Do we expect that the the validity period will be something we inevitably have to maintain as the ecosystem changes?

bh2smith (approved):

orderbook/src/fee.rs
fleupold: Yes, the API will have to guarantee that it accepts a fee that it originally quoted for some time (to allow for the time between quoting and signing the order without rejecting the order if the gas price went up in the meantime). The concrete duration of this time may change.

PR #337, opened 2021-02-25
[Trivial] Remove Arc<Database>

fleupold:
Feedback to remove this unnecessary change in #314 was not incorporated.

### Test Plan
CI


vkgnosis (approved):

bh2smith (commented):
Sorry, I must not have understood the issue here. I really thought I had taken care of it... Although, it was only in the e2e-test that I thought we didn't need Arc. I don't really understand how it was removed everywhere, my compiler was complaining.

PR #340, opened 2021-03-02
[Baseline 1] Add in/out estimation to Uniswap Pool struct

fleupold:
This PR implements the get_amounts_in/out methods based on the Uniswap smart contract logic. I added fee information to the pool struct, so that the in/out computation can be fee aware.

### Test Plan
Some unit tests bases on [this simulation](https://docs.google.com/spreadsheets/d/1K-tu5K43i3gBdlGs_LJNqgMGjEbm6j0Uf9YxxgoMJQg/edit#gid=0).

### DIFF
--- a/shared/src/uniswap_pool.rs
+++ b/shared/src/uniswap_pool.rs
@@ -1,7 +1,8 @@
 use std::collections::HashSet;
 
 use contracts::{UniswapV2Factory, UniswapV2Pair};
-use ethcontract::{batch::CallBatch, Http, Web3, H160};
+use ethcontract::{batch::CallBatch, Http, Web3, H160, U256};
+use num::rational::Rational;
 use web3::signing::keccak256;
 
 use hex_literal::hex;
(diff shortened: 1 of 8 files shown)


shared/src/uniswap_pool.rs
```diff
@@ -19,10 +20,72 @@ pub trait PoolFetching: Send + Sync {
     async fn fetch(&self, token_pairs: HashSet<TokenPair>) -> Vec<Pool>;
 }
 
-#[derive(Clone)]
+#[derive(Clone, Hash)]
 pub struct Pool {
     pub tokens: TokenPair,
     pub reserves: (u128, u128),
+    pub fee: Rational,
```
vkgnosis, line 27: Rational is Ratio<isize>. The fee has to be positive so we could use Ratio<usize> or Ratio<u32>. Just pointing out, not super important.

shared/src/uniswap_pool.rs
```diff
@@ -19,10 +20,72 @@ pub trait PoolFetching: Send + Sync {
...
+        } else {
+            assert!(token_in == self.tokens.get().1, "Token not part of pool");
+            (
+                U256::from(self.reserves.1),
+                U256::from(self.reserves.0),
+                self.tokens.get().0,
+            )
+        };
+
+        let amount_in_with_fee = amount_in * (self.fee.denom() - self.fee.numer());
```
vkgnosis, line 57: I think this and the other method are missing some overflow checks like in https://github.com/gnosis/gp-v2-services/blob/c7b301a4fdfe1f92b691a5e5bf64af4ecee31428/solver/src/uniswap.rs (form an unmerged PR) .
Might also be nice to have amount_out and amount_in from that file with the way the arguments don't include any H160. That way those methods can't panic and testing is a bit nicer and we have the match on the H160 separately from the math.

vkgnosis (approved):

shared/src/uniswap_pool.rs
fleupold: 👍 on the missing overflow check. 

For the publicly facing interface, I find it nice to give callers the opportunity to get an amount without having to query the right reserve amount themselves from the outside. Would you suggest to split this into two methods then?

vkgnosis: Right you can still keep the current public method but have the inner in addition (maybe private).

PR #346, opened 2021-03-03
[MultiSolver 1] Compute objective value for settlement

fleupold:
Implements surplus calculation for buy and sell orders according to 6a, 6b of page 10 https://github.com/gnosis/dex-research/blob/master/BatchAuctionOptimization/batchauctions.pdf. 

For now the objective value is just the sum of surplus. Note, that this is prone to scaling issues (as the price vector could be arbitrarily scaled by a factor to also scale this value). The question on how to normalize the price vector of different settlements to be in the same base unit is still subject to discussion.

In practice, this currently leads to favouring the naive solver (as token prices are denominated in reserve amounts which are likely higher than the traded amounts) which is ok as it's the only one supporting COW at the moment.

### Test Plan
Unit tests at the end of the file. @twalth3r could you double check they are expressing the correct expectations?

### DIFF
--- a/solver/src/settlement.rs
+++ b/solver/src/settlement.rs
@@ -34,6 +34,26 @@ impl Trade {
             fee_discount: 0,
         }
     }
+
+    // The difference between the minimum you were willing to buy/maximum you were willing to sell, and what you ended up buying/selling
+    pub fn surplus(&self, sell_token_price: U256, buy_token_price: U256) -> Option<U256> {
+        match self.order.kind {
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):

twalth3r (approved):
Double checked and seems all good! You could factor out the `executed_amounts` in both equations, but doesn't matter for correctness.

bh2smith (approved):

PR #347, opened 2021-03-03
[MultiSovler 2] Allow driver to work with multiple solvers

fleupold:
Depends on #346. Now that we can rank different settlements, we can have multiple solver "compete" internally for the best solution. This PR makes the driver take a vector of solvers and execute each one of them in parallel, afterwards sorting the resulting solutions by objective value and submitting the best one.

### Test Plan
Ran the binary locally and saw that we can both handle a multi-hop uniswap trade (UniswapBaseline) and a direct pair COW trade (NaiveSolver) in the same process.

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -9,7 +9,7 @@ use anyhow::{Context, Result};
 use contracts::GPv2Settlement;
 use futures::future::join_all;
 use gas_estimation::GasPriceEstimating;
-use std::time::Duration;
+use std::{cmp::Reverse, time::Duration};
 use tracing::info;
 
 // There is no economic viability calculation yet so we're using an arbitrary very high cap to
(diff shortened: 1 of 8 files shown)


solver/src/driver.rs
```diff
@@ -71,34 +73,56 @@ impl Driver {
...
+                    "{} found solution with objective value: {}",
+                    solver,
+                    settlement.objective_value()
+                );
+                Some((solver, settlement))
+            })
+            .collect();
+
+        // Sort by key yields ascending order, but we want to start with the highest objective value, so reverse iterator in loop
+        settlements.sort_by_key(|(_, settlement)| settlement.objective_value());
```
vkgnosis, line 101: You can use `std::cmp::Reverse` to wrap objective_value and avoid having to reverse the iterator

vkgnosis (approved):

PR #348, opened 2021-03-04
Use UniswapSolver for fee estimation (sellToken -> ETH)

fleupold:
This PR changes the price estimation trait to also take an amount and order type (buy vs sell). This is important to account for more accurate fee estimations of our baseline solution (e.g. large trades might make it worthwhile going over multiple hops, but would cost more gas to execute).

While that particular benefit requires a change in the fee endpoint (we need to pass buy token, amount and order type as well), it can immediately benefit the estimation of how much sell token we need to convert for ETH by pricing an ETH buy order from the sell token with the amount of ETH needed for the settlement (gas_price * GAS_PER_ORDER).

The PR is using the shared baseline Uniswap solver logic and as such also enabling sell tokens which don't have a direct pool to the native token.

This should make the component ready to be used for a general price estimation endpoint (cc @alfetopito)

### Test Plan
Unit tests

### DIFF
--- a/orderbook/src/fee.rs
+++ b/orderbook/src/fee.rs
@@ -90,7 +90,12 @@ impl MinFeeCalculator {
         let gas_price = self.gas_estimator.estimate().await?;
         let token_price = match self
             .price_estimator
-            .estimate_price(token, self.native_token)
+            .estimate_price(
+                token,
+                self.native_token,
+                U256::from_f64_lossy(gas_price * GAS_PER_ORDER),
+                model::order::OrderKind::Buy,
+            )
             .await
         {
             Ok(price) => price,
@@ -148,6 +153,7 @@ impl MinFeeStoring for InMemoryFeeStore {
 #[cfg(test)]
 mod tests {
     use chrono::Duration;
+    use model::order::OrderKind;
     use std::sync::Arc;
 
     use super::*;
@@ -155,7 +161,7 @@ mod tests {
     struct FakePriceEstimator(f64);
     #[async_trait::async_trait]
     impl PriceEstimating for FakePriceEstimator {
-        async fn estimate_price(&self, _: H160, _: H160) -> Result<f64> {
+        async fn estimate_price(&self, _: H160, _: H160, _: U256, _: OrderKind) -> Result<f64> {
             Ok(self.0)
         }
     }
--- a/orderbook/src/fee.rs
+++ b/orderbook/src/fee.rs
@@ -100,7 +100,10 @@ impl MinFeeCalculator {
             .await
         {
             Ok(price) => price,
-            Err(_) => return Ok(None),
+            Err(err) => {
+                tracing::warn!("Failed to estimate sell token price: {}", err);
+                return Ok(None);
+            }
         };
 
         Ok(Some(U256::from_f64_lossy(fee_in_eth * token_price)))
--- a/orderbook/src/fee.rs
+++ b/orderbook/src/fee.rs
@@ -88,12 +88,13 @@ impl MinFeeCalculator {
 
     async fn compute_min_fee(&self, token: H160) -> Result<Option<U256>> {
         let gas_price = self.gas_estimator.estimate().await?;
+        let fee_in_eth = gas_price * GAS_PER_ORDER;
         let token_price = match self
             .price_estimator
             .estimate_price(
                 token,
                 self.native_token,
-                U256::from_f64_lossy(gas_price * GAS_PER_ORDER),
+                U256::from_f64_lossy(fee_in_eth),
                 model::order::OrderKind::Buy,
             )
             .await
@@ -102,9 +103,7 @@ impl MinFeeCalculator {
             Err(_) => return Ok(None),
         };
 
-        Ok(Some(U256::from_f64_lossy(
-            gas_price * token_price * GAS_PER_ORDER,
-        )))
+        Ok(Some(U256::from_f64_lossy(fee_in_eth * token_price)))
     }
 
     // Returns true if the fee satisfies a previous not yet expired estimate, or the fee is high enough given the current estimate.
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -11,7 +11,9 @@ use orderbook::{
     serve_task, verify_deployed_contract_constants,
 };
 use shared::uniswap_pool::PoolFetcher;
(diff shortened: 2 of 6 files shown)


vkgnosis (approved):

orderbook/src/price_estimate.rs
```diff
@@ -1,56 +1,155 @@
...
+            sell_token,
+            buy_token,
+            buy_amount,
+            |amount, path, pools| Reverse(estimate_sell_amount(amount, path, pools)),
+            estimate_sell_amount,
+        )
+        .await
+    }
+
+    async fn best_execution<A, C, O>(
```
alfetopito, line 109: Is it common in Rust to use single letter generic types? (I assume are generic types, not that far yet in the Rust book).

It's a bit hard to read IMO

fleupold: Technically it could be any CamelCase, but in practice you see mostly one letter types. Here is a somewhat related discussion: https://users.rust-lang.org/t/one-character-generics/12317

I'm happy to rename it.

alfetopito: I see there's no consensus on any particular approach.

I personally prefer slightly more descriptive names when the types are not immediately clear (like when a single generic type `T`, or key value types `K, V`).

But that's my outsider view of someone who's not familiar with the code base nor the language.

orderbook/src/fee.rs
```diff
@@ -88,18 +88,25 @@ impl MinFeeCalculator {
 
     async fn compute_min_fee(&self, token: H160) -> Result<Option<U256>> {
         let gas_price = self.gas_estimator.estimate().await?;
+        let fee_in_eth = gas_price * GAS_PER_ORDER;
```
bh2smith, line 91: Assuming that `GAS_PER_ORDER` is a constant upper bound on the maximum gas an order could cost, how did we arrive at this value? It could also be that the constant is really the gas cost for just settling the single order, but wouldn't it be more appropriate to factor in the distributed fee incurred by executing a bunch of different interactions. If so, then we would also have to limit the scope of interactions, since they can essentially run arbitrary code. With this in mind, I would think that the gas per order is some sort of upper bound determined by the block gas limit over the max number of orders which can be matched.

orderbook/src/main.rs
```diff
@@ -92,11 +94,18 @@ async fn main() {
     )
     .await
     .expect("failed to create gas price estimator");
-    let price_estimator = UniswapPriceEstimator::new(Box::new(PoolFetcher {
-        factory: uniswap_factory,
-        web3,
-        chain_id,
-    }));
+
+    let mut base_tokens = HashSet::from_iter(args.shared.base_tokens);
```
bh2smith, line 98: So base tokens are something we specify as a runtime argument? What types of base tokens do we have in mind? Is this just a short list of reputable tokens or some very long list of everything we consider potentially tradable?

orderbook/src/price_estimate.rs
```diff
@@ -1,56 +1,155 @@
...
-            None => return Ok(1.0),
-        };
-        let pool = self
+    async fn estimate_price(
+        &self,
+        sell_token: H160,
+        buy_token: H160,
+        amount: U256,
+        kind: OrderKind,
+    ) -> Result<f64> {
```
bh2smith, line 53: Could we include a doc comment on this method about which price is being estimated? I mean if its a sell order, then I assume the price is that of the buy token relative to the sell token (and vice-versa?). I just have the impression that the method would be self explanatory if, say, there were only a single token being passed in (but then it would be pegged to some reference token). Perhaps this obscurity could be mitigated by renaming the input parameters `buy_token -> token` and `sell_token -> reference_token` or maybe we could use `base` and `quote`.

orderbook/src/price_estimate.rs
```diff
@@ -68,33 +167,95 @@ mod tests {
...
+                .unwrap(),
+            1.0
+        );
+        assert_approx_eq!(
+            estimator
+                .estimate_price(token_a, token_a, U256::exp10(18), OrderKind::Sell)
+                .await
+                .unwrap(),
             1.0
         );
```
bh2smith, line 191: I like these tests. price of X relative to X is 1 no matter if you're buying or selling.

orderbook/src/price_estimate.rs
```diff
@@ -68,33 +167,95 @@ mod tests {
...
         assert_approx_eq!(
-            estimator.estimate_price(token_a, token_b).await.unwrap(),
-            10.0
+            estimator
+                .estimate_price(token_a, token_b, U256::exp10(18), OrderKind::Buy)
+                .await
+                .unwrap(),
+            10.03,
+            1.0e-4
```
bh2smith, line 198: I assume these prices came from the pool's initial setup `(10u128.pow(30), 10u128.pow(29))`

bh2smith (approved):
Love this! Just a few general comments and one spot where I think we could clarify what to expect from the price estimate method.

orderbook/src/price_estimate.rs
bh2smith, line 53: Could we include a doc comment on this method about which price is being estimated? I mean if its a sell order, then I assume the price is that of the buy token relative to the sell token (and vice-versa?). I just have the impression that the method would be self explanatory if, say, there were only a single token being passed in (but then it would be pegged to some reference token). Perhaps this obscurity could be mitigated by renaming the input parameters buy_token -> token and sell_token -> reference_token or maybe we could use base and quote.

orderbook/src/price_estimate.rs
```diff
@@ -1,56 +1,155 @@
...
+    }
+}
+
+impl UniswapPriceEstimator {
+    pub async fn best_execution_sell_order(
+        &self,
+        sell_token: H160,
+        buy_token: H160,
+        sell_amount: U256,
+    ) -> Result<(Vec<H160>, U256)> {
```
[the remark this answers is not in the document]
fleupold: When we implement #350 we need to know how many hops the best route entails to properly estimate the gas fee of this order.

orderbook/src/main.rs
fleupold: Base tokens are copied from here: https://github.com/gnosis/gp-v2-swap/blob/a62ded4838775c8c53f934cd3d957e3704fe85ad/src/constants/index.ts#L66-L70

Yes, it's just a list of most liquid tokens for each network.

PR #350, opened 2021-03-04
[API] More expressive fee route

fleupold:
cc @anxolin @W3stside @alfetopito 

This PR suggests a change to the fee API (technically it's creating a new endpoint so we can keep the old one running for a bit).

The main reason is that we cannot accurately estimate the min fee if just given the sell token. The minimal fee should be equivalent to the "baseline solution" (the simple non COW, non fancy interactions path that is in line with the price estimate we show in the UI). The issue is that this path can be very sensitive in terms of execution costs (e.g. a Uniswap direct pair trade is less than half as expensive in terms of gas as a Uniswap two hop trade).

Therefore I believe we need more information about the order (in particular the buy token and the amount intended to trade).

Looking forward to hearing your thoughts, concerns or alternative suggestions.

### DIFF
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -160,9 +160,9 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/FeeInformation"
+                $ref: "#/components/schemas/LegacyFeeInformation"
         404:
-          description: sellToken not existent
+          description: sellToken non-existent
   /api/v1/trades:
     get:
       summary: Get existing Trades.
@@ -239,7 +239,7 @@ paths:
               schema:
                 $ref: "#/components/schemas/FeeInformation"
         404:
-          description: Token not existent or not connected to native token
+          description: Token non-existent or not connected to native token
 components:
   schemas:
     Address:
@@ -255,6 +255,20 @@ components:
       type: string
       example: "1234567890"
     FeeInformation:
+      description: |
+        Provides the information to calculate the fees.
+      type: object
+      properties:
+        expirationDate:
+          description: |
+            Expiration date of the offered fee. Order service might not accept
+            the fee after this expiration date. Encoded as ISO 8601 UTC.
+          type: string
+          example: "2020-12-03T18:35:18.814523Z"
+        amount:
+          description: Absolute amount of fee charged per order in specified sellToken
+          $ref: "#/components/schemas/TokenAmount"
+    LegacyFeeInformation:
       description: |
         Provides the information to calculate the fees.
       type: object
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -204,6 +204,42 @@ paths:
                 type: array
                 items:
                   $ref: "#/components/schemas/Order"
+  /api/v1/fee:
+    get:
+      description: |
+        The fee that is charged for placing an order.
+        The fee is described by a minimum fee - in order to cover the gas costs for onchain settling - and
+        a feeRatio charged to the users for using the service.
+      parameters:
+        - name: sellToken
+          in: query
+          schema:
+            $ref: "#/components/schemas/Address"
+          required: true
+        - name: buyToken
+          in: query
+          schema:
+            $ref: "#/components/schemas/Address"
+          required: true
+        - name: amount
+          in: query
+          schema:
+            $ref: "#/components/schemas/TokenAmount"
+          required: true
+        - name: kind
+          in: query
+          schema:
+            $ref: "#/components/schemas/OrderType"
+          required: true
+      responses:
+        200:
+          description: the fee
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/FeeInformation"
+        404:
+          description: Token not existent or not connected to native token
 components:
   schemas:
     Address:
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):
Code wise this is fine, so approving based on that but ofc we can still discuss.

I feel being more accurate with the fee is probably a good thing because it makes the fee more accurate so instead of having to charge a higher fee on average we charge exactly what is needed. Disadvantage is that it is less predictable when you see in the frontend that you change the amount a little and it makes fee jump.

orderbook/openapi.yml
```diff
@@ -204,6 +204,42 @@ paths:
...
+        - name: buyToken
+          in: query
+          schema:
+            $ref: "#/components/schemas/Address"
+          required: true
+        - name: amount
+          in: query
+          schema:
+            $ref: "#/components/schemas/TokenAmount"
+          required: true
```
alfetopito, line 228: 1. What is this amount's unit? Integer atoms or decimal units
2. Is it the sell amount or buy amount? I assume it matches the kind (sell amount for sell order, etc)

fleupold: Yes, it would be the same as in the order endpoint for now. For convenience it would probably be nice to let people flip-flop between base units and atoms but for now we don't have token info in the backend.

anxolin (commented):
If the backend wants to give more fine-grained estimation, makes all the sense 👌.

The endpoint proposal looks good to me. The only part I don't really get, is why do we need two split the fee in to it's components  ("feeFactor" and "protocolFee"). I think, it made sense before, because we were giving the clients the components for dynamically calculate the fee using a simple formula, but this new endpoint moves that responsibility to the API, so I think it doesn't make sense any more to return the "feeFactor" and "protocolFee" (we could put them together into just `fee`). This way, clients become dummies, asking for a fee, and applying it. 

Other than that, I only have the concern for the performance of the possible implementation. If clients become dummies, this endpoint needs to be very fast, otherwise it will impact in the UX. Just good to keep in mind for the implementation.

orderbook/openapi.yml
W3stside: seconding allowing flip flopping, can obv be later no issue there

orderbook/openapi.yml
```diff
@@ -204,6 +204,42 @@ paths:
...
+      responses:
+        200:
+          description: the fee
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/FeeInformation"
+        404:
+          description: Token not existent or not connected to native token
```
W3stside, line 242:
```suggestion
          description: Token non-existent or not connected to native token
```

W3stside:
```suggestion
          description: Token does not exist or not connected to native token
```

W3stside:
> If the backend wants to give more fine-grained estimation, makes all the sense .
> 
> The endpoint proposal looks good to me. The only part I don't really get, is why do we need two split the fee in to it's components ("feeFactor" and "protocolFee"). I think, it made sense before, because we were giving the clients the components for dynamically calculate the fee using a simple formula, but this new endpoint moves that responsibility to the API, so I think it doesn't make sense any more to return the "feeFactor" and "protocolFee" (we could put them together into just `fee`). This way, clients become dummies, asking for a fee, and applying it.
> 
> Other than that, I only have the concern for the performance of the possible implementation. If clients become dummies, this endpoint needs to be very fast, otherwise it will impact in the UX. Just good to keep in mind for the implementation.

where are you seeing the split fee? from what I can see the `FeeInformation` schema (response shape) hasn't changed:
```yml
FeeInformation:
      description: |
        Provides the information to calculate the fees.
      type: object
      properties:
        expirationDate:
          description: |
            Expiration date of the offered fee. Order service might not accept
            the fee after this expiration date. Encoded as ISO 8601 UTC.
          type: string
          example: "2020-12-03T18:35:18.814523Z"
        minimalFee:
          description: Absolute amount of minimal fee charged per order in specified sellToken
        feeRatio:
          description: The fee ratio charged on a sellAmount. Denoted in basis points
          example: 10
          type: number
          format: int32
```

anxolin:
> where are you seeing the split fee? from what I can see the FeeInformation schema (response shape) hasn't changed:

That's my point :) 
It hasn't change, so it still contains the 2 components of the fee.

orderbook/openapi.yml
```diff
@@ -255,6 +255,20 @@ components:
...
+        expirationDate:
+          description: |
+            Expiration date of the offered fee. Order service might not accept
+            the fee after this expiration date. Encoded as ISO 8601 UTC.
+          type: string
+          example: "2020-12-03T18:35:18.814523Z"
+        amount:
+          description: Absolute amount of fee charged per order in specified sellToken
+          $ref: "#/components/schemas/TokenAmount"
+    LegacyFeeInformation:
```
anxolin, line 271: nit: I don't care that much, since u will kill this as soon as we are migrated. But you could just use the extend it https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism 

totally fine keeping it as u suggest

anxolin (approved):

orderbook/openapi.yml
fleupold: I'd do it if LegacyFeeInformation directly extended the new FeeInformation. However, there we renamed min_fee to amount so I'd have to introduce a third type that both can extend, just to merge them into one again later.

PR #360, opened 2021-03-09
Allow for Uniswap Spot price estimation

fleupold:
In #348 I handled the case when amount is 0 by just rounding it up to 1 (the smallest amount to buy/sell). While this leads to arithmetically valid results they are not very useful in practice because of integer division rounding errors for such small amounts (e.g if token B is worth more than token A, selling 1 of token A gives you 0 token B).

This PR adds functionality to the Pool struct, the uniswap shared methods and the price estimator to compute the spot price (relation of reserves) for a given path. This allows more accurate price estimation for 0 amounts which might also be useful when estimating the current value of a certain token (e.g. for display in the block explorer)

### Test Plan
Added unit tests for this case.

### DIFF
--- a/orderbook/src/price_estimate.rs
+++ b/orderbook/src/price_estimate.rs
@@ -2,9 +2,11 @@ use anyhow::{anyhow, Result};
 use ethcontract::{H160, U256};
 use model::{order::OrderKind, TokenPair};
 use shared::{
+    conversions::big_rational_to_float,
     uniswap_pool::{Pool, PoolFetching},
     uniswap_solver::{
-        estimate_buy_amount, estimate_sell_amount, path_candidates, token_path_to_pair_path,
+        estimate_buy_amount, estimate_sell_amount, estimate_spot_price, path_candidates,
+        token_path_to_pair_path,
     },
 };
 use std::{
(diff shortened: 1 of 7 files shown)


orderbook/src/price_estimate.rs
```diff
@@ -44,6 +46,7 @@ impl UniswapPriceEstimator {
 impl PriceEstimating for UniswapPriceEstimator {
     // Estimates the price using the direct pool between sell and buy token.
     // Returns an error if no pool exists between sell and buy token.
+    // Incorporates uniswap fee unless amount is 0 in which case it returns the best spot price.
```
vkgnosis, line 49: For the spot price it looks like the OrderKind does not matter. Maybe it would make sense to make this a separate function `estimate_spot_price` in that case? I'm not sure, just something I noticed when reading.

vkgnosis (approved):

orderbook/src/price_estimate.rs
fleupold: We could do that. What should then be the behavior of estimate price when invoked with 0 amount? Would it call into that method (in this case it would just be an additional method on PriceEstimating) or should it returns an error? I feel the rounding to 1 we do right now will lead to very bad results.

vkgnosis: I would say that you are not allowed to call estimate_price with 0 amount because it is not a meaningful value for this function.

fleupold: Right, I'm slightly worried that in this case the caller will have to always do a if amount == 0 estimate_price else estimate spot_price (since e.g. the amount for gas estimation can be 0 if the gas price is 0, or U256 user input can always be 0.

That's why I'm slightly favouring handling it inside the estimate_price abstraction directly (although I see your point that it's not great to implicitly apply fundamentally different logic for slightly different values).

vkgnosis: That's reasonable too. I'm fine with either decision. Mostly wanted to point out the idea in case you hadn't thought about it.

PR #361, opened 2021-03-10
Updates http solver to reflect recent API changes.

marcovc:
Namely:

* Removes default_fee
* Adds fee (0.1%) and cost (fixed gas cost) for orders.
* Adds cost (fixed gas cost) for uniswaps.
* Adds execution plan to uniswaps.

### Test Plan
Unit tests.

TODO:
Some wires are missing:
Is there a common interface for the execution plan?
Same for costs?

### DIFF
--- a/solver/src/http_solver.rs
+++ b/solver/src/http_solver.rs
@@ -118,6 +118,12 @@ impl HttpSolver {
                     buy_amount: order.buy_amount,
                     allow_partial_fill: order.partially_fillable,
                     is_sell_order: matches!(order.kind, OrderKind::Sell),
+                    // TODO: map order fee and fixed cost
+                    fee: 0.0,
+                    cost: CostModel {
+                        amount: 0,
+                        token: self.token_to_string(&order.sell_token)
+                    }
                 };
                 (index.clone(), order)
             })
@@ -141,6 +147,11 @@ impl HttpSolver {
                     balance1: amm.reserves.0,
                     balance2: amm.reserves.1,
                     fee: *amm.fee.numer() as f64 / *amm.fee.denom() as f64,
+                    // TODO: map uniswap fixed cost
+                    cost: CostModel {
+                        amount: 0,
+                        token: self.token_to_string(&amm.tokens.get().0)
+                    },
                     mandatory: false,
                 };
                 (index.clone(), uniswap)
@@ -167,7 +178,6 @@ impl HttpSolver {
             tokens: self.token_models(&tokens),
             orders: self.order_models(&limit_orders),
             uniswaps: self.amm_models(&amm_orders),
-            default_fee: 0.0,
         };
         let context = SettlementContext {
             tokens,
@@ -338,6 +348,8 @@ mod tests {
         let uniswap = settled.uniswaps.values().next().unwrap();
         assert!(uniswap.balance_update1 < 0);
         assert_eq!(uniswap.balance_update2 as u128, base(2));
+        assert_eq!(uniswap.exec_plan.sequence, 0);
+        assert_eq!(uniswap.exec_plan.position, 0);
 
         assert_eq!(settled.prices.len(), 2);
     }
--- a/solver/src/http_solver.rs
+++ b/solver/src/http_solver.rs
@@ -122,7 +122,7 @@ impl HttpSolver {
                     fee: 0.0,
                     cost: CostModel {
                         amount: 0,
-                        token: self.token_to_string(&order.sell_token)
+                        token: self.token_to_string(&order.sell_token),
                     }
                 };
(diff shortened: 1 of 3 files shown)


vkgnosis:
> Is there a common interface for the execution plan?

If you mean whether there is already a higher level abstraction around something like execution plans then no. What I think we should do is in the function where you added ` // TODO: handle execution plan.` sort the amm orders according to the execution plan and then "execute" them as the function already does, just in the new order.

vkgnosis (approved):

solver/src/http_solver.rs
```diff
@@ -141,6 +147,11 @@ impl HttpSolver {
                     balance1: amm.reserves.0,
                     balance2: amm.reserves.1,
                     fee: *amm.fee.numer() as f64 / *amm.fee.denom() as f64,
+                    // TODO: map uniswap fixed cost
+                    cost: CostModel {
+                        amount: 0,
+                        token: self.token_to_string(&amm.tokens.get().0),
```
fleupold, line 153: I believe here the cost should be in the native token.

solver/src/http_solver/model.rs
```diff
@@ -42,6 +43,12 @@ pub struct TokenInfoModel {
     pub decimals: u32,
 }
 
+#[derive(Debug, Serialize)]
+pub struct CostModel {
+    pub amount: u128,
```
fleupold, line 48: can the cost not be negative in case of EOA orders where the fee is larger than their own marginal execution cost (since e.g. the fee was estimated with a multi-hop baseline in mind). Thus including this order has a negative cost.

solver/src/http_solver/model.rs
```diff
@@ -67,4 +74,11 @@ pub struct UpdatedUniswapModel {
     pub balance_update1: i128,
     #[serde(with = "serde_with::rust::display_fromstr")]
     pub balance_update2: i128,
+    pub exec_plan: ExecPlanCoordsModel,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct ExecPlanCoordsModel {
```
fleupold, line 81: nit: I'm not a fan of abbreviations (they save you a little time to type but take more time to comprehend later on). Can we call this ExecutionPlanModel or ExecutionPlanCoordinates or ExecutionPlanCoordinatesModel?

vkgnosis:
That's true. In the end we do need to send a specific sequence to the smart contract so we have to have this information loss at some point. Since other solvers aren't aware of different execution plans doing it directly in the code I mentioned will be practical for the near future.

Then about the costs: Iirc we want the cost to be expressed in the native token (ether) which HttpSolver already gets. So we still need to get the gas price and gas amount for uniswaps into the solver. I think we can do that in another PR too because it will be more involved with the rest of the code base.

fleupold (approved):
🎉  Nice first PR!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FvsdMcLXabvW7QQJf9r%2Fgiphy.gif&token=firststeps" data-gitmeme-token="firststeps"></a>

solver/src/http_solver/model.rs
twalth3r: There's a separate `fee` field to make things more explicit.

solver/src/http_solver.rs
twalth3r: For the optimization model, it can be any token (that is part of the batch and connected to the main trading island).

solver/src/http_solver/model.rs
marcovc: Not sure I understand. This models the fixed cost of including the order (the gas cost). Whatever the fee we charge to our customers is, our gas cost is always a positive number right?

twalth3r: I think initially, the idea was to add this cost field for Uniswaps, where it simply represents the gas cost. For orders, we also collect a fee, and we were considering to mix this fee into the order cost (thus making the cost negative). But later we decided to introduce a separate `fee` field for orders, so that `costs` indeed only represents the marginal execution cost.

twalth3r (approved):

solver/src/http_solver/model.rs
marcovc: Right, the fee in the AMM model is the fee the AMM charges to us, while the fee in the orders is the fee we charge our customers. They are both proportional, and they both affect the optimization. I don't mind call it differently though, do you have any suggestion?

fleupold: If I'm the only one confused by this I guess it's fine to keep it like this. Maybe `protocolFee` would express that this is the fee for our protocol?

PR #364, opened 2021-03-10
[FeeRefactor] Prepare fee measurements for buy token and amount

fleupold:
First step towards implementing the change in #350

The database needs to be able to store fee measurements based on buy token and sell_amount (I assume buy order's amounts will be converted to sell amounts before performing fee estimation - this will allow reusing cache hits from sell orders).

For the moment all those fields are still nullable as long as we are maintaining the old endpoint (until the frontend has switched). In this PR the logic doesn't change and we fill the extra fields with `None`.

In the next PR I will implement the new route which will populate non None values. Note, that as long as the validation doesn't use buyToken and amount to verify the min fee, it has still not affect on whether a fee accepted when the order is created.

### Test Plan
Adjusted Unit Tests

### DIFF
--- a/orderbook/src/database/fees.rs
+++ b/orderbook/src/database/fees.rs
@@ -21,7 +21,7 @@ impl MinFeeStoring for Database {
             "INSERT INTO min_fee_measurements (sell_token, buy_token, sell_amount, expiration_timestamp, min_fee) VALUES ($1, $2, $3, $4, $5);";
         sqlx::query(QUERY)
             .bind(sell_token.as_bytes())
-            .bind(buy_token.map(|t| t.as_bytes().to_owned()))
+            .bind(buy_token.as_ref().map(|t| t.as_bytes()))
             .bind(sell_amount.map(|a| u256_to_big_decimal(&a)))
             .bind(expiry)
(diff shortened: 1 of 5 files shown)


orderbook/src/database/fees.rs
```diff
@@ -11,14 +11,18 @@ use ethcontract::{H160, U256};
...
         expiry: DateTime<Utc>,
         min_fee: U256,
     ) -> Result<()> {
         const QUERY: &str =
-            "INSERT INTO min_fee_measurements (token, expiration_timestamp, min_fee) VALUES ($1, $2, $3);";
+            "INSERT INTO min_fee_measurements (sell_token, buy_token, sell_amount, expiration_timestamp, min_fee) VALUES ($1, $2, $3, $4, $5);";
         sqlx::query(QUERY)
-            .bind(token.as_bytes())
+            .bind(sell_token.as_bytes())
+            .bind(buy_token.map(|t| t.as_bytes().to_owned()))
```
vkgnosis, line 24: this is how we can avoid converting to vector
```suggestion
            .bind(buy_token.as_ref().map(|t| t.as_bytes()))
```

vkgnosis (approved):

PR #365, opened 2021-03-10
[FeeRefactor] Add the new route's implementation

fleupold:
Implements the changes in #350 

Prefixes the existing endpoint into legacy and creates new endpoint, which is passing non None informations into the MinFee calculator. Amounts are converted into sell amounts and rounded to the midpoint of their order of magnitude (e.g. 110 -> 500, etc.) to allow more cache hits (I'm not sure we actually want this).

The logic for the legacy endpoint is unchanged.

### Test Plan
Unit tests + testing locally that http://localhost:8080/api/v1/fee?sellToken=0x44fA8E6f47987339850636F88629646662444217&buyToken=0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d&amount=5000000&kind=buy

works and returns the same result as sell orders (it's trading wxDAI vs bridged xDAI)

### DIFF
--- a/orderbook/src/fee.rs
+++ b/orderbook/src/fee.rs
@@ -23,13 +23,22 @@ pub trait MinFeeStoring: Send + Sync {
     // Stores the given measurement. Returns an error if this fails
     async fn save_fee_measurement(
         &self,
-        token: H160,
+        sell_token: H160,
+        buy_token: Option<H160>,
+        sell_amount: Option<U256>,
         expiry: DateTime<Utc>,
         min_fee: U256,
     ) -> Result<()>;
 
     // Return a vector of previously stored measurements for the given token that have an expiry >= min expiry
-    async fn get_min_fee(&self, token: H160, min_expiry: DateTime<Utc>) -> Result<Option<U256>>;
+    // If buy_token or sell_amount is not specified, it will return the lowest estimate matching the values provided.
+    async fn get_min_fee(
+        &self,
+        sell_token: H160,
+        buy_token: Option<H160>,
+        sell_amount: Option<U256>,
+        min_expiry: DateTime<Utc>,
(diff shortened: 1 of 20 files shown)


orderbook/src/fee.rs
```diff
@@ -140,6 +158,29 @@ impl MinFeeCalculator {
...
+                    .await?,
+        ),
+        OrderKind::Sell => amount,
+    })
+}
+
+fn round_amount(amount: U256) -> U256 {
+    let order_of_magnitude = amount.to_string().len() - 1;
+    U256::exp10(order_of_magnitude) * 5
+}
```
vkgnosis, line 182: > Amounts are converted into sell amounts and rounded to the midpoint of their order of magnitude (e.g. 110 -> 500, etc.) to allow more cache hits (I'm not sure we actually want this).

I am not sure either. The way this is implemented here means that all possible sell amounts come in factors of 10 like 5, 50, 500, 5000 etc. Since it rounds a sell amount can be changed by up to a factor of 5. This is nice for the cache but has the drawback that fee is less accurate.
We could make it more accurate while keeping this type of rounding by using exponents of 2 instead of 10 with U256's `bits` and `pow` methods.

vkgnosis (approved):

orderbook/src/fee.rs
fleupold: After thinking about this a bit more, it might be better for now to not optimize cache hits and only add "bucketing" if we see we are getting too many requests. I think it's generally safer and easier to debug to not do the buy_amount -> sell_amount order conversion (as prices might change between the fee request and order placement) and store a model in the DB that is closer to the actual request. I will rework #364 to store amount and order kind instead of always assuming sell_orders.

I do like the suggestion of using power of two instead of power of tens.

PR #367, opened 2021-03-11
More expressive fee 2

fleupold:
We are currently unable to estimate fees for e.g. USDC: https://protocol-mainnet.dev.gnosisdev.com/api/v1/tokens/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48/fee

The reason is that #348 introduced a more sophisticated fee estimation that tries to compute the amount of sell token required to purchase the amount of native token required for the gas (via a buy order).

In our buy order logic we sort the possible uniswap paths by their required sell amount in reverse order (the less you have to sell the better). However, since this method returns an Option<U256> `Reverse` considers None to be the smaller (as such best) path. None however means no path exists. This leads to the effect that as long as there is one invalid path candidate (e.g. with a missing pool), we cannot estimate the fee. 

### Test Plan
Added a regression test and made sure the above request now returns a valid fee again.

### DIFF
--- a/database/sql/V005__extend_fee_measurements_with_order_details.sql
+++ b/database/sql/V005__extend_fee_measurements_with_order_details.sql
@@ -1,0 +1,5 @@
+ALTER TABLE min_fee_measurements RENAME COLUMN token TO sell_token;
+
+ALTER TABLE min_fee_measurements
+  ADD COLUMN buy_token bytea,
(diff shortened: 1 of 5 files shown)


fleupold:
Messed up the PR link in my VSCode. This PR was done on the wrong branch

PR #369, opened 2021-03-11
Make event sync optional

fleupold:
When restarting the database from scratch (e.g. to test migrations or switch between networks) it takes a significant time for the Event Updater to catch up (>10 minutes on my local machine). In the meantime running the e2e integration locally can lead to weird effects as e.g. trade events that come from recent test trades are only picked up after the event updater has caught up (until then the solver tries to submit the orders over and over again).

This PR adds an optional flag to the orderbook which makes it skip all past events. For local development where we only care about orders that we create once the orderbook/solver have been running this is fine and much faster

### Test Plan
Run orderbook with and without `--skip-event-sync`


vkgnosis (approved):
Thanks. I forgot about implementing this since we talked about it last.

PR #372, opened 2021-03-11
[Easy] Better error message when order already exists

fleupold:
Related [discussion on slack](https://gnosisinc.slack.com/archives/CPZA1AGKY/p1615464592060700)

At the moment we return a 500 internal server error when someone tries to place the same order twice. Internally we run into a primary key violation error in the DB which we swallow behind a general purpose internal error. This PR makes it so that we return the proper error reason.

### Test Plan
Place order once via the UI, copy the cURL request from dev tools, run again in shell and observe:

> {"errorType":"DuplicatedOrder","description":"order already exists"}

### DIFF
--- a/orderbook/src/database.rs
+++ b/orderbook/src/database.rs
@@ -20,6 +20,18 @@ pub struct Database {
     pool: PgPool,
 }
 
+#[derive(Debug)]
+pub enum InsertionError {
+    DuplicatedRecord,
+    DbError(sqlx::Error),
+}
+
+impl From<sqlx::Error> for InsertionError {
+    fn from(err: sqlx::Error) -> Self {
+        Self::DbError(err)
+    }
+}
+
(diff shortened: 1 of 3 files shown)


vkgnosis (approved):
It would be better to handle this in `insert_order` (`orderbook/src/database/orders.rs`) directly. There is already a comment there about this. That way there is no race condition because everything happens in one query.

I am still going to approve because this is an improvement just wanted to mention the proper fix.

fleupold:
Could you take another look. I moved the check into the same transaction as the insertion. Alternatively I could inspect the error message, but since it's not strongly typed it seems more error prone than checking for duplication specifically.

fleupold:
Now doing the check by inspecting the error code of the insert query. Please re-review 😄

vkgnosis (approved):

PR #373, opened 2021-03-11
Ranking solutions.

marcovc:
* Collects a vector of estimated prices from a PriceEstimating.
* Sends vector to objective_value()
* Changes base for computing objective from U256 -> BigRational

### Test Plan
Unit test.

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -22,7 +22,6 @@ use tracing::info;
 // protect against a gas estimator giving bogus results that would drain all our funds.
 const GAS_PRICE_CAP: f64 = 500e9;
 
-#[allow(dead_code)]
 pub struct Driver {
     settlement_contract: GPv2Settlement,
     orderbook: OrderBookApi,
@@ -32,6 +31,7 @@ pub struct Driver {
     gas_price_estimator: Box<dyn GasPriceEstimating>,
     target_confirm_time: Duration,
     settle_interval: Duration,
+    native_token: H160,
 }
 
 impl Driver {
@@ -45,6 +45,7 @@ impl Driver {
         gas_price_estimator: Box<dyn GasPriceEstimating>,
         target_confirm_time: Duration,
         settle_interval: Duration,
+        native_token: H160,
     ) -> Self {
         Self {
             settlement_contract,
@@ -55,6 +56,7 @@ impl Driver {
             gas_price_estimator,
             target_confirm_time,
             settle_interval,
+            native_token,
         }
     }
 
@@ -84,9 +86,9 @@ impl Driver {
             .dedup()
             .collect();
 
-        // For ranking purposes it doesn't matter how the external price vector is scaled.
-        // If we use native_token here instead, should it come as another ctor parameter?
-        let denominator_token: H160 = tokens[0];
+        // For ranking purposes it doesn't matter how the external price vector is scaled,
+        // but native_token is used here anyway for better logging/debugging.
+        let denominator_token: H160 = self.native_token;
 
         let estimated_prices = self
             .price_estimator
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -1,10 +1,19 @@
-use crate::{liquidity::{LimitOrder, Liquidity, uniswap::UniswapLiquidity}, orderbook::OrderBookApi, settlement::Settlement, settlement_submission, solver::Solver};
+use crate::{
+    liquidity::{uniswap::UniswapLiquidity, LimitOrder, Liquidity},
+    orderbook::OrderBookApi,
+    settlement::Settlement,
+    settlement_submission,
+    solver::Solver,
+};
 use anyhow::{Context, Result};
 use contracts::GPv2Settlement;
 use futures::future::join_all;
 use gas_estimation::GasPriceEstimating;
 use itertools::Itertools;
+use num::BigRational;
+use primitive_types::H160;
 use shared::price_estimate::PriceEstimating;
+use std::collections::HashMap;
 use std::{cmp::Reverse, sync::Arc, time::Duration};
 use tracing::info;
 
@@ -58,25 +67,30 @@ impl Driver {
         }
     }
 
-    async fn collect_prices(&self, limit_orders: &Vec<LimitOrder>) {
+    // Should this go here or on price_estimate.rs?
+    async fn collect_estimated_prices(
+        &self,
+        limit_orders: &[LimitOrder],
+    ) -> Result<HashMap<H160, BigRational>> {
         // Computes set of traded tokens (limit orders only).
-        let tokens = limit_orders
-            .into_iter()
-            .flat_map(|lo| vec![lo.sell_token, lo.buy_token].iter())
+        let tokens: Vec<H160> = limit_orders
+            .iter()
+            .flat_map(|lo| vec![lo.sell_token, lo.buy_token].into_iter())
             .sorted()
-            .dedup().collect();
+            .dedup()
+            .collect();
 
-        //web3 = ...
+        // For ranking purposes it doesn't matter how the external price vector is scaled.
+        // If we use native_token here instead, should it come as another ctor parameter?
+        let denominator_token: H160 = tokens[0];
 
-        /*let native_token = WETH9::deployed(&web3)
+        let estimated_prices = self
+            .price_estimator
+            .estimate_prices(tokens.as_slice(), denominator_token)
             .await
-            .expect("couldn't load deployed native token");*/
+            .context("failed estimating prices")?;
 
-        let estimated_prices = self.price_estimator.best_execution_spot_prices(
-            tokens, native_token).await;
-        
-        
-
+        Ok(tokens.into_iter().zip(estimated_prices).collect())
     }
 
     pub async fn single_run(&mut self) -> Result<()> {
@@ -96,10 +110,13 @@ impl Driver {
         tracing::debug!("got {} AMMs", amms.len());
 
         let liquidity: Vec<Liquidity> = limit_orders
+            .clone()
             .into_iter()
             .map(Liquidity::Limit)
             .chain(amms.into_iter().map(Liquidity::Amm))
             .collect();
+
+        let estimated_prices = self.collect_estimated_prices(&limit_orders).await?;
 
         let mut settlements: Vec<(&Box<dyn Solver>, Settlement)> =
             join_all(self.solver.iter().map(|solver| {
@@ -113,16 +130,15 @@ impl Driver {
                 info!(
                     "{} found solution with objective value: {}",
                     solver,
-                    settlement.objective_value(/* estimated_prices*/)
+                    settlement.objective_value(&estimated_prices)
                 );
                 Some((solver, settlement))
             })
             .collect();
 
         // Sort by key in descending order
-        settlements.sort_by_key(|(_, settlement)| {
-            Reverse(settlement.objective_value(/* estimated_prices*/))
-        });
+        settlements
+            .sort_by_key(|(_, settlement)| Reverse(settlement.objective_value(&estimated_prices)));
         for (solver, settlement) in settlements {
             info!("{} computed {:?}", solver, settlement);
             if settlement.trades.is_empty() {
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -1,14 +1,9 @@
-use crate::{
-    liquidity::{uniswap::UniswapLiquidity, Liquidity},
-    orderbook::OrderBookApi,
-    settlement::Settlement,
-    settlement_submission,
-    solver::Solver,
-};
+use crate::{liquidity::{LimitOrder, Liquidity, uniswap::UniswapLiquidity}, orderbook::OrderBookApi, settlement::Settlement, settlement_submission, solver::Solver};
 use anyhow::{Context, Result};
 use contracts::GPv2Settlement;
 use futures::future::join_all;
 use gas_estimation::GasPriceEstimating;
+use itertools::Itertools;
 use shared::price_estimate::PriceEstimating;
 use std::{cmp::Reverse, sync::Arc, time::Duration};
 use tracing::info;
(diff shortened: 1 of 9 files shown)


solver/src/driver.rs
```diff
@@ -63,6 +68,35 @@ impl Driver {
...
+        let tokens: Vec<H160> = limit_orders
+            .iter()
+            .flat_map(|lo| vec![lo.sell_token, lo.buy_token].into_iter())
+            .sorted()
+            .dedup()
+            .collect();
+
+        // For ranking purposes it doesn't matter how the external price vector is scaled.
+        // If we use native_token here instead, should it come as another ctor parameter?
+        let denominator_token: H160 = tokens[0];
```
fleupold, line 89: I think it makes sense to use the native token here, as it will likely be the most liquid and therefore give the most accurate price estimations. Can be passed into the driver as ctor arg as you suggested.

solver/src/settlement.rs
```diff
@@ -205,98 +258,117 @@ mod tests {
     }
 
     #[test]
+    #[allow(clippy::just_underscores_and_digits)]
     fn test_buy_order_surplus() {
         // Two goods are worth the same (100 each). If we were willing to pay up to 60 to receive 50,
         // but ended paying the price (1) we have a surplus of 10 sell units, so a total surplus of 1000.
+
+        // I hope there is a better way :)
```
fleupold, line 266: You could define yourself a macro or a small helper function.

solver/src/settlement.rs
```diff
@@ -139,35 +180,47 @@ impl Settlement {
...
+    buy_amount_limit: &BigRational,
+    executed_amount: &BigRational,
+) -> Option<BigRational> {
+    let res = executed_amount
         .checked_mul(sell_amount_limit)?
         .checked_div(buy_amount_limit)?
         .checked_mul(sell_token_price)?
-        .checked_sub(executed_amount.checked_mul(buy_token_price)?)
+        .checked_sub(&executed_amount.checked_mul(buy_token_price)?)?;
+    // Should we simply return 0 when the order fails to satisfy the limit price,
```
fleupold, line 194: I think it's fine to treat this the same way as an overflow in the computation. We should probably log an error so we are being made aware of it.

solver/src/settlement.rs
```diff
@@ -111,26 +117,61 @@ impl Settlement {
...
+    // Objective is re-computed using external prices.
+    fn objective_value_v1(
+        &self,
+        external_prices: &HashMap<H160, BigRational>,
+    ) -> Option<BigRational> {
+        self.total_surplus(external_prices)
+    }
+
+    // Objective is scaled by (harmonic) mean of external/found price ratio.
+    #[allow(dead_code)]
```
fleupold, line 141: I think we should check this in on a separate branch/commit but not have dead_code in the main codebase.

fleupold (commented):
Looks really good! I don't think extra end to end tests are necessary. Maybe you can use the simple examples from https://docs.google.com/document/d/1Toc7DGA8z3pfSBeVjeF9xl5VsWRarx-8K2O5O_aSSDY/edit and implement them as a test case to make sure we have the relative ordering between two different solutions right?

solver/src/settlement.rs
```diff
@@ -205,98 +258,117 @@ mod tests {
     }
 
     #[test]
+    #[allow(clippy::just_underscores_and_digits)]
     fn test_buy_order_surplus() {
         // Two goods are worth the same (100 each). If we were willing to pay up to 60 to receive 50,
         // but ended paying the price (1) we have a surplus of 10 sell units, so a total surplus of 1000.
+
+        // I hope there is a better way :)
+        let _5000 = BigRational::from_u32(5000).unwrap();
```
vkgnosis, line 267: I don't think so because BigRational does not implement From<u32> and so on directly. I would use a small helper functions (or macro) here to avoid the repetition. Something like 

```rust
        fn r(u: u128) -> BigRational {
            BigRational::from_u128(u).unwrap()
        }

        let _5000 = r(5000);
```

and at that point you can put that directly in the assert instead of going through a variable.

solver/src/settlement.rs
```diff
@@ -111,26 +117,61 @@ impl Settlement {
...
                 .expect("Solution with trade but without price for sell token");
-            let buy_token_price = self
-                .clearing_prices
+            let buy_token_external_price = prices
                 .get(&trade.order.buy_token)
-                .expect("Solution with trade but without price for sell token");
-            acc?.checked_add(trade.surplus(*sell_token_price, *buy_token_price)?)
-        }) {
+                .expect("Solution with trade but without price for buy token");
+            acc?.checked_add(&trade.surplus(sell_token_external_price, buy_token_external_price)?)
```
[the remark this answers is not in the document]
marcovc: I thought about that too (btw I haven't found an easy to check that from the code since Ratio<T> is coded for a generic T which doesn't overflow for the specific case of T=BigInt).

However. I believe divisions by 0 are still coerced into None, so we would still need the Option...

solver/src/driver.rs
marcovc: Yes, but notice that we don't care how accurate the actual price estimations are for the purposes of ranking. On the other hand, it might be better for logging/debugging. Done here e41c35.

solver/src/settlement.rs
marcovc: Done in a2a4e.

solver/src/settlement.rs
marcovc: Done!

solver/src/settlement.rs
marcovc: Removed v2 here: f97dd2c.

solver/src/driver.rs
```diff
@@ -63,6 +68,35 @@ impl Driver {
...
+        &self,
+        limit_orders: &[LimitOrder],
+    ) -> Result<HashMap<H160, BigRational>> {
+        if limit_orders.is_empty() {
+            return Ok(hashmap![]);
+        }
+        // Computes set of traded tokens (limit orders only).
+        let tokens: Vec<H160> = limit_orders
+            .iter()
+            .flat_map(|lo| vec![lo.sell_token, lo.buy_token].into_iter())
```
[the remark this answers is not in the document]
marcovc: Perhaps this is one of those cases where the penalty in efficiency doesn't compensate the penalty in readability?

solver/src/driver.rs
```diff
@@ -97,16 +134,15 @@ impl Driver {
...
                 Some((solver, settlement))
             })
             .collect();
 
         // Sort by key in descending order
-        settlements.sort_by_key(|(_, settlement)| {
-            Reverse(settlement.objective_value(/* estimated_prices*/))
-        });
+        settlements
+            .sort_by_key(|(_, settlement)| Reverse(settlement.objective_value(&estimated_prices)));
```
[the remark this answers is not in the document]
marcovc: Done in fd1b65. Perhaps an even better solution would be to store the objective value in the settlement object upon construction, since otherwise we still compute it twice .

solver/src/settlement.rs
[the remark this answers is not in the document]
marcovc: Actually I'm not sure about what happens for div 0. Will check the source!

solver/src/settlement.rs
marcovc: Thanks! It's here a2a4e.

marcovc:
Fixed the main function that scores the settlement - we can't simply recompute using external price vectors, but instead we need to use internal prices for computing surplus, and only in the end rescale using external prices. An extreme example where this makes a difference is when the external price vector fails to satisfy limit constraints.

Added a unit test.

solver/src/driver.rs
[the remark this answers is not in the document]
vkgnosis: I don't have a strong preference, just slightly prefer the non allocating version. So if you prefer this readability then it's fine if we keep it. Maybe compiler optimizes this out anyway.

solver/src/settlement.rs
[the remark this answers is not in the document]
marcovc: Found it: https://docs.rs/num-rational/0.1.42/num_rational/struct.Ratio.html#method.checked_div . I guess this means we need to keep the Option, right?

marcovc:
@twalth3r It would be great if you could also check this PR. Thanks!

solver/src/driver.rs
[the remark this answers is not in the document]
marcovc: What do you think about either implementing a small function or using something like this: https://danielkeep.github.io/rust-grabbag/doc/grabbag_macros/macro.collect!.html ?

solver/src/settlement.rs
```diff
@@ -204,99 +251,249 @@ mod tests {
         assert!(settlement.encode_trades().is_some());
     }
 
+    // Helper function to save some repeatition below.
+    fn r(u: u128) -> BigRational {
```
bh2smith, line 255: Nit: I guess this has been reduced to just r because calling it rat would also lead the reader to have to check what it is anyway and rationalize is far too long. If not, I might suggest calling it rational or at the very least rat. I know it's mathematically sound and much more concise as is, but I'm just hoping to prevent the reader from actually having to jump to the definition.

solver/src/driver.rs
```diff
@@ -9,15 +9,19 @@ use anyhow::{Context, Result};
 use contracts::GPv2Settlement;
 use futures::future::join_all;
 use gas_estimation::GasPriceEstimating;
+use itertools::Itertools;
+use maplit::hashmap;
```
bh2smith, line 13: What the difference between this and the standard library hash map?

solver/src/driver.rs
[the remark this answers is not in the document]
bh2smith: Is this collect more robust that the standard collect?

solver/src/settlement.rs
```diff
@@ -111,26 +117,53 @@ impl Settlement {
...
+    // Objective is re-computed using external prices.
+    fn objective_value_recomputed(
+        &self,
+        external_prices: &HashMap<H160, BigRational>,
+    ) -> Option<BigRational> {
+        self.total_surplus(external_prices)
+    }
+
+    // For now this computes the total surplus of all EOA trades.
+    pub fn objective_value(&self, external_prices: &HashMap<H160, BigRational>) -> BigRational {
```
bh2smith, line 161: I see that this method is now just encapsulating its recomputed counterpart with an additional overflow handling, do we really need this or could it all fit into a single place?

bh2smith (approved):
Really nice work here! Made a few comments and questions that could be treated as suggestions but no strong opinions either way.

twalth3r (approved):
Looks all good to me, thanks!

solver/src/settlement.rs
marcovc: I think calling it "rat" could tranform

https://github.com/gnosis/gp-v2-services/blob/79b12404652c7467002858ff079aeb0b5d8924dd/solver/src/settlement.rs#L406-L447

into

tenor

:)

(now seriously, @vkgnosis suggested the name, but I think one letter is already too many letters - there should be an implicit conversion in this case)

solver/src/driver.rs
marcovc: I'm guessing that the HashMap is the struct, and this maplit::hashmap is just a macro to create literal hashmaps?

solver/src/settlement.rs
marcovc: Yes, fixed in 779d70 . Thanks!

solver/src/driver.rs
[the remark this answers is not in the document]
vkgnosis: How does that macro help here?

solver/src/settlement.rs
vkgnosis: > hoping to prevent the reader from actually having to jump to the definition

The function is one line long and just above the tests so my personal preference is as short as possible because it makes the asserts more readable.

solver/src/driver.rs
vkgnosis: That's correct and in these cases I would suggest just checking the docs. From the use we see that the crate is called `maplit` so searching for that on crates.io or docs.rs shows the docs where it is explained.

solver/src/settlement.rs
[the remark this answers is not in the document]
vkgnosis: I still see a bunch of checked_... that are not checked_div so we could start by removing those and seeing how many options we can remove. For the divs it depends on whether the divisor can really be 0 which I am not sure about from the top of my head.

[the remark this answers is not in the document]
marcovc: Done here:  119b062

solver/src/settlement.rs
marcovc: Done here: 119b062

vkgnosis (approved):

solver/src/driver.rs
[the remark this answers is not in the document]
marcovc: How about this? b83fc75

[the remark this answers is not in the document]
vkgnosis: I see. I'm fine with although it will hopefully no longer be needed in the next rust version.

solver/src/driver.rs
```diff
@@ -63,6 +71,36 @@ impl Driver {
         }
     }
 
+    // Should this go here or on price_estimate.rs?
```
fleupold, line 74: no, I think this is very specific to the driver, and should stay here.

solver/src/driver.rs
```diff
@@ -80,11 +118,14 @@ impl Driver {
         tracing::debug!("got {} AMMs", amms.len());
 
         let liquidity: Vec<Liquidity> = limit_orders
+            .clone()
```
fleupold, line 121: nanonit: if you move `self.collect_estimated_prices(&limit_orders).await?;` above you shouldn't need to clone.

solver/src/util.rs
```diff
@@ -0,0 +1,22 @@
+// This macro allows to build literal iterations without allocating
+// a container such as a vector:
+// chain![1, 2, 3]
```
fleupold, line 3: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2F4Z5lEBKx7WMfEja9In%2Fgiphy.gif&token=chapeau" data-gitmeme-token="chapeau"></a>

solver/src/settlement.rs
```diff
@@ -36,21 +38,25 @@ impl Trade {
     }
 
     // The difference between the minimum you were willing to buy/maximum you were willing to sell, and what you ended up buying/selling
-    pub fn surplus(&self, sell_token_price: U256, buy_token_price: U256) -> Option<U256> {
+    pub fn surplus(
+        &self,
+        sell_token_price: &BigRational,
+        buy_token_price: &BigRational,
+    ) -> Option<BigRational> {
```
fleupold, line 45: nit: I wonder if this should return a Result instead of an Option so that we can pass the information about which trade is causing problems which would make it easier to debug. Can happen in a separate PR though as it's already an Option in the current codebase.

fleupold (approved):
Really good stuff!

solver/src/settlement.rs
marcovc: Will do in subsequent PR.

PR #374, opened 2021-03-12
Computes price estimates in the driver.

marcovc:
I split the PR in two: this one is just to compute the price estimates every batch and send it to the solver.

Apologies in advance :/

Questions:

Should UniswapPriceEstimator be factored out of orderbook (like put in shared/) ?


orderbook/src/price_estimate.rs
```diff
@@ -138,6 +140,29 @@ impl UniswapPriceEstimator {
...
+        Returns a vector of (rational) prices for the given tokens
+        denominated in denominator_token.
+    */
+    pub async fn best_execution_spot_prices(
+        &self,
+        tokens: &Vec<H160>,
+        denominator_token: H160,
+    ) -> Result<Vec<BigRational>> {
+        let res = join_all(tokens.iter().map(|token| {
+            self.best_execution(
```
fleupold, line 153: Would you not be able to reuse `self.best_execution_spot_price` here?

solver/Cargo.toml
```diff
@@ -36,6 +36,7 @@ tokio = { version = "0.2", features =["macros", "rt-threaded", "time"] }
 tracing = "0.1"
 web3 = { version = "0.15", default-features = false, features = ["http-tls"] }
 itertools = "0.10"
+orderbook= { path = "../orderbook" }
```
fleupold, line 39: solver and orderbook should not depend on another. Instead all shared components should live in the shared crate. Maybe in a separate PR, could you move `orderbook/src/price_estimate.rs` to `shared/src/price_estimate.rs `?

fleupold (commented):
You are on the right path. I missed that the price estimator is not yet available in the shared crate which needs to happen before we can use it in the driver.

orderbook/src/price_estimate.rs
marcovc: It seems best_execution_spot_price is returning a float. Not sure what the best option is actually, I was thinking that for comparing solutions we could get away with big rationals and big ints without going into floats. Wdyt?

solver/Cargo.toml
marcovc: Yes, that was my feeling too, I'll do that.

marcovc:
It seems "clippy" is complaining that the Driver constructor has now too many arguments:

```rust
    pub fn new(
        settlement_contract: GPv2Settlement,
        uniswap_liquidity: UniswapLiquidity,
        orderbook: OrderBookApi,
        price_estimator: UniswapPriceEstimator,
        solver: Vec<Box<dyn Solver>>,
        gas_price_estimator: Box<dyn GasPriceEstimating>,
        target_confirm_time: Duration,
        settle_interval: Duration,
    ) -> Self {
```

What to do?

vkgnosis:
One way to fix it is to silence the warning by putting `#[allow(clippy::too_many_arguments)]` right on top of the function. This is what we can do if we feel the warning is too limiting in this case. An alternative could be to group all the arguiments into a new struct like `DriverParameters` and pass the struct as the only argument. Not sure if that really makes it cleaner. For this PR I am fine with the #allow solution.

marcovc:
> An alternative could be to group all the arguiments into a new struct like DriverParameters and pass the struct as the only argument.

Thanks. We wouldn't have the same problem in the constructor of the DriverParameters struct?

vkgnosis:
That struct wouldn't need a constructor function because it just stores plain old data. So you would create it using the `Struct { ... }`
syntax instead of `Struct::new(...)`.

e2e/tests/uniswap_trade_test.rs
```diff
@@ -129,16 +129,16 @@ async fn test_with_ganache() {
...
             factory: uniswap_factory.clone(),
             web3: web3.clone(),
             chain_id,
         }),
         HashSet::new(),
-    );
+    ));
     let fee_calculator = Arc::new(MinFeeCalculator::new(
-        Box::new(price_estimator),
+        price_estimator.clone(),
```
marcovc, line 141: Does this increases the ref count? I've also seen `Arc::clone(...)`, is it the same?.

solver/Cargo.toml
marcovc: done in another PR, now merged into main and this one.

e2e/tests/uniswap_trade_test.rs
vkgnosis: Yes it's the same thing. Either way we are calling `Clone::clone` (the `clone` function of the `Clone` trait). It is just using different syntax. `.clone()` is more commonly used.

solver/src/driver.rs
```diff
@@ -79,6 +85,24 @@ impl Driver {
             .chain(amms.into_iter().map(Liquidity::Amm))
             .collect();
 
+        /*
```
fleupold, line 88: If you want to merge this PR (I can see auto-merge is already enabled) can you please remove this code?

e2e/tests/uniswap_trade_test.rs
fleupold: Actually this is the e2e test, so here it doesn't really matter if we use the same PriceEstimator or a different one (in reality the API and driver run in different processes). Can't hurt though to have these components sharable. I'm sure there will be more use cases soon.

shared/src/price_estimate.rs
```diff
@@ -137,6 +138,29 @@ impl UniswapPriceEstimator {
...
 
+    /*
+        Returns a vector of (rational) prices for the given tokens
+        denominated in denominator_token.
+    */
+    pub async fn best_execution_spot_prices(
+        &self,
+        tokens: &[H160],
+        denominator_token: H160,
+    ) -> Result<Vec<BigRational>> {
```
fleupold, line 149: I'd think floats here is fine since we only use it for one multiplication (it also be slightly more consistent since the functions almost sound the same). I used BigRational to avoid compounding rounding errors when multiplying reserve ratios. Otherwise we could also make the other method return Rationals and do the float conversion in the callsites.

fleupold (commented):
Looks good. Mainly not accepting because of the auto-enabled merge and dummy code still present.

shared/src/price_estimate.rs
marcovc: > Otherwise we could also make the other method return Rationals and do the float conversion in the callsites.

I'd prefer this way since there is no information loss in case we want to use rationals (as we will in the second part of this PR [here](https://github.com/gnosis/gp-v2-services/pull/373)).

Should I do it in this PR?

marcovc: Nevermind, it actually doesn't matter - there's no info loss. I'll do it here.

marcovc: oh wait, there is because the float is only 64 bits. Hmm, I think I prefer to do everything in rationals and convert to float in the call sites then.

marcovc: Done in 237ed55.

shared/src/price_estimate.rs
```diff
@@ -128,13 +132,25 @@ impl UniswapPriceEstimator {
...
-                big_rational_to_float(price)
-                    .ok_or_else(|| anyhow!("Cannot convert price ratio to float"))?,
-            ))
-        })
+    }
+
+    /*
+        Returns a vector of (rational) prices for the given tokens
+        denominated in denominator_token.
+    */
```
fleupold, line 140: nanonit: can we keep the same format for docstrings as in the other public methods? That is just start each line with //

shared/src/price_estimate.rs
```diff
@@ -128,13 +132,25 @@ impl UniswapPriceEstimator {
...
+        tokens: &[H160],
+        denominator_token: H160,
+    ) -> Result<Vec<BigRational>> {
+        let res = join_all(
+            tokens
+                .iter()
+                .map(|token| self.best_execution_spot_price(*token, denominator_token)),
+        )
+        .await;
+        let res: Result<Vec<(Vec<H160>, BigRational)>, anyhow::Error> = res.into_iter().collect();
```
fleupold, line 152: nit: maybe document that this method will return an error as soon as a single estimate fails (it's somewhat logical looking at the signature, but can't hurt to emphasise).

fleupold (approved):
Looks good to me! 

You could add a unit test for the newly added function but I think it's really optional as the code only reusing existing logic and not really adding much new (it could be a nice exercise for completeness though).

marcovc:
Good that I've tried to add a unit test. If I do:

```rust
    async fn best_execution_spot_prices() {
        let token_a = H160::from_low_u64_be(1);
        let token_b = H160::from_low_u64_be(2);
        let token_c = H160::from_low_u64_be(3);
        let pool_ab = Pool::uniswap(
            TokenPair::new(token_a, token_b).unwrap(),
            (10u128.pow(30), 10u128.pow(29)),
        );
        let pool_bc = Pool::uniswap(
            TokenPair::new(token_b, token_c).unwrap(),
            (10u128.pow(30), 10u128.pow(29)),
        );

        let pool_fetcher = Box::new(FakePoolFetcher(vec![pool_ab, pool_bc]));
        let estimator = UniswapPriceEstimator::new(pool_fetcher, hashset!());

        let res = estimator.best_execution_spot_prices(&[token_a, token_b, token_c], token_c).await;
        assert!(res.is_ok());
    }
```

I get a "panic" :

```
thread 'price_estimate::tests::best_execution_spot_prices' panicked at 'token list contains same token in a row', shared/src/uniswap_solver.rs:115:50
```

I guess I am assuming something that is not true.

marcovc:
Right, it appears we can't do `best_execution_spot_price(tokenA, tokenA)`

vkgnosis (approved):

shared/src/price_estimate.rs
```diff
@@ -166,13 +169,27 @@ impl UniswapPriceEstimator {
...
+        &self,
+        tokens: &[H160],
+        denominator_token: H160,
+    ) -> Result<Vec<BigRational>> {
+        let res = join_all(tokens.iter().map(|token| async move {
+            if *token != denominator_token {
+                self.best_execution_spot_price(*token, denominator_token)
+                    .await
+            } else {
+                Ok((vec![*token, *token], BigRational::from_integer(1.into())))
```
fleupold, line 187: Ubernit: Instead of returning a tuple of Vec<H160>, BigRational here, you could only return the BigRational (and in the if branch use `.map(|t| t.1)`.

I find it slightly awkward to temporarily create an invalid path.

fleupold (approved):
Nice find on the issue when the denominator token is part of the list.

shared/src/price_estimate.rs
marcovc: Thanks, I didn't realize that introducing the sync block allowed for a much simpler implementation.

PR #376, opened 2021-03-12
Cached Pool Fetcher

fleupold:
Now that we are using the uniswap pools in a lot of places (price estimate, fee estimate, base line solver, objective value computation), it might make sense to cache pool information as least until a new block arrives.

This PR introduces a CachedPoolFetcher which does exactly that. It uses the current_block_stream to invalidate its cache and otherwise performs a read_through logic with its inner fetcher.

I had to change current_block_stream to take a concrete Transport instead of a DynTransport. If there is value in having DynTransport we should probably change it everywhere in the codebase where we are passing Web3.

### Test Plan
Unit tests for the new component.

### DIFF
--- a/shared/src/uniswap_pool.rs
+++ b/shared/src/uniswap_pool.rs
@@ -1,12 +1,17 @@
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 
 use contracts::{UniswapV2Factory, UniswapV2Pair};
 use ethcontract::{batch::CallBatch, Http, Web3, H160, U256};
+use futures::{select, stream::FusedStream, Stream, StreamExt as _};
 use num::{rational::Ratio, BigInt, BigRational, Zero};
+use std::sync::Arc;
+use tokio::sync::Mutex;
 use web3::signing::keccak256;
 
 use hex_literal::hex;
 use model::TokenPair;
+
+use crate::current_block::Block;
 
 const UNISWAP_PAIR_INIT_CODE: [u8; 32] =
     hex!("96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f");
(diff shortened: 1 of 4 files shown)


shared/src/uniswap_pool.rs
```diff
@@ -138,6 +143,64 @@ impl Pool {
...
+    async fn fetch(&self, token_pairs: HashSet<TokenPair>) -> Vec<Pool> {
+        let fetch = || self.fetch_inner(token_pairs);
+        let clear_cache = || self.clear_cache();
+        let mut stream = self.block_stream.lock().await;
+        loop {
+            select! {
+                _ = stream.next() => clear_cache().await,
+                default => return fetch().await,
+            };
+        }
```
vkgnosis, line 200: If I understand this right this is supposed to clear the cache if stream is immediately ready and otherwise return `fetch()`. It might be possible to express this more obviously with `FutureExt::now_or_never` instead of `select!`.

shared/src/uniswap_pool.rs
```diff
@@ -138,6 +143,64 @@ impl Pool {
     }
 }
 
+// Read though Pool Fetcher that keeps previously fetched pools in a cache, which get invalidated whenever there is a new block
+pub struct CachedPoolFetcher<S> {
```
vkgnosis, line 147: I am worried about the current implementation is that it relies on the stream.next future to not be pending if there is a new block but I could imagine an implementation having to for example acquire a mutex which could block even if there is a new block.

Solution 1: Spawn a task that clears the cache whenever the stream yields a new block.

Solution 2: Adapt CurrentBlockStream so that it not only provides a stream that can be reacted to, but also allows querying the current block. Code like this only has to check whether the block has *changed* instead of having to react to every individual update. This way we can compare the most recent block hash from CurrentBlockStream to the most recent block hash we have updated the cache with. If the hashes differ we have to clear the cache.

I like solution 2 because I expect there will be other code that has the same use case in the future. Will make a PR as a proof of concept.

vkgnosis (approved):

PR #377, opened 2021-03-12
Add log splitter binary for kubernetes alerts

vkgnosis:
After some discussion with devops we want to make this part of the
docker image instead of inserting it through configmap later.
Reimplemented in rust because we're a rust code base.

Once this is merged the plan is to change the kubernetes command to `sh -c "orderbook | log-splitter"`.

### Test Plan
new unit test

### DIFF
--- a/docker/Dockerfile.binary
+++ b/docker/Dockerfile.binary
@@ -5,11 +5,18 @@ WORKDIR /usr/src/oba-services
 COPY . .
 RUN cargo build --target x86_64-unknown-linux-musl --release
 
+RUN \
+    cd .. && \
+    git clone https://github.com/gnosis/regex-stream-split.git && \
+    cd regex-stream-split && \
+    cargo build --target x86_64-unknown-linux-musl --release
+
 # Extract Binary
 FROM alpine:latest
 
 # Handle signal handlers properly
 RUN apk add --no-cache tini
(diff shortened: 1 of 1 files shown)


fleupold (approved):
Looks good, although I can how this is easily useful for any rust project that we run inside gnosis (e.g. even the IDO services, Safe gateway, etc).

Can we maybe publish this as a GPv2 independent binary crate? Otherwise I suspect people will just-copy paste it 😞

vkgnosis:
That is a good point. Will make a new repo for it.

vkgnosis:
Requesting review again because I changed this to pull from the other repository.

docker/Dockerfile.binary
```diff
@@ -5,11 +5,18 @@ WORKDIR /usr/src/oba-services
 COPY . .
 RUN cargo build --target x86_64-unknown-linux-musl --release
 
+RUN \
+    cd .. && \
+    git clone https://github.com/gnosis/regex-stream-split.git && \
```
fleupold, line 10: Should we consider pinning to a commit?

fleupold (approved):
Nice!

docker/Dockerfile.binary
vkgnosis: added

PR #378, opened 2021-03-12
Make block stream FusedStream

vkgnosis:
This avoids other code having to call `.fuse()` on it.

### Test Plan
new unit test

### DIFF
--- a/shared/src/current_block.rs
+++ b/shared/src/current_block.rs
@@ -1,6 +1,6 @@
 use anyhow::{anyhow, Context as _, Result};
 use ethcontract::transport::DynTransport;
-use futures::Stream;
+use futures::{stream::FusedStream, Stream};
 use primitive_types::H256;
 use std::{
     future::Future,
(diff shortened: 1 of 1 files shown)


shared/src/current_block.rs
```diff
@@ -62,22 +60,35 @@ pub fn current_block_stream(
...
     tokio::task::spawn(update_future);
-    CurrentBlockStream { receiver }
+    CurrentBlockStream::new(receiver)
 }
 
 #[derive(Clone)]
-struct CurrentBlockStream {
+pub struct CurrentBlockStream {
     receiver: watch::Receiver<Option<Block>>,
+    is_terminated: bool,
```
vkgnosis, line 69: Should we make this AtomicBool so that the struct becomes Sync?

fleupold: yes.

fleupold (approved):

shared/src/current_block.rs
vkgnosis: turns out it is already Sync + Send because (in hindsight obviously) is_terminated only being mutated through &mut self makes it so.

PR #379, opened 2021-03-12
Make CurrentBlockStream able to return most recently seen block

vkgnosis:
This is useful for users of the struct that do not need to react to the
stream in realtime but only poll whether the block has changed some
times.

As part of this the initialization function has become async so that it can fetch the initial block immediately. This also plays nicely with the channel because we no longer need a `None` default value.

### Test Plan
adjusted test


fleupold (approved):

fleupold:
Is this ready to be merged, so I can depend on it in #376?

PR #384, opened 2021-03-15
Move order signing logic into a trait for reuse

bh2smith:
Because we plan to sign `OrderCancellation`, it will be important to refactor the order signing logic in a way that generalizes the common methods into a trait (`EIP712Signing`). This PR performs the refactor for `OrderCreation`so that less work needs to be done when introducing Order Cancellation.

I am not entirely sure it is appropriate to include this new unit test since it requires an ethcontract import into models. Perhaps now that the bug has been uncovered, we should really test `OrderBuilder.sign_with`

### DIFF
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -94,7 +94,6 @@ impl Signature {
 
     pub fn validate(&self, domain_separator: &DomainSeparator, digest: &[u8; 32]) -> Option<H160> {
         let v = self.v & 0x1f;
-        // Does it make sense to include just order_uid here?
         let message = self.signing_digest(domain_separator, digest);
         let recovery = Recovery::new(message, v as u64, self.r, self.s);
         let (signature, recovery_id) = recovery.as_signature()?;
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -12,12 +12,34 @@ use primitive_types::{H160, H256};
 use serde::{de, Deserialize, Serialize};
 use std::fmt;
 use web3::signing;
+use web3::signing::{Key, SecretKeyRef};
+use web3::types::Recovery;
 
 #[derive(Eq, PartialEq, Clone, Copy, Debug, Default, Hash)]
 pub struct Signature {
     pub r: H256,
     pub s: H256,
     pub v: u8,
+}
+
+pub trait EIP712Signing {
+    fn digest(&self) -> [u8; 32];
+    fn signature(&self) -> Signature;
+
+    fn sign_self_with(&self, domain_separator: &DomainSeparator, key: &SecretKeyRef) -> Signature {
+        let message = Signature::signing_digest_message(domain_separator, &self.digest());
+        // Unwrap because the only error is for invalid messages which we don't create.
+        let signature = Key::sign(key, &message, None).unwrap();
+        Signature {
+            v: signature.v as u8 | 0x80,
+            r: signature.r,
+            s: signature.s,
+        }
+    }
+
+    fn validate_signature(&self, domain_separator: &DomainSeparator) -> Option<H160> {
+        self.signature().validate(domain_separator, &self.digest())
+    }
 }
 
 impl Signature {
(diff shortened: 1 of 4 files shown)


model/src/order.rs
```diff
@@ -539,4 +491,29 @@ mod tests {
         assert_eq!(uid.to_string(), expected);
         assert_eq!(format!("{}", uid), expected);
     }
+
+    #[test]
+    fn order_signature_recovery() {
+        const PRIVATE_KEY: [u8; 32] =
+            hex!("0000000000000000000000000000000000000000000000000000000000000001");
+        let public_key = Account::Offline(PrivateKey::from_raw(PRIVATE_KEY).unwrap(), None);
```
vkgnosis, line 499: I think this can be `H160(PRIVATE_KEY)` so there is no need to go through the ethcontract types.

model/src/order.rs
```diff
@@ -113,7 +110,8 @@ impl OrderBuilder {
     pub fn sign_with(mut self, domain_separator: &DomainSeparator, key: SecretKeyRef) -> Self {
         self.0.order_meta_data.owner = key.address();
         self.0.order_meta_data.uid = self.0.order_creation.uid(&key.address());
-        self.0.order_creation.sign_self_with(domain_separator, &key);
+        self.0.order_creation.signature =
+            self.0.order_creation.sign_self_with(domain_separator, &key);
```
bh2smith, line 114: This was the Bug from the previous PR. I hadn't changed this line which was causing the `OrderBuilder.sign_with` to be incorrect

model/src/order.rs
vkgnosis: nvm, I see that this has to calculate the public key from the private key. We could hardcode the pubic key or see if we can calulate it with secpk.

bh2smith: Not so sure about this. Also, think we use the fact that it is an `Account`as well here.
Screenshot 2021-03-15 at 13 37 35

bh2smith: I could maybe use something like this though:

```
let context = &SECP256K1;
        let s: key::SecretKey = key::SecretKey::from_slice(context, &secret[..])?;
        let pub_key = key::PublicKey::from_secret_key(context, &s)?;
        let serialized = pub_key.serialize_vec(context, false);
        ```

File: model/src/lib.rs [hunk 3]
```diff
@@ -20,6 +22,26 @@ pub struct Signature {
...
+        // Unwrap because the only error is for invalid messages which we don't create.
+        let signature = Key::sign(key, &message, None).unwrap();
+        Signature {
+            v: signature.v as u8 | 0x80,
+            r: signature.r,
+            s: signature.s,
+        }
+    }
+
+    fn validate_signature(&self, domain_separator: &DomainSeparator) -> Option<H160> {
```
vkgnosis, line 40: Add comment here explaining that this returns the public key if the signature is well formed, to make it completely obvious.

model/src/order.rs
vkgnosis: I think that makes sense and it's not too messy.

vkgnosis (approved):

model/src/order.rs
bh2smith: Just need to compare `PublicKey` with `H160` now.

model/src/lib.rs
```diff
@@ -37,6 +60,46 @@ impl Signature {
...
+        if self.v & 0x80 == 0 {
+            Signature::signing_digest_typed_data(domain_separator, digest)
+        } else {
+            Signature::signing_digest_message(domain_separator, digest)
+        }
+    }
+
+    pub fn validate(&self, domain_separator: &DomainSeparator, digest: &[u8; 32]) -> Option<H160> {
+        let v = self.v & 0x1f;
+        // Does it make sense to include just order_uid here?
```
bh2smith, line 97:
```suggestion
```

model/src/order.rs
```diff
@@ -539,4 +490,35 @@ mod tests {
...
+
+    pub fn h160_from_public_key(key: PublicKey) -> H160 {
+        let hash = keccak256(
+            &key.serialize_uncompressed()[1..], /* cut '04' */
+        );
+        H160::from_slice(&hash[12..])
+    }
+
+    #[test]
+    fn order_signature_recovery() {
```
bh2smith, line 502:
```suggestion
    fn order_builder_signature_recovery() {
```

PR #385, opened 2021-03-15
Cancellation API

### LINKED ISSUE #320: Soft (offchain) Order Cancellation

bh2smith:
Closes #320

Implement request and response handlers for order cancellation.

### Test Plan
New tests for cancellation API.

### DIFF
--- a/orderbook/src/api/cancel_order.rs
+++ b/orderbook/src/api/cancel_order.rs
@@ -1,29 +1,20 @@
+use crate::api::extract_payload;
 use crate::orderbook::{OrderCancellationResult, Orderbook};
 use anyhow::Result;
 use model::order::OrderCancellation;
 use std::{convert::Infallible, sync::Arc};
 use warp::{hyper::StatusCode, Filter, Rejection, Reply};
 
-const MAX_JSON_BODY_PAYLOAD: u64 = 1024 * 16;
-
-fn extract_cancellation() -> impl Filter<Extract = (OrderCancellation,), Error = Rejection> + Clone
-{
-    // (rejecting huge payloads)...
-    warp::body::content_length_limit(MAX_JSON_BODY_PAYLOAD).and(warp::body::json())
-}
-
 pub fn cancel_order_request(
 ) -> impl Filter<Extract = (OrderCancellation,), Error = Rejection> + Clone {
     warp::path!("orders")
         .and(warp::delete())
-        .and(extract_cancellation())
+        .and(extract_payload())
 }
 
 pub fn cancel_order_response(result: Result<OrderCancellationResult>) -> impl Reply {
     let (body, status_code) = match result {
-        Ok(OrderCancellationResult::Cancelled) => {
-            (warp::reply::json(&"Cancelled"), StatusCode::OK)
-        }
+        Ok(OrderCancellationResult::Cancelled) => (warp::reply::json(&"Cancelled"), StatusCode::OK),
         Ok(OrderCancellationResult::InvalidSignature) => (
             super::error("InvalidSignature", "Likely malformed signature"),
             StatusCode::BAD_REQUEST,
@@ -75,5 +66,30 @@ mod tests {
             .json(&cancellation);
         let result = request.filter(&filter).await.unwrap();
         assert_eq!(result, cancellation);
+    }
+
+    #[tokio::test]
+    async fn cancel_order_response_ok() {
+        let response =
+            cancel_order_response(Ok(OrderCancellationResult::Cancelled)).into_response();
+        assert_eq!(response.status(), StatusCode::OK);
+    }
+
+    #[tokio::test]
+    async fn cancel_order_response_err() {
+        let response =
+            cancel_order_response(Ok(OrderCancellationResult::InvalidSignature)).into_response();
+        assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+
+        let response =
+            cancel_order_response(Ok(OrderCancellationResult::WrongOwner)).into_response();
+        assert_eq!(response.status(), StatusCode::UNAUTHORIZED);
+
+        let response =
+            cancel_order_response(Ok(OrderCancellationResult::OrderNotFound)).into_response();
+        assert_eq!(response.status(), StatusCode::NOT_FOUND);
+
+        let response = cancel_order_response(Err(anyhow::Error::msg("test error"))).into_response();
+        assert_eq!(response.status(), StatusCode::INTERNAL_SERVER_ERROR);
     }
 }
--- a/orderbook/src/api/cancel_order.rs
+++ b/orderbook/src/api/cancel_order.rs
@@ -22,7 +22,7 @@ pub fn cancel_order_request(
 pub fn cancel_order_response(result: Result<OrderCancellationResult>) -> impl Reply {
     let (body, status_code) = match result {
         Ok(OrderCancellationResult::Cancelled) => {
-            (warp::reply::json(&"Cancelled"), StatusCode::ACCEPTED)
+            (warp::reply::json(&"Cancelled"), StatusCode::OK)
         }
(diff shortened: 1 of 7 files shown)


orderbook/src/api/cancel_order.rs
```diff
@@ -1 +1,75 @@
...
\ No newline at end of file
+use crate::orderbook::{Orderbook, OrderCancellationResult};
+use anyhow::Result;
+use model::order::OrderCancellation;
+use std::{convert::Infallible, sync::Arc};
+use warp::{hyper::StatusCode, Filter, Rejection, Reply};
+
+const MAX_JSON_BODY_PAYLOAD: u64 = 1024 * 16;
+
+fn extract_cancellation() -> impl Filter<Extract = (OrderCancellation,), Error = Rejection> + Clone {
```
fleupold, line 9: This can probably be shared with `extract_user_order` and become a more abstract `fn extract_payload<T: DeserializeOwned + Send>() -> impl Filter<Extract = (T,), Error = Rejection> + Clone { ... }
`

orderbook/src/api/cancel_order.rs
```diff
@@ -1 +1,75 @@
...
+pub fn cancel_order_response(result: Result<OrderCancellationResult>) -> impl Reply {
+    let (body, status_code) = match result {
+        Ok(OrderCancellationResult::Cancelled) => (warp::reply::json(&"Cancelled"), StatusCode::ACCEPTED),
+        Ok(OrderCancellationResult::InvalidSignature) => (
+            super::error("InvalidSignature", "Likely malformed signature"),
+            StatusCode::BAD_REQUEST,
+        ),
+        Ok(OrderCancellationResult::OrderNotFound) => (
+            super::error("OrderNotFound", "order not located in database"),
+            StatusCode::BAD_REQUEST,
```
fleupold, line 30: This should be a Not Found

orderbook/src/api/cancel_order.rs
```diff
@@ -1 +1,75 @@
...
+        ),
+        Ok(OrderCancellationResult::WrongOwner) => (
+            super::error("WrongOwner", "Signature recovery's owner doesn't match order's"),
+            StatusCode::BAD_REQUEST,
```
fleupold, line 34: This should be Unauthorized

orderbook/src/api/cancel_order.rs
```diff
@@ -1 +1,75 @@
...
+pub fn cancel_order_request() -> impl Filter<Extract = (OrderCancellation,), Error = Rejection> + Clone
+{
+    warp::path!("orders")
+        .and(warp::delete())
+        .and(extract_cancellation())
+}
+
+pub fn cancel_order_response(result: Result<OrderCancellationResult>) -> impl Reply {
+    let (body, status_code) = match result {
+        Ok(OrderCancellationResult::Cancelled) => (warp::reply::json(&"Cancelled"), StatusCode::ACCEPTED),
```
fleupold, line 23: According to https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

> The request has been accepted for processing, but the processing has not been completed. The request might or might not be eventually acted upon, and may be disallowed when processing occurs.

which I'm not sure is exactly true in this case.

I don't care too much about the actual code we use, but given our OpenAPI spec says we return 200 in this case  let's maybe do as documented?

orderbook/src/database/orders.rs
```diff
@@ -305,8 +311,14 @@ mod tests {
             .await
             .unwrap();
         assert_eq!(db_orders[0].order_meta_data.invalidated, true);
-        // TODO - cancel twice and verify that first cancellation date isn't over written
-        // This will require querying the DB for the cancellation_date.
+
+        let query = "SELECT cancellation_timestamp FROM orders;";
```
fleupold, line 315: nit: a slightly better way to test this is instead of using `Utc::now()` in the database implementation, we inject our own abstract `now` function that is is using `Utc::now()` in production, but that we can mock in our unit tests.

An example can be found here: https://github.com/gnosis/oba-services/blob/b6de9c3b3193d6e23b217a2442428429a54e690c/orderbook/src/fee.rs#L20

and a test that manipulates the time: https://github.com/gnosis/oba-services/blob/b6de9c3b3193d6e23b217a2442428429a54e690c/orderbook/src/fee.rs#L292

Otherwise this test might not be catching issues on a very fast machine where both inserts happen in the same millisecond.

fleupold (changes requested):
Looks good, but let's please make the return codes match the openapi spec of this endpoint.

bh2smith:
I hadn't tagged any reviewers yet, but certainly will do.

orderbook/src/database/orders.rs
bh2smith: Very good point and thanks for the pointers here. My assumption then would be to include this now field on the `OrderCancellation` struct so that it can be used in the query? Or is the a more appropriate place for this?

bh2smith: I am beginning to think this change might be a more involved refactoring of how we use time on the entire database. If its alright with you I might suggest leaving this change for a separate PR.

vkgnosis: Maybe a lot simpler than using a trait would be to just pass the current time `now` into `cancel_order` as a parameter instead of creating it there. Should only be a  2 line change.

bh2smith: Actually, I think I may have found something that works by implementing an `async now` method on Database that can be used as `Database::now().await`. Not sure if this is what you had in mind, but please let me know your thoughts.

vkgnosis (approved):

orderbook/src/database/orders.rs
bh2smith: Thanks @vkgnosis - I will go with that.

vkgnosis: I meant changing

```
 pub async fn cancel_order(&self, order_uid: &OrderUid) -> Result<()>
```

to 

```
 pub async fn cancel_order(&self, order_uid: &OrderUid, now: DateTime<Utc>) -> Result<()>
```

That way who ever calls the function has control over the time. In the test you can then manually increment the time while in the orderbook you would set the parameter to `Utc::now()`.

bh2smith: Awesome. This looks much better.

orderbook/src/database/orders.rs
```diff
@@ -83,7 +83,7 @@ impl Database {
             })
     }
 
-    pub async fn cancel_order(&self, order_uid: &OrderUid) -> Result<()> {
+    pub async fn cancel_order(&self, order_uid: &OrderUid, now: DateTime<Utc>) -> Result<()> {
```
fleupold, line 86: nice: this is much simpler than my suggestion 👍

orderbook/src/database/orders.rs
```diff
@@ -298,15 +303,39 @@ mod tests {
...
-        // This will require querying the DB for the cancellation_date.
+
+        let query = "SELECT cancellation_timestamp FROM orders;";
+        let first_cancellation: CancellationQueryRow =
+            sqlx::query_as(query).fetch_one(&db.pool).await.unwrap();
+        assert_eq!(cancellation_time, first_cancellation.cancellation_timestamp);
+
+        // Cancel again and verify that cancellation timestamp was not changed.
+        let irrelevant_time =
+            DateTime::<Utc>::from_utc(NaiveDateTime::from_timestamp(1234567890, 1), Utc);
```
fleupold, line 326: nit: Cutting it tight with 1ns difference (I'm not sure all postgres configs support this granularity). Maybe you could use a time that's a bit more different?

fleupold (approved):

orderbook/src/database/orders.rs
vkgnosis: That is a good point. I'd say increment the seconds by 1 to be sure.

Issue #382, opened 2021-03-12
Allow for some slippage in the Uniswap Baseline Solver


fleupold:
https://etherscan.io/tx/0xa32b4b2f2041151d81017f8db7e83882e28460aa861a5e75cd2347923143cfdc failed because of `INSUFFICIENT_OUTPUT_AMOUNT`.

This is likely related to us setting 0 slippage tolerance inside the Uniswap baseline solver (so any tiny alteration in the pool reserves against or favor will lead to reversion). From a user perspective, the order will likely get settled in the next run (as new up to date prices are computed), however our solver can burn easily a lot of gas in the process (about 80$ in the current case)

One challenge given the current contracts is that we have to decide on a clearing price at the time of solution authoring (not when it is mined). The user will get tokens according to that price. At the moment, we keep positive slippage (between solution computation and mining) inside the protocol and fail in case of negative slippage. 

We should probably allow for a fixed small amount of negative slippage in our solution (funds would be take from fees that we collected prior). Maybe that amount could be related to the gas cost we would incur if the solution submission fails?

PR #387, opened 2021-03-15
Add Slippage to Uniswap interaction

fleupold:
Quick fix for #382 

This PR adds a fixed 0.1% slippage to our Uniswap interaction.

### Test Plan
Unit tests + doing a trade on xDAI and observing that the min amount was smaller than the received amount by a 0.999 factor:

Screenshot 2021-03-15 at 18 14 47
Screenshot 2021-03-15 at 18 15 03

### DIFF
--- a/solver/src/liquidity/uniswap.rs
+++ b/solver/src/liquidity/uniswap.rs
@@ -218,6 +218,7 @@ mod tests {
 
     #[test]
     fn test_out_amount_with_slippage() {
+        assert_eq!(out_amount_with_slippage(0.into()), 0.into());
         assert_eq!(out_amount_with_slippage(100.into()), 99.into());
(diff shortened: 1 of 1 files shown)


bh2smith (approved):
Looks good. Just wondering Does this affect the user's configured slippage from the interface?

vkgnosis (approved):
Code looks fine but maybe deserves a comment somewhere about why we do this.

fleupold:
> Does this affect the user's configured slippage from the interface?

It does not. I briefly looked into what it would take to handle the slippage in each solver directly (this way we could e.g. use the user's actual limit price) but it's a bit more involved as we would have to compute the slippage for each hop.

@nlordell is also working on a uniswap fast-path wrapper that allows computing the prices on chain and giving the slippage to the user in those cases.

PR #388, opened 2021-03-15
[API] Define Price Estimation Endpoint

fleupold:
This PR suggests a new price estimation endpoint that can be consumed by clients instead of doing their own estimation (as e.g. currently in the frontend).

I'm planning to use this endpoint when setting up the testing bots that will place random orders every once in a while. A similar endpoint was also already requested by Andrey and Dima when they built their iOS prototype and Leandro was planning to use this endpoint to estimate the value of surplus and other token units in the explorer.

The implementation will follow...

### Test Plan
OpenAPI format is valid.

### DIFF
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -240,6 +240,41 @@ paths:
                 $ref: "#/components/schemas/FeeInformation"
         404:
           description: Token non-existent or not connected to native token
+  /api/v1/price/{baseToken}/{quoteToken}:
+    get:
+      description: |
+        The estimated price for selling an optionally specified amount of baseToken or buying that amount of quote token. Price is denominated in quote token.
+        E.g. the price for WETH/USDC (0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48) might be around 1800 USDC.
+      parameters:
+        - name: baseToken
+          in: path
+          required: true
+          schema:
+            $ref: "#/components/schemas/Address"
+        - name: quoteToken
+          in: path
+          required: true
+          schema:
+            $ref: "#/components/schemas/Address"
+        - name: amount
+          in: query
+          schema:
+            $ref: "#/components/schemas/TokenAmount"
+            default: "0"
+        - name: kind
+          in: query
+          schema:
+            $ref: "#/components/schemas/OrderType"
+            default: sell
+      responses:
+        200:
+          description: the price denominated in quote token
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/PriceEstimate"
+        404:
+          description: Token non-existent or no valid price found
 components:
   schemas:
     Address:
@@ -371,6 +406,18 @@ components:
         signature:
           description: "OrderCancellation signed by owner"
           $ref: "#/components/schemas/Signature"
+    PriceEstimate:
+      description: |
+        Provides the information about an estimated price.
+      type: object
+      properties:
+        price:
+          description: The estimated price
+          type: number
+          format: double
+        currency:
+          description: "The currency in which the price is given"
+          $ref: "#/components/schemas/Address"
     Trade:
       description: |
         Trade data such as executed amounts, fees, order id and block number.
@@ -431,11 +478,6 @@ components:
       properties:
         errorType:
           type: string
-          enum:
-            [
-                InvalidSignature,
-                WrongOwner,
-                OrderNotFound,
-            ]
+          enum: [InvalidSignature, WrongOwner, OrderNotFound]
         description:
           type: string
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -240,11 +240,10 @@ paths:
                 $ref: "#/components/schemas/FeeInformation"
         404:
           description: Token non-existent or not connected to native token
-  /api/v1/markets/{baseToken}-{quoteToken}/price:
+  /api/v1/markets/{baseToken}-{quoteToken}/{kind}/{amount}:
     get:
       description: |
-        The estimated price for selling an optionally specified amount of baseToken or buying that amount of quote token. Price is denominated in quote token.
-        E.g. the price for WETH-USDC (0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2-0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48) might be around 1800 USDC.
+        The estimated amount received for selling a specified amount of baseToken or the estimate amount require to sell for buying that amount of quote token.
       parameters:
         - name: baseToken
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):

orderbook/openapi.yml
```diff
@@ -240,6 +240,41 @@ paths:
                 $ref: "#/components/schemas/FeeInformation"
         404:
           description: Token non-existent or not connected to native token
+  /api/v1/price/{baseToken}/{quoteToken}:
```
anxolin, line 243: what about `/markets/{baseToken}-{quoteToken}/price`

i.e `/markets/WETH-DAI/price`

anxolin: For me, price is one property of the market. Also, makes it easier to implement different prices. 
I believe we will want to implement different prices in the future:
- spot prices
- market price
- mid market price
- some median of the last xxx period
- bid price
- ask price
- ... no idea what prices are relevant

Implementing these endpoints as properties of the market feels more natural. 
Also normally is best not to use singular form ("price"). It would be a "prices"

orderbook/openapi.yml
```diff
@@ -240,6 +240,41 @@ paths:
...
+    get:
+      description: |
+        The estimated price for selling an optionally specified amount of baseToken or buying that amount of quote token. Price is denominated in quote token.
```
anxolin, line 246: Ok, so it's optional just to get the mid market price, or some market price calculation. If this is the case, I would consider to make it mandatory and be explicit on what is exactly this price. 

From the definition, I would think is a "spot price", since we give the amount and the direction of the order.

orderbook/openapi.yml
```diff
@@ -240,6 +240,41 @@ paths:
...
+        - name: amount
+          in: query
+          schema:
+            $ref: "#/components/schemas/TokenAmount"
+            default: "0"
+        - name: kind
+          in: query
+          schema:
+            $ref: "#/components/schemas/OrderType"
+            default: sell
```
anxolin, line 268: I wouldn't specify a default direction of the order

orderbook/openapi.yml
```diff
@@ -240,6 +240,41 @@ paths:
...
+        - name: quoteToken
+          in: path
+          required: true
+          schema:
+            $ref: "#/components/schemas/Address"
+        - name: amount
+          in: query
+          schema:
+            $ref: "#/components/schemas/TokenAmount"
+            default: "0"
```
anxolin, line 263: What would this mean, to sell 0 tokens?

orderbook/openapi.yml
```diff
@@ -371,6 +406,18 @@ components:
...
           $ref: "#/components/schemas/Signature"
+    PriceEstimate:
+      description: |
+        Provides the information about an estimated price.
+      type: object
+      properties:
+        price:
+          description: The estimated price
+          type: number
+          format: double
```
anxolin, line 417: Can this be a string in decimal format?
Like '105.1234567433' for example

anxolin (commented):
Great you are working on this, and the bots too

orderbook/openapi.yml
fleupold: What's the advantage? Wouldn't a float be more easily processable?

orderbook/openapi.yml
fleupold: It would mean to get the spot price. I specified defaults so that the endpoint without any arguments can be used to estimate the spot price. But we can make everything mandatory and expose a separate spot price route.

orderbook/openapi.yml
vkgnosis: I think this only matters for precision because many json libraries force "numbers" to be f32 or f64. f32 could be a problem if there are tokens with very different number of decimals. So to be safe we might want this to be a string which also allows us to use BigRational in the backend if we want to.

anxolin: Yes, my issue would be that double and floats have some issues with precision. They don't represent the value we actually mean for example:

image

anxolin: In JS people can use BigNumber libraries, in Python Decimal

Btw, I think https://docs.python.org/3/library/decimal.html library explains great why this model is better. People will also be able to parse these strings as doubles, but the other way around cannot be done without loosing information

orderbook/openapi.yml
anxolin: For me this behavior would be ok, and we can leave it. Maybe It would be good to mention in the description if we want to keep it like this.

I have some misconception about prices, maybe it would help if we have a glossary that we agree on with the different relevant prices. Let me explain. A price can be for example "the best ask", "best buy", the "mid point of those", or it can be "amount dependant", or even it can be defined as the "last buy", or the "last sell". 
So, it's hard to talk about prices as an absolute number. 

My understanding is, all those prices will have "different names", and I'm not the best at choosing those, but I think doing so wisely is important. So my point is. should we ask for this amount to be mandatory and positive, and use different "endpoint name" for the price that has 0 as the amount, and this endpoint?

I think it would help to start giving names to these prices, and decide the ones we want to do now.

orderbook/openapi.yml
fleupold: Makes sense 👍

orderbook/openapi.yml
fleupold: If we force amount > 0 and not have a default order kind, this should maybe not return a price but an amount instead? This way we can do things like `/api/v1/markets/{baseToken}-{quoteToken}/sell/1000`

anxolin: that's a convenient endpoint yep, to know how much you are getting. Although, I would still add one for prices. Even if they are different ways of seeing the same info

But maybe what's best, is to check how other APIs return the price. And choose one we think is good :)

anxolin: Also, other alternative is:

`/api/v1/markets/{baseToken}-{quoteToken}/sell-price?amount=1000`

returns:
```js
{
   price: '125.2345678',
   sellAmount: '1000',
   buyAmount: '125234.5678'
}
```
I think this would be quite easy to read

fleupold: The thing I don't like about the price + amount together endpoint is that it would require two units. Sell/buy token and price currency. Turning a buy and sell amount into a price is something every client should be able to do locally, and we can expose best bid/best ask/price endpoints later on.

fleupold:
Changed the endpoint to `/api/v1/markets/{baseToken}-{quoteToken}/{kind}/{amount}`  e.g. 

`/api/v1/markets/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2-0x6b175474e89094c44da98b954eedeac495271d0f/sell/1000000000000000000` 

would return something like:

```
{
  "token": 0x6b175474e89094c44da98b954eedeac495271d0f
  "amount": 1800000000000000000000
}
```

PR #389, opened 2021-03-16
Specify full docker registry url for images

vkgnosis:
This is good form to make it clear where the images come from (in our
case the default docker registry). I encountered this when trying to use
the docker alternative podman which forces users to be explicit about this.

### Test Plan
image still builds


fleupold (approved):
Can you also update Dockerfile.migration?

PR #390, opened 2021-03-16
Put startup script that uses correct log regex into image

vkgnosis:
Since we have the binary itself in the container it makes to also have
the regex so that we can update it directly if we ever change the log
message format.

### Test Plan
Ran the image manually to make sure it still works


bh2smith (approved):
Cute!

fleupold (approved):

PR #391, opened 2021-03-16
Transaction Hash on Trade API

### LINKED ISSUE #334: Transaction Hash in Trade Event
For many reasons, it will be important to include transaction hash in the trades table. Also, in order to get a unique `batchId` we would like to store the `log_index` of the settlement event as well.

bh2smith:
Closes #334 

In order to return transaction hash along with trades via the api service, we introduce a new `settlements` table in the database which includes columns `block_number | log_index | tx_hash | solver`. This table is linked to the trades query via block number and log index as follows:

- `trade.block_number` = `settlement.block_number`
- `settlement.log_index` = smallest log index of all settlements in the same block for which 
        `trade.log_index < settlement.log_index`

The trade struct now includes an optional field `transaction_hash` which (when available) is returned along with a trade query. Note that all trades occurring after this deployment will always have a transaction hash, while all trades occurring before will have none. This means that the Optional aspect of this field could eventually be phased out, if we were to successfully populate with the missing settlement data.

Furthermore, we introduce contract event handler for Settlements which is responsible for settlement insertion in response to event emission and deletion in the case of reorgs.

### Test Plan
Relevant new tests introduced here demonstrate that
- trades without `tx_hash`  and
- trades linked to multiple settlements in the same block
can be queried as expected


database/sql/V007__extend_trades_with_txhash.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE trades
+    ADD COLUMN tx_hash bytea NOT NULL,
+    ADD COLUMN settlement_index bigint NOT NULL;
```
fleupold, line 3: If these are NOT NULL you might have to think about what to set them to for the records that are already in the table.

Also given that these two columns together will act as a pseudo foreign key into the batches table, I was wondering if it would make sense to replace them with a single column batch_id (but not sure how this might affect the simplicity of this PR in itself).

orderbook/openapi.yml
```diff
@@ -244,8 +244,11 @@ paths:
           description: Token non-existent or not connected to native token
 components:
   schemas:
+    TransactionHash:
+      description: 32 byte digest encoded as a hex with `0x` prefix.
+      example: "0xd51f28edffcaaa76be4a22f6375ad289272c037f3cc072345676e88d92ced8b5"
     Address:
-      description: Ethereum 40 byte address encoded as a hex with `0x` prefix.
+      description: 20 byte Ethereum address encoded as a hex with `0x` prefix.
```
fleupold, line 251: 👍

orderbook/src/database/events.rs
```diff
@@ -150,7 +151,10 @@ async fn insert_trade(
                 .bind(event.order_uid.0.as_ref())
                 .bind(u256_to_big_decimal(&event.sell_amount_including_fee))
                 .bind(u256_to_big_decimal(&event.buy_amount))
-                .bind(u256_to_big_decimal(&event.fee_amount)),
+                .bind(u256_to_big_decimal(&event.fee_amount))
+                .bind(&event.tx_hash.as_bytes())
+                // TODO - get the settlement log index from transaction...
```
fleupold, line 156: This will be the tricky part 🙈

database/sql/V007__extend_trades_with_txhash.sql
bh2smith: I was under the impression from past experience with Postgres that if you make a column required at a later point in time the database won't complain that those pre existing records could remain as they were. Lucky for us though the settlement index for every batch so far is probably zero anyway (although I would have to double check) so we could simpleton perform a write to update all the null values to zero. I would also have to double check this other assumption about pre existing record that is probably not true (especially for a required field that is involved as a foreign key)

Hopefully the settlement index is zero for all previous settlement events.

bh2smith:
So it turns out that the contracts we are currently using do not emit a settlement event. See for example [this](https://etherscan.io/tx/0xd51f28edffcaaa76be4a22f6375ad289272c037f3cc072345676e88d92ced8b5#eventlog) transaction.

This got me thinking that we would surely have to start the DB from scratch when we start using the new contracts. This may be an opportunity to consolidate the schema definition before alpha. @fleupold not sure when exactly this is planned to happen, but I would create an issue to reorg the `database/sql/*` files when the time comes.

As for the current setup, this also means we have the opportunity to move the settlement event to the beginning of the transaction logs before mainnet deployment (as it is currently the last)

https://github.com/gnosis/gp-v2-contracts/blob/e5af5269d27d5e976d9750e5108878e10f11afa9/src/contracts/GPv2Settlement.sol#L114-L134

In theory, it shouldn't really matter if its first last or in the middle, but it would make detection of double settlement a bit easier and also help us to know exactly where in the logs to be looking. Of course, we can always parse the transaction logs by `topic` as we should be doing anyway so this isn't really a major concern. However, this would make the migration a bit easier if we can simply update all the settlement indices with 0 right from the start.

orderbook/src/database/events.rs
bh2smith: So tricky in fact, that it is currently not possible since the older contracts do not emit a settlement event.

bh2smith: Given that this PR is functional as is and can't be pushed beyond until we are using the most recent contracts, I might propose that we merge it (along with an `UPDATE trades SET settlement_index = 0` and maybe some dummy txHash) as is so that we can at least have txHash for the alpha release.

orderbook/openapi.yml
```diff
@@ -244,8 +244,11 @@ paths:
           description: Token non-existent or not connected to native token
 components:
   schemas:
+    TransactionHash:
+      description: 32 byte digest encoded as a hex with `0x` prefix.
+      example: "0xd51f28edffcaaa76be4a22f6375ad289272c037f3cc072345676e88d92ced8b5"
```
bh2smith, line 249: PS - this is the transaction hash of our biggest COW yet. We may want to update this to the new contracts at some point.

https://etherscan.io/tx/0xd51f28edffcaaa76be4a22f6375ad289272c037f3cc072345676e88d92ced8b5

vkgnosis:
For every event (be it trade or settlement) we know the log index of the event in the block and we already store this for trades. For settlement events we would also store block number, log index, transaction hash.
Then to associate a trade with a settlement we make a query that looks at all settlements in the same block and then takes the settlement with the lowest log index that is > the trade's log index (assuming that settlement events come after trades, if they came before it would the other way around). In postgres we will have an index for settlements that makes this an efficient query.
This way we do not need any foreign keys or store the transaction hash with trades.

fleupold:
This is an interesting idea (I like the idea of not having a forced foreign key as it keeps the tables less coupled). 

Ben and I discussed the PR in our 1:1 (before Valentin's suggestion) and we concluded that it would be nice to not make this task dependent on the smart contract update and instead "fake" the settlement events for now (they only carry the `msg.sender` which we already get on the transaction). This should also work in the suggested proposal, we would just have to use a large log index (e.g. the last trade event +1), but it's a bit more hacky.

One minor advantage of a foreign key relation is that the key can be somewhat opaque and doesn't have to correspond to correct "real world" data (e.g. the foreign key could work with a null log index instead of a fake value if we don't have it yet).

vkgnosis:
> the foreign key could work with a null log index instead of a fake value if we don't have it yet

With my idea this would still work. We would just not be able to return find the settlement corresponding to the log index. So a hypothetical new field `settlement transaction hash` in the Trade struct would be Option::None. I like that this correctly models the case where we do not know about the settlement yet although we do have the trade.

fleupold:
> I like that this correctly models the case where we do not know about the settlement yet although we do have the trade.

I think it might be nice to process trade events for a settlement in a single db transaction to have a consistent state of the database (and e.g. not require the frontend to request information again if the tx hash wasn't found temporarily).

vkgnosis:
But you would just refresh in explorer's trade page if that happens, right (which should be rare, anyway)? I suspect trying to enforce that would lead to awkward code in the event manager because the way we process is sequential and there is no real guarantee to the order in which we get events from the node.

bh2smith:
Thanks @fleupold  and @vkgnosis for your valuable input here. In order to pull all the thoughts so far together I have summarized a few options for the layout:

For both scenarios, it looks like we could go with the following table structure:

- Settlements contains `txHash | block_number | log_index | solver`

- Trades is extended by `settlement_index`, but not `txHash`


### Option 1: No enforced Foreign Key

Trades are linked to unique settlements via limit supremum (smallest greater than) of `log_index` for settlements occurring in the same block.

Migration of existing records is not necessary, but could be spoofed by dropping the trades table replaying all trade events, creating settlement records from trade events and falsely recording the settlement index to the max value. 

_This should work now assuming there have not yet been more than one settlement in a single block so far!*_


### Option 2: Use the Foreign key 

Given the table structure as above and the fact that we no longer have txHash in the trades table, I can only imagine that `(block_number, settlement_index)` would be an appropriate foreign key, but I think that this can easily fail in the case where two settlements occur in the same block and the settlement event log indices are the same.


Option 3: Keep the redundant `txHash` in the trades table and use `(txHash, settlementIndex)` as the Foreign Key. We could make the `settlementIndex` Optional (i.e. allowed to be Null) and include a field on the Trades struct which would be `Option<TxHash>` but this option would only be relevant for old records


To summarize:

There are a lot of possible scenarios at play in the options above, some of them can be mixed and matched through each option.

- I am not terribly fond of having an `Option<txHash>` on the trade struct because it is only relevant for pre-release records.
- That being said, I am keen to try and remove the redundant `txHash` from the trades table (if possible).
- Not exactly sure what the general consensus is here with regards to the options above. They all have their pros and cons. I would think that `Option 1` is the most sound and checks most of the boxes.
- There is also the discussion happening here about how the events should be processed and I do, very much, like the idea that events can all be processed independently (which would again be Option 1)

vkgnosis:
Iirc we talked about this on Monday but you couldn't make it to the meeting. So far we've agreed on option 1 in which case the trades table doesn't need to be extended at all. To find the settlement when querying an order we extend the query to find the supremum in the settlement table.

bh2smith:
> trades table doesn't need to be extended at all

Right, yes the trades table wouldn't need `settlement_index`, but I am beginning to believe we may need `transaction_index` in case there were two settlements in the same block.

vkgnosis:
I don't think any extension is needed. The log index orders all events in a block. We know that a settlement event always comes directly after the trades. So the rule to find the settlement of a trade is to find the smallest settlement log index that is still larger than the trade's index. That should work even with multiple settlements in one block, or multiple settlements in one transaction.

bh2smith:
Sorry, I wasn't aware that the log index did this. That's good enough for me!

fleupold:
We could potentially extend the trades table with the transaction hash (leading to some redundancy), but given that the new smart contracts are basically ready it might make more sense to just base this work stream on top of that PR.

orderbook/src/database/events.rs
```diff
@@ -153,6 +154,34 @@ async fn insert_trade(
                 .bind(u256_to_big_decimal(&event.fee_amount)),
         )
         .await?;
+    // TODO: Temporarily insert_settlement here with the limited information we have
+    insert_settlement(transaction, index).await?;
```
bh2smith, line 158: The insert settlement method could be slightly more general here if we spoof the index being passed in. We could also introduce the `SettlementEvent` Struct here and pass a spoofed/dummy value so the `insert_settlement` could already be written in its most general form.

orderbook/src/database/trades.rs
```diff
@@ -27,8 +27,15 @@ impl Database {
...
-                o.sell_token \
+                o.sell_token, \
+                s.tx_hash \
             FROM trades t \
+            LEFT OUTER JOIN settlements s \
+            ON t.block_number = s.block_number
+            AND t.log_index < (
+                SELECT MIN(log_index) FROM settlements s2
+                WHERE t.block_number = s2.block_number
+            )
```
vkgnosis, line 38: This turned out more complicated than I thought:

```suggestion
            LEFT OUTER JOIN LATERAL \
                ( \
                    SELECT tx_hash \
                    FROM settlements s \
                    WHERE \
                        s.block_number = t.block_number AND \
                        s.log_index > t.log_index \
                    ORDER BY s.log_index ASC \
                    LIMIT 1 \
                ) settlement \
                ON true \
```

see https://www.postgresql.org/docs/13/queries-table-expressions.html ctr+f for `lateral`.

vkgnosis (commented):
Could you mark this as merging into the contracts-upgrade branch so that the diff is cleaner?

orderbook/src/database/events.rs
```diff
@@ -1,22 +1,24 @@
...
 use model::order::OrderUid;
 use sqlx::{Connection, Executor, Postgres, Transaction};
 use std::convert::TryInto;
 
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy)]
 pub struct EventIndex {
     pub block_number: u64,
     pub log_index: u64,
+    pub transaction_hash: H256,
```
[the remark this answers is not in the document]
bh2smith: Oh yes! Very cool point. Sorry I missed this. Silly me.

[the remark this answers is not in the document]
bh2smith: This was actually left over from the hacky solution needed before the contract upgrade.

orderbook/src/database/trades.rs
```diff
@@ -295,4 +343,121 @@ mod tests {
...
+        let event_index_a = EventIndex {
+            block_number: 0,
+            log_index: 0,
+        };
+        let trade_a = add_order_and_trade(
+            &db,
+            owners[0],
+            order_ids[0],
+            event_index_a,
+            Option::from(settlement_hash),
```
vkgnosis, line 379:
```suggestion
            Some(settlement_hash),
```
This happens a couple more times below.

vkgnosis (approved):

fleupold (approved):
Very nice! Just a small comment about the logic of the unit tests that I will find hard to understand in a month from now when revisiting this code.

One general worry (unrelated to this PR) I have with the current code factoring is that adding a new table requires updates in a bunch of scattered places, e.g.:

- events.rs `block_number_of_most_recent_event`
- database.rs `clear()`

It's easy to forget one of these places.

Maybe one thing we can consider in the future is to create a `trait DbTable` (and maybe another `trait EventTable`) that define some of the common functionality that all tables need to fulfill. There could be a single enumerable ALL_TABLES type which the places that rely on this behaviour could use to enumerate and call into all tables.

orderbook/src/database/trades.rs
```diff
@@ -27,8 +27,16 @@ impl Database {
                 t.sell_amount - t.fee_amount as sell_amount_before_fees,\
                 o.owner, \
                 o.buy_token, \
-                o.sell_token \
+                o.sell_token, \
+                settlement.tx_hash \
             FROM trades t \
+            LEFT OUTER JOIN LATERAL ( \
```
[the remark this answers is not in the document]
bh2smith: Credit goes to @vkgnosis here who somehow pulled this trick out of his hat!

orderbook/openapi.yml
```diff
@@ -464,6 +467,9 @@ components:
         buyAmount:
           description: "Total amount of buyToken received in this trade."
           $ref: "#/components/schemas/TokenAmount"
+        transactionHash:
```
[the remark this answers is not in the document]
bh2smith: Yes, I was trying to look into our existing API definitions for some sort of optional type but was unsuccessful. This looks like what we need here.

orderbook/src/database/trades.rs
```diff
@@ -115,19 +127,23 @@ mod tests {
         return (owners, order_ids);
     }
 
-    async fn add_trade(db: &Database, owner: H160, order_uid: OrderUid, log_index: u64) -> Trade {
+    async fn add_trade(
+        db: &Database,
+        owner: H160,
+        order_uid: OrderUid,
+        event_index: EventIndex,
+        tx_hash: Option<H256>,
```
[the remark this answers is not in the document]
bh2smith: Yes, this is all a bit counter intuitive at this point. Although there are now so many combinations to go along with it. We would have `add_trade`, `add_order_and_trade`, `add_order_trade_and_settlement`, `add_trade_and_settlement`, possibly more.  I will try and come up with something a bit cleaner here.

orderbook/openapi.yml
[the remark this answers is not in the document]
bh2smith: I looked at the docs a little further to see how to make null-able properties of objects and it appears all properties are optional by default so we will have to make all the other fields required instead:

> By default, all object properties are optional. You can specify the required properties in the required list:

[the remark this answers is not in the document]
fleupold: Nice find. Yes, I think we should specify what objects are guaranteed to be set and which ones aren't (can happen in a separate PR though).

orderbook/src/database/trades.rs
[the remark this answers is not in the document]
bh2smith: I have made the unit tests more readable/less confusing by adding `add_settlement`. Anything other than this (i.e. combined helpers like add_trade_and_settlement`) didn’t make sense since they are very much independent tasks.

bh2smith:
This is ready to go, but is waiting on the merge of its predecessor #413

orderbook/openapi.yml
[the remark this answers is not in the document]
bh2smith: Created an issue for this and will pick it up ASAP.

https://github.com/gnosis/gp-v2-services/issues/437

bh2smith:
@fleupold - I also created an issue (https://github.com/gnosis/gp-v2-services/issues/438) for your concern regarding common table traits.

PR #396, opened 2021-03-18
Do not error whole whole settlement if one price estimate fails

### LINKED ISSUE #395: Don't error full settlement if price of some tokens can't be determined
> 2021-03-18T11:48:39.695Z ERROR solver::driver: single run errored: failed estimating pricesCaused by:    no valid path found

Just noticed this in the alerts. We should continue with tokens for which we are able to get a price.

vkgnosis:
Fixes #395 

### Test Plan
Don't think a regression test is needed here as the function signature has changed to no longer return result.

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -11,7 +11,6 @@ use contracts::GPv2Settlement;
 use futures::future::join_all;
 use gas_estimation::GasPriceEstimating;
 use itertools::Itertools;
-use maplit::hashmap;
 use num::BigRational;
 use primitive_types::H160;
 use shared::price_estimate::PriceEstimating;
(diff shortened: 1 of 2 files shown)


solver/src/driver.rs
```diff
@@ -94,30 +90,44 @@ impl Driver {
...
         tracing::debug!("starting single run");
-        let limit_orders = self
+        let mut limit_orders = self
             .orderbook
             .get_liquidity()
             .await
             .context("failed to get orderbook")?;
         tracing::debug!("got {} orders", limit_orders.len());
 
+        let estimated_prices = self.collect_estimated_prices(&limit_orders).await;
```
fleupold, line 117: nanonit: Maybe log number of orders after the pruning?

fleupold (approved):
Thanks a lot for fixing!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FlVBtp4SRW6rvDHf1b6%2Fgiphy-downsized-medium.gif&token=fix" data-gitmeme-token="fix"></a>

solver/src/driver.rs
vkgnosis: added

PR #398, opened 2021-03-18
Creates MIP solver.

marcovc:
* Creates MIP solver enum.
* Exposts the solver API url as a parameter.

### DIFF
--- a/solver/src/http_solver.rs
+++ b/solver/src/http_solver.rs
@@ -18,7 +18,6 @@ use std::{
 
 // TODO: exclude partially fillable orders
 // TODO: set settlement.fee_factor
-// TODO: find correct ordering for uniswap trades
 // TODO: gather real token decimals and store them in a cache
 // TODO: special rounding for the prices we get from the solver?
 
@@ -31,6 +30,13 @@ pub struct SolverConfig {
 }
 
 impl SolverConfig {
+    pub fn new(max_nr_exec_orders: u32, time_limit: u32) -> Self {
+        SolverConfig {
+            max_nr_exec_orders,
+            time_limit,
+        }
+    }
+
     fn add_to_query(&self, url: &mut Url) {
         url.query_pairs_mut()
             .append_pair(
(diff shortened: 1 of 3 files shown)


solver/src/http_solver.rs
```diff
@@ -31,6 +30,13 @@ pub struct SolverConfig {
 }
 
 impl SolverConfig {
+    pub fn new(max_nr_exec_orders: u32, time_limit: u32) -> Self {
+        SolverConfig {
+            max_nr_exec_orders,
+            time_limit,
+        }
+    }
```
vkgnosis, line 38: Nit: I'd keep the struct without constructor because it is just plain old data. No need to go through a function to create it.

vkgnosis (approved):

solver/src/http_solver.rs
marcovc: This was just to avoid making the struct fields public (would be required since it is being created from a different place). Would you still prefer the no-ctor version?

vkgnosis: My rule of thumb is that if it is plain old data with no further invariants then no constructor and public fields instead.

solver/src/solver.rs
```diff
@@ -18,15 +22,27 @@ arg_enum! {
...
+) -> Vec<Box<dyn Solver>> {
     solvers
         .into_iter()
         .map(|solver_type| match solver_type {
             SolverType::Naive => Box::new(NaiveSolver {}) as Box<dyn Solver>,
             SolverType::UniswapBaseline => Box::new(UniswapSolver::new(base_tokens.clone())),
+            SolverType::MIP => Box::new(HttpSolver::new(
+                mip_solver_url.clone(),
+                None,
+                SolverConfig::new(100, 100),
```
fleupold, line 43: Is this time-limit in seconds? I think we should not give the solver more than 30s at the moment. Otherwise this can massively delay the other settlements as well given they wait on all to complete.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FHw8vYF4DNRCKY%2Fgiphy.gif&token=ship_it" data-gitmeme-token="ship_it"></a>

bh2smith:
There are a lot of todos left over in this merged branch... do they all correspond to some open issues? If not, I would propose creating issues for each of the todos and maybe linking them directly in the commented code.

PR #407, opened 2021-03-22
Replaces IERC20 by ERC20.

marcovc:
Because IERC20 doesn't have decimals().

NOTE: I have no idea about if what I did is correct:

1. Created a ERC20.json file in contracts with [this](https://ethereumdev.io/abi-for-erc20-contract-on-ethereum/).
2. Removed the IERC20.json file that was there.
3. Did a project wide find-and-replace, to replace IERC20 to ERC20.
4. Removed ".view()" call to please the compiler.


fleupold (approved):
Looks good to me. I'm a bit surprised that GPv2Settlement.json was also updated (that shouldn't be necessary). Maybe you can revert it so we don't run the risk of some subtle bug

marcovc:
That got updated in the find-and-replace step. Is it ok if it still references the IERC20?

fleupold:
Yes, I think we should use the original ABI as it's in the npm package (the smart contract might only depend on the IERC20 as it's not using decimals()).

bh2smith (approved):

PR #410, opened 2021-03-23
Fix panic in naive solver

vkgnosis:
We saw in the alerts that compute_uniswap_in paniced. Looking at the function we see that it can overflow. 
This fix does the multiplication in BigInt and propagates potential errors through an Option.
There might be a smarter fix by removing the offending order and continuing but I wasn't sure how to best implement that. In the meantime this should fix the panic at least.

### Test Plan
New test that used to panic but now doesn't

### DIFF
--- a/solver/src/naive_solver/multi_order_solver.rs
+++ b/solver/src/naive_solver/multi_order_solver.rs
@@ -6,8 +6,9 @@ use anyhow::{anyhow, Result};
 use liquidity::{AmmOrder, LimitOrder};
 use model::order::OrderKind;
 use num::{bigint::Sign, BigInt};
+use primitive_types::U256;
 use std::collections::HashMap;
-use web3::types::{Address, U256};
+use web3::types::Address;
 
 #[derive(Debug)]
 struct TokenContext {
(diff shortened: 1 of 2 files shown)


nlordell (approved):

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -237,8 +236,11 @@ fn compute_uniswap_out(shortage: &TokenContext, excess: &TokenContext) -> U256 {
 /// of tokens to be sent to the pool.
 /// Taken from: https://github.com/Uniswap/uniswap-v2-periphery/blob/4123f93278b60bcf617130629c69d4016f9e7584/contracts/libraries/UniswapV2Library.sol#L53
 ///
-fn compute_uniswap_in(out: U256, shortage: &TokenContext, excess: &TokenContext) -> U256 {
-    U256::from(1000) * out * excess.reserve / (U256::from(997) * (shortage.reserve - out)) + 1
+fn compute_uniswap_in(out: U256, shortage: &TokenContext, excess: &TokenContext) -> Option<U256> {
+    let numerator = 1000 * u256_to_big_int(&out) * u256_to_big_int(&excess.reserve);
+    let denominator = 997 * u256_to_big_int(&(shortage.reserve - out));
+    let result: BigInt = numerator / denominator + 1;
+    bigint_to_u256(&result).ok()
```
nlordell, line 243: Why not just use the `checked_${operator}` methods directly here?

vkgnosis: My thinking was that using bigint would allow temporary overflow that is resolved by the division. Not sure which way is better.

nlordell: So the Uniswap contract uses `uint112` with checked operations, so overflowing the `BigInt` would probably result in a revert on-chain.

fleupold (approved):
Thanks for the quick fix, and nice regression test. In case it's relevant I believe this was the order causing problems:

> {"creationDate":"2021-03-23T10:23:01.847189Z","owner":"0x52df85e9de71aa1c210873bcf37ec46d36c99dc2","uid":"0x871041ad049d4ca6adf26e2c08f5c96f9a4e9629168c229f7beafad6f10c668752df85e9de71aa1c210873bcf37ec46d36c99dc26059c3da","availableBalance":"3556737749936925943202","executedBuyAmount":"0","executedSellAmount":"0","executedSellAmountBeforeFees":"0","executedFeeAmount":"0","invalidated":false,"sellToken":"0x5592ec0cfb4dbc12d3ab100b257153436a1f0fea","buyToken":"0x1f9840a85d5af5bf1d1762f925bdaddc4201f984","sellAmount":"1947707421601709851474","buyAmount":"2000000000000000000","validTo":1616495578,"appData":0,"feeAmount":"987537336368036480","kind":"buy","partiallyFillable":false,"signature":"0x97588a214e3ecec247230345510e2bad008518dc62ff4ea27fd340a309b02add4d2ac7d6a6556657870d2de74cdfb8f3e01e840a8b849dad4705cb7ee911665c1b"}

PR #412, opened 2021-03-23
Adds token information to mip solver requests.

marcovc:
Questions:

I've created a function to get the token info for one token at a time. Should I add the function to get multiple tokens in one batch? If so, can I remove the single one? Doesn't make much sense to me in having both.

### Test Plan
* Tested by running a local setup with everything running and creating some orders on the frontend. Couldn't find an easy way to add a test for it...


solver/src/http_solver.rs
```diff
@@ -91,11 +96,27 @@ impl HttpSolver {
...
-        tokens
+    async fn token_models(
+        &self,
+        tokens: &HashMap<String, H160>,
+    ) -> Result<HashMap<String, TokenInfoModel>> {
+        let addresses: Vec<H160> = tokens.values().into_iter().cloned().collect();
+        let token_infos = self
+            .token_info_fetcher
+            .get_token_infos(addresses.as_slice())
+            .await?;
```
marcovc, line 107: Wanted to ask this again: do we really want to replace missing token info by some default (e.g. 0 decimals) and try the solver anyway instead?

solver/src/http_solver.rs
```diff
@@ -91,11 +96,27 @@ impl HttpSolver {
...
+        Ok(tokens
             .iter()
-            .map(|(index, _)| (index.clone(), TokenInfoModel { decimals: 18 }))
-            .collect()
+            .map(|(index, address)| {
+                let token_info = token_infos.get(address).unwrap();
+                (
```
vkgnosis, line 112: Shouldn't unwrap (maybe related to discussion above).

solver/src/http_solver.rs
vkgnosis: I'm fine with either approach. If we skip orders with tokens without decimals then the solver can't handle them at all. Is there a disadvantage to using a default value? Does it make the solver worse at the other tokens if one of them has very wrong decimals?

vkgnosis (approved):

solver/src/http_solver.rs
marcovc: >  Does it make the solver worse at the other tokens if one of them has very wrong decimals?

That may very well happen yes. We are currently relying on number of decimals for avoiding numerical issues in the solver, so having a wrong number for one token can be enough to invalidate the full solution.

marcovc: I guess a more reliable solution would be to make number of decimals optional for the solver, but given that we're building a PoC and this is not supposed to be common, I'd say to just skip the MIP solver whenever there are orders involving tokens without decimals for now. @twalth3r  What is your opinion?

vkgnosis: I would filter out the orders for which we don't have decimals and continue with the rest.

twalth3r: I'd also be fine with both, hoping it wouldn't happen too often anyways.

If we're going for the default-approach, maybe using a default of 18 decimals is safer here!? Most tokens have 18 decimals, so chances are good we're on the right path. If the token actually has fewer decimals, then the order amounts should be accordingly small, and our solver would just treat them as negligible under its 18-decimal assumption. Does that make sense?

shared/src/token_info.rs
```diff
@@ -0,0 +1,81 @@
...
+}
+
+#[async_trait]
+impl TokenInfoFetching for CachedTokenInfoFetcher {
+    async fn get_token_info(&self, address: &H160) -> Result<TokenInfo> {
+        let mut cache = self.cache.lock().await;
+        if cache.contains_key(address) {
+            Ok(cache[address])
+        } else {
+            let token_info = self.inner.get_token_info(address).await?;
```
fleupold, line 76: Here is where we could potentially add an optimization that inspects the error and if it's non transient (e.g. method does not exist), it also caches the fact that this will always error. For now, maybe a TODO note would be nice.

solver/src/http_solver.rs
fleupold: I think I'd also slightly prefer filtering out non-compliant tokens. We can add support once we see a real use case and this way we don't risk unexpected solver behavior.

shared/src/token_info.rs
```diff
@@ -0,0 +1,81 @@
...
+#[automock]
+#[async_trait]
+pub trait TokenInfoFetching: Send + Sync {
+    /// Retrieves some token information from a token address.
+    async fn get_token_info(&self, address: &H160) -> Result<TokenInfo>;
+
+    /// Retrieves all token information.
+    /// Default implementation calls get_token_info for each token and ignores errors.
+    async fn get_token_infos(&self, addresses: &[H160]) -> Result<HashMap<H160, TokenInfo>> {
+        let mut result = HashMap::new();
```
fleupold, line 32: It would be more efficient to fetch all infos in one roundtrip to the node (by putting them in a single batch call like we do in the `PoolFetcher`), but given that we already cache responses I think it's not that important here.

On the other hand, since we don't really use the single get_token_info, we might as well only expose the plural one on the trait.

shared/src/token_info.rs
```diff
@@ -0,0 +1,81 @@
...
+use mockall::*;
+
+#[cfg_attr(test, derive(Eq, PartialEq))]
+#[derive(Copy, Clone, Debug)]
+pub struct TokenInfo {
+    pub decimals: u8,
+}
+
+pub struct TokenInfoFetcher {
+    //pub factory: UniswapV2Factory,
```
fleupold, line 18: nit: please remove before merging

shared/src/token_info.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub web3: Web3<Http>,
+    //pub chain_id: u64,
+}
+
+#[automock]
```
fleupold, line 23: nit: this can be made "test-only" by using:

```rust
#[cfg_attr(test, mockall::automock)]
```

(cf. e.g. `BalanceFetching`)

vkgnosis: Problem is that it is in a separate crate so when the `solver` is compiled in test, it's dependency `shared` will not be in test. Iirc there is an outstanding cargo issue/feature to allow using the same dependency also as dev-depencency (used by tests) with different features.

solver/src/http_solver.rs
marcovc: This was changed now: driver passes a list with all tokens and all orders to solver, and the solver does its best with it, which may fallback to ignoring the tokens without decimals.

shared/src/token_info.rs
marcovc: Done!

marcovc:
This is what it was done to address the reviews:

* A TokenInfo now has an Option<F> for each field F. This option is None when we couldn't fetch this particular bit of info. Ideally this would be a Result, which would allow for logic to either continuing to refetch the property or assume it is permanently unavailable. Unfortunately it seems Result::Error are not clonable, and therefore I can't cache them. As a workaround I thought about storing an Enum<F> which could be Ok(value), TempErr(err), or PermanentErr(err), but all this really seems like not super priority.

* Batched the requests to the smartcontract. I don't love the current implementation, since it is not so easy to generalize to the case when we want multiple calls for each token info, but it's a start.

shared/src/token_info.rs
```diff
@@ -0,0 +1,97 @@
...
+            })
+            .collect::<Vec<_>>();
+
+        batch.execute_all(MAX_BATCH_SIZE).await;
+
+        addresses
+            .iter()
+            .zip(join_all(futures).await.into_iter())
+            .map(|r| {
+                if (r.1.is_err()) {
```
fleupold, line 50: This is a bit risky. If there was a transient error (e.g. the node shutting down or a timeout), we will cache the token as not having any decimals until we restart. For the time being it might be better to not cache errors and create a TODO that we could inspect the error to decide if it's transient or not.

shared/src/token_info.rs
```diff
@@ -0,0 +1,97 @@
...
+        // Add token infos to cache.
+        cache.extend(fetched);
+
+        // Return token infos from the cache.
+        addresses
+            .iter()
+            .map(|address| (*address, cache[address]))
+            .collect()
+    }
+}
```
fleupold, line 97: nit: Maybe a unit test for the cached fetcher?

fleupold (approved):
I think if we cached errors we would need to check that the error is non-transient (but that the contract is actually not implementing the method e.g.). That goes beyond the scope of this PR, but to be safe I'd rather just refetch "bad tokens" for now.

Also a unit test or two for the cached fetcher would be cool (can be done in a different PR if you want to get this one in before).

shared/src/token_info.rs
marcovc: You're right - changed it.

shared/src/token_info.rs
marcovc: Done!

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FCjmvTCZf2U3p09Cn0h%2Fgiphy-downsized-medium.gif&token=ready" data-gitmeme-token="ready"></a>

PR #413, opened 2021-03-23
Upgrade contracts version

vkgnosis:
Edit: This PR has already been reviewed. I merged the ethsign Pr into this so that there is one up to date branch.

Before merging: Make ethcontract release

Follow ups:

* reintroduce something like the EIP712 trait
* handle OrderInvalidated events

### Test Plan
e2e test


model/src/lib.rs
```diff
@@ -11,40 +12,33 @@ use lazy_static::lazy_static;
...
+}
+
 #[derive(Eq, PartialEq, Clone, Copy, Debug, Default, Hash)]
 pub struct Signature {
     pub r: H256,
     pub s: H256,
     pub v: u8,
 }
 
-pub trait EIP712Signing {
```
vkgnosis, line 25: I am removing this trait because with the new signing schemes I found it hard to wrap my head around implementing them and keeping this trait consistent. In the produce of adding support for other signing schemes we will see how we can generalize again to go back to having a trait.

model/src/appdata_hexadecimal.rs
```diff
@@ -0,0 +1,51 @@
+use serde::{de, Deserializer, Serializer};
+use std::fmt;
+
+pub fn serialize<S>(value: &[u8; 32], serializer: S) -> Result<S::Ok, S::Error>
+where
```
vkgnosis, line 5: With the next rust version we will get const generics with which we might be able to implement a single serialize thing for all the places we want to encode bytes as hex (h160, signature, appdata).

Cargo.toml
```diff
@@ -15,3 +15,7 @@ default-members = [
     "shared",
     "solver",
 ]
+
+[patch.crates-io]
+ethcontract = { git = "https://github.com/gnosis/ethcontract-rs.git", branch = "abiv2_" }
+ethcontract-generate = { git = "https://github.com/gnosis/ethcontract-rs.git", branch = "abiv2_" }
```
vkgnosis, line 21: This will go away when I make an ethcontract release

model/src/appdata_hexadecimal.rs
```diff
@@ -0,0 +1,51 @@
...
+                    "failed to decode {:?} as hex appdata 32 bytes: {}",
+                    s, err
+                ))
+            })?;
+            Ok(value)
+        }
+    }
+
+    deserializer.deserialize_str(Visitor {})
+}
```
fleupold, line 51: Can we add some small unit test here?

model/src/lib.rs
```diff
@@ -61,44 +55,31 @@ impl Signature {
...
+        struct_hash: &[u8; 32],
+    ) -> Option<H160> {
+        let message = hashed_eip712_message(domain_separator, struct_hash);
+        let recovery = Recovery::new(message, self.v as u64, self.r, self.s);
+        let (signature, recovery_id) = recovery.as_signature()?;
+        signing::recover(&message, &signature, recovery_id).ok()
     }
 
-    pub fn signing_digest_message(
+    pub fn sign(
```
fleupold, line 69: Is there still a unit test that check we can sign and later validate the signers address?

orderbook/src/orderbook.rs
```diff
@@ -109,7 +109,10 @@ impl Orderbook {
             None => return Ok(OrderCancellationResult::OrderNotFound),
         };
 
-        match cancellation.validate_signature(&self.domain_separator) {
+        match cancellation
```
fleupold, line 112: nanonit: can we keep the `validate_signature` method on the cancellation struct? I find this nicer to interact with than having the orderbook know how cancellations are validated.

solver/src/encoding.rs
```diff
@@ -1,101 +1,61 @@
...
-        assert_eq!(encoded[74..106], u8_as_32_bytes_be(8));
-        assert_eq!(encoded[106], 0b11);
-        assert_eq!(encoded[107..139], u8_as_32_bytes_be(3));
-        assert_eq!(encoded[139..141], [0, 4]);
-        assert_eq!(encoded[141], 9);
-        assert_eq!(encoded[142..174], u8_as_32_bytes_be(10));
-        assert_eq!(encoded[174..206], u8_as_32_bytes_be(11));
-    }
-}
+pub type EncodedInteraction = (
```
fleupold, line 57: This is convenient.

fleupold (approved):
Looks good to me. One thing no note that before we add the message signing scheme back in this change is not going to work with the trading bot that I'm working on (as hardhat atm doesn't support typed signing for "real" networks).

I assume on the frontend we are only relying on typed signatures so starting with that scheme is fine?

database/sql/V007__contracts_upgrade.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TYPE SigningScheme AS ENUM ('eip712');
+
+DELETE FROM orders;
```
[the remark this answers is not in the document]
vkgnosis: Signatures for the old orders wouldn't be valid in the new scheme so I don't think there is a point. Of course we could migrate them but then we might be violating the assumption that orders in the database are valid.

model/src/appdata_hexadecimal.rs
vkgnosis: It is tested through the roundtrip for the full order struct so imo no need for an extra test.

model/src/lib.rs
vkgnosis: Yes, `order_builder_signature_recovery` in `order.rs`.

model/src/appdata_hexadecimal.rs
fleupold: I see. Isn't it nicer though to test the code closer to where it is defined (otherwise e2e tests covering e.g. serialization might suffice)? 
With a unit test here, we could e.g. document the behaviour for a hex string that is too long or contains invalid characters, which I assume an order roundtrip test doesn't care about.

orderbook/src/orderbook.rs
vkgnosis: fixed.

model/src/appdata_hexadecimal.rs
vkgnosis: True. We haven't done this for the h160 serialization code either which I copied here. Like I said above, with the next rust release we can hopefully have the hex serialization for byte slices only once which can then be tested in one place. Will add a new test because it should be useful then too.

orderbook/src/database/orders.rs
```diff
@@ -48,28 +48,54 @@ impl DbOrderKind {
...
 impl Database {
     pub async fn insert_order(&self, order: &Order) -> Result<(), InsertionError> {
         const QUERY: &str = "\
             INSERT INTO orders (
-                uid, owner, creation_timestamp, sell_token, buy_token, sell_amount, buy_amount, \
-                valid_to, app_data, fee_amount, kind, partially_fillable, signature) \
-            VALUES ( \
-                $1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13);";
+                uid, owner, creation_timestamp, sell_token, buy_token, receiver, sell_amount, buy_amount, \
+                valid_to, app_data, fee_amount, kind, partially_fillable, signature, signing_scheme) \
```
bh2smith, line 77: Nit: If each of these were on a different line it would be easier to detect which new components had been added or removed.

model/src/lib.rs
```diff
@@ -61,44 +55,31 @@ impl Signature {
...
-        hash_data[60..92].copy_from_slice(digest);
-        signing::keccak256(&hash_data)
-    }
-
-    fn signing_digest(&self, domain_separator: &DomainSeparator, digest: &[u8; 32]) -> [u8; 32] {
-        // This is fallback for wallets that don't support EIP712
-        if self.v & 0x80 == 0 {
-            Signature::signing_digest_typed_data(domain_separator, digest)
-        } else {
-            Signature::signing_digest_message(domain_separator, digest)
```
bh2smith, line 91: Have we somehow generalized the signing scheme to no longer need both scenarios here?

model/src/lib.rs
```diff
@@ -11,40 +12,33 @@ use lazy_static::lazy_static;
...
 use std::fmt;
-use web3::signing;
+use web3::signing::{self};
 use web3::signing::{Key, SecretKeyRef};
 use web3::types::Recovery;
 
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Deserialize, Serialize, Hash)]
+#[serde(rename_all = "lowercase")]
+pub enum SigningScheme {
+    EIP712,
```
bh2smith, line 22: Right, so we are only supporting this signing scheme for the moment.

solver/src/encoding.rs
```diff
@@ -1,101 +1,61 @@
...
+    Vec<u8>,  // signature
+);
 
 /// Creates the data which the smart contract's `decodeTrade` expects.
 pub fn encode_trade(
     order: &OrderCreation,
-    sell_token_index: u8,
-    buy_token_index: u8,
+    sell_token_index: usize,
+    buy_token_index: usize,
```
bh2smith, line 25: What was the purpose for this change? Does the `usize`hold an integer representation of the token's address? How was this a `u8` before?

bh2smith (approved):
Wow, this is big one! It looks like the only changes to order struct were `receiver`, `app_data` and `signing_scheme`.  Pretty cool.

orderbook/src/database/orders.rs
vkgnosis: Agree it would be easier to detect but also annoying to read because it would be so many lines. Not sure what is better.

model/src/lib.rs
vkgnosis: No. This PR assumes eip712 and I am going to make another one to add EthSign support back.

solver/src/encoding.rs
vkgnosis: You can see in the contract's settlement code that the index is a U256 now. We are never going to have that many trades so I am using usize.

bh2smith:
This looks to be in good shape now, my only outstanding questions are about how this will work if it lands directly in main. 

1. Does merging this mean we suddenly drop support for the old contracts as if they never existed? 
2. Do we have to do any sort of special coordination with devOps or dex-web team for this?

vkgnosis:
1: Yes.
2: Have to coordinate with dex-web so that the frontend continues to work. https://github.com/gnosis/gp-swap-ui/issues/338

PR #415, opened 2021-03-25
Support ethsign signatures

vkgnosis:
### Test Plan
unit tests are testing both signatures

### DIFF
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -20,6 +20,7 @@ use web3::types::Recovery;
 #[serde(rename_all = "lowercase")]
 pub enum SigningScheme {
     Eip712,
+    EthSign,
 }
 
 #[derive(Eq, PartialEq, Clone, Copy, Debug, Default, Hash)]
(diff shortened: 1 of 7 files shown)


database/sql/V008__contracts_upgrade.sql
```diff
@@ -0,0 +1 @@
+ALTER TYPE SigningScheme ADD VALUE 'ethsign';
```
fleupold, line 1: nit: the migration name should probably be "add ethsign signature or the like" (unless it goes in the same file as the contract upgrade)

model/src/lib.rs
```diff
@@ -37,6 +38,24 @@ fn hashed_eip712_message(domain_separator: &DomainSeparator, struct_hash: &[u8;
...
 }
 
+fn hashed_ethsign_message(domain_separator: &DomainSeparator, struct_hash: &[u8; 32]) -> [u8; 32] {
+    let mut message = [0u8; 60];
+    message[..28].copy_from_slice(b"\x19Ethereum Signed Message:\n32");
+    message[28..].copy_from_slice(&hashed_eip712_message(domain_separator, struct_hash));
+    signing::keccak256(&message)
+}
+
+fn signing_message(
```
fleupold, line 48: naming nit: to be more inline with the ethsign/eip712 methods maybe `hashed_signing_message()` (I kind of like the `digest` word e.g. `ethsign_message_digest` in the previous version).

fleupold (approved):
Maybe we can change the e2e test to use one order of each signature kind to make sure it works with the smart contracts?

bh2smith (approved):

vkgnosis:
I am merging this so that there is only a single, up to date contracts upgrade PR. Both PRs have been reviewed individually already.

PR #418, opened 2021-03-26
Fix new Rust version warnings

vkgnosis:
### Test Plan
still compiles

### DIFF
--- a/solver/src/liquidity/offchain_orderbook.rs
+++ b/solver/src/liquidity/offchain_orderbook.rs
@@ -20,18 +20,18 @@ impl OrderBookApi {
     }
 }
 
-impl Into<LimitOrder> for OrderCreation {
-    fn into(self) -> LimitOrder {
+impl From<OrderCreation> for LimitOrder {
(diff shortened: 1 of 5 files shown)


solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -20,18 +20,18 @@ impl OrderBookApi {
     }
 }
 
-impl Into<LimitOrder> for OrderCreation {
-    fn into(self) -> LimitOrder {
+impl From<OrderCreation> for LimitOrder {
+    fn from(order: OrderCreation) -> Self {
```
bh2smith, line 24: What was this issue here? Is Into no longer a thing?

bh2smith (approved):
Looks good, just one question

solver/src/liquidity/offchain_orderbook.rs
vkgnosis: This is the warning I got from clippy https://rust-lang.github.io/rust-clippy/master/index.html#from_over_into .

PR #426, opened 2021-03-29
Sends external price vector to the solver.

marcovc:
Could we send the price vector to the solver?

It can be used for:
* Better aligning the driver and solver objective.
* Computing fixed costs in a token that is not traded.

### DIFF
--- a/shared/src/price_estimate.rs
+++ b/shared/src/price_estimate.rs
@@ -7,7 +7,7 @@ use crate::uniswap_solver::{
 use anyhow::{anyhow, Result};
 use ethcontract::{H160, U256};
 use futures::future::join_all;
-use mockall::*;
+use mockall::automock;
 use model::{order::OrderKind, TokenPair};
 use num::{BigRational, ToPrimitive};
 use std::{
--- a/shared/src/price_estimate.rs
+++ b/shared/src/price_estimate.rs
@@ -7,7 +7,6 @@ use crate::uniswap_solver::{
 use anyhow::{anyhow, Result};
 use ethcontract::{H160, U256};
 use futures::future::join_all;
-use mockall::automock;
 use model::{order::OrderKind, TokenPair};
 use num::{BigRational, ToPrimitive};
 use std::{
@@ -17,7 +16,6 @@ use std::{
 
 const MAX_HOPS: usize = 2;
 
-#[automock]
 #[async_trait::async_trait]
 pub trait PriceEstimating: Send + Sync {
     // Price is given in how much of sell_token needs to be sold for one buy_token.
@@ -257,6 +255,18 @@ impl UniswapPriceEstimator {
                 ))
             })?,
         ))
+    }
+}
+
+pub struct FakePriceEstimator(pub BigRational);
+#[async_trait::async_trait]
+impl PriceEstimating for FakePriceEstimator {
+    async fn estimate_price(&self, _: H160, _: H160, _: U256, _: OrderKind) -> Result<BigRational> {
+        Ok(self.0.clone())
+    }
+
+    async fn estimate_gas(&self, _: H160, _: H160, _: U256, _: OrderKind) -> Result<U256> {
+        Ok(100_000.into())
     }
 }
 
--- a/shared/src/price_estimate.rs
+++ b/shared/src/price_estimate.rs
@@ -7,14 +7,17 @@ use crate::uniswap_solver::{
 use anyhow::{anyhow, Result};
 use ethcontract::{H160, U256};
 use futures::future::join_all;
+use mockall::*;
 use model::{order::OrderKind, TokenPair};
 use num::{BigRational, ToPrimitive};
 use std::{
     cmp::Reverse,
     collections::{HashMap, HashSet},
 };
+
 const MAX_HOPS: usize = 2;
 
+#[automock]
 #[async_trait::async_trait]
 pub trait PriceEstimating: Send + Sync {
     // Price is given in how much of sell_token needs to be sold for one buy_token.
--- a/solver/src/http_solver.rs
+++ b/solver/src/http_solver.rs
@@ -9,6 +9,7 @@ use crate::{
 };
 use ::model::order::OrderKind;
 use anyhow::{ensure, Context, Result};
+use futures::join;
 use num::{BigRational, ToPrimitive};
 use primitive_types::H160;
 use reqwest::{header::HeaderValue, Client, Url};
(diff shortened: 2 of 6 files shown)


bh2smith (approved):
Seems fine with me!

solver/src/http_solver.rs
```diff
@@ -190,6 +203,16 @@ impl HttpSolver {
             .get_token_infos(addresses.as_slice())
             .await;
 
+        let price_estimates: HashMap<H160, Result<BigRational>> = addresses
+            .iter()
+            .cloned()
+            .zip(
+                self.price_estimator
```
fleupold, line 210: nit: estimate_price and the other async operation (get_token_infos) could happen in parallel and be awaited on via a join!

fleupold (approved):

shared/src/price_estimate.rs
```diff
@@ -7,14 +7,17 @@ use crate::uniswap_solver::{
 use anyhow::{anyhow, Result};
 use ethcontract::{H160, U256};
 use futures::future::join_all;
+use mockall::*;
```
vkgnosis, line 10: We're only using `automock` so better to just import that.

solver/src/http_solver.rs
```diff
@@ -190,6 +203,16 @@ impl HttpSolver {
...
+                    .estimate_prices(addresses.as_slice(), addresses[0])
```
vkgnosis, line 211: `addresses[0]` will panic if there are 0 addresses in total. Just saying because I'm not sure if we guarantee that there is at least one address at this point.

vkgnosis (approved):

solver/src/http_solver.rs
marcovc: I believe it must because of [this](https://github.com/gnosis/gp-v2-services/blob/112c13ac2b04d6401d5092c5f495148627cc2d14/solver/src/http_solver.rs#L333-L335), right?

vkgnosis: Yes, looks right

solver/src/http_solver.rs
marcovc: Wouldn't I need to wrap these two futures into some common abstraction in order to aggregate them?

vkgnosis: In this case no.

```rust
let (token_infos, price_estimates) = futures::join!(
    self.token_info_fetcher
        .get_token_infos(addresses.as_slice()),
    self.price_estimator
        .estimate_prices(addresses.as_slice(), addresses[0]),
);

let price_estimates: HashMap<H160, Result<BigRational>> =
    addresses.iter().cloned().zip(price_estimates).collect();
```
I am using the join! macro https://docs.rs/futures/0.3.13/futures/future/fn.join.html . There is also a plain function https://docs.rs/futures/0.3.13/futures/macro.join.html . The macro has the advantage of being able to take any number of arguments.

marcovc: Thanks, that seems to work (even if my editor complains that join! is a unresolved macro call).

vkgnosis: I think you have to enable the rust analyzer setting "proc macro".

marcovc: Thanks, it seems it is on:

Screenshot_20210330_115124

But this is minor. Thanks!

solver/src/http_solver.rs
```diff
@@ -58,6 +64,7 @@ impl HttpSolver {
         config: SolverConfig,
         native_token: H160,
         token_info_fetcher: &Arc<dyn TokenInfoFetching>,
+        price_estimator: &Arc<dyn PriceEstimating>,
```
nlordell, line 67: Kind of unrelated to this PR, but in this situation, its every so slightly better to move the `Arc<...>` instead of passing a reference.

The reason for this is imagine the code:
```rust
fn foo(val: &Arc<...>) {
  // ...
  val.clone();
  // ...
}

// Some where else:
let val = Arc::new(...);
foo(&val);
```

In this case, we simultaneously have 2 shared pointers to `val` (one in the outer method and one in the inner method). So, reference counting in general is very fast, so the overhead is minimal, but there is one key optimization that you prevent `foo` from doing, and that is using [`Arc::get_mut`](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.get_mut) to get an internal mutable reference, which only works when there is only a single reference to the value.

In general, WRT to ownership, I think its better to ask what the method needs (here an owned `Arc`) instead of a reference and unconditionally `clone`-ing. This gives more control to the caller and can help with overall code optimizations.

This is very small and very nit-picky, and its totally fine as it is here.

solver/src/http_solver.rs
```diff
@@ -185,10 +199,16 @@ impl HttpSolver {
...
+
+        let (token_infos, price_estimates) = join!(
+            self.token_info_fetcher
+                .get_token_infos(addresses.as_slice()),
+            self.price_estimator
+                .estimate_prices(addresses.as_slice(), addresses[0]),
+        );
+
+        let price_estimates: HashMap<H160, Result<BigRational>> =
+            addresses.iter().cloned().zip(price_estimates).collect();
```
nlordell, line 211: Gotta love iterators!

nlordell (approved):
Looks very good.

solver/src/http_solver.rs
marcovc: Thanks for the tip!

I thought it was the opposite. When I see `foo(val)`, without reference, I immediately think that the program will be incrementing the shared ptr count when calling the function, and decrementing when returning, which can be avoided by passing by reference, since that doesn't require cloning the shared_ptr (which changes the ref count under the hoods). 

I suppose this doesn't happen when passing "by move" instead of "by value", which I'm still getting used to :). And if I pass "by move" then I suppose the compiler knows there will be only one pointer to the value (unless it has been cloned before?). so the `get_mut` part makes sense to me. 

So, in general, what is the use case to pass things by const reference instead of moving them?

vkgnosis: As you said, Rust is by move by default whereas c++ is by copy. In this case the struct is going to store an owned Arc anyway so there is no reason to pass in a reference to this function and then clone internally. By passing in a value you give the caller maximum choice. If they already have an owned Arc that they no longer need they can pass it without copy and if they want to copy they can.

You would pass a reference when you all you need is a reference because then callers don't need to give you an owned value.

nlordell: > I thought it was the opposite. When I see foo(val), without reference, I immediately think that the program will be incrementing the shared ptr count when calling the function, and decrementing when returning,

C++ copy semantics :stuck_out_tongue:. So in Rust, when not using a reference it will `move` instead of `copy`. This means that the reference count will not be incremented, but as you mentioned, it will get decremented once `foo` returns in my example. If the calling context would need to use `val` after calling `foo`, then it would have to:
```rust
let val = Arc::new(...);
foo(val.clone());
// ...
val.do_stuff();
```

This is desired as it allows the caller to control how to best share the resource, i.e. it can decide if it doesn't need it anymore and can move it into `foo`, or if it needs it after `foo`, then

On the other hand, in the code here, the `HttpSolver` needs to unconditionally take ownership of the `Arc`. Passing it by reference here, removes choice from the caller (i.e. the `Arc` will always be cloned, even when it doesn't need to be in the case the caller doesn't need the value any more). Another difference is that since `HttpSolver` is taking ownership of `val`, this means that the shared `Arc` pointer would only be `drop`-ped when the `HttpSolver` gets `drop`-ped.

Going back to my example. If, for example, `foo` only needed a temporary reference to `val`, then you could implement it like:
```rust
fn foo(val: &...) {
  // ...
}

// Some where else:
let val = Arc::new(...);
foo(&val);
```

This allows the caller to provide any reference to the value, regardless if its boxed in a `Box` or a smart pointer like `Arc` or `Rc`. This is possible because [`Arc<T>` implements `Deref<Target = T>`](https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-Deref).


> So, in general, what is the use case to pass things by const reference instead of moving them?

When you only need a reference. In the `HttpSolver::new()` case, it needs an owned value, so the method signature should express that. (Although, to be fair, it is a very nit-picky suggestion, since cloning `Arc` is practically for free).

PR #427, opened 2021-03-29
[NaiveSolver] Using BigRational in in/out comp. to avoid rounding issues

### LINKED ISSUE #421: [Naive Solver] Invalid Solution when pool reserves << excess amount
The issue was raised by this [alert](https://gnosisinc.slack.com/archives/CPZA1AGKY/p1616769723024400).

For the instance with a single [trade](https://protocol-explorer.dev.gnosisdev.com/rinkeby/orders/0x3ebebb0e80b7fa47f7f371ed7b4acbfb153385ba2bf99e3ec877a2f4a978971204a66cbba0485d7b21af836f52b711401300fddb605ddeea), both the Naive as well as the BaselineSolver found a solution. The naive solver claimed a higher objective value and therefore tried to submit first. 

The settlement simulation failed because we were trying to send a much larger amount of the buyTokens to the trader than we received from uniswap. This is because the pool used only had very small reserves (~10e7 when we were trying to trade ~10e11). Therefore the direct pool is unusable and the solution found by the Baseline solver was eventually accepted.

The naive solver still found the invalid solution, because when verifying a solution candidate in the code (cf. `is_valid_solution`) we only check if the limit prices are satisfied, not if the amounts traded lead to token conservation.

I tried to make change (https://github.com/gnosis/gp-v2-services/compare/naive_solver_bug?expand=1) that adds token conservation to the checks, however, I'm running into rounding errors leading to marginally unconserved token balances.

I'm not sure at the moment how to proceed. One option would be to pass in a "buffer size" and allow token conservation to be violated within this buffer. Another option would be to adjust the amount computation in the smart contract to always take a bit more and pay out a bit less to the traders to avoid small negative balances in the contract.

cc @twalth3r and @marcovc as you guys have probably thought about rounding issues with regards to token conservation as well.

fleupold:
Hopefully 🤞 Fixes #421 

For more history on the algorithm, cf. https://docs.google.com/document/d/1jS22wxbCqo88fGsqEMZgRQgiAcHlPqxoMw3CJTHst6c/edit

See added test case for the problematic instance. Let me try to explain what happens:

In the instance we are trading a very large amount over a very small (in terms of reserve values) AMM:
- In order to balance excess supply and shortage demand we have to provide **all** our excess tokens into the AMM. 
- This results in moving the uniswap price to almost 0/Inf (depending from which side you look) and a `uniswap_out` value of `|reserve| - epsilon`, with **epsilon << 1** (note that an exact `|reserve|` amount can only be reached if we provide an infinite amount of excess token). 
- Due to the division inside `compute_uniswap_out` the whole integer `uniswap_out` value is `|reserve| - 1`. This value is in turn used to compute the exact `uniswap_in` amount based on the Uniswap formula.
-  Because the tiny epsilon was rounded to 1 and the crazy prices, this results in an `in_amount`  << `excess.sell_volume` (although those values should be roughly equal).
- This then down the line leads to a "valid solution" which respects the limit price but ultimately would fail in the token conservation (a property that we currently don't validate in the driver).

I tried different approaches such as validating token conservation in `is_valid_solution`, incrementing the computed `uniswap_out` amount to deal with the rounding error, trying to spot when our `uniswap_in` computation is less than expected, but always ran into issues with other test cases (requiring some form of buffer and knowledge of that buffer in the test cases).

Since the problem stems from the fact that `|reserve| - epsilon` becomes `|reserve| - 1` in `compute_uniswap_out` and then leads to a insufficient `uniswap_in` amount, the solution proposed in this PR is to compute these values using rationals (thus conserving the very small epsilon) and only convert to U256 for computing the final interactions. 

This addresses the problem without requiring knowledge of buffers or another larger change to the settlement logic.

### Test Plan
Added Test Case

### DIFF
--- a/shared/src/conversions.rs
+++ b/shared/src/conversions.rs
@@ -1,10 +1,18 @@
-use num::bigint::Sign;
+use anyhow::{anyhow, Result};
 use num::BigInt;
+use num::{bigint::Sign, Zero};
 use num::{BigRational, ToPrimitive as _};
 use primitive_types::U256;
 
 pub fn big_rational_to_float(ratio: &BigRational) -> Option<f64> {
     Some(ratio.numer().to_f64()? / ratio.denom().to_f64()?)
+}
+
+pub fn big_rational_to_u256(ratio: &BigRational) -> Result<U256> {
+    if ratio.denom() == &BigInt::zero() {
+        return Err(anyhow!(" Division by 0 in BigRational to U256 conversion"));
+    }
+    big_int_to_u256(&(ratio.numer() / ratio.denom()))
 }
 
 pub fn u256_to_big_int(input: &U256) -> BigInt {
@@ -16,6 +24,17 @@ pub fn u256_to_big_int(input: &U256) -> BigInt {
 pub fn u256_to_big_rational(input: &U256) -> BigRational {
     let as_bigint = u256_to_big_int(input);
     BigRational::new(as_bigint, 1.into())
+}
+
+pub fn big_int_to_u256(input: &BigInt) -> Result<U256> {
+    let (sign, bytes) = input.to_bytes_be();
+    if sign == Sign::Minus {
+        return Err(anyhow!("Negative BigInt to U256 conversion"));
+    }
+    if bytes.len() > 32 {
+        return Err(anyhow!("BigInt too big for U256 conversion"));
+    }
+    Ok(U256::from_big_endian(&bytes))
 }
 
 // Convenience:
--- a/shared/src/conversions.rs
+++ b/shared/src/conversions.rs
@@ -1,4 +1,4 @@
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, ensure, Result};
 use num::BigInt;
 use num::{bigint::Sign, Zero};
 use num::{BigRational, ToPrimitive as _};
@@ -9,9 +9,10 @@ pub fn big_rational_to_float(ratio: &BigRational) -> Option<f64> {
 }
 
 pub fn big_rational_to_u256(ratio: &BigRational) -> Result<U256> {
-    if ratio.denom() == &BigInt::zero() {
-        return Err(anyhow!(" Division by 0 in BigRational to U256 conversion"));
-    }
+    ensure!(
+        !ratio.denom().is_zero(),
+        "Division by 0 in BigRational to U256 conversion"
+    );
     big_int_to_u256(&(ratio.numer() / ratio.denom()))
 }
 
@@ -28,12 +29,8 @@ pub fn u256_to_big_rational(input: &U256) -> BigRational {
 
 pub fn big_int_to_u256(input: &BigInt) -> Result<U256> {
     let (sign, bytes) = input.to_bytes_be();
-    if sign == Sign::Minus {
-        return Err(anyhow!("Negative BigInt to U256 conversion"));
-    }
-    if bytes.len() > 32 {
-        return Err(anyhow!("BigInt too big for U256 conversion"));
-    }
+    ensure!(sign == Sign::Plus, "Negative BigInt to U256 conversion");
+    ensure!(bytes.len() <= 32, "BigInt too big for U256 conversion");
     Ok(U256::from_big_endian(&bytes))
 }
 
--- a/shared/src/conversions.rs
+++ b/shared/src/conversions.rs
@@ -1,4 +1,4 @@
-use anyhow::{anyhow, ensure, Result};
+use anyhow::{ensure, Result};
 use num::BigInt;
 use num::{bigint::Sign, Zero};
 use num::{BigRational, ToPrimitive as _};
--- a/solver/src/naive_solver/multi_order_solver.rs
+++ b/solver/src/naive_solver/multi_order_solver.rs
@@ -229,8 +229,8 @@ fn compute_uniswap_out(shortage: &TokenContext, excess: &TokenContext) -> BigRat
     let numerator_subtrahend = 1000
         * (u256_to_big_int(&shortage.sell_volume) - u256_to_big_int(&shortage.buy_volume))
         * u256_to_big_int(&excess.reserve);
-    let denominator = (1000 * u256_to_big_int(&excess.reserve))
-        + (997 * (u256_to_big_int(&excess.sell_volume) - u256_to_big_int(&excess.buy_volume)));
+    let denominator = 1000 * u256_to_big_int(&excess.reserve)
+        + 997 * (u256_to_big_int(&excess.sell_volume) - u256_to_big_int(&excess.buy_volume));
     BigRational::new(numerator_minuend - numerator_subtrahend, denominator)
 }
 
--- a/solver/src/naive_solver/multi_order_solver.rs
+++ b/solver/src/naive_solver/multi_order_solver.rs
@@ -4,7 +4,7 @@ use crate::{
 };
 use liquidity::{AmmOrder, LimitOrder};
 use model::order::OrderKind;
-use num::BigRational;
+use num::{BigInt, BigRational, Zero};
 use primitive_types::U256;
 use shared::conversions::{big_rational_to_u256, u256_to_big_int, U256Ext};
 use std::collections::HashMap;
@@ -229,8 +229,12 @@ fn compute_uniswap_out(shortage: &TokenContext, excess: &TokenContext) -> BigRat
     let numerator_subtrahend = 1000
         * (u256_to_big_int(&shortage.sell_volume) - u256_to_big_int(&shortage.buy_volume))
         * u256_to_big_int(&excess.reserve);
-    let denominator = 1000 * u256_to_big_int(&excess.reserve)
+    let denominator: BigInt = 1000 * u256_to_big_int(&excess.reserve)
         + 997 * (u256_to_big_int(&excess.sell_volume) - u256_to_big_int(&excess.buy_volume));
+    assert!(
+        !denominator.is_zero(),
+        "reserve values cannot be zero (it's impossible to fully drain a pool)"
+    );
     BigRational::new(numerator_minuend - numerator_subtrahend, denominator)
 }
 
--- a/solver/src/naive_solver/multi_order_solver.rs
+++ b/solver/src/naive_solver/multi_order_solver.rs
@@ -2,15 +2,15 @@ use crate::{
     liquidity,
     settlement::{Interaction, Settlement, Trade},
 };
-use anyhow::{anyhow, Result};
 use liquidity::{AmmOrder, LimitOrder};
 use model::order::OrderKind;
-use num::{bigint::Sign, BigInt};
+use num::BigRational;
 use primitive_types::U256;
+use shared::conversions::{big_rational_to_u256, u256_to_big_int};
 use std::collections::HashMap;
 use web3::types::Address;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 struct TokenContext {
     address: Address,
     reserve: U256,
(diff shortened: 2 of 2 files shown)


solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -43,9 +43,11 @@ pub fn solve(
     let mut orders: Vec<LimitOrder> = orders.collect();
     while !orders.is_empty() {
         let (context_a, context_b) = split_into_contexts(orders.clone().into_iter(), pool);
-        let solution = solve_orders(orders.clone().into_iter(), &pool, &context_a, &context_b)?;
-        if is_valid_solution(&solution) {
-            return Some(solution);
+        if let Some(valid_solution) =
```
fleupold, line 46: This change is somewhat unrelated, but it seems wrong to return if the current set of orders isn't solvable (note the ? before). Instead we should use the same strategy as with invalid solutions (remove order with worst limit price and try again).

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -217,30 +222,32 @@ fn split_into_contexts(
 /// The derivation of this formula is described in https://docs.google.com/document/d/1jS22wxbCqo88fGsqEMZgRQgiAcHlPqxoMw3CJTHst6c/edit
 /// It assumes GP fee (φ) to be 1 and Uniswap fee (Φ) to be 0.997
 ///
-fn compute_uniswap_out(shortage: &TokenContext, excess: &TokenContext) -> U256 {
+fn compute_uniswap_out(shortage: &TokenContext, excess: &TokenContext) -> BigRational {
```
fleupold, line 225: This is the main change.

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -217,30 +222,32 @@ fn split_into_contexts(
...
 /// Taken from: https://github.com/Uniswap/uniswap-v2-periphery/blob/4123f93278b60bcf617130629c69d4016f9e7584/contracts/libraries/UniswapV2Library.sol#L53
+/// Not adding + 1 in the end, because we are working with rationals and thus don't round up.
 ///
-fn compute_uniswap_in(out: U256, shortage: &TokenContext, excess: &TokenContext) -> Option<U256> {
-    U256::from(1000)
-        .checked_mul(out)?
-        .checked_mul(excess.reserve)?
-        .checked_div(U256::from(997).checked_mul(shortage.reserve.checked_sub(out)?)?)?
-        .checked_add(U256::from(1))
+fn compute_uniswap_in(
```
fleupold, line 243: And this

shared/src/conversions.rs
```diff
@@ -1,12 +1,20 @@
...
 use primitive_types::U256;
 
 pub fn big_rational_to_float(ratio: &BigRational) -> Option<f64> {
     Some(ratio.numer().to_f64()? / ratio.denom().to_f64()?)
 }
 
+pub fn big_rational_to_u256(ratio: &BigRational) -> Result<U256> {
+    if ratio.denom() == &BigInt::zero() {
+        return Err(anyhow!(" Division by 0 in BigRational to U256 conversion"));
+    }
```
vkgnosis, line 14: Just wondering if that is the canonical representation of zero in BigRational or if for example numer being 0 with nonzero denom could also be it. Could maybe use https://docs.rs/num-traits/0.2.12/num_traits/identities/trait.Zero.html ?
Or even better use BigRational::to_integer and then check if that is zero or conver it.

vkgnosis (approved):

marcovc:
Do I understand correctly that this instance could never happen in reality anyway? https://github.com/Uniswap/uniswap-v2-core/blob/4dd59067c76dea4a0e8e4bfdda41877a6b16dedc/contracts/UniswapV2Pair.sol#L162

fleupold:
Right, it's impossible to withdraw all funds from a pool. In reality the amount out would still `|reserve| - 1` (when rounding the big ration to U256), however it would use a very high in amount thus making the price close to 0, which should violate any reasonable limit price.

shared/src/conversions.rs
marcovc: I recall there is a `num::zero()` as well.

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -217,30 +222,32 @@ fn split_into_contexts(
...
-        * u256_to_bigint(&excess.reserve);
-    let denominator = (1000 * u256_to_bigint(&excess.reserve))
-        + (997 * (u256_to_bigint(&excess.sell_volume) - u256_to_bigint(&excess.buy_volume)));
-    bigint_to_u256(&((numerator_minuend - numerator_subtrahend) / denominator))
-        .expect("uniswap_out should always be U256 compatible if excess is chosen correctly")
+        * (u256_to_big_int(&shortage.sell_volume) - u256_to_big_int(&shortage.buy_volume))
+        * u256_to_big_int(&excess.reserve);
+    let denominator = (1000 * u256_to_big_int(&excess.reserve))
+        + (997 * (u256_to_big_int(&excess.sell_volume) - u256_to_big_int(&excess.buy_volume)));
+    BigRational::new(numerator_minuend - numerator_subtrahend, denominator)
```
marcovc, line 234: Can denominator be zero here?

twalth3r (approved):
Looks good to me - I think it's a pretty elegant solution!

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -217,30 +222,32 @@ fn split_into_contexts(
...
-        .checked_div(U256::from(997).checked_mul(shortage.reserve.checked_sub(out)?)?)?
-        .checked_add(U256::from(1))
+fn compute_uniswap_in(
+    out: BigRational,
+    shortage: &TokenContext,
+    excess: &TokenContext,
+) -> BigRational {
+    BigRational::from_integer(1000.into()) * out.clone() * u256_to_big_int(&excess.reserve)
+        / (BigRational::from_integer(997.into())
+            * (BigRational::from(u256_to_big_int(&shortage.reserve)) - out))
```
marcovc, line 250: I believe there is, or used to be, a `u256_to_big_rational`

shared/src/conversions.rs
```diff
@@ -1,12 +1,20 @@
...
 use num::{BigRational, ToPrimitive as _};
 use primitive_types::U256;
 
 pub fn big_rational_to_float(ratio: &BigRational) -> Option<f64> {
     Some(ratio.numer().to_f64()? / ratio.denom().to_f64()?)
 }
 
+pub fn big_rational_to_u256(ratio: &BigRational) -> Result<U256> {
+    if ratio.denom() == &BigInt::zero() {
+        return Err(anyhow!(" Division by 0 in BigRational to U256 conversion"));
```
nlordell, line 13:
```suggestion
        return Err(anyhow!("Division by 0 in BigRational to U256 conversion"));
```

or even better, you can use the `ensure!` macro, which is an `assert` which errors instead of panicking:
```rust
    ensure!(ratio.denom() == &BigInt::zero(), "Division by 0 in BigRational to U256 conversion");
```

shared/src/conversions.rs
```diff
@@ -18,6 +26,17 @@ pub fn u256_to_big_rational(input: &U256) -> BigRational {
     BigRational::new(as_bigint, 1.into())
 }
 
+pub fn big_int_to_u256(input: &BigInt) -> Result<U256> {
+    let (sign, bytes) = input.to_bytes_be();
+    if sign == Sign::Minus {
```
nlordell, line 31: nanonit: I think both these `if` statements can be replaced with `ensure` as well.

nlordell (approved):

marcovc (approved):
Looks good. I still fail to see how this is not a problem for uniswap as well, and if it is, why can't we solve it like they do. It seems the function to compute the input from the output is this one: https://github.com/Uniswap/uniswap-v2-periphery/blob/master/contracts/libraries/UniswapV2Library.sol#L53-L58 . I wonder if that "+1" in the end is used to solve this problem?

solver/src/naive_solver/multi_order_solver.rs
bh2smith: Technically, since 
```
denominator = (1000 * reserve) + (997 * (sell_volume - buy_volume))
```

It looks like there is a whole 2d plane on which this could be zero. One solution, in fact, is when excess reserve, sell and buy volume are all zero.

solver/src/naive_solver/multi_order_solver.rs
```diff
@@ -217,30 +222,32 @@ fn split_into_contexts(
...
-        * (u256_to_bigint(&shortage.sell_volume) - u256_to_bigint(&shortage.buy_volume))
-        * u256_to_bigint(&excess.reserve);
-    let denominator = (1000 * u256_to_bigint(&excess.reserve))
-        + (997 * (u256_to_bigint(&excess.sell_volume) - u256_to_bigint(&excess.buy_volume)));
-    bigint_to_u256(&((numerator_minuend - numerator_subtrahend) / denominator))
-        .expect("uniswap_out should always be U256 compatible if excess is chosen correctly")
+        * (u256_to_big_int(&shortage.sell_volume) - u256_to_big_int(&shortage.buy_volume))
+        * u256_to_big_int(&excess.reserve);
+    let denominator = (1000 * u256_to_big_int(&excess.reserve))
+        + (997 * (u256_to_big_int(&excess.sell_volume) - u256_to_big_int(&excess.buy_volume)));
```
bh2smith, line 233: Nit: I think the brackets used before the 1000 and the 997 are unnecessary.

bh2smith (approved):

solver/src/naive_solver/multi_order_solver.rs
fleupold: reserve values can not be zero (since it's not possible to extract the entire amount of a Uniswap pool). Therefore denominator can't be zero. I will add a comment in the code...

bh2smith: What about the case when `buy_volume` is much larger than `sell_volume` causing the second term here to be negative enough that it negates or nullifies the first term?

Another integer solution to this is 
```
reserve = 997
sell_volume = 1
buy_volume = 1001
```

fleupold:
>  I still fail to see how this is not a problem for uniswap as well, and if it is, why can't we solve it like they do.

@marcovc our problem is that we do **two calculations** (back and forth on the pool balances):
1. How much *out_amount* do we need to get from uniswap so that demand and supply match
2. How much *in_amount* do we need to supply into uniswap so that we get *out_amount* computed in 1)

If we round down in 1) and use the same formula as Uniswap (with +1 rounding) for computing *amount_in* in 2) we might get an *amount_in* that is much smaller than the *amount_in* that an exact computation would have yielded (in our example case *amount_in* should have been the entire sell amount, but due to extreme prices was only a fraction of it). 

As such we are not actually matching supply and demand (token conservation will be violated in the end). This is how the code behaved before this change.

If we round up in 1) we start descend into a rabbit hole of failing test cases because we are slightly violating constraints by very small epsilons (we are assuming to get slightly more out of uniswap than it will give us).

Therefore using rational values and deferring rounding to later seemed like the best option.

marcovc:
> our problem is that we do **two calculations** (back and forth on the pool balances):

Thanks this was the part I was missing. All clear now, and it makes perfect sense that the first step uses rationals.

solver/src/naive_solver/multi_order_solver.rs
fleupold: Good point. Maybe it's safe to return a result here. Will fix in a follow up PR

Issue #420, opened 2021-03-26
Divide by 0 panic computing objective value


nlordell:
We encountered the following panic computing the objective value:
```
ERROR (pod: dev-dfusion-v2-solver-rinkeby-85d5b74797-mvvmk):
2021-03-26T15:05:43.197Z ERROR shared::tracing: thread 'main' panicked at 'attempt to divide by zero', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/num-bigint-0.4.0/src/biguint/division.rs:160:9:
  0: std::panicking::begin_panic
  1: num_bigint::biguint::division::div_rem_ref
  2: num_bigint::bigint::division::<impl core::ops::arith::Div for num_bigint::bigint::BigInt>::div
  3: <num_rational::Ratio<T> as core::ops::arith::Div>::div
  4: solver::settlement::Settlement::objective_value
  5: solver::driver::Driver::single_run::{{closure}}
  6: solver::main::{{closure}}
  7: tokio::runtime::enter::Enter::block_on
  8: tokio::runtime::thread_pool::ThreadPool::block_on
  9: tokio::runtime::context::enter
 10: tokio::runtime::handle::Handle::enter
 11: solver::main
```

marcovc:
I suppose we don't have the example that caused this, right?

nlordell:
I'm not sure where to look for it :see_no_evil:

marcovc:
It appears there is no division on function #4 in the "stack trace" above. Could it be that the code is inlined due to compiler optimizations?

vkgnosis:
There are divisions in the functions objective_value calls so that's probably it. Although I am surprised that we don't get line numbers here. Will investigate.

marcovc:
> There are divisions in the functions objective_value calls so that's probably it. 

But if the division is happening from a function called by the objective_value, I was expecting that that function is also part of the stack trace. Unless it got inlined by the compiler or something.

vkgnosis:
Yes, I mean the inlining is probably the cause for that.

bh2smith:
The division occurs in three places from within `objective_value`

1. `buy_order_surplus` here 
http://github.com/gnosis/gp-v2-services/blob/a7cd0ed35f3bfa1954d8e943978da1784224f6f4/solver/src/settlement.rs#L178-L179
2. `sell_order_surplus` here - 
[...]

PR #428, opened 2021-03-30
Include line debug information in docker build

vkgnosis:
By default rust does not include line debug information in release
builds which we have seen for example in the divison by 0 panic in the
objective_value calcuation.
By setting the env variable in this commit the line number information
is included. This increases the size of the binaries and thus the docker
image significantly. The image goes from 55 MB to 210 MB.
I feel that this is worth the tradeoff.

See https://github.com/gnosis/gp-v2-services/issues/420
https://doc.rust-lang.org/cargo/reference/profiles.html#debug

### Test Plan
Locally tested by building and changing code to include a panic

### DIFF
--- a/docker/Dockerfile.binary
+++ b/docker/Dockerfile.binary
@@ -3,7 +3,7 @@ WORKDIR /usr/src/oba-services
 
 # Copy and Build Code
 COPY . .
-RUN cargo build --target x86_64-unknown-linux-musl --release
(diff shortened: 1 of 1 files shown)


fleupold (approved):
Agree, binary size is not really sensitive at the moment.

bh2smith (approved):

PR #429, opened 2021-03-30
Event handler: OrderInvalidation

bh2smith:
With the contract upgrades in place we are now able to listen for and handle the `OrderInvalidation` event whose logic was already implemented. This PR just connects the listener to the handler. 

### Test Plan
Existing CI.

Test the logic by simulating an invalidation locally

1. Setup a new database (if you have been running the services software on a different network in the past)
2. Run the order book service locally on rinkeby with

```sh
cargo run --bin orderbook -- --node-url https://rinkeby.infura.io/v3/<YOUR_INFURA_KEY> --log-filter warn   
```

3. Run the trading interface locally with API pointed at 'http://0.0.0.0:8080/api/v1' (note you will have to manually edit the URLS [here](https://github.com/gnosis/gp-swap-ui/blob/00a0ed063d6434efd793b34f3aa730a7a44002ba/src/custom/utils/operator.ts#L10-L14))

```sh
yarn && yarn start
```

4. Place an order via the [local interface](http://localhost:3000/#/swap) and copy the `UID`

5. Navigate to the smart contract on [etherscan](https://rinkeby.etherscan.io/address/0x3328f5f2cEcAF00a2443082B657CedEAf70bfAEf#writeContract) and invalidate this order via the "Write Contract" interface

6. After this transaction has been mined, you can query your local database for the order to see if the invalidated field is set to `true`.

```sh
curl -X GET "http://localhost:8080/api/v1/orders/0xb6db1950bd36f2fbd27042d746a3233f229a5807df1a9e2a79ae201f76df92c47f01d9b227593e033bf8d6fc86e634d27aa8556860661201" -H  "accept: application/json"
```

### DIFF
--- a/orderbook/src/event_updater.rs
+++ b/orderbook/src/event_updater.rs
@@ -1,11 +1,14 @@
 use crate::database::{
-    Database, Event as DbEvent, EventIndex as DbEventIndex, Settlement as DbSettlement,
-    Trade as DbTrade,
+    Database, Event as DbEvent, EventIndex as DbEventIndex, Invalidation as DbInvalidation,
+    Settlement as DbSettlement, Trade as DbTrade,
 };
 use anyhow::{anyhow, Context, Error, Result};
 use contracts::{
     g_pv_2_settlement::{
-        event_data::{Settlement as ContractSettlement, Trade as ContractTrade},
+        event_data::{
+            OrderInvalidated as ContractInvalidation, Settlement as ContractSettlement,
+            Trade as ContractTrade,
+        },
         Event as ContractEvent,
     },
     GPv2Settlement,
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):

fleupold (approved):
Could you do a manual test where you actually invalidate the order onchain (e.g. on Rinkeby) and observe it no longer being returned from the orderbook ("hard cancellation")?

bh2smith:
Since this PR is based on the upgraded contracts, I think the manual e2e test will have to wait until all the components are working together.

nlordell (approved):

bh2smith:
@fleupold - Unfortunately it is going to be a challenge for us to simulate this locally because I would have to run the orderbook with event sync. Also, My local front is is not able to communicate with my local backend at the moment (failing to send orders). I may not get back to this until next week. 

Logging this here for now.
```
Order Not accepted by the operator

Error handling a 400 error. Likely a problem deserialising the JSON response
```

PR #432, opened 2021-03-30
Better defend against potential 0 division in naive solver

fleupold:
Fixes https://github.com/gnosis/gp-v2-services/pull/427#discussion_r603978442

Adding a checked new helper for Ratio since it seems like this is a safety mechanism we might want to do elsewhere too.

### Test Plan
Unit tests still passing

### DIFF
--- a/shared/src/conversions.rs
+++ b/shared/src/conversions.rs
@@ -1,6 +1,6 @@
 use anyhow::{ensure, Result};
-use num::BigInt;
 use num::{bigint::Sign, Zero};
+use num::{rational::Ratio, BigInt};
 use num::{BigRational, ToPrimitive as _};
 use primitive_types::U256;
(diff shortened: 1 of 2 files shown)


vkgnosis (approved):

marcovc (approved):
Seems good to me!

bh2smith (approved):

shared/src/conversions.rs
```diff
@@ -36,6 +36,20 @@ pub fn big_int_to_u256(input: &BigInt) -> Result<U256> {
 
 // Convenience:
 
+pub trait RatioExt<T> {
```
nlordell, line 39: fancy :jeans:

nlordell (approved):
Looks great!

PR #433, opened 2021-03-30
Fix div by zero

### LINKED ISSUE #420: Divide by 0 panic computing objective value

marcovc:
Closes #420 .

Added test case.

### DIFF
--- a/solver/src/settlement.rs
+++ b/solver/src/settlement.rs
@@ -393,7 +393,6 @@ mod tests {
         );
     }
 
-
     #[test]
     fn test_computing_objective_value_with_zero_prices() {
         // Test if passing a clearing price of zero to the objective value function does
@@ -412,7 +411,7 @@ mod tests {
         };
 
         let trade = Trade {
-            order: order,
+            order,
             executed_amount: 10.into(),
             ..Default::default()
         };
@@ -420,7 +419,7 @@ mod tests {
         let clearing_prices = maplit::hashmap! {token0 => 1.into(), token1 => 0.into()};
 
         let settlement = Settlement {
-            clearing_prices: clearing_prices,
+            clearing_prices,
             trades: vec![trade],
             ..Default::default()
         };
--- a/solver/src/settlement.rs
+++ b/solver/src/settlement.rs
@@ -140,6 +140,13 @@ impl Settlement {
                 .get(&trade.order.buy_token)
                 .expect("Solution with trade but without price for buy token");
 
+            if trade.order.kind == OrderKind::Sell && buy_token_clearing_price.is_zero() {
+                return None;
+            }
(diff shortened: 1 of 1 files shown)


marcovc:
Note: The MIP solver is not supposed to generate zero prices. I guess the other two should be investigated for this case as well.

solver/src/settlement.rs
```diff
@@ -140,6 +140,13 @@ impl Settlement {
                 .get(&trade.order.buy_token)
                 .expect("Solution with trade but without price for buy token");
 
+            if trade.order.kind == OrderKind::Sell && buy_token_clearing_price.is_zero() {
+                return None;
+            }
+            if trade.order.kind == OrderKind::Buy && sell_token_clearing_price.is_zero() {
+                return None;
+            }
```
vkgnosis, line 148: Another way to write this could be

```suggestion
            if match trade.order.kind {
                OrderKind::Buy => &sell_token_clearing_price,
                OrderKind::Sell => &buy_token_clearing_price,
            }
            .is_zero()
            {
                return None;
            }

```

I like that because it uses match which I find clearer than if because it makes it obvious that the branches are exclusive and that all enum cases are handled. Either way is fine, just wanted to suggest it.

vkgnosis (approved):
Looks good. Agree that zero prices don't make sense and other solvers shouldn't be emitting them. Maybe something we can eventually enforce on the Settlement struct.

fleupold (approved):
I wonder if instead of using these one-off 0 checks, we should rather start using checked_div whenever we divide and then just `?` at the end.

bh2smith (approved):

solver/src/settlement.rs
marcovc: Cool, thanks!

PR #434, opened 2021-03-30
[Easy] Log OrderIds of all used order

fleupold:
Adding an opaque id string field to the LimitOrder type so that we can print order identifiers in the driver. This makes it easier to debug instances as it allows easy lookup of the orders involved.

### Test Plan
CI

Running the solver on Rinkeby shows the following output now:

```
2021-03-30T13:05:08.316Z DEBUG solver::driver: starting single run
2021-03-30T13:05:08.676Z DEBUG solver::driver: got 2 orders
2021-03-30T13:05:09.113Z  WARN solver::driver: failed to estimate price for token 0xeeee…eeee: No valid path found between 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee and 0xc778417e063141139fce010982780140aa0cd5ab
2021-03-30T13:05:09.114Z DEBUG solver::driver: pruned 1 orders: ["0x0735c673e3fc4e5b7045538ddb8d86ce4ccb7ed9a5188c16471422ecea7c4f2052df85e9de71aa1c210873bcf37ec46d36c99dc260a43a6a"]
2021-03-30T13:05:09.114Z DEBUG solver::driver: Using orders: ["0xac6411f925653cff65c8fd643cc16a588ca6e27cb1caf1263b24e6ebc618cf5152df85e9de71aa1c210873bcf37ec46d36c99dc260a41388"]
```

### DIFF
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -314,7 +314,7 @@ impl Display for OrderUid {
         let mut bytes = [0u8; 2 + 56 * 2];
         bytes[..2].copy_from_slice(b"0x");
         // Unwrap because the length is always correct.
-        hex::encode_to_slice(self.0, &mut bytes[2..]).unwrap();
+        hex::encode_to_slice(&self.0, &mut bytes[2..]).unwrap();
         // Unwrap because the string is always valid utf8.
         let str = std::str::from_utf8(&bytes).unwrap();
         f.write_str(str)
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -296,7 +296,7 @@ impl Default for OrderMetaData {
 }
 
 // uid as 56 bytes: 32 for orderDigest, 20 for ownerAddress and 4 for validTo
-#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Copy, Eq, Hash, PartialEq)]
 pub struct OrderUid(pub [u8; 56]);
 
 impl FromStr for OrderUid {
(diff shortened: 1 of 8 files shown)


model/src/order.rs
```diff
@@ -309,15 +309,26 @@ impl FromStr for OrderUid {
     }
 }
 
-impl Display for OrderUid {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+impl From<&OrderUid> for String {
+    fn from(uid: &OrderUid) -> Self {
```
vkgnosis, line 313: Docs on `std::string::ToString` say
> This trait is automatically implemented for any type which implements the Display trait. As such, ToString shouldn't be implemented directly: Display should be implemented instead, and you get the ToString implementation for free.

so we should probably do it like that too.

solver/src/liquidity.rs
```diff
@@ -21,6 +21,7 @@ pub enum Liquidity {
 /// Basic limit sell and buy orders
 #[derive(Clone)]
 pub struct LimitOrder {
+    pub id: String,
```
vkgnosis, line 24: Would like a comment to explain what this is. Like "string used to identify order for debugging purposes". Or we could make this OrderUid if that is what it always is.

solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -15,23 +15,24 @@ impl OrderBookApi {
...
-            .map(|order| order.order_creation.into())
+            .map(|order| order.into())
             .collect())
     }
 }
 
-impl From<OrderCreation> for LimitOrder {
-    fn from(order: OrderCreation) -> Self {
+impl From<Order> for LimitOrder {
+    fn from(order: Order) -> Self {
```
vkgnosis, line 24: I'd keep this as OrderCreation and use OrderCreation::uid() to get the id.

bh2smith (approved):

vkgnosis (approved):

model/src/order.rs
fleupold: Good catch!

solver/src/liquidity.rs
fleupold: I prefer using an opaque string here. In the future we might have other liquidity that behaves like a limit order but isn't attached to one of our off-chain orders. Will add the comment

solver/src/liquidity/offchain_orderbook.rs
fleupold: The issue is that `uid()` needs the owner (which is also on metadata). We could recover owner from signature but that would require knowledge of the domain_separator here.

vkgnosis: I forgot it needed the owner. Makes sense to use the full order struct in that case.

model/src/order.rs
```diff
@@ -314,13 +314,19 @@ impl Display for OrderUid {
         let mut bytes = [0u8; 2 + 56 * 2];
         bytes[..2].copy_from_slice(b"0x");
         // Unwrap because the length is always correct.
-        hex::encode_to_slice(&self.0, &mut bytes[2..]).unwrap();
+        hex::encode_to_slice(self.0, &mut bytes[2..]).unwrap();
```
nlordell, line 317: Why this change?

nlordell (approved):

model/src/order.rs
fleupold: This change snuck in when moving the method implementation back and forth. Will remove.

PR #435, opened 2021-03-30
Create Tenderly Simulation on failed solution

### LINKED ISSUE #422: Generate Tenderly simulation on failed settlement simulations
This issue captures the work to generate a Tenderly simulation on failed settlement simulations.

Luckily, its actually quite easy! Tenderly allows configuring a simulation with URL parameters. For example (note escaped newlines with `\` for readability):
```
https://dashboard.tenderly.co/${organization}/${project}/simulator/new?\
block=8306882&\
blockIndex=0&\
from=0x0e1401b7037163764b161fd989bdcf224bb5dcc4&\
gas=8000000&\
gasPrice=0&\
value=0&\
contractAddress=0x4e608b7da83f8e9213f554bdaa77c72e125529d0&\
contractFunction=0x7caf7af0&\
functionInputs=%5B%220x6d7f0754ffeb405d23c51ce938289d4835be3b14%22%2C%220xd9ba894e0097f8cc2bbc9d24d308b98e36dc6d02%22%5D&\
functionInputs=%5B%2262688440296451781%22%2C%222500007499%22%5D&\
functionInputs=0x0100000000000000000000000000000000000000000000000eda9474d6355960d0ef0000000000000000000000000000000000000000000000000000001053ab84d3605ddeeaf6a005bd000000000000000000000000000000000000000000000000a1a1d4559a86e00000000000000000000000000000000000000000000000000eda9474d6355960d0ef00009c2a60b3405de291ca8826ec7e9cd44fdb74c4f8c6324f28801544fa858e2d8c80540fc186c12c4f9d1c1b838e9be39c0038b704acd0838870fca7e5363f8be1d7&\
functionInputs=0x7a250d5630b4cf539739df2c5dacb4c659f2488d00010438ed173900000000000000000000000000000000000000000000000000deb6cde764aec50000000000000000000000000000000000000000000000000000000094dcf0a300000000000000000000000000000000000000000000000000000000000000a00000000000000000000000004e608b7da83f8e9213f554bdaa77c72e125529d0ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000d9ba894e0097f8cc2bbc9d24d308b98e36dc6d020000000000000000000000006d7f0754ffeb405d23c51ce938289d4835be3b14&\
functionInputs=0x&\
network=4
```

Would open a new simulation with the specified parameters for the specified organization or account and project name (note that this requires first "adding" the settlement contract - which can be done by address as its already verified on Etherscan).

So a log message with these populated parameters would be enough! Easy! We just need a Gnosis org and a GPv2 project and this should be fairly trivial to add.

fleupold:
Fixes #422

We should now see a link in our alerts that makes us 1 click away from creating the tenderly simulation for the failed settlement. I believe this will speed up debugging by an order of magnitude

Note, that in order to profit from this I need to add your tenderly account to our organization. Let me know your tenderly account via slack.

### Test Plan
Run rinkeby solver with bad private key. See tenderly link in logs and when following it land on a page where all we need to do is click "Simulate Transaction" (all other values are already filled out).

### DIFF
--- a/solver/src/settlement_submission.rs
+++ b/solver/src/settlement_submission.rs
@@ -5,6 +5,7 @@ use self::retry::{CancelSender, SettlementSender};
 use crate::settlement::Settlement;
 use anyhow::{anyhow, Context, Result};
 use contracts::GPv2Settlement;
+use ethcontract::{dyns::DynTransport, transaction::TransactionBuilder, Web3};
 use gas_estimation::GasPriceEstimating;
 use gas_price_stream::gas_price_stream;
 use primitive_types::{H160, U256};
@@ -92,8 +93,34 @@ async fn simulate_settlement(
     contract: &GPv2Settlement,
 ) -> Result<()> {
     let method = retry::settle_method_builder(contract, settlement.clone());
-    let data = method.tx.data.as_ref().expect("no data").0.as_slice();
-    tracing::info!("Settlement call: {}", hex::encode(data));
-    method.call().await.context("settle simulation failed")?;
-    Ok(())
+    let tx = method.tx.clone();
+    let result = method.call().await;
+    match &result {
+        Ok(_) => Ok(()),
+        Err(_) => {
+            let tenderly_link = tenderly_link(&contract.raw_instance().web3(), tx)
+                .await
+                .unwrap_or_else(|err| format!("Unable to create simulation link due to: {}", err));
+            result
+                .map(|_| ())
+                .context(format!("Settle simulation failed. Link: {}", tenderly_link))
+        }
+    }
+}
+
+// Creates a simulation link in the gp-v2 tenderly workspace
+async fn tenderly_link(
+    web3: &Web3<DynTransport>,
+    tx: TransactionBuilder<DynTransport>,
+) -> Result<String> {
+    let current_block = web3.eth().block_number().await?;
+    let network_id = web3.net().version().await?;
+    Ok(format!(
+        "https://dashboard.tenderly.co/gp-v2/staging/simulator/new?block={}&blockIndex=0&from={:#x}&gas=8000000&gasPrice=0&value=0&contractAddress={:#x}&rawFunctionInput=0x{}&network={}",
+        current_block,
+        tx.from.unwrap().address(),
+        tx.to.unwrap(),
+        hex::encode(tx.data.unwrap().0),
+        network_id
+    ))
 }
--- a/solver/src/settlement_submission.rs
+++ b/solver/src/settlement_submission.rs
@@ -5,7 +5,12 @@ use self::retry::{CancelSender, SettlementSender};
 use crate::settlement::Settlement;
 use anyhow::{anyhow, Context, Result};
 use contracts::GPv2Settlement;
-use ethcontract::{dyns::DynTransport, transaction::TransactionBuilder, Web3};
+use ethcontract::{
+    dyns::DynTransport,
+    errors::{ExecutionError, MethodError},
+    transaction::TransactionBuilder,
+    Web3,
+};
 use gas_estimation::GasPriceEstimating;
 use gas_price_stream::gas_price_stream;
 use primitive_types::{H160, U256};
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):
Nice. This is going to be useful.

* Could pass in network id because it doesn't change so that retrieving it cannot fail.
* Current block has the problem that we don't know at which block the transaction actually failed so we might be setting a block that is later. Since we cannot guarantee correctness we could retrieve it before `simulate_settlement` too so that creating the tenderly link cannot fail.
* Could inspect the error (`MethodError`) and only create tenderly link for `Revert` and `InvalidOpcode` because the others aren't smart contract errors.

fleupold:
> Could pass in network id because it doesn't change so that retrieving it cannot fail.

My feeling was that this would unnecessarily leak a debug only change through a much larger part of the codebase (all the way through settlement and driver logic). Since we only do this extra call when simulation already failed (which should be rare) I thought it was ok to redo it in the method directly (I assume these failures to be rare).

> Current block has the problem that we don't know at which block the transaction actually failed so we might be setting a block that is later

That's a good point. We could do this extra call before every simulation and then pin the simulation to that block number. However, I'm worried that this could cause issues with load balancers and would increase an extra roundtrip for every run. Maybe knowing that the block number in the simulation link is not super accurate and +/- 1 it when debugging is enough though?

> Could inspect the error (MethodError) and only create tenderly link for Revert and InvalidOpcode because the others aren't smart contract errors.

Good point. Will also have to do it for ExecutionError::Failure

nlordell:
> We could do this extra call before every simulation and then pin the simulation to that block number.

Why not do a batch call and run the simulation and get the block being simulated on in the same JSON RPC request?

nlordell (approved):
Very nice! It would be cool if we could do a batch call to get the block for the `eth_call` as well as the result in a single request, but I think that could also be done in a follow up PR.

As a side question, does Tenderly decode the `rawFunctionInput` when the contract was added to the project?

solver/src/settlement_submission.rs
```diff
@@ -92,8 +98,45 @@ async fn simulate_settlement(
...
+
+// Creates a simulation link in the gp-v2 tenderly workspace
+async fn tenderly_link(
+    web3: &Web3<DynTransport>,
+    tx: TransactionBuilder<DynTransport>,
+) -> Result<String> {
+    let current_block = web3.eth().block_number().await?;
+    let network_id = web3.net().version().await?;
+    Ok(format!(
+        "https://dashboard.tenderly.co/gp-v2/staging/simulator/new?block={}&blockIndex=0&from={:#x}&gas=8000000&gasPrice=0&value=0&contractAddress={:#x}&rawFunctionInput=0x{}&network={}",
```
nlordell, line 135: Also tiny nit: the org and project name should likely be configurable (in case others want to run solvers for example). If we pass down these values as configuration for the Tenderly link, then we can also do as @vkgnosis suggested and pass the network ID as well (to avoid a failure for creating the link).

fleupold:
> Why not do a batched Web3 request and run the simulation and get the block being simulated on in the same JSON RPC request?

That's possible however the user friendly batch API is only available for ViewMethods (cannot be mixed with generic eth calls). I feel like writing a manual batch request is going to create a lot of extra code (and I'm not yet sure the very exact block number will often be of interest).

solver/src/settlement_submission.rs
fleupold: This would require making it configurable all the way up to the command line, wouldn't it (I guess we can leave default values, yet it seems intrusive, requiring documentation etc. all for such a non core functionality)? 

Maybe we can add that once we have someone who really wants to use this link but is in a different org?

nlordell: Of course. This wasn't a suggesting doing it in this PR. Just something to keep in mind (maybe just add an issue)?

nlordell:
> I feel like writing a manual batch request is going to create a lot of extra code

I don't imagine it being that bad. Needless to say, I think this PR can be merged as-is, it was more of a suggestion for a future improvement.

vkgnosis:
I'm definitely fine with merging as is. Suggested improvements can be remembered in issue.

bh2smith:
Should we be worried at all about rate limiting with their services when, for example, we have continuously failing batches over a long weekend? Could get expensive!

bh2smith (approved):

solver/src/settlement_submission.rs
bh2smith: I love cleanup!

nlordell:
> Should we be worried at all about rate limiting with their services when

This change only displays the link but doesn't actually run any simulations. Additionally, the link takes you to the "new simulation wizard", so it will only actually run the simulation until you click a button on their page. In that sense running simulations requires human interaction, so I don't think rate limiting would be a concern.

PR #439, opened 2021-03-31
Fix CORS headers not being set on rejections

vkgnosis:
Warp's cors functionality does not apply to rejected requests which is
arguably a bug. The workaround for this is to turn all rejections into
replies so that cors applies. Warp internally already knows how to
convert its rejections like when serializing a json body fails into a
response. In order to reuse that code we use a fork of warp where such a
public method has been exposed.

The issue in warp: https://github.com/seanmonstar/warp/issues/518
PR that allows turning rejections into requests: https://github.com/seanmonstar/warp/pull/827
My fork that uses the same idea: https://github.com/vkgnosis/warp/commit/c329629f1130c190c10fa1e8ec9fb63fc81e45b0

We are using my fork because we are still on an older version of warp (0.2.0) (because rust-web3 does not yet support tokio 1.0).

### Test Plan
I manually `curl`ed a request that fails decoding the url query

```
curl 'http://127.0.0.1:8080/api/v1/fee?sellToken=a&buyToken=a&amount=1&kind=buy'  -H 'Origin: https://protocol-rinkeby.dev.gnosisdev.com' -v
```

and saw that the response now includes the header `access-control-allow-origin: https://protocol-rinkeby.dev.gnosisdev.com`.

### DIFF
--- a/orderbook/src/api.rs
+++ b/orderbook/src/api.rs
@@ -58,6 +58,7 @@ pub fn handle_all_routes(
         .with(cors)
 }
 
+// We turn Rejection into Reply to workaround warp not setting CORS headers on rejections.
 async fn handle_rejection(err: Rejection) -> Result<impl Reply, Infallible> {
     Ok(err.default_response())
 }
--- a/orderbook/src/api.rs
+++ b/orderbook/src/api.rs
@@ -16,7 +16,7 @@ use primitive_types::H160;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use shared::price_estimate::PriceEstimating;
-use std::{str::FromStr, sync::Arc};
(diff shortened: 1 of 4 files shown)


orderbook/src/api.rs
```diff
@@ -38,17 +38,28 @@ pub fn handle_all_routes(
...
+                .or(get_solvable_orders)
+                .or(get_trades)
+                .or(cancel_order)
+                .or(get_amount_estimate),
+        )
+        .recover(handle_rejection)
+        .with(cors)
+}
+
+async fn handle_rejection(err: Rejection) -> Result<impl Reply, Infallible> {
```
fleupold, line 61: Maybe a comment here on why we add this method would be useful?

fleupold (approved):
Nice find and good fix!

PR #440, opened 2021-03-31
[Easy] Listing required fields on object types

### LINKED ISSUE #437: [Swagger] Make required fields explicitly required
After reading the docs about making certain object properties optional, it appears all properties are optional by default. This means we should explicitly make so we will have to make all the other fields required instead:

Example: https://swagger.io/docs/specification/data-models/data-types/#required

bh2smith:
Closes #437 and based/depends on #391

Because 

> By default, all object properties are optional. You can specify the required properties in the required list

We have to explicitly declare which fields are required as opposed to which are nullable.

### Test Plan
Open up the rendered version of the open-api spec doc and see that there is a red asterisk next to the fields which are now explicitly required.

Before:
Screenshot 2021-03-31 at 12 12 15

After:
Screenshot 2021-03-31 at 12 11 03


vkgnosis:
Reading through the openapi documentation again, I believe for json objects `nullable` and `required` are different things. The former means a key-value can be set to `null` while the latter means the whole key value must be exist.
The way we have implemented the api everything is always required but also some fields can be nullable. So I like the change of adding required but we shouldn't remove `nullable`.

vkgnosis (approved):

orderbook/openapi.yml
```diff
@@ -471,6 +502,16 @@ components:
...
+      required:
+        - blockNumber
+        - logIndex
+        - orderUid
+        - owner
+        - sellToken
+        - buyToken
+        - sellAmount
+        - sellAmountBeforeFees
+        - buyAmount
```
fleupold, line 514: If there is no transaction hash do we return null or do we omit to serialized the field? In the former case the field would still be required, but nullable.

fleupold (approved):

orderbook/openapi.yml
bh2smith: Interesting fact. We return null. Will include this then. Are there any other fields that fit this description I am missing?

bh2smith, line 514:
```suggestion
        - buyAmount
        - transactionHash
```

Issue #414, opened 2021-03-24
[API] allow optional from address parameter on order creation


fleupold:
This would make it easier to debug issues with signature recovery. At the moment, if we provide a slightly incorrect signature (e.g. MSB of `v` not set despite using message encoding) we will recover a valid but random address that has no balance or allowance.

Likely this order is not useful and it makes it hard for the user of the API to realize they are doing something wrong.

An optional fromAddress field on the OrderCreation struct could allow an assertion that the signature resolves to the expected address and otherwise lead to a rejection of the order.

PR #441, opened 2021-03-31
Enforce minimum balance on order creation

### LINKED ISSUE #248: Reject new orders when they don't have available balance
due to not giving allowance or not having funds. This is to protect database against spam.
Can be implemented after Felix's Order Balance Fetcher PR.

vkgnosis:
This helps protect against user error and adds some ddos protection. We
use the InsufficientFee error that we already planned for.

Fixes #248 

### Test Plan
e2e test shows that adding orders still works

It would be nice to test rejection in an e2e test too (and probably more smaller tests) but for that we need to refactor the e2e test a bit first to make it easier to setup the components.

### DIFF
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -87,9 +87,11 @@ impl Orderbook {
         if matches!(payload.from, Some(from) if from != order.order_meta_data.owner) {
             return Ok(AddOrderResult::WrongOwner(order.order_meta_data.owner));
         }
-        self.balance_fetcher
-            .register(order.order_meta_data.owner, order.order_creation.sell_token)
-            .await;
+
+        if !new_order_has_balance(&order, self.balance_fetcher.as_ref()).await {
+            return Ok(AddOrderResult::InsufficientFunds);
+        }
+
         match self.database.insert_order(&order).await {
             Ok(()) => Ok(AddOrderResult::Added(order.order_meta_data.uid)),
             Err(InsertionError::DuplicatedRecord) => Ok(AddOrderResult::DuplicatedOrder),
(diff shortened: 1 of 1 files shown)


orderbook/src/orderbook.rs
```diff
@@ -82,9 +82,21 @@ impl Orderbook {
...
+        let minimum_balance = if order.order_creation.partially_fillable {
+            1.into()
+        } else {
+            order
+                .order_creation
+                .sell_amount
+                .saturating_add(order.order_creation.fee_amount)
+        };
+        let balance = self.balance_fetcher.get_balance(owner, token);
+        if !matches!(balance, Some(balance) if balance >= minimum_balance) {
```
fleupold, line 97: With this match statement, it's not clear to me what happens when balance is None.

fleupold (approved):
Maybe we can move the order validation into its own method and start adding some unit tests for it? We are adding at least 3 branches here (partial order, nonpartial order, unable to fetch balance) which could be documented by tests.

orderbook/src/orderbook.rs
nlordell: I read this as "if `balance` doesn't match `Some` balance that is bigger than the minimum". So a `None` balance would enter the if block (and return `InsufficientFunds` result).

nlordell (approved):

orderbook/src/orderbook.rs
```diff
@@ -82,9 +82,21 @@ impl Orderbook {
             Some(order) => order,
             None => return Ok(AddOrderResult::InvalidSignature),
         };
-        self.balance_fetcher
-            .register(order.order_meta_data.owner, order.order_creation.sell_token)
-            .await;
+        let owner = order.order_meta_data.owner;
```
bh2smith, line 85: If we already have `owner` in the order metadata then I think this could be used in place #414 to detect a `WrongOwner` error on order creation (i.e. instead of introducing Optional from address). I have already started a PR for this #449 but we might be able to make use of this instead.

fleupold (approved):
Thanks for adding the tests!

PR #446, opened 2021-04-01
[Trivial] Omit transport debug logs from default filter

fleupold:
Otherwise running the api/solver locally becomes too noisy

In combination with explicitly enabling it in staging: https://gitlab.gnosisdev.com/devops/gnosis-staging/-/merge_requests/422

### Test Plan
Run it and see no request logs.


bh2smith:
I found the order book was also quite noisy with debug logs for this reason too.

bh2smith (approved):

vkgnosis (approved):

PR #447, opened 2021-04-01
[OrderPlacement API] make receiver field optional

fleupold:
Slight usability improvement for people using the API. The field is already nullable so it might as well be completely optional

### Test Plan
Before the following request:

> curl 'http://localhost:8080/api/v1/orders' \
                                                 --data-raw '{
                                                 "sellToken": "0x6810e776880c02933d47db1b9fc05908e5386b96",
                                                 "sellAmount": "1234567890",
                                                 "buyAmount": "1234567890",
                                                 "validTo": 2617291477,
                                                 "feeAmount": "12345678900000000",
                                                 "kind": "buy",
                                                 "partiallyFillable": true,
                                                 "signature": "0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
                                                 "signingScheme": "eip712"
                                               }' --compressed

Would return

> Request body deserialize error: missing field `receiver` at line 13 column 1

With this change it complains about a the fee (which happens after parsing the order).

### DIFF
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -356,7 +356,6 @@ components:
             owner (i.e. the order signer).
           $ref: "#/components/schemas/Address"
           nullable: true
-          required: false
         sellAmount:
           description: "Amount of sellToken to be sold in atoms"
           $ref: "#/components/schemas/TokenAmount"
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -349,6 +349,7 @@ components:
             owner (i.e. the order signer).
           $ref: "#/components/schemas/Address"
           nullable: true
+          required: false
         sellAmount:
           description: "Amount of sellToken to be sold in atoms"
           $ref: "#/components/schemas/TokenAmount"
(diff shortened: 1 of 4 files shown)


orderbook/openapi.yml
```diff
@@ -349,6 +349,7 @@ components:
             owner (i.e. the order signer).
           $ref: "#/components/schemas/Address"
           nullable: true
+          required: false
```
bh2smith, line 352: Didn't we just learn that all parameters of an object are optional by default? We shouldn't have to make it explicitly not required....

bh2smith: If this is the case then I slightly prefer these semantics rather than having a list of all required fields below.

bh2smith (approved):

bh2smith:
I believe this is related to #440.

orderbook/openapi.yml
bh2smith: I think the error may be caused by the fact that `required` is meant to be at the end of the list and usually indented one less. At least according to the errors I am seeing in the build this is what I had on this failing build in a different PR

https://github.com/gnosis/gp-v2-services/runs/2235945020

vkgnosis (approved):

orderbook/openapi.yml
fleupold: makes sense. Removed it...

PR #448, opened 2021-04-01
[Easy] Make one unit test more concise

bh2smith:
This PR fixes and easy todo that was missed when merging the settlements/transaction hash PR #334 where we verify that no trades are returned if there are no orders.

By introducing a drop_table method, we can simply drop the orders table at the end of a more relevant test and make one last assertion when the orders table is truncated.

This PR also starts to make me think a bit more about #438, which I will try and tackle next. Especially since it is incredibly unprofessional to be passing a string into this public `drop_table` method. We should surely be passing an enum here and only allow very specific strings to be passed into this method.

### Test Plan
Existing CI

### DIFF
--- a/orderbook/src/database.rs
+++ b/orderbook/src/database.rs
@@ -44,7 +44,7 @@ impl Database {
     /// Delete all data in the database. Only used by tests.
     pub async fn clear(&self) -> Result<()> {
         use sqlx::Executor;
-        self.pool.execute(sqlx::query("TRUNCATE orders;")).await?;
+        self.drop_orders().await?;
         self.pool.execute(sqlx::query("TRUNCATE trades;")).await?;
         self.pool
             .execute(sqlx::query("TRUNCATE invalidations;"))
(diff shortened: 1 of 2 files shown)


orderbook/src/database.rs
```diff
@@ -57,4 +57,10 @@ impl Database {
             .await?;
         Ok(())
     }
+
+    pub async fn drop_table(&self, table_name: &str) -> Result<()> {
+        use sqlx::Executor;
+        self.pool.execute(sqlx::query(format!("TRUNCATE {};", table_name).as_str())).await?;
```
bh2smith, line 63: ⚠️ - This is super dangerous! For example I could pass in the string

```
table_name = "orders; TRUNCATE trades" and we would be losing both tables. Of course there are probably much worse things that could be done.

Perhaps for now I should just make a `drop_orders` method... at least to be safe for this PR.

fleupold (commented):
I'm not a huge fan of this change, as it's changing the story of the test case from "If there is a trade event processed, but no order event, things work as expected" to "If there was an order event and a trade event, but then we drop the order table, everything works".

It's not obvious those are testing the same thing. The order processing logic might e.g. write to other tables (which we don't drop) and that's why things are working.

I personally preferred the old test as it was more like what we will be seeing in practice (a trade event without an order event being created on a clean state).

PR #449, opened 2021-04-06
[API] Introduction of optional "from" in Order Creation

### LINKED ISSUE #414: [API] allow optional from address parameter on order creation

bh2smith:
Fixes #414

Introducing from address as optional parameter for order creation to make it easier to debug when signatures are not valid. This check would not insert orders whose recovered owner that don't match the provided from address.

We also introduce `WrongOwner(address)` error as a response to order creation so to separate this from `InvalidSignature`.

### Test Plan
Existing CLI should suffice (especially the e2e test with ganache demonstrates that the changes are backwards compatible).

### DIFF
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -144,6 +144,7 @@ pub struct OrderCreation {
     pub sell_token: H160,
     #[serde(with = "h160_hexadecimal")]
     pub buy_token: H160,
+    #[serde(default)]
     #[serde_as(as = "Option<HexadecimalH160>")]
     pub receiver: Option<H160>,
     #[serde(with = "u256_decimal")]
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -68,9 +68,9 @@ impl Order {
                 if from == owner {
                     return Some(construct_order(owner, order_creation, domain));
                 }
-                return None;
+                None
             }
-            None => Some(construct_order(owner, order_creation, domain)),
+            None => Some(construct_order(owner, order_creation, domain))
         }
     }
 }
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -169,15 +169,6 @@ pub struct OrderCreationPayload {
     pub from: Option<H160>,
 }
 
-impl From<OrderCreation> for OrderCreationPayload {
-    fn from(order_creation: OrderCreation) -> Self {
-        Self {
-            order_creation,
-            from: None,
-        }
-    }
-}
-
 impl Default for OrderCreation {
     // Custom implementation to make sure the default order is valid
     fn default() -> Self {
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -164,6 +164,7 @@ pub struct OrderCreation {
 
 #[derive(Clone, Copy, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
 pub struct OrderCreationPayload {
+    #[serde(flatten)]
     pub order_creation: OrderCreation,
     pub from: Option<H160>,
 }
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -168,6 +168,15 @@ pub struct OrderCreationPayload {
     pub from: Option<H160>,
 }
 
+impl From<OrderCreation> for OrderCreationPayload {
+    fn from(order_creation: OrderCreation) -> Self {
+        Self {
+            order_creation,
+            from: None,
+        }
+    }
+}
+
 impl Default for OrderCreation {
     // Custom implementation to make sure the default order is valid
     fn default() -> Self {
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -41,20 +41,37 @@ impl Order {
         order_creation: OrderCreation,
         domain: &DomainSeparator,
     ) -> Option<Self> {
+        fn construct_order(
+            owner: H160,
+            order_creation: OrderCreation,
+            domain: &DomainSeparator,
+        ) -> Order {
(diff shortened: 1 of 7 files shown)


model/src/order.rs
```diff
@@ -159,6 +176,7 @@ pub struct OrderCreation {
     pub partially_fillable: bool,
     pub signature: Signature,
     pub signing_scheme: SigningScheme,
+    pub from: Option<H160>,
```
fleupold, line 179: I don't like that we will carry this field around to all the callsites that should really be using OrderMetadata.owner (from will be sometimes set which might make people believe they could be accessing it).

Maybe we can introduce a new type

```rust
struct OrderCreationPayload {
  #[serde(flatten)]
  pub order_creation: OrderCreation,
  pub from: Option<H160>,
}
```

This way the API would be like currently (flat with an optional from field) but all APIs that return orders would still only return OrderCreation without the From field (which can be read from Metadata)

fleupold (commented):
I think this is a good change, however I'd like to avoid carrying this information downstream to all the other call sites using order creation data.

Alternatively we could make it non-optional and remove it from the metadata (however, this would be a breaking change).

model/src/order.rs
bh2smith: I fixed this up according to your suggestion. Looks a lot better now and require less overall changes.

orderbook/openapi.yml
```diff
@@ -40,7 +40,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/OrderCreation"
+              $ref: "#/components/schemas/OrderCreationPayload"
```
bh2smith, line 43: Isn't this also a breaking change @fleupold ?

bh2smith: All good now, I have reverted this to make the changes backwards compatible.

orderbook/src/orderbook.rs
```diff
@@ -79,7 +81,15 @@ impl Orderbook {
             return Ok(AddOrderResult::InsufficientFee);
         }
         let order = match Order::from_order_creation(order, &self.domain_separator) {
-            Some(order) => order,
```
vkgnosis, line 82: I'd like to keep the match on from_order_creation and payload separate because it's easier to read:
This can stay as before:

```rust
        let order = match Order::from_order_creation(order, &self.domain_separator) {
            Some(order) => order,
            None => return Ok(AddOrderResult::InvalidSignature),
        };
```
and then we add either
```rust
        if matches!(payload.from, Some(from) if from != order.order_meta_data.owner) {
            return Ok(AddOrderResult::WrongOwner(order.order_meta_data.owner));
        }
```
or
```rust
        if let Some(from) = payload.from {
            if from != order.order_meta_data.owner {
                return Ok(AddOrderResult::WrongOwner(order.order_meta_data.owner));
            }
        }
```
Not sure which one you find clearer.

vkgnosis (approved):

orderbook/src/orderbook.rs
bh2smith: Cool! Thanks.

orderbook/openapi.yml
```diff
@@ -384,6 +384,8 @@ components:
           $ref: "#/components/schemas/Signature"
         signingScheme:
           $ref: "#/components/schemas/SigningScheme"
+        from:
+          $ref: "#/components/schemas/Address"
```
fleupold, line 388: nanonit: maybe add nullable (to me this makes it more obvious that the value can be omitted)?

orderbook/src/orderbook.rs
```diff
@@ -82,6 +84,9 @@ impl Orderbook {
             Some(order) => order,
             None => return Ok(AddOrderResult::InvalidSignature),
         };
+        if matches!(payload.from, Some(from) if from != order.order_meta_data.owner) {
```
fleupold, line 87: I guess I'm the only one who has trouble reading these `matches!` statements with `if` blocks inside. Guess, I will learn to cope with them 👨‍🎓

model/src/order.rs
```diff
@@ -162,6 +162,22 @@ pub struct OrderCreation {
...
 }
 
+#[derive(Clone, Copy, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
+pub struct OrderCreationPayload {
+    #[serde(flatten)]
+    pub order_creation: OrderCreation,
+    pub from: Option<H160>,
+}
+
+impl From<OrderCreation> for OrderCreationPayload {
```
fleupold, line 172: Why is this needed? I don't really see the use case for upcasting an order creation to OrderCreationPayload. If this is for the e2e test we should probably provide the `from` address, as this is the first part where this feature will become useful (remember the time you were trying to debug the e2e test because of some signature glitch).

fleupold (approved):
Looks great!

All backwards compatible and not very intrusive but will be super helpful for debugging 👏

model/src/order.rs
bh2smith: Sorry this was left behind from the change in my last commit before backwards compatibility. Will remove.

PR #451, opened 2021-04-06
Token Deny List

### LINKED ISSUE #450: Token-Deny List as runtime argument
Some tokens behave badly and don't work with our infrastructure even though they might be tradable in AMM pools. Such tokens are currently blocking our services over extended periods of time while orders containing misbehaving tokens (as buy or sell token) are valid and being passed onto solvers. 

This issue proposes a token deny list which is passed into the services at runtime and will be handled in the following parts of the code
- OrderCreation -> Deny creation of orders containing denied tokens
- fee estimation -> Skip fee estimation for denied tokens
- Exclude orders containing denied tokens in the solvable_orders filter (so not to pass these orders onto the solver).

@fleupold - may be able to provide some additional context along token examples.

bh2smith:
Fixes #450

This PR introduces a token deny list which is passed as a comma separated list of Ethereum addresses to the `token_filter` argument. The deny list is handled in three separate parts of the service namely

1. `UniswapPriceEstimator` where these tokens are ignored for both price and gas estimations
2. `Order Creation` when calling `add_order`, the response has been updated with two `AddOrderResponse` errors (`BuyTokenDenied` and `SellTokenDenied`) each returning the token address with the error message
3. `get_solvable_orders` here we introduce a new boolean flag `exclude_deny_tokens` into the order filter which is set to `true` when fetching solvable orders.

Furthermore, we assert that the `deny_list` doesn't overlap with the `base_tokens` list on startup.

### Test Plan

There are three different callsites where this change should be tested. All of which are verified with the following instructions:

1. Run the orderbook on rinkeby with `--token-filter 0xd0dab4e640d95e9e8a47545598c33e31bdb53c7c` (this is rGNO)
Run the web interface locally 

2. (Testing Price Estimator) In the interface select GNO as the sell token

Find this warning on price estimation in the services logs

```
WARN orderbook::fee: Failed to estimate sell token price: Sell Token 0xd0da…3c7c Denied!
```
Alternatively, you can make these requests to the API:
```sh
curl -X GET "http://localhost:8080/api/v1/fee?sellToken=0xd0dab4e640d95e9e8a47545598c33e31bdb53c7c&buyToken=0x4DBCdF9B62e891a7cec5A2568C3F4FAF9E8Abe2b&amount=1234567890&kind=sell" -H  "accept: application/json"
```

```sh
curl -X GET "http://localhost:8080/api/v1/fee?sellToken0x4DBCdF9B62e891a7cec5A2568C3F4FAF9E8Abe2b=&buyToken=0xd0dab4e640d95e9e8a47545598c33e31bdb53c7c&amount=1234567890&kind=sell" -H  "accept: application/json"
```

which will both respond with

```
{"errorType":"NotFound","description":"Fee path for given Token pair was not found"}%
```

3. (Testing Token Denied on order placement) Now select `GNO` as the buy token and choose something else for the sell token
When attempting to place the order find the following Response with code 400
```
{"errorType":"TokenDenied","description":"Buy Token 0xd0da…3c7c"}
```

4. (Testing OrderFilter) Now place an order selling USDC for DAI (or whatever isn't currently denied), but record the address of one (or both) of the tokens. We will use `DAI-0x5592EC0cfb4dbc12D3aB100b257153436a1f0FEa`, `USDC-0x4DBCdF9B62e891a7cec5A2568C3F4FAF9E8Abe2b`

Once this order is placed, restart the orderbook services and add one (or both) of these tokens to the token filter
```sh
cargo run --bin orderbook -- --node-url https://rinkeby.infura.io/v3/c8a386aaa40349eda89325818dbc40e9 --token-filter 0x5592EC0cfb4dbc12D3aB100b257153436a1f0FEa
```

Run:
```sh
curl -X GET "http://localhost:8080/api/v1/orders?includeFullyExecuted=false&includeInvalidated=false&includeInsufficientBalance=false&includeDeniedTokens=false" -H  "accept: application/json"
```

to get nothing returned and 

```sh
curl -X GET "http://localhost:8080/api/v1/orders?includeFullyExecuted=false&includeInvalidated=false&includeInsufficientBalance=false&includeDeniedTokens=true" -H  "accept: application/json"
```

to see the order you had placed.

Note that - We haven't managed to test a Sell Token Denied (can't get sell token denied through the interface because the failed fee estimate prevents us from placing orders this way).

### DIFF
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -14,6 +14,7 @@ use secp256k1::key::ONE_KEY;
 use serde::{de, Deserialize, Serialize};
 use serde::{Deserializer, Serializer};
 use serde_with::serde_as;
+use std::collections::HashSet;
 use std::fmt::{self, Display};
 use std::str::FromStr;
 use web3::signing::{self, Key, SecretKeyRef};
@@ -55,6 +56,10 @@ impl Order {
             },
             order_creation,
         })
+    }
+    pub fn contains_token_from(&self, token_list: &HashSet<H160>) -> bool {
+        token_list.contains(&self.order_creation.buy_token)
+            || token_list.contains(&self.order_creation.sell_token)
     }
 }
 
@@ -421,6 +426,7 @@ mod tests {
     use super::*;
     use chrono::NaiveDateTime;
     use hex_literal::hex;
+    use maplit::hashset;
     use primitive_types::H256;
     use secp256k1::{PublicKey, Secp256k1, SecretKey};
     use serde_json::json;
@@ -579,6 +585,24 @@ mod tests {
         let expected = "0x01000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000ff";
         assert_eq!(uid.to_string(), expected);
         assert_eq!(format!("{}", uid), expected);
+    }
+
+    #[test]
+    fn order_contains_token_from() {
+        let order = Order::default();
+        assert_eq!(
+            order.contains_token_from(&hashset!(order.order_creation.sell_token)),
+            true
+        );
+        assert_eq!(
+            order.contains_token_from(&hashset!(order.order_creation.buy_token)),
+            true
+        );
+        assert_eq!(order.contains_token_from(&HashSet::new()), false);
+        let other_token = H160::from_low_u64_be(1);
+        assert_ne!(other_token, order.order_creation.sell_token);
+        assert_ne!(other_token, order.order_creation.buy_token);
+        assert_eq!(order.contains_token_from(&hashset!(other_token)), false);
     }
 
     pub fn h160_from_public_key(key: PublicKey) -> H160 {
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -87,9 +87,9 @@ paths:
           schema:
             type: boolean
             default: false
-        - name: includeDeniedTokens
+        - name: includeUnsupportedTokens
           in: query
-          description: Should the orders containing denied tokens be included?
+          description: Should the orders containing unsupported tokens be included?
           schema:
             type: boolean
             default: false
@@ -554,8 +554,7 @@ components:
               PastValidTo,
               InsufficientFunds,
               InsufficientFee,
-              BuyTokenDenied,
-              SellTokenDenied,
+              UnsupportedToken,
               WrongOwner,
             ]
         description:
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -539,6 +539,8 @@ components:
               PastValidTo,
               InsufficientFunds,
               InsufficientFee,
+              BuyTokenDenied,
+              SellTokenDenied,
             ]
         description:
           type: string
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -81,6 +81,12 @@ paths:
           schema:
             type: boolean
             default: false
+        - name: includeDeniedTokens
+          in: query
+          description: Should the orders containing denied tokens be included?
+          schema:
+            type: boolean
+            default: false
         - name: minValidTo
           in: query
           description: |
--- a/orderbook/src/database/trades.rs
+++ b/orderbook/src/database/trades.rs
@@ -410,9 +410,6 @@ mod tests {
         )
         .await;
         assert_trades(&db, &TradeFilter::default(), &[trade_a, trade_b]).await;
-
-        // TODO - drop orders and make last assertion again.
-        // Then remove test postgres_trades_with_same_settlement_no_orders
     }
 
     #[tokio::test]
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -123,9 +123,17 @@ async fn main() {
     .await
     .expect("failed to create gas price estimator");
 
+    let deny_tokens = HashSet::from_iter(args.shared.token_filter);
     let mut base_tokens = HashSet::from_iter(args.shared.base_tokens);
     // We should always use the native token as a base token.
     base_tokens.insert(native_token.address());
+    assert!(
+        deny_tokens
+            .intersection(&base_tokens)
+            .collect::<HashSet<_>>()
+            .is_empty(),
+        "Overlapping base and deny tokens not allowed!"
+    );
 
     let current_block_stream = current_block_stream(web3.clone()).await.unwrap();
     let pool_fetcher = CachedPoolFetcher::new(
@@ -139,6 +147,7 @@ async fn main() {
     let price_estimator = Arc::new(UniswapPriceEstimator::new(
         Box::new(pool_fetcher),
         base_tokens,
+        deny_tokens.clone(),
     ));
     let fee_calculator = Arc::new(MinFeeCalculator::new(
         price_estimator.clone(),
@@ -154,6 +163,7 @@ async fn main() {
         event_updater,
         Box::new(balance_fetcher),
         fee_calculator.clone(),
+        deny_tokens,
     ));
     check_database_connection(orderbook.as_ref()).await;
 
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -147,7 +147,6 @@ async fn main() {
     let price_estimator = Arc::new(UniswapPriceEstimator::new(
         Box::new(pool_fetcher),
         base_tokens,
-        deny_tokens.clone(),
     ));
     let fee_calculator = Arc::new(MinFeeCalculator::new(
         price_estimator.clone(),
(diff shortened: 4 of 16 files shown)


model/src/order.rs
```diff
@@ -56,6 +58,18 @@ impl Order {
...
+        token_list
+            .intersection(&hashset!(
+                self.order_creation.buy_token,
+                self.order_creation.sell_token
+            ))
+            .collect::<HashSet<_>>()
+            .is_empty()
+        // The above evaluation is more "symmetric" while the commented one is more precise
+        // token_list.contains(&self.order_creation.buy_token)
+        //     || token_list.contains(&self.order_creation.sell_token)
```
bh2smith, line 71: While the intersection is slightly nicer, I might actually opt for the commented version of this condition so not to introduce the `maplit` dependency here.

orderbook/src/main.rs
```diff
@@ -123,6 +123,8 @@ async fn main() {
     .await
     .expect("failed to create gas price estimator");
 
+    // TODO - should we assert that token_filter doesn't overlap with base_tokens?
+    let deny_tokens = HashSet::from_iter(args.shared.token_filter);
```
bh2smith, line 127: It is slightly unclear how the system should respond to denied base tokens given that they should be ignored, but are also being used in price estimation... I would think that these should not overlap, or that one of the two should trump the other. Looking for suggestions.

shared/src/price_estimate.rs
```diff
@@ -98,6 +114,10 @@ impl PriceEstimating for UniswapPriceEstimator {
         amount: U256,
         kind: OrderKind,
     ) -> Result<BigRational> {
+        match self.token_validation(&buy_token, &sell_token) {
+            Ok(_) => (),
+            Err(validation_err) => return Err(validation_err),
+        }
```
bh2smith, line 120: I am not terribly impressed with the lack of savings obtained with the `token_validation` method given that I needed a match statement to parse this. I am sure there is a cleaner way to express the same logic here, but I am not sure what it is.

model/src/order.rs
vkgnosis: I like the commented version because no temporary set is created. Could also write it as
```rust
        [self.order_creation.buy_token, self.order_creation.sell_token]
        .iter()
        .any(|token| token_list.contains(token))
```

orderbook/src/main.rs
vkgnosis: I like asserting that they don't overlap.

orderbook/src/orderbook.rs
```diff
@@ -57,17 +60,25 @@ impl Orderbook {
...
             balance_fetcher,
             fee_validator,
+            deny_tokens,
         }
     }
 
     pub async fn add_order(&self, order: OrderCreation) -> Result<AddOrderResult> {
+        if self.deny_tokens.contains(&order.buy_token) {
+            return Ok(AddOrderResult::BuyTokenDenied(order.buy_token));
+        }
```
vkgnosis, line 78: If you put `contains_token_from` from before on `OrderCreation` instead of `Order` you can reuse it here but I guess then we can't differentiate between sell token being the error or buy token. I'd also be fine with having only one error instead of two. No need to change anything, just mentioning.

shared/src/price_estimate.rs
vkgnosis, line 120:
```suggestion
        self.token_validation(&buy_token, &sell_token)?;
```

shared/src/price_estimate.rs
```diff
@@ -141,6 +161,10 @@ impl PriceEstimating for UniswapPriceEstimator {
         amount: U256,
         kind: OrderKind,
     ) -> Result<U256> {
+        match self.token_validation(&buy_token, &sell_token) {
```
vkgnosis, line 164: here too

shared/src/price_estimate.rs
```diff
@@ -403,12 +430,42 @@ mod tests {
...
+            .await
+            .is_err());
+
+        // // TODO - make more concise assertion about which error we are getting.
+        // let expected = anyhow!("Buy Token {} Denied!", token_a);
+        // let result = estimator
+        //     .estimate_price(token_b, token_a, 1.into(), OrderKind::Buy)
+        //     .await.unwrap_err();
+        // // Unfortunately we can't compare anyhow errors.
+        // assert_eq!(expected, result);
```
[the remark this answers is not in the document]
vkgnosis: The correct way to do it is to use an error enum instead of anyhow. https://docs.rs/thiserror/1.0.24/thiserror/ is nice for that. `anyhow` is meant to be used precisely when you do not need to distinguish errors which is the case for most of our code base.
On the other hand since the distinction is only useful for the test I'm fine with keeping it an anyhow::Error. You could inspect the error string directly through `err.to_string()`.

model/src/order.rs
bh2smith: Cool, I was gonna switch to this anyway! Your suggestion is intriguing, but maybe more cryptic than it needs to be. Lets see what others think and I'll just use the commented code for now.

orderbook/src/orderbook.rs
bh2smith: I like the idea of reuse, but I think its nice to tell the user exactly why there order was rejected if we can. Open to influence if anyone has strong opinion, but would otherwise leave it as is.

orderbook/src/database/trades.rs
```diff
@@ -410,9 +410,6 @@ mod tests {
         )
         .await;
         assert_trades(&db, &TradeFilter::default(), &[trade_a, trade_b]).await;
-
-        // TODO - drop orders and make last assertion again.
-        // Then remove test postgres_trades_with_same_settlement_no_orders
```
bh2smith, line 415: Related to #448 - which was closed as we are not going to do this.

fleupold (changes requested):
I really like the description (especially the detailed test cases) 🎉 


I think this change could be more minimal, if we give the pool fetcher the deny list and just ignore "fetch request" for pools that contain a deny listed token. This way we won't find any Uniswap pools for this token and as such should fail fee estimation and price estimation.

The error would be less descriptive, as it would be a generic "Not Found", but I think this is fine (it's a rare case for someone to trade this token). Also I'm generally about worried about explicit "deny" error message as they suggest centralisation.

Requesting changes so we can discuss this proposal...

vkgnosis (approved):

orderbook/openapi.yml
```diff
@@ -87,6 +87,12 @@ paths:
           schema:
             type: boolean
             default: false
+        - name: includeDeniedTokens
```
fleupold, line 90: Sorry for being nitpicky here on the naming. Can we call it UnsupportedTokenList? This way it seems more like a technical limitation (which it is) rather than something we do arbitrarily.

orderbook/openapi.yml
```diff
@@ -539,6 +545,8 @@ components:
               PastValidTo,
               InsufficientFunds,
               InsufficientFee,
+              BuyTokenDenied,
```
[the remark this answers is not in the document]
fleupold: Needs updating

orderbook/src/orderbook.rs
```diff
@@ -33,6 +33,8 @@ pub enum AddOrderResult {
     PastValidTo,
     InsufficientFunds,
     InsufficientFee,
+    BuyTokenDenied(H160),
```
fleupold, line 36: nit: we likely don't need the differentiation here neither.

shared/src/arguments.rs
```diff
@@ -46,6 +46,10 @@ pub struct Arguments {
     /// Should be the most liquid tokens of the given network.
     #[structopt(long, env = "BASE_TOKENS", use_delimiter = true)]
     pub base_tokens: Vec<H160>,
+
+    /// List of token addresses to be ignored throughout service
+    #[structopt(long, env = "TOKEN_FILTER", use_delimiter = true)]
```
fleupold, line 51: here also I'd go for UNSUPPORTED_TOKENS

shared/src/uniswap_pool.rs
```diff
@@ -29,6 +29,31 @@ pub trait PoolFetching: Send + Sync {
     async fn fetch(&self, token_pairs: HashSet<TokenPair>) -> Vec<Pool>;
 }
 
+pub struct FilteredPoolFetcher {
```
fleupold, line 32: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia4.giphy.com%2Fmedia%2Fl3q2slcE8854Yqqg8%2Fgiphy.gif&token=noice" data-gitmeme-token="noice"></a>

fleupold (approved):
Looks good. Some minor cleanups and a naming request: Can we try to avoid the "deny" name and rather call it unsupported tokens? 

I feel that this is closer to why we filter them and doesn't have as much of a censoring connotation.

Issue #425, opened 2021-03-29
Prometheus Metrics


vkgnosis:
We don't have any metrics right now and there a couple of easy targets like number of orders in the system, api requests.
Let's use this issue to keep track of progress on the metrics front and plan which ones we want to implement.

PR #454, opened 2021-04-07
Collecting metrics for our API requests

fleupold:
Part of #425 

This PR wraps our warp routing system with a function to log all incoming request/responses to prometheus. This single metric will allow us to get a bunch of interesting insights
- Number of orders created per time interval
- Failure Rate (based on status code) per route
- Execution time (histogram information)

I'm currently still struggling on how to do the labelling of the requests (see inline discussion).

### Test Plan
Run orderbook, query a few routes. Then query `http://localhost:9586/metrics` and see collected metrics.

### DIFF
--- a/orderbook/src/metrics.rs
+++ b/orderbook/src/metrics.rs
@@ -1,0 +1,48 @@
+use std::{convert::Infallible, sync::Arc, time::Instant};
+
+use prometheus::{HistogramOpts, HistogramVec, Registry};
+use warp::Filter;
+use warp::{reply::Response, Reply};
+
+pub struct Metrics {
+    requests: HistogramVec,
+}
+
+impl Metrics {
+    pub fn new(registry: &Registry) -> Self {
+        let opts = HistogramOpts::new(
+            "gp_v2_api_requests",
+            "API Request durations labelled by route and response status code",
+        );
+        let requests = HistogramVec::new(opts, &["response", "request_type"]).unwrap();
+        registry
+            .register(Box::new(requests.clone()))
+            .expect("Failed to register metric");
+        Self { requests }
+    }
+}
+
+// Response wrapper needed because we cannot inspect the reply's status code without consuming it
+struct MetricsReply {
+    response: Response,
+}
+
+impl Reply for MetricsReply {
+    fn into_response(self) -> Response {
+        self.response
+    }
+}
+
+pub fn start_request() -> impl Filter<Extract = (Instant,), Error = Infallible> + Clone {
+    warp::any().map(Instant::now)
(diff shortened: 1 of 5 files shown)


orderbook/src/metrics.rs
```diff
@@ -0,0 +1,48 @@
...
+pub fn start_request() -> impl Filter<Extract = (Instant,), Error = Infallible> + Clone {
+    warp::any().map(Instant::now)
+}
+
+pub fn end_request(metrics: Arc<Metrics>, timer: Instant, reply: impl Reply) -> impl Reply {
+    let response = reply.into_response();
+    let elapsed = timer.elapsed().as_secs_f64();
+    metrics
+        .requests
+        .with_label_values(&[response.status().as_str(), "TODO"])
```
fleupold, line 45: I would like to record which route was requested so that we can slice our data by request type. The problem with using the full path here, is that labels are supposed to be finite. We have however an infinite number of paths (e.g. `/orders/<uid>`) so using the full path is likely not a good idea.

Ideally I'd like to get the name of the "handler" as defined in api.rs (e.g. create_orders, legacy_fee_info, get orders, etc). However, this doesn't seem easily possible.

I thought about adding the label as a header into the response and then extracting it in `end_request` (might leak it to the consumer, although we could also remove it since we inspect the response already anyways)/

Another option would be to do a translation from "path" to label. However, I didn't yet find a good way of colocating this code with the actual route handler so that people that create a new route are somewhat forced to also think about adding a label for metrics.

Maybe someone has a cool idea on how to address this...

vkgnosis: In v1 we added the label as another value when we combine the filters https://github.com/gnosis/dex-services/blob/6c67eee81dc067f106ca447b40479770401e8a49/price-estimator/src/filter.rs#L27 . You might be able to do the same thing here. We didn't use wrap_fn back then so not sure if it works out with the filter types. For it to work the filters have to return the same type (like `Json` instead of `impl Reply`) so that we can use `unify`.

fleupold: Thanks for this pointer, I missed the `unify` function when trying to get things to work (I was getting compiler error on the Either type). 

A solution that works with `impl Reply` is to create a wrapper struct `LabelledReply` and boxing the resulting Reply of the filter together with a label inside `handle_all_routes`

bh2smith (commented):
This is some fancy stuff. I am not confident enough in my understanding of what is happening to give a legitimate approval. ✅ Have seen. Looks good though.

vkgnosis (approved):

Issue #444, opened 2021-04-01
Make orderbook maintenance HPA compatible


fleupold:
At the moment, we can run into a primary key violation if two API pods are running at the same time (cf. https://gnosisinc.slack.com/archives/CQZBU9T5J/p1617221794003700).

In the [logs](https://logs.gnosisdev.com/goto/7155413645283910869a632983c2045e) we can see that two pods were operating around the time the error happened. This can happen in case we restart pods (rollin update making sure that one pod has to be fully ready before the existing one is shut down) or more likely when we use high availability auto scaling and create multiple instances of our pod (at the moment this happens on xDAI when our trading bot runs, because it's invoking a lot of price estimates).

The issue seems to be coming from the fact that we first delete all events in one transaction and then insert new events in a separate transaction:

https://github.com/gnosis/gp-v2-services/blob/afec9d968d5e0d824122f7509d42b7072c2e63fe/orderbook/src/event_updater.rs#L72-L83

This allows for race conditions in case there are two instances inserting on the same db.

One way to mitigate would be to bundle the updates in a transaction or add some other way for pods to communicate with one another to signal who is taking care of maintenance.

vkgnosis:
The existing logic bundles events into chunks of

```rust
const INSERT_EVENT_BATCH_SIZE: usize = 250;
```

events. The deletion of all events in the range and the insertion of the first chunk happens in one transaction. It surprises me that we ran into this issue because if our assumptions are correct that means we had more than 250 events in the last 25 blocks. Alternatively
* postgres transactions don't work as we think they do and the two transactions ran at the same time
* the node returned the same event twice

I can think of these fixes:

1. Increase the batch size significantly which should make the insertions not landing in the same transaction extremely unlikely. This is an easy fix with no disadvantage as 250 I can't imagine inserting even 10k events together would be a problem.
2. Change the query so that it does not error if the event already exists. Presumably the events are identical so it doesn't matter whether they already exist. In the unlikely case that the two instances disagree on the state of the chain this would be resolved the next time the events update.
3. Do not make event updating part of HPA. HPA makes sense for the http api but it does not for updating events. We can create a separate pod that is only responsible for updating events and make the http api pod only read events.

We should definitely do 1 because it has no drawbacks. I like 3 over 2 because it is a proper solution and less of a band aid. It makes sense to only have one source of event updating and it doesn't make sense to scale this with HPA.

fleupold:
[...]

PR #455, opened 2021-04-08
Increase event insertion batch size

vkgnosis:
To make it less likely that events are added outside of the replacing
postgres transaction we increase the batch size. Should help in case
multiple pods are updating the events at the same time.

See https://github.com/gnosis/gp-v2-services/issues/444

### Test Plan
CI


fleupold (approved):
Can you maybe add a run a local test to make sure such a large batch isn't causing any issues (I don't think it should, but maybe better to try it before committing).

bh2smith (approved):

vkgnosis:
Locally made a new test that inserts 10k invalidation events.

PR #456, opened 2021-04-08
[Easy] Static Analysis (dedup code block)

bh2smith:
Just a simple cleanup suggested by my IDE.

### DIFF
--- a/shared/src/uniswap_pool.rs
+++ b/shared/src/uniswap_pool.rs
@@ -90,7 +90,7 @@ impl Pool {
                 self.tokens.get().1,
             )
         } else {
-            assert!(base_token == self.tokens.get().1, "Token not part of pool");
+            assert_eq!(base_token, self.tokens.get().1, "Token not part of pool");
             (
                 BigInt::from(self.reserves.1),
                 BigInt::from(self.reserves.0),
--- a/shared/src/uniswap_pool.rs
+++ b/shared/src/uniswap_pool.rs
@@ -20,6 +20,10 @@ const HONEYSWAP_PAIR_INIT_CODE: [u8; 32] =
     hex!("3f88503e8580ab941773b59034fb4b2a63e86dbc031b3633a925533ad3ed2b93");
 const MAX_BATCH_SIZE: usize = 100;
 
+/// This type denotes `(reserve_a, reserve_b, token_b)` where
+/// `reserve_a` refers to the reserve of the excluded token.
+type RelativeReserves = (U256, U256, H160);
+
 #[async_trait::async_trait]
 pub trait PoolFetching: Send + Sync {
     async fn fetch(&self, token_pairs: HashSet<TokenPair>) -> Vec<Pool>;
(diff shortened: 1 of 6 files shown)


shared/src/uniswap_pool.rs
```diff
@@ -69,25 +55,29 @@ impl Pool {
...
     pub fn get_amount_in(&self, token_out: H160, amount_out: U256) -> Option<(U256, H160)> {
+        let (reserve_out, reserve_in, token_in) = self.get_reserves(token_out);
+        Some((
+            self.amount_in(amount_out, reserve_in, reserve_out)?,
+            token_in,
+        ))
+    }
+
+    fn get_reserves(&self, token: H160) -> (U256, U256, H160) {
         // https://github.com/Uniswap/uniswap-v2-periphery/blob/master/contracts/libraries/UniswapV2Library.sol#L53
```
vkgnosis, line 66: IMO with this signature it's hard to understand what the function is doing without seeing how it is called and I feel the code became harder to read despite eliminating some duplicated code. It's unclear what the parameter is and how it relates to the return type.
Maybe it can be refactored in a way that that is clear or we keep the duplicate code.

bh2smith: I would argue the code was already hard to read and that this change doesn't make it any harder. I will aim to make this more clear by providing better description, names and function signature.

bh2smith: I have gone and fully understood the purpose of the method, renamed it, defined a type and documented the functionality. I think it is much more clear now. Could probably cut a bit of the `get_relative_reserves` doc-string since the type is very well defined. Let me know what you think.

vkgnosis (approved):

shared/src/uniswap_pool.rs
vkgnosis: Your doc comments makes it very clear, thanks.

shared/src/uniswap_pool.rs
```diff
@@ -69,25 +58,35 @@ impl Pool {
     /// Given an output amount and token, returns a required input amount and address of the other asset.
     /// Returns None if operation not possible due to arithmetic issues (e.g. over or underflow, reserve too small)
     pub fn get_amount_in(&self, token_out: H160, amount_out: U256) -> Option<(U256, H160)> {
+        let (reserve_out, reserve_in, token_in) = self.get_relative_reserves(token_out);
```
fleupold, line 61: I'm personally not a huge fan of splitting this out into its own method, because it disconnects the logic from the ordering of the tuple. If we want the extra method, I'd slightly prefer a named struct as a return type to avoid mixup of `reserve_in`, `reserve_out` (both are U256). I can see how this is a lot of boilerplate code for arguably little benefit as all of this is implementation detail to this component and likely not used outside. 

But then my general question is: What is the benefit of the extra method. Anyways, super nitty and I'm fine either.

fleupold (approved):
Thanks

🧽 

✨

shared/src/uniswap_pool.rs
bh2smith: I consider it a valuable method that could potentially be used for another purpose at some point. Not generally a fan of performing swaps like this in a method that should be easy follow. In my opinion such swaps should always be delegated to their own helper methods. Happy to restructure if you'd like but I find the readability of the amount in and amount out methods now much simpler to follow.

PR #457, opened 2021-04-09
Expose full order struct on Trade

fleupold:
Prerequisite for solver metrics. I'd like to measure how long orders take from creation to successful settlement. For this we need to have access to the order's creation date from the settlement struct. We already expose `OrderCreation` data on the settlements via the trades field, however the creation timestamp is part of `OrderMetadata`

This PR therefore modifies the Trade struct to contain the full order instead of just the creation data.

### Test Plan
CI


vkgnosis (approved):
Was thinking about this too for my PR where I needed meta info :+1:

PR #458, opened 2021-04-09
Settle orders after a minimum wait period to increase cow chance

vkgnosis:
Currently the driver attempts settlement every 30 seconds and settles
what the solvers produce immediately. This makes it harder to get
coincidence of wants because the timing of order submission has to be
right as we have seen in the friday tests.
Instead of using a long settlement interval this commit only applies
settlements if they contain at least one order that is older than some
minimum age like 1 minute. This increases the cow chance and speeds of
settlements of several individual settlements one after another as the
wait time between them is decreases.

Once there are multiple solvers this might be something that needs more discussion or dao rule but until then this seems like a strict improvement to the existing logic.

### Test Plan
e2e test still works. couldn't find a great way to factor this out and the logic is simply.

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -14,9 +14,7 @@ use itertools::Itertools;
 use num::BigRational;
 use primitive_types::H160;
 use shared::price_estimate::PriceEstimating;
-use std::collections::HashMap;
-use std::{cmp::Reverse, sync::Arc, time::Duration};
-use tracing::info;
+use std::{cmp::Reverse, collections::HashMap, sync::Arc, time::Duration};
 
 // There is no economic viability calculation yet so we're using an arbitrary very high cap to
 // protect against a gas estimator giving bogus results that would drain all our funds.
@@ -32,6 +30,7 @@ pub struct Driver {
     target_confirm_time: Duration,
     settle_interval: Duration,
     native_token: H160,
+    min_order_age: Duration,
 }
 
 impl Driver {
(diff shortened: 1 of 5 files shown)


solver/src/driver.rs
```diff
@@ -179,12 +185,35 @@ impl Driver {
...
+                continue;
+            }
+
+            // If all orders are younger than self.min_order_age skip settlement. Orders will still
+            // be settled once they have been in the order book for longer. This makes coincidence
+            // of wants more likely.
+            let should_be_settled_immediately = settlement.trades.iter().any(|trade| {
+                let meta = &orders
+                    .iter()
+                    .find(|order| order.order_creation == trade.order)
```
vkgnosis, line 204: This is optimal from a big-o perspective but we know that the number of possible trades in a single settlement won't be huge and it makes the code easier than for example creating a hash map of order uid.

solver/src/main.rs
```diff
@@ -66,6 +66,15 @@ struct Arguments {
         use_delimiter = true,
     )]
     solvers: Vec<SolverType>,
+
+    /// Every how often we should execute the driver's run loop
```
fleupold, line 70: copy paste error

fleupold (approved):
Nice change. We might want to make the wait time depending on other factors (e.g. the AppID used or the fee amount paid). 

One idea of Martin was to give people a larger fee discount if they are willing to wait significantly longer (similar to how Uber pool launched in the early days). The balancer interface users would likely (at least initially) want to be as close to their normal experience and therefore have a very short wait time.

But this can be added later on...

PR #459, opened 2021-04-09
Solver Metrics

fleupold:
For now, last part of #425.

Adding some hopefully useful metrics to the driver which can give insight on how many orders of what type are currently considered, how many trades are happening and how long people wait on average to see their trade matched locall.

### Test Plan
Run locally and do a e2e swap on CowSwap. Afterwards check `http://localhost:9587/metrics` to see the relevant metrics

### DIFF
--- a/solver/src/metrics.rs
+++ b/solver/src/metrics.rs
@@ -1,0 +1,107 @@
+use std::{convert::TryInto, time::Instant};
+
+use anyhow::Result;
+use model::order::Order;
+use prometheus::{IntCounter, IntCounterVec, IntGaugeVec, Opts, Registry};
+use strum::{AsStaticRef, VariantNames};
+
+use crate::liquidity::Liquidity;
+
+pub trait SolverMetrics {
+    fn liquidity_fetched(&self, liquidity: &[Liquidity]);
+    fn settlement_computed(&self, solver_type: &str, start: Instant);
+    fn order_settled(&self, order: &Order);
+}
+
+// TODO add labeled interaction counter once we support more than one interaction
+pub struct Metrics {
+    trade_counter: IntCounter,
+    order_settlement_time: IntCounter,
+    solver_computation_time: IntCounterVec,
+    liquidity: IntGaugeVec,
+}
+
+impl Metrics {
+    pub fn new(registry: &Registry) -> Result<Self> {
+        let trade_counter =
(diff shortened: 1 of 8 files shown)


solver/src/metrics.rs
```diff
@@ -0,0 +1,107 @@
...
+impl Metrics {
+    pub fn new(registry: &Registry) -> Result<Self> {
+        let trade_counter =
+            IntCounter::new("gp_v2_solver_trade_counter", "Number of trades settled")?;
+        registry.register(Box::new(trade_counter.clone()))?;
+
+        let order_settlement_time = IntCounter::new(
+            "gp_v2_solver_order_settlement_time_seconds",
+            "Counter for the number of seconds between creation and settlement of an order",
+        )?;
```
vkgnosis, line 33: Maybe better as a histogram?

vkgnosis: This and the trade counter could also both be calculated based on the events we already have in the database. Not sure if Grafana can easily visualize that though so it's fine if we have it here too.

solver/src/metrics.rs
```diff
@@ -0,0 +1,107 @@
...
+        registry.register(Box::new(order_settlement_time.clone()))?;
+
+        let solver_computation_time = IntCounterVec::new(
+            Opts::new(
+                "gp_v2_solver_computation_time_ms",
+                "Ms each solver takes to compute their solution",
+            ),
+            &["solver_type"],
+        )?;
```
vkgnosis, line 42: Also seems like it should be a histogram unless I can't imagine the right visualization. I don't mind changing these types later either if we want to merge this and make a dashboard first.

vkgnosis (approved):

solver/src/metrics.rs
fleupold: I was thinking about using a histogram (and agree that long term it's the better visualization), but then thought that given we will only have measurements in the beginning it might be nice to see individual times rather than bucketed ones. I also didn't really know what a good bucketing strategy would be so maybe we make it more lossy as data increases and it becomes more clear what relevant buckets could be?

For aggregated past data (e.g. report what was our average wait time in the last months), maybe querying db events might be better (I don't think it can be easily visualized in Grafana though).

solver/src/metrics.rs
fleupold: Here is was afraid that different solvers might have very different values so finding good bucket sizes is even harder than for the order settlement time.

PR #465, opened 2021-04-13
Add WETH unwrap interaction

nlordell:
This PR adds a new `Interaction` implementation for unwrapping WETH. This is a small step towards buying ETH support.

/cc @fedgiac 

### Test Plan

Added new unit test verifying encoding.

### DIFF
--- a/solver/src/interactions/weth.rs
+++ b/solver/src/interactions/weth.rs
@@ -1,0 +1,45 @@
+use crate::{encoding::EncodedInteraction, settlement::Interaction};
+use contracts::WETH9;
+use primitive_types::U256;
+
+#[derive(Debug)]
+pub struct UnwrapWethInteraction {
+    pub weth: WETH9,
+    pub amount: U256,
+}
+
+impl Interaction for UnwrapWethInteraction {
+    fn encode(&self) -> Vec<EncodedInteraction> {
+        let method = self.weth.withdraw(self.amount);
+        let calldata = method.tx.data.expect("no calldata").0;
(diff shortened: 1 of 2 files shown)


bh2smith (approved):
Nice and pretty.

vkgnosis (approved):

solver/src/interactions/weth.rs
```diff
@@ -0,0 +1,45 @@
...
+pub struct UnwrapWethInteraction {
+    pub weth: WETH9,
+    pub amount: U256,
+}
+
+impl Interaction for UnwrapWethInteraction {
+    fn encode(&self) -> Vec<EncodedInteraction> {
+        let method = self.weth.withdraw(self.amount);
+        let calldata = method.tx.data.expect("no calldata").0;
+        vec![(self.weth.address(), 0.into(), calldata)]
```
fleupold, line 15: Unrelated to this PR, but wondering if we should refactor EncodedInteraction to a named struct to make this read a bit nicer?

fleupold (approved):

solver/src/interactions/weth.rs
nlordell: One issue here is that `EncodedInteraction` represents the type that is generated by `ethcontract` (because it doesn't support ABI named tuples).

I think the best way to achieve this would actually be to add support for named tuples in `ethcontract` instead of adding an intermediate `InteractionData` representation (although the latter may be a nice stepping stone for now).

PR #468, opened 2021-04-13
Send order/uniswap costs to http solver

marcovc:
I've left a few questions in the code.


vkgnosis (commented):
I feel that the gas price should be part of the Solver trait solver function because it's always relevant information (even if other solvers don't use it) and it means we don't have to make decisions here like what parameters to give the gas estimation or what to do if estimation is impossible. What do you think?

solver/src/http_solver.rs
```diff
@@ -23,9 +24,11 @@ use std::{
     sync::Arc,
 };
 
+const GAS_PER_ORDER: u128 = 37676;
```
fleupold, line 27: Maybe add a comment on how you derived at the values. I feel they are a bit low, given that a settlement with one trade and Uniswap already costs > 180k gas.

solver/src/http_solver.rs
```diff
@@ -272,6 +289,22 @@ impl HttpSolver {
         serde_json::from_str(text.as_str())
             .with_context(|| format!("failed to decode response json, {}", context()))
     }
+
+    async fn order_cost(&self) -> Result<u128> {
+        // Q: should we use estimate_with_limits below instead?
+        // Q: are the units (gwei) correct?
+        let gas_price_gwei = self.gas_price_estimator.estimate().await? * 1e9;
```
fleupold, line 296: I think the price estimator returns the price in WEI so no need to multiply with 1e9

fleupold (commented):
Looks good! I like @vkgnosis suggestion of passing in the current gas price in the `solve(&self, orders: Vec<Liquidity>)` function as it seems relevant to a lot of solvers. 

The gas price is also be relevant when we evaluate the objective function in the end (it doesn't yet, but should eventually also take execution cost into account)

solver/src/driver.rs
```diff
@@ -162,14 +162,21 @@ impl Driver {
 
         self.metrics.liquidity_fetched(&liquidity);
 
+        let gas_price = self
+            .gas_price_estimator
+            .estimate()
+            .await
+            .context("failed to estimate gas price")?;
+        tracing::debug!("solving with gas price of {}", gas_price);
```
marcovc, line 170: Should we use a default price instead?

solver/src/http_solver.rs
marcovc: Done. For "better" estimations we should probably use a table close to the values we are expecting (I guess we aren't expecting 50 trades).

solver/src/driver.rs
vkgnosis: I think it's correct to fail in that case. We definitely need the gas price estimation to work for submitting our transactions.

vkgnosis (approved):

PR #469, opened 2021-04-13
Check that balance is transferable on new order

vkgnosis:
We can improve checking for available balance by checking whether the
balance is transferable. This can detect many weird erc20 tokens like the kick token where the result of getBalance is inconsistent with what is checked during the transfer.
The transfer check implies balance existing so I removed the balance check.
A todo is that we should also do this as part of maintenance for example for tokens that can be "paused" which disables all trading until unpaused. I wasn't yet sure how to organize that code correctly and wanted the base functionality independently reviewed.

### Test Plan
e2e test still works and tested new function with the 2 ignored tests on real networks

### DIFF
--- a/orderbook/src/account_balances.rs
+++ b/orderbook/src/account_balances.rs
@@ -168,7 +168,7 @@ impl BalanceFetching for Web3BalanceFetcher {
         let block = Some(BlockId::Number(BlockNumber::Latest));
         let response = self.web3.eth().call(call_request, block).await;
         response
-            .map(|bytes| is_empty_or_nonzero(bytes.0.as_slice()))
+            .map(|bytes| is_empty_or_truthy(bytes.0.as_slice()))
             .unwrap_or(false)
     }
 }
--- a/orderbook/src/account_balances.rs
+++ b/orderbook/src/account_balances.rs
@@ -5,6 +5,7 @@ use futures::future::{join3, join_all};
 use primitive_types::{H160, U256};
 use shared::Web3;
 use std::{collections::HashMap, sync::Mutex};
+use web3::types::{BlockId, BlockNumber, CallRequest};
 
 const MAX_BATCH_SIZE: usize = 100;
 
@@ -153,14 +154,30 @@ impl BalanceFetching for Web3BalanceFetcher {
 
     async fn can_transfer(&self, token: H160, from: H160, amount: U256) -> bool {
         let instance = ERC20::at(&self.web3, token);
-        instance
+        let calldata = instance
             .transfer_from(from, self.settlement_contract, amount)
-            .view()
-            .from(self.allowance_manager)
-            .call()
-            .await
+            .tx
+            .data
+            .unwrap();
+        let call_request = CallRequest {
+            from: Some(self.allowance_manager),
+            to: Some(token),
+            data: Some(calldata),
+            ..Default::default()
+        };
+        let block = Some(BlockId::Number(BlockNumber::Latest));
+        let response = self.web3.eth().call(call_request, block).await;
+        response
+            .map(|bytes| is_empty_or_nonzero(bytes.0.as_slice()))
             .unwrap_or(false)
     }
+}
+
+fn is_empty_or_nonzero(bytes: &[u8]) -> bool {
+    if bytes.is_empty() {
+        return true;
+    }
+    bytes.iter().any(|byte| *byte > 0)
 }
 
 #[cfg(test)]
--- a/orderbook/src/account_balances.rs
+++ b/orderbook/src/account_balances.rs
@@ -174,10 +174,11 @@ impl BalanceFetching for Web3BalanceFetcher {
 }
 
 fn is_empty_or_nonzero(bytes: &[u8]) -> bool {
-    if bytes.is_empty() {
-        return true;
+    match bytes.len() {
+        0 => true,
+        32 => bytes.iter().any(|byte| *byte > 0),
+        _ => false,
     }
-    bytes.iter().any(|byte| *byte > 0)
 }
 
 #[cfg(test)]
@@ -242,7 +243,8 @@ mod tests {
             .await
             .expect("MintableERC20 deployment failed");
 
-        let fetcher = Web3BalanceFetcher::new(web3, allowance_target.address(), H160::zero());
+        let fetcher =
+            Web3BalanceFetcher::new(web3, allowance_target.address(), H160::from_low_u64_be(1));
 
         // Not available until registered
         assert_eq!(fetcher.get_balance(trader.address(), token.address()), None,);
--- a/orderbook/src/account_balances.rs
+++ b/orderbook/src/account_balances.rs
@@ -173,7 +173,7 @@ impl BalanceFetching for Web3BalanceFetcher {
     }
 }
 
-fn is_empty_or_nonzero(bytes: &[u8]) -> bool {
+fn is_empty_or_truthy(bytes: &[u8]) -> bool {
     match bytes.len() {
         0 => true,
         32 => bytes.iter().any(|byte| *byte > 0),
--- a/orderbook/src/account_balances.rs
+++ b/orderbook/src/account_balances.rs
@@ -1,12 +1,10 @@
 use anyhow::Result;
+use contracts::ERC20;
 use ethcontract::batch::CallBatch;
 use futures::future::{join3, join_all};
-use std::collections::HashMap;
-use std::sync::Mutex;
-
-use contracts::ERC20;
 use primitive_types::{H160, U256};
 use shared::Web3;
+use std::{collections::HashMap, sync::Mutex};
 
 const MAX_BATCH_SIZE: usize = 100;
 
@@ -25,11 +23,19 @@ pub trait BalanceFetching: Send + Sync {
 
     // Called periodically to perform potential updates on registered balances
     async fn update(&self);
+
+    // Check if the allowance manager would be able to call transfer_from with these parameters.
+    // This is useful for tokens that are not consistent about their internal checks and what they
+    // report as balance and allowance. By checking whether the actual transfer would suceed we can
+    // be more certain (but still not 100%) that the balance really is available to the settlement
+    // contract.
+    async fn can_transfer(&self, token: H160, from: H160, amount: U256) -> bool;
 }
 
 pub struct Web3BalanceFetcher {
     web3: Web3,
     allowance_manager: H160,
+    settlement_contract: H160,
     // Mapping of address, token to balance, allowance
     balances: Mutex<HashMap<SubscriptionKey, SubscriptionValue>>,
 }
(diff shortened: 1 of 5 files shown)


orderbook/src/account_balances.rs
```diff
@@ -25,11 +25,19 @@ pub trait BalanceFetching: Send + Sync {
 
     // Called periodically to perform potential updates on registered balances
     async fn update(&self);
+
+    // Check if the allowance manager would be able to call transfer_from with these parameters.
+    // This is useful for tokens that are not consistent about their internal checks and what they
+    // report as balance and allowance. By checking whether the actual transfer would suceed we can
+    // be more certain (but still not 100%) that the balance really is available to the settlement
+    // contract.
+    async fn can_transfer(&self, token: H160, from: H160, amount: U256) -> bool;
```
vkgnosis, line 34: Not sure if BalanceFetching is the best place for this but seeing as I plan on making this part of the maintenance I think it makes sense to integrate it with the batch calls here in a future PR.

fleupold: What amount are you planning to use for the maintenance job?

fleupold (commented):
From a user perspective, I still think these tokens should be deny listed so that we cannot even get the user excited about quoting a potential exchange rate (it would suck to see you have $400 worth of KICK tokens just to then get an InsufficientFunds error when you are trying to place an order).

But the change is a strict improvement and prevents bad orders from causing alerts on the solver. I'm starting to get slightly worried about the increasing complexity of the add order function. Maybe all these checks should be placed in a more structured component and extensively tested (not as part of this PR).

orderbook/src/account_balances.rs
vkgnosis: The same balance that we already fetch. Which is a little annoying because we have to fetch the balance first and then do another batch call with the transfer. Alternative we could just try `1` to catch paused tokens.

orderbook/src/account_balances.rs
```diff
@@ -143,6 +152,28 @@ impl BalanceFetching for Web3BalanceFetcher {
...
+            from: Some(self.allowance_manager),
+            to: Some(token),
+            data: Some(calldata),
+            ..Default::default()
+        };
+        let block = Some(BlockId::Number(BlockNumber::Latest));
+        const TRUE: [u8; 32] =
+            hex!("0000000000000000000000000000000000000000000000000000000000000001");
+        match self.web3.eth().call(call_request, block).await {
+            Ok(bytes) => bytes.0 == TRUE,
```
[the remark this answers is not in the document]
vkgnosis: I see. This is the same thing handleod in https://github.com/gnosis/gp-v2-contracts/blob/26e5eb19e639bc62f50c604a1dd38348799e520c/src/contracts/libraries/GPv2SafeERC20.sol#L110 . I wonder if we can call this same library function from here to get exactly the same behavior.

orderbook/src/account_balances.rs
```diff
@@ -143,6 +152,28 @@ impl BalanceFetching for Web3BalanceFetcher {
...
     }
+
+    async fn can_transfer(&self, token: H160, from: H160, amount: U256) -> bool {
+        let instance = ERC20::at(&self.web3, token);
+        let calldata = instance
+            .transfer_from(from, self.settlement_contract, amount)
+            .tx
+            .data
+            .unwrap();
+        let call_request = CallRequest {
```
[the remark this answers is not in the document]
vkgnosis: fixed

orderbook/src/account_balances.rs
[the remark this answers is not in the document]
vkgnosis: I think we can't because it's an internal library (no public function).

orderbook/src/account_balances.rs
fleupold: Yeah, I think this is not super obvious. Fetching the whole allowance/balance might be too restrictive if a token e.g. implements a 0.1% transfer fee we could still sell up to 99.9% of our balance. Maybe we can add the total of all orders of that token together (similar to how we would compute the available balance a user has in a given token)? But `1` sounds like a simple and good enough solution to catch paused or frozen tokens.

orderbook/src/account_balances.rs
```diff
@@ -143,15 +150,64 @@ impl BalanceFetching for Web3BalanceFetcher {
         };
         let _ = self._register_many(subscriptions.into_iter()).await;
     }
+
+    async fn can_transfer(&self, token: H160, from: H160, amount: U256) -> bool {
+        let instance = ERC20::at(&self.web3, token);
+        instance
```
fleupold, line 156: 👍

fleupold (approved):

orderbook/src/account_balances.rs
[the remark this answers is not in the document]
vkgnosis: @fleupold What about this thread? Merge even though we would fail on those tokens? Not sure how many tokens are still affected now that it's 2 years after that blog post.
If we wanted to handle this better I'd have to go back to using rust-web3 directly without ethcontract and manually inspect the response bytes.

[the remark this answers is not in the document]
fleupold: Right, I forgot about this! How does ethcontract handle a contract that is returning no bytes when it expects a boolean (cc @nlordell)? The most known token to me in that list is OMG (which still has reasonable liquidity so we could test with it). 

Maybe we could actually go back to the manual call implementation and consider a 0 length return data a success?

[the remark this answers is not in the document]
vkgnosis: From having touched the token decoding code recently I believe ethcontract would return this as a decoding error. I think your last suggestion makes sense.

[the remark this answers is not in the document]
nlordell: > From having touched the token decoding code recently I believe ethcontract would return this as a decoding error

That is my assumption as well.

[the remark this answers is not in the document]
nlordell: I think, what would make the most sense here would be to do a raw `eth_call`:
```rust
let response = web3.eth().call(
    CallRequest {
                from: Some(owner_address),
                to: Some(token_address),
                data: token.transfer(amount).tx.data,
    },
    Some(block_id),
).await?;
ensure!(is_empty_bytes_or_true(response), "transfer failed"); 
```

orderbook/src/account_balances.rs
nlordell: Should this also always have `from` be the allowance manager and `to` be the settlement contract address (in case it checks transfers for specific addresses? The [OpenZeppelin ERC20 implementation disallows transferring to address 0](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/3157aff844dfc19fdd68988d9c859959f2d0a4f9/contracts/token/ERC20/ERC20.sol#L213) for example).

orderbook/src/account_balances.rs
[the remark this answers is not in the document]
vkgnosis: Changed it back to this version.

orderbook/src/account_balances.rs
vkgnosis: Yes always hardcoded.

orderbook/src/orderbook.rs
```diff
@@ -247,25 +263,16 @@ fn solvable_orders(mut orders: Vec<Order>, balances: &HashMap<(H160, H160), U256
...
-    let owner = order.order_meta_data.owner;
-    let token = order.order_creation.sell_token;
-    balance_fetcher.register(owner, token).await;
-    let minimum_balance = if order.order_creation.partially_fillable {
-        1.into()
+// Mininum balance user must have in sell token for order to be accepted. None if no balance is
+// sufficient.
+fn minimum_balance(order: &Order) -> Option<U256> {
+    if order.order_creation.partially_fillable {
+        Some(U256::from(1))
```
nlordell, line 270: Doesn't this potentially lead to problems if the user can transfer N tokens, but has a balance of N + M, where the balance check would pass (since the transfer check would succeed) but the solver would be given an incorrect available balance of N + M?

orderbook/src/account_balances.rs
```diff
@@ -169,7 +242,7 @@ mod tests {
             .await
             .expect("MintableERC20 deployment failed");
 
-        let fetcher = Web3BalanceFetcher::new(web3, allowance_target.address());
+        let fetcher = Web3BalanceFetcher::new(web3, allowance_target.address(), H160::zero());
```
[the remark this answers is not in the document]
vkgnosis: Changed to an arbitrary nonzero address

orderbook/src/orderbook.rs
vkgnosis: But then you can't place partially fillable orders for more than your current balance. Agree it's not perfect but not sure which the better tradeoff is. Also depends on the decision what amount we are going to check for our regular balance updates (see first thread).

nlordell: Yeah, its a tough one. I'm inclined to just require the entire remaining amount for now as its "safer" (doesn't produce invalid settlements). On other hand, I do agree that in the long term, the requirement should be relaxed.

Since we don't do partially fillable orders in the FE now anyways, I don't think its super important.

orderbook/src/account_balances.rs
```diff
@@ -143,15 +151,81 @@ impl BalanceFetching for Web3BalanceFetcher {
...
+        response
+            .map(|bytes| is_empty_or_nonzero(bytes.0.as_slice()))
+            .unwrap_or(false)
+    }
+}
+
+fn is_empty_or_nonzero(bytes: &[u8]) -> bool {
+    match bytes.len() {
+        0 => true,
+        32 => bytes.iter().any(|byte| *byte > 0),
```
nlordell, line 179: Because the LSB tends to be `1` you could do 
```suggestion
        32 => bytes.iter().rev().any(|byte| *byte > 0),
```

Just a fun little optimization trick, not actually a suggestion for change.

orderbook/src/account_balances.rs
```diff
@@ -143,15 +151,81 @@ impl BalanceFetching for Web3BalanceFetcher {
...
+        };
+        let block = Some(BlockId::Number(BlockNumber::Latest));
+        let response = self.web3.eth().call(call_request, block).await;
+        response
+            .map(|bytes| is_empty_or_nonzero(bytes.0.as_slice()))
+            .unwrap_or(false)
+    }
+}
+
+fn is_empty_or_nonzero(bytes: &[u8]) -> bool {
```
nlordell, line 176: ubernit: *feel free to disregard* - I would call this
```suggestion
fn is_empty_or_truthy(bytes: &[u8]) -> bool {
```

Since it does not accept any non-zero byte slice.

nlordell (approved):

orderbook/src/orderbook.rs
nlordell: From my point of view this doesn't block merging the PR (since the FE doesn't really support these kinds of orders anyway for the moment). Just something to be aware of.

PR #471, opened 2021-04-14
Make weth available to order settlement handling

fedgiac:
This PR makes a weth instance available when computing a limit order. It doesn't change the behavior of the current code.
It will later be used to replace the placeholder ETH address with the correct WETH address, and to help changing the settlement procedure.

/cc @nlordell 

### Test Plan
Existing unit tests.

### DIFF
--- a/solver/src/orderbook.rs
+++ b/solver/src/orderbook.rs
@@ -1,3 +1,4 @@
+use contracts::WETH9;
 use model::order::Order;
 use reqwest::{Client, Url};
 use std::time::Duration;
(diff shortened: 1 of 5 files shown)


vkgnosis (approved):

solver/src/orderbook.rs
```diff
@@ -21,19 +27,29 @@ impl OrderBookApi {
         url.set_path(PATH);
         self.client.get(url).send().await?.json().await
     }
+
+    pub fn get_native_token(&self) -> WETH9 {
+        self.native_token.clone()
+    }
```
fleupold, line 33: What's the overhead here of cloning the actual Contract instance vs passing around an Arc?

fleupold (approved):
Looks nice!

solver/src/orderbook.rs
nlordell: They are lightweight. The actual ABI is kept in a `lazy_static` object so that cloning just means cloning the transport (which, AFAIK, boils down to cloning an `Arc`).

Issue #445, opened 2021-04-01
Solver restart causes to lose track of onging settlement


fleupold:
When the solver restarts while it is waiting for a settlement to be mined, the new process doesn't find the pending transaction and just re-issues a new transaction with the same nonce. This can lead to the following error in case the gas price has stayed flat or even decreased compared to the pending tx:

> settlement: settlement transaction failed
Caused by:
   0: method 'settle(address[],uint256[],(uint256,uint256,address,uint256,uint256,uint32,bytes32,uint256,uint256,uint256,bytes)[],(address,uint256,bytes)[][3])' failure: web3 error: RPC error: Error { code: ServerError(-32010), message: "Transaction gas price supplied is too low. There is another transaction with same nonce in the queue. Try increasing the gas price or incrementing the nonce.", data: None }

Maybe we can add logic to fetch the pending transaction with in the retry code and use its gas price as a starting point for the  new transaction (+ min increase).

Tangentially related to https://github.com/gnosis/gp-transaction-retry/issues/3

vkgnosis:
I don't think there is a reasonable way to recover pending transactions from the node, looking at the eth rpc api and taking load balancing between several nodes into account. To properly fix this we need to keep track of transactions we've sent out in the database. This is also related to being able to react to transactions timing out because the node failed to propagate them.
I envision a table that keeps track of transactions with hash, nonce, gas price. With that we can detect nodes forgetting about transactions and recover previous still pending transactions. Then with this knowledge we can cancel old transactions with noop or overwrite them with new ones.
This makes the retry component more complicated because it is just responsible for one transaction but also has to keep track of state for multiple.

fleupold:
Does this really require persistence? The approach for finding the pending tx could be to get `eth_getBlockByNumber("pending", true)` and check in the list of txs if there is one of our transactions. This would only have to happen once at cold start (otherwise we can keep track in memory).

I agree that this is not guaranteed to work if we are behind a load balancer but in this case the node that doesn't think we have a pending tx should also accept our new transaction (so eventually we should either find the existing one or succeed at sending our new one).

I'm worried that adding persistence to the driver (and every other system that wants to use the retry component) is quite a big hammer and I'm not sure we tried more low hanging fruits yet.

vkgnosis:
> in this case the node that doesn't think we have a pending tx should also accept our new transaction

[...]

PR #472, opened 2021-04-14
Use gas price from pending transaction if detected

vkgnosis:
This is a partial fix for https://github.com/gnosis/gp-v2-services/issues/445 . It should work fine in case of restarts but it might still print an error if the pending transactions gets mined first. It can also help against the case where the node forgets a transaction.

A more proper solution should go into the transaction-retry crate but that's harder to figure out how to do so I'd like to have this in the mean time.

### Test Plan
Hard to test until we see this actually happening. I did test the getBlock(Pending) call to see that it really returns pending transactions and I saw that the total gas use of pending transactions was well above 12 mil so it seems to be bigger than just "one" pending block.

### DIFF
--- a/solver/src/settlement_submission.rs
+++ b/solver/src/settlement_submission.rs
@@ -14,11 +14,13 @@ use ethcontract::{
     transaction::TransactionBuilder,
     Web3,
 };
+use futures::stream::StreamExt;
 use gas_estimation::GasPriceEstimating;
 use gas_price_stream::gas_price_stream;
 use primitive_types::{H160, U256};
 use std::time::Duration;
 use transaction_retry::RetryResult;
+use web3::types::{BlockId, BlockNumber};
 
 const GAS_PRICE_REFRESH_INTERVAL: Duration = Duration::from_secs(15);
 const ESTIMATE_GAS_LIMIT_FACTOR: f64 = 1.2;
@@ -31,12 +33,23 @@ pub async fn submit(
     gas_price_cap: f64,
     settlement: Settlement,
 ) -> Result<()> {
-    let nonce = transaction_count(contract)
-        .await
-        .context("failed to get transaction_count")?;
     let settlement = encode_settlement(&settlement)?;
     // Check that a simulation of the transaction works before submitting it.
     simulate_settlement(&settlement, contract).await?;
+
+    let nonce = transaction_count(contract)
+        .await
+        .context("failed to get transaction_count")?;
+    let address = &contract
+        .defaults()
+        .from
+        .clone()
+        .expect("no default sender address")
+        .address();
+    let web3 = contract.raw_instance().web3();
+    let pending_gas_price = recover_gas_price_from_pending_transaction(&web3, &address, nonce)
+        .await
+        .context("failed to get pending gas price")?;
 
     // Account for some buffer in the gas limit in case racing state changes result in slightly more heavy computation at execution time
     let gas_limit = retry::settle_method_builder(contract, settlement.clone())
@@ -55,7 +68,30 @@ pub async fn submit(
     };
     // We never cancel.
     let cancel_future = std::future::pending::<CancelSender>();
-    let stream = gas_price_stream(target_confirm_time, gas_price_cap, gas_limit, gas);
+    if let Some(gas_price) = pending_gas_price {
+        tracing::info!(
+            "detected existing pending transaction with gas price {}",
+            gas_price
+        );
+    }
+
+    // It is possible that there is a pending transaction we don't know about because the driver
+    // got restarted while it was in progress. Sending a new transaction could fail in that case
+    // because the gas price has not increased. So we make sure that the starting gas price is at
+    // least high enough to accommodate. This isn't perfect because it's still possible that that
+    // transaction gets mined first in which case our new transaction would fail with "nonce already
+    // used".
+    let pending_gas_price = pending_gas_price.map(|gas_price| {
+        transaction_retry::gas_price_increase::minimum_increase(gas_price.to_f64_lossy())
+    });
+    let stream = gas_price_stream(
+        target_confirm_time,
+        gas_price_cap,
+        gas_limit,
+        gas,
+        pending_gas_price,
+    )
+    .boxed();
 
     match transaction_retry::retry(settlement_sender, cancel_future, stream).await {
         Some(RetryResult::Submitted(result)) => {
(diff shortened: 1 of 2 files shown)


solver/src/settlement_submission.rs
```diff
@@ -143,3 +172,19 @@ async fn tenderly_link(
...
+) -> Result<Option<U256>> {
+    let block = web3
+        .eth()
+        .block_with_txs(BlockId::Number(BlockNumber::Pending))
+        .await?
+        .ok_or_else(|| anyhow!("empty block"))?;
+    let transaction = block
+        .transactions
+        .iter()
+        .find(|transaction| transaction.from == *address);
```
fleupold, line 188: Should we also check nonce here to make sure we are finding the right transaction?

solver/src/settlement_submission.rs
```diff
@@ -31,13 +33,22 @@ pub async fn submit(
...
 
+    let nonce = transaction_count(contract)
+        .await
+        .context("failed to get transaction_count")?;
+    let address = match &contract.defaults().from {
+        Some(account) => account.address(),
+        None => return Err(anyhow!("no default sender address")),
+    };
+    let web3 = contract.raw_instance().web3();
+    let pending_gas_price = recover_gas_price_from_pending_transaction(&web3, &address)
```
fleupold, line 48: nit: Do we need to fail in case we cannot recover the pending transaction? Would it not be "safe" to assume there is none?

solver/src/settlement_submission.rs
```diff
@@ -31,13 +33,22 @@ pub async fn submit(
...
     let settlement = encode_settlement(&settlement)?;
     // Check that a simulation of the transaction works before submitting it.
     simulate_settlement(&settlement, contract).await?;
 
+    let nonce = transaction_count(contract)
+        .await
+        .context("failed to get transaction_count")?;
+    let address = match &contract.defaults().from {
+        Some(account) => account.address(),
+        None => return Err(anyhow!("no default sender address")),
```
fleupold, line 45: Curious on why error here instead of crash? Wouldn't it be a non-recoverable configuration error if from is not set?

solver/src/settlement_submission.rs
```diff
@@ -55,7 +66,25 @@ pub async fn submit(
...
+    // transaction gets mined first in which case our new transaction would fail with "nonce already
+    // used".
+    if let Some(gas_price) = pending_gas_price {
+        tracing::info!(
+            "detected existing pending transaction with gas price {}",
+            gas_price
+        );
+        let gas_price = gas_price.to_f64_lossy();
+        let gas_price = transaction_retry::gas_price_increase::minimum_increase(gas_price);
+        stream = futures::stream::once(std::future::ready(gas_price))
```
fleupold, line 84: Isn't retry expecting a monotonically increasing gas price stream? Here is looks as if it was possible to have the previous (potentially high value, followed by a lower value).

Maybe this would be a good moment to refactor `gas_price_stream` to a proper struct which can be composed like:

```rust
let stream = GasPriceStream::new(target_confirm_time, gas_limit, estimator)
  .capped(gas_price_cap);

if let Some(gas_price) = pending_gas_price {
  stream = stream.previous_value(gas_price)
}

stream = stream.minimum_increase();
```

Of course this can be done as functions as well.

fleupold:
As for testing, I think it's reasonable for this case to do a local mainnet test (with the testing ETH we have) and stop and restart the solver within a short amount of time. Could hardcode the gas price to be somewhat low to be sure that it isn't mined too quickly.

solver/src/settlement_submission.rs
vkgnosis: I wasn't sure what was more appropriate. Changed it to panic now.

solver/src/settlement_submission.rs
vkgnosis: An error in that function means communication with the node failed which is very different from no gas price so I prefer it working this way.

solver/src/settlement_submission.rs
vkgnosis: The nonce would only be wrong if there are multiple pending transactions right? I felt we could ignore that case for now because it should never happen. If we wanted to handle it we could make sure to always take over the lowest pending nonce.

solver/src/settlement_submission.rs
vkgnosis: Oh right, for some reason I thought the enforce_min_increase adapter would handle this but obviously it is only applied to the second stream. I have changed `gas_price_stream` to take another parameter which passes an optional intial gas price. Will think about refactoring it more now.

vkgnosis:
That's a good idea for a test. Will do it once PR is approved.

solver/src/settlement_submission.rs
fleupold: We need to outbid the gas price of the transaction with the same nonce so even if we had more than one pending tx I think the nonce == check would be correct (instead of checking the gas price of the lowest pending nonce):

> "Transaction gas price supplied is too low. There is another transaction with same nonce in the queue. Try increasing the gas price or incrementing the nonce.",

It's true that currently this is not necessary but it feels like a very small addition to pass the nonce and do the check. Given this code path is already an edge case it is risky to forget/not test this if we want to handle multiple parallel settlement transactions.

solver/src/settlement_submission.rs
```diff
@@ -31,13 +33,22 @@ pub async fn submit(
...
-        .await
-        .context("failed to get transaction_count")?;
     let settlement = encode_settlement(&settlement)?;
     // Check that a simulation of the transaction works before submitting it.
     simulate_settlement(&settlement, contract).await?;
 
+    let nonce = transaction_count(contract)
+        .await
+        .context("failed to get transaction_count")?;
+    let address = match &contract.defaults().from {
```
fleupold, line 43: nanonit: why not expect?

solver/src/settlement_submission.rs
fleupold: I think it's good that the method returns Result<Option> just not sure an error should shortcut and fail the runloop. We could log the error and continue assuming there is no pending transaction. But we can also do that if we see this happening.

fleupold (approved):
Looks good!

nlordell (approved):

solver/src/settlement_submission.rs
nlordell: I would also slightly prefer expect here.

solver/src/settlement_submission.rs
fleupold: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FQW4aFaN0GSLshdzEtA%2Fgiphy-downsized-medium.gif&token=metoo" data-gitmeme-token="metoo"></a>

solver/src/settlement_submission.rs
vkgnosis: added

solver/src/settlement_submission.rs
vkgnosis: The reason was that this is a reference but unwrap/exect consume so can't use them. Changed it to clone + expect.

vkgnosis:
Still have to test this with mainnet before merging.

vkgnosis:
Tested on mainnet by creating a low gas price transaction, restarting the binary and seeing that it gets discovered. Then I printed the gas price out to see that it first uses previous_gas_price * 12.5 and then goes back to the estimator. For the test I just did noop transactions to save on gas.

PR #473, opened 2021-04-14
Add seperate groups of interactions for settlement

nlordell:
This PR refactors the settlement interactions to use "interaction groups" to allow for smarter combinations. This allows, for example, to combine multiple unwraps into a single one without requiring separate implementations of the solver. Additionally, this avoids some implicit ordering that would otherwise be needed in all solvers (they would all require unwraps to come after AMM interactions for example).

/cc @fedgiac 

### Test Plan

Added new unit tests.

### DIFF
--- a/solver/src/interactions/weth.rs
+++ b/solver/src/interactions/weth.rs
@@ -1,6 +1,9 @@
-use crate::{encoding::EncodedInteraction, settlement::Interaction};
+use crate::{
+    encoding::EncodedInteraction,
+    settlement::{Interaction, UnwrapInteraction},
+};
 use contracts::WETH9;
-use primitive_types::U256;
+use primitive_types::{H160, U256};
 
 #[derive(Debug)]
 pub struct UnwrapWethInteraction {
@@ -13,6 +16,25 @@ impl Interaction for UnwrapWethInteraction {
         let method = self.weth.withdraw(self.amount);
         let calldata = method.tx.data.expect("no calldata").0;
         vec![(self.weth.address(), 0.into(), calldata)]
+    }
+}
+
+impl UnwrapInteraction for UnwrapWethInteraction {
+    fn token_address(&self) -> H160 {
+        self.weth.address()
+    }
+
+    fn amount(&self) -> U256 {
+        self.amount
+    }
+
+    fn increase_amount(&mut self, amount: U256) {
+        // We are OK with panicking on overflow in this situation as the the
+        // native wrapped token for all our supported networks is trusted.
+        self.amount = self
+            .amount
+            .checked_add(amount)
+            .expect("no one is that rich");
     }
 }
 
@@ -41,5 +63,15 @@ mod tests {
         let withdraw_signature = hex!("2e1a7d4d");
         assert_eq!(call[0..4], withdraw_signature);
         assert_eq!(U256::from_big_endian(&call[4..36]), amount);
+    }
+
+    #[test]
+    #[should_panic]
+    fn panics_when_overflow_on_increase_amount() {
+        let mut unwrap = UnwrapWethInteraction {
+            weth: WETH9::at(&dummy_web3::dummy_web3(), H160([0x42; 20])),
+            amount: U256::max_value(),
+        };
+        unwrap.increase_amount(1.into());
     }
 }
--- a/solver/src/settlement.rs
+++ b/solver/src/settlement.rs
@@ -1,10 +1,11 @@
 use crate::encoding::{self, EncodedInteraction, EncodedTrade};
-use anyhow::Result;
 use model::order::{Order, OrderKind};
 use num::{BigRational, Signed, Zero};
 use primitive_types::{H160, U256};
 use shared::conversions::U256Ext;
-use std::collections::HashMap;
+#[cfg(test)]
+use std::fmt::{self, Formatter};
+use std::{collections::HashMap, fmt::Debug};
 
 #[derive(Clone, Debug, Default, Eq, PartialEq)]
 pub struct Trade {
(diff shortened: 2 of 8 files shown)


fleupold (commented):
Will take a close look tomorrow. 

Just at a high level question: Have you considered using runtime inspection of the concrete interaction type to do the ordering in the end vs. exposing the different groups on the struct level? 
Now each solver still needs "know" that unwrap interactions are something different than other interactions (and as you said on the task this might be true for limit order weth unwrap, but not for cDAI DAI unwrap).

nlordell:
> Have you considered using runtime inspection of the concrete interaction type to do the ordering in the end vs. exposing the different groups on the struct level?

The issue I see with that would require leaking a bunch of `Interaction` implementation types to the `Settlement`.

> Now each solver still needs "know" that unwrap interactions are something different than other interactions

The idea is that `unwrap_interactions` are specific to `LimitOrder`s. Since `UnwrapWethInteraction: Interaction` as well, a solver can include one as part of its execution strategy (so it won't go at the end).

All in all, I'm still not 100% happy about this solution, since it still doesn't solve the case of "user buy ETH for DAI, solver emits AMM interaction that swaps DAI for ETH (UniV1 for example), solver emits a 'wrap ETH -> WETH' interaction, limit order emits a 'unwrap WETH -> ETH' interaction" case. For now, I thought this solution was OK since it:
- No longer has implicit ordering rules between `LimitOrder` interactions and `AMM` interactions
- Allows collapsing multiple unwraps into one
- Does not leak `Interaction` implementation type information to `Settlement`.

I also thought about it later, but `LimitOrder`s aren't always necessarily EOA orders, so I think the `settle` needs to **also** return a `Vec<Box<dyn Interaction>>` for those cases that come before the interactions (I do think this can be added back in later though, since we don't make use of it yet).

vkgnosis (approved):
This representation and your arguments make sense to me. My intuition is that this works better than runtime inspection of Interaction trait objects.

solver/src/interactions/weth.rs
```diff
@@ -16,6 +19,25 @@ impl Interaction for UnwrapWethInteraction {
...
+impl UnwrapInteraction for UnwrapWethInteraction {
+    fn token_address(&self) -> H160 {
+        self.weth.address()
+    }
+
+    fn amount(&self) -> U256 {
+        self.amount
+    }
+
+    fn increase_amount(&mut self, amount: U256) {
```
fleupold, line 31: nit: Maybe a "can_merge(&self, other: &UnwrapInteraction)" logic would make sense which could check that the token addresses are the same and could also skip merging if it was to overflow?

solver/src/settlement.rs
```diff
@@ -59,22 +60,105 @@ impl Trade {
...
+impl Debug for MockUnwrapInteraction {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        f.write_str("MockUnwrapInteraction")
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct SettlementInteractions {
+    amm_interactions: Vec<Box<dyn Interaction>>,
+    unwrap_interactions: HashMap<H160, Box<dyn UnwrapInteraction>>,
```
fleupold, line 108: I find these name quite specific (it's true now but rather coincidental to the reason for these fields). Maybe call them `execution_plan`, `pre_transfer_interactions` or something like that?

solver/src/settlement.rs
```diff
@@ -59,22 +60,105 @@ impl Trade {
...
 }
 
+#[cfg(test)]
+impl Debug for MockInteraction {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        f.write_str("MockInteraction")
+    }
+}
+
+pub trait UnwrapInteraction: Interaction {
```
fleupold, line 76: Is this abstraction really useful if each solver needs to know the specific Interaction type a specific type of liquidity returns? Maybe in this case the LimtiOrder type should return the concrete UnwrapWethInteraction. How do you see it be reused?

fleupold (commented):
Change looks good, however I'm still not super happy about the solver liquidity coupling (this might more generally question the conclusions from: https://github.com/gnosis/gp-v2-services/issues/194)

Another suggestion on how knowledge in each solver about the different "phases of interactions" could be avoided would be to give each interaction an "execution group" (could be an enum or a number exposed on the top level `Interaction` trait) and then implement sorting based on this. 

This way the LimitOrderWethUnwrapInteraction could have a different group than e.g. DAI -> cDAI unwrapping, merging could still be implemented for each group (this might require introspection) and priority would be handled by the liquidity object (rather than by the solvers).

nlordell:
> however I'm still not super happy about the solver liquidity coupling

I agree. Settlement building changes needed to be implemented in multiple places for this change which is a bit unfortunate.

solver/src/settlement.rs
nlordell: Yes, maybe it makes sense for limit orders to not return a boxed trait but a concrete type instead.

solver/src/interactions/weth.rs
nlordell: That's a really nice suggestion. The reason I had it like this was to use a token hash-map for unwrapping, but searching through a vector linearly would be fine as well given the expected size, and clean up the abstraction as well.

fleupold:
> where each solver needs to implement its Settlement building code which is largely the same

I was envisioning the sorting/merging to be a post-processing step which could happen in the driver. As far as the solvers are concerned they just add their abstract interactions to the vector (which gets sorted, each stage merged, later on in common code).

nlordell:
> As far as the solvers are concerned they just add their abstract interactions to the vector

Their also responsible for converting `LimitOrder -> (Trade, Vec<Interaction>)` which I don't think should be part of each solver either. Furthermore, this would also not address the issue of implicit ordering. While, with this decision, the ordering would not be important, a future refactor might make it so, coming back to the issue of "the compiler doesn't tell me where I need to change code when I refactor". Additionally, some interactions need strict ordering and others don't (execution plan vs unwrapping).

nlordell:
Closed in favour of #483

PR #479, opened 2021-04-19
Refactor settlement generation in solvers

nlordell:
This unfortunately large PR refactors how settlements are built in the solver implementations. Specifically, there is an inversion of control, where liquidity doesn't return things (trade and interactions) to be added to the settlement, but instead gets passed a `SettlementEncoder` and adds things directly to it. This is nice as:
- We can have a single `SettlementHandling` trait used for all liquidity
- Extending the settlement intermediate representation will not require changes to existing liquidity
- Solvers are now completely agnostic to how liquidity encodes itself in the solution, both `LimitOrder`s and `AmmOrder`s get added with the same `with_liquidity` method.

Furthermore, this opens the door for implementing a "settlement optimizer" which would take the intermediate `SettlementEncoder` representation and spit out an optimized solution (by, for example, collapsing interactions that can be collapsed).

This is based off of a discussion with @fleupold where we decided that having a "settlement builder" type interface would be beneficial and allow keeping solvers completely how liquidity actually get included into the settlement, only caring about how the liquidity is modeled.

Suggestions open to suggestions on how to make this smaller. Unfortunately, the `Settlement` struct and liquidity gets encoded into it was used in a lot of source and tests.

### Test Plan

CI still passes. This PR is a refactor, which modifies existing traits and how they are called.

### DIFF
--- a/solver/src/liquidity.rs
+++ b/solver/src/liquidity.rs
@@ -19,20 +19,21 @@ pub enum Liquidity {
     Amm(AmmOrder),
 }
 
-/// A trait associating some liquidity model to how it is executed. This allows
+/// A trait associating some liquidity model to how it is executed and encoded
+/// in a settlement (through a `SettlementHandling` reference). This allows
 /// different liquidity types to be modeled the same way.
-pub trait LiquidityModel {
+pub trait Settleable {
     type Execution;
 
     fn settlement_handling(&self) -> &dyn SettlementHandling<Self>;
 }
 
 /// Specifies how a liquidity exectution gets encoded into a settlement.
-pub trait SettlementHandling<Model>: Send + Sync
+pub trait SettlementHandling<L>: Send + Sync
 where
-    Model: LiquidityModel,
+    L: Settleable,
 {
-    fn encode(&self, execution: Model::Execution, encoder: &mut SettlementEncoder) -> Result<()>;
+    fn encode(&self, execution: L::Execution, encoder: &mut SettlementEncoder) -> Result<()>;
 }
 
 /// Basic limit sell and buy orders
@@ -59,7 +60,7 @@ impl LimitOrder {
     }
 }
 
-impl LiquidityModel for LimitOrder {
+impl Settleable for LimitOrder {
     type Execution = U256;
 
     fn settlement_handling(&self) -> &dyn SettlementHandling<Self> {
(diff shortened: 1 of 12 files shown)


solver/src/liquidity.rs
```diff
@@ -21,6 +18,23 @@ pub enum Liquidity {
...
+
+    fn settlement_handling(&self) -> &dyn SettlementHandling<Self>;
+}
+
+/// Specifies how a liquidity exectution gets encoded into a settlement.
+pub trait SettlementHandling<Model>: Send + Sync
+where
+    Model: LiquidityModel,
+{
+    fn encode(&self, execution: Model::Execution, encoder: &mut SettlementEncoder) -> Result<()>;
```
fleupold, line 35: I really like this extra layer of indirection. While it might be a little harder to understand, it really decouples liquidity from solvers and as such is a great step towards allowing independent liquidity/solver development.

fleupold (approved):
Impressive refactoring considering you are "new" to the codebase.

🎩

solver/src/settlement.rs
```diff
@@ -67,56 +62,70 @@ pub trait Interaction: std::fmt::Debug + Send {
...
+impl SettlementEncoder {
+    /// Creates a new settlement encoder with the specified prices.
+    ///
+    /// The prices must be provided up front in order to ensure that all tokens
+    /// included in the settlement are known when encoding trades.
+    fn new(clearing_prices: HashMap<H160, U256>) -> Self {
+        // Explicitely define a token ordering based on the supplied clearing
+        // prices. This is done since `HashMap::keys` returns an iterator in
+        // arbitrary order ([1]), meaning that we can't rely that the ordering
+        // will be consistent across calls. The list is sorted so that
```
vkgnosis, line 90: Alternatively we could use BTreeMap which is always sorted by key. But I see below we wouldn't be able to efficiently find the index (position) of a key with just the BTreeMap api.

vkgnosis (approved):

solver/src/liquidity.rs
nlordell: The other thing I like about it, is we can now add a `unwrap_native_token` to the `SettlementEncoder` and only modify the `impl SettlementHandling<LimitOrder> for Order` implementation (since we don't need to extend the trait, but only the implementation of where it applies). In theory, if there were other liquidity that were modeled as `LimitOrder`s, they would also not need to be changed :smile:.

solver/src/settlement.rs
nlordell: Since we create the vector once and sort it, I find it easier to just use a `Vec` since it has the same runtime characteristics. I think, that if we would insert into the list while building, then a binary tree might be nice.

PR #480, opened 2021-04-20
Re-enable gas price from pending transactions with fix

vkgnosis:
This PR contains the original reverted commit and then  a second commit on top to fix the issue on rinkeby:

> // This works better than rust-web3's `block_with_txs` because some nodes set the `miner` field to
> // `null`. This does not follow the ethrpc specification and fails deserialization in rust-web3.

https://eth.wiki/json-rpc/API#returns-27

For the staging pods this happens on the rinkeby node but not on the mainnet node (didn't test xdai).

### Test Plan
Added test

### DIFF
--- a/solver/src/lib.rs
+++ b/solver/src/lib.rs
@@ -6,6 +6,7 @@ pub mod liquidity;
 pub mod metrics;
 pub mod naive_solver;
 pub mod orderbook;
+pub mod pending_transactions;
 pub mod settlement;
 pub mod settlement_submission;
 pub mod solver;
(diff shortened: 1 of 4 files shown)


fleupold (approved):
This is fine, however given that it's Infura that is behaving weird, it might be nice to make the Web3 code a little bit more forgiving (it seems to already handle blocks where the miner field is not set at all).

Maybe this is something we can push upstream as it would affect any project that wants to do proper tx management based on pending blocks?

vkgnosis:
Correct, it's only `miner` explicitly being set to null that's the problem. Will make an issue/pr for it on rust-web3.

PR #481, opened 2021-04-20
Preparing Uniswap Resources to be interchangable with Sushiswap

bh2smith:
**Phase I of Sushiswap integration**

Together with @fleupold, we introduce a few generalizations to the way `factory` and `router` are shared among orderbook, solver and pool fetching logic. In brief we include a

 -  `UniswapLikeRouter` ABI (interface) to act as a generalized router
 - new struct `UniswapResource` which is responsible for the `factory`, `chain_id` and ultimately construction of token `pair_address`
 - PoolFetcher now operates on the `AmmResource` rather than `factory` and `chain_id` separately

### Test Plan
Existing CI should still pass given that no new logic was introduced, there are no new tests.

### DIFF
--- a/e2e/tests/uniswap_trade_test.rs
+++ b/e2e/tests/uniswap_trade_test.rs
@@ -14,12 +14,11 @@ use orderbook::{
 };
 use secp256k1::SecretKey;
 use serde_json::json;
-use shared::uniswap_pool::UniswapResource;
 use shared::{
     current_block::current_block_stream,
     price_estimate::UniswapPriceEstimator,
     transport::LoggingTransport,
-    uniswap_pool::{CachedPoolFetcher, PoolFetcher},
+    uniswap_pool::{CachedPoolFetcher, PoolFetcher, UniswapPairProvider},
     Web3,
 };
 use solver::{liquidity::uniswap::UniswapLiquidity, metrics::NoopMetrics, orderbook::OrderBookApi};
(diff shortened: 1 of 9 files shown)


e2e/tests/uniswap_trade_test.rs
```diff
@@ -224,13 +228,16 @@ async fn test_with_ganache() {
     assert_eq!(placement.unwrap().status(), 201);
 
     // Drive solution
+    let uniswap_resource = Arc::new(UniswapResource {
```
fleupold, line 231: nit: the comment above seems off.

shared/src/uniswap_pool.rs
```diff
@@ -15,11 +15,29 @@ use crate::current_block::{Block, CurrentBlockStream};
 
 const UNISWAP_PAIR_INIT_CODE: [u8; 32] =
     hex!("96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f");
-
 const HONEYSWAP_PAIR_INIT_CODE: [u8; 32] =
     hex!("3f88503e8580ab941773b59034fb4b2a63e86dbc031b3633a925533ad3ed2b93");
 const MAX_BATCH_SIZE: usize = 100;
 
+pub trait AmmResource: Send + Sync + 'static {
```
fleupold, line 22: Not sure if Resource is a good name here. All it really does no is give us pair addresses. Maybe we can call it AmmPairProvider?

solver/src/interactions/uniswap.rs
```diff
@@ -27,32 +27,34 @@ impl Interaction for UniswapInteraction {
...
         );
         let calldata = method.tx.data.expect("no calldata").0;
-        (self.contract.address(), 0.into(), calldata)
+        (self.router.address(), 0.into(), calldata)
     }
 
     fn web3(&self) -> web3::Web3<ethcontract::transport::DynTransport> {
-        self.contract.raw_instance().web3()
+        // TODO - This is a cheeky way of fetching web3
+        self.settlement.raw_instance().web3()
```
vkgnosis, line 49: It's fine imo. Can keep it this way.

vkgnosis (approved):

solver/src/interactions/uniswap.rs
bh2smith: Should I remove the TODO?

bh2smith:
```suggestion
        self.settlement.raw_instance().web3()
```

vkgnosis: If you agree that it's fine I'd say remove it.

PR #483, opened 2021-04-20
Add support for native token unwrapping to settlement builder

nlordell:
Supersedes #473

This PR re-implements the aforementioned PR using with the new `SettlementEncoder` abstraction, adding an additional `unwraps` field for keeping track of native token unwrapping.

### Test Plan

Added unit tests for new code. Note that code to add these to the encoder are still missing from EOA orders. I believe @fedgiac would be adding that next :smile:.

### DIFF
--- a/solver/src/interactions/weth.rs
+++ b/solver/src/interactions/weth.rs
@@ -1,5 +1,5 @@
 use crate::{encoding::EncodedInteraction, settlement::Interaction};
-use anyhow::{anyhow, Context as _, Result};
+use anyhow::{ensure, Result};
 use contracts::WETH9;
 use primitive_types::U256;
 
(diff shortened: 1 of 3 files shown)


solver/src/interactions/weth.rs
```diff
@@ -1,13 +1,33 @@
...
 pub struct UnwrapWethInteraction {
     pub weth: WETH9,
     pub amount: U256,
 }
 
+impl UnwrapWethInteraction {
+    /// Tries to merge the specified unwrap with the current one, returning
+    /// `true` if the merge was successful, and `false` otherwise.
+    ///
+    /// Returns an error on arithmetic overflow.
```
fleupold, line 16: nanonit: I understand arithmetic overflows are practically impossible. However instead of handling the error case, we could just return false to encode it as its own transaction?

fleupold (approved):
Nice! I like this way of integrating much better!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2Fl3q2XhfQ8oCkm1Ts4%2Fgiphy.gif&token=applause" data-gitmeme-token="applause"></a>

solver/src/interactions/weth.rs
nlordell: > I understand arithmetic overflows are practically impossible

I think they should be impossible on the networks we support given the native currency ERC20 token implementations (WETH and wxDAI). They could happen on test-networks or unsupported configurations.

I just thought that since `SettlementHandling` already returns a `Result` that I might as well propagate the arithmetic overflow. Alternatively we can `panic` as a way to say "we really don't expect this to happen ever" so that we get a louder alert so we can 

Personally, I would prefer panicking over treating it in the same way we do if the `merge` was not possible (i.e. the token address was different).

WDYT?

bh2smith (approved):
Bravo and Welcome back to dex-services!

solver/src/interactions/weth.rs
fleupold: Panicing sounds ok to me in that case (I'm fine either way though)

vkgnosis (approved):

PR #485, opened 2021-04-21
Ignore existing event when inserting to fix primary key violation

### LINKED ISSUE #444: Make orderbook maintenance HPA compatible

vkgnosis:
Fixes #444 

### Test Plan
new unit test

### DIFF
--- a/orderbook/src/database/events.rs
+++ b/orderbook/src/database/events.rs
@@ -137,8 +137,12 @@ async fn insert_invalidation(
     index: &EventIndex,
     event: &Invalidation,
 ) -> Result<(), sqlx::Error> {
+    // We use ON CONFLICT so that multiple updates running at the same do not error because of
+    // events already existing. This can happen when multiple orderbook apis run in HPA.
+    // See #444 .
     const QUERY: &str =
-        "INSERT INTO invalidations (block_number, log_index, order_uid) VALUES ($1, $2, $3);";
+        "INSERT INTO invalidations (block_number, log_index, order_uid) VALUES ($1, $2, $3) \
+         ON CONFLICT DO NOTHING;";
     transaction
         .execute(
             sqlx::query(QUERY)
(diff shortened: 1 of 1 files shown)


bh2smith (approved):
Seems like a cheeky fix but should work without a doubt. Did we ever determine where the conflicts were coming from?

nlordell (approved):
Code looks fine. Same question as :point_up:.

vkgnosis:
See the issue that I linked. Two HPA pods run at same time.

orderbook/src/database/events.rs
```diff
@@ -138,7 +138,8 @@ async fn insert_invalidation(
     event: &Invalidation,
 ) -> Result<(), sqlx::Error> {
     const QUERY: &str =
-        "INSERT INTO invalidations (block_number, log_index, order_uid) VALUES ($1, $2, $3);";
+        "INSERT INTO invalidations (block_number, log_index, order_uid) VALUES ($1, $2, $3) \
```
fleupold, line 141: Can we add a comment on why we do the on conflict do nothing here and below (maybe just a link to the issue).

fleupold (approved):
Alternative would be to acquire a lock early in the transaction (before doing the delete). This might be slightly better to avoid more complex race conditions (e.g. non repeatable reads, etc), but for now this seems like a simple and good enough fix.

orderbook/src/database/events.rs
vkgnosis: added

PR #487, opened 2021-04-21
Chose AmmOrder with deepest liquidity

fleupold:
This PR is in preparation of adding SushiSwap liquidity, which can lead to having more than 1 AMM order per token pair. At the moment one of them would be chosen rather arbitrarily (based on hashing/ordering in the liqudity list).

This PR makes it so that we chose the most liquid pool. While this is not necessarily the best strategy (e.g. a slightly less liquid pool could have a "better spot" price at the moment) it seems like a good heuristic (trusting in market efficiency to arbitrage away large differences in prices).

Regardless, we should adopt the Naive Solver and baseline solver to deal with a list of liquidity on the same pair.

### Test Plan
Added a unit test for the filtering logic.

### DIFF
--- a/solver/src/naive_solver.rs
+++ b/solver/src/naive_solver.rs
@@ -19,15 +19,14 @@ impl Solver for NaiveSolver {
         liquidity: Vec<Liquidity>,
         _gas_price: f64,
     ) -> Result<Option<Settlement>> {
-        let mut limit_orders = Vec::new();
         let uniswaps = extract_deepest_amm_liquidity(&liquidity);
-        for liquidity in liquidity {
-            match liquidity {
-                Liquidity::Limit(order) => limit_orders.push(order),
-                Liquidity::Amm(_) => continue,
-            }
-        }
-        Ok(settle(limit_orders.into_iter(), uniswaps).await)
+        let limit_orders = liquidity
+            .into_iter()
+            .filter_map(|liquidity| match liquidity {
+                Liquidity::Limit(order) => Some(order),
+                _ => None,
+            });
+        Ok(settle(limit_orders, uniswaps).await)
     }
 }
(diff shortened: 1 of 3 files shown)


solver/src/naive_solver.rs
```diff
@@ -19,17 +20,11 @@ impl Solver for NaiveSolver {
         _gas_price: f64,
     ) -> Result<Option<Settlement>> {
         let mut limit_orders = Vec::new();
-        let mut uniswaps = HashMap::new();
+        let uniswaps = extract_deepest_amm_liquidity(&liquidity);
```
bh2smith, line 23: Nit: Maybe swap these two lines so that limit_orders are declared and populated together without uniswaps in between

bh2smith (approved):
Fantastic!

solver/src/naive_solver.rs
```diff
@@ -19,17 +20,11 @@ impl Solver for NaiveSolver {
...
-                Liquidity::Amm(uniswap) => {
-                    let pair = uniswap.tokens;
-                    uniswaps.insert(
-                        TokenPair::new(pair.get().0, pair.get().1).expect("Invalid Pair"),
-                        uniswap,
-                    );
-                }
+                Liquidity::Amm(_) => continue,
             }
         }
```
nlordell, line 29: ubernit: Now that this code is purely for "extracting limit orders from liquidity", maybe something like:
```
let limit_orders = liquidity.iter.filter_map(|liquidity| match liquidity {
    Liquidity::Limit(order) => Some(order),
    _ => None,
}).collect();
```

nlordell (approved):
Very nice!

vkgnosis (approved):

PR #488, opened 2021-04-21
Refactor liquidity collection into its own component

fleupold:
This PR moves the code of collecting the `Vec<Liquidity>` from our different sources from the driver into its own component. This is to not overload the number of dependencies we pass into the driver further (we already need to allow clippy's `too_many_arguments` warning)

We left the "price estimation filtering" logic in the driver, as we didn't want to pass a price estimator into the liquidity collector.

### Test Plan
e2e tests and CI (no logic change).

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -1,14 +1,13 @@
-use crate::{chain, liquidity_collector::LiquidityCollector, metrics::SolverMetrics};
 use crate::{liquidity::Liquidity, settlement::Settlement, settlement_submission, solver::Solver};
+use crate::{liquidity_collector::LiquidityCollector, metrics::SolverMetrics};
 use anyhow::{Context, Result};
 use contracts::GPv2Settlement;
 use futures::future::join_all;
 use gas_estimation::GasPriceEstimating;
-use itertools::Itertools;
 use num::BigRational;
 use primitive_types::H160;
 use shared::price_estimate::PriceEstimating;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::{
     cmp::Reverse,
     sync::Arc,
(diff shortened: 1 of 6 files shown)


solver/src/driver.rs
```diff
@@ -80,13 +71,21 @@ impl Driver {
...
     async fn collect_estimated_prices(
         &self,
-        limit_orders: &[LimitOrder],
+        liquidity: &[Liquidity],
     ) -> HashMap<H160, BigRational> {
         // Computes set of traded tokens (limit orders only).
-        let tokens: Vec<H160> = limit_orders
+        let tokens: Vec<H160> = liquidity
             .iter()
             // .flat_map(|lo| vec![lo.sell_token, lo.buy_token].into_iter())
```
bh2smith, line 79: Probably don't need this anymore. Unless its a hint at whats going on below in "plain rust"

bh2smith (approved):
Very nice! Looking forward to building on this.

solver/src/driver.rs
```diff
@@ -80,13 +71,21 @@ impl Driver {
...
-            .flat_map(|lo| chain![lo.sell_token, lo.buy_token])
+            .flat_map(|liquidity| {
+                let iter: Box<dyn Iterator<Item = H160>> = match liquidity {
+                    Liquidity::Limit(limit_order) => {
+                        Box::new(chain![limit_order.sell_token, limit_order.buy_token])
+                    }
+                    _ => Box::new(std::iter::empty()),
```
vkgnosis, line 85: This is starting to look weird, maybe better to go the imperative route:

```rust
        let mut tokens = HashSet::new();
        for liquid in liquidity {
            if let Liquidity::Limit(limit_order) = liquid {
                tokens.insert(limit_order.sell_token);
                tokens.insert(limit_order.buy_token);
            }
        }
        let tokens: Vec<_> = tokens.drain().collect();
```

vkgnosis (approved):

PR #489, opened 2021-04-21
Use Sushiswap AMM

bh2smith:
Introducing Sushiswap as a valid and usable liquidity source for the solver.

### Test Plan
Existing CI.

1. Run the solver, orderbook and swap-ui locally and place an order selling UNI for SUSHI.
2. Observe that we use uniswap to trade UNI<>ETH and then Sushiswap to trade ETH<>SUSHI

https://etherscan.io/tx/0xdd6220190d91240eb61eda5f07e4b966fca32aa18deb9137011c6d60cb2c7674


solver/src/liquidity_collector.rs
```diff
@@ -6,6 +6,7 @@ use crate::{
 
 pub struct LiquidityCollector {
     pub uniswap_liquidity: UniswapLikeLiquidity,
+    pub sushiswap_liquidity: UniswapLikeLiquidity,
```
fleupold, line 9: just thinking out loud that we could also have a list of `UniswapLikeLiquidity` objects (in case we don't e.g. create SushiSwap liquidity for the e2e test or it isn't available on a certain network). This would also allow us to configure the UniswapLikeLiquidity which is something I'd prefer (in case we run into issues with the Uni/Sushi mix and we want to turn off one source). But this could also be achieved with an optional field.

fleupold (commented):
I don't necessarily think adding this to the e2e is necessary (unless we plan to do something similar for all types of liquidity we are adding).

However, I would prefer if we could configure the source of our UniswapLikeLiquidity so we could turn it off in case something goes wrong without making a code change.

solver/src/liquidity_collector.rs
bh2smith: Right. I had included this as an option on the PR description. Definitely planning to change this.

bh2smith:
Thanks for your review Felix. I will make the suggested change and not include it in the e2e test and have something ready by tomorrow morning.

solver/src/liquidity_collector.rs
```diff
@@ -0,0 +1,38 @@
...
+
+        let mut amms = vec![];
+        for liquidity in self.uniswap_like_liquidity.iter() {
+            amms.extend(
+                liquidity
+                    .get_liquidity(limit_orders.iter())
+                    .await
+                    .context("failed to get pool")?,
+            );
+        }
```
bh2smith, line 29: I tried to use `flat_map` here, but ran into some issue with the async calls inside. @vkgnosis any tricks up your sleeve that could make this more rustic?

nlordell: I think you might have to convert the iterator into a `Stream`.

bh2smith: Thanks I'll give this a try.

nlordell: The [`StreamExt::flat_map` example](https://docs.rs/futures/0.3.14/futures/stream/trait.StreamExt.html#examples-13) is almost exactly what you want.

contracts/src/bin/deploy.rs
```diff
@@ -63,6 +63,13 @@ async fn run() -> Result<()> {
     let uniswap_factory = deploy!(UniswapV2Factory(accounts[0]));
     deploy!(UniswapV2Router02(uniswap_factory.address(), weth.address()));
 
+    log::info!("deploying Sushiswap contracts");
```
fleupold, line 66: Is this still needed if we are not using them in the e2e tests?

solver/src/main.rs
```diff
@@ -173,7 +192,7 @@ async fn main() {
         price_estimator.clone(),
     );
     let liquidity_collector = LiquidityCollector {
-        uniswap_liquidity,
+        uniswap_like_liquidity: vec![uniswap_liquidity, sushiswap_liquidity],
```
fleupold, line 195: Can we decide the contents of this vector based on a config value for now? This way we can turn off specific liquidity without a code change.

solver/src/liquidity_collector.rs
bh2smith: That's even better.

vkgnosis: Nothing wrong with the current imperative code either. Feel free to use which one you like more.

fleupold: `StreamExt::flat_map` looks nice. The way I read the docs however, it's not actually parallelising the inner stream. 

> The provided closure which produce inner streams is executed over all elements of stream as last inner stream is terminated and next stream item is available.

But it's not super important here and might be easier to read than what we do in the [PoolFetcher](https://github.com/gnosis/oba-services/blob/089d91775bfa64043dddcc6e3355d44cb769b757/shared/src/pool_fetching.rs#L249-L270) to achieve concurrency of the inner futures.

fleupold (commented):
Looks good, just missing the ability to configure via command line/env var

solver/src/liquidity_collector.rs
nlordell: I actually personally _ever so slightly_ prefer the current implementation (without `stream::iter`) as I find the `for` loop to be more readable in this case.

Either way, if you prefer using `Stream` and `flat_map` go for it.

bh2smith: Nope. I'm gonna leave it as is. Much easier for me to read also.

bh2smith:
Awesome. I have made the amm sources configurable, but had to write a little helper method `build_amm_artifacts' which is a little bit lopsided at the moment since we exclusively use uniswap for price estimation. There is a todo in the code to fix this.

solver/src/main.rs
```diff
@@ -192,3 +197,59 @@ async fn main() {
...
     driver.run_forever().await;
 }
+
+async fn build_amm_artifacts(
+    sources: Vec<AmmSources>,
+    chain_id: u64,
+    settlement_contract: contracts::GPv2Settlement,
+    base_tokens: HashSet<H160>,
+    web3: web3::Web3<LoggingTransport<Http>>,
+) -> (PoolFetcher, Vec<UniswapLikeLiquidity>) {
```
bh2smith, line 207: This method is a bit ugly at the moment because it is lopsided as the poolfetcher is constructed via the uniswap pair provider... Open to other suggestions of course.

solver/src/main.rs
```diff
@@ -192,3 +197,59 @@ async fn main() {
...
+        .await
+        .expect("couldn't load deployed uniswap router");
+    let uniswap_factory = contracts::UniswapV2Factory::deployed(&web3)
+        .await
+        .expect("couldn't load deployed uniswap router");
+    let uniswap_pair_provider = Arc::new(UniswapPairProvider {
+        factory: uniswap_factory,
+        chain_id,
+    });
+    let pool_fetcher = PoolFetcher {
```
fleupold, line 221: Maybe you can create the pool fetcher outside of this method. I don't think it's a problem to instantiate two pair providers in separate places.

fleupold (approved):
Looks good!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FHw8vYF4DNRCKY%2Fgiphy.gif&token=ship_it" data-gitmeme-token="ship_it"></a>

vkgnosis (approved):

nlordell (approved):

PR #491, opened 2021-04-21
Make BaselineSolver pool agnostic

fleupold:
This change is a bit of refactoring which can benefit both the native Balancer as well as the Sushiswap integration. My suggestion is to have the baseline solver operate on an abstract trait which only knows how much in/out amount for a given "piece of liquidity" can be expected. This can then be implemented e.g. by Balancer pools (which likely behave different to the current `Pool` object, the latter should probably be renamed to `UniswapLikePool` when we get there).

Later on other liquidity sources such as Curve or even market makers could implement `BaselineSolvable`.

The next step would be to make the code in the shared baseline_solver module actually operate on a `Vec<impl BaselineSolvable>`.

### Test Plan
Existing tests, as this is just refactoring things.

### DIFF
--- a/shared/src/baseline_solver.rs
+++ b/shared/src/baseline_solver.rs
@@ -1,18 +1,31 @@
+//! Module containing basic path-finding logic to get quotes/routes for the best onchain liquidity.
+
 use ethcontract::{H160, U256};
 use model::TokenPair;
 use num::BigRational;
 use std::collections::{HashMap, HashSet};
 
-use crate::pool_fetching::Pool;
-
 type PathCandidate = Vec<H160>;
+
+pub trait BaselineSolvable {
+    // Given the desired output token, the amount and token input, return the expected amount of output token.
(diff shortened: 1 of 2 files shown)


bh2smith (approved):
Awesome!

nlordell (approved):
I really like this abstraction.

We can also add a:
```rust
pub struct BestPriceBaselineSolvable(Vec<Box<dyn BaselineSolvable>>);
```

When we want it to take the "best price".

fleupold:
> We can also add a:
`pub struct BestPriceBaselineSolvable(Vec<Box<dyn BaselineSolvable>>);`
When we want it to take the "best price".

This is a nice suggestion! I think in this case the BaselineSolvable would have to return amount and `impl BaselineSolvable` so that the solver in the end know which part of its liquidity was used?

nlordell:
> that the solver in the end know which part of its liquidity was used

Yes.

PR #495, opened 2021-04-22
Refactor driver module

vkgnosis:
The single_run function was very long so I split it up into parts.

The raw diff is a little confusing. Makes more sense side by side looking at the new `single_run`.

### Test Plan
e2e test still works

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -1,10 +1,11 @@
 use crate::{
     liquidity::{offchain_orderbook::BUY_ETH_ADDRESS, Liquidity},
+    liquidity_collector::LiquidityCollector,
+    metrics::SolverMetrics,
     settlement::Settlement,
     settlement_submission,
     solver::Solver,
 };
-use crate::{liquidity_collector::LiquidityCollector, metrics::SolverMetrics};
 use anyhow::{Context, Result};
 use contracts::GPv2Settlement;
 use futures::future::join_all;
@@ -12,9 +13,8 @@ use gas_estimation::GasPriceEstimating;
 use num::BigRational;
 use primitive_types::H160;
 use shared::price_estimate::PriceEstimating;
-use std::collections::{HashMap, HashSet};
 use std::{
-    cmp::Reverse,
+    collections::{HashMap, HashSet},
     sync::Arc,
     time::{Duration, Instant},
 };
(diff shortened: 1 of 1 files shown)


solver/src/driver.rs
```diff
@@ -104,91 +105,56 @@ impl Driver {
...
                 Err(err) => {
-                    tracing::error!("solver {} error: {:?}", solver, err);
-                    None
+                    tracing::error!("solver {} error: {:?}", self.solver[index].to_string(), err);
+                    return None;
                 }
+            }?;
+            if settlement.trades().is_empty() {
+                return None;
+            }
```
vkgnosis, line 136: This is a small logic change because we were previously doing this in the settlement transaction submission loop below and printing "skipping empty settlement". I figured this makes more sense as we are also filtering `None` out here and there is no reason to log this happening.

solver/src/driver.rs
```diff
@@ -197,29 +163,94 @@ impl Driver {
...
-                settlement,
+                settlement.settlement,
             )
             .await
             {
                 Ok(_) => {
-                    // TODO: order validity checks
-                    // Decide what is handled by orderbook service and what by us.
-                    // We likely want to at least mark orders we know we have settled so that we don't
-                    // attempt to settle them again when they are still in the orderbook.
```
vkgnosis, line 214: Removed comment because it's not relevant anymore or will only be relevant when do speculative settlements (relying on previous going through) in the future.

solver/src/driver.rs
```diff
@@ -197,29 +163,94 @@ impl Driver {
...
+            .await;
+        // Sort by key in descending order. Reversed by doing b.cmp(a) instead of a.cmp(b).
+        settlements.sort_unstable_by(|a, b| b.objective_value.cmp(&a.objective_value));
+
+        for settlement in settlements.iter() {
+            tracing::info!(
+                "solver {} found solution with objective value {}: {:?}",
+                self.solver[settlement.index],
+                settlement.objective_value,
+                settlement.settlement,
```
vkgnosis, line 217: This is now a single log entry that prints both the objective value and the settlement contents.

solver/src/driver.rs
```diff
@@ -104,91 +105,57 @@ impl Driver {
...
             // of wants more likely.
-            let should_be_settled_immediately = settlement
-                .trades()
-                .iter()
-                .any(|trade| trade.order.order_meta_data.creation_date <= settle_orders_older_than);
+            let should_be_settled_immediately =
+                settlement.settlement.trades().iter().any(|trade| {
+                    trade.order.order_meta_data.creation_date <= settle_orders_older_than
+                });
             if !should_be_settled_immediately {
```
nlordell, line 159: Maybe all settlement filtering should happen in one place? Right now we filter:
- settlements with no trades
- settlements with no "old enough" orders

in two separate places. I think, it would be nice to have a `should_settle` function (with a better name :stuck_out_tongue:) that does this.

nlordell: Also, for example, there is no need to compute the objective value of a settlement with "too recent" orders.

nlordell (approved):
Smaller functions are nice and easier to follow!

Just one minor nit about whether or not a settlement is filtered is done in separate places and could be combined into a single block or function.

solver/src/driver.rs
vkgnosis: One reason for this is that this check is time dependent so it has to come after anything async like running the solver. But yes this part could be improved more.

solver/src/driver.rs
```diff
@@ -104,91 +105,57 @@ impl Driver {
...
+        prices: &HashMap<H160, BigRational>,
+    ) -> Vec<SolverSettlement> {
+        join_all(self.solver.iter().enumerate().map(|(index, solver)| {
+            let liquidity = liquidity.clone();
+            let metrics = &self.metrics;
+            async move {
+                let start_time = Instant::now();
+                let settlement = solver.solve(liquidity, gas_price).await;
+                metrics.settlement_computed(solver.to_string().as_str(), start_time);
+                (index, settlement)
```
fleupold, line 121: Why do we pass the index around instead of a reference to the solver itself?

solver/src/driver.rs
```diff
@@ -197,29 +164,94 @@ impl Driver {
...
+        }
+
+        self.submit_settlements(settlements).await;
+
         Ok(())
     }
 }
+
+// Filter limit orders for which we don't have price estimates as they cannot be considered for the objective criterion
+fn liquidity_with_price(
```
fleupold, line 229: this free method could be nicely unit tested now.

fleupold (approved):
Very nice cleanup. Breaking the driver up into smaller pieces is a good idea!

solver/src/driver.rs
vkgnosis: If we used references then the SolverSettlement struct would have to have a lifetime and then result of this function would be tied to &'a Self which means you wouldn't be able to call other &mut self methods. Overall it felt easier to use an index.

solver/src/driver.rs
vkgnosis: Agree. Won't add in this PR but definitely good to have.

PR #496, opened 2021-04-22
Enable native eth orders

### LINKED ISSUE #461: Allow buying ETH
The smart contract sets the buy token address to `0xeeee...eee` to indicate the user wants to buy the native token (ETH on mainnent, DAI on xDAI).

The solver currently doesn't support this magic number and treats it as a regular address (for which any trade will likely fail as we cannot find any uniswap base liquidity). This task captures adding support for it in both the API (for fee/amount estimations - for now it might be enough to just handle it as WETH and ignore the extra gas cost of unwrapping) as well as in the driver (adding an "unwrap" interaction to the execution).

[`LimitOrderSettlementHandling`](https://github.com/gnosis/oba-services/blob/7cbe3c376b08a2658287d26ed2dda002489c2509/solver/src/liquidity.rs#L38) already allows returning a list of interactions ` when settle is invoked (method to get the smart contract relevant trade information). We can add an interaction for unwrap in https://github.com/gnosis/oba-services/blob/7cbe3c376b08a2658287d26ed2dda002489c2509/solver/src/liquidity/offchain_orderbook.rs#L44

Currently the interactions just get appended to the settlement interaction list. This can cause issues with the execution plan if e.g. we try to unrwap before we have gotten the WETH from the trade (assuming there is not WETH buffer in the smart contract).

We could try to order interactions in a post processing step in the driver and put unrwap interactions at the end of the list ( + potentially combining them so we only invoke `WETH.unwrap` once in case multiple people trade in the same batch).

fedgiac:
Closes #461.

Last of a series of PRs that enables users buying native eth.
This PR adds a new unwrap interaction to the settlement for every order that buys native eth.

/cc @nlordell 

### Test Plan
New unit tests check that the settlement is working as intended.
Note that the overall logic (flag address, computing executed buy amount) is untested, this is left to a future e2e test.


solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -47,6 +51,194 @@ pub fn normalize_limit_order(order: Order, native_token: WETH9) -> LimitOrder {
...
+                executed_amount,
+                self.native_token.clone(),
+            )?;
+
+            encoder.add_trade(self.order.clone(), executed_amount)?;
+
+            encoder.add_unwrap(interaction);
+        } else {
+            encoder.add_trade(self.order.clone(), executed_amount)?;
+        }
```
nlordell, line 67: nit: Why not factor out the common `add_trade`?
```suggestion
            encoder.add_unwrap(interaction);
        }
        encoder.add_trade(self.order.clone(), executed_amount)?;
```

solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -47,6 +51,194 @@ pub fn normalize_limit_order(order: Order, native_token: WETH9) -> LimitOrder {
...
+    clearing_prices: &HashMap<H160, U256>,
+    order: &Order,
+    executed_amount: U256,
+    weth: WETH9,
+) -> Result<UnwrapWethInteraction> {
+    let sell_price = if let Some(&price) = clearing_prices.get(&order.order_creation.sell_token) {
+        price
+    } else {
+        return Err(anyhow!("sell price not available"));
+    };
```
nlordell, line 83: nit: This can be simplified with `ok_or_else` to take advantage of "try-sugar":
```suggestion
    let sell_price = clearing_prices.get(&order.order_creation.sell_token)
        .ok_or_else(|| anyhow!("sell price not available"))?;
```

Same applies to other `Option::None` handing :point_down:

solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -47,6 +51,194 @@ pub fn normalize_limit_order(order: Order, native_token: WETH9) -> LimitOrder {
...
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::interactions::dummy_web3;
+    use crate::settlement::assert_settlement_encoded_with;
+    use maplit::hashmap;
+    use model::order::OrderCreation;
+
```
nlordell, line 125: Maybe add a unit test for `executed_buy_amount` on arithmetic overflow when computing executed amount?

solver/src/settlement.rs
```diff
@@ -328,6 +335,30 @@ fn sell_order_surplus(
     }
 }
 
+#[cfg(test)]
```
nlordell, line 338: ubernit: Maybe put this in `pub mod tests`? I'm also totally fine with leaving it here, so feel free to ignore.

nlordell (commented):
Looks good!

Would like to see a unit tests checking overflow of price calculation of `executed_buy_amount`.

solver/src/liquidity/offchain_orderbook.rs
fedgiac: The repetition is there because I wanted to guarantee that if the result is `Err` then no changes have been made to `encoder` (under the assumption that if `add_trade` returns `Err`, then the same is true).
With the proposed ordering, if `add_trade` fails then there could still be a new interaction added to `encoder`.

nlordell: Ah, that makes perfect sense.

fleupold (approved):
Nice, agree with the syntactic sugar comments, logic looks great! Before merging could you run it locally and test that it works on e.g. Rinkeby?

solver/src/settlement.rs
```diff
@@ -10,6 +10,9 @@ use primitive_types::{H160, U256};
 use shared::conversions::U256Ext;
 use std::{collections::HashMap, iter};
 
+#[cfg(test)]
+use crate::liquidity::SettlementHandling;
```
nlordell, line 14: This can be moved into the `mod tests { ... }` as well.

nlordell (approved):
Awesome!

fleupold (commented):
I think we are still missing logic to treat/rename the buyToken as the native wrapped token when it is set to 0xee..ee. 

All other parts of the codebase (e.g price estimations. solvers, etc.) don't read 0xeee as something special and might therefore not do any matching.

fedgiac:
Very good point. I fixed it, now the solver should see all eth orders like weth orders.

fleupold (commented):
Cool thanks for fixing. We might have to do the same "trick" in the MinFeeCalculator that is quoting the fee (otherwise the frontend might not be able to place orders as it won't find a route).
Let's see if the manual e2e test works

fedgiac:
I've noticed an important issue with this code: it assumes that the `encode` function of the trait `OrderSettlementHandler` is only called after a solver has computed a solution. This is true for the HTTP solver, but not true for the naive solver. In this case, if the solver tries to call `encode` for a sell order that buys eth the encoding fails because the prices needed to compute the executed sell amount cannot be computed in advance. You can try this out in the `e2e-weth-no-price-bug` branch ([diff](https://github.com/gnosis/gp-v2-services/compare/enable-native-eth-orders...e2e-weth-no-price-bug?expand=1)).

Note that the naive solver doesn't rebuild the settlement after computing it, so we can't just use two different version of the `encode` function. At this point, I believe we would need to change how the naive solver computes the final settlement.

solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -47,6 +54,262 @@ pub fn normalize_limit_order(order: Order, native_token: WETH9) -> LimitOrder {
...
+    clearing_prices: &HashMap<H160, U256>,
+    order: &Order,
+    executed_amount: U256,
+    weth: WETH9,
+) -> Result<UnwrapWethInteraction> {
+    let sell_price = *clearing_prices
+        .get(&order.order_creation.sell_token)
+        .ok_or_else(|| anyhow!("sell price not available"))?;
+    let buy_price = *clearing_prices
+        .get(&order.order_creation.buy_token)
```
nlordell, line 86: I think this might be the issue you mentioned, where its trying to get the price of `0xeeeeeee...eeee` instead of `WETH.address()` which is what the solver outputs.

I think that, in order to fix this, we need to change the encoder to add native token address and price to the list.

nlordell:
> but not true for the naive solver.

It should be true, since the solver does `Settlement::new(prices)` before being able to `settlement.add_liquidity(...)` which enternally calls `encode`.

I think the issue is that the order is for `0xeee...` but the solver is only adding a price for `WETH.address()`.

vkgnosis (approved):

nlordell:
@fedgiac Should be ready to merge now once the manual E2E test with the FE is done.

PR #497, opened 2021-04-22
Introduce 1Inch HTTP Client

nlordell:
This PR introduces an HTTP client to the 1Inch API. This client will eventually be used by a "1Inch solver" that would be able to match single EOA orders to a 1Inch swap.

The HTTP client is currently incomplete and does not:
- Include all query parameters
- Parse the response JSON in its entirety

Although, it is enough for what we need for a 1Inch solver for now.

### Test Plan

Added new unit tests. To do a real HTTP request:
```
$ cargo test -- --ignored --nocapture oneinch_swap
...
running 1 test
Swap {
    ...
}
test oneinch_solver::api::tests::oneinch_swap ... ok
...
```


solver/src/oneinch_solver/api.rs
```diff
@@ -0,0 +1,367 @@
...
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+/// A 1Inch API quote query parameters.
+///
+/// These parameters are currently incomplete, and missing parameters can be
+/// added incrementally as needed.
+pub struct SwapQuery {
```
fleupold, line 55: Is it possible to model buy orders?

fleupold (approved):
Nice!

solver/src/oneinch_solver/api.rs
nlordell: Good question...

nlordell: According to https://docs.1inch.io/api/quote-swap I don't see any obvious way of doing it :disappointed:

vkgnosis (approved):

PR #498, opened 2021-04-23
Move SettlementEncoder into own file

vkgnosis:
While working with the settlement file I felt it got very big so I split
out the encoder into it's own file.

### Test Plan
no logic change, existing tests


fleupold (approved):
Likely conflicting with #500

PR #499, opened 2021-04-23
Add 0xee..ee to pre auction price vector

fleupold:
Needed for #461

I will highlight the main logic change, as it's quite small. It adds the native token alias (0xeee...eee) to our pre-auction price list so that it can be used to compute the objective value if traded.

The rest of the PR is factoring collect_estimated_prices into a free method and adding unit tests for it.

### Test Plan
Unit-tests + I made sure that this change was allowing e2e eth buy trades when combined with #496 ([tx](https://rinkeby.etherscan.io/tx/0xaf96177b579f3ffede03d49eba4b02504fce63156dc920d756a2b246102d7049))

### DIFF
--- a/shared/src/price_estimate.rs
+++ b/shared/src/price_estimate.rs
@@ -258,27 +258,43 @@ impl UniswapPriceEstimator {
     }
 }
 
-pub struct FakePriceEstimator(pub BigRational);
-#[async_trait::async_trait]
-impl PriceEstimating for FakePriceEstimator {
-    async fn estimate_price(&self, _: H160, _: H160, _: U256, _: OrderKind) -> Result<BigRational> {
-        Ok(self.0.clone())
+pub mod mocks {
+    use super::*;
+
+    pub struct FakePriceEstimator(pub BigRational);
+    #[async_trait::async_trait]
+    impl PriceEstimating for FakePriceEstimator {
+        async fn estimate_price(
+            &self,
+            _: H160,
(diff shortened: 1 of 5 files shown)


solver/src/driver.rs
```diff
@@ -223,3 +194,133 @@ impl Driver {
...
+                tracing::warn!("failed to estimate price for token {}: {:?}", token, err);
+                None
+            }
+        })
+        .collect();
+
+    // If the wrapped native token is in the price list (e.g. WETH), so should be the placeholder for its native counterpart
+    if let Some(price) = prices.get(&native_token).cloned() {
+        prices.insert(BUY_ETH_ADDRESS, price);
+    }
```
fleupold, line 236: This is the real change!

vkgnosis (approved):

shared/src/price_estimate.rs
```diff
@@ -270,6 +270,18 @@ impl PriceEstimating for FakePriceEstimator {
     }
 }
 
+pub struct FailingPriceEstimator();
+#[async_trait::async_trait]
+impl PriceEstimating for FailingPriceEstimator {
```
nlordell, line 275: maybe these "mock" implemetnations should go in a 

```
mod mocks {
  pub struct FakePriceEstimator;
  pub struct FailingPriceEstimator;
}
```

nlordell (approved):
Looks good.

PR #504, opened 2021-04-23
[do not merge] Modify e2e test to use weth

fedgiac:
This PR e2e tests the buy eth logic implemented in the recent PRs.

It will be made into an independent e2e test in the future, so it should not be merged.


fleupold:
Given we are already doing two trades in the e2e test, maybe we can make one of them buy the native token? This way we wouldn't have to run a second test...

fedgiac:
I think it makes sense to have a separate test, since it would be nice to test both buy and sell orders buying eth.

PR #505, opened 2021-04-26
Add support for 1Inch flag to disable balance checks

nlordell:
This PR adds the `disableEstimate` flag which disables balance and allowance checks when computing swap transaction.

/cc @twalth3r 

### Test Plan

New unit tests, and changed test to use settlement contract which has 0 Ether balance yet is able to retrieve a swap transaction selling Eth.

### DIFF
--- a/solver/src/oneinch_solver/api.rs
+++ b/solver/src/oneinch_solver/api.rs
@@ -66,6 +66,8 @@ pub struct SwapQuery {
     pub from_address: H160,
     /// Limit of price slippage you are willing to accept.
     pub slippage: Slippage,
+    /// Flag to disable checks of the required quantities.
+    pub disable_estimate: Option<bool>,
 }
 
 impl SwapQuery {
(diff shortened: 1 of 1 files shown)


solver/src/oneinch_solver/api.rs
```diff
@@ -66,6 +66,8 @@ pub struct SwapQuery {
     pub from_address: H160,
     /// Limit of price slippage you are willing to accept.
     pub slippage: Slippage,
+    /// Flag to disable checks of the required quantities.
+    pub disable_estimate: Option<bool>,
```
vkgnosis, line 70: If None is the same as Some(false) maybe we have the raw bool instead of Option. Not sure what is better.

vkgnosis (approved):

fleupold (approved):

solver/src/oneinch_solver/api.rs
nlordell: The query accepts `disableEstimate=false` as well. In theory, its an optional flag with a default value which is `false` ATM, but we may want to force it to `false` if that were to change. So semantically there are 3 options:
- `None`: use default
- `Some(false)`: force enable estimate
- `Some(true)`: force disable estimate

Also, it keeps it more 1:1 with the API.

PR #506, opened 2021-04-26
Handle mip solutions with no execution plan.

marcovc:
This PR allows the driver code for the http solver to handle solutions that don't include an execution plan (essentially just treating them as trivial solutions).

There was the question about doing it here, or on the http solver itself. I'm fine both ways, and @twalth3r prefers that the solver still outputs a solution even if there is no exec plan, since we will be interested in the solution for debugging purposes.

The c++ version of the solver we've been working on guarantees that a uniswap is used only if it participates in a feasible execution plan, so this decision will not be important there.

Closes https://github.com/gnosis/gp-v2-solver/issues/119


nlordell (approved):
Looks good to me.

vkgnosis (approved):

fleupold (approved):

Issue #402, opened 2021-03-19
Handle backlog of "single trades" more efficiently


fleupold:
At the moment the majority of our solvers will settle trades one by one. Since we wait 30 seconds between batches and mining the solution tx takes some time, we can effectively handle about 0.5-1 individual trades per minute. 

This has become noticable in today's testing challenge (worsened by the fact that we were stuck 5 minutes without a solution submission) when multiple people tried to trade non-overlapping tokens at the same time, some trades got very delayed.

This might be important to address, before we get real traffic which might be spiky (e.g. around tweets, etc).

A simple approach would be to make the solver trade return a list of solutions, which we could batch up into a single settlement (or multiple parallel txs). However, this bears the risk of one solution effecting e.g. the uniswap state of another solution and might therefore be invalid by the time the first is mined (maybe we can base all our queries on the pending block which could already have the first solution applied).

We should at least skip the 30s pause when there has been a solution submission in the current run loop.

fleupold:
We discussed this in the team on Monday. We already reduced the time between runloops as long as there are orders in the system that had been created more than 30s ago. Yet in our last test we could easily encounter user wait times of up to 10 minutes.

A next immediate step to address this issue is to change the solver trait to allow returning a `Vec<Settlements>` (e.g. the baseline solver would return one settlement per limit order).

For now, the driver would merge Settlements into a larger "multi trade" settlement and submit it with the current tx logic. A concern about having many independent trades in the same settlement is that failing to settle a single trade, would cause a failure to all involved trades. Since this compounds with number of trades per settlement, we might "upper bound" the merging logic to include no more than n trades per settlement.

Mid-term it probably makes sense to refactor the tx submission logic to allow tracking multiple transactions in parallel (so that any subset of them could be mined/reverted independently without blocking others). This also works nicely with the fact that the MIP solver already returns independent sequences.

The question of how to update the state of e.g. the AMM pools for the trade n+1 assuming that trade n gets mined first can be an implementation detail of each solver. We might want to refactor the SettlementHandling trait to take a `&mut self` so that liquidity can update itself based on information on how it is used in an earlier sequence of the settlement. 

For now we consider this less urgent as material changes to the underlying reserves (in a way that effect prices significantly) require large trade sizes.

PR #507, opened 2021-04-26
Return multiple settlements in Solver trait and handle this in Driver

vkgnosis:
Part of https://github.com/gnosis/gp-v2-services/issues/402 .

This PR makes the solver trait return a vector of settlements which in a future PR can be merged back into a single settlement.
The driver is updated so that it handles more settlements better even as it does not yet merge.
Previously we would go through the settlements ordered by objective value and try to submit them one by one. As part of the submission the settlement would get simulated.
Now we simulate all settlements (in parallel) and attempt to submit only the highest objective value settlement that passed the simulation.

This is advantageous because
* We can log simulation errors even for settlements that we did not attempt to submit. We want to see those errors so we can fix them in the solvers.
* The driver run loop only ever submits one settlement. If we attempt to submit a transaction which later fails it is very likely that the other settlements are out of date (would fail too or could be improved) at that point.

With this PR we are immediately making use of the naive solver potentially returning multiple settlements. The new code can pick the highest objective value in this case whereas the old code only observed an arbitrary single settlement.

### Test Plan
CI

### DIFF
--- a/e2e/tests/uniswap_trade_test.rs
+++ b/e2e/tests/uniswap_trade_test.rs
@@ -246,6 +246,7 @@ async fn test_with_ganache() {
         uniswap_like_liquidity: vec![uniswap_liquidity],
         orderbook_api: create_orderbook_api(&web3),
     };
+    let network_id = web3.net().version().await.unwrap();
     let mut driver = solver::driver::Driver::new(
         gp_settlement.clone(),
         liquidity_collector,
@@ -258,6 +259,7 @@ async fn test_with_ganache() {
         Duration::from_secs(0),
         Arc::new(NoopMetrics::default()),
         web3.clone(),
+        network_id,
     );
     driver.single_run().await.unwrap();
 
--- a/e2e/tests/uniswap_trade_test.rs
+++ b/e2e/tests/uniswap_trade_test.rs
@@ -257,6 +257,7 @@ async fn test_with_ganache() {
         native_token,
         Duration::from_secs(0),
         Arc::new(NoopMetrics::default()),
+        web3.clone(),
     );
     driver.single_run().await.unwrap();
 
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -35,6 +35,7 @@ pub struct Driver {
     min_order_age: Duration,
     metrics: Arc<dyn SolverMetrics>,
     web3: Web3,
+    network_id: String,
 }
 impl Driver {
     #[allow(clippy::too_many_arguments)]
@@ -50,6 +51,7 @@ impl Driver {
         min_order_age: Duration,
         metrics: Arc<dyn SolverMetrics>,
         web3: Web3,
+        network_id: String,
     ) -> Self {
         Self {
             settlement_contract,
@@ -63,6 +65,7 @@ impl Driver {
             min_order_age,
             metrics,
             web3,
+            network_id,
         }
     }
 
@@ -162,6 +165,7 @@ impl Driver {
                 .map(|settlement| settlement.settlement.clone().into()),
             &self.settlement_contract,
             &self.web3,
+            &self.network_id,
         )
         .await
         .context("failed to simulate settlements")?;
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -3,7 +3,7 @@ use crate::{
     liquidity_collector::LiquidityCollector,
     metrics::SolverMetrics,
     settlement::Settlement,
-    settlement_submission,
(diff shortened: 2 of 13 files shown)


solver/src/driver.rs
```diff
@@ -261,9 +297,15 @@ fn liquidity_with_price(
...
     // Index in the Driver::solver vector
     index: usize,
+    settlements: Vec<SolverSettlement>,
+}
+
+// Each individual settlement has an objective value.
+struct SolverSettlement {
     settlement: Settlement,
     objective_value: BigRational,
 }
```
vkgnosis, line 311: I am unsure about these structs. They will most likely change when settlements get merged so I didn't think too hard about proper naming and encapsulation.

solver/src/settlement_simulation.rs
```diff
@@ -0,0 +1,72 @@
+use crate::encoding::EncodedSettlement;
```
vkgnosis, line 1: The code in this module has moved from settlement_submission.rs and been changed to simulate multiple settlements in batches.

solver/src/settlement_submission.rs
```diff
@@ -29,9 +24,7 @@ pub async fn submit(
     gas_price_cap: f64,
     settlement: Settlement,
 ) -> Result<()> {
-    let settlement = settlement.into();
-    // Check that a simulation of the transaction works before submitting it.
-    simulate_settlement(&settlement, contract).await?;
+    let settlement: EncodedSettlement = settlement.into();
```
vkgnosis, line 27: Note that this means that this function `submit` no longer does any simulation because the driver simulates all settlements beforehand. I'd also be okay to simulate anyway for extra safety.

fleupold (approved):
PR looks good. While number of lines is not crazy high there is a lot going on here and I think it would have been nice to split it into multiple PRs. e.g:

1. Simulation logic refactoring
2. Adding Vec to Settlements + runloop refactoring (maybe even that could be split up)

It also seems like a lot of this logic isn't really unit tested (I see that this is hard to do), but as the logic becomes more involved, I think it also become important to test these parts.

solver/src/settlement.rs
```diff
@@ -73,7 +73,7 @@ impl Interaction for EncodedInteraction {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
vkgnosis: (responding below)

solver/src/settlement/settlement_encoder.rs
```diff
@@ -15,12 +15,15 @@ use std::{collections::HashMap, iter};
...
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct SettlementEncoder {
     tokens: Vec<H160>,
     clearing_prices: HashMap<H160, U256>,
     trades: Vec<Trade>,
-    execution_plan: Vec<Box<dyn Interaction>>,
+    // This is an Arc so that this struct is Clone. Cannot require `Interaction: Clone` because it
+    // would make the trait not be object safe which prevents using it through `dyn`.
+    // TODO: Can we fix this in a better way?
```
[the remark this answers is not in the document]
vkgnosis: Not  going to be possible once we do merging because I imagine we'll want the option to still settle individual settlements if the merged one fails to simulate.

solver/src/settlement_simulation.rs
```diff
@@ -0,0 +1,72 @@
...
+            (method.view().batch_call(&mut batch), transaction_builder)
+        })
+        .collect::<Vec<_>>();
+
+    // TODO: It would be nice to add this to the underlying web3 batch transport call used for the
+    // simulations.
+    let ((), current_block, network_id) = futures::join!(
+        batch.execute_all(SIMULATE_BATCH_SIZE),
+        web3.eth().block_number(),
+        web3.net().version()
```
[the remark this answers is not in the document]
vkgnosis: Agree. Will add this as another PR on top of this one.

e2e/tests/uniswap_trade_test.rs
```diff
@@ -246,6 +246,7 @@ async fn test_with_ganache() {
         uniswap_like_liquidity: vec![uniswap_liquidity],
         orderbook_api: create_orderbook_api(&web3),
     };
+    let network_id = web3.net().version().await.unwrap();
```
nlordell, line 249: We should probably start using chain ID instead of network ID everywhere, especially since it is tied to the signatures.

Just a general comment that doesn't have to be addressed in this PR.

nlordell (approved):
Code looks very good.

PR #509, opened 2021-04-26
Support settlement merging

vkgnosis:
Adds a basic way of merging settlements. Prices and trades are combined
and it is checked that the prices are the same and no trade is used
twice. Interactions are appended. This could be more efficient because
in some case it might be possible to for example merge uniswap
interactions the way we merge unwrapping but the current simple
implementation is what we want in the short term.

### Test Plan
new unit tests

### DIFF
--- a/solver/src/settlement/settlement_encoder.rs
+++ b/solver/src/settlement/settlement_encoder.rs
@@ -5,7 +5,10 @@ use model::order::{Order, OrderKind};
 use num::{BigRational, Zero};
 use primitive_types::{H160, U256};
 use shared::conversions::U256Ext;
-use std::{collections::HashMap, iter};
+use std::{
+    collections::{hash_map::Entry, HashMap},
+    iter,
+};
 
 /// An intermediate settlement representation that can be incrementally
 /// constructed.
@@ -17,8 +20,12 @@ use std::{collections::HashMap, iter};
 /// (e.g. collapsing two interactions into one equivalent one).
 #[derive(Debug)]
 pub struct SettlementEncoder {
+    // Make sure to update the `merge` method when adding new fields.
+
+    // Invariant: tokens is all keys in clearing_prices sorted.
     tokens: Vec<H160>,
     clearing_prices: HashMap<H160, U256>,
+    // Invariant: Every trade's buy and sell token has an entry in clearing_prices.
     trades: Vec<Trade>,
     execution_plan: Vec<Box<dyn Interaction>>,
     unwraps: Vec<UnwrapWethInteraction>,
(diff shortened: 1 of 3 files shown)


solver/src/settlement/settlement_encoder.rs
```diff
@@ -373,4 +421,67 @@ pub mod tests {
...
+    fn merge_ok() {
+        let prices = hashmap! { token(1) => 1.into(), token(3) => 3.into() };
+        let mut encoder0 = SettlementEncoder::new(prices);
+        let mut order13 = OrderBuilder::default()
+            .with_sell_token(token(1))
+            .with_buy_token(token(3))
+            .build();
+        order13.order_meta_data.uid.0[0] = 0;
+        encoder0.add_trade(order13, 13.into()).unwrap();
+        encoder0.append_to_execution_plan(NoopInteraction {});
```
fleupold, line 439: Can we also add an unwrap here for completeness?

solver/src/settlement/settlement_encoder.rs
```diff
@@ -214,14 +225,51 @@ impl SettlementEncoder {
...
+        self.trades.append(&mut other.trades);
+        self.sort_tokens_and_update_indices();
+
+        self.execution_plan.append(&mut other.execution_plan);
+
+        for unwrap in other.unwraps {
+            self.add_unwrap(unwrap);
+        }
+
+        Ok(self)
```
fleupold, line 261: Not sure how much logic is in these helper methods, but maybe it would be safer here to create a `SettlementEncoder` and set all fields manually. This way the compiler would yell at us if we create a new field and forget to update the merging logic?

fleupold (approved):

bh2smith (approved):

solver/src/settlement/settlement_encoder.rs
vkgnosis: The function doesn't return a result so there is nothing to unwrap.

solver/src/settlement/settlement_encoder.rs
vkgnosis: Feels weird to destructure self and then restructure just for that. I will add a comment to the field declarations noting `merge`.

solver/src/settlement/settlement_encoder.rs
fleupold: I meant an `UnwrapInteraction`. Can you add `encoder0.add_unwrap(...)` to the testcase? Otherwise, I think we could "forget" to push the unwraps in the `merge` implementations and the unit test would still pass.

vkgnosis: Will do in another PR since I had auto merge enabled here.

PR #511, opened 2021-04-26
Add Ganache wrapper for e2e tests

fedgiac:
Adds a wrapper function to be used when running end to end tests in order to reset the state of the blockchain to that before the test was run. The state is notably reset also if the test fails.
This is helpful to run all e2e tests starting from a clean state, particularly in light of a future e2e test which will change the eth balances of the funded accounts.

The code is copied from [ido-services](https://github.com/gnosis/ido-services/pull/85), in turn suggested by Nick. As mentioned by Alex, it would be a good candidate to be moved to a standalone utils crate.

### Test Plan

The existing unit test still passes.

To observe that a failing test still restores the snapshot on Ganache:
1. Start database and Ganache.
2. Add a `panic!` at the start of the function `onchain_settlement`. 
2. Run `cargo test --locked --package e2e` with Ganache running.
3. Observe that the test fails and the Ganache output reports:
```
Saved snapshot #1
evm_revert
Reverting to snapshot #1
```

### DIFF
--- a/e2e/tests/ganache.rs
+++ b/e2e/tests/ganache.rs
@@ -1,0 +1,94 @@
+use ethcontract::futures::FutureExt;
+use ethcontract::{Http, U256};
+use shared::{transport::LoggingTransport, Web3};
+use std::{
+    fmt::Debug,
+    future::Future,
+    panic::{self, AssertUnwindSafe},
+};
+use web3::{api::Namespace, helpers::CallFuture, Transport};
+
+const NODE_HOST: &str = "http://127.0.0.1:8545";
+
+/// *Testing* function that takes a closure and executes it on Ganache.
+/// Before each test, it creates a snapshot of the current state of the chain.
+/// The saved state is restored at the end of the test.
+///
+/// This function must not be called again until the current execution has
+/// terminated.
+pub async fn test<F, Fut>(f: F)
+where
+    F: FnOnce(Web3) -> Fut,
+    Fut: Future<Output = ()>,
+{
+    let http = LoggingTransport::new(Http::new(NODE_HOST).expect("transport failure"));
+    let web3 = Web3::new(http);
+    let resetter = Resetter::new(&web3).await;
+
+    // Hack: the closure may actually be unwind unsafe; moreover, `catch_unwind`
+    // does not catch some types of panics. In this cases, the state of the node
+    // is not restored. This is not considered an issue since this function
+    // is supposed to be used in a test environment.
+    let result = AssertUnwindSafe(f(web3.clone())).catch_unwind().await;
+
+    resetter.reset().await;
+
+    if let Err(err) = result {
+        panic::resume_unwind(err);
+    }
+}
+
+struct Resetter<T> {
+    ganache: GanacheApi<T>,
+    snapshot_id: U256,
+}
+
+impl<T: Transport> Resetter<T> {
+    async fn new(web3: &web3::Web3<T>) -> Self {
+        let ganache = web3.api::<GanacheApi<_>>();
+        let snapshot_id = ganache
+            .snapshot()
+            .await
+            .expect("Test network must support evm_snapshot");
+        Self {
+            ganache,
+            snapshot_id,
+        }
+    }
+
+    async fn reset(&self) {
+        self.ganache
+            .revert(&self.snapshot_id)
+            .await
+            .expect("Test network must support evm_revert");
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct GanacheApi<T> {
+    transport: T,
+}
+
+impl<T: Transport> Namespace<T> for GanacheApi<T> {
+    fn new(transport: T) -> Self
+    where
+        Self: Sized,
+    {
+        GanacheApi { transport }
+    }
+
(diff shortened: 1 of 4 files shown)


bh2smith (approved):
Very cool. Reminds me of east back in the day when I published an npm package called ganache-snapshot for exactly this purpose! Looks much nicer in rust.

e2e/tests/ganache.rs
```diff
@@ -0,0 +1,94 @@
...
+{
+    let http = LoggingTransport::new(Http::new(NODE_HOST).expect("transport failure"));
+    let web3 = Web3::new(http);
+    let resetter = Resetter::new(&web3).await;
+
+    // Hack: the closure may actually be unwind unsafe; moreover, `catch_unwind`
+    // does not catch some types of panics. In this cases, the state of the node
+    // is not restored. This is not considered an issue since this function
+    // is supposed to be used in a test environment.
+    let result = AssertUnwindSafe(f(web3.clone())).catch_unwind().await;
```
fleupold, line 32: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FHDNcjt5ELkJSE%2Fgiphy.gif&token=fancy" data-gitmeme-token="fancy"></a>

e2e/tests/onchain_settlement.rs
```diff
@@ -29,20 +28,23 @@ use solver::{
...
     hex!("0000000000000000000000000000000000000000000000000000000000000002");
 
-const NODE_HOST: &str = "http://127.0.0.1:8545";
 const API_HOST: &str = "http://127.0.0.1:8080";
 const ORDER_PLACEMENT_ENDPOINT: &str = "/api/v1/orders/";
 
 #[tokio::test]
-async fn test_with_ganache() {
+async fn ganache_onchain_settlement() {
+    ganache::test(onchain_settlement).await;
```
fleupold, line 43: Just curious, could this be also done via an annotation similar to `#[tokio:test]`?

fleupold (approved):

vkgnosis (approved):
Good idea. One thing to keep in mind when adding more e2e tests is that rust runs them in parallel by default which will not work with a reset like this (but we probably don't want them in parallel anyway).
I wonder if ganache has a "reset everything" function which could be used instead of snapshot. Then we wouldn't need to catch panics as we could reset at the start of each test like what we do for the postgres tests.

e2e/tests/onchain_settlement.rs
nlordell: Yes, but it would require moving this code to a `proc-macro` crate.

nlordell: Its actually quite fun :smile:. I wouldn't mind helping get someone do a PoC with a pair programming session on ecosystem-friday if anyone is interested :smile:

nlordell:
> e2e tests is that rust runs them in parallel by default

Maybe we need a global mutex to ensure tests are run in sequence?

nlordell (approved):

fedgiac:
I couldn't find any way to completely reset Ganache to a fresh state, which would be a better solution since right now if a test is interrupted while in progress (`Ctrl+C`) then the state of Ganache is not reset. 

<details><summary>How the mutex was tested</summary>

Add the code that follows to `onchain_settlement.rs`. Then remove the line starting with `shared::tracing::initialize` from the same file.

```
#[tokio::test]
async fn ganache_onchain_settlement_2() {
    ganache::test(onchain_settlement).await;
}

#[tokio::test]
async fn ganache_onchain_settlement_3() {
    ganache::test(onchain_settlement).await;
}

#[tokio::test]
async fn ganache_onchain_settlement_4() {
    ganache::test(onchain_settlement).await;
}

#[tokio::test]
async fn ganache_onchain_settlement_5() {
    ganache::test(|_| async {
        panic!();
    })
    .await;
}
```
</details>

I'm happy to try to make this into a `#[macro]` on a Friday. :smile:

e2e/tests/ganache.rs
```diff
@@ -0,0 +1,106 @@
...
+///
+/// Note that tests calling with this function will not be run aymultaneously.
+pub async fn test<F, Fut>(f: F)
+where
+    F: FnOnce(Web3) -> Fut,
+    Fut: Future<Output = ()>,
+{
+    // The mutex guarantees that no more than a test at a time is running on
+    // Ganache.
+    // Note that the mutex is expected to become poisoned if a test panics. This
```
nlordell, line 31: Wouldn't the `panic` get caught with the `catch_unwind()` :point_down: and then have the `Mutex` not get poisoned? Either way, I like not caring about poisoned mutex here since [`catch_unwind` doesn't work for all panics](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html):

> Additionally, this function is not guaranteed to catch all panics

nlordell (approved):
Approve the mutex changes.

e2e/tests/ganache.rs
fedgiac: By trying it out it looks like the mutex becomes poisoned in all tests after the first one failing, I wasn't able to find a way to "unpoison" a mutex manually.

PR #512, opened 2021-04-26
[Baseline Solver] Allow multiple liquidities on the same token pair

fleupold:
This PR allows for the baseline solver to take multiple liquidity sources for the same token pair and internally choose the best one for a given hop.

This will be useful for correctly integrating sushi swap liquidity since right now we have to decide for a single pool to pass into the solver before knowing the traded amounts & direction. At the moment we choose the pool with the least slippage. While this is a reasonable heuristic, this may lead to suboptimal routes (e.g. if the less liquid pool has a lower spot price and the amount traded is small enough to not cause a lot of slippage).

Right now, all BaselineSolvable ingredients have to be the same type (`Pool` as of right now). As we add more `impl BaselineLiquidity` we might want to make the `estimate_xyz` methods take a `Vec<Box<dyn BaselineSolvable>>` instead. Note however, that the driver needs to reference back to the original concrete liquidity (to encode its interaction into a settlement) and therefore some for of downcasting would be needed.

Given that this is not yet required I only introduced the `Vector<dyn BaselineSolvable>` and didn't box it yet. This way the compiler can infer the concrete type and no downcasting is needed.

### Test Plan
Added two unit tests.

### DIFF
--- a/shared/src/baseline_solver.rs
+++ b/shared/src/baseline_solver.rs
@@ -98,7 +98,8 @@ pub fn estimate_sell_amount<'a, L: BaselineSolvable>(
         )
         .map(|(amount, _, liquidity)| Estimate {
             value: amount,
-            path: liquidity,
+            // Since we reversed the path originally, we need ret re-reverse it here.
+            path: liquidity.into_iter().rev().collect(),
         })
 }
 
@@ -365,7 +366,7 @@ mod tests {
         let sell_estimate = estimate_sell_amount(1000.into(), &path, &pools).unwrap();
         assert_eq!(
             sell_estimate.path,
-            [&second_hop_low_slippage, &first_hop_low_price]
+            [&first_hop_low_price, &second_hop_low_slippage]
         );
 
         let spot_price = estimate_spot_price(&path, &pools).unwrap();
@@ -385,7 +386,7 @@ mod tests {
         let sell_estimate = estimate_sell_amount(1000.into(), &path, &pools).unwrap();
         assert_eq!(
             sell_estimate.path,
-            [&first_hop_high_price, &second_hop_low_slippage]
+            [&second_hop_low_slippage, &first_hop_high_price]
         );
 
         let spot_price = estimate_spot_price(&path, &pools).unwrap();
--- a/shared/src/baseline_solver.rs
+++ b/shared/src/baseline_solver.rs
@@ -20,78 +20,145 @@ pub trait BaselineSolvable {
     fn get_spot_price(&self, base_token: H160, quote_token: H160) -> Option<BigRational>;
 }
 
-// Given a path and sell amount (first token of the path) estimates the buy amount (last token of the path).
+pub struct Estimate<'a, V, L> {
+    // The result amount of the estimate
+    pub value: V,
+    // The liquidity path used to derive at that estimate
+    pub path: Vec<&'a L>,
+}
+
+// Given a path and sell amount (first token of the path) estimates the buy amount (last token of the path) and
+// the path of liquidity that yields this result
 // Returns None if the path is invalid or pool information doesn't exist.
-pub fn estimate_buy_amount(
+pub fn estimate_buy_amount<'a, L: BaselineSolvable>(
     sell_amount: U256,
     path: &[H160],
-    liquidity: &HashMap<TokenPair, impl BaselineSolvable>,
-) -> Option<U256> {
-    let sell_token = path.first()?;
-    path.iter()
-        .skip(1)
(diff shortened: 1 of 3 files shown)


shared/src/baseline_solver.rs
```diff
@@ -20,78 +20,145 @@ pub trait BaselineSolvable {
     fn get_spot_price(&self, base_token: H160, quote_token: H160) -> Option<BigRational>;
 }
 
-// Given a path and sell amount (first token of the path) estimates the buy amount (last token of the path).
+pub struct Estimate<'a, V, L> {
+    // The result amount of the estimate
+    pub value: V,
+    // The liquidity path used to derive at that estimate
+    pub path: Vec<&'a L>,
+}
```
bh2smith, line 28: Very fancy looking struct! Much generic! Such Lifetime!

shared/src/baseline_solver.rs
```diff
@@ -20,78 +20,145 @@ pub trait BaselineSolvable {
...
+            |previous, current| {
+                let previous = match previous {
+                    Some(previous) => previous,
+                    None => return None,
+                };
+
+                match liquidity.get(&TokenPair::new(*current, previous.1)?) {
+                    Some(liquidity) => liquidity
+                        .iter()
+                        .max_by_key(|item| item.get_amount_out(*current, previous.0, previous.1))
```
bh2smith, line 52: So from what I understand, we are indirectly getting the better pool based on slippage and spot price by "greedily" maximizing the output amount! This is very clever and much simpler than working with the spot prices and pool size directly!

bh2smith (approved):
This is really nice work here. Well done! We are "en train" to become an incredible meta-dex-aggregator!

shared/src/baseline_solver.rs
```diff
@@ -20,78 +20,145 @@ pub trait BaselineSolvable {
...
+                                            .into_iter()
+                                            .chain(std::iter::once(best))
+                                            .collect(),
+                                    )
+                                })
+                        }),
+                    None => None,
+                }
+            },
+        )
```
vkgnosis, line 70: I have simplified the part in `fold`. You should be able to do something similar for the sell amount case.
```suggestion
        .fold(
            Some((sell_amount, *sell_token, Vec::new())),
            |previous, current| {
                let (amount, previous, mut path) = previous?;
                let best = liquidity
                    .get(&TokenPair::new(*current, previous)?)?
                    .iter()
                    .map(|l| (l, l.get_amount_out(*current, amount, previous)))
                    .max_by(|a, b| a.1.cmp(&b.1))?;
                path.push(best.0);
                Some((best.1?, *current, path))
            },
        )
```

vkgnosis (approved):

shared/src/baseline_solver.rs
```diff
@@ -20,78 +20,145 @@ pub trait BaselineSolvable {
...
-                    .map(|amount| (amount, *current)),
-                None => None,
-            }
+        .fold(
+            Some((sell_amount, *sell_token, Vec::new())),
+            |previous, current| {
+                let previous = match previous {
+                    Some(previous) => previous,
+                    None => return None,
+                };
```
nlordell, line 47: I think you should be able can you use `previous?` here since the closure returns `Option`:
```suggestion
                let previous = previous?;
```

shared/src/baseline_solver.rs
```diff
@@ -20,78 +20,145 @@ pub trait BaselineSolvable {
...
+
+                match liquidity.get(&TokenPair::new(*current, previous.1)?) {
```
nlordell, line 49: One thing you can do in order to avoid indexing tuples (which makes it hard to remember which field is what) is to destructure early on:
```rust
let (sell_amount, sell_token, path) = previous?;
```

I'm also OK as is, since all tuple fields have different types, so if the wrong one is used somewhere, then you will get a type error anyway.

nlordell (approved):
Looks nice!

Maybe `estimate_sell_amount` and `estimate_buy_amount` could be refactored to share a good chunk of code. For another PR though :smile:

shared/src/baseline_solver.rs
```diff
@@ -98,7 +98,8 @@ pub fn estimate_sell_amount<'a, L: BaselineSolvable>(
         )
         .map(|(amount, _, liquidity)| Estimate {
             value: amount,
-            path: liquidity,
+            // Since we reversed the path originally, we need ret re-reverse it here.
```
bh2smith, line 101:
```suggestion
            // Since we reversed the path originally, we need to re-reverse it here.
```

PR #513, opened 2021-04-27
Make baseline solver return multiple settlements

vkgnosis:
Part of https://github.com/gnosis/gp-v2-services/issues/402 

### Test Plan
existing tests still work. I did not add a new test for the multi settlement case because the code change is small and it is so much boilerplate to set up two independent settlements.


fleupold (approved):
Nice!

PR #515, opened 2021-04-27
Merge multiple settlements in driver

vkgnosis:
Finally enabled merging of multiple settlements.

How do you think we should test this? I could adapt the e2e test so that we assert that in a single run it settles 4 orders instead of the current two.

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -1,3 +1,6 @@
+mod solver_settlements;
+
+use self::solver_settlements::RatedSettlement;
 use crate::{
     liquidity::{offchain_orderbook::BUY_ETH_ADDRESS, Liquidity},
     liquidity_collector::LiquidityCollector,
@@ -36,6 +39,7 @@ pub struct Driver {
     metrics: Arc<dyn SolverMetrics>,
     web3: Web3,
     network_id: String,
+    max_merged_settlements: usize,
 }
 impl Driver {
     #[allow(clippy::too_many_arguments)]
@@ -52,6 +56,7 @@ impl Driver {
         metrics: Arc<dyn SolverMetrics>,
         web3: Web3,
         network_id: String,
+        max_merged_settlements: usize,
     ) -> Self {
         Self {
             settlement_contract,
@@ -66,6 +71,7 @@ impl Driver {
             metrics,
             web3,
             network_id,
+            max_merged_settlements,
         }
     }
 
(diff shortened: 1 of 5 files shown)


solver/src/driver.rs
```diff
@@ -314,14 +337,24 @@ struct RatedSettlement {
...
+        .next()
+        .ok_or_else(|| anyhow!("no settlements"))?;
+    for settlement in settlements.take(MAX_MERGED_SETTLEMENTS - 1) {
+        merged = merged.merge(settlement)?;
+    }
+    Ok(merged)
+}
+
 #[cfg(test)]
 mod tests {
```
nlordell, line 352: Might be worth adding some unit tests for the new logic.

nlordell (commented):
Changes look good.

I would like to see a couple of unit tests that demonstrates how merging affects the `SolverWIthSettlements` result, but it may be quite verbose to setup. Maybe if more of the code in the `filter_map` closure were in a function, it might be easier to test?

solver/src/driver.rs
```diff
@@ -314,14 +337,24 @@ struct RatedSettlement {
     objective_value: BigRational,
 }
 
+// Goes through the settlements in order and tries to merge at most MAX_SETTLEMENT_MERGES together.
+fn merge_settlements(mut settlements: impl Iterator<Item = Settlement>) -> Result<Settlement> {
+    let mut merged = settlements
+        .next()
+        .ok_or_else(|| anyhow!("no settlements"))?;
+    for settlement in settlements.take(MAX_MERGED_SETTLEMENTS - 1) {
+        merged = merged.merge(settlement)?;
```
fleupold, line 346: If a specific settlement cannot be merged into the existing bunch, shouldn't we still try to merge the rest of the settlements? This way we can still reduce the number of total transactions we are waiting for.

solver/src/driver.rs
```diff
@@ -23,6 +23,8 @@ use std::{
 // protect against a gas estimator giving bogus results that would drain all our funds.
 const GAS_PRICE_CAP: f64 = 500e9;
 
+const MAX_MERGED_SETTLEMENTS: usize = 5;
```
fleupold, line 26: Should this be configurable via command line so it's easier to adapt?

fleupold (approved):
Second the suggestion to add unit tests. The driver logic is becoming more and more complex and doesn't yet have good coverage.

Otherwise, great PR 🎉 !

bh2smith (approved):

solver/src/driver.rs
vkgnosis: Fixed.

vkgnosis:
I added two more commits which:
* move some of the settlement handling logic into another file
* split out merge related function and added unit test
* make max merge count configurable

solver/src/driver.rs
```diff
@@ -204,25 +175,31 @@ impl Driver {
...
-                    settlement.settlement,
-                );
-            }
-        }
+            .run_solvers(liquidity, gas_price)
+            .await
+            .filter_map(solver_settlements::filter_bad_settlements)
+            .inspect(|(name, settlements)| {
+                for settlement in settlements.iter() {
+                    tracing::debug!("solver {} found solution:\n {:?}", name, settlement,);
```
nlordell, line 183: ubernit: trailing comma
```suggestion
                    tracing::debug!("solver {} found solution:\n {:?}", name, settlement);
```

solver/src/driver/solver_settlements.rs
```diff
@@ -0,0 +1,204 @@
+use crate::settlement::Settlement;
+use anyhow::Result;
+use num::BigRational;
+use primitive_types::H160;
+use std::{collections::HashMap, time::Duration};
+
+// Return None if the result is an error or there are no settlements remaining after removing
+// settlements with no trades.
+pub fn filter_bad_settlements(
```
nlordell, line 9: "bad" - so judgemental :stuck_out_tongue:.

nlordell (approved):
Looks amazing

PR #516, opened 2021-04-27
Have an explicit Solver::name function

vkgnosis:
I find this cleaner than requiring Display and it allows the trait to
be automockable.

### Test Plan
still compiles

### DIFF
--- a/solver/src/baseline_solver.rs
+++ b/solver/src/baseline_solver.rs
@@ -8,10 +8,7 @@ use shared::{
     },
     pool_fetching::Pool,
 };
-use std::{
-    collections::{HashMap, HashSet},
-    fmt,
-};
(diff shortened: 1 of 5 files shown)


nlordell (approved):
I also like this much better.

bh2smith (approved):
So does this change the way it looks in the logs or is this just an internal change?

nlordell:
Just an internal change. Now you do
```rust
println!("{}", solver.name());
```

Instead of
```rust
println!("{}", solver);
```

The reason I prefer it like this is that:
1. I find the `solver.name()` code easier to read: "print the solver name" is less ambiguous than "print the solver".
2. I think `Solver: Display` was abusing the `Display` trait a bit.

fleupold (approved):

PR #517, opened 2021-04-27
Separate ERC20 approval into its own interaction

nlordell:
This tiny PR just adds a ERC20 approval interaction and refactors the Uniswap one to use it.

While this PR is not very complex, when looking through the code I noticed that each liquidity implementation would have to manage approvals themselves. I believe this is better suited for the `SettlementEncoder` so that it can accumulate and group together separate approvals, as well as verify whether or not existing allowances suffice. That is a larger change that I started, but led to a thread that I kept on pulling and pulling. For now, I would rather just duplicate allowance checking code in the 1Inch solver and then work towards moving the responsibility towards the settlement encoder in separate PRs after the 1Inch solver is in place.

### Test Plan

Added unit test.

### DIFF
--- a/solver/src/interactions/erc20.rs
+++ b/solver/src/interactions/erc20.rs
@@ -1,0 +1,56 @@
+//! Module continaing ERC20 token interaction implementations.
+
+use crate::{encoding::EncodedInteraction, settlement::Interaction};
+use contracts::ERC20;
+use primitive_types::{H160, U256};
+
+#[derive(Debug)]
+pub struct Erc20ApproveInteraction {
+    pub token: ERC20,
+    pub owner: H160,
+    pub spender: H160,
+    pub amount: U256,
+}
+
+impl Erc20ApproveInteraction {
+    pub fn as_encoded(&self) -> EncodedInteraction {
+        let method = self.token.approve(self.spender, self.amount);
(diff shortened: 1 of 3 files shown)


solver/src/interactions/erc20.rs
```diff
@@ -0,0 +1,56 @@
+//! Module continaing ERC20 token interaction implementations.
+
+use crate::{encoding::EncodedInteraction, settlement::Interaction};
+use contracts::ERC20;
+use primitive_types::{H160, U256};
+
+#[derive(Debug)]
+pub struct Erc20ApproveInteraction {
+    pub token: ERC20,
+    pub owner: H160,
```
fleupold, line 10: Why is owner needed on the struct?

fleupold (approved):
Make sense that this should probably refactored in the future to benefit multiple liquidity sources easily. I'm not sure if it needs to go on the `SettlementEncoder` (as each liquidity source likely has some custom logic on which address to use) but at least it should be easy to add the approval check to a new liquidity source.

solver/src/interactions/erc20.rs
nlordell: :see_no_evil: It isn't.

vkgnosis (approved):

PR #520, opened 2021-04-27
1Inch Solver

nlordell:
This PR adds an implementation for a 1Inch solver. It works by selecting a portion of orders and using the 1Inch API to directly get the transaction call data to use for an interaction to swap on 1Inch.

### Test Plan

Added a couple unit tests verifying the `solve` logic. Additionally one ignored test to get an actual settlement for a fictional order:
```
$ cargo test -p solver -- --ignored --nocapture solve_order_on_one
...
running 1 test
Settlement {
...
}
test oneinch_solver::tests::solve_order_on_oneinch ... ok
...
```

An end-to-end test that actually settles against 1Inch will be done when the solver is enabled.

### DIFF
--- a/solver/src/interactions/dummy_web3.rs
+++ b/solver/src/interactions/dummy_web3.rs
@@ -1,7 +1,8 @@
 use contracts::WETH9;
 use jsonrpc_core::Call as RpcCall;
 use serde_json::Value;
-use web3::{api::Web3, types::H160, Transport};
+use shared::transport::LoggingTransport;
+use web3::{api::Web3, transports::Http, types::H160, Transport};
 
 // To create an ethcontract instance we need to provide a web3 even though we never use it. This
 // module provides a dummy transport and web3.
@@ -23,4 +24,18 @@ pub fn dummy_web3() -> Web3<DummyTransport> {
 
 pub fn dummy_weth(addr: impl Into<H160>) -> WETH9 {
     WETH9::at(&dummy_web3(), addr.into())
+}
+
+pub fn infura(network: impl AsRef<str>) -> shared::Web3 {
+    let infura_project_id =
+        std::env::var("INFURA_PROJECT_ID").expect("Missing INFURA_PROJECT_ID env variable");
+    let node_url = format!(
+        "https://{}.infura.io/v3/{}",
+        network.as_ref(),
+        infura_project_id
+    );
+
+    Web3::new(LoggingTransport::new(
+        Http::new(&node_url).expect("transport creation failed"),
+    ))
 }
--- a/solver/src/interactions/dummy_web3.rs
+++ b/solver/src/interactions/dummy_web3.rs
@@ -29,13 +29,15 @@ pub fn dummy_weth(addr: impl Into<H160>) -> WETH9 {
 pub fn infura(network: impl AsRef<str>) -> shared::Web3 {
     let infura_project_id =
         std::env::var("INFURA_PROJECT_ID").expect("Missing INFURA_PROJECT_ID env variable");
-    let node_url = format!(
+    from_node_url(&format!(
         "https://{}.infura.io/v3/{}",
         network.as_ref(),
         infura_project_id
-    );
+    ))
+}
 
+fn from_node_url(node_url: &str) -> shared::Web3 {
     Web3::new(LoggingTransport::new(
-        Http::new(&node_url).expect("transport creation failed"),
+        Http::new(node_url).expect("transport creation failed"),
     ))
 }
--- a/solver/src/liquidity/slippage.rs
+++ b/solver/src/liquidity/slippage.rs
@@ -3,10 +3,10 @@
 use ethcontract::U256;
 
 /// Constant maximum slippage of 10 BPS (0.1%) to use for on-chain liquidity.
-pub const MAX_SLIPPAGE_BPS: u32 = 10;
+pub const MAX_SLIPPAGE_BPS: u16 = 10;
 
 /// Basis points in 100%.
-const BPS_BASE: u32 = 10000;
+const BPS_BASE: u16 = 10000;
 
 /// Apply the constant slippage to the specified amount.
 pub fn amount_with_max_slippage(amount: U256) -> U256 {
--- a/solver/src/oneinch_solver.rs
+++ b/solver/src/oneinch_solver.rs
@@ -152,7 +152,7 @@ mod tests {
     use contracts::{GPv2Settlement, WETH9};
     use ethcontract::H160;
     use hex_literal::hex;
-    use model::order::OrderKind;
+    use model::order::{Order, OrderCreation, OrderKind};
 
     fn dummy_solver() -> OneInchSolver {
         let web3 = dummy_web3::dummy_web3();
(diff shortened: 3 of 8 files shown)


solver/src/oneinch_solver.rs
```diff
@@ -4,3 +4,137 @@
...
+    }
+}
+
+#[async_trait::async_trait]
+impl Solver for OneInchSolver {
+    async fn solve(
+        &self,
+        liquidity: Vec<Liquidity>,
+        _gas_price: f64,
+    ) -> Result<Option<Settlement>> {
```
fleupold, line 117: This should probably return a List now?

solver/src/oneinch_solver.rs
```diff
@@ -4,3 +4,137 @@
...
+            order.sell_token => swap.to_token_amount,
+            order.buy_token => swap.from_token_amount,
+        });
+
+        settlement.with_liquidity(order, order.sell_amount)?;
+
+        if existing_allowance < order.sell_amount {
+            settlement
+                .encoder
+                .append_to_execution_plan(Erc20ApproveInteraction {
```
fleupold, line 92: Maybe settlement can also expose a method to append an arbitrary interaction to the execution plan?

fleupold (commented):
Looks nice and straight forward! Curious if you have already tested it e2e and can share a tx hash with a 1Inch routed settlement?

nlordell:
> Curious if you have already tested it e2e and can share a tx hash with a 1Inch routed settlement?

No, and it appears 1Inch is mainnet only (at least the API appears to be mainnet only).

solver/src/oneinch_solver.rs
```diff
@@ -4,3 +4,221 @@
...
+                .map(|sell_order| self.settle_order(sell_order)),
+        )
+        .await;
+
+        Ok(settlements
+            .into_iter()
+            .filter_map(|settlement| match settlement {
+                Ok(settlement) => Some(settlement),
+                Err(err) => {
+                    // It could be that 1Inch can't find match an order and would
```
fleupold, line 128:
```suggestion
                    // It could be that 1Inch can't match an order and would
```

fleupold (approved):
Looks great!

solver/src/oneinch_solver.rs
```diff
@@ -4,3 +4,221 @@
...
+impl Solver for OneInchSolver {
+    async fn solve(&self, liquidity: Vec<Liquidity>, _gas_price: f64) -> Result<Vec<Settlement>> {
+        let settlements = future::join_all(
+            liquidity
+                .into_iter()
+                .filter_map(|liquidity| match liquidity {
+                    Liquidity::Limit(order) if order.kind == OrderKind::Sell => Some(order),
+                    _ => None,
+                })
+                .take(MAX_SETTLEMENTS)
```
vkgnosis, line 118: It could be interesting to randomize which 5 orders get considered just in case there are some unsolvable orders that always happen to be at the start of the liquidity list.

vkgnosis (approved):

solver/src/oneinch_solver.rs
nlordell: Added. I added the `rand` crate as a dependency (we were already pulling it in anyway).

solver/src/interactions/dummy_web3.rs
```diff
@@ -24,3 +25,19 @@ pub fn dummy_web3() -> Web3<DummyTransport> {
...
+        network.as_ref(),
+        infura_project_id
+    ))
+}
+
+fn from_node_url(node_url: &str) -> shared::Web3 {
+    Web3::new(LoggingTransport::new(
+        Http::new(node_url).expect("transport creation failed"),
+    ))
+}
```
bh2smith, line 43: Is this stuff really being used for dummy web3 related functionality?

solver/src/liquidity/slippage.rs
```diff
@@ -3,10 +3,10 @@
 use ethcontract::U256;
 
 /// Constant maximum slippage of 10 BPS (0.1%) to use for on-chain liquidity.
-pub const MAX_SLIPPAGE_BPS: u32 = 10;
+pub const MAX_SLIPPAGE_BPS: u16 = 10;
 
 /// Basis points in 100%.
-const BPS_BASE: u32 = 10000;
+const BPS_BASE: u16 = 10000;
```
bh2smith, line 9: Does this save us something in the long run, or just good practice.

solver/src/oneinch_solver.rs
```diff
@@ -4,3 +4,232 @@
...
+    fn encode(&self) -> Vec<EncodedInteraction> {
+        vec![(self.tx.to, self.tx.value, self.tx.data.clone())]
+    }
+}
+
+/// Maximum number of sell orders to consider for settlements.
+///
+/// This is mostly out of concern to avoid rate limiting and because 1Inch
+/// requests take a non-trivial amount of time.
+const MAX_SETTLEMENTS: usize = 5;
```
bh2smith, line 107: Nit: Should this be moved to the top of the file?

nlordell:
Merging this PR as the 1Inch solver is currently not functional (just taking up space in the binary :stuck_out_tongue:). A proper E2E test will be done once the solver can be enabled (in a followup PR).

solver/src/oneinch_solver.rs
```diff
@@ -4,3 +4,232 @@
...
+            .get_swap(SwapQuery {
+                from_token_address: order.sell_token,
+                to_token_address: order.buy_token,
+                amount: order.sell_amount,
+                from_address: self.settlement_contract.address(),
+                slippage: Slippage::basis_points(MAX_SLIPPAGE_BPS).unwrap(),
+                // Disable balance/allowance checks, as the settlement contract
+                // does not hold balances to traded tokens.
+                disable_estimate: Some(true),
+            })
```
bh2smith, line 68: I recall that 1inch allows to specify best price or gas efficient as a quote parameter. I would imagine the default is best price and that is also what we would choose on the user's behalf here, but I am not seeing any place where this is passed through to the 1inch client.

bh2smith (approved):
Looks great! Couple questions inline.

solver/src/interactions/dummy_web3.rs
nlordell: Its used for unit tests. I want to move these to a "testing utilities" file but that would touch too much code for this PR and planed on doing it in a follow up.

solver/src/liquidity/slippage.rs
nlordell: No, its just to match the type for the `oninch_solver::api::Slippage` type.

solver/src/oneinch_solver.rs
nlordell: Personally I prefer it here, so its closer to the code that uses it. I don't feel very strongly about it. In general, the order things show up in a file isn't so important, AFAIK there is no "accepted style" for it, and I tend to prefer reading order in general.

solver/src/oneinch_solver.rs
nlordell: That flag is currently not supported by the API implementation. We can definitely add it (but in a separate PR I think).

solver/src/interactions/dummy_web3.rs
nlordell: https://github.com/gnosis/gp-v2-services/pull/537

solver/src/oneinch_solver.rs
nlordell: https://github.com/gnosis/gp-v2-services/issues/538

PR #521, opened 2021-04-27
Use multiple AMMs on same token pair for BaselineSolver

fleupold:
This PR moves `extract_deepest_amm_liquidity` into the Naive Solver and makes the baseline solver operate with the full list of AMMs it can have for a token pair.

This should lead to us always finding the optimal route when deciding between SushiSwap and Uniswap liquidity.

I made the `AmmOrder` implement `BaselineSolvable` by converting itself into a `Pool` and deferring to its methods.

### Test Plan
Added a pool with larger k but worse price to the unit tests.

### DIFF
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -176,6 +176,17 @@ impl TokenPair {
         self.0 == *token || self.1 == *token
     }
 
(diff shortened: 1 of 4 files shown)


bh2smith (approved):

vkgnosis (approved):

nlordell (approved):
Nice!

PR #522, opened 2021-04-28
Split tenderly link error message into its own error

nlordell:
This PR just splits the tenderly alert error into its own log. This is done because including the settlement + tenderly link is too long, and leads to the link getting cut off in the alert message (making it harder to debug issues).

image

To address this, the settlement gets dumped in the following log statement so that it can still be retrieved in the logs.

### Test Plan

Tenderly links in alerts don't get cut off.

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -175,11 +175,8 @@ impl Driver {
             .filter_map(|(settlement, simulation)| match simulation {
                 Ok(()) => Some(settlement),
                 Err(err) => {
-                    tracing::error!(
-                        "simulation of settlement failed\nsettlement: {:?}\n error:{:?}",
-                        settlement.settlement,
-                        err
-                    );
+                    tracing::error!("settlement simulation failed\n error: {:?}", err);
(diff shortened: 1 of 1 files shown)


bh2smith (approved):
Does this mean the debug message will be a subthread in the alert channel or just the next one? I assume it will be the next one or not appear in the alerts channel at all because its only a debug message. I guess then on call would get the error log, go to the actual logs and copy the tenderly link from elsewhere. It would be cute if there was someway to put the tenderly link as a sub message of the alert, but that it probably too much to ask! Nice work here.

nlordell:
> Does this mean the debug message will be a subthread in the alert channel or just the next one?

It will not appear in the channel. I find the tenderly link to be much more useful for debugging solutions (since you get to debug how it is run in the SC). Also the calldata contains the complete settlement anyway and is visible in Tenderly, and you can use @fedgiac's tool https://github.com/gnosis/gp-v2-contracts/pull/596 for a very nice formatted and colorized settlement description.

vkgnosis (approved):

fleupold (approved):

PR #525, opened 2021-04-28
Enforce a minimum order validity period

vkgnosis:
The motivation for this that our solver takes some time to run and the transaction time to get mined. If an order is only valid for a short time we would be including it in the solution but then not be able to use it anymore.
Based on this I thought it would make sense to only consider orders that are still valid for some period when solving and then also to not accept orders that would immediately not be considered because of this.

### Test Plan
changed e2e test to set a realistic validity period

### DIFF
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -17,7 +17,9 @@ use shared::{
     price_estimate::UniswapPriceEstimator,
     transport::LoggingTransport,
 };
-use std::{collections::HashSet, iter::FromIterator as _, net::SocketAddr, sync::Arc};
+use std::{
+    collections::HashSet, iter::FromIterator as _, net::SocketAddr, sync::Arc, time::Duration,
+};
 use structopt::StructOpt;
 use tokio::task;
 use url::Url;
@@ -40,6 +42,15 @@ struct Arguments {
 
     #[structopt(long, env = "FEE_DISCOUNT_FACTOR", default_value = "1")]
     fee_discount_factor: f64,
+
+    /// The minimum amount of time an order has to be valid for.
+    #[structopt(
+        long,
+        env = "MIN_ORDER_VALIDITY_PERIOD",
+        default_value = "60",
+        parse(try_from_str = shared::arguments::duration_from_seconds),
+    )]
+    min_order_validity_period: Duration,
 }
 
 pub async fn orderbook_maintenance(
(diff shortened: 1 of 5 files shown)


orderbook/src/orderbook.rs
```diff
@@ -80,7 +84,9 @@ impl Orderbook {
         if self.unsupported_tokens.contains(&order.sell_token) {
             return Ok(AddOrderResult::UnsupportedToken(order.sell_token));
         }
-        if !has_future_valid_to(shared::time::now_in_epoch_seconds(), &order) {
+        if order.valid_to
+            < shared::time::now_in_epoch_seconds() + self.min_order_validity_period.as_secs() as u32
```
fleupold, line 88: We should probably spare a separate error code or rename past valid to to `InusfficientValidTo` to not confuse users of the API.

orderbook/src/main.rs
```diff
@@ -40,6 +42,15 @@ struct Arguments {
 
     #[structopt(long, env = "FEE_DISCOUNT_FACTOR", default_value = "1")]
     fee_discount_factor: f64,
+
+    /// The minimum amount of time an order has to be valid for.
+    #[structopt(
+        long,
+        env = "MIN_ORDER_VALIDITY_PERIOD",
+        default_value = "120",
```
fleupold, line 50: nit (and good that this is configurable): I'd personally say 60s would be sufficient in case a market maker wanted to specifically place a counter-order against an order they found in the orderbook.

fleupold (approved):

orderbook/src/orderbook.rs
vkgnosis: Agree. Changed to that name.

orderbook/src/main.rs
vkgnosis: Changed.

nlordell (approved):
Short and sweet.

fleupold:
Not sure if this had happened already before, but it just 30 minutes ago...

https://gnosis-protocol.io/orders/0x6deab0237ed71829f1a7dbfdd94b7da19cebf2c7aae03e22a317e66a03c4648dcfcdcde4d1c654ed980b4b3f6ff782194b3aa666608ab05b

Nice foresight!

vkgnosis:
Cool. I just remembered we had at least one instance of simulation failing because of an expired order.

PR #526, opened 2021-04-28
Sends network_id to http solver.

marcovc:
General rust question related with this PR:

In this PR I 'm passing a String around, since last time I asked I got the feeling that pass-by-move is more idiomatic than pass-by-reference.

However say, I have a couple of functions:

```rust
 f_1(s: String)
 f_2(s: String)
etc
```

I want to pass the same string to all of them.

```rust
let s: String = "blabla";
f_1(s);
f_2(s);
etc
```

The above doesn't compile since I can't move s multiple times.

So it appears that in this case my `f_i` should take a reference instead, to avoid cloning a bunch of times. Right? (assume that s is a really long string, or some object that is costly to clone).

If so, this means that whenever I am defining a function I have to think about what else the calling function(s) will do with the parameters I'm receiving in order to decide if I should take a reference or a moved value. That doesn't seem nice, so what am I missing?


vkgnosis:
Rust forces you think to think about ownership. So in these cases the question is whether the functions need to own the the thing. If you don't need an owned value then usually you should use a reference. This is a little more complicated for async code or storing references in structs because then Rust also forces you to think about lifetime which can be tedious. Escape hatches are clowning or using reference counting like Rc, Arc.

> what else the calling function(s) will do with the parameters 

It should depend on the callee not the caller.

vkgnosis (approved):

solver/src/http_solver.rs
```diff
@@ -244,6 +247,9 @@ impl HttpSolver {
             tokens: token_models,
             orders: order_models,
             uniswaps: uniswap_models,
+            metadata: Some(MetadataModel {
+                environment: Some(self.network_id.clone()),
```
fleupold, line 251: nit: given this is already a string maybe we can translate this into human readable network names (mainnet, rinkeby, xdai)?

fleupold (approved):

solver/src/http_solver.rs
marcovc: I thought that network_id was already "mainnet", "rinkeby", etc. Is there a mapping somewhere in the code?

fleupold: If you check where `network_id` is set in the code, you will find that it queries the node for [net_version](https://eth.wiki/json-rpc/API) which returns a string that is a number.

I don't think we have that translation logic (1 => Mainnet, 4=> Rinkeby etc.) yet, but it might be a nice thing to add to the utils crate.

nlordell: naming-nit: Maybe call this `network` instead of `environment`? To me `environment` is more like "production"/"staging".

nlordell (approved):
Yay!

marcovc:
> To me environment is more like "production"/"staging".

That was on purpose. In the future we might want to add "production/staging" or which L2 are we in, or whatever, and for the solver it is just a string with runtime information. The other option would be to discriminate each of these things in the solver API, but I don't think that is better.

nlordell:
> In the future we might want to add "production/staging" or which L2 are we in, or whatever,

Makes sense. We should leave it as `environment` then.

solver/src/http_solver.rs
marcovc: > I don't think we have that translation logic (1 => Mainnet, 4=> Rinkeby etc.) yet, but it might be a nice thing to add to the utils crate.

Sure. Do you mean the "shared" subproject? Or is the "utils" crate on a different repo?

fleupold: Wherever it's convenient in this repo. There is solver/src/utils.rs e.g.

PR #528, opened 2021-04-29
Add eth equivalence to fee estimate

fedgiac:
Return valid fee estimates when buying native eth.

### Test Plan

Start database + services pointing to Rinkeby, then:

```
$ sell_dai_for() {
   curl -X 'GET' \
     'http://localhost:8080/api/v1/fee?sellToken=0x5592ec0cfb4dbc12d3ab100b257153436a1f0fea&buyToken='"$1"'&amount=12&kind=buy' \
     -H 'accept: application/json'
   echo
 }
$ sell_dai_for 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE # eth
{"expirationDate":"2021-04-29T10:59:27.939324397Z","amount":"256281330953542944"}
$ sell_dai_for 0xc778417E063141139Fce010982780140Aa0cD5Ab # weth
{"expirationDate":"2021-04-29T10:59:27.972065197Z","amount":"256281330953542944"}
$ sell_dai_for 0x5592ec0cfb4dbc12d3ab100b257153436a1f0fea # dai
{"expirationDate":"2021-04-29T10:59:27.999053809Z","amount":"0"}

```


nlordell (approved):
Looks great!

vkgnosis (approved):

orderbook/src/fee.rs
```diff
@@ -84,6 +84,12 @@ impl MinFeeCalculator {
         amount: Option<U256>,
         kind: Option<OrderKind>,
     ) -> Result<Option<Measurement>> {
+        let buy_token = if buy_token == Some(BUY_ETH_ADDRESS) {
+            Some(self.native_token)
+        } else {
+            buy_token
+        };
+
```
fleupold, line 92: I think we need the same logic inside `PriceEstimating` (maybe even only there is fine as I believe MinFeeCalculating calls into it)

fleupold (commented):
Not sure if adding this logic in price estimating should be separate PR or could even replace this one?

bh2smith (approved):

bh2smith:
Looking back on this it seems like a fine thing to do but it also isn't really all that much additional effort for a front end to simply query the WETH token for price estimates when buying ETH right?

PR #534, opened 2021-04-30
Add pod file and swap UI docker file

nlordell:
This PR introduces a pod file for starting up an entire GPv2 system (including CowSwap) for local debugging.

This PR is still a work in progress, just wanted to share what I have so far.

### Test Plan

TBD

### DIFF
--- a/docker/debug/pod.yaml
+++ b/docker/debug/pod.yaml
@@ -5,13 +5,23 @@ metadata:
 
 spec:
   containers:
-    - name: api
-      image: gnosispm/gp-v2-services
-      args: ["orderbook"]
-
     - name: solver
       image: gnosispm/gp-v2-services
       args: ["solver"]
+      env:
+        - name: PRIVATE_KEY
+          value: "0x4f3edf983ac636a65a842ce7c78d9aa706d3b113bce9c46f30d7d21715b23b1d"
+        - name: AMM_SOURCES
+          value: Uniswap
+
+    - name: api
+      image: gnosispm/gp-v2-services
+      args: ["orderbook"]
+      env:
+        - name: DB_URL
+          value: "postgresql://localhost/?user=me&password="
+      ports:
+        - hostPort: 8080
 
     - name: migrations
       image: gnosispm/gp-v2-services-migrations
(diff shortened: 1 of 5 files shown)


vkgnosis:
Ohh even with the UI. That is cool.

docker/debug/swapui.env
```diff
@@ -0,0 +1,21 @@
...
+REACT_APP_NETWORK_URL_100=https://rpc.xdaichain.com
+REACT_APP_NETWORK_URL_1337=http://localhost:8545
+
+# Domain
+REACT_APP_DOMAIN_REGEX_DEV=".*"
+
+# API
+REACT_APP_API_URL_STAGING_MAINNET=http://localhost:8080/api
+REACT_APP_API_URL_STAGING_RINKEBY=http://localhost:8080/api
+REACT_APP_API_URL_STAGING_XDAI=http://localhost:8080/api
```
fleupold, line 18: Does this need one for 1337 as well?

docker/debug/pod.yaml
```diff
@@ -0,0 +1,52 @@
...
+spec:
+  containers:
+    - name: solver
+      image: gnosispm/gp-v2-services
+      args: ["solver"]
+      env:
+        - name: PRIVATE_KEY
+          value: "0x4f3edf983ac636a65a842ce7c78d9aa706d3b113bce9c46f30d7d21715b23b1d"
+        - name: AMM_SOURCES
+          value: Uniswap
```
fleupold, line 15: Why that restriction? Because sushi swap is not yet deployed?

fleupold (commented):
Can you add a section to the readme how to use this? I think we commonly want to run at least one service on the host system so that we can set break points and change code without having to rebuild the container.

docker/debug/pod.yaml
nlordell: We don't deploy SushiSwap on Ganache. (with `cargo run --bin deploy`).

docker/debug/swapui.env
nlordell: Good question. I'll dig around.

nlordell:
> Can you add a section to the readme how to use this?

For sure!

PR #539, opened 2021-04-30
Log settlement simulation errors only if fails with original block

### LINKED ISSUE #527: Suppress transaction revert errors for `UniswapV2Router: INSUFFICIENT_OUTPUT_AMOUNT`
When a transaction simulation succeeds, but then proceeds to revert with `UniswapV2Router: INSUFFICIENT_OUTPUT_AMOUNT`, we should suppress the alert and simply use Prometheus metrics to count this:
- This would give us a better overall picture of how many transactions revert because of price movement
- Reduces noise in the alerts channel.

vkgnosis:
It is annoying when we get alerts for settlements that failed to
simulate on the current block at time of submission but would have
succeeded at time of liquidity fetching.
Here the solver worked correctly but the chain moved in unfavorably.
This is not an error so it doesn't make sense to alert.

Fixes #527

### Test Plan
Ran the adjusted test with real node to see that tenderly link is still generated.
report_simulation_errors doesn't influence the correctness of the solver so I am planning on manually checking the logs

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -9,10 +9,11 @@ use crate::{
     settlement_simulation, settlement_submission,
     solver::Solver,
 };
-use anyhow::{Context, Result};
+use anyhow::{Context, Error, Result};
 use contracts::GPv2Settlement;
 use futures::future::join_all;
 use gas_estimation::GasPriceEstimating;
+use itertools::{Either, Itertools};
 use num::BigRational;
 use primitive_types::H160;
 use shared::{price_estimate::PriceEstimating, Web3};
@@ -125,11 +126,11 @@ impl Driver {
         }
     }
 
-    // Returns only settlements for which simulation succeeded.
+    // Split settlements into successfully simulating ones and errors.
     async fn simulate_settlements(
         &self,
         settlements: Vec<RatedSettlement>,
-    ) -> Result<Vec<RatedSettlement>> {
+    ) -> Result<(Vec<RatedSettlement>, Vec<(RatedSettlement, Error)>)> {
         let simulations = settlement_simulation::simulate_settlements(
             settlements
                 .iter()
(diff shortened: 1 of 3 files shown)


solver/src/driver.rs
```diff
@@ -125,38 +126,89 @@ impl Driver {
...
+                Ok(()) => continue,
+                Err(err) => err,
+            };
+            tracing::error!(
+                "settlement simulation failed right before submission AND for block {} which was current when liquidity was fetched:\n{:?}",
+                current_block_during_liquidity_fetch, error_at_earlier_block
+            );
+            // This is an additional debug log so that the log message doesn't get too long as
+            // settlement information is recoverable through tenderly anyway.
+            tracing::debug!("settlement failure for: \n{:#?}", settlement.settlement,);
```
fleupold, line 198: maybe still warn level?

solver/src/driver.rs
```diff
@@ -125,38 +126,89 @@ impl Driver {
...
+        }
     }
 
     pub async fn single_run(&mut self) -> Result<()> {
         tracing::debug!("starting single run");
         let liquidity = self.liquidity_collector.get_liquidity().await?;
+        let current_block_during_liquidity_fetch = self
```
fleupold, line 205: I think it's an important follow up task to pass this block into the liquidity fetcher to make sure all pools are queried at the same block.

solver/src/driver.rs
```diff
@@ -125,38 +126,89 @@ impl Driver {
...
+                return;
+            }
+        };
+
+        for ((settlement, _previous_error), result) in errors.into_iter().zip(simulations) {
+            let error_at_earlier_block = match result {
+                Ok(()) => continue,
+                Err(err) => err,
+            };
+            tracing::error!(
```
fleupold, line 192: Maybe we can even "store" the settlement in some form and avoid logging an error multiple times for the same settlement/order?

solver/src/driver.rs
```diff
@@ -125,38 +126,89 @@ impl Driver {
...
+                    current_block_during_liquidity_fetch,
+                    err
+                );
+                return;
+            }
+        };
+
+        for ((settlement, _previous_error), result) in errors.into_iter().zip(simulations) {
+            let error_at_earlier_block = match result {
+                Ok(()) => continue,
```
fleupold, line 189: We should still log these cases in some metric (separate PR).

solver/src/driver.rs
```diff
@@ -125,38 +126,89 @@ impl Driver {
...
+            &self.settlement_contract,
+            &self.web3,
+            &self.network_id,
+            settlement_simulation::Block::FixedWithTenderly(current_block_during_liquidity_fetch),
+        )
+        .await
+        {
+            Ok(simulations) => simulations,
+            Err(err) => {
+                tracing::error!(
```
fleupold, line 178: When does this error happen? I initially confused this for "simulation failed" but it's more of a meta level error isn't it? Maybe we can make it a bit clearer in the error message.

solver/src/settlement_simulation.rs
```diff
@@ -1,12 +1,20 @@
...
 use ethcontract::{batch::CallBatch, dyns::DynTransport, transaction::TransactionBuilder};
 use futures::FutureExt;
 use shared::Web3;
+use web3::types::{BlockId, BlockNumber};
 
 const SIMULATE_BATCH_SIZE: usize = 10;
 
+pub enum Block {
+    // Simulate the transactions at this block and attach a tenderly link to errors.
+    FixedWithTenderly(u64),
```
fleupold, line 13: nano-naming-nit: Sounds to me like the simulation is done via tenderly (but any more descriptive name like FixedWithTenderlyLinkErrorContext sounds a bit silly).

fleupold (approved):
I think this should go together with passing a block into the liquidity fetcher to make sure we are actually simulating on the real state.

Also we still need some way to track how often we become victim of race-conditions so we can react if this number is too high.

These two things can happen in separate PRs though.

solver/src/settlement_simulation.rs
vkgnosis: Agree name is not perfect which is why I added the comment above. Imo good enough.

vkgnosis:
Will make an issue to track the follow up tasks you mentioned.

PR #540, opened 2021-04-30
[Config Changes] Enable 1Inch solver

nlordell:
This PR adds the final glue code to enable the 1Inch solver :tada:.

### Test Plan

Ran a successful E2E test on mainnet:
https://etherscan.io/tx/0xbdc26ed8dfb6aa45a90f990f1e3f15238140e7ff44a71a23e7135b9673d073af

Notice that we traded GNO on Uniswap over Bancor

### DIFF
--- a/solver/src/oneinch_solver.rs
+++ b/solver/src/oneinch_solver.rs
@@ -73,6 +73,10 @@ impl OneInchSolver {
             })
             .await?;
 
+        ensure!(
+            swap.to_token_amount >= order.buy_amount,
+            "order limit price not respected",
+        );
         let mut settlement = Settlement::new(hashmap! {
             order.sell_token => swap.to_token_amount,
             order.buy_token => swap.from_token_amount,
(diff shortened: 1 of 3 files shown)


solver/src/oneinch_solver.rs
```diff
@@ -92,6 +101,20 @@ impl OneInchSolver {
     fn web3(&self) -> DynWeb3 {
         self.settlement_contract.raw_instance().web3()
     }
+
+    async fn chain_id(&self) -> Result<u64> {
```
fleupold, line 105: We already load the chain_id on startup [here](https://github.com/gnosis/oba-services/blob/bed59910aedace2e04fd28e7d04e1ed5a5114f78/solver/src/main.rs#L132). Can we not pass it in from above?

fleupold (approved):

solver/src/oneinch_solver.rs
nlordell: That makes a lot of sense.

vkgnosis (approved):

bh2smith (approved):

PR #542, opened 2021-04-30
Remove legacy fee

fleupold:
This PR removes the legacy fee route and makes the "fee verification logic" upon order placement verify that the fee is actually valid for the given path / amount.

We should keep this change in staging for a while and make sure trades are still happening / we are not seeing errors because of fee estimation errors (as the check is now stricter than before)

### Test Plan
Unit tests are still passing.

### DIFF
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -156,27 +156,6 @@ paths:
           description: Invalid signature
         404:
           description: Order was not found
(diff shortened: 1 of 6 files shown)


vkgnosis (approved):

bh2smith (approved):
This is so much better 🧹 🗜️

PR #547, opened 2021-05-03
SushiSwap Price Estimation

### LINKED ISSUE #546: [Price Estimation] Integrate Sushiswap
Initially, I had thought that the changes could be envisioned as follows:

- the main `fetch` method for `PoolFetcher`: 
    `PoolFetcher` takes a list of `AmmPairProvider` (one Uniswap and one Sushiswap) and the fetch method returns still `Vec<Pool>` but then there could/would be multiple pools for the same `TokenPair`.
- `UniswapPriceEstimator`: 
    would be renamed to `UniswapLikePriceEstimator` and updated to handle the case where two pools for the same `TokenPair` exist. In particular the `best_execution` method which makes used of `PoolFetcher::fetch` will be adjusted to handle the case of pools with multiplicity.

However, by simply renaming `UniswapPriceEstimator` to `UniswapLikePriceEstimator`, we could just instantiate one of these for each pair provider (sushi and uni) then implement `PriceEstimating` for a new struct like `MultiPriceEstimator` which consists of multiple `UniswapLikePriceEstimators`. From here, we would implement the trait methods on the values returned by the individual estimates. This would be more robust for the future, when incorporating `PriceEstimating` for sources with different logic, this may be the way to go.

The problem with the apparently more clever option is that having the independent pool fetchers means that the routes for price estimate are sticly bound to be contained in a single platform. Meaning, if the best route uses one Uniswap and one sushiswap pool. this price estimate wouldn't be found. In order to truely provide the best price estimate, the pools need to be mixed so to get the best price of each pool along the route.

Looking for suggestions on the proposal @fleupold, @vkgnosis...

bh2smith:
Fixes #546

Introducing `PoolAggregator` which can handle a collection of PoolFetchers (e.g. Sushi and Uniswap) so to aggregate different sources for price estimation. This PR also introduces Sushiswap for price estimation in both the orderbook api and solver code.

## Test Plan:

Query the Price for selling 1000 WETH for SUSHI via:
https://protocol-mainnet.dev.gnosisdev.com/api/v1/markets/0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2-0x6B3595068778DD592e39A122f4f5a5cF09C90fE2/sell/1000000000000000000000

Get response like:
```
{
  "amount": "99632636745739619801741",
  "token": "0x6b3595068778dd592e39a122f4f5a5cf09c90fe2"
}
```

Run the orderbook locally with 

```sh
cargo run --bin orderbook -- --node-url https://mainnet.infura.io/v3/YOUR-INFURA-KEY
```
and make the same query to your local instance

http://localhost:8080/api/v1/markets/0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2-0x6B3595068778DD592e39A122f4f5a5cF09C90fE2/sell/1000000000000000000000

Getting

```
{
  "amount": "241969510572526553848496",
  "token": "0x6b3595068778dd592e39a122f4f5a5cf09c90fe2"
}
```

Which is an order of magnitude larger.
```
 99632636745739619801741
241969510572526553848496
```


shared/src/price_estimate.rs
```diff
@@ -268,6 +268,69 @@ impl UniswapPriceEstimator {
...
+        for estimate in estimates {
+            if let Ok(price) = estimate {
+                valid_estimates.push(price)
+            }
+        }
+
+        match valid_estimates.len() {
+            0 => Err(anyhow!("Failed to estimate price on all estimators")),
+            _ => Ok(valid_estimates.iter().min().unwrap().clone()),
+        }
```
vkgnosis, line 302: Here is an equivalent shorter version. Can probably do something like this for the version below too.
```suggestion
        estimates
            .into_iter()
            .filter_map(Result::ok)
            .min()
            .ok_or_else(|| anyhow!("Failed to estimate prices on all estimators"))
```

fleupold (commented):
I think we should build a "mix and match" approach like we did for the solver already. Eventually it would be nice if our baseline solver could even split hops between multiple liquidity source (e.g. trading 60% on Uni and 40% on Sushi) if this make sense pricewise.

orderbook/src/main.rs
```diff
@@ -97,6 +99,9 @@ async fn main() {
     let uniswap_factory = UniswapV2Factory::deployed(&web3)
         .await
         .expect("couldn't load deployed uniswap factory");
+    let sushiswap_factory = SushiswapV2Factory::deployed(&web3)
+        .await
+        .expect("couldn't load deployed uniswap factory");
```
bh2smith, line 104:
```suggestion
        .expect("couldn't load deployed sushiswap factory");
```

shared/src/pool_collector.rs
```diff
@@ -0,0 +1,34 @@
...
+use model::TokenPair;
+use std::collections::HashSet;
+use std::sync::Arc;
+
+pub struct PoolCollector {
+    pub pool_fetchers: Vec<PoolFetcher>,
+}
+
+impl PoolCollector {
+    pub fn new(pair_providers: Vec<Arc<dyn AmmPairProvider>>, web3: Web3) -> Self {
```
fleupold, line 13: Here I would pass in a list of PoolFetchers directly. ~This way we can reuse the `FilteredPoolFetcher<CachedPoolFetcher<PoolFetcher>>` layered component we already create in main.~

Actually, I see now how those are layered on top. I think we might need a separate trait interface when we want to generalize from just fetching `Pools` to fetching arbitrary `BaselineSolvable` objects, but we can cross this bridge when we get there. 

Yet, to leave this component very abstract (it aggregates different `impl PoolFetching` together), we could pass a `Vec<Arc<dyn PoolFetching>>`?

shared/src/pool_collector.rs
```diff
@@ -0,0 +1,34 @@
...
+                pair_provider,
+                web3: web3.clone(),
+            })
+        }
+        Self { pool_fetchers }
+    }
+}
+
+#[async_trait::async_trait]
+impl PoolFetching for PoolCollector {
```
fleupold, line 26: This is very clever!

shared/src/pool_collector.rs
```diff
@@ -0,0 +1,34 @@
...
+use crate::Web3;
+use model::TokenPair;
+use std::collections::HashSet;
+use std::sync::Arc;
+
+pub struct PoolCollector {
+    pub pool_fetchers: Vec<PoolFetcher>,
+}
+
+impl PoolCollector {
```
fleupold, line 12: naming nit: `PoolAggregator`?

solver/src/main.rs
```diff
@@ -169,14 +169,15 @@ async fn main() {
...
+    });
+    let price_estimator = Arc::new(BaselinePriceEstimator::new(
         Box::new(PoolFetcher {
-            pair_provider: Arc::new(UniswapPairProvider {
-                factory: contracts::UniswapV2Factory::deployed(&web3)
-                    .await
-                    .expect("couldn't load deployed uniswap router"),
-                chain_id,
-            }),
+            pair_provider: uniswap_pair_provider,
```
fleupold: We should probably also use a filtered/cached pool fetcher here (but can be in a separate PR).

solver/src/main.rs
```diff
@@ -169,14 +169,15 @@ async fn main() {
     .expect("failed to create gas price estimator");
 
     // TODO - Currently we use only Uniswap for price estimation.
-    let price_estimator = Arc::new(UniswapPriceEstimator::new(
+    let uniswap_pair_provider = Arc::new(UniswapPairProvider {
```
fleupold, line 171: Let's also use Sushiswap here.

bh2smith: I think I have already made this change... Is it possible this comment is already outdated?

shared/src/pool_collector.rs
bh2smith: Right, if we are going to do that then there really is no need for this new method since its only trying to reuse the web3 for `AmmPairProviders`. In that case I will just remove the new method so not to favour one implementation of PoolFetcher over another.

solver/src/main.rs
bh2smith: Yes, I agree although at this moment I believe we are. I will get this PR fixed up and in order and revist if this is not the case. I would like to include a TODO in the right place so I can find and and do it immediately after this is merged.

solver/src/main.rs
bh2smith: Yes. Sushiswap is directly below

contracts/artifacts/IUniswapLikePair.json
```diff
@@ -0,0 +1,1431 @@
+{
```
vkgnosis, line 1: Where does this file come from? Our other contracts abis are defined in `contracts/src/bin/vendor.rs` but this one is not.

bh2smith: Oh... So I am supposed to provide a source for it here? I had just pasted it in there ad I also did with `IUniswapLikeRouter` am I supposed to do something different?

bh2smith: This is just an interface contract that I got from here:

@uniswap/v2-core/contracts/interfaces/IUniswapV2Pair.sol

vkgnosis: If it's not in a npm package or something like that it's fine I guess. I was just curious. Not super important, but I think it would be nice to minimize the json though since we won't be editing it by hand. The vendored artifacts are all only one line without whitespaces.

bh2smith: I guess I'll look and see if I can do that. I suppose removing all the white space would clear that up.

fleupold: Good catch @vkgnosis . I believe this artefact can be vendored from npm just like https://github.com/gnosis/oba-services/blob/10e301a3122b5437cc4e29065d4b74c3085441c8/contracts/src/bin/vendor.rs#L30-L33. It looks like the second value in the `NPM_CONTRACTS` lists can be used to create filenames with arbitrary name overrides (e.g. IUniswapLikePair.json)

There is also a README with more instructions.

fleupold: https://github.com/gnosis/gp-v2-services/pull/559 for reference (there were some issues with the vendor script which require to update the patched ethcontract version)

shared/src/pool_aggregating.rs
```diff
@@ -0,0 +1,68 @@
...
+            let pair_provider: Arc<dyn AmmPairProvider>;
+            match source {
+                PriceEstimationSources::Uniswap => {
+                    pair_provider = Arc::new(UniswapPairProvider {
+                        factory: contracts::UniswapV2Factory::deployed(&web3)
+                            .await
+                            .expect("couldn't load deployed uniswap router"),
+                        chain_id,
+                    });
+                    tracing::info!("Constructed Uniswap Pair Provider");
```
bh2smith, line 38: Not sure if these logs and the two that follow are really all that necessary, but they helped me debug, will remove if requested.

shared/src/arguments.rs
```diff
@@ -50,6 +50,17 @@ pub struct Arguments {
     /// List of token addresses to be ignored throughout service
     #[structopt(long, env = "UNSUPPORTED_TOKENS", use_delimiter = true)]
     pub unsupported_tokens: Vec<H160>,
+
+    /// Which Liquidity sources to be used by Price Estimator.
+    #[structopt(
```
fleupold, line 55: Here I'd say we should reuse AMM_SOURCES (maybe it can be renamed to BASELINE_SOURCES)

shared/src/pool_aggregating.rs
fleupold: They only get logged once on startup, but I wouldn't consider them useful in prod. Please remove.

solver/src/main.rs
```diff
@@ -100,7 +100,7 @@ struct Arguments {
 
     /// Which AMM sources to use. Multiple sources are supported alone or simultaneously
     #[structopt(
-    long,
+        long,
```
fleupold, line 103: Let's remove this one in favor of a shared config BASELINE_SOURCES

fleupold (approved):
LGTM, I think we should have the same set of Liqiudity sources for price estimation and baseline solver so no need to differentiate with different config values imo.

vkgnosis (approved):

PR #548, opened 2021-05-03
Fix switch up of owner and token address on new order

vkgnosis:
The order of the arguments was wrong. This issue did not surface because
when we fetch balances later we insert missing owner-token pairs anyway.
I found this because I was wondering why we had a lot of "unable to
fetch allowance" errors in the logs.

### Test Plan
see that we get less of the errors in the logs afterwards

### DIFF
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -126,7 +126,7 @@ impl Orderbook {
             _ => (),
         }
         self.balance_fetcher
-            .register(order.order_creation.sell_token, order.order_meta_data.owner)
+            .register(order.order_meta_data.owner, order.order_creation.sell_token)
             .await;
         Ok(AddOrderResult::Added(order.order_meta_data.uid))
     }


orderbook/src/orderbook.rs
```diff
@@ -126,7 +126,7 @@ impl Orderbook {
             _ => (),
         }
         self.balance_fetcher
-            .register(order.order_creation.sell_token, order.order_meta_data.owner)
+            .register(order.order_meta_data.owner, order.order_creation.sell_token)
```
fleupold, line 129: This is such a classic error (same type, unnamed arguments) 🙈. Maybe we should create a named struct instead to make sure this doesn't happen again in the future

fleupold (approved):

orderbook/src/orderbook.rs
vkgnosis: Personally I don't think it's needed but I'm not very against it either.

bh2smith: So... how were we ever getting correct information somtimes?

PR #549, opened 2021-05-03
Attempts to execute settlements from liquidity internal buffers first.

marcovc:
This PR tries a simple strategy to use our own buffers instead of external liquidity provision. 

This would be step 0 from the following plan:

0 - Quick and dirty solution where we simply simulate the settlement without any interactions, and go with it if it succeeds. Pros: can save gas and uniswap fees. Cons: not optimal - see next step.

1 - Partially fill excess orders from our buffers, and send the remaining to uniswap. Pros: can safe more uniswap fees.

2 - Additionally rebalance in case we have empty buffers. Pros: our buffers will track external exchange rates / rebalancing + fullfilling the user order costs the same gas in most situations.

Step 1 will likely rewrite much of this PR, but lets start simple.


solver/src/driver.rs
```diff
@@ -197,7 +201,51 @@ impl Driver {
             self.min_order_age,
             &mut settlements,
         );
-        let settlements = self.simulate_settlements(settlements).await?;
+
+        // Remove liquidity provision interactions from settlements.
```
fleupold, line 205: Instead of adding all this code here could we maybe do the extra simulation just before we submit the settlement?


```rust
if let Some(settlement) = settlements
  .into_iter()
  .max_by(|a, b| a.objective_value.cmp(&b.objective_value))
{
  // If we have enough buffer in the settlement contract to not use on-chain interactions, remove those
  if can_settle(settlement.without_liquidity()) {
    settlement = settlement.without_liquidity()
  }
  self.submit_settlement(settlement).await;
}
```


I think it depends a little bit how phase 1 and 2 would be implemented. Would the logic of which/how to replace interactions go in the SettlementEncoder or would this logic here stay around?

marcovc: Right, I was getting ahead of myself since on the second phase we won't need the extra simulation, and will do everything before the final simulation. But I agree it makes more sense that we don't make it unnecessarily complex in this PR. Done in c9118ec.

solver/src/driver.rs
```diff
@@ -125,6 +126,18 @@ impl Driver {
         }
     }
 
+    async fn can_settle(&self, settlement: RatedSettlement) -> Result<bool> {
+        let simulations = settlement_simulation::simulate_settlements(
+            chain![settlement.settlement.clone().into()],
```
fleupold, line 131: nanonit: maybe Rated settlement could become clone directly and implement Into<EncodedSettlement> to avoid this long chain.

solver/src/driver/solver_settlements.rs
```diff
@@ -37,6 +37,15 @@ pub struct RatedSettlement {
     pub objective_value: BigRational,
 }
 
+impl RatedSettlement {
+    pub fn without_liquidity(&self) -> Self {
```
fleupold, line 41: Naming nit: `without_onchain_liquidity` ?

fleupold (approved):
Looks good to me for phase 0. Did you test this yet (maybe locally connecting to rinkeby), we could also add an e2e test where we transfer some of the buy token to the smart contract beforehand.

We also need to make sure we withdraw WETH from our settlement contract before merging this into prod (@fedgiac might be able to help with that as he wrote a script to do so).

marcovc:
> we could also add an e2e test where we transfer some of the buy token to the smart contract beforehand.

I'll try that.

> We also need to make sure we withdraw WETH from our settlement contract before merging this into prod (@fedgiac might be able to help with that as he wrote a script to do so).

Why we'd want to do that? Are you concerned that this could consume all the WETH there due to a bug, or do you want to limit the amount of buffer that can be used for some other reason?

solver/src/driver.rs
```diff
@@ -134,7 +148,7 @@ impl Driver {
         let simulations = settlement_simulation::simulate_settlements(
             settlements
                 .iter()
-                .map(|settlement| settlement.settlement.clone().into()),
+                .map(|settlement| settlement.clone().into()),
```
bh2smith, line 151: I am surprised settlement.settlement was actually doing anything before....

bh2smith (approved):
Looks cool, but I see we use `without_onchain_liquidity` on a settlement rather than an individual order. what happens if the settlement happens to consist of a collection of orders and only one of those orders can be settled with the buffer? Wouldn't we rather determine this buffer usage on a per-order basis rather than an entire settlement?

marcovc:
> what happens if the settlement happens to consist of a collection of orders and only one of those orders can be settled with the buffer? Wouldn't we rather determine this buffer usage on a per-order basis rather than an entire settlement?

If you mean that we could do better than what is currently in the PR, then I agree. This is a all-or-nothing optimization in the sense that we either do the full settlement without external liquidity, or we use all external liquidity as before. The idea was to do more clever things in subsequent PRs.

fleupold:
> Why we'd want to do that?

Because ETH is the only token we hold a large amount in at the moment and we might not want to try out this feature with a large attack vector.

solver/src/driver.rs
marcovc: The outer settlement is a "RatedSettlement { settlement: Settlement }", and now there is a coercion between the two by means of the  new "From<>" trait.

marcovc:
I've added a test that is failing. It essentially tries to trade between one trader and the solver, and it is failing since apparently the order is not cleared. Any hints what this could be?

solver/src/settlement/settlement_encoder.rs
```diff
@@ -68,6 +68,17 @@ impl SettlementEncoder {
         result
     }
 
+    // Returns a copy of self without any liquidity provision interaction.
+    pub fn without_onchain_liquidity(&self) -> Self {
```
josojo, line 72: Really cool PR! great start!

Maybe the next step is to write a fn: `without_onchain_stablecoin_liquidity` where we just replace the onchain stable coin liquidity but leave all other interactions, because the current solution might not work on WETH->USDC->DAI trading pairs if the buffer is not sufficient on WETH->USDC.

vkgnosis:
> I've added a test that is failing. It essentially tries to trade between one trader and the solver, and it is failing since apparently the order is not cleared. Any hints what this could be?

Looking at the logs I think the reason might be that we don't clear the database after the first e2e test. We need to call `Database::clear` (from `orderbook/src/database.rs`) somewhere. Maybe inside `Ganache::test` as that is already responsible for ensuring multiple e2e tests don't conflict e2e/tests/ganache.rs .

solver/src/settlement/settlement_encoder.rs
marcovc: It can be even better. We can also use buffers for one order even if the order is not totally covered by them since that will save us uniswap fees.

marcovc:
Thanks @vkgnosis . I've tried inserting a couple of Database::clear here and there but it doesn't seem to work. Of course doing

```rust
    let db = orderbook::database::Database::new("postgresql://").unwrap();
    db.clear().await.unwrap();
```

right before the failing assertion makes the test pass, but it looks to me that it would defeat the purpose of the assertion.

I guess I don't know enough about the overall arch to debug this:

* What triggers an order deletion in the orderbook? 
* If orderbook is reacting to the blockchain, then `driver.single_run` should update the blockchain that in turn should update the orderbook. Right?

fleupold:
> I guess I don't know enough about the overall arch to debug this:
> 1. What triggers an order deletion in the orderbook?
>  2. If orderbook is reacting to the blockchain, then driver.single_run should update the blockchain that in turn should update the orderbook. Right?

When the order is traded, the blockchain emits a `Trade` event which is processed by the orderbook in the  `run_maintenance` method. If your order was fully traded (which you would expect in this test) it would be removed.

My first step at debugging this was to `assert!(orders.len() == 1);` before the run (which passes) to make sure we are not dealing with old state from a previous test.

Next, I set a breakpoint on the failing line and when I inspect the returned order it looks like the one we created. On the metadata I can see that it wasn't traded at all. Also, the `// Check matching` section in the test case checks that the balances have not changed as far as I can tell (trade should however have 100 of token B afterwards).

From that, I'd say the test just fails because we don't trade that order. Also this log line:

> solver::driver: 0 settlements passed simulation and 0 failed

You can set further breakpoints in the driver to see why no solutions are returned.

marcovc:
I'm in a situation where using the debugger influences the code path (schrodinger style). Could it be that there are some timeouts on what goes through Ganache?

fleupold:
> I'm in a situation where using the debugger influences the code path (schrodinger style). Could it be that there are some timeouts on what goes through Ganache?

It's possible but hard to say without more context on what the different behavior is and where you set your breakpoints in order to experience those.

koeppelmann:
I am in favor of this PR but we should all know that this will allow an attacker to drain all the fees/buffer from the contract.
The attack is simple:
create a fake token where you control 100% of the supply.
Create a Uniswap pool - fake/ ETH

Trade fake->ETH (or any other token the settlement contract has) on Cowswap - With this PR we will accept the fake token and give the user the valuable tokens - after that, the attacker can remove liquidity from Uni.

The way to mitigate this is to use a whitelist of tokens - only if all tokens that we effectively buy (==all the users in the batch sell) are on the whitelist this shortpath is allowed.


By the way - as long as we take fees in any token the same trick can be used to effectively circumvent our fee (pay the fee in a fake token) and still buy any token.

marcovc:
>  use a whitelist of tokens

For sure. Additionally, only if the token also belongs to the list of the 10-15 most traded tokens:

image

to avoid having our buffers exposed to high price fluctuations of tokens we never heard about.

PR #551, opened 2021-05-04
Return unsupported token error fee route

fleupold:
Following #550 this PR now makes it so we return this error for unsupported tokens

### Test Plan
Added unit test + starting orderbook locally with `--unsupported-tokens 0xfad45e47083e4607302aa43c65fb3106f1cd7607` and on 

```
http://localhost:8080/api/v1/fee?sellToken=0xfad45e47083e4607302aa43c65fb3106f1cd7607&buyToken=0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d&amount=1000000&kind=buy
```

Getting

> {"errorType":"UnsupportedToken","description":"Token address 0xfad45e47083e4607302aa43c65fb3106f1cd7607"}

### DIFF
--- a/orderbook/src/api/get_fee_info.rs
+++ b/orderbook/src/api/get_fee_info.rs
@@ -50,7 +50,7 @@ pub fn get_fee_info_response(
             StatusCode::NOT_FOUND,
         )),
         Err(MinFeeCalculationError::UnsupportedToken(token)) => Ok(reply::with_status(
-            super::error("UnsupportedToken", format!("Token address {}", token)),
+            super::error("UnsupportedToken", format!("Token address {:?}", token)),
             StatusCode::BAD_REQUEST,
         )),
         Err(MinFeeCalculationError::Other(err)) => {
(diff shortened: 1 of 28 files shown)


orderbook/src/fee.rs
```diff
@@ -97,6 +100,19 @@ impl MinFeeCalculator {
...
+        }
+
+        if buy_token
+            .map(|t| self.unsupported_tokens.contains(&t))
+            .unwrap_or_default()
+        {
+            return Err(MinFeeCalculationError::UnsupportedToken(
+                buy_token.expect("Must exist"),
+            ));
+        }
```
vkgnosis, line 114:
```suggestion
        if let Some(token) = buy_token {
            if self.unsupported_tokens.contains(&token) {
                return Err(MinFeeCalculationError::UnsupportedToken(token));
            }
        }
```

vkgnosis (approved):

orderbook/src/fee.rs
fleupold: Right, I felt the upper was easier to merge with #542

orderbook/src/api/get_fee_info.rs
```diff
@@ -50,7 +50,7 @@ pub fn get_fee_info_response(
             StatusCode::NOT_FOUND,
         )),
         Err(MinFeeCalculationError::UnsupportedToken(token)) => Ok(reply::with_status(
-            super::error("UnsupportedToken", format!("Token address {}", token)),
+            super::error("UnsupportedToken", format!("Token address {:?}", token)),
```
bh2smith, line 53: Whoops! My mistake. Does this mean the token address wasn't being displayed before?

bh2smith (approved):

orderbook/src/api/get_fee_info.rs
fleupold: Actually this was my mistake in #550

PR #552, opened 2021-05-04
Extend api description of the order post data from field

vkgnosis:
With the current openapi version (3.0) this does not yet show up in the
generated documentation because you cannot overwrite fields from the
$ref but in the next version (3.1) it should work.


bh2smith (approved):
Love this!

Issue #545, opened 2021-05-03
Follow up tasks for better settlement simulation error logging


vkgnosis:
From #539

* [x] Have a metric for settlements that succeed at earlier block but fail at current block.
* [ ] Use a fixed block for the liquidity fetcher to ensure pool blockchain information comes from the same block.

fleupold:
Here is an example where a settlement simulation failed right before submission (12426718) AND for block 12426716 which was current when liquidity was fetched, but would have succeeded at an intermediate block (12426717) which is likely the block at which the liquidity was fetched.

Using a fixed block for the liquidity fetcher would have avoided this alert.

https://gnosisinc.slack.com/archives/CUD4MUCUF/p1620918277001000

PR #554, opened 2021-05-04
Add metric for settlement simulation succcess and failures

vkgnosis:
We are not emitting alerts anymore when settlements suceed on their
original block but fail on the current block before creating the
transaction. This metric allows us to keep track of how often this
occurs.

Part of #545

### Test Plan
new unit test for metrics

### DIFF
--- a/solver/src/metrics.rs
+++ b/solver/src/metrics.rs
@@ -11,6 +11,8 @@ pub trait SolverMetrics {
     fn liquidity_fetched(&self, liquidity: &[Liquidity]);
     fn settlement_computed(&self, solver_type: &str, start: Instant);
     fn order_settled(&self, order: &Order);
+    fn settlement_simulations_succeeded(&self, count: usize);
+    fn settlement_simulations_failed(&self, count: usize);
 }
 
 // TODO add labeled interaction counter once we support more than one interaction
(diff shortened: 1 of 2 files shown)


solver/src/metrics.rs
```diff
@@ -104,4 +129,21 @@ impl SolverMetrics for NoopMetrics {
     fn liquidity_fetched(&self, _liquidity: &[Liquidity]) {}
     fn settlement_computed(&self, _solver_type: &str, _start: Instant) {}
     fn order_settled(&self, _order: &Order) {}
+    fn settlement_simulation_successes(&self, _count: usize) {}
```
fleupold, line 132: naming nit: to be inline with the other methods maybe this can be called settlement_simulation_succeeded/failed

fleupold (approved):
Is there value in differentiating between the ones that failed on both blocks and the ones that only failed on Latest?

vkgnosis:
> Is there value in differentiating between the ones that failed on both blocks and the ones that only failed on Latest?

I don't think so. If they fail on the initial block we already error log which is an alert so it doesn't need a special place in the metrics.

solver/src/metrics.rs
vkgnosis: changed

PR #556, opened 2021-05-04
Separate reusable components in e2e test

fedgiac:
This PR refactors the current end to end test to make some of the existing components more reusable.

CC @marcovc.

### Test Plan
Current e2e test still works.


vkgnosis (approved):
Good start and I think we can continue to refactor things here as we see fit when we add another e2e test.

fleupold (approved):
Very nice!

PR #557, opened 2021-05-04
Adds costs to the objective function.

marcovc:
* It refactors code, to make sure rating settlements is after we know which settlements simulate ok.
* Estimates gas for each settlement, and uses that estimation for a more accurate objective value.

I've left some comments in the code that need answering.

Not sure what I could do to test it?

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -205,25 +205,27 @@ impl Driver {
         settlements: Vec<Settlement>,
         prices: &HashMap<H160, BigRational>,
     ) -> Vec<RatedSettlement> {
-        use futures::stream::{StreamExt};
-        futures::stream::iter(settlements).filter_map(|settlement| async {
-            let surplus = settlement.total_surplus(prices);
-            if let Ok(gas_estimate) = settlement_submission::estimate_gas(
-                &self.settlement_contract,
-                &settlement.clone().into(),
-            ).await {
-                Some(
-                    RatedSettlement {
+        use futures::stream::StreamExt;
+        futures::stream::iter(settlements)
+            .filter_map(|settlement| async {
+                let surplus = settlement.total_surplus(prices);
+                if let Ok(gas_estimate) = settlement_submission::estimate_gas(
+                    &self.settlement_contract,
+                    &settlement.clone().into(),
+                )
+                .await
+                {
+                    Some(RatedSettlement {
                         settlement,
                         surplus,
                         gas_estimate,
-                    }
-                )
-            }
-            else {
-                None
-            }
-        }).collect::<Vec<_>>().await
+                    })
+                } else {
+                    None
+                }
+            })
+            .collect::<Vec<_>>()
+            .await
     }
 
     pub async fn single_run(&mut self) -> Result<()> {
@@ -286,8 +288,7 @@ impl Driver {
             .settlement_simulations_succeeded(settlements.len());
         self.metrics.settlement_simulations_failed(errors.len());
 
-        let rated_settlements = self
-            .rate_settlements(settlements, &estimated_prices).await;
+        let rated_settlements = self.rate_settlements(settlements, &estimated_prices).await;
 
         if let Some(settlement) = rated_settlements.into_iter().max_by(|a, b| {
             a.objective_value(gas_price)
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -16,7 +16,7 @@ use gas_estimation::GasPriceEstimating;
 use itertools::{Either, Itertools};
 use num::BigRational;
 use primitive_types::H160;
-use shared::{price_estimate::PriceEstimating, Web3};
+use shared::{conversions::U256Ext, price_estimate::PriceEstimating, Web3};
 use std::{
     collections::{HashMap, HashSet},
     sync::Arc,
@@ -129,12 +129,12 @@ impl Driver {
     // Split settlements into successfully simulating ones and errors.
     async fn simulate_settlements(
         &self,
-        settlements: Vec<RatedSettlement>,
-    ) -> Result<(Vec<RatedSettlement>, Vec<(RatedSettlement, Error)>)> {
+        settlements: Vec<Settlement>,
+    ) -> Result<(Vec<Settlement>, Vec<(Settlement, Error)>)> {
         let simulations = settlement_simulation::simulate_settlements(
             settlements
                 .iter()
-                .map(|settlement| settlement.settlement.clone().into()),
+                .map(|settlement| settlement.clone().into()),
             &self.settlement_contract,
             &self.web3,
             &self.network_id,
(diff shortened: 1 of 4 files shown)


marcovc:
Regarding naming, since a settlement is not actually rated until we pass the `gas_price` in the call to `objective_value`, I suggest instead either pass the gas_price to RatedSettlement, or change the name RatedSettlement->SimulatedSettlement. Let me know if you like this idea, and which one.

solver/src/driver/solver_settlements.rs
```diff
@@ -27,14 +27,25 @@ pub fn filter_bad_settlements(
...
+    pub settlements: Vec<Settlement>,
 }
 
 // Each individual settlement has an objective value.
 #[derive(Debug)]
 pub struct RatedSettlement {
     pub settlement: Settlement,
-    pub objective_value: BigRational,
+    pub surplus: BigRational,
+    pub gas_estimate: BigRational,
```
fleupold, line 38: nano-nit: I'd find it slightly more natural to be the integer type and only convert it to Rational when needed.

solver/src/driver/solver_settlements.rs
```diff
@@ -27,14 +27,25 @@ pub fn filter_bad_settlements(
...
+}
+
+impl RatedSettlement {
+    pub fn objective_value(&self, gas_price: f64) -> BigRational {
+        // I can't see why a f64 couldn't be represented as a BigRational. Thoughts?
```
fleupold, line 43: IEEE 754 floating point numbers can represent positive or negative infinity, and NaN. Those cannot be represented as BigRationals

solver/src/driver/solver_settlements.rs
```diff
@@ -44,28 +55,16 @@ pub fn merge_settlements(
...
-        let objective_value = settlement.objective_value(prices);
-        settlements.push(RatedSettlement {
-            settlement,
-            objective_value,
-        });
+    let mut settlements = settlements;
+
+    settlements.sort_by_cached_key(|a| -a.total_surplus(prices));
+
+    // I'd like a comment here: what is this doing? If there were N settlements, after
```
fleupold, line 62: It should take the top `max_merged_settlements` (in terms of surplus) and try to merge them into an additional solution which gets appended to the list of potential settlement (resulting in n+1 settlements). This is a short term way of increasing throughput for multiple disconnected settlements until we have parallel tx execution.

Maybe the sorting logic could go inside the `merge_at_most_settlements` function (to become part of the unit testing logic) even though it means `max_by_key` will be marginally less efficient as it will operate on unsorted data.

solver/src/settlement_submission.rs
```diff
@@ -41,9 +51,7 @@ pub async fn submit(
         .context("failed to get pending gas price")?;
 
     // Account for some buffer in the gas limit in case racing state changes result in slightly more heavy computation at execution time
-    let gas_limit = retry::settle_method_builder(contract, settlement.clone())
-        .tx
-        .estimate_gas()
+    let gas_limit = estimate_gas(contract, &settlement)
```
fleupold, line 54: nanonit: if submit took a RatedSettlement it could reuse the gas estimate from that struct instead of estimating it again.

fleupold (approved):
Great change. I really like the separation of merging and rating settlements and am surprised who small the required change for this feature is.

solver/src/driver.rs
```diff
@@ -195,10 +195,33 @@ impl Driver {
...
+        prices: &HashMap<H160, BigRational>,
+    ) -> Result<Vec<RatedSettlement>> {
+        let mut result: Vec<RatedSettlement> = Vec::new();
+        for settlement in settlements {
+            let surplus = settlement.total_surplus(prices);
+            let gas_estimate = settlement_submission::estimate_gas(
+                &self.settlement_contract,
+                &settlement.clone().into(),
+            )
+            .await?;
```
vkgnosis, line 215: It would be nice to do this as a batched call like with the simulations. Doesn't have to be in this PR. Also not sure if ethcontract's estimateGas works with batch transport.

vkgnosis, line 215: The way this is written if only one of the gas estimates fails we fail rating all of them. While I think it's unlikely that a single one would fail maybe we should be more robust and error log while continuing with the rest.
Doing this is related to the batch transport suggestion too.

solver/src/driver/solver_settlements.rs
```diff
@@ -44,28 +55,16 @@ pub fn merge_settlements(
...
-        settlements
-            .iter()
-            .map(|settlement| settlement.settlement.clone()),
-    ) {
-        let objective_value = settlement.objective_value(prices);
-        settlements.push(RatedSettlement {
-            settlement,
-            objective_value,
-        });
+    let mut settlements = settlements;
```
vkgnosis, line 58: You can change it to `mut settlements: Vec<Settlement>,` in the function arguments and remove this line.

vkgnosis (approved):

solver/src/driver/solver_settlements.rs
marcovc: Right of course. I've used unwrap anyway, since we're not expecting any of those values.

solver/src/driver/solver_settlements.rs
marcovc: I'm still unsure about it because I lack the understanding of what merged settlements are and why we're doing (is this written down in some issue?), so I won't try to change it in this PR.

solver/src/driver.rs
marcovc: I think I know what you mean and agree but I'll leave that optimization to another PR to avoid delaying this one too much.

marcovc: If one of the gas estimates fail then the corresponding settlement will have "undefined" objective value. What to do about the "max" part then? Should we ignore the ones with undefined gas, or abort all together?

fleupold: I'd vote for skipping the ones for which gas cannot be estimated. This could happen e.g. if a new block was mined and now the simulation no longer works due to a race condition. In this case we'd also fail to estimate gas.

marcovc: Done. Let me know if it's ok now, as I used some new tricks I just learned.

solver/src/driver.rs
```diff
@@ -195,10 +195,39 @@ impl Driver {
...
+    ) -> Vec<RatedSettlement> {
+        use futures::stream::StreamExt;
+        futures::stream::iter(settlements)
+            .filter_map(|settlement| async {
+                let surplus = settlement.total_surplus(prices);
+                if let Ok(gas_estimate) = settlement_submission::estimate_gas(
+                    &self.settlement_contract,
+                    &settlement.clone().into(),
+                )
+                .await
```
fleupold, line 216: nit: Instead of if/else you might be able to get away with ?

```rust
let gas_estimate = ettlement_submission::estimate_gas(
                    &self.settlement_contract,
                    &settlement.clone().into(),
                )
                .await?
Ok(RatedSettlement {
                        settlement,
                        surplus,
                        gas_estimate,
                    })

```

fleupold (approved):
LGTM

PR #558, opened 2021-05-05
Gas aware price estimations in API

fleupold:
This PR makes it so that the price estimator is not returning marginally better prices when the baseline execution for this path is more expensive than the price improvement.

This is achieved by changing the "maximization criterion" in the price estimator from just looking at the value to look at the value - tx_cost (for sell orders) and at the (cost + tx_cost) for buy orders.

This change has no effect on the baseline solver in the driver yet, as it could prevent it from finding solutions that the frontend has suggested to the user. Once the frontend uses our price estimation API we can do the same optimisation to the driver's baseline solver.

### Test Plan
Added unit tests.


vkgnosis (approved):
Only thing I am worried about is making requests to GasNow too often. I started working on implementing a cache + rate limits in the GasNow estimator for this reason.

shared/src/pool_fetching.rs
```diff
@@ -183,6 +184,10 @@ impl BaselineSolvable for Pool {
             price
         })
     }
+
+    fn gas_cost(&self) -> usize {
+        POOL_SWAP_GAS_COST
```
bh2smith, line 189: I assume this is a fixed cost for Uniswap-Like-Pools. Is this value substantially different for 1inch and will this need to be adjusted for Balancer?

shared/src/price_estimate.rs
```diff
@@ -170,13 +183,24 @@ impl UniswapPriceEstimator {
...
         self.best_execution(
             sell_token,
             buy_token,
             sell_amount,
             |amount, path, pools| {
-                estimate_buy_amount(amount, path, pools).map(|estimate| estimate.value)
+                estimate_buy_amount(amount, path, pools).map(|estimate| {
+                    let proceeds_in_native_token =
+                        estimate.value.to_big_rational() * buy_token_price_in_native_token.clone();
+                    let tx_cost_in_native_token = U256::from_f64_lossy(gas_price).to_big_rational()
```
bh2smith, line 200: ubernit: Given that native assets aren't actually tokens, I might propose names more like `native_tx_cost`, `native_proceeds`

solver/src/baseline_solver.rs
```diff
@@ -43,6 +43,10 @@ impl BaselineSolvable for AmmOrder {
     fn get_spot_price(&self, base_token: H160, quote_token: H160) -> Option<BigRational> {
         amm_to_pool(self).get_spot_price(base_token, quote_token)
     }
+
+    fn gas_cost(&self) -> usize {
+        amm_to_pool(self).gas_cost()
```
bh2smith, line 48: interesting I hadn't see `amm_to_pool` before. Now that I see it and realize that `AmmOrder` and `Pool` share the same attributes, couldn't we do something cute that would make them both essentially the same... I mean we should be able to just say `amm_order.gas_cost()` by making gas cost a trait method and implementing it for Pool and AmmOrder

bh2smith (approved):
Formidable! Looks great. Just a little naming nit about the term "native token" which can be completely ignored (given that its hard to come up with a better name)

e2e/tests/onchain_settlement.rs
```diff
@@ -169,7 +169,7 @@ async fn onchain_settlement(web3: Web3) {
         liquidity_collector,
         price_estimator,
         vec![Box::new(solver)],
-        Box::new(web3.clone()),
+        Arc::new(web3.clone()),
```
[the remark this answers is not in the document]
fleupold: Because we want access to the "same" price estimator in multiple components. Box is an exclusive pointer to an object (not directly cloneable), whereas Arc is a shared pointer to the underlying object which can be cloned.

shared/src/pool_fetching.rs
fleupold: Yes, this likely has to be different for each BaselineSolvable (it's also only a rough estimate right now as the concrete swap costs depend on the concrete ERC20 token logic)

shared/src/price_estimate.rs
fleupold: I'd say ETH is still a token (not an ERC20 though), wouldn't you agree?

solver/src/baseline_solver.rs
fleupold: AmmOrder is of type `Liquidity` which at the moment is only exposed in the driver crate and includes logic on how to actually encode a swap into calldata. `BaselineSolvable` (or `Pool` in this example) is the type in the `shared` crate that is used for both the API as well as the driver's BaselineSolver.

Right now the API doesn't have to know about how a liquidity source is encoded as an `Interaction`. However, this may change in the future (if we e.g. want to simulate the Baseline execution to see if tokens can be supported) in which case the refactoring you suggest makes a lot of sense.

nlordell (approved):

PR #560, opened 2021-05-06
Expose solver name on Settlement struct

fleupold:
I'm not sure this approach is the best, but it would be nice to know in our logs which solvers are responsible for the settled trades (to get an idea which ones are performing best and potential later attribution).

Next up, I'd change the metric looging methods to add a solver label to each trade/simulation event.

### Test Plan
CI


vkgnosis:
> Because of "settlement merging" we cannot really do this attribution exact.

At the moment we only merge settlements by the same solver.

I'm not happy with storing the solver name inside of Settlement because that struct shouldn't know about solvers. It would be nicer if we could do this just in a driver wrapper struct liked RatedSettlement. But if it is really awkward there I am ok with doing it this way.

fleupold:
Implemented the suggestion with another wrapper struct. Let me know what you think.

vkgnosis (approved):

bh2smith (commented):
Seems reasonable. I can't think of a better way to do this at the moment.

PR #562, opened 2021-05-06
Update gp estimation dependency to use gas now with cache

vkgnosis:
This is more urgent than I thought because we're already querying the
estimator every time a fee is calculated if the fee isn't cached.

### Test Plan
CI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -898,10 +898,11 @@ dependencies = [
 [[package]]
 name = "gas-estimation"
 version = "0.1.0"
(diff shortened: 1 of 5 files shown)


fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2Fl41JGlWa1xOjJSsV2%2Fgiphy.gif&token=fast" data-gitmeme-token="fast"></a>

PR #564, opened 2021-05-06
Sends fee to the solver.

marcovc:
We'd like to reason with the fee in the solver to be able to obtain less non-trivial solutions as of now, and more profitable solutions in general long-term.

Unfortunately this required more changes that we'd like.

There's still the objective function in the driver: Should we update it now to take into account the fee as well?

Note: Merging this probably requires changing driver startup scripts.


solver/src/http_solver.rs
```diff
@@ -302,6 +308,17 @@ impl HttpSolver {
...
+
+    fn order_fee(&self, order: &LimitOrder) -> u128 {
+        assert!(
+            self.fee_discount_factor > 0.,
+            "Discount factor can't be zero."
+        );
+        let subsidized_fee = (order.fee_amount.to_f64_lossy() / self.fee_discount_factor)
+            .ceil()
+            .to_u128();
+        subsidized_fee.unwrap() // Should we panic or return zero if this happens?
```
fleupold, line 320: Maybe return an error in both failure cases (discount factor 0/float conversion)? The solve method already returns a Result

solver/src/http_solver.rs
```diff
@@ -302,6 +308,17 @@ impl HttpSolver {
     async fn uniswap_cost(&self, gas_price: f64) -> u128 {
         gas_price as u128 * GAS_PER_UNISWAP
     }
+
+    fn order_fee(&self, order: &LimitOrder) -> u128 {
+        assert!(
+            self.fee_discount_factor > 0.,
+            "Discount factor can't be zero."
+        );
+        let subsidized_fee = (order.fee_amount.to_f64_lossy() / self.fee_discount_factor)
```
fleupold, line 317: this is the unsubsidized fee isn't it?

fleupold (approved):
It's unfortunate that FEE_DISCOUNT_FACTOR needs to be set on the driver as well, but I don't see another option at the moment on how the MIP solver can get reasonable input regarding the economic viability of orders.

Maybe @vkgnosis has another idea...

marcovc:
> It's unfortunate that FEE_DISCOUNT_FACTOR needs to be set on the driver as well

If it's not confidential, perhaps it could be exposed in the orderbook http api, and the driver would get it from there?

vkgnosis:
You want to pass this number `order.fee_amount.to_f64_lossy() / self.fee_discount_factor` , right? How come just the fee_amount does not work? If all orders have their fee changed proportionally by the same factor does that matter?
I don't mind setting the fee factor as part of the config for the solver.

fleupold:
> If all orders have their fee changed proportionally by the same factor does that matter?

The issue is that the solver has a notion "economic viability" and will only settle orders if `fee + surplus` is greater than the cost to settle them. Since the fees are artificially low (due to the discount) most orders are actually not economically viable in the strict sense. However we still want the solver to match them as if there was no fee discount (otherwise the user experience might be bad). So we need to show the orders with their "real"/"unsubsidized" fee (alternatively we could also divide all the costs by the subsidy, but that seems a bit weird).

vkgnosis (approved):
That makes sense. I think any solution needs the fee discount factor and like I said I don't think it's unreasonable for the driver to know it. When merging, could you put `Config Change` or something like that in the commit title so we don't forget to set this for the next prod release?

shared/src/arguments.rs
```diff
@@ -51,6 +51,10 @@ pub struct Arguments {
     #[structopt(long, env = "UNSUPPORTED_TOKENS", use_delimiter = true)]
     pub unsupported_tokens: Vec<H160>,
 
+    /// Fee discount factor: 1 means no discount, 0.9 means 10% discount.
```
bh2smith, line 54: I might elaborate here and say something about how this is a value in the range `[0, 1]`. Do we have any assertions that this value is `<= 1` anywhere?

solver/src/http_solver.rs
```diff
@@ -424,6 +437,7 @@ mod tests {
             mock_token_info_fetcher,
             mock_price_estimation,
             "mock_network_id".to_string(),
+            1.,
```
bh2smith, line 440: Would this also be a place to use the default instead of a hard coded 1?

bh2smith (approved):
Looks good!

shared/src/arguments.rs
marcovc: I actually don't really love this name, but since it was already around, I followed along. 

My understanding is that `fee_discount_factor` is a number, that multiplied by the fee we would charge for that order if not subsidizing anything, gives the effective fee that is charged to the user.

```
subsidized_fee = unsubsidized_fee * fee_discount_factor
```

That means the effective fee payed by the user goes down as fee_discount goes down, which is why I think calling it "discount" is misleading (we usually pay less when the discount is larger, and not the other way around).

If I got the above right, as far as I can tell, the domain for `subsidized_fee` is therefore `[0, +oo[` . As you point out, values greater than 1 don't make much sense, and it has the extra shortcoming that when we can't solve the equation for `unsubsidized_fee` when `fee_discount_factor=0`.

To solve all those problems, we could instead model it like this:

```
subsidized_fee = unsubsidized_fee * (1 - fee_discount_factor)
```

and require that `fee_discount_factor` is in `[0,1]` .

solver/src/http_solver.rs
marcovc: You mean that `Default::default()` thing? I assumed that that is just a trait that can get specialized for every type, and for f64 it is zero. And currently zero is the only value forbidden for `fee_discount_factor`  (see previous comment).

shared/src/arguments.rs
vkgnosis: Like you said Marco, it was already this way so no need to fix it in this PR. I do think what you say makes sense so I'm open to that change.

marcovc:
Thanks for all the reviews. I'll do this part in another PR:

> There's still the objective function in the driver: Should we update it now to take into account the fee as well?

PR #566, opened 2021-05-07
Add metric for number of rows in database tables

vkgnosis:
I don't have something specific in mind but this seems useful in general to have metrics for.

### Test Plan
new unit test


orderbook/src/database.rs
```diff
@@ -57,4 +57,45 @@ impl Database {
             .await?;
         Ok(())
     }
+
+    async fn count_rows_in_table(&self, table: &str) -> Result<i64> {
+        let query = format!("SELECT COUNT(*) FROM {};", table);
```
vkgnosis, line 62: Apparently you can't use `$1` in the from position.

orderbook/src/database.rs
```diff
@@ -57,4 +57,45 @@ impl Database {
...
+    }
+
+    pub async fn count_rows_in_tables(&self) -> Result<HashMap<&'static str, i64>> {
+        let tables = [
+            "orders",
+            "trades",
+            "invalidations",
+            "min_fee_measurements",
+            "settlements",
+        ];
```
fleupold, line 74: Can we make Tables an enumerable struct similar to `Liquidity` (EnumVariantNames) so that we don't store our list of "all tables" in multiple places? The truncate queries could use the same list

orderbook/src/main.rs
```diff
@@ -72,6 +74,20 @@ pub async fn orderbook_maintenance(
     unreachable!()
 }
 
+pub async fn database_metrics(metrics: Arc<Metrics>, database: Database) -> ! {
```
fleupold, line 77: Is there a specific reason you don't want to do this as part of orderbook maintenance? I think the less different "tasks" we spawn the easier it will be to reason about what is happening in the system (e.g. race conditions won't happen in a sequential setup). Of course, if there is a good reason in priority or frequency it might make sense but I personally don't see this here.

e2e/tests/services.rs
```diff
@@ -161,12 +162,16 @@ impl OrderbookServices {
             Duration::from_secs(120),
         ));
 
+        let registry = Registry::default();
+        let metrics = Arc::new(Metrics::new(&registry).unwrap());
```
fleupold, line 166: We have NoopMetrics for the e2e tests.

fleupold (approved):

e2e/Cargo.toml
```diff
@@ -12,6 +12,7 @@ hex-literal = "0.3"
 lazy_static = "1.4"
 model = { path = "../model" }
 orderbook = { path = "../orderbook" }
+prometheus = "0.12"
```
bh2smith, line 15: Does the fact that we are adding the metrics in the e2e test mean we only measure it when tests run?

e2e/tests/services.rs
```diff
@@ -161,12 +162,16 @@ impl OrderbookServices {
...
         orderbook::serve_task(
             db.clone(),
             orderbook.clone(),
             fee_calculator,
             price_estimator.clone(),
             API_HOST[7..].parse().expect("Couldn't parse API address"),
+            registry,
+            metrics,
```
bh2smith, line 174: Oh I see that we had to include it in the e2e test because we want to serve the orderbook which requires these metrics. Should we consider making these optional and not send them along with e2e testing?

orderbook/src/database.rs
bh2smith: I am glad this isn't a public method because I suppose if `table = "orders; DROP TABLE orders;"` we could get in trouble here. 

In general, I am under the impression that we should try and avoid making database queries with variable string inputs. Do we have an enum somewhere with table names or could we generate one with a DB query:

```sh
postgres=# \dt
```

bh2smith (approved):
Looks good, but I think we should address the comments regarding tables as an enum and not passing string into DB queries. Also, would like to know if the metrics parameters introduced into serve orderbook could be made optional so we don't need them for e2e testing.

orderbook/src/database.rs
vkgnosis: You should definitely never create sql queries from strings with attacker controlled inputs. In this case we control the possible values for `table` so it's fine.

orderbook/src/database.rs
vkgnosis: I made it a const list.

e2e/tests/services.rs
vkgnosis: True but this also serves as another test that there is no panic in the real metrics code path. Not sure which is better.

e2e/Cargo.toml
vkgnosis: No. Just need to import this crate to create an instance of Registry.

e2e/tests/services.rs
vkgnosis: See Felix' comment about NoopMetrics  above.

PR #569, opened 2021-05-07
Make fee calculator eth aware

fedgiac:
Supersedes and closes #528.

The PR adds eth support to the price estimator.
I noticed that a recently introduced function `is_valid_fee` is also called with `ETH_BUY_ADDRESS` in the e2e test supporting eth. This PR creates a wrapper around a `MinFeeCalculator` that replaces the eth address in `buy_token` by the weth address. If new functions were to be added to `MinFeeCalculator`, this would be easily detected and fixed.
The new trait helps testing this wrapper.

### Test Plan

New unit tests. Also, as in #528:
```
$ sell_dai_for() {
   curl -X 'GET' \
     'http://localhost:8080/api/v1/fee?sellToken=0x5592ec0cfb4dbc12d3ab100b257153436a1f0fea&buyToken='"$1"'&amount=12&kind=buy' \
     -H 'accept: application/json'
   echo
 }
$ sell_dai_for 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE # eth
{"expirationDate":"2021-05-07T19:35:40.857485792Z","amount":"164621342166646656"}
$ sell_dai_for 0xc778417E063141139Fce010982780140Aa0cD5Ab # weth
{"expirationDate":"2021-05-07T19:35:50.551925497Z","amount":"164621342166646656"}
$ sell_dai_for 0x5592ec0cfb4dbc12d3ab100b257153436a1f0fea # dai
{"expirationDate":"2021-05-07T19:36:05.183569640Z","amount":"0"}
```

### DIFF
--- a/orderbook/src/fee.rs
+++ b/orderbook/src/fee.rs
@@ -33,15 +33,6 @@ pub struct MinFeeCalculator {
 #[cfg_attr(test, mockall::automock)]
 #[async_trait::async_trait]
 pub trait MinFeeCalculating {
-    fn new(
-        price_estimator: Arc<dyn PriceEstimating>,
-        gas_estimator: Box<dyn GasPriceEstimating>,
-        native_token: H160,
-        database: Database,
-        discount_factor: f64,
-        unsupported_tokens: HashSet<H160>,
-    ) -> Self;
-
     // Returns the minimum amount of fee required to accept an order selling the specified order
     // and an expiry date for the estimate.
     // Returns an error if there is some estimation error and Ok(None) if no information about the given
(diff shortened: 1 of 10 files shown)


bh2smith:
Is this somehow related or dependent on #528 because I am seeing some similarities with the code. If so. I might suggest comparing this branch with that so I can see the relevant difference better. 

Just want to point out that on xDAI this is all about wxDAI - so maybe we want to consider using "Native Asset" in place of ETH everywhere.

How is this PR different from #528?

orderbook/src/fee.rs
```diff
@@ -23,6 +30,41 @@ pub struct MinFeeCalculator {
     unsupported_tokens: HashSet<H160>,
 }
 
+#[cfg_attr(test, mockall::automock)]
+#[async_trait::async_trait]
+pub trait MinFeeCalculating {
+    fn new(
```
fleupold, line 36: Does new have to be part of the interface? Usually the instantiation dependencies should be an implementation detail of each implementor and not dictated by the trait.

fleupold (approved):
Looks good to me and this PR is actually required in order to make current staging work (cf. https://gnosisinc.slack.com/archives/CC3C5SMB5/p1620504481465800).

The only thing I don't like is that `new` is part of the trait, I think we should make it specific to each implementation. I really like the layered approach in general though (looks like EthAdapter can also be used to implement `PriceEstimating` and the like)!

vkgnosis (approved):

fleupold (approved):
❇️

PR #570, opened 2021-05-07
Add e2e test trading eth

fedgiac:
End to end test with eth being traded.
This new test shows that both buy and sell order with eth as the buy token are working. It also tests that the fee endpoint for the eth flag address is working.

### Test Plan
New code is a test.


e2e/tests/eth_integration.rs
```diff
@@ -0,0 +1,273 @@
...
+    create_orderbook_api, deploy_mintable_token, to_wei, GPv2, OrderbookServices, UniswapContracts,
+    API_HOST,
+};
+
+const TRADER_BUY_ETH_A_PK: [u8; 32] =
+    hex!("0000000000000000000000000000000000000000000000000000000000000001");
+const TRADER_BUY_ETH_B_PK: [u8; 32] =
+    hex!("0000000000000000000000000000000000000000000000000000000000000002");
+const TRADER_SELL_ETH_PK: [u8; 32] =
+    hex!("0000000000000000000000000000000000000000000000000000000000000003");
```
bh2smith, line 31: Could use fewer characters if we just use `[1u8; 32], [2u8; 32], [3u8; 32]` although they are not the exact same strings in the end.

bh2smith (approved):
Cool stuff!

e2e/tests/eth_integration.rs
```diff
@@ -0,0 +1,273 @@
...
+            token.address(),
+            weth.address(),
+            to_wei(100_000),
+            to_wei(100_000),
+            0_u64.into(),
+            0_u64.into(),
+            solver.address(),
+            U256::max_value(),
+        )
+    );
```
fleupold, line 107: Maybe we can share these lines in a util method across tests (could also include the required minting).

e2e/tests/eth_integration.rs
```diff
@@ -0,0 +1,273 @@
...
+                to_wei(42)
+            ))
+            .send()
+            .await
+            .unwrap()
+    };
+    let fee_with_weth = estimate_fee(token.address(), BUY_ETH_ADDRESS).await;
+    assert_eq!(fee_with_weth.status(), 200);
+    // Eth is only supported as the buy token
+    let fee_with_eth = estimate_fee(BUY_ETH_ADDRESS, token.address()).await;
```
fleupold, line 141: I think these variable names are off, aren't they?

e2e/tests/eth_integration.rs
```diff
@@ -0,0 +1,273 @@
...
+        )
+        .build()
+        .order_creation;
+    let placement = client
+        .post(&format!("{}{}", API_HOST, ORDER_PLACEMENT_ENDPOINT))
+        .body(json!(order_buy_eth_a).to_string())
+        .send()
+        .await;
+    assert_eq!(placement.unwrap().status(), 201);
+    let order_buy_eth_b = OrderBuilder::default()
```
fleupold, line 166: why are there two buy eth orders?

e2e/tests/eth_integration.rs
```diff
@@ -0,0 +1,273 @@
...
+        .build()
+        .order_creation;
+    let placement = client
+        .post(&format!("{}{}", API_HOST, ORDER_PLACEMENT_ENDPOINT))
+        .body(json!(order_buy_eth_b).to_string())
+        .send()
+        .await;
+    assert_eq!(placement.unwrap().status(), 201);
+
+    let order_sell_eth = OrderBuilder::default()
```
fleupold, line 187: nit: What extra bit is this testing? Would it suffice to only buy eth with a single order and rely on the AMM to provide the matching? This might make the test significantly shorter?

e2e/tests/services.rs
```diff
@@ -24,27 +24,34 @@ use std::{collections::HashSet, str::FromStr, sync::Arc, time::Duration};
...
+}
 
 pub fn to_wei(base: u32) -> U256 {
     U256::from(base) * U256::from(10).pow(18.into())
 }
 
-pub fn create_orderbook_api(web3: &Web3) -> OrderBookApi {
-    let native_token = WETH9::at(web3, H160([0x42; 20]));
+pub fn create_orderbook_api(web3: &Web3, weth_address: H160) -> OrderBookApi {
+    let weth = WETH9::at(&web3, weth_address);
```
fleupold, line 50: Alternatively, we could deploy WETH as part of our `deploy.rs` script and have it have a shared address across tests (and getting it via `.deployed` instead of `.at`? But no strong feelings.

fleupold (approved):
Great test that already caught a bad issue if we had released to prod without #569!

vkgnosis (approved):

e2e/tests/eth_integration.rs
fedgiac: Good find!
At first I wanted to check that querying the two gave the same result, but then I realized that it wouldn't be helpful because the gas price is zero. So now I just compare a buy eth pair to an invalid pair.

e2e/tests/eth_integration.rs
fedgiac: One is a buy order, the other is a sell order. They use different logic in the services, so I wanted to make sure that both paths are touched.

fedgiac:
Merging now to have the test available, changes to how WETH is handled can be done in another PR.

fedgiac:
Had to merge manually because I changed `.github/workflows`.

PR #573, opened 2021-05-10
[Easy] Doument Solver Trait

### LINKED ISSUE #192: Document solver trait
docs for this public trait (in particular the semantics of the return type) would be nice.

_Originally posted by @fleupold in https://github.com/gnosis/gp-v2-services/pull/174#discussion_r554963825_

bh2smith:
This was already essentially done (cf #560) in the recent refactoring of the solver trait to include `name`. I have merely included a bit of description on top of the trait which should suffice to close #192

### Test Plan
Existing CI.


solver/src/solver.rs
```diff
@@ -15,6 +15,9 @@ use reqwest::Url;
 use shared::{price_estimate::PriceEstimating, token_info::TokenInfoFetching};
 use structopt::clap::arg_enum;
 
+/// Defines shared behaviour for all Solvers.
+/// A `solve` method which transforms a collection of `Liquidity` sources into a `Settlement`
```
fleupold, line 19: into a list of Settlement ?

Generally I don't find this comment much helpful as it's just stating the obvious that can be seen from the method definition (and this sentence will likely forgotten to be updated if the method definition changes). Maybe we could explain a bit more what "Solvers" are how (e.g. that they can use liquidity on top of the provided ones if they have private sources, etc.)

solver/src/solver.rs
```diff
@@ -15,6 +15,9 @@ use reqwest::Url;
 use shared::{price_estimate::PriceEstimating, token_info::TokenInfoFetching};
 use structopt::clap::arg_enum;
 
+/// Defines shared behaviour for all Solvers.
```
fleupold, line 18: It's really the interface that all solvers need to implement (there is no behavior in this trait definition).

vkgnosis (approved):

Issue #475, opened 2021-04-15
Investigate even more erc20 validity checks using batch calls


vkgnosis:
When I made https://github.com/gnosis/gp-v2-services/pull/469 I thought I was only able to simulate (call) a single transaction which limits what can be checked. For example I thought I wouldn't be able to simulate the user first approving the token and then the allowance manager transfering.
Felix suggested that using batch calls this might be possible so I'm using this issue as a reminder to myself to investigate.

fleupold:
Right, this is not possible because the first transaction doesn't actually update any state (it's a read). It might work if it's an actual transaction that is being sent (but in that case the batch might not return until the tx is mined and this is not supported by ethcontracts).

One other option would be to use something like https://github.com/makerdao/multicall which might allow simulating a transaction followed by a static call (but I'm not sure if this fits our use case, as the msg.sender would be the Multicall contract).

nlordell:
To summarize what we talked about in the weekly sync:

The idea would be to add a method to simulate a transfer with a balance check to make sure the ERC20 behaves in the "standard" way (i.e. non-deflationary and does not take fees from the transferred amount). This method could be added to a storage accessible reader contract as to not be part of the actual settlement contract. Here is the existing reader contract: https://github.com/gnosis/gp-v2-contracts/blob/main/src/contracts/reader/SettlementStorageReader.sol

Basically, something like (in pseudo-code):
```solidity
function checkTokenTransfer(interactions, token, expectedTransfer) {
  address randomAddress = address(uint160(uint256(block.hash)));
  uint256 startingBalance = token.balanceOf(randomAddress);
  foreach (interaction of interactions) {
    // NOTE: allow interactions to the allowance manager
    interaction.execute()
  }
  token.transfer(randomAddress, expectedTransfer);
  uint256 finalBalance = token.balanceOf(randomAddress);
  require(finalBalance - startingBalance == expectedTransfer);
}
```

We use interactions so that we can have different ways of transferring the tokens:
- If the order is selling the token we want to test, then we transfer from the user to the settlement contract with the allowance manager
- If the order is a sell order, then we can use whatever is easiest (Uniswap pair for example) to get some of the token into the settlement contract and try transferring it.

vkgnosis:
Link from Nick: https://openethereum.github.io/JSONRPC-trace-module#trace_callmany

fleupold:
[...]

PR #582, opened 2021-05-11
Use trace_callMany api for more accurate bad token detection

vkgnosis:
This api allows us to simulate multiple transactions together. If it is
not available we fall back to the existing method.

Related to #475

### Test Plan
Adjusted the tests for this module which you can run with `cargo test -p orderbook account_balances::tests::mainnet -- --ignored`.

### DIFF
--- a/orderbook/src/account_balances.rs
+++ b/orderbook/src/account_balances.rs
@@ -25,12 +25,12 @@ pub trait BalanceFetching: Send + Sync {
     // Called periodically to perform potential updates on registered balances
     async fn update(&self);
 
-    // Check if the allowance manager would be able to call transfer_from with these parameters.
-    // This is useful for tokens that are not consistent about their internal checks and what they
-    // report as balance and allowance. By checking whether the actual transfer would suceed we can
-    // be more certain (but still not 100%) that the balance really is available to the settlement
-    // contract.
-    async fn can_transfer(&self, token: H160, from: H160, amount: U256) -> bool;
+    // Check that the settlement contract can make use of this user's token balance. This check
+    // could fail if the user does not have enough balance, has not given the allowance to the
+    // allowance manager or if the token does not allow freely transferring amounts around for
+    // for example if it is paused or takes a fee on transfer.
+    // If the node supports the trace_callMany we can perform more extensive tests.
+    async fn can_transfer(&self, token: H160, from: H160, amount: U256) -> Result<bool>;
 }
 
 pub struct Web3BalanceFetcher {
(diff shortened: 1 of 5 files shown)


orderbook/src/account_balances.rs
```diff
@@ -114,6 +114,62 @@ impl Web3BalanceFetcher {
         }
         Ok(())
     }
+
+    async fn can_transfer_call(&self, token: H160, from: H160, amount: U256) -> bool {
+        let instance = ERC20::at(&self.web3, token);
```
vkgnosis, line 119: This function hasn't changed. Just moved from the trait impl to here.

orderbook/src/account_balances.rs
```diff
@@ -114,6 +114,62 @@ impl Web3BalanceFetcher {
...
+
+        let calldata = instance.transfer(from, amount).tx.data.unwrap();
+        let transfer_out = CallRequest {
+            from: Some(recipient),
+            to: Some(token),
+            data: Some(calldata),
+            ..Default::default()
+        };
+
+        crate::trace_many::trace_many(vec![transfer_in, transfer_out], &self.web3).await
```
nlordell, line 171: Do we also need to check that the balance remains unchanged? This system can be tricked by sending in some deflationary token to the settlement contract just so this check passes.

I don't think it necessarily needs to be added here, I think this PR as-is is a great start! Just something to consider as a future improvement (if it is possible with `trace_callMany` to do get some output bytes from a call after the first two calls).

orderbook/src/trace_many.rs
```diff
@@ -0,0 +1,42 @@
...
+    Transport,
+};
+
+// Use the trace_callMany api https://openethereum.github.io/JSONRPC-trace-module#trace_callmany
+// api to simulate whether these call requests if applied together in order would succeed.
+// Err if communication with the node failed.
+// Ok(true) if transactions simulate without reverting
+// Ok(false) if transactions simulate with at least one revert.
+pub async fn trace_many(requests: Vec<CallRequest>, web3: &Web3) -> Result<bool> {
+    use web3::types::{BlockTrace, TraceType};
```
nlordell, line 14: why the scoping?

nlordell (approved):
Awesome!

Just some comments for some followup improvements that could be done. Looks great as-is though.

orderbook/src/account_balances.rs
nlordell: I think, as it is now, it will only catch deflationary tokens for which the settlement contract either has no balance or a very small amount of balance.

In general, I think the assumption is fine, but in the future we can further assert that the settlement contract's balance before and after the two transfers stays the same so that the check continues to catch deflationary tokens for which the settlement contract has some balance in.

The reason for suggesting this is:
- Deflationary token admin adds settlement contract to its exemption list
- Settlement contract trades some deflationary tokens and accrues fees in the token
- Deflationary token admin removes settlement contract from exemption list
- Some trades would be allowed because of the remaining buffers in the settlement contract
- Trades with these tokens drain the settlement contract buffer for these tokens and takes fees.

orderbook/src/account_balances.rs
```diff
@@ -152,24 +208,12 @@ impl BalanceFetching for Web3BalanceFetcher {
...
-            data: Some(calldata),
-            ..Default::default()
-        };
-        let block = Some(BlockId::Number(BlockNumber::Latest));
-        let response = self.web3.eth().call(call_request, block).await;
-        response
-            .map(|bytes| is_empty_or_truthy(bytes.0.as_slice()))
-            .unwrap_or(false)
+    async fn can_transfer(&self, token: H160, from: H160, amount: U256) -> Result<bool> {
+        if let Ok(result) = self.can_transfer_trace(token, from, amount).await {
```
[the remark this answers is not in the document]
nlordell: I like `Result` here.

orderbook/src/account_balances.rs
```diff
@@ -216,14 +265,19 @@ mod tests {
         let settlement = contracts::GPv2Settlement::deployed(&web3).await.unwrap();
         let allowance = settlement.allowance_manager().call().await.unwrap();
         let fetcher = Web3BalanceFetcher::new(web3, allowance, settlement.address());
-        let owner = H160(hex!("c978f4364c03a00352e8c7d9619b42e26b6424ab"));
-        let token = H160(hex!("c12d1c73ee7dc3615ba4e37e4abfdbddfa38907e"));
+        let owner = H160(hex!("78045485dc4ad96f60937dad4b01b118958761ae"));
```
fleupold, line 268: Not sure where this address is coming from, but it looks like the token itself has a balance of its own token. This is likely not going to move (whereas this might transfer out its balance at some point)?

fleupold (approved):
Agree that this is a great start! It doesn't however yet catch failures when people try to buy one of those funny tokens yet, does it?

I wonder if there is an easy way to achieve this (maybe we could find a holder of the buy token by looking at recent Transfer events and simulate the same thing) or if it would make sense to simulate the entire baseline matching (maybe on a slightly altered settlement contract that doesn't enforce limit prices in case the baseline price is not satisfying the limit at the moment). This would also give us a very accurate gas estimate.

orderbook/src/trace_many.rs
```diff
@@ -0,0 +1,41 @@
...
+    let response = transport.execute("trace_callMany", params).await?;
+    let traces: Vec<BlockTrace> = serde_json::from_value(response)?;
+    for trace in traces {
+        let transaction_trace = trace.trace.ok_or_else(|| anyhow!("trace not set"))?;
+        let first = transaction_trace
+            .first()
+            .ok_or_else(|| anyhow!("expected at least one trace"))?;
+        if first.error.is_some() {
+            return Ok(false);
+        }
```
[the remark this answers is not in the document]
vkgnosis: This is not clearly documented in the api but the way I understand it is that the first trace is for the whole transaction and the following for sub transactions. We only care about the whole.

vkgnosis:
> maybe we could find a holder of the buy token

This was the bigger complexity that I wasn't sure how to address yet. We need to somehow get some of the token.

vkgnosis:
Will merge this now to at least catch more bad sell tokens. For future I'm planning on having a separate component that checks if a token is reasonable the first time it is encountered (be it for new order, fee request, etc).

orderbook/src/account_balances.rs
nlordell: Verifying that the random account balance is the transferred out amount would also work.

PR #583, opened 2021-05-11
[1Inch] Filtering solver that removes orders below certain threshold

fleupold:
1Inch solutions can be very gas inefficient (long routes + extra intermediary contract creating an extra layer of transfer indirection). This overhead might only make sense for orders of a certain minimum size.

This PR adds a general purpose solver that filters liquidity below a certain threshold and passes it into an inner solver. The threshold 5 ETH is currently hardcoded, but could potentially passed in as an argument if we wanted.

### Test Plan
Unit tests + running shadow traffic on mainnet and seeing that most orders are filtered (as they are tiny).


bh2smith (approved):

bh2smith (commented):
Very cool!

vkgnosis (approved):
Probably something we'll want to configure but good as is.

PR #584, opened 2021-05-11
Enforce some filters on get_trades and get_orders

vkgnosis:
These routes have the potential return a large number of database rows
which is expensive for the backend and shouldn't be requested by users
anyway (at least until we have pagination).

### Test Plan
adapted unit tests and manually tested with swagger


bh2smith (approved):
Seems alright for now. I guess pagination is a substantially more involved task then.

fleupold (approved):
OpenApi should be updated as well?

PR #588, opened 2021-05-11
Allow disabling trace api and log trace results

vkgnosis:
We want this in prod to reduce alerts but as a safety measure in case something goes wrong this adds more logging and allows us to disable the trace api without rebuilding.

### Test Plan
locally check that the argument gets set

### DIFF
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -50,6 +50,13 @@ struct Arguments {
         parse(try_from_str = shared::arguments::duration_from_seconds),
     )]
     min_order_validity_period: Duration,
+
+    /// Don't use the trace_callMany api that only some nodes support to check whether a token
+    /// should be denied.
+    /// Note that if a node does not support the api we still fallback to the less accurate call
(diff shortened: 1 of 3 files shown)


orderbook/src/main.rs
```diff
@@ -50,6 +50,13 @@ struct Arguments {
         parse(try_from_str = shared::arguments::duration_from_seconds),
     )]
     min_order_validity_period: Duration,
+
+    /// Don't use the trace_callMany api that only some nodes support to check whether a token
+    /// should be denied.
+    /// Note that if a node does not support the api we still fallback to the less accurate call
+    /// api.
+    #[structopt(long, env = "SKIP_TRACE_API")]
+    skip_trace_api: bool,
```
vkgnosis, line 59: I'm using the negation "skip" here instead of "enable" because you cannot define a default value for boolean arguments but I want the default to be that the api is enabled.

fleupold (approved):

nlordell (approved):

PR #593, opened 2021-05-12
Makes pools non-mandatory.

marcovc:
Now (that we think) we solved the island problem in the solver, we can finally make the pools non-mandatory.


vkgnosis (approved):

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2Fw48WeMIN73QsM%2Fgiphy.gif&token=snape" data-gitmeme-token="snape"></a>

marcovc:
Now (that we think) we solved the island problem in the solver, we can finally make the pools non-mandatory.

bh2smith:
How are the islands related to mandatory pools.

marcovc:
> How are the islands related to mandatory pools?

If an amm between token A and B is mandatory (i.e. must be used in a solution) then we're sure A will be connected to B in the solution. So assuming that our problem has only one connected component (with tokens as nodes, mandatory amm's as edges), we know that any solution to that problem will also be one connected component, in other words, there are no multiple disconnected islands in it. Therefore making pools mandatory is one easy way to avoid multiple islands. Fortunately it seems that we might have found another solution for this (this PR) that doesn't have the drawbacks of making all pools mandatory.

PR #580, opened 2021-05-11
Default to creating 1Inch swap without splitting

nlordell:
This PR configures the 1Inch solver make more gas efficient swaps by:
- To use a maximum of 1 swap "path"
- Setting a limit on gas prices
- Only allowing a single connector token.

### Test Plan

Added a test around new `Amount` type, previous tests continue to pass, including:
```
$ cargo test -- --ignored --nocapture solve_order_on_one
...
Spender {
    address: 0x11111112542d85b3ef69ae05771c2dccff4faa26,
}
test oneinch_solver::api::tests::oneinch_spender_address ... ok
Swap {
...
}
test oneinch_solver::api::tests::oneinch_swap_without_amount_checks_and_splitting ... ok
Swap {
...
}
test oneinch_solver::api::tests::oneinch_swap ... ok
Settlement {
...
}
test oneinch_solver::tests::solve_order_on_oneinch ... ok
...
```

PR #594, opened 2021-05-12
Re-adjust 1Inch Solver to be slightly more aggressive

fleupold:
More context: https://gnosisinc.slack.com/archives/CC3C5SMB5/p1620798729034000

We may have oversteered in #580 so this PR makes it so that we can split a trade into three main parts and each part into 3 subparts. Each main path can have two connector tokens, upper bounding our solution at 27 interactions if my math is correct (3 main paths with 2 intermediate hope each being split into 3 again).

We keep the limit of 750k gas per trade (excluding GP overhead). This value could in the future be adjusted depending on current gas price and amount sold.

cc @nlordell @koeppelmann open for suggestions.

### Test Plan
only a config change, so 🤷

### DIFF
--- a/solver/src/oneinch_solver.rs
+++ b/solver/src/oneinch_solver.rs
@@ -77,8 +77,8 @@ impl OneInchSolver {
                 complexity_level: Some(Amount::new(2).unwrap()),
                 // Cap swap gas to 750K.
                 gas_limit: Some(750_000),
-                // Use only 1 main route for cheaper trades.
-                max_route_parts: Some(Amount::new(3).unwrap()),
+                // Use only 3 main route for cheaper trades.
+                main_route_parts: Some(Amount::new(3).unwrap()),
                 parts: Some(Amount::new(3).unwrap()),
             })
             .await?;
--- a/solver/src/oneinch_solver.rs
+++ b/solver/src/oneinch_solver.rs
@@ -73,13 +73,13 @@ impl OneInchSolver {
                 // Disable balance/allowance checks, as the settlement contract
                 // does not hold balances to traded tokens.
                 disable_estimate: Some(true),
-                // Use at most 1 connector token, reducing overall gas usage.
-                complexity_level: Some(Amount::new(1).unwrap()),
+                // Use at most 2 connector tokens
+                complexity_level: Some(Amount::new(2).unwrap()),
                 // Cap swap gas to 750K.
                 gas_limit: Some(750_000),
                 // Use only 1 main route for cheaper trades.
-                max_route_parts: Some(Amount::new(1).unwrap()),
-                parts: Some(Amount::new(1).unwrap()),
+                max_route_parts: Some(Amount::new(3).unwrap()),
+                parts: Some(Amount::new(3).unwrap()),
             })
             .await?;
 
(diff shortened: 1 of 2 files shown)


vkgnosis (approved):
Code wise all fine. What the best settings are, not sure.

solver/src/oneinch_solver.rs
```diff
@@ -73,13 +73,13 @@ impl OneInchSolver {
                 // Disable balance/allowance checks, as the settlement contract
                 // does not hold balances to traded tokens.
                 disable_estimate: Some(true),
-                // Use at most 1 connector token, reducing overall gas usage.
-                complexity_level: Some(Amount::new(1).unwrap()),
+                // Use at most 2 connector tokens
+                complexity_level: Some(Amount::new(2).unwrap()),
                 // Cap swap gas to 750K.
                 gas_limit: Some(750_000),
                 // Use only 1 main route for cheaper trades.
```
bh2smith, line 80: I think this comment should probably be a 3 now.
```suggestion
                // Use only 3 main route for cheaper trades.
```

bh2smith (approved):

bh2smith:
Maybe its already getting a bit out of hand with the run time configuration, but perhaps these parameters could be made part of the `LiquiditySource::1Inch` enum that is passed in at runtime, with some default values. Something like `1inch(1, 3)`

fleupold:
> Does this mean that the route can only be split in thirds?

It can be split into three unequal parts as well (e.g. 60, 30, 10).

Looking at the 1Inch API response, I can see that `protocols` returns a three dimensional array. After some manual trial and error (during which I noticed that `max_route_parts` should be called `main_route_parts` and that `complexityLevel` needs to be escaped as a string not an int 🤦 ) I come to the conclusion that `main_route_parts` is the max length of the first level array, `complexityLevel` is the depth of each main route array (number of connectors) and `parts` the max length of the last array (in how many chunks each hop can be split).

That is 
```js
protocol: Protocol[main_route_parts][complexityLevel][parts]
```

Here is a sample 1Inch route that shows what is what:

Screenshot 2021-05-13 at 12 33 28

This indicates that my original calculation (max 27 interactions) should be correct.

nlordell:
Nice :detective:

nlordell (approved):
Looks great! Nice catch on the URL issues!

solver/src/oneinch_solver/api.rs
```diff
@@ -372,9 +373,9 @@ mod tests {
                 &fromAddress=0x00000000219ab540356cbb839cbe05303d7705fa\
                 &slippage=0.5\
                 &disableEstimate=true\
-                &complexityLevel=1\
+                &complexityLevel=%271%27\
```
bh2smith, line 376: What is 271 and 27? Are these typos?

solver/src/oneinch_solver/api.rs
```diff
@@ -130,16 +130,17 @@ impl SwapQuery {
                 .append_pair("disableEstimate", &disable_estimate.to_string());
         }
         if let Some(complexity_level) = self.complexity_level {
+            // complexity level needs to be encoded as a string despite being an in (https://docs.1inch.io/api/quote-swap)
```
bh2smith, line 133: Despite being an... what?

solver/src/oneinch_solver/api.rs
fleupold: it's escaped ' characters

solver/src/oneinch_solver/api.rs
fleupold: int

PR #595, opened 2021-05-12
Introduce a TokenList struct

fleupold:
This will allow us to easily configure a whitelist of tokens for which we are happy to market make using internal buffers. I'm thinking about using https://tokens.coingecko.com/uniswap/all.json

### Test Plan
Unit tests.

### DIFF
--- a/shared/src/macros.rs
+++ b/shared/src/macros.rs
@@ -1,0 +1,6 @@
+#[macro_export]
+macro_rules! addr {
+    ($val:literal) => {
+        ::ethcontract::H160(::hex_literal::hex!($val))
+    };
+}
--- a/shared/src/http.rs
+++ b/shared/src/http.rs
@@ -1,0 +1,20 @@
+use std::time::Duration;
+
+use anyhow::Result;
+use reqwest::{
+    header::{HeaderValue, ACCEPT},
+    Client, ClientBuilder,
+};
+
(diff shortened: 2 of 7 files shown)


vkgnosis (approved):

bh2smith:
One concern I have with the tokenlist you suggested is that it contains NAWA (a token that was recently rug pulled).

shared/src/macros.rs
```diff
@@ -0,0 +1,6 @@
+#[macro_export]
+macro_rules! addr {
+    ($val:literal) => {
+        ::ethcontract::H160(::hex_literal::hex!($val))
+    };
```
bh2smith, line 5: `Fancy!(); MacroMan!();`

bh2smith (approved):
Very nice `TokenList` struct and implemented functions. I can see this coming in handy for other projects.

shared/src/macros.rs
fleupold: this was just move, so not my 🧑‍🎨

fleupold:
> One concern I have with the tokenlist you suggested is that it contains NAWA (a token that was recently rug pulled). It is quite a vast list and maybe we want to tone the default down a bit.

The list can be configured via config arguments, so we can use whatever condensed list we like.

PR #596, opened 2021-05-12
IntermediateSettlement

marcovc:
Creates a structured representation of the solution that is uniform across solvers.

## Motivation

To be able to implement better post-processing of solver solutions, in particular for using our liquidity buffers, we need to deal with concepts such as executed orders, executed amm's in a uniform way. However, currently solvers output directly into a Settlement, a close-to-the-blockchain representation, where these concepts are already serialized into unstructured "interactions" that contain not only this information, but also other low level uninteresting things like increasing allowances.

To help to give a motivation of what this PR enables in the future, consider an example of a solution with one order and one uniswap:

```
order_execution: 2A for 2B
uniswap_execution: 2B for 2A
```

If we have 1 B in our internal liquidity buffer, we would like to replace the solution by

```
order_execution: 2A for 2B
uniswap_execution: 1B for 1A
```

## Idea

There was already an IntermediateSettlement in the http_solver directory, which was almost what we want. This PR makes this available for all solvers. The process now is the following:

[Problem] -> Solver ->  [IntermediateSettlement] -> PostProcessing -> [Settlement]

It is assumed that post processing does not change prices, and just changes executed amounts and adds/removes uniswaps.

The Solver::solve() method maps "Problem" to "IntermediateSettlement". Additionally, since each solver knows best how to convert an "IntermediateSettlement" to a "Settlement" (aka "finalizing the settlement"), this is also new Solver method. I made a little detour here and created instead a method that returns a SettlementFinalizing class that implements this logic, since this part is common to most solvers.

## Alternatives

To me the current codebase is already more complex than it should, and this PR doesn't help. So I was also considering this alternatives:

1. Add some structure to Interactions: For example a Settlement could be a set of chunks of Interactions, where each chunk would have introspection capabilities so we could ask e.g. what order is responsible for it, and also to update it.
   Pros: No IntermediateSettlement concept.
   Cons: Seems hard to make interactions update themselves without coupling them with the solver that created them.

2. Make each solver responsible for handling internal buffers and whatever other preprocessing we will want to add in the future.
  Pros: No IntermediateSettlement concept. Each solver knows better what is the cost of updating its initial solution.
  Cons: There's going to be postprocessing code repetition, unless we factored it out in which case we will end up with something uniform that represents the solution, or in other orders, the IntermediateSettlement again.

3. We could run the solvers twice. After getting the first solution, we see what we can do with our buffers, decrease the user orders to take that into account, and solve the problem again.
  Pros/Cons: Same as previous.

## Drawbacks or proposed solution

The most relevant in my view, is the fact finalizing an intermediate settlement (i.e. IntermediateSettlement -> Settlement) may be as costly as Problem->IntermediateSettlement. At least this seems to be the case in the 1inch solver in case we modify one of the orders, since I believe it needs to query again 1inch for the best path. Other solvers are fine.

This all turned out to be more complex than I thought, so I'd be fine for trying something completely different in another PR in case I overlooked an easier solution.

### Test Plan

Updated existing tests, and they all pass.


solver/src/internal_liquidity.rs
```diff
@@ -0,0 +1,65 @@
...
+struct InternalLiquidityManager {
+    fetcher: Box<dyn InternalLiquidityFetching>,
+    allowed_tokens: Vec<H160>,
+}
+
+impl InternalLiquidityManager {
+    pub fn transform(settlement: Settlement) -> Settlement {
+        settlement.
+    }
+}
```
marcovc, line 65: Oops I just realized: this file wasn't supposed to be in this PR!

fleupold:
I would like to explore the route a bit more where we add methods to the `Liquidity`/`Settlable` abstractions that we already have. What is it that we need to know about an a piece of liquidity in the post-processing step? 

Is it the executed in/out amounts + tokens (and potentially adjusting them based on buffers we have)? 

Maybe this could be achieved by changing the `Settleable` trait to:

```rust
pub trait Settleable {
    // Some associated type telling us how an execution is represented 
    // could also be a concrete struct (token in, amount in, token_out, amount_out)
    type Execution;

    // Applies the desired execution on the current object (this could have the nice side effect, that e.g. pool weight updates could be reflected if the liquidity is used again in a second pass of e.g. the baseline solver (today price movements by an earlier pass are ignored).
    fn settle(&mut self, execution: Execution) -> Result<()>;

    // Encodes the current "execution" of self in the SettlementEncoder object (e.g. adding the interactions to the execution plan like today)
    fn encode(&self, encoder: &mut SettlementEncoder) -> Result<()>;

    // Allowing post processing code to introspect the execution state of the piece of liquidity
    fn get_current_execution(&self) -> Execution;

    // ... + whatever other methods post processing logic requires
}

```

The solvers would take a `Vec<Liquidity>`, invoke `settle` on the items they care about and add them to the `Settlement` instance (via `with_liquidity`).

The `Settlement` struct already serves as a kind of "intermediate settlement" today. Instead of encoding `Settleable` liquidity into interactions immediately, we could store a `Vec<Settleable>` (to which the desired exec amounts have been applied) on the `Settlement` struct. 
This could then be used by post processing logic (like ordering, merging, "internal market making". We might need to add some more methods to `Settleable` to allow for adjustments to the current execution.

The key advantage of this in my opinion is that if I add a new liquidity source I can implement all `Settleable` methods in one place (without having to know much about the solvers and post-processing logic). My hope is that different types of liquidity can be abstracted under a common umbrella like that and changes can be more local.

The step of turning the intermediate Settlement to the EncodedSettlement would remain canonical (no custom logic in each solver). Once we are done with post-processing we would "encode" the `Settlement` by iterating over all `Settleable` objects and calling `encode` on them.

If we can implement a good enough ordering algorithm we might even be able to get rid of `SettlementEncoder` and assume `encode` will be called in the right order (at the moment each liquidity may know about its relative priority in the execution plan and insert itself accordingly, but if they are sorted correctly, this wouldn't be necessary).

marcovc:
> What is it that we need to know about an a piece of liquidity in the post-processing step?

I'd say at least:

Accessors:
- If it is a limit order, or a liquidity order
- What are the tokens and the executed sell and executed buy amounts.

Modifiers:
- Remove a liquidity order.
- Update sell amount of liquidity order.

> The Settlement struct already serves as a kind of "intermediate settlement" today. Instead of encoding Settleable liquidity into interactions immediately, we could store a Vec<Settleable> (to which the desired exec amounts have been applied) on the Settlement struct.

I'm currently a bit lost on how the settlement encoding works, but for example the 1inch solver adds its main swap to the execution plan, and it doesn't look it's a Settleable.

fleupold:
@nlordell can also comment here, but iirc the main reason we have the SettlementEncoder as an abstraction in the middle is that each Liquidity type may have some proprietary knowledge on where it needs to put its interactions in the list (in terms of ordering).

Therefore, we ask each liquidity instance to "add itself" to the solution by calling `settlable.encode(encoder)`. A trade buying ETH can then insert a custom Unwrap Interactions (which goes at the end of the execution plan).

If we had an execution planner as a post-processing step that could e.g. based on required input and produced output call encode in the right order, we might not need the `SettlementEncoder` object anymore (and could just ask each liquidity to turn itself into a bunch of interactions.

> If it is a limit order, or a liquidity order

Why do you care about the difference between the two? Because you don't want to manipulate the amounts of "protected" orders?

marcovc:
> Why do you care about the difference between the two? Because you don't want to manipulate the amounts of "protected" orders?

Because liquidity orders (orders going to uniswap) we can change, or even remove them completely, at will. On the other hand, limit orders (user orders) we need to fullfill exactly to make sure the objective stays the same (although we can decide how to best fund them).

marcovc:
If I understand correctly, your suggestion doesn't seem to different from this PR: Instead of outputing Interactions, solvers output something (you call it Settleable, I was calling it IntermediateSettlement), that is introspectable and updatable, which later gets output into interactions. 

I guess the main difference is that I was putting the responsability of the "IntermediateThing" -> "Interaction" conversion into the solver (since from what I could tell solvers encode themselves differently), while you're suggesting that this could be factored out, and hence associated with the "IntermediateThing".

I like your approach of not introducing yet another abstraction, as long as we can pull it off :)

bh2smith:
This PR seems to have become stale and outdated @marcovc  - is this something we can close?

marcovc:
Closing as this seems to be not the best way to solve the problem.

PR #597, opened 2021-05-12
Only executing settlements without interactions if all sell tokens are in allowList

fleupold:
To make us less attackable. Using an Option in case we cannot fetch the list to not crash but instead just not market make between our own tokens.

### Test Plan
Added a unit test for the added logic in the driver. E2E test for this feature was adapted and still passes.

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -20,7 +20,7 @@ use itertools::{Either, Itertools};
 use model::order::BUY_ETH_ADDRESS;
 use num::BigRational;
 use primitive_types::H160;
-use shared::{price_estimate::PriceEstimating, Web3};
+use shared::{price_estimate::PriceEstimating, token_list::TokenList, Web3};
 use std::{
     collections::{HashMap, HashSet},
     sync::Arc,
@@ -46,6 +46,7 @@ pub struct Driver {
     network_id: String,
     max_merged_settlements: usize,
     solver_time_limit: Duration,
+    market_makable_token_list: Option<TokenList>,
 }
 impl Driver {
     #[allow(clippy::too_many_arguments)]
@@ -64,6 +65,7 @@ impl Driver {
         network_id: String,
         max_merged_settlements: usize,
         solver_time_limit: Duration,
+        market_makable_token_list: Option<TokenList>,
     ) -> Self {
         Self {
             settlement_contract,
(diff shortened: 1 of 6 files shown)


vkgnosis (approved):

solver/src/main.rs
```diff
@@ -117,6 +118,15 @@ struct Arguments {
...
     min_order_size_one_inch: U256,
+
+    /// The list of tokens our settlement contract is willing to buy when settling trades
+    /// without external liquidity
+    #[structopt(
+        long,
+        env = "MARKET_MAKEABLE_TOKEN_LIST",
+        default_value = "https://tokens.coingecko.com/uniswap/all.json"
+    )]
+    market_makable_token_list: String,
```
bh2smith, line 129: I suppose there are plenty of reasons this should be different from the base tokens we already use. 

I like the default value, but it seems like quite a vast list of tokens I've never heard of. In fact this list even includes NAWA (the token that was the reason for introducing this).

solver/src/driver.rs
```diff
@@ -428,14 +441,31 @@ fn liquidity_with_price(
     liquidity
 }
 
+fn is_only_selling_trusted_tokens(settlement: Settlement, token_list: &TokenList) -> bool {
```
marcovc, line 444: Just a question: why isn't settlement passed by reference? Would that avoid cloning the settlement on the caller?

solver/src/driver.rs
```diff
@@ -156,9 +159,19 @@ impl Driver {
...
+        // We don't want to buy tokens that we don't trust. If no list is set, we settle with external liquidity.
+        if !self
+            .market_makable_token_list
+            .as_ref()
+            .map(|list| is_only_selling_trusted_tokens(settlement.clone().into(), list))
+            .unwrap_or(false)
+        {
+            return Ok(false);
+        }
+
```
marcovc, line 172: Just checking, when we can't get the trusted token list then we never use buffers, right?

marcovc (approved):
Looks good!

solver/src/driver.rs
fleupold: Good point, that's possible (although I need to access the field then directly instead of using into())

solver/src/driver.rs
fleupold: Yes, this is to be more "safe" and have this feature opt-in.

PR #598, opened 2021-05-13
Adds fees to objective computation.

marcovc:
Objective computation in the driver was not taking into account the order fees. This PR fixes that.

### DIFF
--- a/solver/src/settlement.rs
+++ b/solver/src/settlement.rs
@@ -140,8 +140,30 @@ impl Settlement {
         match self.encoder.total_surplus(&external_prices) {
             Some(value) => value,
             None => {
-                tracing::error!("Overflow computing objective value for: {:?}", self);
+                tracing::error!("Overflow computing surplus value for: {:?}", self);
                 num::zero()
+            }
+        }
+    }
+
+    // For now this computes the fees of all EOA trades.
+    pub fn total_fees(&self) -> U256 {
+        match self.encoder.total_fees() {
+            Some(value) => value,
+            None => {
+                tracing::error!("Overflow computing fees value for: {:?}", self);
+                0.into()
+            }
+        }
+    }
+
+    // For now this computes the fees of all EOA trades.
+    pub fn total_unsubsidized_fees(&self, fee_discount_factor: f64) -> U256 {
+        match self.encoder.total_unsubsidized_fees(fee_discount_factor) {
+            Some(value) => value,
+            None => {
+                tracing::error!("Overflow computing fees value for: {:?}", self);
+                0.into()
             }
         }
     }
--- a/solver/src/settlement.rs
+++ b/solver/src/settlement.rs
@@ -149,9 +149,13 @@ impl Settlement {
     // For now this computes the fees of all EOA trades.
     pub fn total_unsubsidized_fees(
         &self,
-        normalizing_prices: &HashMap<H160, BigRational>,        
-        fee_discount_factor: BigRational) -> BigRational {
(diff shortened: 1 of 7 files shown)


vkgnosis (approved):

solver/src/settlement/settlement_encoder.rs
```diff
@@ -208,6 +208,20 @@ impl SettlementEncoder {
         })
     }
 
+    pub fn total_fees(&self) -> Option<U256> {
+        self.trades.iter().fold(Some(0.into()), |acc, trade| {
+            let fees = trade.order.order_creation.fee_amount;
+            Some(acc? + fees)
```
fleupold, line 214: does this not have to be a checked add?

solver/src/settlement.rs
```diff
@@ -140,12 +140,34 @@ impl Settlement {
...
+    }
+
+    // For now this computes the fees of all EOA trades.
+    pub fn total_unsubsidized_fees(&self, fee_discount_factor: f64) -> U256 {
+        match self.encoder.total_unsubsidized_fees(fee_discount_factor) {
+            Some(value) => value,
+            None => {
+                tracing::error!("Overflow computing fees value for: {:?}", self);
+                0.into()
+            }
```
fleupold, line 167: nit: shouldn't the error handling logic be in the encoder? Or do you expect people calling encoder.total_unsubsidized_fees that don't care about the extra context?

solver/src/settlement.rs
```diff
@@ -140,12 +140,34 @@ impl Settlement {
...
             None => {
-                tracing::error!("Overflow computing objective value for: {:?}", self);
+                tracing::error!("Overflow computing surplus value for: {:?}", self);
                 num::zero()
             }
         }
     }
 
+    // For now this computes the fees of all EOA trades.
+    pub fn total_fees(&self) -> U256 {
```
fleupold, line 150: nit: doesn't seem used?

solver/src/settlement/settlement_encoder.rs
```diff
@@ -208,6 +208,20 @@ impl SettlementEncoder {
...
+        })
+    }
+
+    pub fn total_unsubsidized_fees(&self, fee_discount_factor: f64) -> Option<U256> {
```
fleupold, line 218: If you wanted to save some boilerplate you could get rid of this method and just divide total_fees by the fee_discount_factor in the layer above (but not a huge difference).

fleupold: Maybe a small unit test would be nice?

fleupold (approved):

solver/src/settlement/settlement_encoder.rs
nlordell, line 214:
```suggestion
            acc?.checked_add(fees)
```

nlordell (approved):
Looks good. Just missing overflow checking in one spot AFAICT.

solver/src/settlement.rs
marcovc: I was following along the existing logic around `total_surplus()` which defaults to zero for None. Do you prefer to keep it consistent or change this one to return an Option? If the latter, should we change total_surplus in a subsequent PR too?

marcovc:
I just realized there's something wrong here. In a previous PR I assumed that fees of limit orders are always given in the sell token, can't recall how I arrived to that conclusion, probably asked someone about it. But here I'm simply adding the amounts, instead of amounts*p(sell_volume).

bh2smith (approved):

marcovc:
Closed in favour of #692

PR #599, opened 2021-05-13
Makes network native token the ref token for http solver.

marcovc:
Because @twalth3r suggested that this can improve numerical stability of the solver.

### DIFF
--- a/solver/src/http_solver/model.rs
+++ b/solver/src/http_solver/model.rs
@@ -43,6 +43,7 @@ pub struct UniswapModel {
 pub struct TokenInfoModel {
     pub decimals: Option<u32>,
     pub external_price: Option<f64>,
+    pub normalize_priority: Option<u64>,
 }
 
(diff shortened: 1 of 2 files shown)


vkgnosis (approved):

solver/src/http_solver/model.rs
```diff
@@ -43,6 +43,7 @@ pub struct UniswapModel {
 pub struct TokenInfoModel {
     pub decimals: Option<u32>,
     pub external_price: Option<f64>,
+    pub normalize_priority: Option<u64>,
```
fleupold, line 46: Why Option if it's always set?

fleupold (approved):

solver/src/http_solver/model.rs
marcovc: The http solver API doesn't require it (it defaults to zero if it's not set).

PR #604, opened 2021-05-17
Add component to detect bad tokens more accurately

vkgnosis:
Detects whether a token is "bad" (works in unexpected ways that are problematic for solving) by
simulating several transfers of a token. To find an initial address to transfer from we use
the amm pair providers.

In a future PR this will be used as a global bad-token cache that gets populated whenever we encounter a new token.
We can also extend the type of checks done but I don't want to bikeshed those too much in this PR. We can collect them in an issue.

We should keep the existing check on order creation as an additional safe guard and use this new one in addition.

### Test Plan
I added already added a real mainnet test and commented which tokens are wrongly detected as bad.
TODO: add a unit test for handle_response

### DIFF
--- a/orderbook/src/bad_token.rs
+++ b/orderbook/src/bad_token.rs
@@ -161,12 +161,15 @@ impl BadTokenDetector {
         requests.push(call_request(None, token, tx));
         // 3
         let recipient = Self::arbitrary_recipient();
+        let tx = instance.balance_of(recipient).m.tx;
+        requests.push(call_request(None, token, tx));
+        // 4
         let tx = instance.transfer(recipient, amount).tx;
         requests.push(call_request(Some(self.settlement_contract), token, tx));
-        // 4
+        // 5
         let tx = instance.balance_of(self.settlement_contract).m.tx;
         requests.push(call_request(None, token, tx));
-        // 5
+        // 6
         let tx = instance.balance_of(recipient).m.tx;
         requests.push(call_request(None, token, tx));
 
@@ -174,13 +177,13 @@ impl BadTokenDetector {
     }
 
     fn handle_response(traces: &[BlockTrace], amount: U256) -> Result<TokenQuality> {
-        ensure!(traces.len() == 6, "unexpected number of traces");
+        ensure!(traces.len() == 7, "unexpected number of traces");
 
         let gas_in = match ensure_transaction_ok_and_get_gas(&traces[1])? {
             Ok(gas) => gas,
             Err(reason) => return Ok(TokenQuality::Bad { reason }),
         };
-        let gas_out = match ensure_transaction_ok_and_get_gas(&traces[3])? {
+        let gas_out = match ensure_transaction_ok_and_get_gas(&traces[4])? {
             Ok(gas) => gas,
             Err(reason) => return Ok(TokenQuality::Bad { reason }),
         };
@@ -201,7 +204,7 @@ impl BadTokenDetector {
                 })
             }
         };
-        let balance_after_out = match decode_u256(&traces[4]) {
+        let balance_after_out = match decode_u256(&traces[5]) {
             Ok(balance) => balance,
             Err(_) => {
                 return Ok(TokenQuality::Bad {
@@ -209,11 +212,21 @@ impl BadTokenDetector {
                 })
             }
         };
-        let balance_of_receiver = match decode_u256(&traces[5]) {
+
+        let balance_recpient_before = match decode_u256(&traces[3]) {
             Ok(balance) => balance,
             Err(_) => {
                 return Ok(TokenQuality::Bad {
-                    reason: "can't decode recipient balance".to_string(),
+                    reason: "can't decode recipient balance before".to_string(),
+                })
+            }
+        };
+
+        let balance_recipient_after = match decode_u256(&traces[6]) {
+            Ok(balance) => balance,
+            Err(_) => {
+                return Ok(TokenQuality::Bad {
+                    reason: "can't decode recipient balance after".to_string(),
                 })
             }
         };
@@ -233,9 +246,9 @@ impl BadTokenDetector {
                 reason: "balance after out transfer does not match".to_string(),
             });
         }
-        if balance_of_receiver != amount {
+        if balance_recpient_before + amount != balance_recipient_after {
             return Ok(TokenQuality::Bad {
-                reason: "balance of receiver does not match".to_string(),
+                reason: "balance of recipient does not match".to_string(),
             });
         }
 
@@ -357,6 +370,13 @@ mod tests {
             },
             BlockTrace {
                 output: encode_u256(1.into()),
+                trace: None,
+                vm_trace: None,
+                state_diff: None,
+                transaction_hash: None,
+            },
+            BlockTrace {
+                output: encode_u256(0.into()),
                 trace: None,
                 vm_trace: None,
                 state_diff: None,
--- a/orderbook/src/bad_token.rs
+++ b/orderbook/src/bad_token.rs
@@ -18,13 +18,13 @@ use web3::{
 };
 
 #[derive(Debug, Clone, Eq, PartialEq)]
-pub enum TokenInfo {
+pub enum TokenQuality {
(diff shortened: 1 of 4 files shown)


orderbook/src/bad_token.rs
```diff
@@ -0,0 +1,465 @@
...
+                _ => biggest_balance = Some((address, balance)),
+            }
+        }
+        Ok(biggest_balance)
+    }
+
+    fn create_trace_request(&self, token: H160, amount: U256, take_from: H160) -> Vec<CallRequest> {
+        let instance = ERC20::at(&self.web3, token);
+
+        // We transfer one unit (base on decimals) of the token from the amm pool into the
```
nlordell, line 134: Is this true? Don't we transfer `amount` of token?

orderbook/src/bad_token.rs
```diff
@@ -0,0 +1,465 @@
...
+        let tx = instance.transfer(self.settlement_contract, amount).tx;
+        requests.push(call_request(Some(take_from), token, tx));
+        // 2
+        let tx = instance.balance_of(self.settlement_contract).m.tx;
+        requests.push(call_request(None, token, tx));
+        // 3
+        // For the out transfer we use an arbitrary address without balance to detect tokens that
+        // usually apply fees but not if the the sender or receiver is specifically exempt like
+        // their own uniswap pools.
+        let recipient = H160::from_low_u64_le(1);
```
nlordell, line 155: We should make this:
```rust
H160(hex!("cd7328a5D376D5530f054EAF0B9D235a4Fd36059"));
```

Which is:
```rust
PrivateKey::from_hex(keccak256("moo")).unwrap().public_address();
```

(This is mostly a joke and for fun :stuck_out_tongue:).

orderbook/src/bad_token.rs
```diff
@@ -0,0 +1,465 @@
...
+        let gas_out = match ensure_transaction_ok_and_get_gas(&traces[3])? {
+            Ok(gas) => gas,
+            Err(reason) => return Ok(TokenInfo::Bad { reason }),
+        };
+
+        let balance_before_in = match decode_u256(&traces[0]) {
+            Ok(balance) => balance,
+            Err(_) => {
+                return Ok(TokenInfo::Bad {
+                    reason: "can't decode balance".to_string(),
```
nlordell, line 184: nit: maybe qualify which balance can't be decoded:
- `can't decode initial settlement balance`
- `can't decode final settlement balance`
- `can't decode initial recipient balance`
- `can't decode final recipient balance`

nlordell (approved):
Looks great!

orderbook/src/bad_token.rs
```diff
@@ -0,0 +1,465 @@
...
+            }
+        };
+        let amount = match U256::from(10).checked_pow(decimals.into()) {
+            Some(amount) => {
+                tracing::debug!("testing token {:?} with amount {}", token, amount);
+                amount
+            }
+            None => {
+                return Ok(TokenInfo::Bad {
+                    reason: "decimals too large".to_string(),
```
bh2smith, line 64: Is the a max to this field? What make decimals too long here? Are you referring to the actually value passed in as having too much precision for the token? If that's the case I would think we could just truncate to the maximum precision. Although I suppose that would screw up the order signature validation.

orderbook/src/bad_token.rs
```diff
@@ -0,0 +1,465 @@
...
+        };
+
+        let take_from = match self.find_address_owning_token(token).await? {
+            Some((address, balance)) if balance > amount => {
+                tracing::debug!("testing token {:?} with pool {:?}", token, address);
+                address
+            }
+            _ => {
+                return Ok(TokenInfo::Bad {
+                    reason: "no amm pool with base tokens with enough balance".to_string(),
```
bh2smith, line 76: Nit: not sure about having "with" twice so close to each other here. Could replace the second with "having".

bh2smith (approved):
Very nice work here. Looking forward to shipping this masterpiece.

orderbook/src/bad_token.rs
vkgnosis: From the perspective of this function it is amount but where it is called it is always set to this. Will change it to say amount here.

orderbook/src/bad_token.rs
vkgnosis: This error happens when we can't create a base unit of the token (10 ^ decimals) because the resulting number doesn't fit in a U256. I don't think any reasonable token would have it set this way.

orderbook/src/bad_token.rs
vkgnosis: good idea

fleupold (approved):
Nice, do you know how long one detection roughly takes e2e?

orderbook/src/bad_token.rs
```diff
@@ -0,0 +1,577 @@
...
+        let request = self.create_trace_request(token, amount, take_from);
+        let traces = trace_many::trace_many(request, &self.web3).await?;
+        Self::handle_response(&traces, amount)
+    }
+
+    // Based on amm pools find the address with the largest amount of the token.
+    // Err if communication with the node failed.
+    // Ok(None) if there is no pool or getting the balance fails.
+    // Ok(address, balance) for an address that has this amount of balance of the token.
+    async fn find_address_owning_token(&self, token: H160) -> Result<Option<(H160, U256)>> {
```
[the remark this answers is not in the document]
vkgnosis: In some way that is an implementation detail because the function could find an owner of the token in another way. But I don't mind changing the name either.

orderbook/src/bad_token.rs
```diff
@@ -0,0 +1,577 @@
...
+use primitive_types::{H160, U256};
+use shared::{amm_pair_provider::AmmPairProvider, Web3};
+use std::collections::HashSet;
+use web3::{
+    signing::keccak256,
+    types::{BlockTrace, CallRequest, Res},
+};
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum TokenInfo {
```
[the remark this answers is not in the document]
vkgnosis: They are in different modules so I don't think it's a problem to have the same name but your suggestions are fine too.

orderbook/src/bad_token.rs
```diff
@@ -0,0 +1,577 @@
...
+            }
+        }
+        Ok(biggest_balance)
+    }
+
+    // For the out transfer we use an arbitrary address without balance to detect tokens that
+    // usually apply fees but not if the the sender or receiver is specifically exempt like
+    // their own uniswap pools.
+    fn arbitrary_recipient() -> H160 {
+        PrivateKey::from_raw(keccak256(b"moo"))
```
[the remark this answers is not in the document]
vkgnosis: We could, but also if they did fund it we could take their funds seeing as we have the private key.

vkgnosis:
> Nice, do you know how long one detection roughly takes e2e?
In the mainnet test it takes maybe 300 ms. But this will be behind a cache anyway.

orderbook/src/bad_token.rs
```diff
@@ -0,0 +1,577 @@
...
+        };
+
+        let take_from = match self.find_address_owning_token(token).await? {
+            Some((address, balance)) if balance > amount => {
+                tracing::debug!("testing token {:?} with pool {:?}", token, address);
+                address
+            }
+            _ => {
+                return Ok(TokenInfo::Bad {
+                    reason: "no amm pool with base tokens has enough balance".to_string(),
```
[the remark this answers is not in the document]
vkgnosis: Oh yeah, it should be that.

orderbook/src/bad_token.rs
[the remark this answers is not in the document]
vkgnosis: I am going to add another balance check to the recipient address before the send to catch this.

orderbook/src/bad_token.rs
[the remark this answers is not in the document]
fleupold: When using symbol search in the IDE I find it still slightly more convenient to not have the same struct name overloaded multiple times.

PR #605, opened 2021-05-18
Refactors objective function computation.

marcovc:
Additionally adds comments, logging, and tests.

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -267,8 +267,8 @@ impl Driver {
     ) -> Vec<RatedSettlement> {
         use futures::stream::StreamExt;
 
-        // TODO: 1e18 should be 10**(native_token_decimals)
-        let gas_price_wei = BigRational::from_float(gas_price_wei).unwrap();
+        let gas_price_wei = BigRational::from_float(gas_price_wei).unwrap()
+            * prices.get(&self.native_token).unwrap();
 
         futures::stream::iter(settlements)
             .filter_map(|settlement| async {
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -410,6 +410,7 @@ impl Driver {
                 tracing::info!("settlement without onchain liquidity");
             }
 
+            tracing::debug!("winning settlement: {:?}", settlement);
             self.submit_settlement(settlement.clone()).await;
             self.report_matched_but_unsettled_orders(
                 &Settlement::from(settlement),
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -351,7 +351,7 @@ impl Driver {
 
         let rated_settlements = self.rate_settlements(settlements, &estimated_prices).await;
         // TODO: 1e18 should be 10**(native_token_decimals)
-        let gas_price_normalized: BigRational = estimated_prices.get(&self.native_token).unwrap()
+        let gas_price_normalized: BigRational = num::one::<BigRational>()
             / BigRational::from_integer(1_000_000_000_000_000_000_u128.into());
 
         if let Some(mut settlement) = rated_settlements.into_iter().max_by(|a, b| {
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -267,8 +267,11 @@ impl Driver {
     ) -> Vec<RatedSettlement> {
         use futures::stream::StreamExt;
 
-        let gas_price_wei = BigRational::from_float(gas_price_wei).unwrap()
-            * prices.get(&self.native_token).unwrap();
+        // Normalize gas_price_wei to the native token price in the prices vector.
+        let gas_price_wei = BigRational::from_float(gas_price_wei).expect("Invalid gas price.")
+            * prices
+                .get(&self.native_token)
+                .expect("Price of native token must be known.");
 
         futures::stream::iter(settlements)
             .filter_map(|settlement| async {
@@ -287,7 +290,7 @@ impl Driver {
                     gas_price: gas_price_wei.clone(),
                 };
                 tracing::info!(
-                    "Objective value for solver {} is {}: surplus={}, gas_estimate={}, gas_price_normalized={}",
+                    "Objective value for solver {} is {}: surplus={}, gas_estimate={}, gas_price={}",
                     solver_name,
                     rated_settlement.objective_value(),
                     rated_settlement.surplus,
@@ -314,6 +317,8 @@ impl Driver {
         let estimated_prices =
             collect_estimated_prices(self.price_estimator.as_ref(), self.native_token, &liquidity)
                 .await;
+        tracing::debug!("estimated prices: {:?}", estimated_prices);
+
         let liquidity = liquidity_with_price(liquidity, &estimated_prices);
         self.metrics.liquidity_fetched(&liquidity);
 
@@ -382,6 +387,7 @@ impl Driver {
                 settlement = settlement.without_onchain_liquidity();
                 tracing::info!("settlement without onchain liquidity");
             }
+            tracing::debug!("winning settlement: {:?}", settlement);
             self.submit_settlement(settlement).await;
         }
 
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -502,7 +502,7 @@ mod tests {
             }),
         ];
         let prices = collect_estimated_prices(&price_estimator, native_token, &liquidity).await;
-        assert_eq!(prices.len(), 2);
+        assert_eq!(prices.len(), 4);
         assert!(prices.contains_key(&sell_token));
         assert!(prices.contains_key(&buy_token));
     }
@@ -527,7 +527,7 @@ mod tests {
             id: "0".into(),
         })];
         let prices = collect_estimated_prices(&price_estimator, native_token, &liquidity).await;
-        assert_eq!(prices.len(), 0);
+        assert_eq!(prices.len(), 2);
     }
 
     #[tokio::test]
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -301,15 +301,15 @@ impl Driver {
         let liquidity = liquidity_with_price(liquidity, &estimated_prices);
         self.metrics.liquidity_fetched(&liquidity);
 
-        let gas_price = self
+        let gas_price_wei = self
             .gas_price_estimator
             .estimate()
             .await
             .context("failed to estimate gas price")?;
-        tracing::debug!("solving with gas price of {}", gas_price);
+        tracing::debug!("solving with gas price of {}", gas_price_wei);
 
         let settlements = self
-            .run_solvers(liquidity, gas_price)
+            .run_solvers(liquidity, gas_price_wei)
             .await
             .filter_map(solver_settlements::filter_bad_settlements)
             .inspect(|(name, settlements)| {
@@ -350,10 +350,13 @@ impl Driver {
         }
 
         let rated_settlements = self.rate_settlements(settlements, &estimated_prices).await;
+        // TODO: 1e18 should be 10**(native_token_decimals)
+        let gas_price_normalized: BigRational = estimated_prices.get(&self.native_token).unwrap()
+            / BigRational::from_integer(1_000_000_000_000_000_000_u128.into());
 
         if let Some(mut settlement) = rated_settlements.into_iter().max_by(|a, b| {
-            a.objective_value(gas_price)
-                .cmp(&b.objective_value(gas_price))
+            a.objective_value(&gas_price_normalized)
+                .cmp(&b.objective_value(&gas_price_normalized))
         }) {
             // If we have enough buffer in the settlement contract to not use on-chain interactions, remove those
             if self
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -263,8 +263,13 @@ impl Driver {
         &self,
         settlements: Vec<SettlementWithSolver>,
         prices: &HashMap<H160, BigRational>,
+        gas_price_wei: f64,
     ) -> Vec<RatedSettlement> {
         use futures::stream::StreamExt;
+
+        // TODO: 1e18 should be 10**(native_token_decimals)
+        let gas_price_wei = BigRational::from_float(gas_price_wei).unwrap();
+
         futures::stream::iter(settlements)
             .filter_map(|settlement| async {
                 let surplus = settlement.settlement.total_surplus(prices);
@@ -274,11 +279,22 @@ impl Driver {
                 )
                 .await
                 .ok()?;
-                Some(RatedSettlement {
+                let solver_name = settlement.name;
+                let rated_settlement = RatedSettlement {
                     settlement,
                     surplus,
                     gas_estimate,
-                })
+                    gas_price: gas_price_wei.clone(),
+                };
+                tracing::info!(
+                    "Objective value for solver {} is {}: surplus={}, gas_estimate={}, gas_price_normalized={}",
+                    solver_name,
+                    rated_settlement.objective_value(),
+                    rated_settlement.surplus,
+                    rated_settlement.gas_estimate,
+                    rated_settlement.gas_price,
+                );
+                Some(rated_settlement)
             })
             .collect::<Vec<_>>()
             .await
@@ -349,15 +365,14 @@ impl Driver {
             self.metrics.settlement_simulation_failed(settlement.name);
         }
 
-        let rated_settlements = self.rate_settlements(settlements, &estimated_prices).await;
-        // TODO: 1e18 should be 10**(native_token_decimals)
-        let gas_price_normalized: BigRational = estimated_prices.get(&self.native_token).unwrap()
-            / BigRational::from_integer(1_000_000_000_000_000_000_u128.into());
+        let rated_settlements = self
+            .rate_settlements(settlements, &estimated_prices, gas_price_wei)
+            .await;
 
-        if let Some(mut settlement) = rated_settlements.into_iter().max_by(|a, b| {
-            a.objective_value(&gas_price_normalized)
-                .cmp(&b.objective_value(&gas_price_normalized))
-        }) {
+        if let Some(mut settlement) = rated_settlements
+            .into_iter()
+            .max_by(|a, b| a.objective_value().cmp(&b.objective_value()))
+        {
             // If we have enough buffer in the settlement contract to not use on-chain interactions, remove those
             if self
                 .can_settle_without_liquidity(&settlement)
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -351,7 +351,7 @@ impl Driver {
 
         let rated_settlements = self.rate_settlements(settlements, &estimated_prices).await;
         // TODO: 1e18 should be 10**(native_token_decimals)
-        let gas_price_normalized: BigRational = num::one::<BigRational>()
+        let gas_price_normalized: BigRational = estimated_prices.get(&self.native_token).unwrap()
             / BigRational::from_integer(1_000_000_000_000_000_000_u128.into());
 
         if let Some(mut settlement) = rated_settlements.into_iter().max_by(|a, b| {
@@ -384,6 +384,8 @@ pub async fn collect_estimated_prices(
     liquidity: &[Liquidity],
 ) -> HashMap<H160, BigRational> {
     // Computes set of traded tokens (limit orders only).
+    // NOTE: The native token is always added.
+
     let mut tokens = HashSet::new();
     for liquid in liquidity {
         if let Liquidity::Limit(limit_order) = liquid {
@@ -413,10 +415,12 @@ pub async fn collect_estimated_prices(
         })
         .collect();
 
-    // If the wrapped native token is in the price list (e.g. WETH), so should be the placeholder for its native counterpart
-    if let Some(price) = prices.get(&native_token).cloned() {
-        prices.insert(BUY_ETH_ADDRESS, price);
-    }
+    // Always include the native token.
+    prices.insert(native_token, num::one());
+
+    // And the placeholder for its native counterpart.
+    prices.insert(BUY_ETH_ADDRESS, num::one());
+
     prices
 }
 
--- a/solver/src/driver/solver_settlements.rs
+++ b/solver/src/driver/solver_settlements.rs
@@ -271,26 +271,37 @@ mod tests {
 
     #[test]
     fn compute_objective_value() {
-        // Surplus is 0.4 ETH
-        let surplus = BigRational::from_integer(4.into()) / BigRational::from_integer(10.into());
+        // Gas price is 2e-7 ETH
+        let gas_price = BigRational::from_integer(200_000_000_000_u128.into());
 
-        // Gas estimate is 500_000
-        let gas_estimate = BigRational::from_integer(500_000.into());
+        // Surplus1 is 0.4 ETH
+        let surplus1 = BigRational::from_integer(400_000_000_000_000_000_u128.into());
 
-        // Normalized gas price is 2e-7 ETH
-        let gas_price_normalized = BigRational::from_integer(200_000_000_000_u128.into())
-            / BigRational::from_integer(1_000_000_000_000_000_000_u128.into());
+        // Surplus2 is 1.0 ETH
+        let surplus2 = BigRational::from_integer(1_000_000_000_000_000_000_u128.into());
 
-        // Objective value is 4e-1 - 5e5 * 2e-7 = 4e-1 - 1e-1 = 0.3
-        let obj_value = RatedSettlement::compute_objective_value(
-            &surplus,
-            &gas_estimate,
-            &gas_price_normalized,
-        );
+        // Gas estimate1 is 500_000
+        let gas_estimate1 = BigRational::from_integer(500_000.into());
+
+        // Gas estimate2 is 100_000
+        let gas_estimate2 = BigRational::from_integer(100_000.into());
+
+        // Objective value 1 is 4e-1 - 5e5 * 2e-7 = 4e-1 - 1e-1 = 0.3 ETH
+        let obj_value1 =
+            RatedSettlement::compute_objective_value(&surplus1, &gas_estimate1, &gas_price);
 
         assert_eq!(
-            obj_value,
-            BigRational::from_integer(3.into()) / BigRational::from_integer(10.into())
+            obj_value1,
+            BigRational::from_integer(300_000_000_000_000_000_u128.into())
+        );
+
+        // Objective value 2 is 1 - 1e5 * 2e-7 = 1 - 2e-2 = 0.98 ETH
+        let obj_value2 =
+            RatedSettlement::compute_objective_value(&surplus2, &gas_estimate2, &gas_price);
+
+        assert_eq!(
+            obj_value2,
+            BigRational::from_integer(980_000_000_000_000_000_u128.into())
         );
     }
 }
--- a/solver/src/driver/solver_settlements.rs
+++ b/solver/src/driver/solver_settlements.rs
@@ -79,11 +79,19 @@ pub struct RatedSettlement {
 }
 
 impl RatedSettlement {
+    fn compute_objective_value(
+        surplus: &BigRational,
+        gas_estimate: &BigRational,
+        gas_price_normalized: &BigRational,
+    ) -> BigRational {
+        let cost = gas_estimate * gas_price_normalized;
+        surplus - cost
+    }
+
     // gas_price_normalized is the price of gas normalized to the found price vector.
     pub fn objective_value(&self, gas_price_normalized: &BigRational) -> BigRational {
         let gas_estimate = self.gas_estimate.to_big_rational();
-        let cost = gas_estimate * gas_price_normalized;
-        self.surplus.clone() - cost
+        Self::compute_objective_value(&self.surplus, &gas_estimate, gas_price_normalized)
     }
 }
 
@@ -256,5 +264,30 @@ mod tests {
         let settlement = Settlement::new(hashmap! {token0 => 0.into()});
         let settlements = vec![settlement.clone(), settlement];
         assert!(merge_at_most_settlements(1, settlements.into_iter()).is_none());
+    }
+
+    #[test]
+    fn compute_objective_value() {
+        // Surplus is 0.4 ETH
+        let surplus = BigRational::from_integer(4.into()) / BigRational::from_integer(10.into());
+
+        // Gas estimate is 500_000
+        let gas_estimate = BigRational::from_integer(500_000.into());
+
+        // Normalized gas price is 2e-7 ETH
+        let gas_price_normalized = BigRational::from_integer(200_000_000_000_u128.into())
+            / BigRational::from_integer(1_000_000_000_000_000_000_u128.into());
+
+        // Objective value is 4e-1 - 5e5 * 2e-7 = 4e-1 - 1e-1 = 0.3
+        let obj_value = RatedSettlement::compute_objective_value(
+            &surplus,
+            &gas_estimate,
+            &gas_price_normalized,
+        );
+
+        assert_eq!(
+            obj_value,
+            BigRational::from_integer(3.into()) / BigRational::from_integer(10.into())
+        );
     }
 }
--- a/solver/src/driver/solver_settlements.rs
+++ b/solver/src/driver/solver_settlements.rs
@@ -76,22 +76,24 @@ pub struct RatedSettlement {
     pub settlement: SettlementWithSolver,
     pub surplus: BigRational,
     pub gas_estimate: U256,
+    pub gas_price: BigRational,
 }
 
 impl RatedSettlement {
+    // gas_price_normalized is the price of gas normalized to the found price vector.
+    pub fn objective_value(&self) -> BigRational {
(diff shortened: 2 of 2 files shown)


nlordell (approved):

solver/src/driver.rs
```diff
@@ -350,10 +350,13 @@ impl Driver {
         }
 
         let rated_settlements = self.rate_settlements(settlements, &estimated_prices).await;
+        // TODO: 1e18 should be 10**(native_token_decimals)
+        let gas_price_normalized: BigRational = estimated_prices.get(&self.native_token).unwrap()
+            / BigRational::from_integer(1_000_000_000_000_000_000_u128.into());
```
vkgnosis, line 355: Looking at the implementation of `collect_estimated_prices` I don't think it is guaranteed to contain `native_token`. We do however calculate prices relative to the native token 
```rust
let denominator_token: H160 = native_token;
price_estimator.estimate_prices(&tokens, denominator_token)
```
which means it always has price `1`. 
I think we should either document that in `collect_estimated_prices` and hard code the same `1` here or change it to always put the native token in the price hash map.

marcovc: Yes, I just noticed this. What do you prefer?

marcovc: (I'd go with adding it to the price vector since it avoids problems of changing it in one place and not the other later on, but I don't know if it has side effects)

vkgnosis: Agree. It is more robust to always have it in the price list. So I would say we document in `collect_estimated_prices` that the result will always contain the native token.

vkgnosis (approved):

bh2smith (approved):
Good catch!

solver/src/driver.rs
```diff
@@ -350,10 +350,13 @@ impl Driver {
...
 
         if let Some(mut settlement) = rated_settlements.into_iter().max_by(|a, b| {
-            a.objective_value(gas_price)
-                .cmp(&b.objective_value(gas_price))
+            a.objective_value(&gas_price_normalized)
```
fleupold, line 358: Maybe we should change the way we compute the objective value from doing it in two places (once on RatedSettlement assuming surplus has already been cached) and once on Settlement directly so that we have one method where we can unit test our expectations given internal prices external prices and gas price?

fleupold (approved):
Can you add how you noticed this was wrong and what you did to "manually" test it's now showing the correct value? I still find it quite surprising we didn't notice this (e.g. 1Inch solver was used 15 times in the last 12 hours even though it's much less gas efficient than the Naive or Baseline solver and should have therefore not been chosen).

This is also probably a sign that we shouldn't scatter the objective value computation across multiple structs.

solver/src/driver.rs
marcovc: I tried to address this in the commit below, but not sure I did (also added a unit test for that function, hopefully with real world values).

marcovc:
I came across this bug when creating another PR (for adding fees to the objective) and reflecting if I did the "right thing" when adding costs.

The source of the problem was me not doing due diligence and automatically assuming that just because gas price was an `f64` it meant it was denominated in the native token (when in fact is in 1e-18 of the native token).

The consequence of the bug was that, because in our formula for ranking solutions `surplus - gas * gas_price`, gas_price is wrongly multiplied by 1e18, the surplus is largely ignored and we've been ranking the solutions solely based on gas.

In retrospective I think we have been likely saving more gas than we wanted, and potentially haven't offer the best "deal" to the customer. Really sorry about the mess.

@fleupold Let me know if I should merge the fix now, or make further changes.

fleupold (changes requested):
I'm still not sure this is the correct unit now. I added some logging to the driver (which I think we should add in any case to get a better understanding of the solutions and their objValues). 

By looking at prod shadow traffic of this PR I can see:

> 2021-05-18T20:32:07.246Z  WARN solver::driver: Objective for BaselineSolver: 18339349137697730526503740148553633737006678142809381062643/52674325129873780323080500000000000000000, surplus: 183393491376977305265037401485710623016597/526743251298737803230805, gas_estimate: 330874
2021-05-18T20:32:07.247Z  WARN solver::driver: Objective for solver::solver::SellVolumeFilteringSolver<solver::oneinch_solver::OneInchSolver>: 38072799107882993020728968003327982184102284403449290289383/105348650259747560646161000000000000000000, surplus: 38072799107882993020728968003386250206515/105348650259747560646161, gas_estimate: 553097

Surplus is ~0.36**e18** and cost is 0.06 ETH (normalized to 6**e-2**). Before this PR the final objective value was usually negative but greater than -1e18 which makes sense given that we are not considering the fees paid and usually the surplus for a small order is not that big (compared to its execution cost)

Now the values are different by 18 orders of magnitude so to me it looks like we are not considering gas costs at all anymore. The gas price used for computation here was 115000000000 wei.

Here is the associated solutions and price estimates for completeness

<details>

> 2021-05-18T20:32:06.399Z DEBUG solver::driver: solver BaselineSolver found solution:
 Settlement { encoder: SettlementEncoder { tokens: [0x3b73c1b2ea59835cbfcadade5462b6ab630d9890, 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48], clearing_prices: {0x3b73c1b2ea59835cbfcadade5462b6ab630d9890: 24935633384, 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48: 13227422021973872622547}, trades: [Trade { order: Order { order_meta_data: OrderMetaData { creation_date: 2021-05-18T20:31:24.300326Z, owner: 0xef42253884aac0a5df6efdc5fd04ac3cb1f485fd, uid: 0x0fd47ec904285aec4ef1b203a0534b77a0bca5de68cc2985c34717b5f0c1e855ef42253884aac0a5df6efdc5fd04ac3cb1f485fd60a42544, available_balance: Some(228222859775), executed_buy_amount: BigUint { data: [] }, executed_sell_amount: BigUint { data: [] }, executed_sell_amount_before_fees: BigUint { data: [] }, executed_fee_amount: BigUint { data: [] }, invalidated: false }, order_creation: OrderCreation { sell_token: 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48, buy_token: 0x3b73c1b2ea59835cbfcadade5462b6ab630d9890, receiver: Some(0xef42253884aac0a5df6efdc5fd04ac3cb1f485fd), sell_amount: 24935633384, buy_amount: 12573018808727190454846, valid_to: 1621370180, app_data: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1], fee_amount: 64366616, kind: Sell, partially_fillable: false, signature: Signature { r: 0x677eef4df33c0c02669014d35a561754749b1869e04ac11672835d6feaf438e5, s: 0x2ddef6c89f8e388e1dc954051a0717ea2bcbda5d1273cfdea86ad0222499d68a, v: 28 }, signing_scheme: Eip712 } }, sell_token_index: 1, buy_token_index: 0, executed_amount: 24935633384 }], execution_plan: [UniswapInteraction { router: IUniswapLikeRouter(0xd9e1ce17f2641f24ae83637ab66a2cca9c378b9f), settlement: GPv2Settlement(0x3328f5f2cecaf00a2443082b657cedeaf70bfaef), set_allowance: false, amount_out: 7240435876494084118, amount_in_max: 24960569018, token_in: 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48, token_out: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 }, UniswapInteraction { router: IUniswapLikeRouter(0x7a250d5630b4cf539739df2c5dacb4c659f2488d), settlement: GPv2Settlement(0x3328f5f2cecaf00a2443082b657cedeaf70bfaef), set_allowance: false, amount_out: 13227422021973872622547, amount_in_max: 7247676312370578203, token_in: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, token_out: 0x3b73c1b2ea59835cbfcadade5462b6ab630d9890 }], unwraps: [] } }
2021-05-18T20:32:06.401Z DEBUG solver::driver: solver solver::solver::SellVolumeFilteringSolver<solver::oneinch_solver::OneInchSolver> found solution:
 Settlement { encoder: SettlementEncoder { tokens: [0x3b73c1b2ea59835cbfcadade5462b6ab630d9890, 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48], clearing_prices: {0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48: 13252294880261259029821, 0x3b73c1b2ea59835cbfcadade5462b6ab630d9890: 24935633384}, trades: [Trade { order: Order { order_meta_data: OrderMetaData { creation_date: 2021-05-18T20:31:24.300326Z, owner: 0xef42253884aac0a5df6efdc5fd04ac3cb1f485fd, uid: 0x0fd47ec904285aec4ef1b203a0534b77a0bca5de68cc2985c34717b5f0c1e855ef42253884aac0a5df6efdc5fd04ac3cb1f485fd60a42544, available_balance: Some(228222859775), executed_buy_amount: BigUint { data: [] }, executed_sell_amount: BigUint { data: [] }, executed_sell_amount_before_fees: BigUint { data: [] }, executed_fee_amount: BigUint { data: [] }, invalidated: false }, order_creation: OrderCreation { sell_token: 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48, buy_token: 0x3b73c1b2ea59835cbfcadade5462b6ab630d9890, receiver: Some(0xef42253884aac0a5df6efdc5fd04ac3cb1f485fd), sell_amount: 24935633384, buy_amount: 12573018808727190454846, valid_to: 1621370180, app_data: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1], fee_amount: 64366616, kind: Sell, partially_fillable: false, signature: Signature { r: 0x677eef4df33c0c02669014d35a561754749b1869e04ac11672835d6feaf438e5, s: 0x2ddef6c89f8e388e1dc954051a0717ea2bcbda5d1273cfdea86ad0222499d68a, v: 28 }, signing_scheme: Eip712 } }, sell_token_index: 1, buy_token_index: 0, executed_amount: 24935633384 }], execution_plan: [Swap { from_token: Token { address: 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 }, to_token: Token { address: 0x3b73c1b2ea59835cbfcadade5462b6ab630d9890 }, from_token_amount: 24935633384, to_token_amount: 13252294880261259029821, protocols: [[[Protocol { name: "PMM3", part: 100.0, from_token_address: 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48, to_token_address: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 }], [Protocol { name: "UNISWAP_V2", part: 100.0, from_token_address: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, to_token_address: 0x3b73c1b2ea59835cbfcadade5462b6ab630d9890 }]]], tx: Transaction { from: 0x3328f5f2cecaf00a2443082b657cedeaf70bfaef, to: 0x11111112542d85b3ef69ae05771c2dccff4faa26, data: 0x7c025200000000000000000000000000db38ae75c5f44276803345f7f02e95a0aeef594400000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000180000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb480000000000000000000000003b73c1b2ea59835cbfcadade5462b6ab630d9890000000000000000000000000db38ae75c5f44276803345f7f02e95a0aeef59440000000000000000000000003328f5f2cecaf00a2443082b657cedeaf70bfaef00000000000000000000000000000000000000000000000000000005ce4791e80000000000000000000000000000000000000000000002cdb09faff33f7b2e2700000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000011800000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000700000000000000000000000000000000000000000000000000000000000000e0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000005e0000000000000000000000000000000000000000000000000000000000000082000000000000000000000000000000000000000000000000000000000000009800000000000000000000000000000000000000000000000000000000000000ca00000000000000000000000000000000000000000000000000000000000000ee080000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000064eb5625d9000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb4800000000000000000000000095e6f48254609a6ee006f7d493c8e5fb97094cef00000000000000000000000000000000000000000000000000000005ce4791e800000000000000000000000000000000000000000000000000000000800000000000000000000000080bf510fcbf18b91105470639e95610229377120000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000324b4be83d5000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000005ce4791e800000000000000000000000000000000000000000000000000000000000002a0000000000000000000000000a5d07e978398eb1715056d3ca5cb31035c02fdad000000000000000000000000db38ae75c5f44276803345f7f02e95a0aeef5944000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000006499c21c26df1c0000000000000000000000000000000000000000000000000000000005ce4791e8000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000060a425156ac2df4963e0d93cd09f2fb3a50b892ada41eb663c1fa549259e2bf0d91cc85e000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000001e00000000000000000000000000000000000000000000000000000000000000024f47261b0000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000024f47261b0000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb480000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000421bf06ee8abd58db5a6ca81afe5452dff299b547d5cc07f411738c9908493f4f42e1e17fc8468ef615514dc7176bcab0c74e2d18db68ba2673603f9cf376841f8f0030000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000184b3af37c000000000000000000000000000000000000000000000000000000000000000808000000000000000000000000000000000000000000000000000000000000024000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc20000000000000000000000000000000300000000000000000000000000000003000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc20000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000044a9059cbb0000000000000000000000008ec39063fab1fc22c05df2f00d999422a30f2c8300000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000a4c9f12e9d0000000000000000000000008ec39063fab1fc22c05df2f00d999422a30f2c83000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc20000000000000000000000003b73c1b2ea59835cbfcadade5462b6ab630d9890000000000000000000000003db38ae75c5f44276803345f7f02e95a0aeef5944000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000002647f8fe7a000000000000000000000000000000000000000000000000000000000000000808000000000000000000000000000000000000000000000000000000000000044000000000000000000000000db38ae75c5f44276803345f7f02e95a0aeef594400000000000000000000000000000000000000000000000000000000000001e0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000a4059712240000000000000000000000003b73c1b2ea59835cbfcadade5462b6ab630d9890000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000100000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004470bdb9470000000000000000000000003b73c1b2ea59835cbfcadade5462b6ab630d98900000000000000000000000000000000000000000000002ce68894bee1c2e6d3d000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000184b3af37c00000000000000000000000000000000000000000000000000000000000000080a000000000000000000000000000000000000000000000000000000000000024000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb480000000000000000000000000000000100000000000000000000000000000001000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb480000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000044a9059cbb000000000000000000000000910bf2d50fa5e014fd06666f456182d4ab7c8bd200000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000001a4b3af37c0000000000000000000000000000000000000000000000000000000000000008080000000000000000000000000000000000000000000000000000000000000440000000000000000000000003b73c1b2ea59835cbfcadade5462b6ab630d9890000000000000000000000000000000010000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000064d1660f990000000000000000000000003b73c1b2ea59835cbfcadade5462b6ab630d98900000000000000000000000003328f5f2cecaf00a2443082b657cedeaf70bfaef00000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000, value: 0, gas_price: 103100000000, gas: 0 } }], unwraps: [] } }

2021-05-18T20:32:04.439Z  WARN solver::driver: Estimated Prices: {0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48: Ratio { numer: 68104914525020295325337, denom: 233820793162468 }, 0x3b73c1b2ea59835cbfcadade5462b6ab630d9890: Ratio { numer: 280245401710528829697, denom: 526743251298737803230805 }, 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee: Ratio { numer: 1, denom: 1 }, 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2: Ratio { numer: 1, denom: 1 }}
</details>

Here is how this trade got matched in prod: https://gnosis-protocol.io/orders/0x0fd47ec904285aec4ef1b203a0534b77a0bca5de68cc2985c34717b5f0c1e855ef42253884aac0a5df6efdc5fd04ac3cb1f485fd60a42544

solver/src/driver/solver_settlements.rs
```diff
@@ -79,11 +79,19 @@ pub struct RatedSettlement {
...
+        let cost = gas_estimate * gas_price_normalized;
+        surplus - cost
+    }
+
+    // gas_price_normalized is the price of gas normalized to the found price vector.
+    pub fn objective_value(&self, gas_price_normalized: &BigRational) -> BigRational {
         let gas_estimate = self.gas_estimate.to_big_rational();
-        let cost = gas_estimate * gas_price;
-        self.surplus.clone() - cost
+        Self::compute_objective_value(&self.surplus, &gas_estimate, gas_price_normalized)
```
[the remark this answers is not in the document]
marcovc: > giving the function on the settlement struct (which currently computes the surplus) also access to the gas estimate

It already has access to the gas_estimate. Do you mean the gas_price_normalized?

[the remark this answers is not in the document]
fleupold: I think it might be nice to unit test the objective function via some thing that takes:
1. A set of computed prices and trades (e.g. a settlement)
2. The pre-auction prices (for normalization)
3. The current gas price in wei
4. The amount of gas needed for the transaction

And then do some assertions on that holistic computation (what is the objective value given certain individual utilities), at what gas price does a "bad but cheap" solution become higher ranked than another "good but expensive" solution, etc.

This way everything gets computed in one place rather than carrying around intermediate results (like surplus) on helper structs, which can easily lead to confusing the unit of account (and also using the wrong unit of account inside the unit tests).

fleupold:
Just to follow up here with my intuition on why the current version is correct:

Surplus is using traded amounts (which are in wei) and converts them into wei of ETH using the passed in exchange rates (which take decimals into account e.g. price of DAI is ~3e-4 and price of USDC is ~3e8, difference of 12 decimals). Thus the total surplus of a settlement is denominated in wei and therefore cost should also be denominated in wei (which it is at the moment, but not after this PR).

solver/src/driver/solver_settlements.rs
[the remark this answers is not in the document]
marcovc: Refactored below to address this. Still kept the helper function to allow to unit test just the final formula.

marcovc:
As Felix pointed out, things we're fine from the start. I've now added a test, some comments, and refactored a bit to make it more clear. I still kept the product of the gas price by the estimated price of the network token (1), so that things don't break if later we change this constant.

solver/src/driver.rs
```diff
@@ -263,8 +263,13 @@ impl Driver {
         &self,
         settlements: Vec<SettlementWithSolver>,
         prices: &HashMap<H160, BigRational>,
+        gas_price_wei: f64,
     ) -> Vec<RatedSettlement> {
         use futures::stream::StreamExt;
+
+        // TODO: 1e18 should be 10**(native_token_decimals)
+        let gas_price_wei = BigRational::from_float(gas_price_wei).unwrap();
```
bh2smith: Is there any way that this can fail? In my understanding of rust lang, we should avoid unwraps everywhere but tests or at least comment here on why we know that this will never result in an Error.

bh2smith (approved):
This PR has come a long way! Looks much simpler now.

solver/src/driver.rs
```diff
@@ -263,8 +263,13 @@ impl Driver {
         &self,
         settlements: Vec<SettlementWithSolver>,
         prices: &HashMap<H160, BigRational>,
+        gas_price_wei: f64,
     ) -> Vec<RatedSettlement> {
         use futures::stream::StreamExt;
+
+        let gas_price_wei = BigRational::from_float(gas_price_wei).unwrap()
+            * prices.get(&self.native_token).unwrap();
```
nlordell, line 271: nit: I agree with @bh2smith. If this can never fail, we should change these to `expect` (which is unwrap with an error message):
```suggestion
        let gas_price_wei = BigRational::from_float(gas_price_wei).expect("description of broken invariant")
            * prices.get(&self.native_token).description("description of broken invariant");
```

bh2smith: Somehow my comment is no longer attached to this line.

solver/src/driver.rs
```diff
@@ -274,11 +279,22 @@ impl Driver {
...
+                let solver_name = settlement.name;
+                let rated_settlement = RatedSettlement {
                     settlement,
                     surplus,
                     gas_estimate,
-                })
+                    gas_price: gas_price_wei.clone(),
+                };
+                tracing::info!(
+                    "Objective value for solver {} is {}: surplus={}, gas_estimate={}, gas_price_normalized={}",
```
fleupold, line 290: this shouldn't say `gas_price_normalized` should it?

solver/src/driver/solver_settlements.rs
```diff
@@ -76,14 +76,23 @@ pub struct RatedSettlement {
     pub settlement: SettlementWithSolver,
     pub surplus: BigRational,
     pub gas_estimate: U256,
+    pub gas_price: BigRational,
```
fleupold, line 79: Can we add comments on all three numerical fields what units these are?

solver/src/driver/solver_settlements.rs
```diff
@@ -76,14 +76,23 @@ pub struct RatedSettlement {
...
 impl RatedSettlement {
-    pub fn objective_value(&self, gas_price: f64) -> BigRational {
-        let gas_price = BigRational::from_float(gas_price).unwrap();
+    pub fn objective_value(&self) -> BigRational {
         let gas_estimate = self.gas_estimate.to_big_rational();
+        Self::compute_objective_value(&self.surplus, &gas_estimate, &self.gas_price)
+    }
+
+    // Helper function to allow unit testing objective value computation without a Settlement.
+    fn compute_objective_value(
```
fleupold, line 89: Does having this method separate still make sense? feels like it could be on RatedSettlement but no strong feelings if this makes testing it much easier. If you want to keep the method, it should probably become a free method.

solver/src/driver/solver_settlements.rs
```diff
@@ -257,4 +267,40 @@ mod tests {
         let settlements = vec![settlement.clone(), settlement];
         assert!(merge_at_most_settlements(1, settlements.into_iter()).is_none());
     }
+
+    #[test]
+    fn compute_objective_value() {
+        // Gas price is 2e-7 ETH
```
fleupold, line 273: Gas price is most commonly denoted in gwei (rather than in ETH)

solver/src/driver/solver_settlements.rs
```diff
@@ -257,4 +267,40 @@ mod tests {
...
+        let gas_price = BigRational::from_integer(200_000_000_000_u128.into());
+
+        // Surplus1 is 0.4 ETH
+        let surplus1 = BigRational::from_integer(400_000_000_000_000_000_u128.into());
+
+        // Surplus2 is 1.0 ETH
+        let surplus2 = BigRational::from_integer(1_000_000_000_000_000_000_u128.into());
+
+        // Gas estimate1 is 500_000
```
fleupold, line 282: This comment is not good (it's stating the obvious).

solver/src/driver/solver_settlements.rs
```diff
@@ -257,4 +267,40 @@ mod tests {
...
+        let gas_estimate1 = BigRational::from_integer(500_000.into());
+
+        // Gas estimate2 is 100_000
+        let gas_estimate2 = BigRational::from_integer(100_000.into());
+
+        // Objective value 1 is 4e-1 - 5e5 * 2e-7 = 4e-1 - 1e-1 = 0.3 ETH
+        let obj_value1 =
+            RatedSettlement::compute_objective_value(&surplus1, &gas_estimate1, &gas_price);
+
+        assert_eq!(
```
fleupold, line 292: Why are there two asserts here? What is the difference between the two scenarios? Could their difference be documented by using using two descriptively named unit tests. I'd also still like to see the scenario where we have two solutions. One being ranked higher at a certain low gas price, maybe them being equal at some gas price and then the second one ending up higher ranked at a sufficiently high gas price.

solver/src/driver.rs
fleupold: Maybe also add a comment why we do this multiplication (in case the price vector is not in the native token)

fleupold (commented):
Can you please add two more logs to the driver
1. Estimated Prices
2. Which solution ends up getting submitted

With this we should have been able to debug questions around the objective value just using logs.

Could you also please adjust the PR name, given that this is no longer fixing anything, just improving maintainability?

solver/src/driver/solver_settlements.rs
bh2smith: Now that I see the term RatedSettlement mentioned here, I might also suggested calling it `RankedSettlement`

solver/src/driver.rs
vkgnosis: Personally I don't like how the code formatting changes with rustfmt and longer `expect` messages so I prefer unwrap with a comment on top.  Also I don't think every unwrap/expect needs to have an explanation like when it is already commented earlier in the function or obvious. You always know where the panic comes from anyway because you get a line number for the unwrap too.
Not that I'm strongly against the expects, just highlighting my different experience with it.

solver/src/driver/solver_settlements.rs
marcovc: It allows testing "RatedSettlement.objective_value()" without having to build a "RatedSettlement" object, which requires passing in a settlement. This is used in the test I added. Note that this is a static method (doesn't have self) so essentially a free method namespaced on RatedSettlement. I'm happy to move it to outside the class if you like.

fleupold: You are right that the free vs static doesn't make a big difference (especially since it's private). I'd go with the general style of the repo which is using free methods over static methods when in doubt.

solver/src/driver.rs
nlordell: > Personally I don't like how the code formatting changes with rustfmt and longer expect messages

Yeah... its not pretty sometimes...

>  Also I don't think every unwrap/expect needs to have an explanation like when it is already commented earlier in the function or obvious.

I agree, if an `.unwrap` is asserting the same invariant as before, I don't think an expect is needed. In this case though, the `unwrap` could panic on a bad `gas_price_gwei` input to the function (for example, `NaN` for whatever reason).

> unwrap with a comment on top

I think this is fine as well. The reason I prefer `expect` over `unwrap + // comment` is between the two:
```
something.expect("should never fail");
----
// should never fail
something.unwrap();
```

The information is basically the same (i.e. the expect message is the comment), but in the former you actually display the information on `panic`.

This means that when I get an alert I will see:
```
panic: should never fail
```
instead of 
```
panic: unwrap `None` value
```

and then have to look up the file + line number to get the "should never fail" information. Additionally, this may or may not be easy given that prod and staging run different versions, which means you might have to also determine which tag to browse the code at.

That being said, since this shouldn't really panic, its not like we would have to look up the code to understand the invariant that is being broken often, and understanding exactly what is happening typically requires context from the code (which means you would have to look at it anyway). So I'm OK either way. The comment was mostly about seeing `unwrap`s without justification as to why this never fails.

solver/src/driver/solver_settlements.rs
nlordell: Personally I prefer free methods for these kinds of helper functions. For me associated functions (i.e. static functions) are typically for constructing instances of a type.

This is just personal preference. I imagine the compiler treats them in the exact same way.

fleupold (approved):
Could you please still add the two missing logs (prices, winning settlement) to get a full picture of the underlying computation and choice of the driver?

Otherwise, this PR looks good to merge!

marcovc:
I did:

https://github.com/gnosis/gp-v2-services/pull/605/files#diff-d2ca01ce07864051d8cdd4f8ddd0ebc2c4f11bca934ebb1fcb4cdb39c5dbc929R341

https://github.com/gnosis/gp-v2-services/pull/605/commits/7e65e1b7055019f1b8d9070f9702b0eb283ca8d8#diff-d2ca01ce07864051d8cdd4f8ddd0ebc2c4f11bca934ebb1fcb4cdb39c5dbc929R390

(but the last one got lost in the last merge, re-added it)

marcovc:
Thank you all for the valuable feedback as always!

Issue #603, opened 2021-05-17
Balancer Integration: WeightedPool Fetching


bh2smith:
After a few discussions about how best to fetch pools, given the options of 

- API calls to Google Big Query (for pools and reserves)
- Indexing Pool Created events and fetching reserves directly from the EVM

We have decided to go the route that does not rely on any third party service; namely, the latter.

To sketch out a plan for this, I envision that we will implement the following steps sequentially
- [x] Migrate reusable components of event handling for GPv2Settlement into shared crate - #606, #614, #613 and #616 
- [x] Setup Event Receiving (aka Streaming) For `BalancerV2Vault` contract (#641)
- [x] Implement Event Storing for Balancer events (#642).
- [x] `getNormalizedWeights` for each `WeightedPool`. (#696)
- [x] Write and Implement `WeightedPoolFetching` for `BalancerPoolFetcher` (#698)
- [ ] Read/Write Balancer Pool Registry to and from file (for more efficient cold start) - #723. This task is still undergoing debate as to the most appropriate design. We are planning to load Pool Registry from the balancer-v2 [subgraph](https://thegraph.com/explorer/subgraph/balancer-labs/balancer-v2), but also talk about being able to load from file when the graph is unavailable.
- [x] ~~Filtered Event Receiving by topic (cf issue #681)~~ no longer necessary.

Note that whenever the pools are actually needed by the services, we would have to query the reserves at the time of use (to have the most updated info) - but this is outside the scope of this task.

One thing that we may need to be careful of is that both solver and orderbook services will make use of this (one for price estimating and the other for solution finding) so we should take care to avoid data races when updating cached or dumped pool data.

nlordell:
Looks like a very good plan! One suggestion to reduce the scope a bit:

> I would suggest implementing a simple read/write WeightedPools to and from file

This is a very good suggestion for speeding up "warm starts". I think we should leave this for last though as I suspect indexing all `WeightedPool` addresses and static data (pool ID, weights, scales, tokens, etc.) will be quick enough for now that this won't be necessary (until the number of existing pools grows substantially). I would even suggest to keep that as a separate issue for now so we can decide how much it is needed/when it is needed.

fleupold:
Sounds great! I'm not sure how much code can be shared from the existing "event listening" part, maybe the EventStream and the "newBlockStream" can be abstracted behind a common interface to give us an automatically updating event stream (but that might be quite involved given we handle "reorgs" in the custom business logic).

[...]

PR #606, opened 2021-05-18
[Refactor] Generalize Event Handler for Storing and Retrieving

bh2smith:
As a first step towards pool fetching for BalancerV2 integration (Issue #603), we move the common components of our event handler for GPv2Settlement into the shared crate for reuse.

Introducing `EventHandler<C: EventRetrieving, S: EventStoring<C::Event>>` struct and traits 
- `EventStoring<T>`
- `EventRetrieving`

which are implemented for the `EventUpdater` used for GPv2Settlement

There has been some boiler plate code introduced (specifically implementing `EventRetrieving` for `GPv2Settlement` which we intend to remove via a macro)

This PR was written together with @nlordell - whose knowledge of associated types and generics has been enlightening.

### Test Plan
No logic changes, existing CI should pass.


shared/src/event_handling.rs
```diff
@@ -0,0 +1,170 @@
...
+const INSERT_EVENT_BATCH_SIZE: usize = 10_000;
+
+pub struct EventHandler<C: EventRetrieving, S: EventStoring<C::Event>> {
+    contract: C,
+    store: S,
+    last_handled_block: Option<u64>,
+}
+
+#[async_trait::async_trait]
+pub trait EventStoring<T> {
```
nlordell, line 22: Maybe some trait documentation might be helpful here (its obvious to me, but I'm biased because we had discussed it already :stuck_out_tongue:). An example of something that might not be obvious at first glance is the optional `range` parameter in `save_events`.

nlordell (approved):
Looks like there is a formatting issue in CI for some reason.

shared/src/event_handling.rs
```diff
@@ -0,0 +1,181 @@
...
+use ethcontract::contract::{AllEventsBuilder, ParseLog};
+use ethcontract::errors::ExecutionError;
+use ethcontract::{dyns::DynTransport, BlockNumber as Web3BlockNumber, Event as EthcontractEvent};
+use futures::{Stream, StreamExt, TryStreamExt};
+use std::ops::RangeInclusive;
+use web3::Web3;
+
+// We expect that there is never a reorg that changes more than the last n blocks.
+const MAX_REORG_BLOCK_COUNT: u64 = 25;
+// When we insert new trade events into the database we will insert at most this many in one
```
fleupold, line 11: nit: Maybe this comment should reference EventStoring now instead of a database.

shared/src/event_handling.rs
```diff
@@ -0,0 +1,181 @@
...
+/// # Examples
+/// Databases: might transform, filter and classify which events are inserted
+/// HashSet: For less persistent (in memory) storing, insert events into a set.
+#[async_trait::async_trait]
+pub trait EventStoring<T> {
+    /// Returns ok, on successful execution, otherwise an appropriate error
+    ///
+    /// # Arguments
+    /// * `events` the contract events to be saved by the implementer
+    /// * `range` optionally indicates a particular range of blocks on which to operate.
```
fleupold, line 33: I find the information that if range is set we should consider all previously announced events as invalid (basically replace) a bit hidden here. It's very implicit in "if the Options is some, you need to replace".

Maybe we can either put back two methods (`replace_events(events, range)` vs `append_events(events)`) or consider leaving the splitting into `INSERT_EVENT_BATCH_SIZE` up to the event storing information and only call `save_events` once (in which case we could always provide the range) and document that the range might overlap in part with a previously announced event range.

shared/src/event_handling.rs
```diff
@@ -0,0 +1,181 @@
...
+    pub fn new(contract: C, store: S, start_sync_at_block: Option<u64>) -> Self {
+        Self {
+            contract,
+            store,
+            last_handled_block: start_sync_at_block,
+        }
+    }
+
+    async fn event_block_range(&self) -> Result<RangeInclusive<BlockNumber>> {
+        let web3 = self.contract.web3();
```
fleupold, line 63: I think this was already a cheeky way of getting web3 in the previous implementation, but now we are leaking this hack through the public EventRetrieving interface (by requiring every implementation to implement a method that hands out web3). 

Instead, we could just give the EventHandler component its own web3 instance, this way the EventRetrieving interface is easier to implement in multiple places.

vkgnosis (approved):

shared/src/event_handling.rs
```diff
@@ -0,0 +1,181 @@
...
+                "current block number according to node is {} which is more than {} blocks in the \
+                 past compared to last handled block {}",
+                current_block, MAX_REORG_BLOCK_COUNT, last_handled_block
+            )
+        );
+        Ok(BlockNumber::Specific(from_block)..=BlockNumber::Latest(current_block))
+    }
+
+    /// Get new events from the contract and insert them into the database.
+    pub async fn update_events(&mut self) -> Result<()> {
```
fleupold, line 90: Maybe the logic on when to trigger this could also become shared functionality (like an UpdatingEventHandler). It probably makes sense to update events whenever we notice a new block. We already have `current_block_stream` which tells us whenever there is a new block and which I believe currently powers the run_maintenance logic.

fleupold (approved):
Really nice refactoring. I wonder if a `Stream` would be a more rusty way of handling these rather than the `save_events` callback, but this would be much more invasive.

vkgnosis:
The stream idea is interesting but like you said probably requires bigger changes so maybe something for the future.

shared/src/event_handling.rs
bh2smith: Absolutely agree. We had touched on this when refactoring and never came back.

shared/src/event_handling.rs
nlordell: This might make it nice to make the two modes of operation for `EventStoring` more explicit.

shared/src/event_handling.rs
nlordell: > public EventRetrieving interface (by requiring every implementation to implement a method that hands out web3).

Since this is intended to be implemented by contract bindings, I think this is fine, as contract bindings always have a `Web3` instance. Especially given @bh2smith's comment

> specifically implementing EventRetrieving for GPv2Settlement which we intend to remove via a macro

shared/src/event_handling.rs
nlordell: This is a cool idea. We could combine a `current_block_stream` with an `EventHandler` to make something that "just works". I would leave that for a follow up PR though.

shared/src/event_handling.rs
fleupold: To me accessing web3 this way is lying about dependencies. This component needs a web3 to do its logic (in fact it needs a currentBlockFetcher) and if e.g. we wanted to unit test it we could pass in a Fake `AllEventsBuilder` and assert some things about the events that get forwarded.

nlordell: Just to be clear, I also think that passing in a `web3` is the more correct/cleaner approach (or event better `CurrentBlockFetching` if that already exists).

The motivation for not passing it in was just that it was easier and all `EventRetrieving` implementations should be contract bindings and hence have access to a `web3` instance anyway.

shared/src/event_handling.rs
bh2smith: Yes, this is looking like a very nice change. Will update with this suggestion soon.

shared/src/event_handling.rs
bh2smith: If its alright with the reviewers, I will leave this as is for now with expectations to address in the following PR where we remove some of boiler plate code referred to here.

shared/src/event_handling.rs
bh2smith: Created an issue #613 for this point.

PR #607, opened 2021-05-18
Add cache for token detection

vkgnosis:
Edit: Don't review.

Not used yet.

### Test Plan
I did not feel that turning BadTokenDetector into a trait, mocking it, and then taking Box<dyn Trait> in the cache was worth it compared to how simple the logic it. It will still get tested through the e2e test once it is used. I don't midn doing the mocking if other people think it's worth it.

### DIFF
--- a/orderbook/src/bad_token_cache.rs
+++ b/orderbook/src/bad_token_cache.rs
@@ -1,0 +1,53 @@
+use crate::bad_token::{BadTokenDetector, TokenQuality};
+use anyhow::Result;
+use primitive_types::H160;
+use std::{collections::HashMap, sync::Mutex};
+
+pub struct BadTokenCache {
+    detector: BadTokenDetector,
+    // Std mutex is fine because we don't hold lock across await.
+    cache: Mutex<HashMap<H160, TokenQuality>>,
+    // Explicitly allowed tokens that are always treated as good.
+    allow_list: Vec<H160>,
+}
+
+impl BadTokenCache {
+    pub fn new(detector: BadTokenDetector, allow_list: Vec<H160>) -> Self {
+        Self {
+            detector,
+            cache: Default::default(),
+            allow_list,
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


orderbook/src/bad_token_cache.rs
```diff
@@ -0,0 +1,53 @@
...
+impl BadTokenCache {
+    pub fn new(detector: BadTokenDetector, allow_list: Vec<H160>) -> Self {
+        Self {
+            detector,
+            cache: Default::default(),
+            allow_list,
+        }
+    }
+
+    pub async fn is_good(&self, token: H160) -> Result<bool> {
```
vkgnosis, line 23: I decided to keep this a simple bool for now until I know we need the fool quality enum.

vkgnosis:
While continuing this I had a better to implement it so will make new PR.

orderbook/src/bad_token_cache.rs
```diff
@@ -0,0 +1,53 @@
+use crate::bad_token::{BadTokenDetector, TokenQuality};
+use anyhow::Result;
+use primitive_types::H160;
+use std::{collections::HashMap, sync::Mutex};
+
+pub struct BadTokenCache {
+    detector: BadTokenDetector,
+    // Std mutex is fine because we don't hold lock across await.
+    cache: Mutex<HashMap<H160, TokenQuality>>,
```
nlordell, line 9: A std Mutex is fine here, but why not use a `tokio::sync::Mutex` anyway? I would imagine it performs much better.

orderbook/src/bad_token_cache.rs
nlordell: I think that's totally fine. Its super easy to make it return a `TokenQuality` if its needed in the future.

nlordell (approved):
Looks good. Just a question about the choice of mutex (but not blocking the merge from my point of view).

PR #608, opened 2021-05-18
Log matched but unsettled order UIDs in current run loop

fleupold:
We have gotten increased user complaints that the time from order submission to settlement is taking too long. It's likely that we could speed up individual settlements if instead of merging them (risk of single one causing failure for other orders) we submitted them in parallel.

Before we implement this I'd like to get an idea how many orders are currently "delayed" a run loop.

### Test Plan
CI

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -258,12 +258,12 @@ impl Driver {
         }
     }
 
-    // Log the amount of orders that were matched but not settled in this runloop (effectively queued for the next one)
+    // Record metric with the amount of orders that were matched but not settled in this runloop (effectively queued for the next one)
     // Should help us to identify how much we can save by parallelizing execution.
-    fn report_matched_but_unsettled_orders<I: Iterator<Item = Settlement>>(
+    fn report_matched_but_unsettled_orders(
         &self,
         submitted: &Settlement,
-        all: I,
+        all: impl Iterator<Item = Settlement>,
     ) {
         let submitted: HashSet<_> = submitted
             .trades()
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -258,6 +258,27 @@ impl Driver {
         }
     }
 
+    // Log the amount of orders that were matched but not settled in this runloop (effectively queued for the next one)
+    // Should help us to identify how much we can save by parallelizing execution.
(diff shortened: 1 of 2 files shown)


solver/src/driver.rs
```diff
@@ -258,6 +258,27 @@ impl Driver {
         }
     }
 
+    // Log the amount of orders that were matched but not settled in this runloop (effectively queued for the next one)
+    // Should help us to identify how much we can save by parallelizing execution.
+    fn report_matched_but_unsettled_orders<I: Iterator<Item = Settlement>>(
+        &self,
+        submitted: &Settlement,
+        all: I,
+    ) {
```
vkgnosis, line 267:
```suggestion
    fn report_matched_but_unsettled_orders(
        &self,
        submitted: &Settlement,
        all: impl Iterator<Item = Settlement>,
    ) {
```

vkgnosis (approved):

vkgnosis (approved):

solver/src/driver.rs
```diff
@@ -258,6 +258,27 @@ impl Driver {
...
+            .iter()
+            .map(|trade| trade.order.order_meta_data.uid)
+            .collect();
+        let all_matched: HashSet<_> = all
+            .flat_map(|solution| solution.trades().to_vec())
+            .map(|trade| trade.order.order_meta_data.uid)
+            .collect();
+        let matched_but_not_settled: HashSet<_> = all_matched.difference(&submitted).collect();
+        self.metrics
+            .orders_matched_but_not_settled(matched_but_not_settled.len())
```
bh2smith, line 279: The doc string on this method says that we log these values, but the implementation only seems to include it in the metrics.... Do the metrics have a log statement in them?

bh2smith (approved):
All good, just wondering where the actual logging is taking place?

solver/src/driver.rs
vkgnosis: https://doc.rust-lang.org/stable/book/ch10-02-traits.html#traits-as-parameters

It means the same thing but is easier to write.

solver/src/driver.rs
fleupold: By logging I mean report to a prometheus metrics, which grafana collects from our pods periodically. Will update the documentation to clarify this.

nlordell (approved):

PR #609, opened 2021-05-18
Refactor bad token detector into trait and module

vkgnosis:
I plan on using the trait to implement wrapping detectors that can add a
cache and explicit allow and deny lists.

I dropped the gas price from TokenQuality because I am not sure yet how it should work with the allow and deny lists. Can add this back later.

### Test Plan
not needed

### DIFF
--- a/orderbook/src/bad_token/mod.rs
+++ b/orderbook/src/bad_token/mod.rs
@@ -1,0 +1,30 @@
+pub mod trace_call;
+
+use anyhow::Result;
+use primitive_types::H160;
+
+/// How well behaved a token is.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum TokenQuality {
+    Good,
+    Bad { reason: String },
+}
+
+impl TokenQuality {
+    pub fn is_good(&self) -> bool {
+        matches!(self, Self::Good { .. })
+    }
+
+    pub fn bad(reason: impl ToString) -> Self {
+        Self::Bad {
+            reason: reason.to_string(),
+        }
+    }
+}
+
+/// Detect how well behaved a token is.
+#[cfg_attr(test, mockall::automock)]
+#[async_trait::async_trait]
+pub trait BadTokenDetecting: Send + Sync {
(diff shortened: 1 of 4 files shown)


orderbook/src/bad_token/trace_call.rs
```diff
@@ -11,50 +12,42 @@ use ethcontract::{
...
 /// the amm pair providers.
 /// Tokens are bad if:
 /// - we cannot find an amm pool of the token to one of the base tokens
 /// - transfer into the settlement contract or back out fails
 /// - a transfer loses total balance
-pub struct BadTokenDetector {
+pub struct TraceCallDetector {
     pub web3: Web3,
-    pub pools: Vec<Box<dyn AmmPairProvider>>,
+    pub pools: Vec<Arc<dyn AmmPairProvider>>,
```
vkgnosis, line 30: Not needed in this PR but snuck in from another.

orderbook/src/bad_token/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+use primitive_types::H160;
+
+/// How well behaved a token is.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum TokenQuality {
+    Good,
+    Bad { reason: String },
+}
+
+impl TokenQuality {
```
bh2smith, line 13: This is cute! Excited to see it in action!

bh2smith (approved):
Looks excellent. I only wonder if we can expect the logs for bad token quality to come along with the token address as well. I presume this to be true, but don't really see a sign of it here.

vkgnosis:
> Looks excellent. I only wonder if we can expect the logs for bad token quality to come along with the token address as well. I presume this to be true, but don't really see a sign of it here.

Not sure exactly what you mean. This is just some infrastructure for detecting bad tokens. The code that calls it gives it a token address to check so it already knows what the address is and it can do logging how it wants to.

fleupold (approved):
I like this trait and the offered implementations a lot. All of them are nice little unit testable components that are easy to understand in itself.

PR #611, opened 2021-05-18
Add list based bad token detector

vkgnosis:
The purpose of this is to allow us to override the automatic detection
or use it if the node dosen't have the trace api.

### Test Plan
new tests

### DIFF
--- a/orderbook/src/bad_token/list_based.rs
+++ b/orderbook/src/bad_token/list_based.rs
@@ -1,0 +1,116 @@
+use super::{BadTokenDetecting, TokenQuality};
+use anyhow::Result;
+use primitive_types::H160;
+
+/// If a token is neither in the allow nor the deny list treat it this way.
+pub enum UnknownTokenStrategy {
+    Allow,
+    Deny,
+    Forward(Box<dyn BadTokenDetecting>),
+}
+
+/// Classify tokens with explicit allow and deny lists.
+pub struct ListBasedDetector {
+    pub allow_list: Vec<H160>,
+    pub deny_list: Vec<H160>,
+    pub strategy: UnknownTokenStrategy,
+}
+
+#[async_trait::async_trait]
+impl BadTokenDetecting for ListBasedDetector {
+    async fn detect(&self, token: ethcontract::H160) -> Result<TokenQuality> {
+        if self.allow_list.contains(&token) {
+            return Ok(TokenQuality::Good);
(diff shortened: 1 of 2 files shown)


orderbook/src/bad_token/list_based.rs
```diff
@@ -0,0 +1,116 @@
...
+    pub allow_list: Vec<H160>,
+    pub deny_list: Vec<H160>,
+    pub strategy: UnknownTokenStrategy,
+}
+
+#[async_trait::async_trait]
+impl BadTokenDetecting for ListBasedDetector {
+    async fn detect(&self, token: ethcontract::H160) -> Result<TokenQuality> {
+        if self.allow_list.contains(&token) {
+            return Ok(TokenQuality::Good);
```
bh2smith, line 23: I guess we have to be careful not to include any actually bad tokens into the allow list since they essentially default to good.

fleupold (approved):
Very nice and useful component!

bh2smith (approved):
Looks good, I just wonder if we could document the detect method a bit here. From my understanding; 

- check if its in allow list -> if so then Good
- check if its in deny list -> if so then Bad
- compare it with our strategy and act accordingly. 

I guess the strategy is something configured at runtime, so we deterministically know how to treat it. Now that I've written it out, it seems clear to me. Does this mean `Forward` will be our default?

fleupold:
> Looks good, I just wonder if we could document the detect method a bit here.

Personally I feel this is well documented by the unit tests. @taminomara mentioned to me that they are a big fan of doc-tests. Maybe there is a way to combine the two here?

taminomara:
Yeah, `cargo test` will compile and run any code snippet from any docstring, so you could put some simple snippets there. Although for some reason doctests don't work nicely with mocked `BadTokenDetecting`, not sure what's up with that.

vkgnosis:
Probably because the mocks are only generated if the `test` attr is set and it isn't set by doc tests since they are meant to be used as documentation for library users so it shouldnt have access to test exclusive stuff.

taminomara:
Yeah, that was my thought. I usually add things like `DenyAllDetector`, `AllowAllDetector` and `FailDetector`, it's good to have them around even if they're not used by the main codebase. Then we can add convenience methods like `ListBasedDetector::from_allow_list` and `ListBasedDetector::from_deny_list`.

PR #616, opened 2021-05-19
[Refactor] Introducing Service Maintenance

### LINKED ISSUE #613: Refactor EventHandler to recognize when to update
As suggested in https://github.com/gnosis/gp-v2-services/pull/606#discussion_r634382017, it would make sense for the `EventHandler` to manage its own update schedule (i.e. detect when there is a new block).  This should amount to combining a `current_block_stream` with an `EventHandler` which is currently handled in 

http://github.com/gnosis/gp-v2-services/blob/9939d8d31f336a7b0d517ce7ebeba7b79ffc7917/orderbook/src/orderbook.rs#L200-L203

bh2smith:
Closes #613 and is a step toward the first stage of #603.

This change introduces `Maintaining` trait with method `run_maintenance`. Given the current services requiring maintenance, we implement `Maintaining` for `Orderbook`, `Database` and `EventUpdater` and collect them into a cool new `ServiceMaintenance` struct. 

This decouples the `current_block_stream` from the orderbook maintenance in a way that can be reusable elsewhere.

These changes were made together with @nlordell.

### Test Plan
No new tests. Existing CI should suffice.


orderbook/src/orderbook.rs
```diff
@@ -48,7 +45,7 @@ pub enum OrderCancellationResult {
 pub struct Orderbook {
     domain_separator: DomainSeparator,
     database: Database,
-    event_updater: Mutex<EventUpdater>,
+    // event_updater: Mutex<EventUpdater>,
```
nlordell, line 48: nit: these comments can probably be cleaned up.

orderbook/src/orderbook.rs
```diff
@@ -59,7 +56,7 @@ impl Orderbook {
     pub fn new(
         domain_separator: DomainSeparator,
         database: Database,
-        event_updater: EventUpdater,
+        // event_updater: EventUpdater,
```
nlordell, line 59: :point_up:

orderbook/src/orderbook.rs
```diff
@@ -68,7 +65,7 @@ impl Orderbook {
         Self {
             domain_separator,
             database,
-            event_updater: Mutex::new(event_updater),
+            // event_updater: Mutex::new(event_updater),
```
nlordell, line 68: :point_up:

orderbook/src/maintenance.rs
```diff
@@ -0,0 +1,34 @@
+use crate::database::Database;
+use crate::event_updater::EventUpdater;
+use crate::orderbook::Orderbook;
+use anyhow::Result;
+use futures::try_join;
+use shared::current_block::Maintaining;
+use std::sync::Arc;
+
+pub struct ServiceMaintenance {
```
nlordell, line 9: Might be worth adding a comment blurb here (or for the module) explaining that this structure collects all service components that need to run maintenance on every new block.

orderbook/src/maintenance.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+#[async_trait::async_trait]
+impl Maintaining for ServiceMaintenance {
+    async fn run_maintenance(&self) -> Result<()> {
+        try_join!(
+            self.storage.run_maintenance(),
+            self.event_updater.run_maintenance()
+        )?;
+        self.database.run_maintenance().await
```
nlordell, line 32: Just for the other reviewers, we intentionally only make the storage and event updater run concurrently followed serially by the database maintenance as that was how the original `orderbook_maintenance` was run. I don't know if its actually necessary to run this way, or if they all can run concurrently.

Note that both the `event_updater` and the `database` maintenance can cause changes to the Prostgres database.

nlordell (approved):

orderbook/src/maintenance.rs
bh2smith: I could add a comment that this is why we have kept them separate here.

nlordell: It might also be good to know if its still necessary.

bh2smith: While it is probably safer to keep these from being run in parallel, the event updater primarily targets (but it not limited to) writes to the trade table, it also updates existing orders (on `OrderInvalidation`). database maintenance is mostly concerned with removing expired fee quotes.

It is possible that future iterations could change either of these places and a project maintainer may not think to look here, so its probably safer to just keep them apart with a comment.

vkgnosis: They can run at the same time but also it doesn't really matter if they for maintenance that happens every 10 seconds. So I would remove the comment saying that they shouldn't but also it's fine to keep them sequential here.

vkgnosis: One disadvantage of try_join is that

> try_join! is similar to join!, but completes immediately if any of the futures return an error.

So if one of them fails it will cancel the others. Imo it would be better if they all get a chance to run. This is what the previous code did. It just logged the errors but ran all of them.

bh2smith: Cool, so I will remove the comment and use `join!` instead.

bh2smith: I have one question here about what might be best for this.

We need to resolve a Tuple `(Result<(), anyhow::Error>, Result<(), anyhow::Error>)`
because of unused results. In the previous implementation we had just gone with:

```rust
        join!(
            self.storage.run_maintenance(),
            self.event_updater.run_maintenance()
        ).0?;
```

but this is giving preference to the first result and seems like cheating... should I assign this like `res = join!(one, two)` and then do 
```
res.0?;
res.1?;
``` 
in order to be unbiased?

nlordell: In that case, @vkgnosis do you think we should just `join` on all 3 futures, and then print out any of the errors that we encountered?. Something like:
```rust
for result of future::join_all(vec![
            self.storage.run_maintenance(),
            self.event_updater.run_maintenance(),
            self.database.run_maintenance(),
]) {
  result?;
}
```

bh2smith: Right... I just had to make it a vector and use join_all!

Seems good!

nlordell: > but this is giving preference to the first result and seems like cheating...

Remember that this used to do `balance_fetcher.update().await` which can't fail (see `impl Maintaining for Orderbook`) which is why it worked like that before. I think we can ether:

```rust
let result = join!(
    self.storage.run_maintenance(),
    self.event_updater.run_maintenance(),
    self.database.run_maintenance(),
);
result.0?;
result.1?;
result.2?;
```

Or my suggestion above.

vkgnosis: I would say each individual task should be run and if it errors, the error logged. Then you could always return Ok from the bigger function.
Something like join_all should be fine.

orderbook/src/maintenance.rs
```diff
@@ -0,0 +1,37 @@
...
+
+#[async_trait::async_trait]
+impl Maintaining for ServiceMaintenance {
+    async fn run_maintenance(&self) -> Result<()> {
+        join_all(vec![
+            self.storage.run_maintenance(),
+            self.event_updater.run_maintenance(),
+            self.database.run_maintenance(),
+        ])
+        .await;
```
vkgnosis, line 34: Still need to `tracing::error!` log the errors.

nlordell: Also, this doesn't correctly propagate any errors.

vkgnosis: I think if we log them it's fine to not propagate because in that sense the method is always successful. Like before the global maintenance task never returns, it just logs errors and keeps going.

bh2smith: I'm almost there... just left with an "unused map"

nlordell: > fine to not propagate because in that sense the method is always successful

Great point. I think in that case we should probably remove the error trace from `CurrentBlackStream::run_maintenance_on_new_block()`.

> just left with an "unused map"

You should probably use a `for ... in` loop here for a couple reasons:
- `map` is lazy, so you actually will probably not run the closure at all here
- `map` is for converting one type to another, not really for executing some code on each element in the iterator. There is [`inspect`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.inspect) which is a little more appropriate for this kind of thing, but still not quite right here because it is also lazy (so you would have to iterate it somehow, either with `collect` or `for ... in`).

I think the easiest approach is really a for loop here:
```rust
for join_all(vec![
        self.storage.run_maintenance(),
        self.event_updater.run_maintenance(),
        self.database.run_maintenance(),
    ])
    .await
{
    if let Err(err) = result {
        tracing::error!("maintenance failed with: {}", err);
    }
}
```

bh2smith: So I have made the suggested change here, but my only worry now is the unwrap I had to use in `CurrentBlackStream::run_maintenance_on_new_block`. Is this ok? Should it contain some comment that the errors are handled inside maintainer?

nlordell: I don't think we should `unwrap`. Especially if we plan on using this with other `Maintaining` implementations (like the Balancer Vault event indexer in the solver).

Maybe my comment:

> I think in that case we should probably remove the error trace from CurrentBlackStream::run_maintenance_on_new_block().

Was wrong.

nlordell: > However I can imagine we might not always return Ok in other places.

Yeah, I think that is the catch. In the solver for example, we might just run maintenance on the Vault event updater, which in theory could error.

nlordell: One other approach would be to aggregate all errors into one so it gets logged by the caller.

orderbook/src/maintenance.rs
```diff
@@ -0,0 +1,42 @@
...
+
+/// Collects all service components requiring maintenance on each new block
+pub struct ServiceMaintenance {
+    storage: Arc<Orderbook>,
+    database: Database,
+    event_updater: EventUpdater,
+}
+
+impl ServiceMaintenance {
+    pub fn new(storage: Arc<Orderbook>, database: Database, event_updater: EventUpdater) -> Self {
```
fleupold, line 17: Should this be agnostic to the number and kind of components it takes and just work on `impl Iterator<Item = Arc<dyn Maintaining>>`?

vkgnosis (approved):

orderbook/src/main.rs
```diff
@@ -222,12 +200,8 @@ async fn main() {
...
         metrics.clone(),
     );
-    let maintenance_task = task::spawn(orderbook_maintenance(
-        orderbook,
-        database.clone(),
-        settlement_contract,
-        current_block_stream,
-    ));
+    let maintenance_task =
+        task::spawn(current_block_stream.run_maintenance_on_new_block(service_maintainer));
```
fleupold, line 204: I'd prefer if instead of putting logic about maintenance into the current_block_stream component, we would instead pass the current_block_stream into the ServiceMaintenance maintenance component and do the stream processing there.

The ServiceMaintenance already know about the maintenance logic and I'd find it more natural to say this component work on current block stream than having the current block stream also know how to drive a maintenance component.

fleupold (approved):

orderbook/src/main.rs
nlordell: Good suggestion.

orderbook/src/maintenance.rs
nlordell: The original motivation was that we weren't sure whether or not there were ordering requirements (see discussion here: https://github.com/gnosis/gp-v2-services/pull/616#discussion_r635212228). Now that we determined that there aren't any - I agree that act on a collection of `Arc<dyn Maintaining>`s.

We could then even move this into the `shared` crate.

bh2smith: Will do. Although we may find in the future that they can't necessarily all be run in parallel.

bh2smith: With this `impl Iterator` I get the warning that ``impl Trait` not allowed outside of function and inherent method return types [E0562]`. I suspect I will have to go with `Vec<Arc<dyn Maintaining>>` if thats alright with everyone.

fleupold: I believe at least on the struct you need to chose a concrete container type (or make the struct templated, but that's overkill). But Vec everywhere works as well.

PR #618, opened 2021-05-19
Use BadTokenDetecting trait instead of manual deny list

vkgnosis:
We still have the existing deny list but it is now implemented through
the ListBasedDetector and BadTokenDetecting instead of passing the deny
list around manually.
A few changes follwo because a Result had to be added to some of the
functions.
This PR does not yet enable the new tracing based bad token detection.

### Test Plan
existing tests


orderbook/src/main.rs
```diff
@@ -160,34 +160,41 @@ async fn main() {
...
+            .iter()
+            .all(|token| !unsupported_tokens.contains(token)),
+        "same token can't be both allowed and denied"
     );
 
-    let current_block_stream = current_block_stream(web3.clone()).await.unwrap();
+    let bad_token_detector = Arc::new(ListBasedDetector {
+        allow_list: allowed_tokens,
+        deny_list: unsupported_tokens,
+        strategy: UnknownTokenStrategy::Allow,
```
bh2smith, line 182: Don't we wanna use the `UnknownTokenStrategy::Forward` by default?

shared/src/pool_fetching.rs
```diff
@@ -20,30 +21,38 @@ type RelativeReserves = (U256, U256, H160);
...
 #[async_trait::async_trait]
 impl PoolFetching for FilteredPoolFetcher {
-    async fn fetch(&self, token_pairs: HashSet<TokenPair>) -> Vec<Pool> {
-        let filtered_pairs = token_pairs
-            .into_iter()
-            .filter(|pair| !self.unsupported_tokens.iter().any(|t| pair.contains(t)))
-            .collect();
+    async fn fetch(&self, pairs: HashSet<TokenPair>) -> Result<Vec<Pool>> {
+        let mut filtered_pairs = HashSet::new();
+        'outer: for pair in pairs {
```
bh2smith, line 48: This is new for me `outer!

shared/src/pool_aggregating.rs
```diff
@@ -58,11 +59,11 @@ impl PoolAggregator {
...
 impl PoolFetching for PoolAggregator {
-    async fn fetch(&self, token_pairs: HashSet<TokenPair>) -> Vec<Pool> {
+    async fn fetch(&self, token_pairs: HashSet<TokenPair>) -> Result<Vec<Pool>> {
         let mut pools = vec![];
         for fetcher in self.pool_fetchers.iter() {
-            pools.extend(fetcher.fetch(token_pairs.clone()).await);
+            pools.extend(fetcher.fetch(token_pairs.clone()).await?);
         }
-        pools
+        Ok(pools)
```
bh2smith, line 67: It looks like we are just putting `Ok` on all the fetch implementations? Is there at least one place that we didn't do this? Just want to understand the reason for changing the fetch function signature.

bh2smith (approved):
Seems good, just a few questions.

orderbook/src/main.rs
vkgnosis: Forward means it has to forward unknown tokens to another BadTokenDetecting implementation. But this new doesn't exist in this PR yet. This to keep the deny-list behavior the same. I'll make another PR to add the new better bad token detector as the forward target.

shared/src/pool_fetching.rs
vkgnosis: Not sure if it's in the book but I found this documentation https://doc.rust-lang.org/rust-by-example/flow_control/loop/nested.html .

shared/src/pool_aggregating.rs
vkgnosis: Here in line 65 there is a sneaky `?` because the inner fetcher can fail, so the error path can happen. For example in the FilteredPoolFetcher there is `!self.bad_token_detector.detect(token).await?.is_good()`: it fails if the token detector fails.

orderbook/src/fee.rs
```diff
@@ -233,17 +246,9 @@ impl MinFeeCalculating for MinFeeCalculator {
...
-        }
-
-        if buy_token
-            .map(|t| self.unsupported_tokens.contains(&t))
-            .unwrap_or_default()
-        {
-            return Err(MinFeeCalculationError::UnsupportedToken(
-                buy_token.expect("Must exist"),
-            ));
+        self.ensure_token_supported(sell_token).await?;
```
fleupold, line 249: noice

orderbook/src/orderbook.rs
```diff
@@ -51,8 +51,8 @@ pub struct Orderbook {
     event_updater: Mutex<EventUpdater>,
     balance_fetcher: Box<dyn BalanceFetching>,
     fee_validator: Arc<EthAwareMinFeeCalculator>,
-    unsupported_tokens: HashSet<H160>,
     min_order_validity_period: Duration,
+    bad_token: Arc<dyn BadTokenDetecting>,
```
fleupold, line 55: naming nit: calling this bad_token sounds like this is a single H160. Maybe instead call it bad_token_detector or at least bad_tokens?

orderbook/src/orderbook.rs
```diff
@@ -203,6 +206,23 @@ impl Orderbook {
...
 }
 
+async fn filter_unsupported_tokens(
+    mut orders: Vec<Order>,
+    bad_token: &dyn BadTokenDetecting,
+) -> Result<Vec<Order>> {
+    let mut index = 0;
+    'outer: while index < orders.len() {
+        for token in orders[index].order_creation.token_pair().unwrap() {
+            if !bad_token.detect(token).await?.is_good() {
```
fleupold, line 216: The fancy double loop is needed because detect is async and so using things like iter.any doesn't work? Maybe add a comment or consideration if this should run in parallel in the future?

shared/src/price_estimate.rs
```diff
@@ -123,11 +126,10 @@ impl PriceEstimating for BaselinePriceEstimator {
         amount: U256,
         kind: OrderKind,
     ) -> Result<BigRational, PriceEstimationError> {
-        if self.unsupported_tokens.contains(&sell_token) {
-            return Err(PriceEstimationError::UnsupportedToken(sell_token));
-        }
-        if self.unsupported_tokens.contains(&buy_token) {
-            return Err(PriceEstimationError::UnsupportedToken(buy_token));
+        for &token in [sell_token, buy_token].iter() {
```
fleupold, line 129: Why not use the same nice little ensure helper you used in the other component?

shared/src/pool_fetching.rs
```diff
@@ -20,30 +21,38 @@ type RelativeReserves = (U256, U256, H160);
 
 #[async_trait::async_trait]
 pub trait PoolFetching: Send + Sync {
-    async fn fetch(&self, token_pairs: HashSet<TokenPair>) -> Vec<Pool>;
+    async fn fetch(&self, token_pairs: HashSet<TokenPair>) -> Result<Vec<Pool>>;
```
fleupold, line 24: This change makes the use of this component more complicated. Do we actually still need the BadTokenDetector here if we have it in the other components?

shared/src/pool_fetching.rs
```diff
@@ -20,30 +21,38 @@ type RelativeReserves = (U256, U256, H160);
...
-            .filter(|pair| !self.unsupported_tokens.iter().any(|t| pair.contains(t)))
-            .collect();
+    async fn fetch(&self, pairs: HashSet<TokenPair>) -> Result<Vec<Pool>> {
+        let mut filtered_pairs = HashSet::new();
+        'outer: for pair in pairs {
+            for token in pair {
+                if !self.bad_token_detector.detect(token).await?.is_good() {
+                    continue 'outer;
+                }
+            }
```
fleupold, line 53: Maybe you can factor this check into a separate method (or token pair directly). I believe this would simplify your loop here and in the other place where you use `outer.

orderbook/src/orderbook.rs
```diff
@@ -411,4 +433,31 @@ mod tests {
         let orders_ = solvable_orders(orders.clone(), &balances);
         assert_eq!(orders_, orders[..1]);
     }
+
+    #[test]
+    fn filter_unsupported_tokens_() {
```
fleupold, line 438: nice unit test!

fleupold (commented):
Looks good overall. Quite a lot of changes, left some comments (not accepting yet mainly to discuss if we can get rid of the BadTokenDetector in the pool fetching logic)

orderbook/src/main.rs
```diff
@@ -160,34 +160,41 @@ async fn main() {
...
     assert!(
-        unsupported_tokens
-            .intersection(&base_tokens)
-            .collect::<HashSet<_>>()
-            .is_empty(),
-        "Base tokens include at least one unsupported token!"
+        allowed_tokens
+            .iter()
+            .all(|token| !unsupported_tokens.contains(token)),
+        "same token can't be both allowed and denied"
```
nlordell, line 176: nit: Maybe this could move into a `ListBasedDetector::new()` and make the fields private, thus making it illegal to create an instance with tokens that are both simultaneously allowed and denied?

shared/src/pool_fetching.rs
nlordell: One possible solution I see would be to make the `BadTokenDetecting` infallible. The implementation with `trace_callMany` could then give the token the benefit of the doubt on unrelated failures (connection issues for example).

Additionally, I would also be curious, as @fleupold asked, if we need `PoolFetching` to check the quality of the tokens. Wouldn't it make sense for the caller to be required only fetch pools for "good" tokens (i.e. the filtering happening "higher-up"). I understand that the pool fetcher previously filtered bad tokens, so removing it would be unrelated to this change. It could go in a follow up PR?

That being said, I'm OK with returning a `Result` here.

nlordell (approved):
Nice changes! I think this is a very good refactor.

vkgnosis:
I like the suggestion of making callers check for bad tokens. Will prepare another PR that does it that way and compare.

shared/src/pool_fetching.rs
vkgnosis: Fixed.

orderbook/src/main.rs
vkgnosis: added

PR #621, opened 2021-05-20
Remove FilteredPoolFetcher

vkgnosis:
The main component (PriceEstimator) using the pool fetcher already
checks for deny listed tokens so there is no need for it anymore.

The other use is for fetching uniswap liquidity in the driver in
solver/src/liquidity/uniswap.rs:82 which is fine too as we only consider
base_tokens as path candidates and the original orders have already been
filtered by the order book.

This came up in #618  . I think the historical reason for the filtered pool fetcher is that it used to be the only place where we filtered unsupported tokens and eventually we introduced the filtering directly into price estimator and fee calculation.

### Test Plan
adapted test

### DIFF
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -15,7 +15,7 @@ use prometheus::Registry;
 use shared::{
     current_block::{current_block_stream, CurrentBlockStream},
     pool_aggregating::{self, PoolAggregator},
-    pool_fetching::{CachedPoolFetcher, FilteredPoolFetcher},
+    pool_fetching::CachedPoolFetcher,
     price_estimate::BaselinePriceEstimator,
     transport::LoggingTransport,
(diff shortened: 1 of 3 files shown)


fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FxT0GqimU9dTwmE5lra%2Fgiphy.gif&token=red" data-gitmeme-token="red"></a>

nlordell (approved):
Awesome.

bh2smith (approved):
Looks and sounds great. I only wonder if we have to worry about the possibility of old orders which are not yet expired containing unsupported tokens.

vkgnosis:
get solvable orders filters our unsupported tokens so should be fine

PR #625, opened 2021-05-20
Enabled new bad token detection

vkgnosis:
and disable the tracing in account balances since it's no longer
necessary as it's weaker than this one. We'll still keep the call based
one.

### Test Plan
existing tests and manually tested with the markets route

### DIFF
--- a/shared/src/bad_token/trace_call.rs
+++ b/shared/src/bad_token/trace_call.rs
@@ -34,7 +34,9 @@ pub struct TraceCallDetector {
 #[async_trait::async_trait]
 impl BadTokenDetecting for TraceCallDetector {
     async fn detect(&self, token: H160) -> Result<TokenQuality> {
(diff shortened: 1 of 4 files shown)


shared/src/bad_token/trace_call.rs
```diff
@@ -34,7 +34,9 @@ pub struct TraceCallDetector {
 #[async_trait::async_trait]
 impl BadTokenDetecting for TraceCallDetector {
     async fn detect(&self, token: H160) -> Result<TokenQuality> {
-        self.detect_impl(token).await
+        let quality = self.detect_impl(token).await?;
+        tracing::info!("token {:?} quality {:?}", token, quality);
```
fleupold, line 38: good idea to log this!

fleupold (approved):
🏁

PR #626, opened 2021-05-20
Pass fixed block number into liquidity fetcher

### LINKED ISSUE #545: Follow up tasks for better settlement simulation error logging

fleupold:
Fixes #545

Some of the alerts regarding settlements that failed both on the Latest block as well as on the block that liquidity was fetched are incorrect, because we don't enforce that liquidity is actually fully fetched on a given block. It is possible that e.g. Uniswap pools are fetched on block t and Sushiswap pools on block t+1.

As we add more Baseline Liquidity sources the chance for seeing inconsistent state increases.

This PR makes it so that we can "fix" a BlockNumber into our liquidity fetcher to make sure at least the baseline solver is working only with state that comes from one coherent block. Technically we could still run into issues with reorgs, but I'm not sure there is something we can do about.

Theoretically we should also fetch the off-chain orderbook at a specific block (so that balance checks, etc are in line with the remaining path finding). We can add that if we see failures with regards to that.

### Test Plan
CI as this is mainly just propagating a new argument into all implementations of the PoolFetching trait.

### DIFF
--- a/shared/src/pool_fetching.rs
+++ b/shared/src/pool_fetching.rs
@@ -2,7 +2,7 @@ use std::collections::{HashMap, HashSet};
 
 use crate::{baseline_solver::BaselineSolvable, Web3};
 use contracts::{IUniswapLikePair, ERC20};
-use ethcontract::{batch::CallBatch, H160, U256};
+use ethcontract::{batch::CallBatch, BlockNumber, H160, U256};
 use model::TokenPair;
 use num::{rational::Ratio, BigInt, BigRational, Zero};
 use std::sync::Arc;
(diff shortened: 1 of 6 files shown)


shared/src/pool_fetching.rs
```diff
@@ -181,8 +181,13 @@ impl CachedPoolFetcher {
...
     }
 
-    async fn fetch_inner(&self, token_pairs: HashSet<TokenPair>) -> Vec<Pool> {
+    async fn fetch_inner(
+        &self,
+        token_pairs: HashSet<TokenPair>,
+        at_block: BlockNumber,
+    ) -> Vec<Pool> {
         let mut cache = self.cache.lock().await;
+        //TODO check that cache hit also matches at_block parameter
```
fleupold, line 190: https://github.com/gnosis/gp-v2-services/pull/627

vkgnosis (approved):

nlordell (approved):

bh2smith (approved):

PR #627, opened 2021-05-20
Make caching pool fetcher block aware

fleupold:
Following up un #626 the caching pool fetcher should now differentiate at what block its information was fetched (and not e.g. return the Latest state when queried for a past block).

This is achieved by allowing it to store cached pool information for multiple blocks (stored in an LRU cache).

### Test Plan
Added a unit test

### DIFF
--- a/shared/src/pool_fetching.rs
+++ b/shared/src/pool_fetching.rs
@@ -562,4 +562,53 @@ mod tests {
             vec![]
         );
     }
+
+    #[tokio::test]
+    async fn caching_pool_fetcher_invalidates_if_latest_block_reorgs() {
+        let token_a = H160::from_low_u64_be(1);
+        let token_b = H160::from_low_u64_be(2);
+        let pair = TokenPair::new(token_a, token_b).unwrap();
+
+        let pools = Arc::new(Mutex::new(vec![Pool::uniswap(pair, (1, 1))]));
+
+        let starting_block = CurrentBlock {
+            hash: Some(H256::from_low_u64_be(0)),
+            number: Some(0.into()),
+            ..Default::default()
+        };
+
+        let current_block = Arc::new(std::sync::Mutex::new(starting_block.clone()));
+
+        let (_, receiver) = watch::channel::<CurrentBlock>(Default::default());
+        let block_stream = CurrentBlockStream::new(receiver, current_block.clone());
+
+        let inner = Box::new(FakePoolFetcher(pools.clone()));
+        let instance = CachedPoolFetcher::new(inner, block_stream);
+
+        // Read Through
+        assert_eq!(
+            instance.fetch(hashset!(pair), BlockNumber::Latest).await,
+            vec![Pool::uniswap(pair, (1, 1))]
+        );
+
+        // simulate reorg on latest block
+        *current_block.lock().unwrap() = CurrentBlock {
+            hash: Some(H256::from_low_u64_be(1)),
+            number: starting_block.number,
+            ..Default::default()
+        };
+
+        // clear inner, to test we are not using cache
+        pools.lock().await.clear();
+        assert_eq!(
+            instance
+                .fetch(hashset!(pair), BlockNumber::Number(0.into()))
+                .await,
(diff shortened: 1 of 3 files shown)


vkgnosis (approved):
Only problem I see is that pending block won't work correctly for the same reason we need to check that latest block hasn't changed. I don't think we're using the pending block so this won't manifest now but could be a problem later.  An easy fix could be to assert the block isn't pending. Or we could say pending blocks are never cached since they can change at any time.

Another simplification could be to convert Latest into a u64 block number using `self.block_stream.current_block()` so that the cache only ever stores real block numbers.

nlordell (approved):
Nice!

shared/src/pool_fetching.rs
```diff
@@ -165,18 +166,49 @@ impl BaselineSolvable for Pool {
...
+    // we need to prune the cache entry for Block::Latest (which doesn't store any details)
+    latest_block: CurrentBlock,
+    pools: LruCache<Block, HashMap<TokenPair, Pool>>,
+}
+
+/// Helper type to work around the fact that ethcontract::BlockNumber is not Hashable
+#[derive(PartialEq, Eq, Hash)]
+enum Block {
+    Latest,
+    Pending,
```
nlordell, line 189: Maybe `Pending` shoulnd't be allowed in the cache. This way we always re-fetch for `pending` block?

fleupold:
> Another simplification could be to convert Latest into a u64 block number using self.block_stream.current_block() so that the cache only ever stores real block numbers.

That's a nice idea. This would have the nice side effect that pools that were cached for a `BlockNumber::Latest` request don't have to be re-queried if the same block is later queried via `BlockNumber::Number`.

nlordell:
How would this deal with re-orgs?

nlordell:
> if a reorg happens "behind" the latest block (or latest block is not re-queried before the next block is mined) we don't notice the reorg

Created an issue for this: https://github.com/gnosis/gp-v2-services/issues/631. I don't think its critical at the moment but it might cause some hard to debug issues in the future.

shared/src/pool_fetching.rs
```diff
@@ -165,20 +166,38 @@ impl BaselineSolvable for Pool {
...
     inner: Box<dyn PoolFetching>,
-    cache: Arc<Mutex<(Block, TokenPoolsMap)>>,
+    cache: Arc<Mutex<Cache>>,
     block_stream: CurrentBlockStream,
 }
 
+struct Cache {
+    /// Used to store details (e.g. hash) about the latest block. Needed so we know what `BlockNumber::Latest` refers to
+    latest_block: CurrentBlock,
+    pools: LruCache<u64, HashMap<TokenPair, Vec<Pool>>>,
```
bh2smith, line 181: This might also be good to use in `BalancerPools` cache that is currently under construction.

shared/src/pool_fetching.rs
```diff
@@ -189,30 +208,47 @@ impl CachedPoolFetcher {
...
             .cloned()
             .collect();
 
         let mut inner_results = self.inner.fetch(cache_misses, at_block).await;
         for miss in &inner_results {
-            cache.1.entry(miss.tokens).or_default().push(miss.clone());
+            cached_pools
+                .entry(miss.tokens)
+                .or_default()
+                .push(miss.clone());
```
bh2smith, line 237: if the entry is a vector would this be equivalent to 

```
cached_pools.entry(miss.tokens).or_default().extend(inner_results)
```

bh2smith (approved):
Nice chuck of code and  quickly implemented too!

shared/src/pool_fetching.rs
fleupold: Cool, have we already discussed if each liquidity type should have its own cache or if there should be a cache for all `BaselineSolvable` types?

shared/src/pool_fetching.rs
fleupold: I think so, yes.

PR #629, opened 2021-05-20
Make token quality cache expire

vkgnosis:
Because some tokens might dynamically change.

### Test Plan
new test

### DIFF
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -60,6 +60,15 @@ struct Arguments {
     /// Note that if a node does not support the api we still use the less accurate call api.
     #[structopt(long, env = "SKIP_TRACE_API")]
     skip_trace_api: bool,
+
+    /// The amount of time a classification of a token into good or bad is valid for.
+    #[structopt(
+        long,
+        env = "TOKEN_QUALITY_CACHE_EXPIRY_SECONDS",
+        default_value = "600",
+        parse(try_from_str = shared::arguments::duration_from_seconds),
+    )]
+    token_quality_cache_expiry: Duration,
 }
 
 pub async fn database_metrics(metrics: Arc<Metrics>, database: Database) -> ! {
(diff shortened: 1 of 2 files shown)


nlordell (approved):
:heavy_check_mark:

orderbook/src/main.rs
```diff
@@ -60,6 +60,15 @@ struct Arguments {
     /// Note that if a node does not support the api we still use the less accurate call api.
     #[structopt(long, env = "SKIP_TRACE_API")]
     skip_trace_api: bool,
+
+    /// The amount of time a classification of a token into good or bad is valid for.
+    #[structopt(
+        long,
+        env = "TOKEN_QUALITY_CACHE_EXPIRY",
```
fleupold, line 67: for the env var it might be nice to encode the unit (append _sec) so that when someone logs at the staging config they know this is seconds not minutes without having to jump into the codebase.

fleupold (approved):
Code looks good, although we might have a similar challenge here with keeping the cache "warm". Otherwise it's quite likely that when a user comes to cowswap the backend has to perform a lot of checks before it can return the fee/price estimate.

shared/src/bad_token/cache.rs
```diff
@@ -23,19 +28,30 @@ impl BadTokenDetecting for CachingDetector {
...
         Self {
             inner,
             cache: Default::default(),
+            cache_expiry,
         }
     }
 
-    fn get_from_cache(&self, token: &H160) -> Option<TokenQuality> {
-        self.cache.lock().unwrap().get(token).cloned()
+    fn get_from_cache(&self, token: &H160, now: Instant) -> Option<TokenQuality> {
```
bh2smith, line 39: Nice!

bh2smith (approved):

orderbook/src/main.rs
vkgnosis: Added SECONDS postfix.

PR #630, opened 2021-05-20
Add balancer vault artifact

nlordell:
This PR refactors how the artifact vendoring works in order to add new Balancer Vault contract. In particular this PR:
- Strips even more JSON artifact properties that aren't used, namely:
  - `contractName`: the build script was overriding the contract name for all generated bindings
  - `networks`: they were being set for all contracts that we needed `Contract::deployed()` for (with the exception of WETH, but that was a very small amount of data to add).
  - `bytecode`: For contracts that are only needed as an interface and are not deployed for E2E
- Creates `Sushiswap*` to `UniswapV2` symbolic links
  - This is done because they share an interface
  - The Sushiswap contracts weren't getting automatically vendored by the `vendor.rs` script.
  
Furthermore, two contracts for which we had bindings but are unused have been removed (`UniswapV2Pair` and `SushiswapV2Pair`, since we use `IUniswapLikePair` for both those contracts).

### Test Plan

Added new tests to ensure that contracts have deployment information configured for all networks that we care about (to avoid another incident like we had in staging this morning).

Otherwise, no service changes, so CI will hopefully catch all issues.


vkgnosis (approved):
I was confused about ` contracts/artifacts/SushiswapV2Router02.json` because the github diff renders this as if it c ontained the single line `UniswapV2Router02.json`. Actually it's a symbolic link to that file.

nlordell:
> I was confused about contracts/artifacts/SushiswapV2Router02.json because the github diff renders this as if it c ontained the single line UniswapV2Router02.json. Actually it's a symbolic link to that file.

Added info to the PR description

contracts/build.rs
```diff
@@ -15,37 +15,27 @@ fn main() {
...
-            .add_deployment_str(1, "0xd9e1cE17f2641f24aE83637ab66a2cca9C378B9F")
-            .add_deployment_str(4, "0x1b02dA8Cb0d097eB8D57A175b88c7D8b47997506")
-            .add_deployment_str(100, "0x1b02dA8Cb0d097eB8D57A175b88c7D8b47997506")
-    });
-    generate_contract_with_config("SushiswapV2Factory", |builder| {
-        builder
-            .add_deployment_str(1, "0xC0AEe478e3658e2610c5F7A4A2E1777cE9e4f2Ac")
-            .add_deployment_str(4, "0xc35DADB65012eC5796536bD9864eD8773aBc74C4")
-            .add_deployment_str(100, "0xc35DADB65012eC5796536bD9864eD8773aBc74C4")
+            .with_contract_mod_override(Some("balancer_v2_vault"))
```
bh2smith, line 20: Can we just call this balance_v2 since we might wind up using more than just the vault. Also the vault is already in the contract name.

contracts/src/bin/deploy.rs
```diff
@@ -125,11 +124,7 @@ async fn wait_for_node(web3: &Web3<Http>) -> Result<()> {
             "node not responding, retrying in {}s",
             NODE_READY_POLL_INTERVAL.as_secs_f64(),
         );
-
-        // NOTE: Usually a blocking call in a future is bad, but since we block
-        // on this future right at the beginning and have no concurrent fibers,
-        // it should be OK for this simple script.
-        thread::sleep(NODE_READY_POLL_INTERVAL);
+        tokio::time::delay_for(NODE_READY_POLL_INTERVAL).await;
```
bh2smith, line 127: Cool change!

contracts/src/bin/vendor.rs
```diff
@@ -65,38 +22,122 @@ fn main() {
...
-    log::info!("vendoring contract artifacts to '{}'", artifacts.display());
-    for (npm_path, local_path) in NPM_CONTRACTS {
-        log::info!("retrieving {}", npm_path);
-        let source = Source::npm(npm_path.to_string());
+    vendor
+        .full()
+        .github(
+            "BalancerV2Vault",
+            "balancer-labs/balancer-core-v2/v1.0.0-artifacts/\
+             deployments/artifacts/mainnet/Vault.json",
```
bh2smith, line 32: Are we taking advantage of the fact that we use the mainnet abi for all networks.

bh2smith (approved):
Really cool. Will look over the fact stuff again tomorrow morning.

bh2smith:
One general comment, it looks like this kinda tooling could eventually fit more naturally in ethcontract. Given that it has become quite useful and sophisticated, it would be more easily accessible for other projects.

contracts/build.rs
nlordell: The issue is that `ethcontract` doesn't support multiple contracts under the same module. Ideally, we would have:
```
mod balancer_v2 {
  type Vault;
  type WeighedPool;
}
```

But unfortunately, `ethcontract` gives us:
```
mod balancer_v2_vault {
  type Contract;
}
type BalancerV2Vault = balancer_v2_vault::Contract;

mod balancer_v2_weighted_pool {
  type Contract;
}
type BalancerV2WeightedPool = balancer_v2_weighted_pool::Contract;
```

nlordell:
> looks like this kinda tooling could eventually fit more naturally in ethcontract

https://github.com/gnosis/ethcontract-rs/issues/514

contracts/src/bin/vendor.rs
```diff
@@ -65,38 +22,122 @@ fn main() {
...
+
+impl Vendor {
+    fn new() -> Result<Self> {
+        let artifacts = paths::contract_artifacts_dir();
+        log::info!("vendoring contract artifacts to '{}'", artifacts.display());
+        fs::create_dir_all(&artifacts)?;
+        Ok(Self { artifacts })
+    }
+
+    fn full(&self) -> VendorContext {
```
fleupold, line 89: maybe add some comments here that full is needed if you want to deploy the contract (e.g. for e2e tests)

contracts/src/lib.rs
```diff
@@ -1,16 +1,93 @@
...
+            )
+        }
+
+        fn send(&self, _id: RequestId, _request: Call) -> Self::Out {
+            future::ready(Ok(json!(format!("{}", self.0))))
+        }
+    }
+
+    #[test]
+    fn deployment_addresses() {
```
fleupold, line 54: nice, I wonder how we can encourage people to add tests here whenever a new contract is added.

fleupold (approved):

contracts/src/lib.rs
bh2smith: Maybe we could somehow generate these lines from a constant list

```
include!(concat!(env!("OUT_DIR"), "/WETH9.rs"));
```

and include in this test and assertion each element from the list was verified here...

nlordell: #646

PR #633, opened 2021-05-21
[Hot Fix] Not all pools considered for estimate

nlordell:
Fixes #632 

This PR fixes an issue with how we collect our pools into a mapping of token pairs into a vector of pools. The current implementation only considers the last pool for any given token pair instead of all of them.

### Test Plan

Added new unit tests to ensure that the best pool is always chosen.

PR #634, opened 2021-05-21
Fix cached pool fetcher

### LINKED ISSUE #632: Baseline price estimator not considering all pools.
It appears that the baseline price estimator is not considering all pools for the same token pair. In fact it only considers the "last" pool retrieved by its `PoolFetching` for any given token pair. Here is a failing test:

```rust
    fn pool_price(
        pool: &Pool,
        token_out: H160,
        amount_in: impl Into<U256>,
        token_in: H160,
    ) -> BigRational {
        let amount_in = amount_in.into();
        BigRational::new(
            amount_in.to_big_int(),
            pool.get_amount_out(token_out, amount_in, token_in)
                .unwrap()
                .as_u128()
                .into(),
        )
    }

    #[tokio::test]
    async fn price_estimate_uses_best_pool() {
        let token_a = H160([0x0a; 20]);
        let token_b = H160([0x0b; 20]);

        let pools = vec![
            Pool::uniswap(
                TokenPair::new(token_a, token_b).unwrap(),
                (100_000, 100_000),
            ),
            Pool::uniswap(TokenPair::new(token_a, token_b).unwrap(), (100_000, 90_000)),
        ];

        let pool_fetcher = Box::new(FakePoolFetcher(pools.clone()));
        let gas_estimator = Arc::new(FakeGasPriceEstimator(Arc::new(Mutex::new(0.0))));
        let estimator = BaselinePriceEstimator::new(
            pool_fetcher,
            gas_estimator,
            HashSet::new(),
            Arc::new(ListBasedDetector::deny_list(Vec::new())),
            token_a,
        );

        let price = estimator
            .estimate_price(token_a, token_b, 100.into(), OrderKind::Sell)
            .await
            .unwrap();
        // Pool 0 is more favourable for buying token B.
        assert_eq!(price, pool_price(&pools[0], token_b, 100, token_a));

        let price = estimator
            .estimate_price(token_b, token_a, 100.into(), OrderKind::Sell)
            .await
            .unwrap();
        // Pool 1 is more favourable for buying token B.
        assert_eq!(price, pool_price(&pools[1], token_a, 100, token_b));
    }

```

fleupold:
Fixes #632

The CachedPoolFetcher was only caching a single pool per token pair even if multiple pools were fetched by inner. Components relying ont he Cache (e.g. the price estimate) would therefore only get a single pool on cache hits leading to suboptimal price estimates.

### Test Plan
Adjusted the unit test (failed before this change) and made sure we are now getting the best Uniswap price when trying to sell 100 WETH for UNI:

http://localhost:8080/api/v1/markets/0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2-0x1f9840a85d5aF5bf1D1762F925BDADdC4201F984/sell/99995000000000000000

And the best Sushiswap price when trying to sell 100 WETH for SUSHI

http://localhost:8080/api/v1/markets/0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2-0x6B3595068778DD592e39A122f4f5a5cF09C90fE2/sell/99992840000000000000

Compared to expected values by using https://swap.archerdao.io/#/swap

### DIFF
--- a/shared/src/pool_fetching.rs
+++ b/shared/src/pool_fetching.rs
@@ -165,10 +165,12 @@ impl BaselineSolvable for Pool {
     }
 }
 
+type TokenPoolsMap = HashMap<TokenPair, Vec<Pool>>;
+
 // Read though Pool Fetcher that keeps previously fetched pools in a cache, which get invalidated whenever there is a new block
 pub struct CachedPoolFetcher {
     inner: Box<dyn PoolFetching>,
-    cache: Arc<Mutex<(Block, HashMap<TokenPair, Vec<Pool>>)>>,
+    cache: Arc<Mutex<(Block, TokenPoolsMap)>>,
     block_stream: CurrentBlockStream,
 }
 
--- a/shared/src/pool_fetching.rs
+++ b/shared/src/pool_fetching.rs
@@ -168,7 +168,7 @@ impl BaselineSolvable for Pool {
 // Read though Pool Fetcher that keeps previously fetched pools in a cache, which get invalidated whenever there is a new block
 pub struct CachedPoolFetcher {
     inner: Box<dyn PoolFetching>,
-    cache: Arc<Mutex<(Block, HashMap<TokenPair, Pool>)>>,
+    cache: Arc<Mutex<(Block, HashMap<TokenPair, Vec<Pool>>)>>,
     block_stream: CurrentBlockStream,
 }
 
@@ -189,12 +189,13 @@ impl CachedPoolFetcher {
         let cache_results: Vec<_> = cache_hits
             .iter()
             .filter_map(|pair| cache.1.get(pair))
+            .flatten()
             .cloned()
             .collect();
 
         let mut inner_results = self.inner.fetch(cache_misses).await;
         for miss in &inner_results {
-            cache.1.insert(miss.tokens, miss.clone());
+            cache.1.entry(miss.tokens).or_default().push(miss.clone());
         }
         inner_results.extend(cache_results);
         inner_results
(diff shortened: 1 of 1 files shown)


nlordell (approved):
Looks good.

~Does invalidating the cache empty the entire `HashMap`?~ I see the entire hashmap gets cleared. I was just confused because it wasn't part of the diff. My bad!

fleupold:
This is also covered by the unit test.

vkgnosis (approved):

shared/src/pool_fetching.rs
```diff
@@ -189,12 +191,13 @@ impl CachedPoolFetcher {
...
             .iter()
             .filter_map(|pair| cache.1.get(pair))
+            .flatten()
             .cloned()
             .collect();
 
         let mut inner_results = self.inner.fetch(cache_misses).await;
         for miss in &inner_results {
-            cache.1.insert(miss.tokens, miss.clone());
+            cache.1.entry(miss.tokens).or_default().push(miss.clone());
```
bh2smith, line 200: What does the term "miss" represent here?

bh2smith: could we include a comment here on what is happening. I find this very cryptic and I am not sure I even understand the difference between the changes.

bh2smith (commented):
Seems good, but I am lacking some understanding on the cache insertion.

shared/src/pool_fetching.rs
fleupold: > What does the term "miss" represent here?

It represents a cache miss (the opposite of a cache hit).

> I am not sure I even understand the difference between the changes.

It's exactly the same as [in the other fix](https://github.com/gnosis/gp-v2-services/pull/633/files#diff-f6933a68a986a710ace700404bf73cb176ec2a1abac83803335794b8f91db01bR324)

Instead of overriding the hash map with a single value, we load the existing value (which is a list) and append to it. Happy for a suggestion on how to write this in a way that is easier to understand, however I'm not sure a comment is a good idea (comments should focus on why something is happening and not repeat the "what" which is already implicit by the code).

bh2smith: That is already helpful enough. Thanks.

PR #637, opened 2021-05-21
Fix empty pool results not getting cached

vkgnosis:
I was debugging what takes how much time when estimating prices and was confused why we would hit the inner pool fetcher so often. This took me a while to debug but eventually I found that we have a bug where we don't cache the information that the inner pool has returned no pools for a token pair.
This happens because we only create an entry in the cache hash map if there is at least one pool for the pair. This PR fixes this issue by always creating an entry which then stays empty and is counted as a cache hit next call.

### Test Plan
new unit test

### DIFF
--- a/shared/src/pool_fetching.rs
+++ b/shared/src/pool_fetching.rs
@@ -1,12 +1,13 @@
-use std::collections::{HashMap, HashSet};
-
 use crate::{baseline_solver::BaselineSolvable, Web3};
 use contracts::{IUniswapLikePair, ERC20};
 use ethcontract::{batch::CallBatch, BlockNumber, H160, U256};
 use lru::LruCache;
 use model::TokenPair;
 use num::{rational::Ratio, BigInt, BigRational, Zero};
-use std::sync::Arc;
+use std::{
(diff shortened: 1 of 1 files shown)


fleupold (approved):
Very good 🕵️ work and a great improvement!

shared/src/pool_fetching.rs
```diff
@@ -185,21 +186,36 @@ impl CachedPoolFetcher {
...
+                Entry::Occupied(entry) => {
+                    cache_hits.extend_from_slice(entry.get());
+                }
+                Entry::Vacant(entry) => {
+                    cache_misses.insert(pair);
+                    // It is possible that the inner fetcher when queried below does not return any
+                    // pools for a token pair. It is important that this information itself enters
+                    // the cache so that it is remembered on next fetch and we do not attempt to
+                    // fetch pools for the token pair again (until the cache expires).
+                    entry.insert(Default::default());
```
nlordell, line 203: I was surprised to find that there is no `insert_default()`. TIL!

nlordell (approved):

bh2smith (approved):
Fantastic! Super happy this was discovered and resolved. So quick!

PR #641, opened 2021-05-21
Balancer Event Handling:  Part I - Streaming

bh2smith:
Partial fulfillment of Step 2 from #603 

Putting together the wiring to Stream Balancer events into our services. At the moment they are not handled, but just logged

Most of the code changes are in `balancer_event_handler.rs` which introduces
- `enum BalancerEvent`
- `struct PoolRegistered` - The Pool Registered Event
- `enum PoolSpecialization` 
- `struct BalancerPools` - currently empty, but will represent the Store of static data for `WeightedPools`
- `struct BalancerEventUpdater` - analogous to the `EventUpdater` for `GPv2Settlement`(which should probably be renamed to `GPv2SettlementEventUpdater` 
- `impl EventStoring<ContractEvent> for BalancerPools` - This is currently only receiving the event stream and not actually storing them.
- some conversion methods to transform `EthcontractEvents` into `BalancerEvents`

Other minor changes introduced here are 

- moving `EventIndex` into shared and introducing a constructor `From<EventMetadata>`
- generalizing the `event_retrieving` macro to use absolute paths.


contracts/artifacts/Vault.json
```diff
@@ -0,0 +1,1188 @@
+{
+  "_format": "hh-sol-artifact-1",
```
vkgnosis, line 2: I think this file is leftover from rebasing on top of https://github.com/gnosis/gp-v2-services/pull/630 . Because that PR should already have added this artifact.

bh2smith: Yes. I thought I had gotten rid of it. Reason its still here is that the filename there is `BalancerV2Vault` and here is just `Vault` I think. Will take care of this.

bh2smith: Yes. Removed. Good eye 👁️

vkgnosis (approved):

shared/src/balancer_event_handler.rs
```diff
@@ -0,0 +1,142 @@
...
+
+pub struct BalancerEventUpdater(
+    Mutex<EventHandler<DynWeb3, BalancerV2VaultContract, BalancerPools>>,
+);
+
+impl BalancerEventUpdater {
+    pub fn new(
+        contract: BalancerV2Vault,
+        pools: BalancerPools,
+        start_sync_at_block: Option<u64>,
```
nlordell, line 79: This shouldn't be needed as its included in the deployment information of the `BalancerV2Vault`. i.e., you could use `contract.deployment_information()` to get the transaction hash/block number of when the contract got deployed.

orderbook/src/main.rs
```diff
@@ -187,6 +192,7 @@ async fn main() {
             orderbook.clone(),
             Arc::new(database.clone()),
             Arc::new(event_updater),
+            Arc::new(vault_updater),
```
nlordell, line 195: This shouldn't be added the the list of service maintainers because of the outstanding `todo!()` which would cause panics in staging.

orderbook/src/main.rs
```diff
@@ -120,6 +124,7 @@ async fn main() {
 
     let event_updater =
         EventUpdater::new(settlement_contract.clone(), database.clone(), sync_start);
+    let vault_updater = BalancerEventUpdater::new(vault_contract, BalancerPools::default(), None);
```
nlordell, line 127: Should probably leave instantiating + adding this component until it doesn't panic anymore.

shared/src/balancer_event_handler.rs
```diff
@@ -0,0 +1,142 @@
...
+
+#[derive(Default)]
+pub struct BalancerPools {
+    // pools: HashMap<H160, HashSet<WeightedPool>>,
+// Block number of last update
+// last_updated: u64,
+}
+
+impl BalancerPools {
+    pub fn contract_to_balancer_events(
```
nlordell, line 46: I don't think this needs to be `pub`.

orderbook/src/main.rs
bh2smith: Yes. it is in the above checklist to remove before merge. This was only for my local testing to ensure the events were being streamed.

shared/src/balancer_event_handler.rs
bh2smith: I am looking at ethcontract docs and I see that this method returns an Option enum:
```
pub enum DeploymentInformation {
    BlockNumber(u64),
    TransactionHash(TransactionHash),
}
```

Meaning that it is optionally block number or transaction hash... I suppose I should handle either case here.

bh2smith: Seems that if we wind up with `TransactionHash` at the time this new method is called then we have to make a request to fetch the block number which would make this method async.

bh2smith: I have updated this bit here with a somewhat incomplete match statement that leave the `start_sync_at_block` None when the deployment info returns a TransactionHash.

Please take a look and let me know if you had something else in mind.

shared/src/balancer_event_handler.rs
```diff
@@ -0,0 +1,150 @@
...
+    pub fn new(contract: BalancerV2Vault, pools: BalancerPools) -> Self {
+        let mut start_sync_at_block = None;
+        if let Some(deployment_info) = contract.deployment_information() {
+            match deployment_info {
+                DeploymentInformation::BlockNumber(block_number) => {
+                    start_sync_at_block = Some(block_number);
+                }
+                // Having only transaction hash would require a web3 request
+                // to fetch the block number, but this method is not async.
+                DeploymentInformation::TransactionHash(_hash) => (),
```
nlordell, line 86: There is nothing preventing this method from becoming `async` right?

bh2smith: Nothing really... I was just thinking its more logic than may be necessary to introduce. I will go ahead and make it async.

nlordell: Alternatively, `EventHandler` could be extended to take a `Option<DeploymentInformation>` instead of a `Option<BlockNumber>`.

bh2smith: Alright I have made the method async and implemented some kinda ugly looking block number fetching when available. I have a feeling this new code introduced could be more concise. Looking for suggestions.

shared/src/balancer_event_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+    },
+    BalancerV2Vault,
+};
+use ethcontract::common::DeploymentInformation;
+use ethcontract::{dyns::DynWeb3, Event as EthContractEvent, EventMetadata, H160};
+use std::ops::RangeInclusive;
+use tokio::sync::Mutex;
+
+#[derive(Debug)]
+pub enum BalancerEvent {
```
fleupold, line 20: Are you going to listen to other events as well in the short future? If so, which ones? If not, maybe you don't need an enum and can instead work directly with the PoolRegistered struct? (enums with one variant look a bit 🤡 - like a bullet list with one bullet point)

shared/src/balancer_event_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+
+#[derive(Debug, Default)]
+pub struct PoolRegistered {
+    pub pool_id: [u8; 32],
+    pub pool_address: H160,
+    pub specialization: u8,
+}
+
+#[derive(Default)]
+pub struct WeightedPool {
```
fleupold, line 32: Maybe get rid of this until it's needed?

shared/src/balancer_event_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+#[derive(Debug)]
+pub enum BalancerEvent {
+    PoolRegistered(PoolRegistered),
+}
+
+#[derive(Debug, Default)]
+pub struct PoolRegistered {
+    pub pool_id: [u8; 32],
+    pub pool_address: H160,
+    pub specialization: u8,
```
fleupold, line 28: Can you add comments to this struct? Specifically specialization is not intuitive to me (might suffice to link to the balancer docs explaining what this is)

shared/src/event_handling.rs
```diff
@@ -178,6 +180,21 @@ where
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct EventIndex {
```
fleupold, line 184: 👍

shared/src/balancer_event_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+pub struct WeightedPool {
+    // pool_address: H160,
+// tokens: Vec<H160>,
+// pool_id:
+// TODO - other stuff
+}
+
+#[derive(Default)]
+pub struct BalancerPools {
+    // pools: HashMap<H160, HashSet<WeightedPool>>,
```
fleupold, line 41: It would be nice to get an idea what we will be storing here. and maybe populate that already. I assume the actual WeightedPools will be fetched lazily when needed and we will just store a mapping of token to pool addresses and maybe block number (for reorg handling)?

shared/src/balancer_event_handler.rs
bh2smith: I believe we will have to listen to at least one other event (`TokensDeregistered`) because apparently it is possible that the pool could become "inactive" however the Balancer team has mentioned this will likely never happen, we should at least be prepared for the fact that it can, given that it is within the functionality of the deployed contracts. Maybe @nlordell wants to say something about potential other events we might want to listen for and act upon.

shared/src/balancer_event_handler.rs
bh2smith: Yes, this is more relevant to the next PR where we start populating this struct. I can shed a bit on light on it here though. 

I initially thoughts we would only need `HashMap<H160, HashSet<WeightedPool>>` (i.e. Tokens -> WeightedPools), but soon realized we also need to keep another `HashMap<PoolId, WeightedPool>` so we know if these pools have already been already been included. It may be possible that this is not needed, but I'd rather save this discussion for the following PR (https://github.com/gnosis/gp-v2-services/pull/642) still in Draft state.

shared/src/balancer_event_handler.rs
bh2smith: Here is a list of events emitted by the vault:

```
    AuthorizerChanged(IAuthorizer indexed newAuthorizer)
    ExternalBalanceTransfer(IERC20 indexed token, address indexed sender, address recipient, uint256 amount)
    FlashLoan(IFlashLoanRecipient indexed recipient, IERC20 indexed token, uint256 amount, uint256 feeAmount);
    InternalBalanceChanged(address indexed user, IERC20 indexed token, int256 delta)
    PausedStateChanged(bool paused);
    PoolBalanceChanged(bytes32 indexed poolId, address indexed liquidityProvider, IERC20[] tokens, int256[] deltas, uint256[] protocolFeeAmounts);
    PoolBalanceManaged(bytes32 indexed poolId,address indexed assetManager,IERC20 indexed token,int256 cashDelta,int256 managedDelta);
    PoolRegistered(bytes32 indexed poolId, address indexed poolAddress, PoolSpecialization specialization);
    RelayerApprovalChanged(address indexed relayer, address indexed sender, bool approved)
    Swap(bytes32 indexed poolId,IERC20 indexed tokenIn,IERC20 indexed tokenOut,uint256 amountIn,uint256 amountOut);
    TokensDeregistered(bytes32 indexed poolId, IERC20[] tokens);
    TokensRegistered(bytes32 indexed poolId, IERC20[] tokens, address[] assetManagers);
```

nlordell: I'm also OK with leaving it as a placeholder for now.

shared/src/balancer_event_handler.rs
nlordell: This is a good point. I would suggest:
```
pub struct PoolRegistered {
    pub pool_id: H256,
    pub pool_address: H160,
    pub specialization: PoolSpecialization,
}

#[repr(u8)]
pub enum PoolSpecialization {
    General = 0,
    MinimalSwapInfo = 1,
    TwoToken = 2,
}
```

nlordell: In theory, we could even copy [the documentation](https://github.com/balancer-labs/balancer-core-v2/blob/3b8c2fb5c75e987d63c4d44f9dd62351f43e153b/pkg/vault/contracts/interfaces/IVault.sol#L197-L212) from the Balancer repo to keep it close.

shared/src/balancer_event_handler.rs
nlordell: I'm also fine with leaving this empty for now and discussing the best approach in the following PR. As @fleupold suggested, we will likely need to keep some additional info in order to deal with re-orgs.

shared/src/current_block.rs
```diff
@@ -156,6 +157,23 @@ where
             .context("failed to get current block")?
             .as_u64())
     }
+
+    async fn block_number_from_tx_hash(&self, hash: H256) -> Option<u64> {
```
nlordell, line 161: I think this should propagate the failure to the caller.

```suggestion
    async fn block_number_from_tx_hash(&self, hash: H256) -> Result<u64> {
```

Returning just an `Option` makes the `.context(...)` not do anything either.

shared/src/balancer_event_handler.rs
```diff
@@ -0,0 +1,154 @@
...
+
+pub struct BalancerEventUpdater(
+    Mutex<EventHandler<DynWeb3, BalancerV2VaultContract, BalancerPools>>,
+);
+
+impl BalancerEventUpdater {
+    pub async fn new(contract: BalancerV2Vault, pools: BalancerPools) -> Self {
+        let mut deployment_block = None;
+        if let Some(deployment_info) = contract.deployment_information() {
+            match deployment_info {
```
nlordell, line 80: One thing you can do here is use a single `match` instead of `if ... match`:
```rust
let deployment_info = match contract.deployment_info() {
    Some(DeploymentInformation::BlockNumber(block_number)) => {
        Some(block_number)
    }
    Some(DeploymentInformation::TransactionHash(hash)) => {
        Some(contract
            .raw_instance()
            .web3()
            .block_number_from_tx_hash(hash)
            .await?)
    }
    None => None,
}
```

shared/src/balancer_event_handler.rs
bh2smith: Ok, so I have included the suggestions here. Note that I had to quite a bit of manual labor for this `PoolSpecialization`struct (i.e. manual implementation of `Debug` and an explicit `new` constructor which builds the enum from u8, but has to return a result because of Error when the value is too high).

Not sure if I am missing out on some sweet tricks here, but I guess thats all part of the safety guarantees!

I also, don't see the real benefit of using `H256` here for `pool_id` instead of `[u8; 32]` if we aren't going to make use of any functionality inherited from the fact that its an `H256`.

shared/src/balancer_event_handler.rs
```diff
@@ -0,0 +1,198 @@
...
+        match specialization {
+            0 => Ok(Self::General),
+            1 => Ok(Self::MinimalSwapInfo),
+            2 => Ok(Self::TwoToken),
+            t => Err(anyhow!("Invalid PoolSpecialization value {}", t)),
+        }
+    }
+}
+
+impl std::fmt::Debug for PoolSpecialization {
```
nlordell, line 64: Why not just `derive(Debug)`?

AFAIK it just works:
```rust
#[derive(Debug)]
#[repr(u8)]
enum Foo {
    Bar = 0,
    Baz = 1,
}

fn main() {
    println!("{:?} and {:?}", Foo::Bar, Foo::Baz); // prints "Bar and Baz"
}
```

shared/src/balancer_event_handler.rs
```diff
@@ -0,0 +1,198 @@
...
+/// constant product one popularized by Balancer V1. Swap costs are smaller compared to general Pools, and are
+/// independent of the number of registered tokens.
+///
+///  - Two Token: only allows two tokens to be registered. This achieves the lowest possible swap gas cost. Like
+/// minimal swap info Pools, these are called via IMinimalSwapInfoPool.
+#[repr(u8)]
+pub enum PoolSpecialization {
+    General = 0,
+    MinimalSwapInfo = 1,
+    TwoToken = 2,
```
nlordell, line 50: Maybe include a link to the Balancer docs? Also, I would suggest having the different pool speiciliazation docs inline:
```suggestion
/// minimal swap info Pools, these are called via IMinimalSwapInfoPool.
#[repr(u8)]
pub enum PoolSpecialization {
    /// No specialization, suited for all Pools. ...
    General = 0,
    /// IMinimalSwapInfoPool is used instead ...
    MinimalSwapInfo = 1,
    /// Only allows two tokens to be registered. ...
    TwoToken = 2,
```

shared/src/current_block.rs
```diff
@@ -156,6 +159,18 @@ where
...
+        let potential_receipt = self.eth().transaction_receipt(hash).await?;
+        if let Some(receipt) = potential_receipt {
+            // Need to unwrap Option<U64> to convert U64 -> u64
+            if let Some(block_number) = receipt.block_number {
+                return Ok(block_number.as_u64());
+            }
+            return Err(anyhow!("no block number with transaction receipt"));
+        }
+        return Err(anyhow!("no transaction receipt found"));
+    }
```
nlordell, line 173: This can be simplified using `Option::ok_or_else` which promotes the `Option` to a `Result` by specifying an error for `None`:

```suggestion
        Ok(self.eth()
            .transaction_receipt(hash)
            .await?
            .ok_or_else(|| anyhow!("no transaction receipt found"))?
            .block_number
            .ok_or_else(|| anyhow("no block number with transaction receipt"))?
            .as_u64())
```

shared/src/balancer_event_handler.rs
bh2smith: I totally tried that, but maybe it was the ordering that didn't work.

I had
```
#[repr(u8)]
#[derive(Debug)]
```

Will try again

shared/src/current_block.rs
bh2smith: I knew there was something nicer that could be done!

shared/src/balancer_event_handler.rs
nlordell: Just checked an order shouldn't matter: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b0ffd9b7738412ad007638858ee4c1cc

shared/src/balancer_event_handler.rs
nlordell: > Maybe include a link to the Balancer docs?

On second thought... I don't think this has much value.

shared/src/balancer_event_handler.rs
```diff
@@ -0,0 +1,180 @@
...
+                {
+                    Ok(block_number) => Some(block_number),
+                    Err(err) => {
+                        tracing::warn!(
+                            "no deployment block found for transaction hash {}: {}",
+                            hash,
+                            err
+                        );
+                        None
+                    }
```
nlordell, line 114: Why not make this a hard error?

nlordell (commented):
Looks great! Just one small suggestion :point_up:

vkgnosis (approved):

nlordell (approved):

PR #642, opened 2021-05-21
Balancer Event Handling: Part II - Event Storing

bh2smith:
Partial fulfillment of #603

This PR introduces a full implementation of `EventStoring` for `BalancerPool` (append_events, replace_events and last_event_block).

Some particular design decisions (open for discussion) made at this point are as follows:

1. BalancerPool stores the pool data as two hashmaps: namely
```
    pools_by_token: HashMap<H160, HashSet<PoolId>>,
    pools: HashMap<PoolId, WeightedPool>,
```
this gives us O(1) lookup for pool ids for a given token without duplicating the entire `WeightedPool` struct in memory. One could argue that this isn't necessary as this is not expected to be a terribly large amount of storage (Currently the BalancerPool struct is about 41kb when pretty-printed in a text file).

2. Both `TokensRegistered` and `PoolRegistered` events are emmited seperately in the same transaction (see [here](https://github.com/balancer-labs/balancer-v2-monorepo/blob/70843e6a61ad11208c1cfabf5cfe15be216ca8d3/pkg/pool-utils/contracts/BasePool.sol#L128-L130)), but are both needed in order to have complete information about the `WeightedPool` struct. For this reason (because we process events one at a time) I have made a couple of the fields Optional. In particular

```rust
    pool_address: Option<H160>,
    tokens: Option<Vec<H160>>,
    specialization: Option<PoolSpecialization>,
```

While both events contain the `PoolId`, `TokensRegistered` only contains token information while `PoolsRegistered` yields `pool_address` and `specialization`. We thus have

In practice, the plan would be to introduce a helper methods `is_active` to `WeightedPool` returning `pool_address.is_some() && tokens.is_some()`. I can understand that this is not necessarily ideal and that we could approach this differently (e.g. a PreWeightedPool with the optional data that is upgraded to a `WeightedPool` only when all information becomes available). It might be nice to not have these fields optional.

3. On the less controversial side of things, the implementation of `last_event_block` literally iterates over all pools and takes the max block_created. This is analogous to how we evaluate for the postgres database and probably not terribly resource intensive. However, with a bit more technical overhead we could keep this value updated for immediate lookup on every update to the `PoolStore` (however this would require more scattered logic that would also need to be tested).

## Testing 

We unit test all the non-async methods on `BalancerPool` which are responsible for updates called via the async trait methods of `EventStoring`.

- insert_events
- replace_events

Only methods not directly tested are the async trait methods and those which relate to conversion of `EthcontractEvent` -> `BalancerEvent`

Happy to discuss and looking forward to your reviews.

### DIFF
--- a/orderbook/src/event_updater.rs
+++ b/orderbook/src/event_updater.rs
@@ -30,7 +30,7 @@ impl EventStoring<ContractEvent> for Database {
             db_events.len(),
             range.start().to_u64()
         );
-        Database::replace_events(self,range.start().to_u64(), db_events)
+        Database::replace_events(self, range.start().to_u64(), db_events)
             .await
             .context("failed to replace trades")?;
         Ok(())
--- a/orderbook/src/event_updater.rs
+++ b/orderbook/src/event_updater.rs
@@ -18,7 +18,7 @@ pub struct EventUpdater(Mutex<EventHandler<DynWeb3, GPv2SettlementContract, Data
 #[async_trait::async_trait]
 impl EventStoring<ContractEvent> for Database {
     async fn replace_events(
-        &self,
+        &mut self,
         events: Vec<Event<ContractEvent>>,
         range: RangeInclusive<BlockNumber>,
     ) -> Result<()> {
@@ -30,13 +30,13 @@ impl EventStoring<ContractEvent> for Database {
             db_events.len(),
             range.start().to_u64()
         );
-        self.replace_events(range.start().to_u64(), db_events)
+        Database::replace_events(self,range.start().to_u64(), db_events)
             .await
             .context("failed to replace trades")?;
         Ok(())
     }
 
-    async fn append_events(&self, events: Vec<Event<ContractEvent>>) -> Result<()> {
+    async fn append_events(&mut self, events: Vec<Event<ContractEvent>>) -> Result<()> {
         let db_events = self
             .contract_to_db_events(events)
             .context("append - failed to convert events")?;
--- a/shared/src/balancer_event_handler.rs
+++ b/shared/src/balancer_event_handler.rs
@@ -387,7 +387,10 @@ mod tests {
         let lonely_token_registration_index = 2 * n as u64 + 1;
         let stand_alone_token_registration = TokensRegistered {
             pool_id: H256::from_low_u64_be(lonely_token_registration_index),
-            tokens: vec![H160::from_low_u64_be(lonely_token_registration_index), H160::from_low_u64_be(lonely_token_registration_index + 1)],
+            tokens: vec![
+                H160::from_low_u64_be(lonely_token_registration_index),
+                H160::from_low_u64_be(lonely_token_registration_index + 1),
+            ],
         };
 
         let events: Vec<(EventIndex, BalancerEvent)> = vec![
@@ -402,9 +405,15 @@ mod tests {
             (EventIndex::new(3, 0), token_registration_events[2].clone()),
             (EventIndex::new(4, 0), pool_registration_events[2].clone()),
             // Stand alone Token registration
-            (EventIndex::new(5, 0), BalancerEvent::TokensRegistered(stand_alone_token_registration.clone())),
+            (
+                EventIndex::new(5, 0),
+                BalancerEvent::TokensRegistered(stand_alone_token_registration.clone()),
+            ),
             // Stand along Pool Registration
-            (EventIndex::new(6, 0), BalancerEvent::PoolRegistered(stand_alone_pool_registration.clone())),
+            (
+                EventIndex::new(6, 0),
+                BalancerEvent::PoolRegistered(stand_alone_pool_registration.clone()),
+            ),
         ];
 
         let mut pool_store = BalancerPools::default();
@@ -434,18 +443,25 @@ mod tests {
                 &WeightedPool {
                     pool_id: pool_ids[i],
                     pool_address: pool_addresses[i],
-                    tokens: vec![tokens[i], tokens[i+1]],
+                    tokens: vec![tokens[i], tokens[i + 1]],
                     specialization: PoolSpecialization::new(i as u8).unwrap(),
                     block_created: i as u64 + 1
                 },
-                "failed assertion at index {}", i
+                "failed assertion at index {}",
+                i
             );
             assert!(pool_store.pending_pools.get(&pool_ids[i]).is_none());
         }
 
-        assert!(pool_store.pools.get(&stand_alone_token_registration.pool_id).is_none());
+        assert!(pool_store
+            .pools
+            .get(&stand_alone_token_registration.pool_id)
+            .is_none());
         assert_eq!(
-            pool_store.pending_pools.get(&stand_alone_token_registration.pool_id).unwrap(),
+            pool_store
+                .pending_pools
+                .get(&stand_alone_token_registration.pool_id)
+                .unwrap(),
             &WeightedPoolBuilder {
                 pool_registration: None,
                 tokens_registration: Some(stand_alone_token_registration),
@@ -453,9 +469,15 @@ mod tests {
             },
         );
 
-        assert!(pool_store.pools.get(&stand_alone_pool_registration.pool_id).is_none());
+        assert!(pool_store
+            .pools
+            .get(&stand_alone_pool_registration.pool_id)
+            .is_none());
         assert_eq!(
-            pool_store.pending_pools.get(&stand_alone_pool_registration.pool_id).unwrap(),
+            pool_store
+                .pending_pools
+                .get(&stand_alone_pool_registration.pool_id)
+                .unwrap(),
             &WeightedPoolBuilder {
                 pool_registration: Some(stand_alone_pool_registration),
                 tokens_registration: None,
@@ -581,12 +603,7 @@ mod tests {
             assert!(pool_store.pools.get(pool_id).is_none());
         }
         for token in tokens.iter().take(7).skip(4) {
-            assert!(
-                pool_store
-                .pools_by_token
-                .get(token)
-                .unwrap()
-                .is_empty());
+            assert!(pool_store.pools_by_token.get(token).unwrap().is_empty());
         }
 
         // All new data is included.
--- a/shared/src/balancer_event_handler.rs
+++ b/shared/src/balancer_event_handler.rs
@@ -60,7 +60,7 @@ pub struct BalancerPools {
 
 /// There are three specialization settings for Pools,
 /// which allow for cheaper swaps at the cost of reduced functionality:
-#[derive(Debug)]
+#[derive(Debug, Clone, Copy)]
 #[repr(u8)]
 pub enum PoolSpecialization {
     /// no specialization, suited for all Pools. IGeneralPool is used for swap request callbacks,
@@ -373,4 +373,125 @@ fn convert_tokens_registered(
         EventIndex::from(meta),
         BalancerEvent::TokensRegistered(event),
     ))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use maplit::{hashmap, hashset};
+
+    #[test]
+    fn balancer_insert_pool() {
+        let mut pool_store = BalancerPools::default();
+        let index = EventIndex {
+            block_number: 1,
+            log_index: 0
+        };
+        let pool_id = H256::from_low_u64_be(1);
+        let pool_address = H160::from_low_u64_be(1);
+        let specialization = PoolSpecialization::General;
+
+        let registration = PoolRegistered {
+            pool_id,
+            pool_address,
+            specialization,
+        };
+
+        pool_store.insert_pool(index, registration);
+        let expected = hashmap! {H256::from_low_u64_be(1) => WeightedPool {
+            pool_id,
+            pool_address: Some(pool_address),
+            tokens: None,
+            block_created: 1,
+        }};
+        assert_eq!(pool_store.pools, expected);
+        assert_eq!(pool_store.pools_by_token, HashMap::new());
+        
+        // Branch when token event already stored
+        let tokens = vec![H160::from_low_u64_be(2), H160::from_low_u64_be(3)];
+        let mut weighted_pool = WeightedPool {
+            pool_id,
+            pool_address: None,
+            tokens: Some(tokens.clone()),
+            block_created: 1,
+        };
+        let mut pools = hashmap! {H256::from_low_u64_be(1) => weighted_pool.clone() };
+        let pools_by_token = hashmap! {
+            tokens[0] => hashset! { pool_id },
+            tokens[1] => hashset! { pool_id },
+        };
+        let mut pool_store = BalancerPools {
+            pools_by_token: pools_by_token.clone(),
+            pools: pools.clone(),
+            contract_deployment_block: 0,
+        };
+
+        let registration = PoolRegistered {
+            pool_id,
+            pool_address,
+            specialization,
+        };
+
+        pool_store.insert_pool(index, registration);
+        weighted_pool.pool_address = Some(pool_address);
+        *pools.get_mut(&pool_id).unwrap() = weighted_pool;
+        assert_eq!(pool_store.pools, pools);
+        assert_eq!(pool_store.pools_by_token, pools_by_token);
+    }
+
+    #[test]
+    fn balancer_include_token_data() {
+        let mut pool_store = BalancerPools::default();
+        let index = EventIndex {
+            block_number: 1,
+            log_index: 0
+        };
+        let pool_id = H256::from_low_u64_be(1);
+        let tokens = vec![H160::from_low_u64_be(2), H160::from_low_u64_be(3)];
+
+        let registration = TokensRegistered {
+            pool_id,
+            tokens: tokens.clone(),
+        };
+
+        pool_store.include_token_data(index, registration);
+        let expected_pool_map = hashmap! {H256::from_low_u64_be(1) => WeightedPool {
+            pool_id,
+            pool_address: None,
+            tokens: Some(tokens.clone()),
+            block_created: 1,
+        }};
+        let expected_token_map = hashmap! {
+            tokens[0] => hashset! { pool_id },
+            tokens[1] => hashset! { pool_id },
+        };
+        assert_eq!(pool_store.pools, expected_pool_map);
+        assert_eq!(pool_store.pools_by_token, expected_token_map);
+
+
+    }
+
+    #[test]
+    fn balancer_delete_pools() {
+    }
+
+    #[test]
+    fn balancer_insert_events() {
+    }
+
+    #[test]
+    fn balancer_replace_events() {
+    }
+
+    #[test]
+    fn event_conversion() {
+        // tokens registered
+        // pools registered
+    }
+
+    #[tokio::test]
+    #[ignore]
+    async fn event_storing_trait() {}
+
+
 }
--- a/shared/src/balancer_event_handler.rs
+++ b/shared/src/balancer_event_handler.rs
@@ -275,24 +275,24 @@ pub struct BalancerEventUpdater(
 );
 
 impl BalancerEventUpdater {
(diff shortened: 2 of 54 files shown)


shared/src/balancer_event_handler.rs
```diff
@@ -7,31 +7,107 @@ use crate::{
...
+                self.specialization = Some(pool_registration.specialization);
+            }
+            BalancerEvent::TokensRegistered(token_registration) => {
+                self.tokens = Some(token_registration.tokens)
+            }
+        }
+    }
+}
+
+impl From<(EventIndex, PoolRegistered)> for WeightedPool {
```
fleupold, line 71: nanonit: Maybe you can just get away with a single new and then use `update_from_event` for both the  `PoolRegistered` and `TokensRegistered`

shared/src/balancer_event_handler.rs
```diff
@@ -7,31 +7,107 @@ use crate::{
...
+    pub pool_id: H256,
+    pub tokens: Vec<H160>,
+}
+
+#[derive(Clone, Debug, Default, Eq, PartialEq, Hash)]
+pub struct WeightedPool {
+    pool_id: H256,
+    pool_address: Option<H160>,
+    tokens: Option<Vec<H160>>,
+    specialization: Option<PoolSpecialization>,
```
fleupold, line 51: The fact that these are never really None is a pit hacky. Maybe we can have a private builder struct or the like but have the publicly facing struct only contain non-nullable values?

shared/src/balancer_event_handler.rs
```diff
@@ -54,15 +130,124 @@ impl PoolSpecialization {
...
+                existing_pool
+                    .update_from_event(BalancerEvent::TokensRegistered(registration.clone()));
+                tracing::debug!(
+                    "Pool Tokens recovered for existing Balancer pool with id {:?}",
+                    pool_id
+                )
+            }
+        }
+
+        // In either of the above cases we can now populate pools_by_token
```
fleupold, line 197: I see that these events happen in the same transactions, but conecptually shouldn't we only advertise pools that are "completely" initialized?

shared/src/balancer_event_handler.rs
```diff
@@ -54,15 +130,124 @@ impl PoolSpecialization {
...
+            })
+            .collect::<HashSet<H256>>();
+
+        self.pools
+            .retain(|pool_id, _| !pool_ids_to_delete.contains(pool_id));
+        // Remove all deleted pool_ids from token listing.
+        // Note that this could result in an empty set for some tokens.
+        for (_, pool_set) in self.pools_by_token.iter_mut() {
+            *pool_set = pool_set
+                .difference(&pool_ids_to_delete)
```
fleupold, line 234: Could also just union with the remaining pools (then you wouldn't need the filter_map).

shared/src/balancer_event_handler.rs
```diff
@@ -7,31 +7,107 @@ use crate::{
...
+
+/// The BalancerPool struct represents in-memory storage of all deployed Balancer Pools
+#[derive(Debug, Default)]
+pub struct BalancerPools {
+    /// Used for O(1) access to all pool_ids for a given token
+    pools_by_token: HashMap<H160, HashSet<H256>>,
+    /// WeightedPool data for a given PoolId
+    pools: HashMap<H256, WeightedPool>,
+    /// Used mainly as a fallback for `last_event_block` when there are no events.
+    contract_deployment_block: u64,
```
fleupold, line 105: No big deal, but is this really needed given that we pass deployment information with the contract instance? Couldn't we just default to block 0?

shared/src/balancer_event_handler.rs
```diff
@@ -129,19 +345,20 @@ impl EventStoring<ContractEvent> for BalancerPools {
             balancer_events.len(),
             range.start().to_u64()
         );
-        todo!()
+        // Not sure we will event need to replace events as we don't store the events themselves.
```
fleupold, line 334: I think this kind of question shouldn't go in the code but on the PR if possible (it would be weird for future maintainers to read this).

To answer this question specifically, I think technically we need to replace events in case of reorgs (otherwise we might store a pool which was only mined in the uncle chain). Unlikely, but given that the logic is quite straight forward, I'd leave it in.

shared/src/balancer_event_handler.rs
```diff
@@ -167,3 +384,256 @@ fn convert_pool_registered(
...
+        pool_store.delete_pools(3).unwrap();
+        assert_eq!(pool_store.last_event_block(), 2);
+        pool_store.delete_pools(2).unwrap();
+        assert_eq!(pool_store.last_event_block(), 1);
+        pool_store.delete_pools(1).unwrap();
+        assert_eq!(pool_store.last_event_block(), 0);
+    }
+
+    #[test]
+    fn balancer_insert_events() {
```
fleupold, line 560: You could add a unit test with a reorg where a pool was created but later we replace events without the pool creation (asserting that the pool is no longer there)

fleupold (changes requested):
Looks good! My main concern is around the publicly facing WeightedPool struct which exposes a bunch of optional fields although the external consumer of this struct likely will expect all these fields to be non-null.

Maybe we can turn this into a builder and only make readily built pools (without None values) available for the public.

shared/src/balancer_event_handler.rs
bh2smith: Yes, although I think this will be covered by the fact that we will introduce `WeightedPoolBuilder` and only expose those which are built.

shared/src/balancer_event_handler.rs
bh2smith: We could default to zero. I like this suggestion and will go with it.

shared/src/balancer_event_handler.rs
bh2smith: I found this article on rust builder patterns. Will look at it tomorrow and make appropriate changes!

https://doc.rust-lang.org/1.0.0/style/ownership/builders.html

shared/src/balancer_event_handler.rs
bh2smith: Do you mean intersect?

bh2smith: I will make this change because it is a complimentary equivalent and removes the negation (`!`), but I don't see a `filter_map` being used anywhere here.

shared/src/balancer_event_handler.rs
bh2smith: I have tried a few alternatives to this, but none of them seem to turn out very well.

Example:

```rust
fn new(index: EventIndex, pool_registration: Option<PoolRegistration>,  token_registration: Option<TokenRegistration>) -> Self
```
Would be cool if I could override the new method with an implementation for each of the above options.

shared/src/balancer_event_handler.rs
bh2smith: I have taken the route of renaming the Optional `WeightedPool` to `WeightedPoolBuilder` and implementing `TryInto<WeightedPool> for WeightedPoolBuilder` I hope this is what you were going for.

shared/src/balancer_event_handler.rs
```diff
@@ -7,31 +7,132 @@ use crate::{
...
+    }
+}
+
+/// The BalancerPool struct represents in-memory storage of all deployed Balancer Pools
+#[derive(Debug, Default)]
+pub struct BalancerPools {
+    /// Used for O(1) access to all pool_ids for a given token
+    pools_by_token: HashMap<H160, HashSet<H256>>,
+    /// WeightedPool data for a given PoolId
+    pools: HashMap<H256, WeightedPoolBuilder>,
```
vkgnosis, line 130: I feel it would be cleaner to store in this hash map an `enum { WeightedPoolBuilder, WeightedPool }` or have two separate hash maps for each case.
When you get new events you build up the builder and when you know you have both events and can either change the enum or move the pool from one hash map into the other.
It might also be simpler (less code to write) to make WeightedPoolBuilder only store the two relevant events directly like
```rust
pub struct WeightedPoolBuilder {
    pool_registered: Option<PoolRegistered>,
    tokens_registered: Option<TokensRegistered>,
}
```
and then you know you can create the WeightedPool when both options are Some.

bh2smith: I really like the second suggestion and will certainly go for it. As for keeping the two maps, I could see this also working but it will require some internal changes. Nice thing is that the pending pools (or so I imagine we will call it) will nearly always be empty.

bh2smith: I admit I have always started working on this suggestion and it will really make a lot of thing much easier. It's going to be quite a big refactoring off what o have here so won't finish today, but it's super fun.

shared/src/balancer_event_handler.rs
```diff
@@ -93,31 +302,38 @@ pub struct BalancerEventUpdater(
...
         let deployment_block = match contract.deployment_information() {
             Some(DeploymentInformation::BlockNumber(block_number)) => Some(block_number),
-            Some(DeploymentInformation::TransactionHash(hash)) => Some(
-                contract
+            Some(DeploymentInformation::TransactionHash(hash)) => {
+                match contract
                     .raw_instance()
                     .web3()
                     .block_number_from_tx_hash(hash)
-                    .await?,
```
nlordell, line 104: Is this change because of xDAI? We could also make this continue erroring and then configure the xDAI nodes to not specify `--baseline-liquidiy Balancer`.

shared/src/balancer_event_handler.rs
nlordell: I also really like the alternate `WeightedPoolBuilder` representation suggestion.

One thing I was thinking about is what happens if a re-org happens such that:
- Block 100: `PoolCreated` event
- Block 102: `TokensRegistered` event
- `replace_events(101)` - so we need to "undo half" of a weighted pool

I think for this, it would be easier to have a single hash map and keep track of event blocks separately instead of having a single `block_created`.

nlordell (commented):
Looks very good! Just have some questions about reorg behaviour if happens in a way that only one of the pool events get invalidated.

shared/src/balancer_event_handler.rs
bh2smith: Although remember that both events are emitted in the same transaction always.

shared/src/balancer_event_handler.rs
bh2smith: I am not sure I recall what this was for. Will have to look into this one.

shared/src/balancer_event_handler.rs
nlordell: > Although remember that both events are emitted in the same transaction always. So we don't necessarily have to be that considerate.

Ah, cool, then its a non-issue.

In which case I do agree with @vkgnosis that two maps might be nicer.

shared/src/balancer_event_handler.rs
bh2smith: Actually. I think this was the result of a bad merge.

bh2smith:
Good morning fellow reviewers (cc @fleupold, @nlordell, @vkgnosis). I have made the suggested changes and refactored WeightedPoolBuilder to be much simpler. This comes along with another minor change to the way Builders are upgraded that is specifically after inserting event into the pending pools, we "try_upgrade". For this we return the pool builder from the insertion methods and try_upgrade externally. This was mainly due to the borrow checker not allowing to take mutable twice. If this is an uncommon pattern, please let me know and I will make the necessary changes.

Happy reviewing! 🥂

shared/src/balancer_event_handler.rs
```diff
@@ -7,31 +7,94 @@ use crate::{
...
+pub struct WeightedPoolBuilder {
+    pool_registration: Option<PoolRegistered>,
+    tokens_registration: Option<TokensRegistered>,
+    /// Both Pool and Tokens Registered events are emitted in the same transaction
+    /// https://github.com/balancer-labs/balancer-v2-monorepo/blob/70843e6a61ad11208c1cfabf5cfe15be216ca8d3/pkg/pool-utils/contracts/BasePool.sol#L128-L130
+    /// block_number is only contained in the EventIndex
+    block_created: u64,
+}
+
+impl TryInto<WeightedPool> for WeightedPoolBuilder {
```
fleupold, line 66: @vkgnosis convinced me that while these traits are nice (and a sign one knows rust) they are not always practical especially if the object is not passed around publicly (and therefore it's unlikely an external consumer wants to use the TryInto<WeightedPool> abstraction somewhere).

Here I think an `Option<WeightedPool>` would be a slightly more normal return type (but of course Result with void error works as well).

shared/src/balancer_event_handler.rs
```diff
@@ -54,15 +117,127 @@ impl PoolSpecialization {
...
+                self.pools.insert(pool_id, weighted_pool);
+                for token in token_registration.tokens {
+                    self.pools_by_token
+                        .entry(token)
+                        .or_default()
+                        .insert(pool_id);
+                }
+            }
+            Err(_) => {
+                tracing::info!("Pool Builder not yet upgradable");
```
fleupold, line 143: This should probably not log at info level (which is higher signal than `debug`). I don't think we need a log entry here at all, but if it should be `debug` or `trace`.

shared/src/balancer_event_handler.rs
```diff
@@ -54,15 +117,127 @@ impl PoolSpecialization {
...
 }
 
-#[derive(Default)]
-pub struct BalancerPools {}
-
 impl BalancerPools {
+    fn try_upgrade(&mut self, pool_builder: WeightedPoolBuilder) {
+        match pool_builder.clone().try_into() {
+            Ok(weighted_pool) => {
+                let token_registration = pool_builder.tokens_registration.expect("known to exist");
```
fleupold, line 129: nit: why not use pool_id and tokens from the already built weighted pool directly?

shared/src/balancer_event_handler.rs
```diff
@@ -54,15 +117,127 @@ impl PoolSpecialization {
...
+                BalancerEvent::PoolRegistered(event) => self.insert_pool(index, event),
+                BalancerEvent::TokensRegistered(event) => self.insert_token_data(index, event),
+            };
+            // In the future, when processing TokensDeregistered we may have to downgrade the result.
+            self.try_upgrade(pool_builder)
+        }
+        Ok(())
+    }
+
+    fn insert_pool(
```
fleupold, line 161: It's weird that this method mutates self (by adding the pool builder) and also returns a copy of the builder. Why is that? If for testing purposes, could we inspect self.pending_pools in the test directly? I think this method should return `()`

shared/src/balancer_event_handler.rs
```diff
@@ -78,8 +253,16 @@ impl BalancerPools {
                     ContractEvent::PoolRegistered(event) => {
                         Some(convert_pool_registered(&event, &meta))
                     }
+                    ContractEvent::TokensRegistered(event) => {
+                        Some(convert_tokens_registered(&event, &meta))
+                    }
+                    // ContractEvent::TokensDeregistered(event) => {
```
fleupold, line 259: remove?

orderbook/src/event_updater.rs
```diff
@@ -18,7 +18,7 @@ pub struct EventUpdater(Mutex<EventHandler<DynWeb3, GPv2SettlementContract, Data
 #[async_trait::async_trait]
 impl EventStoring<ContractEvent> for Database {
     async fn replace_events(
-        &self,
+        &mut self,
```
fleupold, line 21: Now that I see it (and we talked about it in the office), this change - while not a lot of lines - could have gone separate to keep the PR more focussed.

shared/src/balancer_event_handler.rs
```diff
@@ -167,3 +350,467 @@ fn convert_pool_registered(
...
+
+    struct BalancerPoolSetup {
+        pool_ids: Vec<H256>,
+        pool_addresses: Vec<H160>,
+        tokens: Vec<H160>,
+        specializations: Vec<PoolSpecialization>,
+        pool_store: BalancerPools,
+    }
+
+    fn dummy_balancer_setup(start_block: usize, end_block: usize) -> BalancerPoolSetup {
```
fleupold, line 646: This test scares me as a future maintainer of the codebase. It looks like there is a lot of undocumented setup which is reused in two test cases. If one of them fails I'm not sure I would understand what exactly happens in the setup and what to change to fix on but not the other test.

To me, a good unit test interface of this component would be the public interface, mainly `replace_events` and `append_events`. You could have some helper methods that create an event struct but the concrete "story" of which events are created in what order should be part of each individual test case. After processing the set of events you could make a bunch of assertions making sure the BalancerPools struct is in the expected state.

fleupold (commented):
Continues to look good. Minor suggestions in the code, but I'm a bit scared of maintaining the unit tests.

shared/src/balancer_event_handler.rs
bh2smith: Yes. I mentioned this in my most recent comment on the PR. Will change.

shared/src/balancer_event_handler.rs
bh2smith: Sure, but I am going to handle this case immediately after in a follow up PR.

shared/src/balancer_event_handler.rs
bh2smith: I was struggling a lot with the borrow checker when refactoring this and this is likely left over from several different approaches.

fleupold: You could clone the weighted pool instead of the builder if needed for the borrow checker (you are already cloning the builder, so it's at the end equivalent). If you have a concrete error, I'm sure we can help out.

shared/src/balancer_event_handler.rs
bh2smith: Right, so I would be happy to return `Option<WeightedPool>` and it makes perfect sense, but then I suppose this falls outside the scope of `TryInto`, perhaps this is what `Into` is for.

shared/src/balancer_event_handler.rs
bh2smith: I have rearranged, reduced, removed  and all-round cleaned up the test suite to be maintainable, accurate and exhaustive without duplication and only the public interfaces as suggested. Please let me know what you think!

shared/src/balancer_event_handler.rs
fleupold: The idea is that these pre-existing traits (Into, TryInto, etc) only really make sense when you pass your object to external consumers which may be operating on the more abstract trait type. In other cases it's not bad to use them, but there is no need to shoehorn something that is different into fitting a pre-existing trait.

E.g. a method that is agnostic to weither its using a PoolBuilder or a Pool or something else that can be turned into a Pool, might look like: `foo(impl Into<WeightedPool>)`. This method would benefit from`impl TryInto<WeightedPool> for WeightedPoolBuilder` because it could use both types. 

But here, you are not passing anything around so you might as well just implement `my_try_into` on `Impl WeightedPoolBuilder` directly. I hope that makes sense.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia4.giphy.com%2Fmedia%2FeqBvQQ5nahKfK%2Fgiphy.gif&token=tada" data-gitmeme-token="tada"></a>

shared/src/balancer_event_handler.rs
```diff
@@ -54,15 +115,112 @@ impl PoolSpecialization {
...
+
+    // All insertions happen in one transaction.
+    fn insert_events(&mut self, events: Vec<(EventIndex, BalancerEvent)>) -> Result<()> {
+        for (index, event) in events {
+            match event {
+                BalancerEvent::PoolRegistered(event) => self.insert_pool(index, event),
+                BalancerEvent::TokensRegistered(event) => self.insert_token_data(index, event),
+            };
+            // In the future, when processing TokensDeregistered we may have to downgrade the result.
+            self.try_upgrade();
```
nlordell, line 149: SHould we just upgrade once after processing all events?

nlordell (commented):
Looks great!

nlordell (approved):

PR #645, opened 2021-05-25
Don't use deny list in driver

vkgnosis:
The deny list is only needed for the order book. To avoid confusion
let's not use it in the driver at all.

This came up on the staging config update to clear the deny listed tokens.

### Test Plan
existing tests, shouldn't affect anything

### DIFF
--- a/solver/src/main.rs
+++ b/solver/src/main.rs
@@ -205,6 +205,7 @@ async fn main() {
         Box::new(pool_aggregator),
         gas_price_estimator.clone(),
         base_tokens.clone(),
+        // Order book already filters bad tokens
         Arc::new(ListBasedDetector::deny_list(Vec::new())),
         native_token_contract.address(),
     ));
--- a/solver/src/main.rs
+++ b/solver/src/main.rs
@@ -200,12 +200,12 @@ async fn main() {
 
     let pool_aggregator = PoolAggregator::from_providers(&pair_providers, &web3).await;
 
-    // TODO - use Filtered-Cached PoolFetchers here too.
+    // TODO: use caching pool fetcher
     let price_estimator = Arc::new(BaselinePriceEstimator::new(
         Box::new(pool_aggregator),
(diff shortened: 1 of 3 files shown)


solver/src/main.rs
```diff
@@ -200,12 +200,11 @@ async fn main() {
 
     let pool_aggregator = PoolAggregator::from_providers(&pair_providers, &web3).await;
 
-    // TODO - use Filtered-Cached PoolFetchers here too.
```
bh2smith, line 203: Are we removing this because its no longer necessary? Or were you going to introduce it here?

bh2smith (approved):

solver/src/main.rs
vkgnosis: As I am saying in the commit message I don't think the filtering part ever made sense. But I guess we might still want to cache this. I'll add that part of the comment back.

solver/src/main.rs
```diff
@@ -200,12 +200,12 @@ async fn main() {
...
+    // TODO: use caching pool fetcher
     let price_estimator = Arc::new(BaselinePriceEstimator::new(
         Box::new(pool_aggregator),
         gas_price_estimator.clone(),
         base_tokens.clone(),
-        Arc::new(ListBasedDetector::deny_list(args.shared.unsupported_tokens)),
```
nlordell, line 208: Do we use `unsupported_tokens` in other places in the driver? If not, we should probably move it from the `shared::Options` to be `orderbook` specific.

vkgnosis: It's not used so I will add that to this PR.

nlordell (approved):

solver/src/main.rs
fleupold: Maybe also add a comment why it's fine to not filter additional tokens (the orderbook does the filtering for us)?

fleupold (approved):

PR #647, opened 2021-05-25
Disable private market makers for 1Inch solver

nlordell:
@fleupold noticed that the failed 1Inch transactions were happened when using private market makers. This PR adds a configuration option to disable 1Inch protocols, with a default value of `PMM1` - the private market maker that seems to be causing issues.

### Test Plan

<details><summary>Make sure that the `protocols` parameter is being respected:</summary>

On `main` the route includes SushiSwap:
```
$ git checkout main && cargo test -p solver -- --ignored --nocapture oneinch_swap_without_amount_checks_and_splitting
...
    protocols: [
        [
            [
                Protocol {
                    name: "WETH",
                    part: 100.0,
                    from_token_address: 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee,
                    to_token_address: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2,
                },
            ],
            [
                Protocol {
                    name: "UNISWAP_V2",
                    part: 100.0,
                    from_token_address: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2,
                    to_token_address: 0xa3bed4e1c75d00fa6f4e5e6922db7261b5e9acd2,
                },
            ],
        ],
        [
            [
                Protocol {
                    name: "WETH",
                    part: 100.0,
                    from_token_address: 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee,
                    to_token_address: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2,
                },
            ],
            [
                Protocol {
                    name: "UNISWAP_V3",
                    part: 33.333333333333336,
                    from_token_address: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2,
                    to_token_address: 0x6b175474e89094c44da98b954eedeac495271d0f,
                },
                Protocol {
                    name: "SUSHI",
                    part: 66.66666666666667,
                    from_token_address: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2,
                    to_token_address: 0x6b175474e89094c44da98b954eedeac495271d0f,
                },
            ],
            [
                Protocol {
                    name: "UNISWAP_V2",
                    part: 100.0,
                    from_token_address: 0x6b175474e89094c44da98b954eedeac495271d0f,
                    to_token_address: 0xa3bed4e1c75d00fa6f4e5e6922db7261b5e9acd2,
                },
            ],
        ],
    ],
...
```

With this branch it doesn't:
```
$ git checkout 1inch-protocols && cargo test -p solver -- --ignored --nocapture oneinch_swap_fully_parameterized
...
    protocols: [
        [
            [
                Protocol {
                    name: "WETH",
                    part: 100.0,
                    from_token_address: 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee,
                    to_token_address: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2,
                },
            ],
            [
                Protocol {
                    name: "UNISWAP_V2",
                    part: 100.0,
                    from_token_address: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2,
                    to_token_address: 0xa3bed4e1c75d00fa6f4e5e6922db7261b5e9acd2,
                },
            ],
        ],
        [
            [
                Protocol {
                    name: "WETH",
                    part: 100.0,
                    from_token_address: 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee,
                    to_token_address: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2,
                },
            ],
            [
                Protocol {
                    name: "UNISWAP_V2",
                    part: 100.0,
                    from_token_address: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2,
                    to_token_address: 0x6b175474e89094c44da98b954eedeac495271d0f,
                },
            ],
            [
                Protocol {
                    name: "UNISWAP_V2",
                    part: 100.0,
                    from_token_address: 0x6b175474e89094c44da98b954eedeac495271d0f,
                    to_token_address: 0xa3bed4e1c75d00fa6f4e5e6922db7261b5e9acd2,
                },
            ],
        ],
    ],
...
```

</details>

The logic for filtering private market makers can also be checked:
```
$ cargo test -p solver -- --ignored filters_disabled_protocols
```


vkgnosis (approved):

bh2smith (approved):
So cool! This may very well silence our alerts channel!

fleupold (approved):
Very clever way of getting all but one protocol (which still works without code change on our end when new protocols are added on their side)

nlordell:
@fleupold can you take another look? I changed it to filter specific protocols, and to have it be passed in as a command line option.

solver/src/oneinch_solver.rs
```diff
@@ -78,27 +97,27 @@ impl OneInchSolver {
...
+            to_token_address: order.buy_token,
+            amount: order.sell_amount,
+            from_address: self.settlement_contract.address(),
+            slippage: Slippage::basis_points(MAX_SLIPPAGE_BPS).unwrap(),
+            protocols,
+            // Disable balance/allowance checks, as the settlement contract
+            // does not hold balances to traded tokens.
+            disable_estimate: Some(true),
+            // Use at most 2 connector tokens
+            complexity_level: Some(Amount::new(2).unwrap()),
```
bh2smith, line 111: I thought there was some discussion about removing this particular parameter somewhere. Maybe I misunderstood.

bh2smith (approved):

solver/src/oneinch_solver.rs
nlordell: https://github.com/gnosis/gp-v2-services/pull/649#issuecomment-848575706 was my reply.

We can remove it, but it would be an orthogonal change to this PR.

solver/src/main.rs
```diff
@@ -120,6 +120,12 @@ struct Arguments {
     )]
     min_order_size_one_inch: U256,
 
+    /// The list of disabled 1Inch protocols. By default, the `PMM1` protocol
+    /// (representing a private market maker) is disabled as it seems to
+    /// produce invalid swaps.
+    #[structopt(long, env, default_value = "PMM1")]
```
fleupold, line 126: Does an empty list work here for an env variable?

solver/src/oneinch_solver.rs
```diff
@@ -62,26 +96,28 @@ impl OneInchSolver {
...
-                // Use at most 2 connector tokens
-                complexity_level: Some(Amount::new(2).unwrap()),
-                // Cap swap gas to 750K.
-                gas_limit: Some(750_000),
-                // Use only 3 main route for cheaper trades.
-                main_route_parts: Some(Amount::new(3).unwrap()),
-                parts: Some(Amount::new(3).unwrap()),
-            })
-            .await?;
+        let protocols = self.supported_protocols().await?;
```
fleupold, line 99: Maybe query this outside settleOrder (since we are already running into API timeout issues and this is only required once per batch).

fleupold (approved):

solver/src/main.rs
nlordell: It does not :sob:

bh2smith: Yea, this has also been a concern for our unsupported tokens list where we have to include a dummy zero address. What could we include here as null in the event we ever open up "PMM1"? If we do not ever plan to exclude "PMM1" from the configuration, maybe its best left as a constant pushed onto this list at runtime.

nlordell: So I noticed that we have other arguments that use `Vec<...>` that in principle also doesn't support empty lists when using `env`. Would you be OK with merging this as is? In theory, if you run `DISABLED_ONEINCH_PROTOCOLS="" solver` right now it would disable the `""` protocol, so basically a no-op.

I see that we already use a similar work around for other parameters in our staging configuration:
```yaml
  unsupportedTokens: '0x0000000000000000000000000000000000000000' # None, work around not supporting empty list.
```

nlordell: https://github.com/gnosis/gp-v2-services/issues/660

PR #648, opened 2021-05-25
Always allow placeholder token *

vkgnosis:
This is so the bad token detector doesn't detect it as bad.

### Test Plan
using this token in route doesn't log that token is bad

### DIFF
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -1,5 +1,8 @@
 use contracts::{GPv2Settlement, WETH9};
-use model::{order::OrderUid, DomainSeparator};
+use model::{
+    order::{OrderUid, BUY_ETH_ADDRESS},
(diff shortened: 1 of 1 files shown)


nlordell (commented):
Is there a reason you prefer to handle it this way than adding handling in `ListBasedDetector`?

vkgnosis:
I did consider it but it feels slightly cleaner to me that the detector doesn't know about special addresses and uses exactly what you give it.

nlordell (approved):

bh2smith (approved):

bh2smith:
I suppose this change should be considered when updating the config for next week's release. We will have to revert or remove the `0xeee..eee` from being manually inserted into the allow list.

fleupold (approved):

PR #651, opened 2021-05-26
[Trivial] inconsistency in runtime arguments

bh2smith:
Not sure how much this matters, but the capitalized name was different with the argument name. Given that neither spellings are actually real words I chose the one I thought looked more appealing.

### DIFF
--- a/solver/src/main.rs
+++ b/solver/src/main.rs
@@ -124,7 +124,7 @@ struct Arguments {
     /// without external liquidity
     #[structopt(
         long,
-        env = "MARKET_MAKEABLE_TOKEN_LIST",
+        env = "MARKET_MAKABLE_TOKEN_LIST",
         default_value = "https://tokens.coingecko.com/uniswap/all.json"
     )]
     market_makable_token_list: String,


vkgnosis (approved):
Before merging should check that this isn't set on staging or if it is fix the env var there too.

solver/src/main.rs
```diff
@@ -124,7 +124,7 @@ struct Arguments {
     /// without external liquidity
     #[structopt(
         long,
-        env = "MARKET_MAKEABLE_TOKEN_LIST",
+        env = "MARKET_MAKABLE_TOKEN_LIST",
```
nlordell, line 127: If you just use `env` it will automatically just use `field_name.to_screaming_snake_case()`.

nlordell (approved):

solver/src/main.rs
nlordell: Yes.

bh2smith: I could get behind this. I do it right here in this PR if everyone else is onboard?

bh2smith:
> Before merging should check that this isn't set on staging or if it is fix the env var there too.

I can't see anywhere in staging that this is used yet in any of our configuration files. Seems like its fine to change and won't actually require anything for next release.

I am now considering removing all the `env = CAPS_STUFF` everywhere right here, but this also shouldn't require any action for deployment.

solver/src/main.rs
nlordell: I prefer it this way, but I'm not sure if the `env = "EXPLICIT_NAME"` was intentional or not.

fleupold (approved):

bh2smith:
Given that this PR has received a lot of approvals and will probably be forgotten. I will leave the changes as is and merge whats here. Will follow up with another PR for removal of `env = CAPS_STUFF` as proposed here.

Issue #638, opened 2021-05-21
Keep token quality cache warm


vkgnosis:
We might want to keep the token quality cache warm so that random requests that happen to come after the timeout don't have to wait longer.

For tokens that we have never seen before we can't do this but we could refresh all existing entries in the cache automatically or if we want to allow the cache to shrink maybe only base tokens or tokens that have been traded recently.

nlordell:
This should be addressed by the recent cache block.

PR #652, opened 2021-05-26
Benchmark API

fleupold:
Related to #638 

This PR is introducing a benchmark for the price/fee estimation API into the repo. I used Criterion, even though it's usually better suited for micro-benchmarking in order to repurpose the rust benchmarking support.

A disadvantage is that I don't seem to be able to do multiple parallel requests (which even with autocannon was problematic as the backend started rate limiting us)

### Test Plan
`cargo bench -p e2e`

### DIFF
--- a/e2e/benches/bench.rs
+++ b/e2e/benches/bench.rs
@@ -1,6 +1,5 @@
 use std::time::Duration;
 
-use criterion::async_executor::FuturesExecutor;
 use criterion::{criterion_group, criterion_main, Criterion};
 use ethcontract::U256;
 use rand::seq::SliceRandom as _;
@@ -20,13 +19,12 @@ pub fn criterion_benchmark(c: &mut Criterion) {
     group
         .measurement_time(Duration::from_secs(300))
         .bench_function("Estimate Price", |b| {
-            b.to_async(FuturesExecutor)
-                .iter(|| estimate_fee_and_price_estimate(&token_list));
+            b.iter(|| estimate_fee_and_price_estimate(&token_list));
         });
     group.finish();
 }
 
-async fn estimate_fee_and_price_estimate(token_list: &TokenList) {
+fn estimate_fee_and_price_estimate(token_list: &TokenList) {
     let mut rng = rand::thread_rng();
     let base_token = token_list
         .all()
--- a/e2e/benches/bench.rs
+++ b/e2e/benches/bench.rs
@@ -1,0 +1,67 @@
+use std::time::Duration;
+
+use criterion::async_executor::FuturesExecutor;
+use criterion::{criterion_group, criterion_main, Criterion};
+use ethcontract::U256;
+use rand::seq::SliceRandom as _;
+use shared::token_list::TokenList;
+use tokio::runtime::Runtime;
+
+const TOKEN_LIST: &str = "https://gateway.ipfs.io/ipns/tokens.uniswap.org";
+const BASE_URL: &str = "https://protocol-mainnet.dev.gnosisdev.com/api/v1";
+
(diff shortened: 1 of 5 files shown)


e2e/benches/bench.rs
```diff
@@ -0,0 +1,67 @@
...
+        "{}/markets/{:#x}-{:#x}/{}/{}",
+        BASE_URL, base_token.address, quote_token.address, order_type, amount
+    );
+    let estimate_fee = format!(
+        "{}/fee?sellToken={:#x}&buyToken={:#x}&kind={}&amount={}",
+        BASE_URL, base_token.address, quote_token.address, order_type, amount
+    );
+
+    for request_url in &[estimate_amount, estimate_fee] {
+        let result = reqwest::blocking::get(request_url).expect("Query failed");
```
vkgnosis, line 54: Why reqwest::blocking if we are in an async function?

vkgnosis (approved):
Not sure how applicable criterion is in this case. Usually I see it used for small local cpu bound tasks and not networking. Especially since in this case it goes to our staging environment instead of what is locally running. Maybe a standalone binary would fit better for this use case and it would give you full control over how the benchmark is executed.

On the other hand if this works and other solutions are more complicated then I'm not against it.

nlordell:
Cleaver use of `criterion` (which as @vkgnosis mentioned is more focused around CPU-bound code performance benchmarking). That being said, it seems to work very nicely here!

e2e/benches/bench.rs
nlordell: I am also curious about this.

fleupold: This was my first approach before I learnt how to use async function in criterion. Will remove

fleupold:
> On the other hand if this works and other solutions are more complicated then I'm not against it.

In particular criterion takes care of plotting and comparing against a baseline which is very nice (and would take a while if implemented separately). I changed the base url to be localhost by default so it can be used on local checkouts. We could make it instantiate everything, but it feels more general if it can also be used to benchmark external instances.

nlordell (approved):
Meant to approve earlier :facepalm:

e2e/benches/bench.rs
```diff
@@ -0,0 +1,66 @@
...
+                request_url,
+                result.status(),
+                result.text().expect("No text")
+            );
+        }
+    }
+}
+
+criterion_group!(benches, criterion_benchmark);
+criterion_main!(benches);
```
bh2smith, line 66: In these two lines we use this term `benches` but I can't really tell what it is apart from maybe the directory in which this file is contained

bh2smith (approved):
Looks really good. Should we also introduce a benchmark test for the 1inch API requests as well (given that they have been timing out)? Not necessary here.

Also, I tried to run the test plan:

```
cargo bench -p e2e
```

Which failed because (I think) the orderbook services weren't running locally.

e2e/benches/bench.rs
fleupold: Criterion.rs requires disabling the benchmark harness generated automatically by rustc. Since we've disabled the default benchmark harness, we need to add our own (benches is the folder and criterion_benchmark is the name of the main benchmark function)

e2e/benches/bench.rs
fleupold: It turns out this would need a tokio executor (instead of a futures one). Criterion supports tokio 1.6 but since we use reqwest also in our shared folder this is using tokio 0.2 (which then causes trouble when I use a tokio 1.6 runtime in the benchmark).

Changing back to all blocking code without async criterion.

PR #653, opened 2021-05-26
Make pool fetching error if communication with node fails

vkgnosis:
The current code skips all pools for which a future has errored. This
makes sense when the contract call reverted becauase the pool is
unusuable in this case.
It does not make sense when communication with the node failed. In that
case it is likely that all futures are errors so the function would
return an empty vector which would then be part of the cache.

In order to test this I had to split up PoolFetcher::fetch.

### Test Plan
existing tests and added new one

### DIFF
--- a/shared/src/bad_token/trace_call.rs
+++ b/shared/src/bad_token/trace_call.rs
@@ -1,13 +1,11 @@
 use super::{BadTokenDetecting, TokenQuality};
-use crate::{amm_pair_provider::AmmPairProvider, trace_many, Web3};
+use crate::{
+    amm_pair_provider::AmmPairProvider, ethcontract_error::EthcontractErrorType, trace_many, Web3,
+};
 use anyhow::{anyhow, bail, ensure, Result};
 use contracts::ERC20;
 use ethcontract::{
-    batch::CallBatch,
-    dyns::DynTransport,
-    errors::{ExecutionError, MethodError},
-    transaction::TransactionBuilder,
-    PrivateKey,
+    batch::CallBatch, dyns::DynTransport, transaction::TransactionBuilder, PrivateKey,
 };
 use model::TokenPair;
 use primitive_types::{H160, U256};
@@ -46,9 +44,11 @@ impl TraceCallDetector {
         let decimals = match instance.decimals().call().await {
             Ok(decimals) => decimals,
             Err(err) => {
-                return match EthcontractError::classify(&err) {
-                    EthcontractError::Node => Err(err.into()),
-                    EthcontractError::Contract => Ok(TokenQuality::bad("failed to get decimals")),
+                return match EthcontractErrorType::classify(&err) {
+                    EthcontractErrorType::Node => Err(err.into()),
+                    EthcontractErrorType::Contract => {
+                        Ok(TokenQuality::bad("failed to get decimals"))
+                    }
                 }
             }
         };
@@ -110,9 +110,9 @@ impl TraceCallDetector {
             let balance = match result {
                 Ok(balance) => balance,
                 Err(err) => {
-                    return match EthcontractError::classify(&err) {
-                        EthcontractError::Node => Err(err.into()),
-                        EthcontractError::Contract => Ok(None),
+                    return match EthcontractErrorType::classify(&err) {
+                        EthcontractErrorType::Node => Err(err.into()),
+                        EthcontractErrorType::Contract => Ok(None),
                     }
                 }
             };
@@ -261,23 +261,6 @@ fn ensure_transaction_ok_and_get_gas(trace: &BlockTrace) -> Result<Result<U256, 
         _ => bail!("no error but also no call result"),
     };
     Ok(Ok(call_result.gas_used))
-}
-
-#[derive(Copy, Clone, Debug)]
-enum EthcontractError {
-    // The error stems from communicating with the node.
-    Node,
-    // Communication was successful but the contract on chain errored.
-    Contract,
-}
-
-impl EthcontractError {
-    fn classify(err: &MethodError) -> Self {
-        match &err.inner {
-            ExecutionError::Web3(_) => Self::Node,
-            _ => Self::Contract,
-        }
-    }
 }
 
 #[cfg(test)]
--- a/shared/src/pool_aggregating.rs
+++ b/shared/src/pool_aggregating.rs
@@ -1,6 +1,7 @@
 use crate::amm_pair_provider::{AmmPairProvider, SushiswapPairProvider, UniswapPairProvider};
 use crate::pool_fetching::{Pool, PoolFetcher, PoolFetching};
 use crate::Web3;
+use anyhow::Result;
 use ethcontract::BlockNumber;
 use model::TokenPair;
 use std::collections::HashSet;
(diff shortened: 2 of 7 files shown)


shared/src/bad_token/trace_call.rs
```diff
@@ -263,23 +263,6 @@ fn ensure_transaction_ok_and_get_gas(trace: &BlockTrace) -> Result<Result<U256,
     Ok(Ok(call_result.gas_used))
 }
 
-#[derive(Copy, Clone, Debug)]
-enum EthcontractError {
```
vkgnosis, line 267: Moved to it's own module so it can be reused.

shared/src/pool_fetching.rs
```diff
@@ -299,57 +315,81 @@ impl PoolFetching for PoolFetcher {
...
+                    .balance_of(pair_address)
+                    .block(at_block.into())
+                    .batch_call(&mut batch);
+                let token1_balance = token1
+                    .balance_of(pair_address)
+                    .block(at_block.into())
+                    .batch_call(&mut batch);
+
+                async move {
+                    // Clippy is wrong about this being eval order depndendent.
```
nlordell, line 333: The lints are probably there because of the `await`s.

I think its 100% fine to ignore in this case.

nlordell (approved):
Great change!

shared/src/pool_aggregating.rs
```diff
@@ -59,15 +60,19 @@ impl PoolAggregator {
...
 impl PoolFetching for PoolAggregator {
-    async fn fetch(&self, token_pairs: HashSet<TokenPair>, at_block: BlockNumber) -> Vec<Pool> {
-        futures::future::join_all(
+    async fn fetch(
+        &self,
+        token_pairs: HashSet<TokenPair>,
+        at_block: BlockNumber,
+    ) -> Result<Vec<Pool>> {
+        // vk: Using try join means if any pool fetcher fails we fail too. Alternatively we could
+        // return the succeeding ones but I feel it is cleaner to forward the error.
```
bh2smith, line 69: I guess this boils down to seeing how reliable the service will be with this configuration.

shared/src/pool_fetching.rs
```diff
@@ -299,57 +315,81 @@ impl PoolFetching for PoolFetcher {
...
+    // Helper for the fold below.
+    // If the result is a node error then return it.
+    // If it is a contract error then skip this pool.
+    macro_rules! forward_error_or_skip {
+        ($result:expr, $acc:ident) => {
+            match $result {
+                Ok(t) => t,
+                Err(err) => match EthcontractErrorType::classify(&err) {
+                    EthcontractErrorType::Node => return Err(err.into()),
+                    EthcontractErrorType::Contract => return Ok($acc),
```
bh2smith, line 371: I see the skip based on the ok, but what is this acc inside the ok?

shared/src/pool_fetching.rs
```diff
@@ -655,13 +720,45 @@ mod tests {
...
             .is_empty());
     }
+
+    #[test]
+    fn pool_fetcher_forwards_node_error() {
+        let results = vec![FetchedPool {
+            reserves: Err(ethcontract_error::testing_node_error()),
+            pair: Default::default(),
+            token0_balance: Ok(1.into()),
+            token1_balance: Ok(1.into()),
```
bh2smith, line 741: Isn't it also possible to get an eth contract error on either of these?

bh2smith (approved):
Looks great. Just a few comments and questions.

shared/src/pool_fetching.rs
vkgnosis: Sure but I don't think it's worth it to repeat this test 3 times

fleupold (approved):
Good call and nice abstraction with `EthcontractErrorType::classify`

shared/src/pool_fetching.rs
```diff
@@ -299,57 +315,81 @@ impl PoolFetching for PoolFetcher {
...
+                Err(err) => match EthcontractErrorType::classify(&err) {
+                    EthcontractErrorType::Node => return Err(err.into()),
+                    EthcontractErrorType::Contract => return Ok($acc),
+                },
+            }
+        };
     }
+
+    results.into_iter().try_fold(Vec::new(), |mut acc, pool| {
+        let reserves = forward_error_or_skip!(pool.reserves, acc);
```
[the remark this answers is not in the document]
vkgnosis: Agree this is a little bit harder to understand. The reason in case it wasn't clear is that I didn't want to have the same fairly big match statement 3 times in the fold. And I needed a macro instead of a function so that it could do the return and continue directly.
I am also not sure how other people feel about it. Will try your suggestion.

shared/src/pool_fetching.rs
```diff
@@ -299,57 +315,81 @@ impl PoolFetching for PoolFetcher {
...
+        let token0_balance = forward_error_or_skip!(pool.token0_balance, acc);
+        let token1_balance = forward_error_or_skip!(pool.token1_balance, acc);
+        if U256::from(reserves.0) <= token0_balance && U256::from(reserves.1) <= token1_balance {
```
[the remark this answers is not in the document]
vkgnosis: Thanks for spotting this. Was a copy paste error.

shared/src/pool_fetching.rs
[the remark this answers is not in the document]
vkgnosis: Changed it to use a normal function and 3 matches. I don't think the matches are too bad so I like this compromise.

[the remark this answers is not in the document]
nlordell: I liked the macro :stuck_out_tongue:. It brought up a bunch of nostalgia from the old `try!()` macro.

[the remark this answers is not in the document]
fleupold: @nlordell starring in: "Tell me you are a rust OG without telling me you are a rust OG"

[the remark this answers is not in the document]
nlordell: :rofl:

shared/src/pool_fetching.rs
```diff
@@ -299,57 +315,93 @@ impl PoolFetching for PoolFetcher {
...
+        handle_results(results)
+    }
+}
+
+struct FetchedPool {
+    pair: TokenPair,
+    reserves: Result<(u128, u128, u32), MethodError>,
+    token0_balance: Result<U256, MethodError>,
+    token1_balance: Result<U256, MethodError>,
+}
```
bh2smith, line 359: I wanted to ask here why we keep track of both `reserves` and `token_i_balance`. Shouldn't the reserves already contain these balances? Also, may I assume that the `u32` in the reserves tuple is the `block_last_updated`? Could we benefit from making use of this instead of all the fancy block logic?

bh2smith: Oh I see in the comment that we 

```
// Fetch ERC20 token balances of the pools to sanity check with reserves
```

However I would think that we could perform the sanity check upon pool fetching and then forget about these values.

vkgnosis: This PR didn't change that logic just refactored it so maybe let's discuss it in an issue or on slack so it's not hidden here.

Issue #650, opened 2021-05-26
Settlement Reverts with `GPv2: order filled`


bh2smith:
We have received quite a few alerts recently with Settlement simulations failing on 

```
contract call reverted with message: Some("GPv2: order filled")
```

This was uncovered last week and discussed in our weekly sync. The problem appears to be with the fact that orders which have been filled in a recent settlement have not yet been picked up by our event listener and thus, not yet flagged as having been matched. In the meantime our solver is already working on the next batch and attempts to use this order again.

It was mentioned that a reasonable avenue to mitigate this is to keep a temporary list of recently filled orders in memory (only as long as a single batch) for which the solver can filter these essentially pending/in flight already matched orders.

One additional piece of insight that would make this "hunch" a bit more apparent is if there were larger batches being settled with several orders for which we receive the alert with multiplicity. Unfortunately at this time our batches are usually limited to one or two trades so we have not seen a massive influx of identical alerts simultaneously.

PR #655, opened 2021-05-26
Make BlockStream poll interval configurable

fleupold:
My benchmarks showed that our prod environment has much larger outliers when it comes to querying the API than staging. After investigating a few isolated instances, it looks like prod is much more sensitive to reorgs and therefore running maintenance (and thus cache eviction) much more frequently than staging.

This PR allows us to configure the polling interval which we use for the new block stream (and sets the default from 1s to 5s)

It may be that #650 becomes more frequent as a consequence since we will be updating our trade table less frequently

### Test Plan
I didn't see significant improvements on the benchmarks (locally my times are much better than on prod however).

### DIFF
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -91,11 +91,11 @@ struct Arguments {
     /// The amount of time a classification of a token into good or bad is valid for.
     #[structopt(
         long,
-        env = "BLOCK_STREAM_POLL_INTERVAL_SECONDS",
+        env,
         default_value = "5",
         parse(try_from_str = shared::arguments::duration_from_seconds),
     )]
-    block_stream_poll_interval: Duration,
+    block_stream_poll_interval_seconds: Duration,
 }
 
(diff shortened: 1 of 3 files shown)


nlordell (approved):
Looks good.

vkgnosis (approved):

orderbook/src/main.rs
```diff
@@ -87,6 +87,15 @@ struct Arguments {
     /// thinks they are bad. Base tokens are automatically allowed.
     #[structopt(long, env = "ALLOWED_TOKENS", use_delimiter = true)]
     pub allowed_tokens: Vec<H160>,
+
+    /// The amount of time a classification of a token into good or bad is valid for.
+    #[structopt(
+        long,
+        env = "BLOCK_STREAM_POLL_INTERVAL_SECONDS",
```
bh2smith, line 94: So we were discussion/considering removing the `env = "LONG_CAPS"` stuff, but here would be an example scenario where it is necessary because its actually different than the var name. Will keep in mind when submitting the PR for this.

bh2smith (approved):

PR #658, opened 2021-05-26
[Easy] Propogate EventHandler Maintenance Failure Logs

### LINKED ISSUE #656: Maintenance Error description poorly propogated
We have a long list of recent errors all with minimal information that can't really be debugged.

```
2021-05-26T13:39:06.906Z ERROR shared::maintenance: failed with: event update error
2021-05-26T13:39:44.212Z ERROR shared::maintenance: failed with: event update error
2021-05-26T13:41:20.116Z ERROR shared::maintenance: failed with: event update error
2021-05-26T13:43:43.687Z ERROR shared::maintenance: failed with: event update error
2021-05-26T13:45:54.731Z ERROR shared::maintenance: failed with: event update error
2021-05-26T13:47:12.868Z  WARN shared::current_block: failed to get current block: failed to get current block
Caused by:
    0: RPC error: Error { code: InternalError, message: "request failed or timed out", data: None }
    1: Internal error: request failed or timed out
2021-05-26T13:48:20.740Z ERROR shared::maintenance: failed with: event update error
2021-05-26T13:49:24.940Z ERROR shared::maintenance: failed with: event update error
2021-05-26T13:49:58.469Z ERROR shared::maintenance: failed with: event update error
```

Given that maintenance is run in several different places, it might be valuable to provide a more descriptive error message.

bh2smith:
Fixes #656
Attempting to propagate a more descriptive error out of failed `EventHandler` maintenance. Currently we are getting:

```
shared::maintenance: failed with: event update error
```

Note that this is also being emitted from the other file that was touched which has been updated a bit to be more specific

So, what we would expect to see now for similar occurrences of an error could be any of 

```
shared::maintenance: Service Maintenance Error: failed to get past events
shared::maintenance: Service Maintenance Error: failed to get next chunk of events
shared::maintenance: Service Maintenance Error: replace - failed to convert events
shared::maintenance: Service Maintenance Error: append - failed to convert events
shared::maintenance: Service Maintenance Error: failed to insert trades
shared::maintenance: Service Maintenance Error: failed to replace trades
```

and possibly more.

### Test Plan
This trait implementation for `Mutex<EventHandler<B, C, S>>` seems like quite an elaborate construct I am not sure if it can easily be mocked. Happy to try if reviewers think it should be.

### DIFF
--- a/shared/src/event_handling.rs
+++ b/shared/src/event_handling.rs
@@ -174,7 +174,6 @@ where
             .await
             .update_events()
             .await
-            .context("event update error")
     }
 }
 
--- a/shared/src/event_handling.rs
+++ b/shared/src/event_handling.rs
@@ -131,7 +131,7 @@ where
         // in one transaction.
         let mut have_deleted_old_events = false;
         while let Some(events_chunk) = events.next().await {
-            let unwrapped_events = events_chunk.context("Failed to get next chunk of events")?;
+            let unwrapped_events = events_chunk.context("failed to get next chunk of events")?;
             if !have_deleted_old_events {
                 self.store
                     .replace_events(unwrapped_events, range.clone())
--- a/shared/src/event_handling.rs
+++ b/shared/src/event_handling.rs
@@ -170,10 +170,7 @@ where
     S: EventStoring<C::Event> + Send + Sync,
 {
     async fn run_maintenance(&self) -> Result<()> {
-        self.lock()
-            .await
-            .update_events()
-            .await
+        self.lock().await.update_events().await
     }
 }
 
--- a/shared/src/maintenance.rs
+++ b/shared/src/maintenance.rs
@@ -19,7 +19,7 @@ impl Maintaining for ServiceMaintenance {
     async fn run_maintenance(&self) -> Result<()> {
         for result in join_all(self.maintainers.iter().map(|m| m.run_maintenance())).await {
             if let Err(err) = result {
-                tracing::error!("Service Maintenance failed with: {}", err);
+                tracing::error!("Service Maintenance Error: {}", err);
             }
         }
         Ok(())
--- a/shared/src/maintenance.rs
+++ b/shared/src/maintenance.rs
@@ -19,7 +19,7 @@ impl Maintaining for ServiceMaintenance {
     async fn run_maintenance(&self) -> Result<()> {
         for result in join_all(self.maintainers.iter().map(|m| m.run_maintenance())).await {
             if let Err(err) = result {
-                tracing::error!("Service Maintenance Error: {}", err);
+                tracing::error!("Service Maintenance Error: {:?}", err);
             }
         }
         Ok(())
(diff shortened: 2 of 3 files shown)


shared/src/event_handling.rs
```diff
@@ -170,11 +170,7 @@ where
     S: EventStoring<C::Event> + Send + Sync,
 {
     async fn run_maintenance(&self) -> Result<()> {
-        self.lock()
-            .await
-            .update_events()
-            .await
-            .context("event update error")
```
bh2smith, line 177: This is the 🔑  change to avoid over generalizing the error message and losing track of what really happened.

shared/src/event_handling.rs
```diff
@@ -170,11 +170,7 @@ where
...
+        self.lock().await.update_events().await
```
bh2smith, line 173: This is the 🔑  change to propagate internal errors better

vkgnosis (approved):

shared/src/maintenance.rs
```diff
@@ -19,7 +19,7 @@ impl Maintaining for ServiceMaintenance {
     async fn run_maintenance(&self) -> Result<()> {
         for result in join_all(self.maintainers.iter().map(|m| m.run_maintenance())).await {
             if let Err(err) = result {
-                tracing::error!("failed with: {}", err);
+                tracing::error!("Service Maintenance Error: {}", err);
```
vkgnosis, line 22: For our anyhow errors it is a little better to use the `{:?}` format string than `{}` here because it sometimes includes more relevant information.

bh2smith:
```suggestion
                tracing::error!("Service Maintenance Error: {:?}", err);
```

shared/src/event_handling.rs
fleupold: Isn't this only removing one layer of context? Wouldn't it without the change below now just say 

`shared::maintenance: failed with: replace - failed to convert events`

which still wouldn't tell you much about what's happening?

shared/src/maintenance.rs
```diff
@@ -19,7 +19,7 @@ impl Maintaining for ServiceMaintenance {
     async fn run_maintenance(&self) -> Result<()> {
         for result in join_all(self.maintainers.iter().map(|m| m.run_maintenance())).await {
             if let Err(err) = result {
-                tracing::error!("failed with: {}", err);
+                tracing::error!("Service Maintenance Error: {:?}", err);
```
fleupold, line 22: As far as my understanding goes, this is the key change, which will make us log the entire error chain instead of just the top level context.

shared/src/event_handling.rs
nlordell: I think the other key change is :point_down: where we use the debug instead of display formatter to print the error, thereby printing the entire error chain and not the most recent context (as you also noted).

shared/src/maintenance.rs
bh2smith: Good to know!

shared/src/event_handling.rs
bh2smith: Looking forward to running into these errors again! Will be ready for them now.

PR #662, opened 2021-05-27
Skip Inflight Orders in Single Run

### LINKED ISSUE #650: Settlement Reverts with `GPv2: order filled`

bh2smith:
As proposed to solver the recent `GPv2: Order Filled` simulation failures, we can keep a set of inflight orders from the current settlement which is then passed onto the next single run. These orders are only ignored for at most one batch whether the transaction passes or fails. 

Not yet sure if this is the ideal/intended approach to this problem given that I noticed single run awaits transaction submission. I am guessing it only awaits confirmation of submission but not for the settlement to be mined.

In brief:

`single_run` accepts a set of inflight order ids which it passes onto the liquidity collector who filters these order by ID out of the orderbook. If a settlement is found, it stores the set of new inflight orders which it then returns to be passed into the next single run.

I am partially expecting to get some feedback along the lines of how this isn't directly connected to any check for transactions having been mined and when to actually clear the inflight orders. Perhaps single run could be executed multiple times without having any of its predecessors mined yet. In this case we would probably rather continue appending to inflight orders and only release when we know a corresponding settlement has been mined.

Once correct, this closes #650

### Test Plan
Plan to introduce a test once we have decided if this implementation is the best approach.

### DIFF
--- a/e2e/tests/onchain_settlement.rs
+++ b/e2e/tests/onchain_settlement.rs
@@ -183,7 +183,7 @@ async fn onchain_settlement(web3: Web3) {
         f64::MAX,
         None,
     );
-    driver.single_run(&HashSet::new()).await.unwrap();
+    driver.single_run().await.unwrap();
 
     // Check matching
     let balance = token_b
@@ -210,5 +210,5 @@ async fn onchain_settlement(web3: Web3) {
     assert!(orders.is_empty());
 
     // Drive again to ensure we can continue solution finding
-    driver.single_run(&HashSet::new()).await.unwrap();
+    driver.single_run().await.unwrap();
 }
--- a/e2e/tests/onchain_settlement.rs
+++ b/e2e/tests/onchain_settlement.rs
@@ -183,7 +183,7 @@ async fn onchain_settlement(web3: Web3) {
         f64::MAX,
         None,
     );
-    driver.single_run().await.unwrap();
+    driver.single_run(&HashSet::new()).await.unwrap();
 
     // Check matching
     let balance = token_b
@@ -210,5 +210,5 @@ async fn onchain_settlement(web3: Web3) {
     assert!(orders.is_empty());
 
     // Drive again to ensure we can continue solution finding
-    driver.single_run().await.unwrap();
+    driver.single_run(&HashSet::new()).await.unwrap();
 }
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -250,8 +250,10 @@ impl Driver {
                 Err(err) => err,
             };
             tracing::error!(
-                "settlement simulation failed right before submission AND for block {} which was current when liquidity was fetched:\n{:?}",
-                current_block_during_liquidity_fetch, error_at_earlier_block
+                "{} settlement simulation failed at submission and block {}:\n{:?}",
+                settlement.name,
+                current_block_during_liquidity_fetch,
+                error_at_earlier_block
             );
             // This is an additional debug log so that the log message doesn't get too long as
             // settlement information is recoverable through tenderly anyway.
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -1,6 +1,7 @@
 pub mod solver_settlements;
 
 use self::solver_settlements::{RatedSettlement, SettlementWithSolver};
+use crate::settlement::Trade;
 use crate::{
     chain,
     liquidity::Liquidity,
@@ -14,11 +15,10 @@ use crate::{
 use anyhow::{anyhow, Context, Error, Result};
 use contracts::GPv2Settlement;
 use ethcontract::errors::MethodError;
-use ethcontract::U256;
 use futures::future::join_all;
 use gas_estimation::GasPriceEstimating;
 use itertools::{Either, Itertools};
-use model::order::{InflightOrders, OrderUid, BUY_ETH_ADDRESS};
+use model::order::BUY_ETH_ADDRESS;
 use num::BigRational;
 use primitive_types::H160;
 use shared::{price_estimate::PriceEstimating, token_list::TokenList, Web3};
@@ -45,7 +45,7 @@ pub struct Driver {
     solver_time_limit: Duration,
     gas_price_cap: f64,
     market_makable_token_list: Option<TokenList>,
-    inflight_orders: InflightOrders,
+    inflight_trades: HashSet<Trade>,
 }
 impl Driver {
     #[allow(clippy::too_many_arguments)]
@@ -84,7 +84,7 @@ impl Driver {
             solver_time_limit,
             gas_price_cap,
             market_makable_token_list,
-            inflight_orders: HashMap::new(),
+            inflight_trades: HashSet::new(),
         }
     }
 
@@ -335,15 +335,14 @@ impl Driver {
             .await
             .context("failed to get current block")?
             .as_u64();
+
+        let inflight_trades = &self.inflight_trades.clone();
+        // Reset the inflight orders - event if get_liquidity fails.
+        self.inflight_trades = HashSet::new();
         let liquidity = self
             .liquidity_collector
-            .get_liquidity(
-                current_block_during_liquidity_fetch.into(),
-                &self.inflight_orders,
-            )
+            .get_liquidity(current_block_during_liquidity_fetch.into(), inflight_trades)
             .await?;
-        // Reset the inflight orders after use.
-        self.inflight_orders = HashMap::new();
 
         let estimated_prices =
             collect_estimated_prices(self.price_estimator.as_ref(), self.native_token, &liquidity)
@@ -421,13 +420,13 @@ impl Driver {
 
             tracing::debug!("winning settlement: {:?}", settlement);
             self.submit_settlement(settlement.clone()).await;
-            self.inflight_orders = settlement
+            self.inflight_trades = settlement
                 .settlement
                 .settlement
                 .trades()
                 .iter()
-                .map(|t| (t.order.order_meta_data.uid, t.executed_amount))
-                .collect::<HashMap<OrderUid, U256>>();
+                .cloned()
+                .collect::<HashSet<Trade>>();
 
             self.report_matched_but_unsettled_orders(
                 &Settlement::from(settlement),
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -1,7 +1,6 @@
 pub mod solver_settlements;
 
 use self::solver_settlements::{RatedSettlement, SettlementWithSolver};
-use crate::settlement::Trade;
 use crate::{
     chain,
     liquidity::Liquidity,
@@ -18,7 +17,7 @@ use ethcontract::errors::MethodError;
 use futures::future::join_all;
 use gas_estimation::GasPriceEstimating;
 use itertools::{Either, Itertools};
-use model::order::BUY_ETH_ADDRESS;
+use model::order::{OrderUid, BUY_ETH_ADDRESS};
 use num::BigRational;
 use primitive_types::H160;
 use shared::{price_estimate::PriceEstimating, token_list::TokenList, Web3};
@@ -45,7 +44,7 @@ pub struct Driver {
     solver_time_limit: Duration,
     gas_price_cap: f64,
     market_makable_token_list: Option<TokenList>,
-    inflight_trades: HashSet<Trade>,
+    inflight_trades: HashSet<OrderUid>,
 }
 impl Driver {
     #[allow(clippy::too_many_arguments)]
@@ -425,8 +424,8 @@ impl Driver {
                 .settlement
                 .trades()
                 .iter()
-                .cloned()
-                .collect::<HashSet<Trade>>();
+                .map(|t| t.order.order_meta_data.uid)
+                .collect::<HashSet<OrderUid>>();
 
             self.report_matched_but_unsettled_orders(
                 &Settlement::from(settlement),
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -14,10 +14,11 @@ use crate::{
 use anyhow::{anyhow, Context, Error, Result};
 use contracts::GPv2Settlement;
 use ethcontract::errors::MethodError;
+use ethcontract::U256;
 use futures::future::join_all;
 use gas_estimation::GasPriceEstimating;
 use itertools::{Either, Itertools};
-use model::order::{OrderUid, BUY_ETH_ADDRESS};
+use model::order::{InflightOrders, OrderUid, BUY_ETH_ADDRESS};
 use num::BigRational;
 use primitive_types::H160;
 use shared::{price_estimate::PriceEstimating, token_list::TokenList, Web3};
@@ -44,7 +45,7 @@ pub struct Driver {
     solver_time_limit: Duration,
     gas_price_cap: f64,
     market_makable_token_list: Option<TokenList>,
-    inflight_orders: HashSet<OrderUid>,
+    inflight_orders: InflightOrders,
 }
 impl Driver {
     #[allow(clippy::too_many_arguments)]
@@ -83,7 +84,7 @@ impl Driver {
             solver_time_limit,
             gas_price_cap,
             market_makable_token_list,
-            inflight_orders: HashSet::new(),
+            inflight_orders: HashMap::new(),
         }
     }
 
@@ -342,7 +343,7 @@ impl Driver {
             )
             .await?;
         // Reset the inflight orders after use.
-        self.inflight_orders = HashSet::new();
+        self.inflight_orders = HashMap::new();
 
         let estimated_prices =
             collect_estimated_prices(self.price_estimator.as_ref(), self.native_token, &liquidity)
@@ -425,8 +426,8 @@ impl Driver {
                 .settlement
                 .trades()
                 .iter()
-                .map(|t| t.order.order_meta_data.uid)
-                .collect::<HashSet<_>>();
+                .map(|t| (t.order.order_meta_data.uid, t.executed_amount))
+                .collect::<HashMap<OrderUid, U256>>();
 
             self.report_matched_but_unsettled_orders(
                 &Settlement::from(settlement),
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -339,7 +339,10 @@ impl Driver {
 
         let liquidity = self
             .liquidity_collector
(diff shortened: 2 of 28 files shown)


solver/src/driver.rs
```diff
@@ -414,6 +421,14 @@ impl Driver {
 
             tracing::debug!("winning settlement: {:?}", settlement);
             self.submit_settlement(settlement.clone()).await;
+            inflight_orders = settlement
+                .settlement
+                .settlement
+                .trades()
+                .iter()
+                .map(|t| t.order.order_meta_data.uid)
```
fleupold, line 429: We should probably only remove fully filled orders?

solver/src/liquidity_collector.rs
```diff
@@ -4,17 +4,23 @@ use ethcontract::BlockNumber;
...
-    pub async fn get_liquidity(&self, at_block: BlockNumber) -> Result<Vec<Liquidity>> {
+    pub async fn get_liquidity(
+        &self,
+        at_block: BlockNumber,
+        excluded_orders: &HashSet<OrderUid>,
+    ) -> Result<Vec<Liquidity>> {
         let limit_orders = self
             .orderbook_api
-            .get_liquidity()
+            .get_liquidity(excluded_orders)
```
fleupold, line 23: nit: I would do the filtering here instead of leaking this logic further down the stack.

solver/src/driver.rs
```diff
@@ -323,7 +327,10 @@ impl Driver {
             .await
     }
 
-    pub async fn single_run(&mut self) -> Result<()> {
+    pub async fn single_run(
+        &mut self,
+        inflight_orders: &HashSet<OrderUid>,
```
fleupold, line 332: I'm not entirely sure the driver of this method should manage state about inflight orders. While this approach makes it very clear that inflight orders only stick around for the next run, I wonder if storing them as a field on the driver instance itself would be more contained.

e2e/tests/onchain_settlement.rs
```diff
@@ -210,5 +210,5 @@ async fn onchain_settlement(web3: Web3) {
     assert!(orders.is_empty());
 
     // Drive again to ensure we can continue solution finding
-    driver.single_run().await.unwrap();
+    driver.single_run(&HashSet::new()).await.unwrap();
```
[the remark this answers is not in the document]
bh2smith: Yes, in particular I would think that this might suffice as a test for these changes, unless you think there should be a standalone test for it.

solver/src/liquidity_collector.rs
bh2smith: I will have to check on this again, but noticed this method converts Orders to LimitOrder which no longer seem to have a UId. Will check back on this.

e2e/tests/onchain_settlement.rs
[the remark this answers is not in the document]
bh2smith: Not needed anymore since we don't give it to a single run.

solver/src/driver.rs
bh2smith: Although if we do this, we still could run into the same problem trying to "over fill an order". I would think that missing at most one single run isn't so bad in this case.

fleupold: > I would think that missing at most one single run isn't so bad in this case

It's essentially censoring this order and could potentially be used in an attack. Match some order partially as you know it won't be in the system next round. We could also store the remaining sell amount for those orders and update if they differ from what the orderbook sent us.

bh2smith: Ok, so I am in the process of making this special consideration for partially fillable orders and stumbled upon an outstanding `TODO` in the code (particularly in `normalize_limit_order` which converts `Order` to `LimitOrder`

http://github.com/gnosis/gp-v2-services/blob/51d10253916e41cba150242cd874237bafb8a745/solver/src/liquidity/offchain_orderbook.rs#L60-L62

we see that it remains to discount previously executed amounts here. The question I would have it whether or not the solver takes the `OrderMetaData` into consideration when working with `LimitOrder`. 

More specifically if it would be a problem for settlement encoding to directly edit the `LimitOrder`'s buy and sell amount fields because we are possibly altering the content of the originally signed order. Of course it would probably make the most sense to directly update the `OrderMetaData` with the relevant inflight execution information, but the solver would at least need to cross check the `LimitOrder` it is given with the corresponding `Order`'s metadata (so not to overfill a partially filled solution). My worry here is that once converted to `LimitOrder` the Order data is dropped.

solver/src/driver.rs
```diff
@@ -334,8 +337,13 @@ impl Driver {
             .as_u64();
         let liquidity = self
             .liquidity_collector
-            .get_liquidity(current_block_during_liquidity_fetch.into())
+            .get_liquidity(
+                current_block_during_liquidity_fetch.into(),
+                &self.inflight_orders,
+            )
             .await?;
+        // Reset the inflight orders after use.
```
fleupold, line 345: This way they won't be reset if e.g. get_liquidity fails and returns early above. Maybe this needs to be written to a temp object and reset first thing in the run_single method?

solver/src/driver.rs
```diff
@@ -414,6 +421,14 @@ impl Driver {
 
             tracing::debug!("winning settlement: {:?}", settlement);
             self.submit_settlement(settlement.clone()).await;
+            self.inflight_orders = settlement
```
fleupold, line 424: Not entirely sure and no strong feeling, but maybe it would be less "leaky" to store the full trades in a hashset?

solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -4,21 +4,41 @@ use crate::settlement::SettlementEncoder;
...
     /// Returns a list of limit orders coming from the offchain orderbook API
-    pub async fn get_liquidity(&self) -> Result<Vec<LimitOrder>> {
+    pub async fn get_liquidity(&self, inflight_orders: &InflightOrders) -> Result<Vec<LimitOrder>> {
         Ok(self
             .get_orders()
             .await
             .context("failed to get orderbook")?
             .into_iter()
-            .map(|order| normalize_limit_order(order, self.get_native_token()))
+            .filter_map(|order| {
```
fleupold, line 21: Can we unit test this logic?

solver/src/driver.rs
bh2smith: Yes, although I might propose `HashMap<OrderUid, Trade>` so that we can easily search for `order_id` containment when flitering. Would this be a reasonable compromise?

bh2smith: Given that we aren't actually handling this at the moment, I suppose we could get away without the `HashMap` for now.

solver/src/driver.rs
```diff
@@ -414,6 +421,14 @@ impl Driver {
...
+                .settlement
+                .settlement
+                .trades()
+                .iter()
+                .map(|t| (t.order.order_meta_data.uid, t.executed_amount))
+                .collect::<HashMap<OrderUid, U256>>();
```
nlordell, line 430: Why not use your fancy type alias?
```suggestion
                .collect::<InflightOrders>();
```

Also, I suspect the compiler would be able to figure this out on its own:
```suggestion
                .collect();
```

solver/src/driver.rs
nlordell: I would argue for a `HashSet<OrderUid>` since we don't really use the executed amount for now.

solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -4,21 +4,41 @@ use crate::settlement::SettlementEncoder;
...
+                        // order.order_meta_data.executed_buy_amount += executed_buy_amount;
+                    } else {
+                        // Fully filled, inflight orders are excluded from consideration
+                        return None;
+                    }
+                }
+                Some(normalize_limit_order(order, self.get_native_token()))
+            })
+            // .filter(|order| !excluded_orders.contains(&order.order_meta_data.uid))
+            // .map(|order| normalize_limit_order(order, self.get_native_token()))
```
nlordell, line 41: I expect these comments can be cleaned up.

solver/src/driver.rs
bh2smith: Actually it is going to be eliminated in favour of @fleupold's suggestion above. However in the future it might reappear as

`InflightTrades - HashMap<OrderUid, Trade>`

solver/src/driver.rs
nlordell: I would actually argue this shouldn't get cleared at all. It already get overwritten here: https://github.com/gnosis/gp-v2-services/blob/e0abb426d00af2fc7c4ff70d41ffbe3974d153fe/solver/src/driver.rs#L424-L430

solver/src/driver.rs
nlordell: I'm not fond of `HashSet<Trade>` because trade equality is brittle, and, at least for me, adds a lot of ambiguity to the code without looking around to fully understand the context. For example, imagine an order:
```rust
let order = Order { partially_fillable: true, ... };
let mut trades = HashSet::new();

trades.insert(Trade {
    order: order.clone(),
    sell_token_index: 0,
    buy_token_index: 1,
    executed_amount: 100.into(),
});
// What do you expect to happen here? Its the same order, so should it replace the existing trade in the
// hash set? The token indices are different, what does this mean and is it even possible here?
trades.insert(Trade {
    order: order.clone(),
    sell_token_index: 1,
    buy_token_index: 0,
    executed_amount: 50.into(),
});
```

solver/src/driver.rs
nlordell: I added comments :point_up:

bh2smith:
Included some TODOs in the code with links to the related issue:

https://github.com/gnosis/gp-v2-services/issues/673

I will also flag this as a good first issue and see if @taminomara might be interested in checking it out!

solver/src/driver.rs
nlordell: The other issue I see with `HashMap<OrderUid, Trade>` is the ambiguity around trading a a partially fillable order multiple times in the same settlement. In that case `HashMap<OrderUid, Trade>` would not be enough as only one of the trades would be recorded.

I would either suggest `Vec<Trade>` and with the `.map().collect()` as you do here, or simply just use `HashSet<OrderUid>` for now, until we actually need trade data when filtering.

solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -28,6 +30,28 @@ struct OrderSettlementHandler {
     order: Order,
 }
 
+fn inflight_order_filter(order: Order, inflight_trades: &HashSet<Trade>) -> Option<Order> {
+    // TODO - could model inflight_trades as HashMap<OrderUid, Trade>
+    // https://github.com/gnosis/gp-v2-services/issues/673
+    // would avoid map-collect and make trade data accessible by id.
+    let inflight_trades_map = inflight_trades
```
nlordell, line 37: I'm not a huge fan of allocating a new map for every order being filtered just to do a more efficient search. You might as well just search linearly thought the `inflight_trades` as it would have the same runtime characteristic (`O(n)`) without having to allocate a map temporarily.

solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -28,6 +30,28 @@ struct OrderSettlementHandler {
...
+        .iter()
+        .map(|t| (t.order.order_meta_data.uid, t.clone()))
+        .collect::<HashMap<OrderUid, Trade>>();
+    if inflight_trades_map.contains_key(&order.order_meta_data.uid) {
+        return if order.order_creation.partially_fillable {
+            // TODO - driver logic for Partially Fillable Orders
+            // https://github.com/gnosis/gp-v2-services/issues/673
+            tracing::error!("filtering partially fillable inflight orders unimplemented.");
+            Some(order)
```
nlordell, line 46: I really like this pragmatic approach we settled on.

solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -312,4 +338,44 @@ pub mod tests {
             },
         );
     }
+
+    #[test]
+    fn inflight_order_filter_() {
```
nlordell, line 343: Nice! Love the test

solver/src/liquidity/offchain_orderbook.rs
bh2smith: I am totally fine with this approach. I was just trying to address Felix's comment (https://github.com/gnosis/gp-v2-services/pull/662#discussion_r642221566). Will revert to just a `HashSet<OrderUid>` for now then.

solver/src/settlement.rs
```diff
@@ -13,7 +13,7 @@ use std::collections::HashMap;
 
 pub use settlement_encoder::SettlementEncoder;
 
-#[derive(Clone, Debug, Default, Eq, PartialEq)]
+#[derive(Clone, Debug, Default, Eq, Hash, PartialEq)]
```
bh2smith, line 16:
```suggestion
#[derive(Clone, Debug, Default, Eq, PartialEq)]
```

solver/src/liquidity/offchain_orderbook.rs
nlordell: I added a bunch of comments as to why I don't like `HashSet<Trade>`: https://github.com/gnosis/gp-v2-services/pull/662#discussion_r642390191

solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -28,6 +33,23 @@ struct OrderSettlementHandler {
     order: Order,
 }
 
+fn inflight_order_filter(order: Order, inflight_trades: &HashSet<OrderUid>) -> Option<Order> {
+    // TODO - could model inflight_trades as HashMap<OrderUid, Trade>
```
nlordell, line 37: I think this might be needed:
```suggestion
    // TODO - could model inflight_trades as HashMap<OrderUid, Vec<Trade>>
```

See comment (https://github.com/gnosis/gp-v2-services/pull/662#discussion_r642393327). This is just a comment, so not really important. Just wanted to point out that partially fillable orders will have some weirdness and hard edge cases when we get around to it.

solver/src/driver.rs
```diff
@@ -334,9 +336,13 @@ impl Driver {
             .await
             .context("failed to get current block")?
             .as_u64();
+
+        let inflight_trades = &self.inflight_trades.clone();
+        // Reset the inflight orders - event if get_liquidity fails.
+        self.inflight_trades = HashSet::new();
```
nlordell, line 342: One cool trick you can use to do this more succinctly is:
```suggestion
        let inflight_trades = std::mem::replace(&mut self.inflight_trades, HashSet::new());
```

Which would avoid the `clone`.

Also, I'd like to bring back the point (https://github.com/gnosis/gp-v2-services/pull/662#discussion_r642385489) that I don't think we ever need to clear the inflight trades early. Specifically I think we should just ever clear this on a successful run. So, from line `407`:
```rust
if let Some(mut settlement) = rated_settlements.clone().into_iter().max_by(...) {
    ...
    self.inflight_trades = ...
} else {
    self.inflight_trades = HashSet::new();
}
```
https://github.com/gnosis/gp-v2-services/blob/c368cd97c1d720fd8c508a3c80d29353cdc1e49d/solver/src/driver.rs#L424-L430

With the way it is currently, imagine a situation where:
1. The inflight trades set gets prematurely cleared
2. There is an intermittent network error that causes the fetching the orders to fail
3. Run loops gets restarted right away
4. Orders are fetched containing inflight orders that don't get filtered.

Additionally, I don't even think its a big deal to never clear this - the worse that can happen is the inflight trades become old (but there aren't any new ones), so it would just try to filter out orders that the orderbook API already knows were filled.

bh2smith: Given all these options I would go for the option of not clearing it, but only updating it when necessary. Will wait for @fleupold's input here.

fleupold: The arguments make sense. I can get behind not clearing or only replacing on successful submission.

solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -28,6 +33,23 @@ struct OrderSettlementHandler {
...
 }
 
+fn inflight_order_filter(order: Order, inflight_trades: &HashSet<OrderUid>) -> Option<Order> {
+    // TODO - could model inflight_trades as HashMap<OrderUid, Vec<Trade>>
+    // https://github.com/gnosis/gp-v2-services/issues/673
+    if inflight_trades.contains(&order.order_meta_data.uid) {
+        return if order.order_creation.partially_fillable {
+            // TODO - driver logic for Partially Fillable Orders
+            // https://github.com/gnosis/gp-v2-services/issues/673
+            tracing::error!("filtering partially fillable inflight orders unimplemented.");
```
fleupold, line 43: nano nit: I wouldn't log an error in this case. The worst case that could happen is that our simulation fails due to order filled in which case it would already notify us.

solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -28,6 +33,23 @@ struct OrderSettlementHandler {
     order: Order,
 }
 
+fn inflight_order_filter(order: Order, inflight_trades: &HashSet<OrderUid>) -> Option<Order> {
```
fleupold, line 36: The signature is already designed to later on mutate partially fillable orders? Otherwise a simple bool filter would suffice (but it makes sense to already set this up like that).

fleupold (approved):
LGTM

solver/src/driver.rs
bh2smith: I will go with that then!

solver/src/liquidity/offchain_orderbook.rs
bh2smith: Yes, this was the purpose of the function signature. Good eye!

nlordell (approved):

PR #663, opened 2021-05-27
Instrument Web3 Transport to record round trip times

fleupold:
Adding another implementation of Transport and BatchTransport which reports the roundtrip time to prometheus. This should 
help us verify if our bad response times are due to node slowness or not.

Note, that something like a DynTransport would probably be nice so that our components can be agnostic as to whether they have a LoggingTransport a MetricTransport or a mix/match. However, this doesn't seem trivial and was out of scope for this more important change.

### Test Plan
Run orderbook locally, make a fee request, then check metrics under `http://localhost:9586/metrics`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -471,6 +471,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "derivative"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fcc3dd5e9e9c0b295d6e1e4d811fb6f157d5ffd784b8d202fc62eac8035a770b"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
 name = "derive_more"
 version = "0.99.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2451,6 +2462,7 @@ dependencies = [
  "async-trait",
  "atty",
  "contracts",
+ "derivative",
  "ethcontract",
  "futures",
  "gas-estimation",
--- a/e2e/tests/ganache.rs
+++ b/e2e/tests/ganache.rs
@@ -1,7 +1,7 @@
 use ethcontract::futures::FutureExt;
-use ethcontract::{Http, U256};
+use ethcontract::U256;
 use lazy_static::lazy_static;
-use shared::{transport::LoggingTransport, Web3};
+use shared::{transport::create_test_transport, Web3};
 use std::{
     fmt::Debug,
     future::Future,
@@ -33,7 +33,7 @@ where
     // it but rather in the locked state.
     let _lock = GANACHE_MUTEX.lock();
 
-    let http = LoggingTransport::new(Http::new(NODE_HOST).expect("transport failure"));
+    let http = create_test_transport(NODE_HOST);
     let web3 = Web3::new(http);
     let resetter = Resetter::new(&web3).await;
 
--- a/orderbook/src/account_balances.rs
+++ b/orderbook/src/account_balances.rs
@@ -189,16 +189,14 @@ fn is_empty_or_truthy(bytes: &[u8]) -> bool {
 mod tests {
     use super::*;
     use contracts::ERC20Mintable;
-    use ethcontract::{prelude::Account, Http};
+    use ethcontract::prelude::Account;
     use hex_literal::hex;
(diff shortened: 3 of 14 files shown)


vkgnosis (approved):

bh2smith (approved):
Tried it out. Looks good. I made a single fee request and got the following response:


```
# HELP gp_v2_api_requests API Request durations labelled by route and response status code
# TYPE gp_v2_api_requests histogram
gp_v2_api_requests_bucket{request_type="fee_info",response="200",le="0.005"} 0
gp_v2_api_requests_bucket{request_type="fee_info",response="200",le="0.01"} 0
gp_v2_api_requests_bucket{request_type="fee_info",response="200",le="0.025"} 0
gp_v2_api_requests_bucket{request_type="fee_info",response="200",le="0.05"} 0
gp_v2_api_requests_bucket{request_type="fee_info",response="200",le="0.1"} 0
gp_v2_api_requests_bucket{request_type="fee_info",response="200",le="0.25"} 0
gp_v2_api_requests_bucket{request_type="fee_info",response="200",le="0.5"} 1
gp_v2_api_requests_bucket{request_type="fee_info",response="200",le="1"} 1
gp_v2_api_requests_bucket{request_type="fee_info",response="200",le="2.5"} 1
gp_v2_api_requests_bucket{request_type="fee_info",response="200",le="5"} 1
gp_v2_api_requests_bucket{request_type="fee_info",response="200",le="10"} 1
gp_v2_api_requests_bucket{request_type="fee_info",response="200",le="+Inf"} 1
gp_v2_api_requests_sum{request_type="fee_info",response="200"} 0.441046325
gp_v2_api_requests_count{request_type="fee_info",response="200"} 1
```

Just trying to understand the 1's and 0's here in the buckets.

it appears the one request took 0.441046325 (seconds?) and now there is a 1 in every bucket greater than this. Is that what we want or should it be the compliment of this?

vkgnosis:
That is correct because the buckets are cumulative. See https://prometheus.io/docs/concepts/metric_types/#histogram for how Prometheus histograms work.

bh2smith:
I'm aware that the buckets are cumulative, but I expected the count to be one for the times below and not above. If this is what it's supposed to do then great.

fleupold:
> I'm aware that the buckets are cumulative, but I expected the count to be one for the times below and not above. If this is what it's supposed to do then great.

"le" stands for "less or equal" so it is indeed counting the number of requests that took less than each bucket. Your example query took 441ms so all buckets 500ms and larger have a count of 1 (because 1 query took less than 10s e.g.) and all the ones below have a count of 0 (because 0 queries took less than 100ms).

bh2smith:
So then the measurement should be made with "ge" because I can't see how it is meaningful for know my response time was between 0.5 and 10 seconds. What we should be counting here is how many round trips took longer than each of these buckets. 

Change my mind.

bh2smith:
Let me elaborate a little more to be perfectly clear. If we use this measurement as it is, than a majority of all round trips will wind up being counted in the 10s bucket (Maybe even all the buckets). Now when, on occasion, there is a request that actually take a long time (say 9 seconds) it will be counted in a bucket that is already full from everything else and we won't be able to tell the difference or have any reason to suspect something is wrong. 

Of course if we construct this histogram the other way around we will see a majority of all requests with lower times and the outliers (I.e longer round trips) will be clearly visible in their appropriate buckets.

vkgnosis:
I believe either of way of measuring this (le vs ge) is equivalent (you can turn one into the other). It won't matter for analysis. Regardless of what is more natural, this is just how the prometheus api is defined. Histogram is a standard data type that happens to work this way. It's not something we can change.
The way you know a request took long is by comparing buckets. So if we 1000 counts in the 5 second bucket and then 1001 in the 10 second bucket we know 1 request took 10 seconds.

nlordell:
Right, to get the value of the measurement as if:
> So then the measurement should be made with "ge"

You could, for example, to simulate `be="0.1"`:
```
gp_v2_api_requests_count{request_type="fee_info",response="200"} - gp_v2_api_requests_bucket{request_type="fee_info",response="200",le="0.1"}
```

So, as @e00E metioned, there is no loss of information, its just exposed its counters in a arguably counter-intuitive way. Note that the Grafana dashboard lets you do these kinds of transformations, so we aren't tied to presenting the information exactly as the metrics counters keep track of them.

bh2smith:
Thanks for clarifying, I suppose this is alright.

PR #666, opened 2021-05-28
Add transaction monitoring service

fedgiac:
Introduces a new crate with a binary that can be used to monitor how many failures there have been in the GPv2 contracts.
Later, this crate could be extended to monitor the behavior of third-party solvers.

### Test Plan

Try the service with test parameters:
```
export NODE_URL="https://mainnet.infura.io/v3/${INFURA_KEY}"
export REVERTED_TRANSACTIONS_ALERT_THRESHOLD=0
export REVERTED_TRANSACTIONS_TIME_WINDOW=60
export LOG_FILTER=info
cargo run --bin monitor
```

### DIFF
--- a/monitor/src/main.rs
+++ b/monitor/src/main.rs
@@ -138,12 +138,13 @@ async fn get_blocks_from(
                 tracing::error!("block filter did not return a valid block: {}", e);
                 bail!("invalid block filter");
             }
-            let response = eth.block_with_txs(BlockId::Hash(block_hash?)).await?;
-            if response.is_none() {
-                tracing::warn!("failed to retrieve a block returned by the new block filter");
-                bail!("invalid block from block filter")
+            match eth.block_with_txs(BlockId::Hash(block_hash?)).await? {
+                Some(block) => Ok(block),
+                None => {
+                    tracing::warn!("failed to retrieve a block returned by the new block filter");
+                    bail!("invalid block from block filter")
+                }
             }
-            Ok(response.unwrap())
         }
     }))
 }
--- a/monitor/src/main.rs
+++ b/monitor/src/main.rs
@@ -176,10 +176,8 @@ impl SettlementFailureCounter {
 
     fn prune(&mut self) {
         let threshold = Utc::now() - self.time_window;
-        self.settlements = std::mem::take(&mut self.settlements)
-            .into_iter()
-            .filter(|Settlement { time, .. }| time >= &threshold)
-            .collect();
+        self.settlements
+            .retain(|Settlement { time, .. }| time >= &threshold);
     }
 
     fn push_settlement(&mut self, time: DateTime<Utc>, success: bool) {
--- a/monitor/src/main.rs
+++ b/monitor/src/main.rs
@@ -1,7 +1,7 @@
 use anyhow::bail;
 use chrono::{offset::Utc, DateTime, TimeZone};
 use contracts::GPv2Settlement;
-use futures::{stream, Stream, StreamExt};
+use futures::{pin_mut, stream, Stream, StreamExt};
 use shared::{transport::LoggingTransport, Web3};
 use std::time::Duration;
 use structopt::StructOpt;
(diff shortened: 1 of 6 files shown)


monitor/src/main.rs
```diff
@@ -0,0 +1,289 @@
...
+    web3: &Web3,
+) -> impl Stream<Item = Option<TransactionReceipt>> + '_ {
+    let transactions = block
+        .transactions
+        .into_iter()
+        .filter(move |tx| tx.to == Some(target));
+    stream::unfold(transactions, move |mut txs| {
+        let eth = web3.eth();
+        async move {
+            if let Some(tx) = txs.next() {
```
vkgnosis, line 65: Could probably remove else branch and use `let tx = txs.next()?`.

monitor/src/main.rs
```diff
@@ -0,0 +1,289 @@
...
+    })
+}
+
+// Note: this function does not manage uncle blocks. It could be that more than
+// one block is returned with the same number. Initialization may be subject to
+// race conditions where a few blocks might be skipped or duplicated.
+async fn get_blocks_from(
+    start: DateTime<Utc>,
+    web3: &Web3,
+) -> impl Stream<Item = anyhow::Result<Block<Transaction>>> + '_ {
```
vkgnosis, line 87: Haven't looked into the body in detail but maybe https://docs.rs/futures/0.3.15/futures/stream/trait.TryStreamExt.html trait functions could be used

monitor/src/main.rs
```diff
@@ -0,0 +1,289 @@
...
+
+struct Count {
+    total: usize,
+    failures: usize,
+}
+
+struct SettlementFailureCounter {
+    // Note: settlements are not expected to be sorted by time because of uncle
+    // blocks.
+    settlements: Vec<Settlement>,
```
vkgnosis, line 168: If this makes other logic easier you could enforce the invariant that the vector is always sorted. To uphold the invariant you can use the binary_search functions https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search .

monitor/src/main.rs
```diff
@@ -0,0 +1,289 @@
...
+    start: DateTime<Utc>,
+    web3: &Web3,
+) -> impl Stream<Item = anyhow::Result<Block<Transaction>>> + '_ {
+    let eth = web3.eth();
+    let mut block_number = BlockNumber::Latest;
+    let mut blocks = Vec::new();
+    let filter = EthFilter::new(web3.transport())
+        .create_blocks_filter()
+        .await
+        .expect("unable to register event listener for new blocks");
```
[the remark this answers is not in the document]
fedgiac: I think I encountered this issue with the mainnet staging node, how can I set this up so that it works with load-balanced nodes?

vkgnosis (approved):

monitor/src/main.rs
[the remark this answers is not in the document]
vkgnosis: In general there is just no way to make it reliable. At any time you might get a new node that doesn't have the filter. The workaround is not use filters at all. You could do it like `CurrentBlockStream` and poll the current block yourself. Then based on the numbers you get you know which new blocks have appeared.

monitor/src/main.rs
```diff
@@ -0,0 +1,289 @@
...
+
+    let start =
+        Utc::now() - chrono::Duration::from_std(args.reverted_transaction_time_window).unwrap();
+    // Note: filtering out node error means that we might be ignoring failures.
+    let blocks = get_blocks_from(start, &web3)
+        .await
+        .filter_map(|x| async { x.ok() });
+
+    let mut txs = Box::pin(
+        blocks
```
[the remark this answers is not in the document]
fedgiac: What's the difference between these three? I don't see much in the docs so I wouldn't know which to use when.

monitor/src/main.rs
fedgiac: Having it sorted is nice because `settlements` could be a `VecDeque` and pruning could be done by popping from the left until all old transactions are removed, instead of traversing each time. In terms of readability I think it's better how it is now, in terms of efficiency I don't know, but anyway the numbers are very small, so I think it might be ok to leave it like this.

vkgnosis: Agree, I wouldn't worry about efficiency. Just about code readability.

monitor/src/main.rs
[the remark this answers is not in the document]
vkgnosis: pin_mut! is the most efficient because it does not allocate anything.
The reason Boxing helps in the first place is because the stream is `!Unpin` so it must be pinned before use and boxing makes it Unpin so it no longer has to be pinned.
Box::pin is the same thing as StreamExt::boxed.

fleupold:
As discussed via slack, for things for which we already have prometheus metrics (e.g. number of successfully submitted vs failed transactions), we can create an alert just based on this metric (cf. [staging gitlab](https://gitlab.gnosisdev.com/devops/gnosis-staging/blob/872b6963ec55ab2acde2235c4d4ed53[…]onnet/kube-prometheus/rules-custom/dfusion.rules.libsonnet)).

We don't need to run a separate service.

fedgiac:
Thanks Valentin for the review anyway!

PR #667, opened 2021-05-28
Split up PoolFetching trait

vkgnosis:
When working on the cache I found that I can simplify the implementation
significantly by making use of the fact that we only care about the
latest state of the pools in the order book.
To encode this in the type system I introduce a new trait
LatestPoolFetching that works just like PoolFetching but always operates
on the latest block.

This trait will be used in the following PR.

### Test Plan
not needed

### DIFF
--- a/shared/src/pool_fetching.rs
+++ b/shared/src/pool_fetching.rs
@@ -33,6 +33,22 @@ pub trait PoolFetching: Send + Sync {
     ) -> Result<Vec<Pool>>;
 }
 
+#[async_trait::async_trait]
+pub trait LatestPoolFetching: Send + Sync {
+    /// Like PoolFetching::fetch but only for the current block.
+    async fn fetch_latest(&self, token_pairs: HashSet<TokenPair>) -> Result<Vec<Pool>>;
+}
+
+#[async_trait::async_trait]
+impl<T> LatestPoolFetching for T
+where
+    T: PoolFetching,
+{
+    async fn fetch_latest(&self, token_pairs: HashSet<TokenPair>) -> Result<Vec<Pool>> {
+        self.fetch(token_pairs, BlockNumber::Latest).await
+    }
+}
+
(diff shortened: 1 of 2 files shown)


bh2smith (approved):
Very cool. Shouldn't this change also remove a bunch of places where we had been using classical pool fetching and passing around the current block everywhere? Or were we already not doing this?

vkgnosis:
What places do you mean? This PR already makes the price estimator use the new trait. I don't think there are other places  that only use the latest block.

fleupold (commented):
I was hoping to reuse the same trait but either specify BlockNumber::Latest or even make block number an Option to say "I don't care about the concrete block, give me whatever is cheapest to fetch" - this way we could even reuse an e.g. 3 block old estimate instead even if the token was evicted from LRU.

Can you explain a bit more what complicates making the existing caching pool fetcher run maintenance by itself? I would expect we create another LRU instance to track token pairs and the periodically fetch them from inner at block "latest".

Otherwise we have to maintain two different implementations of the cache and I could see how both could be useful in the same process.  E.g. the price estimate in the solver would be fine to use more relaxed pool state (for estimating token value) whereas the liquidity needs to be fetched on an exact block.

vkgnosis:
> The reason for the old caching pool fetcher, which was only recently introduced and not yet used was so that the driver can "pin" a specific block to make sure we are not querying different information in the same runloop (while allowing it to be cached). 

I misinterpreted that. I thought we needed block specific info for consistent simulations in driver and the cache for the order book which lead to these things combined being more complicated than needed.

Using "latest" as a place holder for best effort recent pools is a good point too. Turning latest into a real block number while staying consistent with automatic updating and not blocking api users for longer than necessary was something I felt got more complicated. But this way this can probably stay simple.

I am going to attempt to reconcile this again together.

vkgnosis:
https://github.com/gnosis/gp-v2-services/pull/670

PR #668, opened 2021-05-28
Use automatically updating pool cache in order book

vkgnosis:
To minimze latency for end user requests that need to fetch pools like
getting the fee and estimating prices we want to use a pool cache that
always contains and updates the last N recently used pairs.

The previous cache that takes block numbers into account has been deleted.
It was already not used in the driver. If we do want to eventually have a cache in
the driver we can either extend this implementation or make a new one that is
optimized for the driver.

### Test Plan
New unit test + manual testing (TODO)


bh2smith (approved):
Wow you are on 🔥 today. Call 911, 🚨 🚒 🧯

fleupold (commented):
The reason for the old caching pool fetcher, which was only recently introduced and not yet used was so that the driver can "pin" a specific block to make sure we are not querying different information in the same runloop (while allowing it to be cached). I don't think it should be removed (e.g. price estimation and liquidity fetching should share the cache instead of querying data twice).

Will comment on the other PR to discuss if we should have two fundamentally different cache implementations.

vkgnosis:
Closing but not deleting the branch yet because likely superseded by https://github.com/gnosis/gp-v2-services/pull/670.

PR #670, opened 2021-05-28
Driver friendly pool cache

vkgnosis:
Replaces the existing cache. Pool fetching takes a significant amount of
time in both the order book and the driver. This cache hopes to reduce
this by caching recently used tokens automatically when the current
block changes, and keeping previous results cached for a number of
blocks.

Follow ups:
- Change the block number parameter in the pool fetching trait to make it clear whether you get a best effort or exact result.
- Maybe make maintenance trait pass block number around.
- Enable it in the driver.
 
### Test Plan
new unit tests, some manual testing

### DIFF
--- a/e2e/tests/services.rs
+++ b/e2e/tests/services.rs
@@ -138,6 +138,7 @@ impl OrderbookServices {
         let pool_fetcher = PoolCache::new(
             NonZeroU64::new(10).unwrap(),
             20,
+            4,
             Box::new(PoolFetcher {
                 pair_provider,
                 web3: web3.clone(),
--- a/e2e/tests/services.rs
+++ b/e2e/tests/services.rs
@@ -10,16 +10,12 @@ use orderbook::{
 };
 use prometheus::Registry;
 use shared::{
-    amm_pair_provider::UniswapPairProvider,
-    bad_token::list_based::ListBasedDetector,
-    current_block::current_block_stream,
-    maintenance::ServiceMaintenance,
-    pool_fetching::{CachedPoolFetcher, PoolFetcher},
-    price_estimate::BaselinePriceEstimator,
-    Web3,
+    amm_pair_provider::UniswapPairProvider, bad_token::list_based::ListBasedDetector,
+    current_block::current_block_stream, maintenance::ServiceMaintenance, pool_cache::PoolCache,
+    pool_fetching::PoolFetcher, price_estimate::BaselinePriceEstimator, Web3,
 };
 use solver::orderbook::OrderBookApi;
-use std::{collections::HashSet, str::FromStr, sync::Arc, time::Duration};
+use std::{collections::HashSet, num::NonZeroU64, str::FromStr, sync::Arc, time::Duration};
 
 pub const API_HOST: &str = "http://127.0.0.1:8080";
 
@@ -132,24 +128,27 @@ impl OrderbookServices {
         let db = Database::new("postgresql://").unwrap();
         db.clear().await.unwrap();
         let event_updater = Arc::new(EventUpdater::new(gpv2.settlement.clone(), db.clone(), None));
-        let current_block_stream = current_block_stream(web3.clone(), Duration::from_secs(1))
-            .await
-            .unwrap();
         let pair_provider = Arc::new(UniswapPairProvider {
             factory: uniswap_factory.clone(),
             chain_id,
         });
-        let pool_fetcher = CachedPoolFetcher::new(
+        let current_block_stream = current_block_stream(web3.clone(), Duration::from_secs(5))
+            .await
+            .unwrap();
+        let pool_fetcher = PoolCache::new(
+            NonZeroU64::new(10).unwrap(),
+            20,
             Box::new(PoolFetcher {
                 pair_provider,
                 web3: web3.clone(),
             }),
             current_block_stream,
-        );
+        )
+        .unwrap();
         let gas_estimator = Arc::new(web3.clone());
         let bad_token_detector = Arc::new(ListBasedDetector::deny_list(Vec::new()));
         let price_estimator = Arc::new(BaselinePriceEstimator::new(
-            Box::new(pool_fetcher),
+            Arc::new(pool_fetcher),
             gas_estimator.clone(),
             HashSet::new(),
             bad_token_detector.clone(),
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -218,6 +218,7 @@ async fn main() {
         PoolCache::new(
             args.shared.pool_cache_blocks,
             args.shared.pool_cache_lru_size,
+            args.shared.pool_cache_maximum_recent_block_age,
             Box::new(pool_aggregator),
             current_block_stream.clone(),
         )
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -23,7 +23,7 @@ use shared::{
     current_block::current_block_stream,
     maintenance::ServiceMaintenance,
     pool_aggregating::{self, PoolAggregator},
-    pool_fetching::CachedPoolFetcher,
+    pool_cache::PoolCache,
     price_estimate::BaselinePriceEstimator,
     transport::create_instrumented_transport,
 };
@@ -214,11 +214,18 @@ async fn main() {
             .unwrap();
 
     let pool_aggregator = PoolAggregator::from_providers(&pair_providers, &web3).await;
-    let pool_fetcher =
-        CachedPoolFetcher::new(Box::new(pool_aggregator), current_block_stream.clone());
+    let pool_fetcher = Arc::new(
+        PoolCache::new(
+            args.shared.pool_cache_blocks,
+            args.shared.pool_cache_lru_size,
+            Box::new(pool_aggregator),
+            current_block_stream.clone(),
+        )
+        .expect("failed to create pool cache"),
+    );
 
     let price_estimator = Arc::new(BaselinePriceEstimator::new(
-        Box::new(pool_fetcher),
+        pool_fetcher.clone(),
         gas_price_estimator.clone(),
         base_tokens,
         bad_token_detector.clone(),
@@ -246,6 +253,7 @@ async fn main() {
             orderbook.clone(),
             Arc::new(database.clone()),
             Arc::new(event_updater),
+            pool_fetcher,
         ],
     };
     check_database_connection(orderbook.as_ref()).await;
--- a/shared/src/arguments.rs
+++ b/shared/src/arguments.rs
@@ -1,7 +1,10 @@
 //! Contains command line arguments and related helpers that are shared between the binaries.
 use crate::{gas_price_estimation::GasEstimatorType, pool_aggregating::BaselineSources};
 use ethcontract::{H160, U256};
-use std::{num::ParseFloatError, time::Duration};
+use std::{
+    num::{NonZeroU64, ParseFloatError},
+    time::Duration,
+};
 use url::Url;
 
 #[derive(Debug, structopt::StructOpt)]
@@ -61,6 +64,14 @@ pub struct Arguments {
         use_delimiter = true
     )]
     pub baseline_sources: Vec<BaselineSources>,
+
+    /// The number of blocks kept in the pool cache.
+    #[structopt(long, env, default_value = "10")]
+    pub pool_cache_blocks: NonZeroU64,
+
+    /// The number of pairs that are automatically updated in the pool cache.
+    #[structopt(long, env, default_value = "200")]
+    pub pool_cache_lru_size: usize,
 }
 
 pub fn duration_from_seconds(s: &str) -> Result<Duration, ParseFloatError> {
--- a/shared/src/arguments.rs
+++ b/shared/src/arguments.rs
@@ -72,6 +72,10 @@ pub struct Arguments {
     /// The number of pairs that are automatically updated in the pool cache.
     #[structopt(long, env, default_value = "200")]
     pub pool_cache_lru_size: usize,
+
+    /// The number of pairs that are automatically updated in the pool cache.
+    #[structopt(long, env, default_value = "4")]
+    pub pool_cache_maximum_recent_block_age: u64,
 }
 
 pub fn duration_from_seconds(s: &str) -> Result<Duration, ParseFloatError> {
--- a/shared/src/pool_cache.rs
+++ b/shared/src/pool_cache.rs
@@ -34,26 +34,36 @@ pub struct PoolCache {
 // When pools are requested we mark all those pools as recently used which potentially evicts other
 // pairs from the pair lru cache. Cache misses are fetched and inserted into the cache.
 // Then when the automatic update runs the next time, we request and cache all recently used pairs.
-// For the order book we only care about the "recent" state of the pools. So we can return any
+// For some consumers we only care about the "recent" state of the pools. So we can return any
 // result from the cache even if it comes from previous blocks.
-// On the other hand for the driver we need to get the pool at exact blocks which is why we keep a
-// cache of previous blocks in the first place as we could simplify this module if it was only used
-// by the order book.
+// On the other hand for others we need to get the pool at exact blocks which is why we keep a cache
+// of previous blocks in the first place as we could simplify this module if it was only used by
+// by the former.
 
 impl PoolCache {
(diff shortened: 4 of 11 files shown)


e2e/tests/services.rs
```diff
@@ -131,24 +128,18 @@ impl OrderbookServices {
...
             chain_id,
         });
-        let pool_fetcher = CachedPoolFetcher::new(
-            Box::new(PoolFetcher {
-                pair_provider,
-                web3: web3.clone(),
-            }),
-            current_block_stream,
-        );
+        let pool_fetcher = OrderbookPoolFetcher::new(Box::new(PoolFetcher {
```
nlordell, line 135: Is this maybe missing parameters? Otherwise I'm a little confused.

Also, does the pool fetcher need to be added to the list of `maintiners` in the `ServiceMaintenance` instance :point_down:?

orderbook/src/orderbook_pool_fetcher.rs
```diff
@@ -0,0 +1,200 @@
...
+            .fetch(
+                cache_misses.clone(),
+                BlockNumber::Number(cache_miss_block.into()),
+            )
+            .await?;
+
+        {
+            let mut mutexed = self.mutexed.lock().unwrap();
+            mutexed.insert(cache_miss_block, pairs.iter().copied(), &uncached_pools);
+            mutexed.mark_used(pairs.iter().copied());
```
fleupold, line 119: Could this not be an implementation detail of the get method?

e2e/tests/services.rs
vkgnosis: I didn't finish integrating it fully which is why it's a draft. Just wanted to compare the different cache implementations.

orderbook/src/orderbook_pool_fetcher.rs
```diff
@@ -0,0 +1,200 @@
...
+            cached_most_recently_at_block: HashMap::new(),
+            pools: BTreeMap::new(),
+            last_update_block: current_block,
+        }
+    }
+}
+
+impl Mutexed {
+    fn get(&self, pair: TokenPair, block: Option<u64>) -> Option<&[Pool]> {
+        let block = block.or_else(|| self.cached_most_recently_at_block.get(&pair).copied())?;
```
fleupold, line 160: Maybe we should more explicitly document this "inexact" matching (I'm slightly afraid that someone will in the future request latest block and assume it is actually going to be the latest block). We could even make the "approximate_matching" a cofigurable boolean on the outer Fetcher struct.

Alternatively we could make BlockNumber in the PoolFetching struct optional, None meaning "I don't care what block"

fleupold (commented):
Looks good overall (some comments in line). Looking for feedback from the others what they think about the different approaches

orderbook/src/orderbook_pool_fetcher.rs
```diff
@@ -0,0 +1,200 @@
...
+                // Panic because we should never write code that does this.
+                panic!("cache does not support earliest and pending");
+            }
+        };
+
+        let mut cache_hits = Vec::new();
+        let mut cache_misses = HashSet::new();
+        let last_update_block;
+        {
+            let mutexed = self.mutexed.lock().unwrap();
```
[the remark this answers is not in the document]
nlordell: Not releasing the lock would not allow simultaneous reads and fetching (in case of multiple clients). My intuition leans towards only holding the lock for reading and updating the cache would perform worse.

nlordell:
> Looking for feedback from the others what they think about the different approaches

I definitely prefer not having separate traits for this (i.e. the approach in this PR). Also, I think the implementation looks very nice (although TBH I didn't review super carefully yet).

orderbook/src/orderbook_pool_fetcher.rs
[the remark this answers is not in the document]
vkgnosis: I should have noted that in a comment.
This is intentional and has the trade offs you mention. There is a chance that some work is wasted. The upside is that it guarantees no one has to wait for other fetches. My intuition is that it is unlikely that the same tokens are queried at the same time. And once it has happened the result is cached.
There are more complicated solutions but I would like to have the base version first before attempting those.

vkgnosis:
Since this got positive feedback I will update the PR to be in mergable state and address comments.

orderbook/src/orderbook_pool_fetcher.rs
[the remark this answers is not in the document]
nlordell: > My intuition is that it is unlikely that the same tokens are queried at the same time.

Given that the base token pairs should be kept up to date with the `run_maintenance`, I also think, as you mentioned, it is more likely to have cache misses for different tokens than the same one. Also, during these cache misses, other would be stuck waiting for tokens that are already cached.

I think its better to optimize for concurrency here than for worse case.

> There are more complicated solutions but I would like to have the base version first before attempting those.

Very much agree. I would even guess that the gains would be fairly marginal from trying to sync the same fetches (as I suggested :point_up:).

orderbook/src/database/fees.rs
```diff
@@ -86,7 +86,7 @@ impl Database {
 
 #[async_trait::async_trait]
 impl Maintaining for Database {
-    async fn run_maintenance(&self) -> Result<()> {
+    async fn run_maintenance(&self, _new_block: u64) -> Result<()> {
```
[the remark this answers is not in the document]
vkgnosis: Removed it. However if the assumption is that maintenance does *not* correspond to the current block changing, then this cache should not implement Maintainable because it should always be updated when the block changes.

orderbook/src/orderbook_pool_fetcher.rs
vkgnosis: Added a comment. Agree that proper solution is to make this explicit in the block number enum. Would like to do that in future PR.

orderbook/src/database/fees.rs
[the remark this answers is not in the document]
fleupold: I see. Maybe it would make sense rename run_maintenance to on_new_block and do this refactoring codebase wide in a separate PR?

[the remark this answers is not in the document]
vkgnosis: We don't have to make maintenance run on new blocks, maybe that isn't relevant for other components. I am not sure. Just saying that if it is, then this component shouldn't be part of it. No strong feelings either way.
If we do want to change maintenance then yes I'd do it in a separate PR.

vkgnosis:
Finished integrating fully with rest of code and added tests.

shared/src/arguments.rs
```diff
@@ -61,6 +64,14 @@ pub struct Arguments {
         use_delimiter = true
     )]
     pub baseline_sources: Vec<BaselineSources>,
+
+    /// The number of blocks kept in the pool cache.
+    #[structopt(long, env, default_value = "10")]
```
fleupold, line 69: Not sure what the best value is, but maybe 4 blocks (~1min on average) would be a good compromise. If we choose this value too high, we might be quoting outdated prices occasionally.

shared/src/pool_cache.rs
```diff
@@ -0,0 +1,461 @@
...
+// going through the api do not have to wait longer than necessary:
+// - The mutex is never held while waiting on an async operation (getting pools from the node).
+// - Automatically updating the cache is decoupled from normal pool fetches.
+// A result of this is that it is possible that the same uncached pair is requested multiple times
+// simultaneously and some work is wasted. This is unlikely to happen in practice and the pool is
+// going to be cached the next time it is needed.
+// When pools are requested we mark all those pools as recently used which potentially evicts other
+// pairs from the pair lru cache. Cache misses are fetched and inserted into the cache.
+// Then when the automatic update runs the next time, we request and cache all recently used pairs.
+// For the order book we only care about the "recent" state of the pools. So we can return any
```
fleupold, line 38: nanonit: I really like the design consideration docs, just for "who uses it how" we could maybe be a bit more generic ("some consumers only care about the recent state"). Otherwise I feel like this comment won't age well with the rest of the codebase changing. But also no big deal

shared/src/pool_cache.rs
```diff
@@ -0,0 +1,461 @@
...
+// result from the cache even if it comes from previous blocks.
+// On the other hand for the driver we need to get the pool at exact blocks which is why we keep a
+// cache of previous blocks in the first place as we could simplify this module if it was only used
+// by the order book.
+
+impl PoolCache {
+    /// number_of_blocks_to_cache: Previous blocks stay cached until the block is this much older
```
fleupold, line 45: nanonit: I believe older blocks would still be cached if they are queried, but not retained on a new block.

shared/src/pool_cache.rs
```diff
@@ -0,0 +1,461 @@
...
+            mutexed.last_update_block = new_block;
+        }
+        Ok(())
+    }
+}
+
+#[async_trait::async_trait]
+impl PoolFetching for PoolCache {
+    async fn fetch(&self, pairs: HashSet<TokenPair>, block: BlockNumber) -> Result<Vec<Pool>> {
+        // We assume that if "Latest" is used it is okay to return an earlier cached version even if
```
fleupold, line 90: I assume you don't want to pass `Option<BlockNumber>` because this would make the PR too large? Can we do this right after (could have also been a preparational PR before)?

shared/src/pool_cache.rs
```diff
@@ -0,0 +1,461 @@
...
+impl Mutexed {
+    fn get(&mut self, pair: TokenPair, block: Option<u64>) -> Option<&[Pool]> {
+        self.recently_used.put(pair, ());
+        let block = block.or_else(|| self.cached_most_recently_at_block.get(&pair).copied())?;
+        self.pools.get(&(block, pair)).map(Vec::as_slice)
+    }
+
+    fn insert(&mut self, block: u64, pairs: impl Iterator<Item = TokenPair>, pools: &[Pool]) {
+        for pair in pairs {
+            self.cached_most_recently_at_block.insert(pair, block);
```
fleupold, line 188: What do we mean by "recently" here? It looks like we could cache a pair for an old block (e.g. 25 blocks in the past) and would consider this most recently. Should we check that cached_most_recently_at_block is not larger than block?

Also maybe add a unit test for this case?

fleupold (approved):
Looks good to me! I'm a bit worried that translating Block::Latest to some old block can lead to confusion once we forget about this PR, so maybe we can address this immediately after (by replacing with Option<BlockNumber> on the trait level or a custom enum).

fleupold:
I was trying to run the benchmark locally but am basically getting NotFound responses for all tokens (on staging the same requests gives a response). Maybe something is not quite right yet?

shared/src/arguments.rs
vkgnosis: We probably want a slightly higher value in the driver so we're sure we still have the pool state when we simulate solver solutions.
On the other hand in the order book as you say this is a bound on how outdated a price can be so it can be lower. Although as long as the token is in the LRU it will have updated pools anyway.

shared/src/pool_cache.rs
vkgnosis: Yes I didn't want to include it in this PR. Good idea to do it first. Will make a PR for that now.

shared/src/pool_cache.rs
vkgnosis: Good point. I think we should have another parameter that decides the maximum age of an entry you can get when requesting "latest". This would be decoupled from the number of blocks cache entries stay alive for.

vkgnosis:
> Maybe something is not quite right yet?

Last time I tried it was working. Maybe the node you're using doesn't support tracing so bad token detection fails and prevents the token from being used? Anyway, I'll test it again.

vkgnosis:
Results of Felix' benchmark. I first tested on main and then on this PR. Configuration:

* made the gas estimator return a constant 33 gwei
* default settings of the new parameters for this PR
* use our developer staging node
* use our staging base tokens
* disable trace api (no bad token detection)

image

We can see that the new cache is faster. But also note that the main feature of this PR which is automatically updating recently used pairs, is not very important for the benchmark.

shared/src/arguments.rs
```diff
@@ -61,6 +64,18 @@ pub struct Arguments {
...
+    pub pool_cache_blocks: NonZeroU64,
```
bh2smith, line 70: TIL: NonZeroU64!

shared/src/pool_cache.rs
```diff
@@ -0,0 +1,524 @@
...
+    mutexed: Mutex<Mutexed>,
+    number_of_blocks_to_cache: NonZeroU64,
+    inner: Box<dyn PoolFetching>,
+    block_stream: CurrentBlockStream,
+}
+
+// Design:
+// The design of this module is driven by the need to always return pools quickly so that end users
+// going through the api do not have to wait longer than necessary:
+// - The mutex is never held while waiting on an async operation (getting pools from the node).
```
bh2smith, line 29: Do you mean it is never blocking or "locked"? Ahhh, you mean "held" like owned elsewhere.

bh2smith (approved):
I like it and look forward to interacting with this interface when we implement `WeightedPoolFetching` (since I imagine we should use it there). Will probably have more comments and questions about convenience of use when that time comes. Nice job here. 

Also, I must say that given the complexity of this change and my current inexperience in comprehending the implications of this change, my approval should only count as 1/2.

orderbook/src/main.rs
```diff
@@ -214,11 +214,19 @@ async fn main() {
...
     let pool_aggregator = PoolAggregator::from_providers(&pair_providers, &web3).await;
-    let pool_fetcher =
-        CachedPoolFetcher::new(Box::new(pool_aggregator), current_block_stream.clone());
+    let pool_fetcher = Arc::new(
+        PoolCache::new(
+            args.shared.pool_cache_blocks,
+            args.shared.pool_cache_lru_size,
+            args.shared.pool_cache_maximum_recent_block_age,
+            Box::new(pool_aggregator),
+            current_block_stream.clone(),
```
nlordell, line 223: Is it an issue that we have separate instances of `current_block_stream` where the "current block" might be different for them because of polling offsets?

For example the `maintenance` block stream could see a new block, but the polling is offset in such a way that the `pool_cache` block stream doesn't see it, and then it will, instead, update for the previous block instead of the next one.

**edit**: I see we are using tokio spmc channels (instead of mpsc) so it is meant for a single producer (the spawned fiber) an multiple receivers, so its not an issue at all.

shared/src/pool_cache.rs
```diff
@@ -0,0 +1,524 @@
...
+        Ok(())
+    }
+}
+
+#[async_trait::async_trait]
+impl PoolFetching for PoolCache {
+    async fn fetch(&self, pairs: HashSet<TokenPair>, block: Block) -> Result<Vec<Pool>> {
+        // We assume that if "Latest" is used it is okay to return an earlier cached version even if
+        // it is not exactly latest. This should be reflected in the block enum parameter in the
+        // future.
```
nlordell, line 101: I believe this comment is out of date.

nlordell (approved):
Looks great! Sorry for taking so long to review.

shared/src/pool_cache.rs
vkgnosis: Technically mutexes can be locked and locks can be held so yeah the language here was a little inaccurate.

shared/src/pool_cache.rs
```diff
@@ -0,0 +1,521 @@
...
+        let uncached_pools = self
+            .inner
+            .fetch(cache_misses.clone(), Block::Number(cache_miss_block))
+            .await?;
+        {
+            let mut mutexed = self.mutexed.lock().unwrap();
+            mutexed.insert(cache_miss_block, cache_misses.into_iter(), &uncached_pools);
+        }
+
+        cache_hits.extend_from_slice(&uncached_pools);
```
fleupold, line 134: As a follow up, maybe we should record a prometheus metric for cache hits and cache misses?

shared/src/pool_cache.rs
fleupold: I still think we should only insert `block` here if it's higher than what we already have saved. Otherwise, we could query for a very old block (would override `cached_most_recently_at_block`) and then would cache miss despite having information available in a "recent" block. Not a super important edge case, but feels easy enough to do this check here.

fleupold (approved):
Looks great! Thanks for the benchmark (baseline seems much faster in your setting than in mine) and refactoring the signature up front.

PR #671, opened 2021-05-29
[Easy] Include Solver Name in Simulation Failure

bh2smith:
Based on several alerts of the form:

```
ERROR solver::driver: settlement simulation failed right before submission AND for block 12527827 which was current when liquidity was fetched:
```

It would be nice to know which solver is failing (given that many of them are rounding errors from HTTP solver)

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -249,7 +249,9 @@ impl Driver {
             };
             tracing::error!(
                 "{} settlement simulation failed at submission and block {}:\n{:?}",
-                settlement.name, current_block_during_liquidity_fetch, error_at_earlier_block
+                settlement.name,
+                current_block_during_liquidity_fetch,
+                error_at_earlier_block
             );
             // This is an additional debug log so that the log message doesn't get too long as
             // settlement information is recoverable through tenderly anyway.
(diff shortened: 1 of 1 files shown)


nlordell:
I'm always for adding more information in the alerts!
Note that the solver name should be there in the following log (if you still need to figure it out), but I agree that it adds an unnecessary extra step when debugging:
image

solver/src/driver.rs
```diff
@@ -248,8 +248,8 @@ impl Driver {
                 Err(err) => err,
             };
             tracing::error!(
-                "settlement simulation failed right before submission AND for block {} which was current when liquidity was fetched:\n{:?}",
-                current_block_during_liquidity_fetch, error_at_earlier_block
+                "settlement simulation for {} solver failed right before submission AND for block {} which was current when liquidity was fetched:\n{:?}",
```
nlordell, line 251: Because the tenderly link is very long, and we sometimes run into maximum message size issues, can we shorten this message to save a few characters? For example:
```suggestion
                "{} settlement simulation failed at submission and block {}:\n{:?}",
```

nlordell (approved):

vkgnosis (approved):

PR #674, opened 2021-05-31
Update web3 and tokio

vkgnosis:
This brings us up to date with dependencies using tokio 1.0.

Details:
* Removed `struct CurrentBlockStream ` because the tokio 1.0 watcher already provides the functionality we need.
* Use ethcontract's `Byte` wrapper in a few places because this is how contract abis are generated now.

TODO: make ethcontract release before merging this

### Test Plan
existing tests still work

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,5 @@ default-members = [
 ]
 
 [patch.crates-io]
-ethcontract = { git = "https://github.com/gnosis/ethcontract-rs.git", rev = "20714100f891663b3f16740df855e34222cc6178" }
-ethcontract-generate = { git = "https://github.com/gnosis/ethcontract-rs.git", rev = "20714100f891663b3f16740df855e34222cc6178" }
-warp = { git = 'https://github.com/vkgnosis/warp.git', rev = "c329629f1130c190c10fa1e8ec9fb63fc81e45b0" }
+# Warp fork with an extra commit to allow turning rejections into responses.
+warp = { git = 'https://github.com/vkgnosis/warp.git', rev = "87a91e24311b0ca6ed67408d2d302c569331dfec" }
--- a/contracts/build.rs
+++ b/contracts/build.rs
@@ -109,9 +109,7 @@ fn generate_contract_with_config(name: &str, config: impl FnOnce(Builder) -> Bui
     println!("cargo:rerun-if-changed={}", artifact.display());
     let mut builder = Builder::new(artifact)
         .with_contract_name_override(Some(name))
-        .with_visibility_modifier(Some("pub"))
-        .add_event_derive("serde::Deserialize")
-        .add_event_derive("serde::Serialize");
+        .with_visibility_modifier(Some("pub"));
 
     if let Ok(address) = fs::read_to_string(&address_file) {
         println!("cargo:rerun-if-changed={}", address_file.display());
--- a/solver/src/interactions/uniswap.rs
+++ b/solver/src/interactions/uniswap.rs
@@ -1,6 +1,7 @@
 use super::Erc20ApproveInteraction;
 use crate::{encoding::EncodedInteraction, settlement::Interaction};
 use contracts::{GPv2Settlement, IUniswapLikeRouter, ERC20};
+use ethcontract::Bytes;
 use primitive_types::{H160, U256};
 
 #[derive(Debug)]
@@ -45,7 +46,7 @@ impl UniswapInteraction {
             U256::MAX,
         );
         let calldata = method.tx.data.expect("no calldata").0;
-        (self.router.address(), 0.into(), calldata)
+        (self.router.address(), 0.into(), Bytes(calldata))
     }
 
     fn web3(&self) -> web3::Web3<ethcontract::transport::DynTransport> {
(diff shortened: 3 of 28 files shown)


contracts/build.rs
```diff
@@ -109,9 +109,7 @@ fn generate_contract_with_config(name: &str, config: impl FnOnce(Builder) -> Bui
     println!("cargo:rerun-if-changed={}", artifact.display());
     let mut builder = Builder::new(artifact)
         .with_contract_name_override(Some(name))
-        .with_visibility_modifier(Some("pub"))
-        .add_event_derive("serde::Deserialize")
-        .add_event_derive("serde::Serialize");
```
vkgnosis, line 114: We don't use this derive in v2.

Cargo.toml
```diff
@@ -17,6 +17,7 @@ default-members = [
 ]
 
 [patch.crates-io]
-ethcontract = { git = "https://github.com/gnosis/ethcontract-rs.git", rev = "20714100f891663b3f16740df855e34222cc6178" }
-ethcontract-generate = { git = "https://github.com/gnosis/ethcontract-rs.git", rev = "20714100f891663b3f16740df855e34222cc6178" }
-warp = { git = 'https://github.com/vkgnosis/warp.git', rev = "c329629f1130c190c10fa1e8ec9fb63fc81e45b0" }
+ethcontract = { git = "https://github.com/gnosis/ethcontract-rs.git", rev = "a13f0d479e65539fd43254d5b3cddab7909ec910" }
+ethcontract-generate = { git = "https://github.com/gnosis/ethcontract-rs.git", rev = "a13f0d479e65539fd43254d5b3cddab7909ec910" }
+# Warp fork with an extra commit to allow turning rejections into responses.
+warp = { git = 'https://github.com/vkgnosis/warp.git', rev = "87a91e24311b0ca6ed67408d2d302c569331dfec" }
```
fleupold, line 23: Why did the warp revision change?

fleupold (approved):
tokio::sync::watch::Receiver is nice!

solver/src/interactions/uniswap.rs
```diff
@@ -92,7 +93,7 @@ mod tests {
         // Verify Approve
         let approve_call = &interactions[0];
         assert_eq!(approve_call.0, token_in);
-        let call = &approve_call.2;
+        let call = &approve_call.2 .0;
```
bh2smith, line 96: Is the extra space here intentional!
```suggestion
        let call = &approve_call.2.0;
```

solver/src/interactions/uniswap.rs
```diff
@@ -101,7 +102,7 @@ mod tests {
         // Verify Swap
         let swap_call = &interactions[1];
         assert_eq!(swap_call.0, router.address());
-        let call = &swap_call.2;
+        let call = &swap_call.2 .0;
```
bh2smith, line 105: Here too.

solver/src/interactions/weth.rs
```diff
@@ -55,7 +56,7 @@ mod tests {
         let withdraw_call = &encoded_interactions[0];
         assert_eq!(withdraw_call.0, weth.address());
         assert_eq!(withdraw_call.1, U256::from(0));
-        let call = &withdraw_call.2;
+        let call = &withdraw_call.2 .0;
```
bh2smith, line 59: I am getting the impression that it is, will stop reporting.

bh2smith (approved):
Wow! Quite the difference between major versions, looks like we benefited from some block related feature enhancements.

nlordell:
Looks like its an issue in the `deploy` contracts script:
```
 error[E0425]: cannot find function `delay_for` in module `tokio::time`
   --> contracts/src/bin/deploy.rs:127:22
    |
127 |         tokio::time::delay_for(NODE_READY_POLL_INTERVAL).await;
    |                      ^^^^^^^^^ not found in `tokio::time`
```

Probably just needs to `s/delay_for/sleep/`.

Cargo.toml
nlordell: I imagine it needs an updated version with the latest Tokio (the old one likely being against `0.2`).

nlordell (approved):

Cargo.toml
vkgnosis: Correct. Before it was warp 0.2 with an extra commit, now it is warp 0.3 with an extra commit.

solver/src/interactions/uniswap.rs
vkgnosis: Looks like rustfmt does it this way so I would just keep it. Apparently the rust compiler used to need the extra space (but not longer does) and rustfmt hasn't been updated to reflect that.

nlordell: AFAIU its because of the nested tuple structs:
```rust
pub struct Foo(Bar);
pub struct Bar(u64);

let foo: Foo = ...;
foo.0.0;
```

The issue is parsing ambiguity on `foo.0.0`:
```
identifier(foo) dot(.) tuple_field(0) dot(.) tuple_field(0)
```
or
```
identifier(foo) dot(.) float(0.0)
```

PR #680, opened 2021-06-01
Use camel case for order cancellation serialization

nlordell:
This PR fixes the order cancellation JSON (de)serialization to use camel case like the rest of the API and as documented in the OpenAPI specification.

Previously, it was looking for a field name `signing_scheme` instead of `signingScheme`.

### Test Plan

Added new unit test.

### DIFF
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -255,7 +255,7 @@ impl OrderCreation {
 
 /// An order cancellation as provided to the orderbook by the frontend.
 #[serde_as]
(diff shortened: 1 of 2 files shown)


alfetopito (approved):
That was fast!

bh2smith (approved):
Thats cool, but how did that not get noticed in the fist place?

nlordell:
> Thats cool, but how did that not get noticed in the fist place?

:man_shrugging:

bh2smith:
Its because `signingScheme` was added later - the original implementation was already in camel case.

Issue #664, opened 2021-05-28
get_fee error: invalid response


bh2smith:
We have been getting a fairly steady flow of the following in-descriptive errors in our logs

```
ERROR orderbook::api::get_fee_info: get_fee error err=Got invalid response: Error("data did not match any variant of untagged enum Response", line: 0, column: 0)
```

which appears to be a long outstanding (non urgent) [issue](https://github.com/serde-rs/serde/issues/773) with `serde-rs`

This is the place in the code is here

http://github.com/gnosis/gp-v2-services/blob/f8739a5178ae75de49bd90575c155fc6c8586d73/orderbook/src/api/get_fee_info.rs#L56-L57

and in the same way again here:

http://github.com/gnosis/gp-v2-services/blob/f8739a5178ae75de49bd90575c155fc6c8586d73/orderbook/src/api/get_fee_info.rs#L123-L124

I have not managed to recover this response in a way that would allow us to adapt the code and catch this error. I am beginning to suspect that the response is actually "no response" so perhaps we could try and work with that.

vkgnosis:
We are probably missing some `.context(...)` somewhere which is why we don't see where the error comes from. I'll go through the get_fee code now to find the culprit.

nlordell:
The issue is a `trace_callMany` call is not returning a valid JSON RPC repsonse.

nlordell:
While the context would help, it would be really nice to see the actual response to know what the node is sending back.

vkgnosis:
We log the response in the transports logs already though. I could add it to the error directly but not sure how useful that is because it could make it very long.

vkgnosis:
Ah we probably don't log anything if it can't decode the outer jsonrpc json.

bh2smith:
Yes, these solutions are definitely an improvement, but it would be really nice to know what exactly we are failing to enumerate and catch this in a way that we can adapt to. I have asked devops for the response data and hope to at least be able to catch this response in a way that we can deal with in the future.

vkgnosis:
Using our own http Transport implementation we would be able to log those responses too.

bh2smith:
> Using our own http Transport implementation we would be able to log those responses too.

Yes, I think we will have to do this (at least temporarily) because devops has not been able to recover the response data for this RPC request.

vkgnosis:
Would be nice to have in general. I'll take stab at it.

fleupold:
> Would be nice to have in general. I'll take stab at it.

[...]

PR #682, opened 2021-06-02
Use a custom http transport implementation

vkgnosis:
This allows us to log raw responses before decoding to json and
customize the transport for example by setting a timeout (not in this
PR).

For the implementation I used our gpv1 custom transport and rust-web3's
default transport as references and read the jsonrpc specification.
I discovered that that neither previous implementation takes into
account that the server is allowed to reorder responses for batch
requests. This is fixed here and something we might want to upstream
into rust-web3.

Related https://github.com/gnosis/gp-v2-services/issues/664

### Test Plan
existing tests, manual tests with staging node (TODO)

### DIFF
--- a/shared/src/http_transport.rs
+++ b/shared/src/http_transport.rs
@@ -1,0 +1,187 @@
+use ethcontract::jsonrpc as jsonrpc_core;
+use futures::{future::BoxFuture, FutureExt};
+use jsonrpc_core::types::{Call, Output, Request, Value};
+use reqwest::{Client, Url};
+use serde::de::DeserializeOwned;
+use std::{
+    collections::HashMap,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    },
+};
+use web3::{error::Error as Web3Error, helpers, BatchTransport, RequestId, Transport};
+
+#[derive(Clone, Debug)]
+pub struct HttpTransport {
+    client: Client,
+    inner: Arc<Inner>,
+}
+
+#[derive(Debug)]
+struct Inner {
+    url: Url,
+    id: AtomicUsize,
+}
+
+impl HttpTransport {
+    pub fn new(url: Url) -> Self {
+        Self::with_client(Client::new(), url)
+    }
+
+    pub fn with_client(client: Client, url: Url) -> Self {
+        Self {
+            client,
+            inner: Arc::new(Inner {
+                url,
+                id: AtomicUsize::new(0),
+            }),
+        }
+    }
+
+    pub fn next_id(&self) -> RequestId {
+        self.inner.id.fetch_add(1, Ordering::SeqCst)
+    }
+
+    pub fn new_request(&self) -> (Client, Url) {
+        (self.client.clone(), self.inner.url.clone())
+    }
+}
+
+// Id is only used for logging.
+async fn execute_rpc<T: DeserializeOwned>(
+    client: Client,
+    url: Url,
+    id: RequestId,
+    request: &Request,
+) -> Result<T, Web3Error> {
+    tracing::debug!(
+        "[id:{}] sending request: {:?}",
+        id,
+        serde_json::to_string(&request)?
+    );
+    let response = client.post(url).json(request).send().await.map_err(|err| {
+        let message = format!("failed to send request: {}", err);
+        tracing::debug!("[id:{}] {}", id, message);
+        Web3Error::Transport(message)
+    })?;
+    let status = response.status();
+    let text = response.text().await.map_err(|err| {
+        let message = format!("failed to get response body: {}", err);
+        tracing::debug!("[id:{}] {}", id, message);
+        Web3Error::Transport(message)
+    })?;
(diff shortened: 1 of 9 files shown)


shared/src/http_transport.rs
```diff
@@ -0,0 +1,186 @@
...
+    })?;
+    let status = response.status();
+    let text = response.text().await.map_err(|err| {
+        let message = format!("failed to get response body: {}", err);
+        tracing::debug!("[id:{}] {}", id, message);
+        Web3Error::Transport(message)
+    })?;
+    // Log the raw text before decoding to get more information on responses that aren't valid
+    // json. Debug encoding so we don't get control characters like newlines in the output.
+    tracing::debug!("[id:{}] received response: {:?}", id, text.trim());
```
vkgnosis, line 76: I'd like to discuss this design choices. Debug encoding makes it harder to read the response directly in the logs because everything is one long line. On the other hand I do think it is necessary for the raw text body as the comment mentions.
An alternative could be to decode into serde_json::Value first and pretty print that. This has the advantage the output is struct-like which makes it easier to read. Only if we can't decode into Value first we would debug log the raw text.
What do you think?

nlordell:
Very good catch! According to the [JSON RPC spec](https://www.jsonrpc.org/specification#batch):
> The Response objects being returned from a batch call MAY be returned in any order within the Array. The Client SHOULD match contexts between the set of Request objects and the resulting set of Response objects based on the id member within each Object.

shared/src/http_transport.rs
nlordell: Personally, I am for this change. It also would resolve the issue @bh2smith opened here: https://github.com/gnosis/gp-v2-services/issues/669.

shared/src/http_transport.rs
```diff
@@ -0,0 +1,187 @@
...
+            ))
+        })
+        .collect::<Result<HashMap<_, _>, Web3Error>>()?;
+    ids.iter()
+        .map(|id| {
+            outputs.remove(id).ok_or_else(|| {
+                Web3Error::InvalidResponse(format!("batch response is missing id {}", id))
+            })
+        })
+        .collect()
```
nlordell, line 145: Given the spec:
> A Response object SHOULD exist for each Request object, except that there SHOULD NOT be any Response objects for notifications.

Should we also check that there are no remaining responses in the `outputs` after all the requests have been matched?

nlordell (approved):
Awesome!

shared/src/http_transport.rs
vkgnosis: If I read the issue right, you are in favor of keeping the debug log so the output stays in one line. (correct me if i am wrong)

shared/src/http_transport.rs
vkgnosis: Will add this.

shared/src/http_transport.rs
```diff
@@ -127,6 +127,11 @@ fn handle_batch_response(
     ids: &[RequestId],
     outputs: Vec<Output>,
 ) -> Result<Vec<RpcResult>, Web3Error> {
+    if ids.len() != outputs.len() {
```
nlordell, line 130: :+1:

nlordell (approved):

shared/src/http_transport.rs
```diff
@@ -0,0 +1,192 @@
...
+        )),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn handles_batch_response_being_in_different_order_than_input() {
```
fleupold, line 171: good 👁️

shared/src/http_transport.rs
```diff
@@ -0,0 +1,192 @@
...
+            client,
+            inner: Arc::new(Inner {
+                url,
+                id: AtomicUsize::new(0),
+            }),
+        }
+    }
+
+    pub fn next_id(&self) -> RequestId {
+        self.inner.id.fetch_add(1, Ordering::SeqCst)
```
fleupold, line 43: Maybe a comment on why this ordering (if I remember we don't care about performance so the most strict one?).

fleupold (approved):
Looks good. Given how similar this transport is to what we wrote in GPv1, would it make sense to factor it in a reusable crate to that the safe gateweay, ido services and whoever else uses Rust and Web3 in the context of Gnosis can reuse it (instead of copying e.g. the GPv1 version with its "bug")?

shared/src/http_transport.rs
vkgnosis: This is the most strict ordering (as in always being correct) so I don't think a comment is needed tbh for people that have encountered atomics before.

vkgnosis:
> Given how similar this transport is to what we wrote in GPv1, would it make sense to factor it in a reusable crate to that the safe gateweay, ido services and whoever else uses Rust and Web3 in the context of Gnosis can reuse it.

I would like to and am working on upstreaming this into rust-web3. Then everyone can benefit. If that doesn't work out then yes we could easily have this as a standalone crate.

PR #683, opened 2021-06-02
Represent pool fetching block number more accurately

vkgnosis:
As a precursor to the better cache PR we want to make it clear what data
you are requesting.

We should discuss what values we want to be valid in the enum. I decided to not just have an Option<BlockNumber> and instead enumerate all the states that made sense to me directly. Thus I removed Earliest and Pending.
Alternatively we could say we still want to support Pending and make the cache pass it through.

### Test Plan
existing tests

### DIFF
--- a/shared/src/pool_fetching.rs
+++ b/shared/src/pool_fetching.rs
@@ -29,14 +29,14 @@ type RelativeReserves = (U256, U256, H160);
 pub enum Block {
     /// The most recent state. This is on a best effort basis so that for example a cache can still
     /// return results that are slightly out of date.
-    Latest,
+    Recent,
     Number(u64),
 }
 
 impl From<Block> for BlockNumber {
     fn from(val: Block) -> Self {
         match val {
-            Block::Latest => BlockNumber::Latest,
+            Block::Recent => BlockNumber::Latest,
             Block::Number(number) => BlockNumber::Number(number.into()),
         }
     }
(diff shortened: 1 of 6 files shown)


shared/src/pool_fetching.rs
```diff
@@ -24,13 +24,27 @@ const POOL_SWAP_GAS_COST: usize = 60_000;
 /// `reserve_a` refers to the reserve of the excluded token.
 type RelativeReserves = (U256, U256, H160);
 
+/// The state of the chain at which information should be retrieved.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq, PartialOrd, Ord)]
+pub enum Block {
+    /// The most recent state. This is on a best effort basis so that for example a cache can still
+    /// return results that are slightly out of date.
+    Latest,
```
fleupold, line 32: Let's not call it the same as BlockNumber::Latest means something quite different (the comment clarifies it, but I'm afraid some future maintainer might shortcut their understanding). Maybe Recent?

fleupold (approved):
I think a separate enum makes sense and is more explicit than an Option.

nlordell (approved):
Looks great. Agree that we shouldn't call it `Latest` as to not overload its existing meaning in Ethereum/JSON RPC (i.e. the latest verified block on the block chain).

shared/src/pool_fetching.rs
vkgnosis: Renamed to Recent.

PR #687, opened 2021-06-03
Use pool cache in driver

vkgnosis:
Enable the pool cache in the driver. I decided to move the lru_size argument into the order book and have the lru be always turned off in the driver. This makes sense to me as in the driver we should be able to fetch all of the needed pools up front together which is less wasteful than auto updating.

### Test Plan
manually checked that maintenance runs, e2e tests otherwise

### DIFF
--- a/solver/src/main.rs
+++ b/solver/src/main.rs
@@ -5,8 +5,8 @@ use reqwest::Url;
 use shared::{
     amm_pair_provider::{SushiswapPairProvider, UniswapPairProvider},
     bad_token::list_based::ListBasedDetector,
-    http_transport::HttpTransport,
     current_block::current_block_stream,
+    http_transport::HttpTransport,
     maintenance::ServiceMaintenance,
     metrics::serve_metrics,
     network::network_name,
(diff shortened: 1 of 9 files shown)


nlordell (approved):

solver/src/main.rs
```diff
@@ -201,14 +204,27 @@ async fn main() {
...
+    let current_block_stream =
+        current_block_stream(web3.clone(), args.shared.block_stream_poll_interval_seconds)
+            .await
+            .unwrap();
+
     let pair_providers =
         pool_aggregating::pair_providers(&args.shared.baseline_sources, chain_id, &web3).await;
-
     let pool_aggregator = PoolAggregator::from_providers(&pair_providers, &web3).await;
+    let pool_fetcher = Arc::new(
```
fleupold, line 215: Pool fetcher is also used inside `UniswapLikeLiquidity` (https://github.com/gnosis/oba-services/blob/52a710f9bf000a77ff70c1e4c53311382030f59a/solver/src/liquidity/uniswap.rs#L83) for when liquidity_collector uses it. Maybe there is a way to share the cache between the two callsites (might require more refactoring though)?

fleupold (approved):

vkgnosis:
Added metrics.

solver/src/main.rs
vkgnosis: At first glance that it is more involved with refactoring so I am going to to do it in a separate PR.

PR #688, opened 2021-06-03
Add metrics for pool cache

vkgnosis:
### Test Plan
manually tested that counters show up and increase

### DIFF
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -221,6 +221,7 @@ async fn main() {
             args.shared.pool_cache_maximum_recent_block_age,
             Box::new(pool_aggregator),
             current_block_stream.clone(),
+            metrics.clone(),
         )
         .expect("failed to create pool cache"),
     );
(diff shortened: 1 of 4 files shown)


nlordell (approved):

orderbook/src/metrics.rs
```diff
@@ -14,6 +14,8 @@ pub struct Metrics {
     db_table_row_count: IntGaugeVec,
     /// Outgoing RPC request metrics
     rpc_requests: HistogramVec,
+    pool_cache_hits: IntCounter,
```
fleupold, line 17: Nanonit: could also use one metric with two labels to keep this struct a bit smaller (feel free to ignore)

orderbook/src/main.rs
```diff
@@ -221,6 +221,7 @@ async fn main() {
             args.shared.pool_cache_maximum_recent_block_age,
             Box::new(pool_aggregator),
             current_block_stream.clone(),
+            metrics.clone(),
```
fleupold, line 224: We should probably start using the cache in the driver as well, shouldn't we?

fleupold (approved):

orderbook/src/main.rs
fleupold: #687 already done 🙈

orderbook/src/metrics.rs
vkgnosis: Thought about that but I find it more natural and in line with how Prometheus docs describe new name vs new label https://prometheus.io/docs/practices/naming/ .

PR #691, opened 2021-06-04
Retry failed requests in cached pool fetcher

vkgnosis:
to get rid of annoying alerts as an interim solution.

Also turned the pool cache config into a struct because clippy
complained and I agree with it that this is easier to read and less
likely to have order of arguments errors.

### Test Plan
Manually tested that price estimation still works. Didn't write a test for retry logic because it is simple and uses sleep so I didn't want all the boilerplate for testing.

### DIFF
--- a/e2e/tests/services.rs
+++ b/e2e/tests/services.rs
@@ -14,7 +14,7 @@ use shared::{
     bad_token::list_based::ListBasedDetector,
     current_block::{current_block_stream, CurrentBlockStream},
     maintenance::ServiceMaintenance,
-    pool_cache::PoolCache,
+    pool_cache::{PoolCache, PoolCacheConfig},
     pool_fetching::PoolFetcher,
     price_estimate::BaselinePriceEstimator,
     Web3,
@@ -144,9 +144,12 @@ impl OrderbookServices {
             .await
             .unwrap();
         let pool_fetcher = PoolCache::new(
-            NonZeroU64::new(10).unwrap(),
-            20,
-            4,
+            PoolCacheConfig {
+                number_of_blocks_to_cache: NonZeroU64::new(10).unwrap(),
+                number_of_pairs_to_auto_update: 20,
+                maximum_recent_block_age: 4,
+                ..Default::default()
+            },
             Box::new(PoolFetcher {
                 pair_provider,
                 web3: web3.clone(),
--- a/shared/src/arguments.rs
+++ b/shared/src/arguments.rs
@@ -73,6 +73,14 @@ pub struct Arguments {
     #[structopt(long, env, default_value = "4")]
     pub pool_cache_maximum_recent_block_age: u64,
 
+    /// How often to retry requests in the pool cache.
+    #[structopt(long, env, default_value = "5")]
+    pub pool_cache_maximum_retries: u32,
+
+    /// How long to sleep between retries in the pool cache.
+    #[structopt(long, env, default_value = "1", parse(try_from_str = duration_from_seconds))]
+    pub pool_cache_delay_between_retries_seconds: Duration,
+
     /// How often we poll the node to check if the current block has changed.
     #[structopt(
         long,
--- a/shared/src/pool_cache.rs
+++ b/shared/src/pool_cache.rs
@@ -10,6 +10,7 @@ use std::{
     collections::{hash_map::Entry, BTreeMap, HashMap, HashSet},
     num::NonZeroU64,
     sync::{Arc, Mutex},
+    time::Duration,
 };
 
 /// Caching pool fetcher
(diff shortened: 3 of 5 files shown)


shared/src/arguments.rs
```diff
@@ -73,6 +73,14 @@ pub struct Arguments {
     #[structopt(long, env, default_value = "4")]
     pub pool_cache_maximum_recent_block_age: u64,
 
+    /// How often to retry requests in the pool cache.
+    #[structopt(long, env, default_value = "5")]
+    pub pool_cache_maximum_retries: u32,
```
vkgnosis, line 78: Note that without retrying we wouldn't get a response at all so waiting up to ~5 seconds is still better than nothing.

bh2smith: ~~We might want to include "seconds" in the name here, I interpreted this as number of retries. unless that's what it is... but the doc string seems to conflict with the name.~~

Given the next config value, I am beginning to suspect it is the doc string that is incorrect here.

shared/src/pool_cache.rs
```diff
@@ -103,6 +126,21 @@ impl PoolCache {
...
+
+    // Sometimes nodes requests error when we try to get state from what we think is the current
+    // block when the node has been load balanced out to one that hasn't seen the block yet. As a
+    // workaround we repeat the request up to N times while sleeping in between.
+    async fn fetch_inner(&self, pairs: HashSet<TokenPair>, block: Block) -> Result<Vec<Pool>> {
+        let fetch = || self.inner.fetch(pairs.clone(), block);
+        for _ in 0..self.maximum_retries {
+            match fetch().await {
+                Ok(pools) => return Ok(pools),
+                Err(err) => tracing::debug!("retrying pool fetch because error: {:?}", err),
```
bh2smith, line 138: Super glad this wont be an logged error anymore! Cheers to quiet, but informative alert channels.

solver/src/main.rs
```diff
@@ -212,9 +212,13 @@ async fn main() {
     let pool_aggregator = PoolAggregator::from_providers(&pair_providers, &web3).await;
     let pool_fetcher = Arc::new(
         PoolCache::new(
-            args.shared.pool_cache_blocks,
-            0,
-            args.shared.pool_cache_maximum_recent_block_age,
+            PoolCacheConfig {
+                number_of_blocks_to_cache: args.shared.pool_cache_blocks,
+                number_of_pairs_to_auto_update: 0,
```
bh2smith, line 217: any particular reason we hard code the zero here?

bh2smith (approved):
So excited to have this included! On call duties will actually become fun!

e2e/tests/services.rs
```diff
@@ -144,9 +144,12 @@ impl OrderbookServices {
             .await
             .unwrap();
         let pool_fetcher = PoolCache::new(
-            NonZeroU64::new(10).unwrap(),
-            20,
-            4,
+            PoolCacheConfig {
```
nlordell, line 147: I love this change!

shared/src/pool_cache.rs
nlordell: It should still be an error if all the retries fail.

solver/src/main.rs
nlordell: I think its because the driver doesn't auto-update pairs on new blocks, but instead forces the cache to get primed with a specific block number on each run. This way all the components during the run dont:
- need to go to the EVM in order to get data
- use consistent data from the same block

nlordell (approved):
Looks very good.

solver/src/main.rs
fleupold: Maybe good to put this thought into a comment there?

shared/src/pool_cache.rs
```diff
@@ -103,6 +126,21 @@ impl PoolCache {
...
+            }
+            tokio::time::sleep(self.delay_between_retries).await;
```
fleupold, line 140: A bit sad that this breaks testability at least of this code path :-( I guess we don't have our mockable sleeping trait yet.

fleupold (approved):
Nice and simple.

shared/src/pool_cache.rs
vkgnosis: Note that the only thing we're not testing is the error. The update function itself is still being tested through the ok path.

solver/src/main.rs
vkgnosis: > but instead forces the cache to get primed with a specific block number on each run

That is an idea for a future change. Either way the driver does not need auto updating so it is disabled. I already disabled it in the PR that made use of the cache in the driver. Added the comment.

PR #692, opened 2021-06-04
[Objective Criterion] Add solver fees to objective value

fleupold:
Our current objective value deducts the costs (gas) from the achieved surplus but doesn't offset its cost by the accrued fees. This may lead to batches where the objective criterion is actually negative (for smaller amounts the surplus tends to be much smaller than the gas costs).

This may lead to weird behaviour that simpler, single trade solutions are picked over complex CoW solutions, because the single trade solution is less negative (and therefore better) than matching many negative trades at once.

This PR adds the solver_fees to the objective value and also scales them by the discount factor (pretending users were paying the real min fee)

### Test Plan
Added a unit test for both individual trade fee computation as well as the total fee computation for a settlement (containing normalizing prices).

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -52,6 +52,7 @@ pub struct Driver {
     market_makable_token_list: Option<TokenList>,
     inflight_trades: HashSet<OrderUid>,
     block_stream: CurrentBlockStream,
+    fee_discount_factor: f64,
 }
 impl Driver {
     #[allow(clippy::too_many_arguments)]
@@ -73,6 +74,7 @@ impl Driver {
         gas_price_cap: f64,
         market_makable_token_list: Option<TokenList>,
         block_stream: CurrentBlockStream,
+        fee_discount_factor: f64,
     ) -> Self {
         Self {
             settlement_contract,
@@ -93,6 +95,7 @@ impl Driver {
             market_makable_token_list,
             inflight_trades: HashSet::new(),
             block_stream,
+            fee_discount_factor,
         }
     }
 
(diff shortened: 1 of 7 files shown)


solver/src/driver.rs
```diff
@@ -309,6 +312,9 @@ impl Driver {
         futures::stream::iter(settlements)
             .filter_map(|settlement| async {
                 let surplus = settlement.settlement.total_surplus(prices);
+                // Given the fee discount the solver fees by themselves may not be sufficient to make a solution economically viable (leading to a negative objective value)
```
nlordell, line 315: I didn't understand the sentence. Did you mean this?

```suggestion
                // Given the fee discount the solver applies by themselves may not be sufficient to make a solution economically viable (leading to a negative objective value)
```

nlordell (commented):
Looks good at a first glance. Will take a more careful look on Monday.

bh2smith (approved):

solver/src/settlement.rs
```diff
@@ -46,6 +46,20 @@ impl Trade {
         }
     }
 
+    // Returns the relative fee amount given the executed amount
+    // cf. https://github.com/gnosis/gp-v2-contracts/blob/964f1eb76f366f652db7f4c2cb5ff9bfa26eb2cd/src/contracts/GPv2Settlement.sol#L370-L371
+    pub fn fee(&self) -> U256 {
```
vkgnosis, line 51: This is the executed fee amount, right? Maybe easier to understand if is called that instead of relative fee.

solver/src/driver.rs
```diff
@@ -309,6 +312,9 @@ impl Driver {
...
+                // We therefore reverse apply the fee discount to simulate unsubsidized fees for ranking.
+                let solver_fees = settlement.settlement.total_fees(prices) / BigRational::from_float(self.fee_discount_factor).expect("Discount factor is not a rational");
```
marcovc, line 317: If you'd like to make it more explicit, I was having this as a SettlementEncoder member called `unsubsidized_fees` in the other [PR](https://github.com/gnosis/gp-v2-services/pull/598/files#diff-2c61b8ac9597c7fc4f938bf785c1f2b9777cd3b65d04f444be9860e30364f0a1R223-R230).

marcovc (approved):

vkgnosis (approved):

nlordell (approved):

PR #695, opened 2021-06-08
Sends a new parameter to the http solver with the instance name.

marcovc:
This is used for logging and generating intermediate files (which will end up in the S3 bucket).

### Test Plan
Tested locally.

### DIFF
--- a/solver/src/http_solver.rs
+++ b/solver/src/http_solver.rs
@@ -36,6 +36,7 @@ const GAS_PER_UNISWAP: u128 = 94696;
 pub struct SolverConfig {
     pub max_nr_exec_orders: u32,
     pub time_limit: u32,
+    pub instance_name: String,
     // TODO: add more parameters that we want to set
 }
 
@@ -46,7 +47,8 @@ impl SolverConfig {
                 "max_nr_exec_orders",
                 self.max_nr_exec_orders.to_string().as_str(),
             )
-            .append_pair("time_limit", self.time_limit.to_string().as_str());
+            .append_pair("time_limit", self.time_limit.to_string().as_str())
+            .append_pair("instance_name", self.instance_name.as_str());
     }
 }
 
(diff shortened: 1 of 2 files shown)


vkgnosis (approved):

nlordell (approved):

solver/src/http_solver.rs
```diff
@@ -46,7 +47,8 @@ impl SolverConfig {
                 "max_nr_exec_orders",
                 self.max_nr_exec_orders.to_string().as_str(),
             )
-            .append_pair("time_limit", self.time_limit.to_string().as_str());
+            .append_pair("time_limit", self.time_limit.to_string().as_str())
+            .append_pair("instance_name", self.instance_name.as_str());
```
fleupold, line 51: Why does the driver generate this? Is it logged somewhere here so that it can be cross-referenced?

marcovc: The driver generates this because the solver doesn't know have the info to properly name the instance (it doesn't know about networks and chains). Logging is a [good idea](https://github.com/gnosis/gp-v2-services/pull/701).

fleupold: I see, chain and network id are equivalent (or at least uniquely identify the network) so just passing a header with this could have been enough. But if we log it in the driver, it should be simple to find the exact file name in the AWS bucket when debugging a specific instance.

marcovc: It's bit confusing to me. By [this table](https://github.com/gnosis/gp-v2-services/blob/main/shared/src/network.rs) for example 

```
(network_id, chain_id) => network_name
-------------------------------------------------
("1", 1) => "Ethereum", 
("1", 101) => "EtherInc",
 ("3", 3) => "Ethereum",
```
from which I inferred that

* network_id != chain_id
* chain_id might encode everything (but is not user friendly)
* network_name perhaps could be improved in the table

So I decided to just include chain_id and network_name.

PR #696, opened 2021-06-08
[Refactor] Event Handling for Balancer WeightedPools

### LINKED ISSUE #681: Filtered Balancer Event Streaming
The Balancer Vault contract emits several events that are not relevant to construction of Weighted Pool data

```
    AuthorizerChanged(IAuthorizer indexed newAuthorizer)
    ExternalBalanceTransfer(IERC20 indexed token, address indexed sender, address recipient, uint256 amount)
    FlashLoan(IFlashLoanRecipient indexed recipient, IERC20 indexed token, uint256 amount, uint256 feeAmount);
    InternalBalanceChanged(address indexed user, IERC20 indexed token, int256 delta)
    PausedStateChanged(bool paused);
    PoolBalanceChanged(bytes32 indexed poolId, address indexed liquidityProvider, IERC20[] tokens, int256[] deltas, uint256[] protocolFeeAmounts);
    PoolBalanceManaged(bytes32 indexed poolId,address indexed assetManager,IERC20 indexed token,int256 cashDelta,int256 managedDelta);
    PoolRegistered(bytes32 indexed poolId, address indexed poolAddress, PoolSpecialization specialization);
    RelayerApprovalChanged(address indexed relayer, address indexed sender, bool approved)
    Swap(bytes32 indexed poolId,IERC20 indexed tokenIn,IERC20 indexed tokenOut,uint256 amountIn,uint256 amountOut);
    TokensDeregistered(bytes32 indexed poolId, IERC20[] tokens);
    TokensRegistered(bytes32 indexed poolId, IERC20[] tokens, address[] assetManagers);
 ```
 
Specifically, we are only interested in 

```
    PoolRegistered(bytes32 indexed poolId, address indexed poolAddress, PoolSpecialization specialization);
    TokensRegistered(bytes32 indexed poolId, IERC20[] tokens, address[] assetManagers);
    TokensDeregistered(bytes32 indexed poolId, IERC20[] tokens);
```

in order to reduce our requests, we should refactor `EventReceiving` to only request events by topicId.

### PR DESCRIPTION
**Pull request by bh2smith**
Partial fulfillment of #603

Due to discussion raised here, we have taken a more concise route Weighted Pools fetching. That is listening for PoolCreated events from the `WeightedPoolFactory` contract.

This event emits only the `pool_address`from which we recover the `pool_id` and `normalized_weights` from the WeightedPool contract itself and the `pool_id` from the Vault contract.

This greatly simplifies the logic for maintaining the pool store since we are not expected to gather scattered information over several events (but do now rely more on a sturdy web3 connection). 

Furthermore, this change eliminates the need for filtering Balancer Vault event stream (i.e. closes #681).

Adding a new field `normalized_weights` to WeightedPool and fetching this information from the Pool contract via eth_call. 

### Test Plan

No new tests introduced here, they were just updated to accommodate the refactoring.

### CODE REVIEW

File: contracts/src/bin/vendor.rs [hunk 1]
```diff
@@ -28,8 +28,11 @@ fn run() -> Result<()> {
...
         .github(
             "BalancerV2Vault",
-            "balancer-labs/balancer-core-v2/v1.0.0-artifacts/\
-             deployments/artifacts/mainnet/Vault.json",
+            "balancer-labs/balancer-v2-monorepo/a3b570a2aa655d4c4941a67e3db6a06fbd72ef09/pkg/deployments/deployed/mainnet/Vault.json",
+        )?
+        .github(
+            "BalancerV2WeightedPool",
+            "balancer-labs/balancer-v2-monorepo/a3b570a2aa655d4c4941a67e3db6a06fbd72ef09/pkg/deployments/extra-abis/WeightedPool.json",
         )?
```
nlordell, line 36: This probably requires only vendoring the ABI and should be added to the `vendor.abi_only()` list below :point_down:.

vkgnosis (approved):

shared/src/balancer_event_handler.rs
```diff
@@ -63,31 +67,62 @@ pub struct WeightedPoolBuilder {
...
+    ) -> Result<WeightedPool> {
+        if let (Some(pool_registration), Some(tokens_registration)) = (
+            self.pool_registration.clone(),
+            self.tokens_registration.clone(),
+        ) {
+            return Ok(WeightedPool {
                 pool_id: pool_registration.pool_id,
                 pool_address: pool_registration.pool_address,
                 tokens: tokens_registration.tokens,
+                normalized_weights: weight_fetcher
```
fleupold, line 82: Has there been a discussion as to whether this should happen on event processing or lazily when the pool is actually used? A reason to do it here would be that we only have to do it once (it looks immutable in the contract). 
However, the purpose of the pool registry is mainly to "know" which pools exist and maybe we can save a bunch of queries (one roundtrip per pool on coldstart) if we instead fetch these infos batched when we also query the reserves.

For this it might also matter if/how we plan on handling non-weighted pools. Right now it seems like this registry is agnostic to the specializtion, but with this PR it would only work for WeightedPools (what happens if we pick up a more general pool creation even)? I'd prefer if this registry stays pool agnostic, but if we want to have it here it should be more up-front about which types of pools it can handle (and not issue a RPC call if it knows it cannot succeed).

fleupold (dismissed):
It looks like we are mixing general pool handling with some assumptions about certain specialised (in particular "weighted") pool handling in the same component, which I'd like to discuss before pursuing.

shared/src/balancer_event_handler.rs
bh2smith: Very good point to bring up. I thought it would be nice to have this information right off the bat, but had not considered the fact that General pools are the only ones with weights. I guess I was assuming that the other pools would still contain this information although the weights are implied by the pool type and wouldn't need to be fetched. I'd be happy to keep the pools agnostic and move the weight fetching to the pool fetching when we get the reserves. This would only really incur one additional EVM query (depending on the pool type). 

I do recall @nlordell mentioning that we were only aiming to support `WeightedPools` initially (i.e. those with General Specialization). Maybe we should wait and see what he is thinking.

bh2smith: I have taken your comment into consideration and created a new PR which addresses this. This new approach fetches the pool weights during poolFetching when seeking the pool's reserves. This then allows us to filter and separate for different pools until they are fully supported, meanwhile only constructing the public facing WeightedPool Structure at the time all of this information is known.

Please see here for potentially superseding changes https://github.com/gnosis/gp-v2-services/pull/702 (Note that these changes incorporate the work of both this PR and #698).

nlordell: I actually think we should keep this as is. The reason I say this is that it is that a general Balancer pool registry is not super useful without knowing exactly what kind of pool is being used. The `specialization` doesn't give enough information in that regard.

I checked what Balancer does in their subgraph <https://github.com/balancer-labs/balancer-subgraph-v2> and they listen for `PoolCreated` events on the `WeighedPoolFactory` instead of the `Vault` in order to get a list of weighted pools.

fleupold: This would be an argument for making the PoolRegistry practically a WeightedPool registry. However, I'd still question if we should do this roundtrip on every pool creation event or lazily (and then batched) in the dynamic state fetching logic?

nlordell: I would also prefer lazily, but I do think that the registry should be `WeightedPool` speific.

bh2smith: I will rework this and use the weighted pool factory here.

contracts/build.rs
```diff
@@ -33,6 +33,24 @@ fn main() {
                 )),
             )
     });
+    generate_contract_with_config("BalancerV2WeightedPoolFactory", |builder| {
+        builder
+            .with_contract_mod_override(Some("balancer_v2_factory"))
```
nlordell, line 38: naming-nit: I would maybe specify that this is specifically a weighted pool factory (as we may eventually add support for other pools like stable pools, which would mean we we need a `balancer_v2_stable_pool_factory`).

```suggestion
            .with_contract_mod_override(Some("balancer_v2_weighted_pool_factory"))
```

contracts/src/lib.rs
```diff
@@ -3,6 +3,10 @@ pub mod paths;
 
 include!(concat!(env!("OUT_DIR"), "/BalancerV2Vault.rs"));
 include!(concat!(env!("OUT_DIR"), "/BalancerV2WeightedPool.rs"));
+include!(concat!(
```
nlordell, line 6: Can you add tests for networks and bytecode :point_down: in this file?

contracts/src/bin/vendor.rs
```diff
@@ -30,6 +30,10 @@ fn run() -> Result<()> {
             "BalancerV2Vault",
             "balancer-labs/balancer-v2-monorepo/a3b570a2aa655d4c4941a67e3db6a06fbd72ef09/pkg/deployments/deployed/mainnet/Vault.json",
         )?
+        .github(
+            "BalancerV2WeightedPoolFactory",
```
nlordell, line 34: :+1: I think this is in the right spot as we would eventually need the bytecode for E2E test.

shared/src/balancer_event_handler.rs
```diff
@@ -3,195 +3,183 @@ use crate::{
...
+        let pool_contract = BalancerV2WeightedPool::at(self, pool_address);
+        Ok(H256::from(
+            pool_contract.methods().get_pool_id().call().await?.0,
+        ))
+    }
+
+    async fn get_pool_data(&self, pool_address: H160) -> Result<WeightedPoolData> {
+        let pool_id = self.get_pool_id(pool_address).await?;
+        let tokens = self.get_pool_tokens(pool_id).await?;
+        let weights = self.get_normalized_weights(pool_address).await?;
```
bh2smith, line 111: Only two of these can be parallelized... may not be worth the effort to make this more efficient (since tokens requires pool_id).

shared/src/balancer_event_handler.rs
```diff
@@ -3,195 +3,183 @@ use crate::{
...
     pool_address: H160,
     tokens: Vec<H160>,
-    specialization: PoolSpecialization,
+    normalized_weights: Vec<U256>,
     block_created: u64,
 }
 
-#[derive(Clone, Debug, Default, PartialEq)]
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq, Hash)]
 pub struct WeightedPoolBuilder {
```
nlordell, line 43: Is this still needed now that we have a single event?

shared/src/balancer_event_handler.rs
```diff
@@ -224,42 +213,31 @@ impl BalancerPools {
...
-                    ContractEvent::TokensDeregistered(event) => {
-                        tracing::error!("unexpected Token Deregistration event {:?}", event);
-                        None
-                    }
-                    _ => {
-                        // TODO - Not processing other events at the moment.
-                        // https://github.com/gnosis/gp-v2-services/issues/681
-                        None
-                    }
+                    ContractEvent::PoolCreated(event) => convert_pool_created(&event, &meta),
```
bh2smith, line 225: Still using the match to unpack the wrapped data, but there is only one possibility for this match... Is there something simpler I could do to unpack it?

shared/src/balancer_event_handler.rs
```diff
@@ -3,195 +3,183 @@ use crate::{
...
+    weights: Vec<U256>,
+}
+
+#[automock]
+#[async_trait::async_trait]
+trait PoolDataFetching: Send + Sync {
+    async fn get_normalized_weights(&self, pool_address: H160) -> Result<Vec<U256>>;
+    async fn get_pool_tokens(&self, pool_id: H256) -> Result<Vec<H160>>;
+    async fn get_pool_id(&self, pool_address: H160) -> Result<H256>;
+    async fn get_pool_data(&self, pool_address: H160) -> Result<WeightedPoolData>;
```
nlordell, line 76:
```suggestion
    async fn get_pool_data(&self, pool_address: H160) -> Result<WeightedPoolData>;
```

Since we intend to get the data "all or nothing", I think it makes more sense for this to be a single method. It would also allow you to make a batch call to read all the pool data together (instead of individual calls).

shared/src/balancer_event_handler.rs
```diff
@@ -3,195 +3,183 @@ use crate::{
...
-                self.pending_pools.remove(&pool_id);
-                for token in weighted_pool.tokens {
-                    self.pools_by_token
-                        .entry(token)
-                        .or_default()
-                        .insert(pool_id);
-                }
+    async fn try_upgrade(&mut self) -> Result<()> {
+        for pool_builder in self.pending_pools.clone() {
+            let weighted_pool = pool_builder.into_pool(&*self.data_fetcher).await?;
```
bh2smith, line 137: This part of the EVM calls could probably be parallelized. Although I would prefer to do this in a separate PR.

shared/src/balancer_event_handler.rs
```diff
@@ -3,195 +3,183 @@ use crate::{
...
-    fn new(specialization: u8) -> Result<Self> {
-        match specialization {
-            0 => Ok(Self::General),
-            1 => Ok(Self::MinimalSwapInfo),
-            2 => Ok(Self::TwoToken),
-            t => Err(anyhow!("Invalid PoolSpecialization value {} (> 2)", t)),
-        }
-    }
+    /// Temporary storage for WeightedPools having errored on data fetching
+    pending_pools: HashSet<WeightedPoolBuilder>,
```
nlordell, line 129: I'm not sure I understand the use of a `HashSet` here - since we don't expect two created events to come in for the same pool right? Also is this really needed now that we only need a single event in order to register a weighted pool?

nlordell (commented):
Looks like the fact that its a single event greatly simplified some things. I don't think the `WeightedPoolBuilder` abstraction is still needed anymore because of that (since we don't need to wait for an event pair anymore).

Also, could you split this PR into:
- A PR for handling `WeightedPoolFactory` events instead of `Vault` events for adding pool addresses to the pool
- A PR for fetching pool data from a pool address (essentially the `PoolDataFetching` trait)
- A PR for combining the two together (either by fetching immediately on a new pool - or by doing it lazily as @fleupold suggested)

The reason for this is I think the suggestion from @fleupold to keep the `BalancerEventUpdater` as a simple registry for `WeightedPool` addresses and have a separate component for fetching immutable data (normalized weights, tokens, ID, etc) deserves a separate discussion. If we split this PR into those 3 pieces, we can already start merging the first two (which are simpler and more straight forward) all the while starting the discussion for the 3rd one.

shared/src/balancer_event_handler.rs
vkgnosis: No, you always need the match.

contracts/build.rs
bh2smith: I can do this, although it would have been the plan to change it when necessary.

shared/src/balancer_event_handler.rs
bh2smith: Well, we do still need to make the web3 requests for additional data, and this would allow us to parallelize/batch those queries. What do you think?

shared/src/balancer_event_handler.rs
nlordell: Personally I also find the `match` nice in this case as the code guarantees that there is only a single event (if there were more variants, you would get a compiler error).

shared/src/balancer_event_handler.rs
bh2smith: I have answered this above but will rework pending pools into a vector instead. The reason for this as a set instead of a map is that we no longer have pool_id at event emission time. So I turned it from HashMap to HashSet. The only other option for use here is to use a Vector if that is something you would prefer. It is assumed that we would never get the same event twice since I don't believe that could ever happen.

shared/src/balancer_event_handler.rs
```diff
@@ -3,195 +3,183 @@ use crate::{
...
-                BalancerEvent::PoolRegistered(event) => self.insert_pool(index, event),
-                BalancerEvent::TokensRegistered(event) => self.insert_token_data(index, event),
-            };
+    async fn insert_events(&mut self, events: Vec<(EventIndex, PoolCreated)>) -> Result<()> {
+        for (index, creation) in events {
+            self.insert_pool(index, creation)
         }
-        // In the future, when processing TokensDeregistered we may have to downgrade the result.
-        self.try_upgrade();
+        self.try_upgrade().await?;
```
fleupold, line 157: Why do we still need the try upgrade logic? Can't we now that there is only one event not always immediately create the struct from the event data?

shared/src/balancer_event_handler.rs
```diff
@@ -3,195 +3,183 @@ use crate::{
...
-                pool_address: pool_registration.pool_address,
-                tokens: tokens_registration.tokens,
-                specialization: pool_registration.specialization,
-                block_created: self.block_created,
-            });
-        }
-        None
+    async fn into_pool(self, data_fetcher: &dyn PoolDataFetching) -> Result<WeightedPool> {
+        let pool_address = self.pool_creation.pool_address;
+        let pool_data = data_fetcher.get_pool_data(pool_address).await?;
```
fleupold, line 52: Didn't we say this is something we want to do lazily when the pool is actually used not on every pool registration event (would add a roundtrip per pool on coldstart)

contracts/build.rs
nlordell: Maybe `balancer_v2_pool_factory` at least, since I find just `factory` a little vague (is it for the `Vault`? For BAL token? etc.).

shared/src/balancer_event_handler.rs
```diff
@@ -3,195 +3,183 @@ use crate::{
...
+            weights,
+        })
     }
 }
 
 /// The BalancerPool struct represents in-memory storage of all deployed Balancer Pools
-#[derive(Debug, Default)]
+#[derive(Derivative)]
+#[derivative(Debug)]
 pub struct BalancerPools {
```
fleupold, line 123: Maybe this struct should now become a WeightedBalancerPoolRegistry?

contracts/build.rs
```diff
@@ -33,6 +33,24 @@ fn main() {
                 )),
             )
     });
+    generate_contract_with_config("BalancerV2WeightedPoolFactory", |builder| {
```
fleupold, line 36: can we remove the contract that we were listening to previously?

shared/src/balancer_event_handler.rs
nlordell: I don't think we even need this in the first place - to me it looks like these are added to the `pending_pools` member temporarily in `insert_events` only to be upgraded at the end. I think we can just construct a list of `WeightedPool`s from a list of `PoolCreated` events with a block number right?

shared/src/balancer_event_handler.rs
nlordell: But what advantage does this have over just using `(PoolCreated, u64)`?

fleupold (commented):
I think the BalancerPool component should just become a registry that knows for a token pair which pool IDs are relevant. We can then have a separate component that fetches the per pool information (mutable and immutable) whenever they are used. This will allow more efficient batching of calls and make this component very simple (complexity will be in the other component then).

I see that we currently also fetch the token information in a separate call (instead of getting it from an event). Is there a way how we can still get the token list even based in this component?

bh2smith:
> Looks like the fact that its a single event greatly simplified some things. I don't think the `WeightedPoolBuilder` abstraction is still needed anymore because of that (since we don't need to wait for an event pair anymore).

I have commented above that I do still feel the need for having the data stored as a builder before the remaining information is known and this allows us to parallelize the EVM queries to fetch this information. 


That being said, I would like to point out that I do not believe there is any value in storing a weighted pool addresses without the token information. This token information absolutely must be known at the time of balance fetching.

> Also, could you split this PR into:
> 
>     * A PR for handling `WeightedPoolFactory` events instead of `Vault` events for adding pool addresses to the pool
>     * A PR for fetching pool data from a pool address (essentially the `PoolDataFetching` trait)
>     * A PR for combining the two together (either by fetching immediately on a new pool - or by doing it lazily as @fleupold suggested)

Given my proposal that PoolBuilder and the upgrade method is still useful I would insist that the PR can't really be split up into these chunks. I will, however, give it a try and make a very simple pool struct that only knows addresses.

contracts/build.rs
bh2smith: We still need it to get the pool_id

shared/src/balancer_event_handler.rs
bh2smith: Yes, I was trying to keep this massive PR as small as possible and would have done this in a separate PR.

contracts/build.rs
nlordell: We will need the `Vault` contract still for building swap calldata.

shared/src/balancer_event_handler.rs
bh2smith: Yes, you have mentioned this lazy approach in the past, but I am not convinced there is any value in storing a pool address without the known token information. Given that I think we need the token addresses immediately, I just thought we might as well fetch the rest of the pool info while we are at it.

shared/src/balancer_event_handler.rs
nlordell: I actually like `BalancerPools`, as we can extend this to instead of return a:
```
enum BalancerPool {
    Weighted(...),
    Stable(...),
}
```

When we start supporting more pool types.

shared/src/balancer_event_handler.rs
bh2smith: The upgrade logic is meant to make the pool data fetching a parallelized EVM call.

shared/src/balancer_event_handler.rs
bh2smith: As I have mentioned above, this try_upgrade logic with the pool builder has been left here so that we can make batch_calls to the EVM for pool token data.

shared/src/balancer_event_handler.rs
bh2smith: Ok sure, we may not need a `WeightedPoolBuilder` struct anymore, but if we don't keep it in pending_pools, where are we going to keep it? Or should we just perform try_upgrade on the list of tuples immediately? I see what you mean here.

nlordell: My point is that we can create the `WeighedPool` directly from the `PoolCreated` event.

nlordell: > Or should we just perform try_upgrade on the list of tuples immediately? I see what you mean here.

:+1:

nlordell:
> I would insist that the PR can't really be split up into these chunks.

If we want to have a `TokenPair -> Pool` mapping, then I agree.

shared/src/balancer_event_handler.rs
bh2smith: Ok. Yes we can in fact we can do all this directly inside insert_events instead of needing `try_upgrade`

shared/src/balancer_event_handler.rs
bh2smith: Cool idea.

bh2smith:
> If we want to have a TokenPair -> Pool mapping, then I agree.


Yes, @nlordell, This is the content of the followup PR for pool fetching.

```rust
pub fn pools_containing_pair(&self, token_pair: TokenPair) -> HashSet<RegisteredPool> 
```

shared/src/balancer_event_handler.rs
```diff
@@ -3,195 +3,140 @@ use crate::{
...
-                for token in weighted_pool.tokens {
-                    self.pools_by_token
-                        .entry(token)
-                        .or_default()
-                        .insert(pool_id);
-                }
+    async fn insert_events(&mut self, events: Vec<(EventIndex, PoolCreated)>) -> Result<()> {
+        for (index, creation) in events {
+            let weighted_pool =
+                WeightedPool::from_event(index.block_number, creation, &*self.data_fetcher).await?;
```
bh2smith, line 114: In a follow up PR I would propose that these be parallelized, but I would like to keep the changes here contained to the storing of the weighted pools.

bh2smith: After performing some naive benchmarks, I see no reason to believe that our data fetching is or will ever be a bottle neck even when it is not batched or parallelized. In fact, from what we see here (of course this is only data from a product which is not yet live), there would be no improvement gained from batched our requests. Only when several pools are deployed within the same 100 (25 minutes w. 15 second block times) blocks can we expect to see any marginal performance gain from batching requests.

I have found from a single simulation contained in the data logs below that fetching token data for 2 pools and syncing events for about 500 blocks takes < 0.4 seconds (including all other processes that the orderbook API is running in parallel). Only when several pools are deployed within the same 100 blocks (25 minutes w. 15 second block times) can we expect to see any marginal performance gain from batching requests. 

Say, for example that 100 pools are deployed in 25 minutes... then it is expected to take 20 seconds to process and store these events (but of course this would happen without notice at different times during that 25 minute interval).

## Summary of finding from logs:

Spans over:
    1.5 months (April 21 to June 9) - 316_781 blocks

Total time to sync all pools: 8 minutes.

## Points of interest:
- There is only 2 occurrences of > 1 pool created over a period of 1000 block (and these are the real time gaps).
- The real time gap here is not with the pool data fetching, but merely with event sync
      - Between 11 and 12 (128_443 blocks) we have a time gap of 2 minutes.
      - Between 1 and 3: total time gap of 0.4 seconds to fetch the pool info for 2 new pools.

## Comparison with Uniswap V2:

- There are 41_928 pools total. (dune query: `select count(*) from uniswap_v2."Factory_evt_PairCreated"`)
- Total processing time in terms of data fetching (plus 500 block sync) 41_928 * 0.2 (seconds) = 2.3 hours
- For a cold start, this is not ideal!
- However, we can easily handle 5000 pools in 16 minutes.
By this point, I would imagine that we can read/write to an from file to to avoid such issues.

Of course, when comparing with UniswapV2, this looks terrible (especially for a cold start). However, by the time BalancerV2 reaches this magnitude, it should be fairly straightforward to feed a backup of this data into the program at runtime.

## File Size: 
The current amount of data when stored as JSON (without white space or line breaks) is 18kb (2kb compressed). For 20 pools, this makes an average of 1kb per pool. The file store size is then linear with the number of pools. So even when/if balancer reaches the magnitude of uniswapV2 this file should not exceed 42 mb (or 4.2 mb compressed). See file [BalancerPools.txt](https://github.com/gnosis/gp-v2-services/files/6634167/BalancerPools.txt) and compressed [BalancerPools.zip](https://github.com/gnosis/gp-v2-services/files/6634178/BalancerPools.zip)


<details><summary>raw data from naive benchmark</summary>
2021-06-10T19:29:39.822Z Last event block 12285749 - Balancer Pool State: 1  pools in store
2021-06-10T19:29:40.201Z Last event block 12286276 - Balancer Pool State: 3  pools in store
2021-06-10T19:29:57.191Z Last event block 12297008 - Balancer Pool State: 4  pools in store
2021-06-10T19:30:00.055Z Last event block 12299382 - Balancer Pool State: 5  pools in store
2021-06-10T19:30:05.540Z Last event block 12303593 - Balancer Pool State: 6  pools in store
2021-06-10T19:30:06.723Z Last event block 12304403 - Balancer Pool State: 7  pools in store
2021-06-10T19:30:23.915Z Last event block 12318074 - Balancer Pool State: 8  pools in store
2021-06-10T19:30:49.069Z Last event block 12331542 - Balancer Pool State: 9  pools in store
2021-06-10T19:30:54.249Z Last event block 12331765 - Balancer Pool State: 10 pools in store
2021-06-10T19:32:16.895Z Last event block 12376911 - Balancer Pool State: 11 pools in store
2021-06-10T19:34:51.197Z Last event block 12505354 - Balancer Pool State: 12 pools in store
2021-06-10T19:34:54.090Z Last event block 12509636 - Balancer Pool State: 13 pools in store
2021-06-10T19:36:11.955Z Last event block 12563369 - Balancer Pool State: 14 pools in store
2021-06-10T19:36:27.052Z Last event block 12567967 - Balancer Pool State: 15 pools in store
2021-06-10T19:37:14.006Z Last event block 12595481 - Balancer Pool State: 16 pools in store
2021-06-10T19:37:15.167Z Last event block 12595681 - Balancer Pool State: 17 pools in store
2021-06-10T19:37:25.202Z Last event block 12602530 - Balancer Pool State: 20 pools in store
</details>

nlordell (approved):
Looks good to me.

> I see no reason to believe that our data fetching is or will ever be a bottle neck even when it is not batched or parallelized.

I'm not sure I agree. From the benchmarking it showed that most of the time querying events from the node. However as the number of pools grows, then the time spent querying static pool data will also increase and may impact the total sync time. That being said, it doesn't seem to be an issue right now.

However, if we implement faster "cold starting" (with the Graph for example), then I agree that I don't see this becoming an issue.

bh2smith:
I will get started on using The Graph for cold starts immediately after we have landed #698 as this will be useful to have in place so not to block @fedgiac from his ongoing task.

PR #697, opened 2021-06-08
Single Order Solver abstraction

fleupold:
Preparing for the Paraswap solver, I noticed that a common abstraction will likely be "sequential" solvers that solve the entire set of liquidity individually (e.g Matcha, Balancer SOR, Paraswap, 1Inch).

This PR therefore extracts the split logic into a separate component. In the future this component could also do logic like time management (extrapolating how many more orders can be attempted).

We might also want to pass the remaining liquidity to the SingleOrderSolving trait, however since it's not yet needed we can cross that bridge later.

Also note, that we are now fetching available protocol on every fetch. If this is an issue, we could either start caching the protocols, or extend the trait with some `prepare` method that can return an associated `Context` type which we would pass in into each individual run of the SingleOrderSolving.

### Test Plan
CI (right now this component is too dumb to justify unit tests, as the only logic - extract limit orders from liquidity - is enforced by the compiler.


bh2smith (approved):
Looks good, excited to see what's comes next!

vkgnosis (approved):

solver/src/single_order_solver.rs
```diff
@@ -0,0 +1,80 @@
...
+#[async_trait::async_trait]
+impl Solver for SingleOrderSolver {
+    async fn solve(&self, liquidity: Vec<Liquidity>, _gas_price: f64) -> Result<Vec<Settlement>> {
+        let mut orders = liquidity
+            .into_iter()
+            .filter_map(|liquidity| match liquidity {
+                Liquidity::Limit(order) => Some(order),
+                _ => None,
+            })
+            .collect::<Vec<_>>();
```
bh2smith, line 46: I know this was the point of the single order solver to only operate on actual limit orders. However, I am now struck with the question "how can a single order be matched without any other liquidity sources to consider?" Is this single order solver's purpose entirely for matching orders directly with the settlement contract's reserves? Or is this intended to be for COWs? If the answer is the latter, I don't see how there can be a COW with only a single order.

bh2smith (approved):
Looks really nice. I have included one question with some context, but also realized this is meant to be used for the 1inch and soon paraswap solvers.

solver/src/single_order_solver.rs
fleupold: I think it would make sense to pass the remaining liquidity as an additional argument to the SingleOrderSolver. However, since it's not used now (we are just invoking solvers that have knowledge of off-chain liquidity themselves) I didn't do it just yet.

solver/src/single_order_solver.rs
```diff
@@ -0,0 +1,80 @@
...
+    fn name(&self) -> &'static str;
+}
+
+/// Maximum number of sell orders to consider for settlements.
+///
+/// This is mostly out of concern to avoid rate limiting and because
+/// requests may a non-trivial amount of time.
+const MAX_SETTLEMENTS: usize = 5;
+
+pub struct SingleOrderSolver {
```
nlordell, line 26: ubernit: I think in this case, it makes sense to make this generic on the single order solving type.

nlordell (approved):
Nice abstraction! Only minor nit is that this seems like a good place to use generics instead of a boxed trait object. Fine to merge as is though.

PR #698, opened 2021-06-08
Balancer Weighted Pool Fetching

bh2smith:
One of many steps towards #603 

As a follow up to #696 we introduce 

- struct `FetchedWeightedPool`
- struct `BalancerPoolFetcher`
- trait `WeightedPoolFetching`
and implement `WeightedPoolFetching` for `BalancerPoolFetcher`

In doing so, this introduces a minor change to include a helper method `pools_containing_pair` for `BalancerPoolStore` which was renamed here from `BalancerPools`

Furthermore, since this has become quite an extensive suite of components we create a new sub-module in the shared crate called `balancer`.

### Test Plan

- Unit test `pools_containing_token_pairs`
- Unit test `fetch`


shared/src/pool_fetching.rs
```diff
@@ -243,18 +243,18 @@ struct FetchedPool {
...
+// Node errors should be bubbled up but contract errors should lead to the pool being skipped.
+pub fn handle_contract_error<T>(result: Result<T, MethodError>) -> Result<Option<T>> {
+    match result {
+        Ok(t) => Ok(Some(t)),
+        Err(err) => match EthcontractErrorType::classify(&err) {
+            EthcontractErrorType::Node => Err(err.into()),
+            EthcontractErrorType::Contract => Ok(None),
+        },
     }
+}
```
bh2smith, line 255: Moving this out to use in `balancer/pool_fetching.rs` as well.

shared/src/balancer/pool_fetching.rs
```diff
@@ -0,0 +1,80 @@
...
+use contracts::BalancerV2Vault;
+use ethcontract::batch::CallBatch;
+use ethcontract::errors::MethodError;
+use ethcontract::{BlockId, Bytes, H160, U256};
+use itertools::Itertools;
+
+pub struct FetchedWeightedPool {
+    pub pool: WeightedPool,
+    /// getPoolTokens returns (Tokens, Balances, LastBlockUpdated)
+    pub reserves: Result<(Vec<H160>, Vec<U256>, U256), MethodError>,
```
fleupold, line 17: This should probably not be of type Result (but instead the values directly, and we just don't have a pool if the reserves fail).

I think this struct should be named WeightedPool (the other one is really just static metadata about the pool). Also maybe we can expose the important information in a flat data structure (right now tokens are e.g. exposed twice). Maybe @fedgiac can already give some insights on how this struct would be most conveniently used for the math he is working on?

shared/src/balancer/event_handler.rs
```diff
@@ -155,7 +156,30 @@ impl PoolSpecialization {
     }
 }
 
-impl BalancerPools {
+impl BalancerPoolStore {
```
fleupold, line 159: nano-naming-nit (since you are already touching it). Maybe this is more a PoolRegistry?

shared/src/balancer/event_handler.rs
```diff
@@ -48,11 +49,11 @@ pub struct TokensRegistered {
 
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
 pub struct WeightedPool {
-    pool_id: H256,
-    pool_address: H160,
+    pub pool_id: H256,
+    pub pool_address: H160,
+    pub normalized_weights: Vec<U256>,
+    pub specialization: PoolSpecialization,
```
fleupold, line 55: Shouldn't the other fiels also be accessible?

shared/src/balancer/pool_fetching.rs
bh2smith: RIght, this was really only meant to be a temp struct that is used before the results are handled. This struct has since been renamed `PreFetchedWeighted` pool and made non-public. The recent update has its reserves field as just `Vec<U256>`. 

On your other point, I will see how I can rename the other weighted pool to accommodate the rename here and flatten the struct.

shared/src/balancer/event_handler.rs
bh2smith: I suppose so, I was only trying to edit the parts that I needed for this PR and we will also find that soon this struct is only for internal use anyway.

shared/src/balancer/event_handler.rs
bh2smith: Should we also eliminate the term `Balancer` from this struct given that it is now in its own "balancer" module

shared/src/balancer/pool_fetching.rs
bh2smith: I have included a doc comment on this struct:

```
/// only an internal temporary struct used during pool fetching to handle any errors when fetching.
```

I hope this helps clear it up. Will refactor and flatted the publicly exposed struct for @fedgiac.

fedgiac: In terms of usability, a HashMap would probably be best, since starting from in and out tokens we need the in and out balances for only these tokens.

bh2smith: So something like 
```rust
reserves: `HashMap<Token, PoolBalance>`
```

shared/src/balancer/event_handler.rs
```diff
@@ -155,7 +169,30 @@ impl PoolSpecialization {
     }
 }
 
-impl BalancerPools {
+impl PoolRegistry {
+    // Since all the fields are private, we expose helper methods to fetch relevant information
+    pub fn pools_containing_pair(&self, token_pair: TokenPair) -> HashSet<RegisteredPool> {
```
bh2smith, line 174: I have decided to return HashSet here since All fields in `RegisteredPool` are hashable and this makes the unit tests more deterministic on this function's output.

bh2smith:
Closing because this relies on Event Listener from the Vault Contract which is known have issues (too many events, doesn't provide specific pool type)

shared/src/balancer/pool_fetching.rs
```diff
@@ -0,0 +1,149 @@
...
+struct FetchedWeightedPool {
+    pool_data: RegisteredWeightedPool,
+    /// getPoolTokens returns (Tokens, Balances, LastBlockUpdated)
+    reserves: Result<(Vec<H160>, Vec<U256>, U256), MethodError>,
+}
+
+fn handle_results(results: Vec<FetchedWeightedPool>) -> Result<Vec<WeightedPool>> {
+    results
+        .into_iter()
+        .try_fold(Vec::new(), |mut acc, fetched_pool| {
```
nlordell, line 110: fancy!

shared/src/balancer/event_handler.rs
```diff
@@ -30,24 +32,24 @@ pub struct PoolCreated {
...
-    pool_address: H160,
-    tokens: Vec<H160>,
-    normalized_weights: Vec<U256>,
+#[derive(Clone, Debug, Default, Eq, PartialEq, Hash)]
+pub struct RegisteredWeightedPool {
+    pub(crate) pool_id: H256,
+    pub(crate) pool_address: H160,
+    pub(crate) tokens: Vec<H160>,
+    pub(crate) normalized_weights: Vec<U256>,
     block_created: u64,
```
vkgnosis: Why not have everything pub? This seems like the standard "just data" struct where everything could be pub because we return it from other public functions in this file.

nlordell (approved):
Looks very nice!

shared/src/balancer.rs
```diff
@@ -0,0 +1,2 @@
+pub mod event_handler;
+pub mod pool_fetching;
```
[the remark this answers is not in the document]
bh2smith: I bet you're happy I went this this instead of `mod.rs` I am considerate of your concerns with having too many mod.rs files. Although I am not sure why you would ever be searching for a mod.rs file.

shared/src/balancer/event_handler.rs
```diff
@@ -468,4 +509,98 @@ mod tests {
...
+        //     },
+        //     pool_ids[1] => WeightedPool {
+        //         tokens: vec![tokens[1], tokens[2]],
+        //         ..other fields
+        //     }
+        //     pool_ids[2] => WeightedPool {
+        //         tokens: vec![tokens[2]],
+        //         ..other fields
+        //     }
+        // };
```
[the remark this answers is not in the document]
bh2smith: Not dead code, its pseudo code so that the reader doesn't have to think about what the loop is doing.

shared/src/balancer/pool_fetching.rs
bh2smith: I borrowed this from @vkgnosis.

shared/src/balancer/event_handler.rs
```diff
@@ -98,20 +100,61 @@ impl PoolDataFetching for Web3 {
...
+                    .expect("failed iterating over known pools")
+                    .clone()
+            })
+            .collect::<HashSet<RegisteredWeightedPool>>()
+    }
+
+    pub fn pools_containing_token_pairs(
+        &self,
+        token_pairs: HashSet<TokenPair>,
+    ) -> HashSet<RegisteredWeightedPool> {
```
vkgnosis, line 142: I would like to make it clear in a documentation comment whether this returns
1. all pools that contain any of the tokens(current implementation)
2. all pools that contain all of the tokens

shared/src/balancer/event_handler.rs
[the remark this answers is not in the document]
nlordell: :+1:

shared/src/balancer/event_handler.rs
bh2smith: I have included a doc comment on both methods `pools_containing_token_pairs`and `pools_containing_token_pair`

vkgnosis (approved):

bh2smith:
@fedgiac - with this part of the integration landed, you should be able to connect your pool arithmetic to the WeightedPool defined and fetched in `shared/src/balancer/pool_fetching.rs`

PR #699, opened 2021-06-08
Fixes call to get current utc time.

marcovc:
Fixes a mistake in querying utc time (coincidently when I tested this, the current code was returning the right thing :) . Sorry about that!

### DIFF
--- a/solver/src/http_solver.rs
+++ b/solver/src/http_solver.rs
@@ -320,7 +320,7 @@ impl HttpSolver {
     }
 
     pub fn generate_instance_name(network_id: &str, chain_id: u64) -> String {
-        let now = chrono::offset::Utc::now();
+        let now = chrono::Utc::now();
         format!(
             "{}_{}_{}",
             now.to_string().replace(" ", "_"),


solver/src/http_solver.rs
```diff
@@ -320,7 +320,7 @@ impl HttpSolver {
     }
 
     pub fn generate_instance_name(network_id: &str, chain_id: u64) -> String {
-        let now = chrono::offset::Utc::now();
+        let now = chrono::Utc::now();
```
vkgnosis, line 323: Does this really fix anything? From looking at the chrono code I believe this two calls are the same thing.

marcovc: I'm slightly confused about the time in rust. I've added a log message (in other PR) to print the name of the instance I'm creating with the results from this function. I just saw this:

```
2021-06-09T08:43:13.508Z  INFO solver::http_solver: http solver instance name is 2021-06-09_08:26:22.218115801_UTC_Ethereum_1
```

Note the discrepancy between the log time and the instance name. This is the first time I see a discrepancy of minutes (is even more puzzling) but I was seeing discrepancy of hours before. While hovering over this function in vscode, I got the impression that it was computing the difference between local time and utc time:

```
chrono
mod offset

The time zone, which calculates offsets from the local time to UTC.
```

But now I see the problem is something else.

vkgnosis: I see what the problem is. `instance_name` is set in `struct SolverConfig`. This is only created once and never updated so the instance name is always this exact string and it is set to the current time when the Solver struct is instantiated. What we want is to always use the current time when solving.

vkgnosis: I would in Http::solver send call generate_instance_name and then add this to the url query part in line 274. You can see how `Config` adds stuff to the query. Config was meant as static global configuration.

marcovc: oh... of course, now everything makes sense. Will close this and open another PR with the fix. Thanks!

marcovc:
No need to fix anything.

PR #702, opened 2021-06-09
Balancer Pool Fetching Part III - Pool Fetching (Alternative Option)

bh2smith:
As an alternative option to #696 which was fetching pools weights at event storing time we incorporate the weight fetching at the time of balance/reserves fetching. If this becomes the accepted means of approach, this PR supersedes both #696 and #698. Unfortunately now that we are combining the work of two PRs, it has become a bit larger I could reduce the diff here by upgrading the ethcontracts repo in a different PR.

I have since created a new branch to vendor contracts outside of this PR. Another thing I can do to make this diff much smaller to do the renaming of structs before hand. I will do this in the meantime.

As a follow up to #696 we introduce

- struct `PoolTokenState` (for balances and weights) and `WeightedPool`
- struct `BalancerPoolFetcher` consisting of the `PoolRegistry`, `Web3` and the `BalancerVault` contract.
- trait `WeightedPoolFetching`
and implement `WeightedPoolFetching` for `BalancerPoolFetcher`

In doing so, this introduces a minor change to include a helper method `pools_containing_pair` for `PoolRegistry`.

### Test Plan

- Unit test pools_containing_pair
- Unit test fetch method return type handler.


shared/src/balancer/event_handler.rs
```diff
@@ -572,7 +607,83 @@ mod tests {
                 block_created: new_event_block
             }
         );
+
```
bh2smith, line 610:
```suggestion

```

shared/src/pool_fetching.rs
```diff
@@ -243,18 +243,18 @@ struct FetchedPool {
...
+// Node errors should be bubbled up but contract errors should lead to the pool being skipped.
+pub fn handle_contract_error<T>(result: Result<T, MethodError>) -> Result<Option<T>> {
+    match result {
+        Ok(t) => Ok(Some(t)),
+        Err(err) => match EthcontractErrorType::classify(&err) {
+            EthcontractErrorType::Node => Err(err.into()),
+            EthcontractErrorType::Contract => Ok(None),
+        },
     }
+}
```
bh2smith, line 255: Moving this out of `handle_results` and making public to be used in balancer pool fetching.

shared/src/balancer/pool_fetching.rs
```diff
@@ -0,0 +1,169 @@
...
+        &self,
+        token_pairs: HashSet<TokenPair>,
+        at_block: Block,
+    ) -> Result<Vec<WeightedPool>> {
+        let mut batch = CallBatch::new(self.web3.transport());
+        let block = BlockId::Number(at_block.into());
+        let futures = token_pairs
+            .into_iter()
+            .flat_map(|pair| self.pool_data.pools_containing_pair(pair))
+            .unique_by(|pool| pool.pool_id)
```
bh2smith, line 77: This is the reason for the `itertools` dependency!

shared/src/balancer/event_handler.rs
```diff
@@ -125,6 +138,29 @@ impl PoolSpecialization {
 }
 
 impl PoolRegistry {
+    // Since all the fields are private, we expose helper methods to fetch relevant information
+    pub fn pools_containing_pair(&self, token_pair: TokenPair) -> HashSet<RegisteredPool> {
+        let empty_set = HashSet::new();
+        let pools_0 = self
+            .pools_by_token
+            .get(&token_pair.get().0)
+            .unwrap_or(&empty_set);
```
fleupold, line 147: nit: unwrap_or_default should do.

shared/src/balancer/pool_fetching.rs
```diff
@@ -0,0 +1,169 @@
...
+    results
+        .into_iter()
+        .try_fold(Vec::new(), |mut acc, fetched_pool| {
+            let balances = match handle_contract_error(fetched_pool.reserves)? {
+                // We only keep the balances entry of reserves query.
+                Some(reserves) => reserves.1,
+                None => return Ok(acc),
+            };
+            let weights = match handle_contract_error(fetched_pool.weights)? {
+                // We only keep the balances entry of reserves query.
```
fleupold, line 129: copy pasta error

shared/src/balancer/pool_fetching.rs
```diff
@@ -0,0 +1,169 @@
...
+        let mut batch = CallBatch::new(self.web3.transport());
+        let block = BlockId::Number(at_block.into());
+        let futures = token_pairs
+            .into_iter()
+            .flat_map(|pair| self.pool_data.pools_containing_pair(pair))
+            .unique_by(|pool| pool.pool_id)
+            .map(|weighted_pool| {
+                let pool_contract =
+                    BalancerV2WeightedPool::at(&self.web3, weighted_pool.pool_address);
+                let weights = pool_contract
```
fleupold, line 81: Maybe we could fetch this via a component which can do some caching per pool id (given that the values are immutable) - maybe a good to do

shared/src/balancer/pool_fetching.rs
fleupold: Maybe the pool_data interface should already allow you to specify a set of pairs and give you unique pool IDs already?

shared/src/balancer/pool_fetching.rs
```diff
@@ -0,0 +1,169 @@
...
+                        weights: weights.await,
+                    }
+                }
+            })
+            .collect::<Vec<_>>();
+        batch.execute_all(MAX_BATCH_SIZE).await;
+
+        let mut results = Vec::new();
+        for future in futures {
+            results.push(future.await);
```
fleupold, line 105: nit: maybe comment that given the batch is executed these awaits should resolve immediately? Otherwise a future maintainer might wonder why they are not parallelised.

shared/src/balancer/event_handler.rs
```diff
@@ -45,11 +46,23 @@ pub struct TokensRegistered {
...
-    specialization: PoolSpecialization,
-    block_created: u64,
+    pub pool_id: H256,
+    pub pool_address: H160,
+    pub specialization: PoolSpecialization,
+    pub tokens: Vec<H160>,
+    pub block_created: u64,
+}
+
+impl RegisteredPool {
```
fleupold, line 56: Should this be only in #[cfg(test)] ? Otherwise you could also derive Default on RegisteredPool

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2Fm2Q7FEc0bEr4I%2Fgiphy.gif&token=brent" data-gitmeme-token="brent"></a>

shared/src/balancer/event_handler.rs
bh2smith: Can't derive default because of specialization and tokens but I'll move it into a test cuz that's where it should be.

shared/src/balancer/event_handler.rs
nlordell: Does it work for references?

shared/src/balancer/pool_fetching.rs
```diff
@@ -0,0 +1,169 @@
...
+        token_pairs: HashSet<TokenPair>,
+        at_block: Block,
+    ) -> Result<Vec<WeightedPool>> {
+        let mut batch = CallBatch::new(self.web3.transport());
+        let block = BlockId::Number(at_block.into());
+        let futures = token_pairs
+            .into_iter()
+            .flat_map(|pair| self.pool_data.pools_containing_pair(pair))
+            .unique_by(|pool| pool.pool_id)
+            .map(|weighted_pool| {
```
nlordell, line 78: This still has the issue that we don't know that the pools here are, in fact, `WeightedPool`s and that they may be some other pool implementation.

Just because they have a `normalizedWeights` contract method doesn't mean that they will follow the logic from the weighted pool that we are expecting.

I added a comment in #696 mentioning this as well, but I think we should instead listen to events from the `WeightedPoolFactory` instead of the `Vault` for pool creation. This way we are guaranteed to have the weighted pools that we expect.

bh2smith: Wouldn't it also be appropriate to simply filter out those events which are not in correspondence with weighted pools? Although I we would also solve the problem of having to filter out all the other events coming from the vault. I will rework the first implementation then.

nlordell: > This may mean that we have to make a web3 query to fetch the tokens

I think that its fine. This is also part of the "immutable" part of the the pool, so we can fetch it at the same time as we do the normalized weights and cache it permanently.

shared/src/balancer/pool_fetching.rs
bh2smith: Relevant changes for this will be found in #698.

shared/src/balancer/pool_fetching.rs
bh2smith: No longer relevant as weights are stored in the `PoolRegistry`now.

shared/src/balancer/pool_fetching.rs
bh2smith: Also will be addressed in #698

shared/src/balancer/event_handler.rs
bh2smith: No longer need `PoolSpecialization` so we can derive Default.

shared/src/balancer/pool_fetching.rs
bh2smith: No longer relevant as the weights are fetched and stored by our Event Handler.

bh2smith:
Closing this in favour of #698 which is currently being rebased onto #696. All suggestions for improvement will be addressed there.

PR #705, opened 2021-06-09
[Easy] Balancer Pool Fetching Part II - Small Refactor

bh2smith:
Just trying to separate #702 into smaller chunks.

We move `balancer_event_handler` into own module in preparation for more to come. 

Renaming structs `BalancerPools` -> `PoolRegistry` and `WeightedPool` -> `RegisteredPool` since these are more suitable names for non-public facing interfaces.

Small rework of `into_pool` method returning `Result` instead of `Option`.

### Test Plan
Refactored unit tests to accommodate change of `into_pool` method.


shared/src/balancer/event_handler.rs
```diff
@@ -63,29 +63,33 @@ pub struct WeightedPoolBuilder {
...
 
 impl WeightedPoolBuilder {
-    fn into_pool(self) -> Option<WeightedPool> {
-        if let (Some(pool_registration), Some(tokens_registration)) =
-            (self.pool_registration, self.tokens_registration)
-        {
-            return Some(WeightedPool {
+    fn into_pool(self) -> Result<RegisteredPool> {
+        if let (Some(pool_registration), Some(tokens_registration)) = (
+            self.pool_registration.clone(),
```
fleupold, line 68: why is this extra clone necessary?

shared/src/balancer/event_handler.rs
```diff
@@ -120,22 +124,22 @@ impl PoolSpecialization {
...
-                tracing::info!("Upgrading Pool Builder with id {:?}", pool_id);
-                self.pools.insert(pool_id, weighted_pool.clone());
-                self.pending_pools.remove(&pool_id);
-                for token in weighted_pool.tokens {
-                    self.pools_by_token
-                        .entry(token)
-                        .or_default()
-                        .insert(pool_id);
-                }
+            let weighted_pool = pool_builder.into_pool()?;
```
fleupold, line 130: won't this return if a single pool fails to be turned into a weighted pool (also is the variable name outdated now?) ? Maybe add a unit test for this if not too much hassle? 

I guess this is fine, because we expect events to be part of the same transaction, but I remember that we used to be able to handle the case even if they were split across multiple pages.

shared/src/balancer/event_handler.rs
```diff
@@ -120,22 +124,22 @@ impl PoolSpecialization {
     }
 }
 
-impl BalancerPools {
-    fn try_upgrade(&mut self) {
+impl PoolRegistry {
+    fn try_upgrade(&mut self) -> Result<()> {
```
fleupold, line 128: nit: maybe some docs on when this should error?

fleupold (commented):
Looks good, just a questions about the erroring logic. Maybe you could explain _why_ we need to return a Result now?

shared/src/balancer/event_handler.rs
bh2smith: This is because of the use of self in the error log.

shared/src/balancer/event_handler.rs
bh2smith: Now that you mention it I do slight prefer returning the option and logging an error when upgrade fails.

bh2smith:
Closing in favour of #696 and #698

shared/src/balancer/event_handler.rs
bh2smith: Will address and test this in #696.

PR #706, opened 2021-06-09
Refactor pool cache to be generic to any kind of on-chain data

nlordell:
In preparation for the Balancer V2 pools, this PR refactors the `PoolCache` into a generic `RecentBlockCache` that applies the same caching semantics to any arbitrary key-value pairs of on-chain data. This would allow us to easily reuse this cache for Balancer V2 weighted pools :tada:.

In order to make the recent block cache completely generic on keys, we needed to add some additional traits for:
- The `CacheKey` (as it needs to be able to produce the `first_ord` value - absolute first value in ordering).
- The `CacheRetrieving` (to describe how cached data gets fetched).
- The `CacheMetrics` (to add caching metrics to the recent block cache - while also allowing for different caches to keep their own metrics separate by keeping the `PoolCacheMetrics` trait).

The code was has minimal modifications, and most changes are a product of using generic `K` key and `V` value types. In order to make this PR easier to review, I split this into this PR that replaces the implementation directly in the `pool_cache.rs` file, while putting the existing pool cache implementation into the `pool_cache_new.rs` file using `#[path = "..."]` renaming to keep module names as they are supposed to be. This is done in order to coax the Github diff UI to show a better diff and make it easier to review. The PR #707 just moves the files without modification into their final places

### Test Plan

No major logic changes - tests ported for new generic type. CI should catch any issues.


shared/src/pool_cache_new.rs
```diff
@@ -0,0 +1,73 @@
...
+    pool_fetching::{Pool, PoolFetching},
+    recent_block_cache::{
+        Block, CacheConfig, CacheFetching, CacheKey, CacheMetrics, RecentBlockCache,
+    },
+};
+use anyhow::Result;
+use model::TokenPair;
+use std::{collections::HashSet, sync::Arc};
+
+pub trait PoolCacheMetrics: Send + Sync {
```
fleupold, line 13: Do we need this extra trait? Couldn't we work on the original one?

fleupold (approved):
Very useful abstraction!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2FxUA7b2KE79co5JD9N6%2Fgiphy-downsized-medium.gif&token=impressed" data-gitmeme-token="impressed"></a>

vkgnosis (approved):

shared/src/lib.rs
```diff
@@ -17,9 +17,12 @@ pub mod maintenance;
 pub mod metrics;
 pub mod network;
 pub mod pool_aggregating;
+#[path = "pool_cache_new.rs"]
 pub mod pool_cache;
 pub mod pool_fetching;
 pub mod price_estimate;
+#[path = "pool_cache.rs"]
+pub mod recent_block_cache;
```
bh2smith, line 25: cute!

shared/src/pool_cache.rs
```diff
@@ -1,125 +1,176 @@
...
+//! simultaneously and some work is wasted. This is unlikely to happen in practice and the value is
+//! going to be cached the next time it is needed.
+//!
+//! When entries are requested we mark all those entries as recently used which potentially evicts
+//! other entries from the lru cache. Cache misses are fetched and inserted into the cache.
+//! Then when the automatic update runs the next time, we request and cache all recently used
+//! entries. For some consumers we only care about the "recent" state of the entries. So we can
+//! return any result from the cache even if it comes from previous blocks.
+//!
+//! On the other hand for others we need to fetch on-chain data at exact blocks which is why we keep
```
bh2smith, line 18: NIT: I think this new paragraph is meant to stay with the last sentence from the previous. Otherwise, I am not sure what "on the other hand" is referring to.

shared/src/pool_cache.rs
```diff
@@ -1,125 +1,176 @@
...
-pub struct PoolCache {
-    mutexed: Mutex<Mutexed>,
+/// Caches on-chain data for a specific number of blocks and automatically updates the N most
+/// recently used entries automatically when a new block arrives.
+pub struct RecentBlockCache<K, V, F, M>
+where
+    K: CacheKey<V>,
+    F: CacheFetching<K, V>,
+{
+    mutexed: Mutex<Mutexed<K, V>>,
```
bh2smith, line 78: Holy moly! Hats off for this.

shared/src/pool_cache.rs
```diff
@@ -130,116 +181,118 @@ impl PoolCache {
...
 #[derive(Debug)]
-struct Mutexed {
-    recently_used: LruCache<TokenPair, ()>,
-    // For quickly finding at which block a pair is cached.
-    cached_most_recently_at_block: HashMap<TokenPair, u64>,
+struct Mutexed<K, V>
+where
+    K: CacheKey<V>,
+{
+    recently_used: LruCache<K, ()>,
```
bh2smith, line 250: I guess `LruCache<K, ()>` is like a map with only keys and the values don't matter.

bh2smith (approved):
Very elaborate and super cool. Nice work here.

shared/src/pool_cache.rs
nlordell: Correct, I think its because there is no `LruSet<T>`.

PR #708, opened 2021-06-10
ParaswapAPI

fleupold:
First step towards a paraswap solver. This PR implements the Paraswap API interactions as documented [here](https://developers.paraswap.network/api/get-rate-for-a-token-pair).

Except for the decoding of src and dest amount, there is little to no logic here (just defining the datastructures and unit tests makes up for a lot of lines)

### Test Plan
Unit tests & e2e test


solver/src/paraswap_solver/api.rs
```diff
@@ -0,0 +1,420 @@
...
+#[async_trait::async_trait]
+pub trait ParaswapApi {
+    async fn price(&self, query: PriceQuery) -> Result<PriceResponse>;
+    async fn transaction(
+        &self,
+        query: TransactionBuilderQuery,
+    ) -> Result<TransactionBuilderResponse>;
+}
+
+struct DefaultParaswapApi {
```
nlordell, line 27: naming-nit: Why not
```suggestion
struct HttpParaswapApi {
```

solver/src/paraswap_solver/api.rs
```diff
@@ -0,0 +1,420 @@
...
+            .expect("unexpectedly invalid URL segment");
+
+        let side = match self.side {
+            Side::Buy => "BUY",
+            Side::Sell => "SELL",
+        };
+
+        url.query_pairs_mut()
+            .append_pair("from", &format!("{:#x}", self.from))
+            .append_pair("to", &format!("{:#x}", self.to))
```
nlordell, line 94: TIL `H160` implements hex formatting.

solver/src/paraswap_solver/api.rs
```diff
@@ -0,0 +1,420 @@
...
+            .append_pair("amount", &self.amount.to_string())
+            .append_pair("side", side)
+            .append_pair("network", "1");
```
nlordell, line 97: Do they support xDAI and Rinkeby? Might be nice to add support for it in the future (doesn't have to be this PR).

solver/src/paraswap_solver/api.rs
```diff
@@ -0,0 +1,420 @@
...
+
+/// A Paraswap API price response.
+#[derive(Clone, Debug, PartialEq)]
+pub struct PriceResponse {
+    pub price_route: Map<String, Value>,
+    pub src_amount: U256,
+    pub dest_amount: U256,
+}
+
+impl TryFrom<Value> for PriceResponse {
```
nlordell, line 110: Why not implement `Deserialize` instead of `TryFrom<Value>`? You should even be able to derive it with `#[serde(with = "u256_decimal")]`.

If a type implements `Deserialize`, then you can also use [`serde_json::from_value()`](https://docs.rs/serde_json/1.0.64/serde_json/fn.from_value.html) to convert a `serde_json::Value` to that type.

nlordell (commented):
Looks good! I would only suggest implementing `Deserialize` instead of `TryFrom<Value>` as it:
- Allows you to directly deserialize from a JSON string
- Also allows you to convert from a `serde_json::Value` using `serde_json::from_value`.

solver/src/oneinch_solver.rs
```diff
@@ -158,51 +156,15 @@ impl Interaction for Swap {
...
-        if sell_orders.len() > MAX_SETTLEMENTS {
-            sell_orders.shuffle(&mut rand::thread_rng());
-        }
-
+impl SingleOrderSolving for OneInchSolver {
+    async fn settle_order(&self, order: LimitOrder) -> Result<Settlement> {
+        ensure!(
+            order.kind == OrderKind::Sell,
+            "1Inch only supports sell orders"
+        );
```
bh2smith, line 165: Any particular reason we are seeing single order solving pop up again here? These changes were already introduced elsewhere. I think you may want to base this on top of that.

solver/src/paraswap_solver/api.rs
```diff
@@ -0,0 +1,420 @@
...
+            .await
+            .context("TransactionBuilderQuery result parsing failed")
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum Side {
+    Buy,
+    Sell,
+}
```
bh2smith, line 61: Could we reuse our `OrderKind` enum from the model crate? Or are we trying to keep this entirely self-contained?

solver/src/paraswap_solver/api.rs
bh2smith: Me too!

bh2smith (approved):
Apart from the fact that I suspect this branch is not based on the right thing, everything looks great!

One small suggestion that we include the link to the the Paraswap API docs in the code here as well

https://developers.paraswap.network/api/get-rate-for-a-token-pair

solver/src/paraswap_solver/api.rs
fleupold: I originally tried implementing my own Deserializer. I thought I could just `deserialize_str` like we already do it [here](https://github.com/gnosis/oba-services/blob/16e30b88f9d89198d4ad607a6f5efec5d61b0365/model/src/h160_hexadecimal.rs#L39-L40) but it complained that it would have to use `deserialize_map`. Then, for a [MapAccess](https://docs.serde.rs/serde/de/trait.MapAccess.html) visitor the code looked so complicated and indirect that I gave up and went for this naive approach. I'd be happy to take a look together with you (or take a pointer to where this has already be done). 

I gotta say serde::Deserialize is definitely not my favorite trait.

solver/src/oneinch_solver.rs
fleupold: Had to rebase after merging the base branch. Should be fixed now

solver/src/paraswap_solver/api.rs
fleupold: They are not directly related, so I wanted to keep them separate (and not couple the Paraswap API with out core data mdoels).

vkgnosis (approved):

solver/src/paraswap_solver/api.rs
nlordell: Implementing `Deserialize` by hand is hard.

> I thought I could just deserialize_str [...] but it complained that it would have to use deserialize_map

This would indicate that your type expects the next value from the deserializer to be a string (which is true for `H160` for example because its represented as a string hex value in JSON), but since this is an object that is being deserialized, the nex value from the deserializer should be a map.

fleupold: Will add the following comment (I though I already mentioned it somewhere, but cannot find it)

> The reason we do not `#[derive(Deserialize)]` is that the structure of `price_route` is opaque and may change over time. In order to get the transaction data from the API, it expects `price_route` to be echoed back in the exact form as it was sent in the price estimation response. We therefore leave it as an opaque mapping and only extract the effective in and out amounts from the struct (which we hope will remain in the same place). We don't implement `Deserialize` manually as writing a custom MapVisitor leads to convoluted code.

nlordell: I would still advocate for deriving the deserialization code. You can just keep a field with the "raw" value for serialization and deserialization:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=1da79c937b3e0bc135b878be12a06d7c

nlordell (approved):

PR #709, opened 2021-06-10
Paraswap Solver

fleupold:
This PR implements a SingleOrderSolver using the Paraswap API.

### Test Plan
Unit tests

### DIFF
--- a/solver/src/paraswap_solver.rs
+++ b/solver/src/paraswap_solver.rs
@@ -55,13 +55,17 @@ impl<F> std::fmt::Debug for ParaswapSolver<F> {
 #[cfg_attr(test, mockall::automock)]
 #[async_trait::async_trait]
 trait AllowanceFetching: Send + Sync {
-    async fn existing_allowance(&self, token: &ERC20, spender: &H160) -> Result<U256>;
+    async fn existing_allowance(&self, token: H160, spender: H160) -> Result<U256>;
 }
 
 #[async_trait::async_trait]
-impl AllowanceFetching for H160 {
-    async fn existing_allowance(&self, token: &ERC20, spender: &H160) -> Result<U256> {
-        Ok(token.allowance(*self, *spender).call().await?)
+impl AllowanceFetching for GPv2Settlement {
+    async fn existing_allowance(&self, token: H160, spender: H160) -> Result<U256> {
+        let token_contract = ERC20::at(&self.raw_instance().web3(), token);
+        Ok(token_contract
+            .allowance(self.address(), spender)
+            .call()
+            .await?)
     }
 }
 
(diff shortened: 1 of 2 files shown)


vkgnosis (approved):

solver/src/paraswap_solver.rs
```diff
@@ -1 +1,383 @@
...
+    client: Box<dyn ParaswapApi + Send + Sync>,
+}
+
+impl ParaswapSolver<H160> {
+    #![allow(dead_code)]
+    pub fn new(
+        settlement_contract: GPv2Settlement,
+        token_info: Arc<dyn TokenInfoFetching>,
+    ) -> Self {
+        let allowance_fetcher = settlement_contract.address();
```
nlordell, line 38: neat! This could maybe also be added to the 1Inch solver eventually :smile:.

One idea (but feel free to ignore, I think it looks good like this as well) would be to instead:
```
trait AllowanceFetching: Send + Sync {
    async fn existing_allowance(&self, token: &H160, spender: &H160) -> result<U256>;
}
impl AllowanceFetching for GPv2Settlement { ... }
```

This way, you wouldn't need a `web3` instance at all to mock `AllowanceFetching` (as there would be no need to construct a `ERC20` instance to pass to the mock fetching) and make it easier to set mock parameter assertions as `H160: Eq`.

nlordell (approved):

solver/src/paraswap_solver.rs
```diff
@@ -1 +1,383 @@
...
+            dest_token: order.buy_token,
+            src_amount: src_amount_with_slippage,
+            dest_amount: price_response.dest_amount,
+            from_decimals: decimals(&order.sell_token)?,
+            to_decimals: decimals(&order.buy_token)?,
+            price_route: price_response.price_route,
+            user_address: self.settlement_contract.address(),
+            referrer: REFERRER.to_string(),
+        };
+        let transaction = self.client.transaction(transaction_query).await?;
```
bh2smith, line 120: So their API service encodes the transaction for us, or is this something you had implemented in the previous PR?

bh2smith (approved):
Nice. Just a couple little questions inline.

solver/src/paraswap_solver.rs
fleupold: Correct, their API gives us the transaction calldata (same as 1Inch)

solver/src/paraswap_solver.rs
fleupold: Agree that this interface would be more appropriate, however we still need the `ERC20` instance to create the `Erc20ApproveInteraction` approve interaction (so we would still need a web3 instance). Unless we refactor `Erc20ApproveInteraction` to take the ERC20 address and encode the calldata with a dummy web3 (which I believe should work).

nlordell: Makes sense, Lets leave it as is then.

PR #711, opened 2021-06-10
Fix price estimation

vkgnosis:
A user reported a bug that the price we give was wrong. For the pair [SushiToken](https://etherscan.io/token/0x6B3595068778DD592e39A122f4f5a5cF09C90fE2) - [USDC](https://etherscan.io/token/0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48) the 1 sushi should be worth  roughly 10 usdc as seen on the etherscsan links.

```
curl "https://protocol-mainnet.dev.gnosisdev.com/api/v1/markets/0x6B3595068778DD592e39A122f4f5a5cF09C90fE2-0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48/sell/100000000000000000000"
{"amount":"540263517","token":"0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48"}
```

Taking the decimals into account: `(540263517 * 1e-6) / (100000000000000000000 * 1e-18) == 5.4` so indeed the price is off.

When debugging this I noticed that `sell_token_price_in_eth` seemed off because `cost_in_native_token` (which is the final buy amount in the native token) was many orders of magnitudes lower than `tx_cost_in_native_token` when intuitively it should be a little larger as we are selling roughly a 1000 dollars worth of sushi. This made the the gas cost effectively the only criteria so the direct connection was always preferred even with a much worse price.

The reason for this is that `estimate_price` returns different units when calculating with 0 input amount (spot price) versus some input amount. I fixed the documentation and swapped the order of the tokens in the native token spot price calculation and changed the test so that it would fail without this commit.

In a future PR it probably makes sense to make this more clear in the names of the functions and also have a separate spot_price function rather than mixing this behavior in the same estimate_price function.

### Test Plan
new unit test and ran the same curl command locally: 

```
curl "http://localhost:8080/api/v1/markets/0x6B3595068778DD592e39A122f4f5a5cF09C90fE2-0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48/sell/100000000000000000000"
{"amount":"928545308","token":"0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48"}
```

`(928545308 * 1e-6) / (100000000000000000000 * 1e-18) == 9.3`

### DIFF
--- a/shared/src/price_estimate.rs
+++ b/shared/src/price_estimate.rs
@@ -34,7 +34,11 @@ pub enum PriceEstimationError {
 
 #[async_trait::async_trait]
 pub trait PriceEstimating: Send + Sync {
-    // Price is given in how much of sell_token needs to be sold for one buy_token.
+    /// For OrderKind::Sell amount is in sell_token and for OrderKind::Buy in buy_token.
+    /// The resulting price is how many units of sell_token needs to be sold for one unit of
+    /// buy_token (sell_amount / buy_amount).
+    /// If amount is 0 then kind is ignored and the result is how many units of buy token you get
+    /// from selling one unit of sell token (buy_amount / sell_amount).
     async fn estimate_price(
         &self,
         sell_token: H160,
@@ -129,9 +133,6 @@ impl BaselinePriceEstimator {
 
 #[async_trait::async_trait]
 impl PriceEstimating for BaselinePriceEstimator {
-    // Estimates the price between sell and buy token denominated in |sell token| per buy token.
-    // Returns an error if no path exists between sell and buy token.
-    // Incorporates uniswap fee unless amount is 0 in which case it returns the best spot price.
     async fn estimate_price(
         &self,
         sell_token: H160,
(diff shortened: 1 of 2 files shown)


shared/src/price_estimate.rs
```diff
@@ -34,7 +34,11 @@ pub enum PriceEstimationError {
 
 #[async_trait::async_trait]
 pub trait PriceEstimating: Send + Sync {
-    // Price is given in how much of sell_token needs to be sold for one buy_token.
+    /// For OrderKind::Sell amount is in sell_token and for OrderKind::Buy in buy_token.
+    /// The resulting price is how many units of sell_token needs to be sold for one unit of
+    /// buy_token (sell_amount / buy_amount).
+    /// If amount is 0 then kind is ignored and the result is how many units of buy token you get
```
fleupold, line 40: Oh wow, this is pretty bad (and totally my fault when implementing this). Agree that we should probably not handle 0 amounts in the same method or at least make it that they return it in the same unit.

fleupold (approved):
Great 🕵️ work and 🎣

bh2smith (approved):
Really nice find and quick fix here @vkgnosis!

nlordell (approved):
Great catch!

PR #712, opened 2021-06-11
Multi-Contract Event Handling

bh2smith:
As a first step to Receiving and Storing multiple pool types for the Balancer integration, we must first implement some way Streaming Events from multiple contracts into the same Storage. This naive first approach extends the event handler to accept a vector of Event Receivers coupled with some additional context which the EventStoring uses to distinguish where the events came from. This adaption is used for our existing single contract event handlers passing along the contract name as a String.

For event updates, this implementation simply loops over each receiver, and handles the events sequentially. This may be something can could be parallelized perhaps with a mutex but it seems we will be fetching the same blocks multiple times   regardless. Happy to get suggestions on this

### Test Plan
No tests introduced here.

### DIFF
--- a/shared/src/event_handling.rs
+++ b/shared/src/event_handling.rs
@@ -18,10 +18,10 @@ pub struct EventHandler<B, C, S>
 where
     B: BlockRetrieving,
     C: EventRetrieving,
-    S: EventStoring<C::Event>,
+    S: EventStoring<C::Event, String>,
 {
     block_retriever: B,
-    contract: C,
+    named_contracts: Vec<(C, String)>,
     store: S,
     last_handled_block: Option<u64>,
 }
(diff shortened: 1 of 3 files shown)


shared/src/event_handling.rs
```diff
@@ -100,59 +107,66 @@ where
...
-            if !have_deleted_old_events {
-                self.store
-                    .replace_events(unwrapped_events, range.clone())
-                    .await?;
-                have_deleted_old_events = true;
-            } else {
-                self.store.append_events(unwrapped_events).await?;
-            };
+        for (contract, name) in &self.named_contracts {
+            let events = self
```
bh2smith, line 115: This is the big change here and also what makes it naive. However, I don't think we can get around fetching the blocks twice from within this codebase.

shared/src/event_handling.rs
```diff
@@ -33,23 +33,30 @@ where
 /// Databases: might transform, filter and classify which events are inserted
 /// HashSet: For less persistent (in memory) storing, insert events into a set.
 #[async_trait::async_trait]
-pub trait EventStoring<T> {
+pub trait EventStoring<T, U> {
```
nlordell, line 36: I'm not sure I understand why the context here is generic. It seems like it has to be `String` for the `EventHandler` so you can't really implement `EventStoring` for anything other than that.

nlordell (commented):
One concern I have is that this seems to only work for a multiple instances of a single contract `EventRetrieving` type and not multiple instances of _different_ contract `EventRetrieving` types that we would need for Balancer V2 (for the different pool factory contracts).

Why not just use multiple instances of `EventHandler` with a single `EventStoring` instance behind an `Arc<Mutex<...>>`?

shared/src/event_handling.rs
bh2smith: I was going to use an enum PoolType for balancer and then () for GPSettelment. but it doesn't matter now, this approach isn't going to work.

PR #713, opened 2021-06-11
Make driver use pool cache for liquidity

vkgnosis:
For this we turn the cache of the aggregator of uniswap and sushiswapinto the aggregator of the cache of uniswap and the cache of sushiswap so that we have two caches instead of one.
I kept the order book as the single cache of the aggregator because we don't need this there.
This code is likely to change further when we integrate balance pools but I wanted to accomplish this first.

### Test Plan
Existing tests still work as this just moves code around without changing logic.

### DIFF
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -1,4 +1,5 @@
 use contracts::{GPv2Settlement, WETH9};
+use futures::{stream, StreamExt};
 use model::{
     order::{OrderUid, BUY_ETH_ADDRESS},
     DomainSeparator,
@@ -15,6 +16,7 @@ use orderbook::{
 use primitive_types::H160;
 use prometheus::Registry;
 use shared::{
+    amm_pair_provider::AmmPairProvider,
     bad_token::{
         cache::CachingDetector,
         list_based::{ListBasedDetector, UnknownTokenStrategy},
@@ -25,6 +27,7 @@ use shared::{
     maintenance::ServiceMaintenance,
     pool_aggregating::{self, PoolAggregator},
     pool_cache::PoolCache,
+    pool_fetching::{PoolFetcher, PoolFetching},
     price_estimate::BaselinePriceEstimator,
     recent_block_cache::CacheConfig,
     transport::create_instrumented_transport,
@@ -171,8 +174,13 @@ async fn main() {
         .expect("failed to create gas price estimator"),
     );
 
-    let pair_providers =
-        pool_aggregating::pair_providers(&args.shared.baseline_sources, chain_id, &web3).await;
+    let pair_providers: Vec<Arc<dyn AmmPairProvider>> = stream::iter(args.shared.baseline_sources)
+        .then(|source| {
+            let web3 = web3.clone();
+            async move { pool_aggregating::pair_provider(source, chain_id, &web3).await }
+        })
+        .collect()
+        .await;
 
     let mut base_tokens = HashSet::from_iter(args.shared.base_tokens);
     // We should always use the native token as a base token.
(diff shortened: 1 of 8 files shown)


fleupold (approved):
Nice rewiring!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2F5aY2MOdRZ2y8kQfIHH%2Fgiphy.gif&token=wire" data-gitmeme-token="wire"></a>

shared/src/pool_aggregating.rs
```diff
@@ -11,53 +11,35 @@ use std::sync::Arc;
...
     pub enum BaselineSources {
         Uniswap,
         Sushiswap,
     }
 }
 
-pub async fn pair_providers(
-    sources: &[BaselineSources],
+pub async fn pair_provider(
+    source: BaselineSources,
```
bh2smith, line 22: nit: Maybe this should be singular now
```suggestion
    source: BaselineSource,
```

orderbook/src/main.rs
```diff
@@ -207,7 +215,17 @@ async fn main() {
...
+        pool_fetchers: pair_providers
+            .into_iter()
+            .map(|pair_provider| {
+                Arc::new(PoolFetcher {
+                    pair_provider,
+                    web3: web3.clone(),
+                }) as Arc<dyn PoolFetching>
+            })
+            .collect(),
+    };
```
bh2smith, line 228: I know we removed the `from_providers` method, but wouldn't it be nice to keep all this enclosed in some method implemented for pool aggregator? Was there some special reason for bringing it out? It appears to be doing nearly the same thing as before but with an Arc.

solver/src/main.rs
```diff
@@ -300,48 +320,35 @@ async fn main() {
...
+        let router = contracts::SushiswapV2Router02::deployed(&web3)
+            .await
+            .expect("couldn't load deployed sushiswap router");
+        res.push(UniswapLikeLiquidity::new(
+            IUniswapLikeRouter::at(&web3, router.address()),
+            settlement_contract.clone(),
+            base_tokens.clone(),
+            web3.clone(),
+            pool_fetcher.clone(),
+        ));
```
bh2smith, line 351: Given that we are no longer matching on the source here, I would think that when introducing a new source, one could easily forget to include the necessary code here and be confused as to why their code isn't working. Is there any way we could somehow circumvent this or is my concern not actually an issue?

nlordell (approved):
Looks good!

solver/src/main.rs
```diff
@@ -208,38 +209,54 @@ async fn main() {
...
+        number_of_blocks_to_cache: args.shared.pool_cache_blocks,
+        // 0 because we don't make use of the auto update functionality as we always fetch
+        // for specific blocks
+        number_of_entries_to_auto_update: 0,
+        maximum_recent_block_age: args.shared.pool_cache_maximum_recent_block_age,
+        max_retries: args.shared.pool_cache_maximum_retries,
+        delay_between_retries: args.shared.pool_cache_delay_between_retries_seconds,
+    };
+    let pool_caches: HashMap<BaselineSources, Arc<PoolCache>> =
+        stream::iter(args.shared.baseline_sources)
```
bh2smith, line 222: This also seems like something that could be contained in some sort of builder method since it is essentially a map from `BaselineSources -> Arc<PoolCache>` maybe something like 

```rust
impl BaselineSource {
    pub fn to_pool_cache_tuple(&self) -> (Self, Arc<PoolCache>)
}
```

This way we could baseline_sources.iter().map(to_pool_cache_tuple)` and collect into a hashmap...

Just a suggestion to keep main from being so complicated.

bh2smith (approved):
Very nice. I have made a few cosmetic suggestions.

orderbook/src/main.rs
vkgnosis: The setup is different in the orderbook main and solver main so I think this way makes sense.

solver/src/main.rs
vkgnosis: Changed code to match

orderbook/src/main.rs
bh2smith: As you wish!

Issue #694, opened 2021-06-07
Return API error when cancelling a finalized order.


nlordell:
If an order gets cancelled after it has already been previously cancelled, or already completely filled, the orderbook API should return an HTTP error (400?).

There might be some funny edge cases that require extra attention (cancelled and then filled because tx with settlement was already getting mined).

fleupold:
Great first couple of Rust PRs 🎉 !!!

PR #715, opened 2021-06-11
694/order status

alfetopito:
# Summary

Part of #694 

Added `status` to the order model.

This has been discussed over slack and on this draft PR https://github.com/gnosis/gp-v2-services/pull/714

This PR introduces a new enum `OrderStatus`.
When orders are loaded from DB, the status is then classified into that enum.
The status is part of the returned API Order data

### Test Plan
Unit tests

### DIFF
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -502,7 +502,7 @@ mod tests {
                 executed_fee_amount: BigUint::from_bytes_be(&[1]),
                 invalidated: true,
 
-            status: OrderStatus::Open,
+                status: OrderStatus::Open,
             },
             order_creation: OrderCreation {
                 sell_token: H160::from_low_u64_be(10),
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -33,7 +33,6 @@ pub struct Order {
     pub order_meta_data: OrderMetaData,
     #[serde(flatten)]
     pub order_creation: OrderCreation,
-    pub status: OrderStatus,
 }
 
 impl Default for Order {
@@ -69,7 +68,6 @@ impl Order {
                 ..Default::default()
             },
             order_creation,
-            status: OrderStatus::Open,
         })
     }
     pub fn contains_token_from(&self, token_list: &HashSet<H160>) -> bool {
@@ -338,6 +336,7 @@ pub struct OrderMetaData {
     #[serde(with = "serde_with::rust::display_fromstr")]
     pub executed_fee_amount: BigUint,
     pub invalidated: bool,
+    pub status: OrderStatus,
 }
 
 impl Default for OrderMetaData {
@@ -352,6 +351,7 @@ impl Default for OrderMetaData {
             executed_sell_amount_before_fees: Default::default(),
             executed_fee_amount: Default::default(),
             invalidated: Default::default(),
+            status: OrderStatus::Open,
         }
     }
 }
@@ -501,6 +501,8 @@ mod tests {
                 executed_sell_amount_before_fees: BigUint::from_bytes_be(&[4]),
                 executed_fee_amount: BigUint::from_bytes_be(&[1]),
                 invalidated: true,
+
+            status: OrderStatus::Open,
             },
             order_creation: OrderCreation {
                 sell_token: H160::from_low_u64_be(10),
@@ -526,7 +528,6 @@ mod tests {
                 },
                 signing_scheme: SigningScheme::Eip712,
             },
-            status: OrderStatus::Open,
         };
         let deserialized: Order = serde_json::from_value(value.clone()).unwrap();
         assert_eq!(deserialized, expected);
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -33,12 +33,22 @@ pub struct Order {
     pub order_meta_data: OrderMetaData,
     #[serde(flatten)]
     pub order_creation: OrderCreation,
+    pub status: OrderStatus,
 }
 
 impl Default for Order {
     fn default() -> Self {
         Self::from_order_creation(OrderCreation::default(), &DomainSeparator::default()).unwrap()
     }
+}
+
+#[derive(Eq, PartialEq, Clone, Debug, Deserialize, Serialize, Hash)]
+#[serde(rename_all = "camelCase")]
+pub enum OrderStatus {
+    Open,
+    Fulfilled,
+    Cancelled,
+    Expired,
 }
 
 impl Order {
@@ -59,6 +69,7 @@ impl Order {
                 ..Default::default()
             },
             order_creation,
+            status: OrderStatus::Open,
         })
     }
     pub fn contains_token_from(&self, token_list: &HashSet<H160>) -> bool {
@@ -477,6 +488,7 @@ mod tests {
             "partiallyFillable": false,
             "signature": "0x0200000000000000000000000000000000000000000000000000000000000003040000000000000000000000000000000000000000000000000000000000000501",
             "signingScheme": "eip712",
+            "status": "open",
         });
         let expected = Order {
             order_meta_data: OrderMetaData {
@@ -514,6 +526,7 @@ mod tests {
                 },
                 signing_scheme: SigningScheme::Eip712,
             },
+            status: OrderStatus::Open,
         };
         let deserialized: Order = serde_json::from_value(value.clone()).unwrap();
         assert_eq!(deserialized, expected);
--- a/orderbook/src/database/orders.rs
+++ b/orderbook/src/database/orders.rs
@@ -201,6 +201,12 @@ struct OrdersQueryRow {
     signing_scheme: DbSigningScheme,
 }
 
+/// Amount which is the minimum to consider an order fully filled.
+/// original amount / FILLED_PERCENTAGE_DENOMINATOR => minimum amount
+/// Fully filled order => amount - executed_amount < minimum_amount
+/// The higher this number, the smaller the amount to consider an order filled.
+const FILLED_PERCENTAGE_DENOMINATOR: u32 = 10_000; // 10_000 == 0.01 %
+
 impl OrdersQueryRow {
     fn calculate_status(&self) -> OrderStatus {
         match self.kind {
--- a/orderbook/src/database/orders.rs
+++ b/orderbook/src/database/orders.rs
@@ -201,12 +201,6 @@ struct OrdersQueryRow {
     signing_scheme: DbSigningScheme,
 }
 
-/// Amount which is the minimum to consider an order fully filled.
-/// original amount / FILLED_PERCENTAGE_DENOMINATOR => minimum amount
-/// Fully filled order => amount - executed_amount < minimum_amount
-/// The higher this number, the smaller the amount to consider an order filled.
-const FILLED_PERCENTAGE_DENOMINATOR: u32 = 10_000; // 10_000 == 0.01 %
-
 impl OrdersQueryRow {
     fn calculate_status(&self) -> OrderStatus {
         match self.kind {
--- a/orderbook/src/database/orders.rs
+++ b/orderbook/src/database/orders.rs
@@ -567,7 +567,7 @@ mod tests {
                 order_meta_data: OrderMetaData {
                     owner: H160::from_low_u64_be(0),
                     uid: OrderUid([0u8; 56]),
-                status: OrderStatus::Expired,
+                    status: OrderStatus::Expired,
                     ..Default::default()
                 },
                 order_creation: OrderCreation {
@@ -581,7 +581,7 @@ mod tests {
                 order_meta_data: OrderMetaData {
                     owner: H160::from_low_u64_be(0),
                     uid: OrderUid([1; 56]),
-                status: OrderStatus::Expired,
+                    status: OrderStatus::Expired,
                     ..Default::default()
                 },
                 order_creation: OrderCreation {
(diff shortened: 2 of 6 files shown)


model/src/order.rs
```diff
@@ -41,6 +42,15 @@ impl Default for Order {
     }
 }
 
+#[derive(Eq, PartialEq, Clone, Debug, Deserialize, Serialize, Hash)]
+#[serde(rename_all = "camelCase")]
+pub enum OrderStatus {
+    Open,
```
bh2smith, line 48: I suppose partially filled falls back to a status of Open. What about partially filled orders that are no longer open? I think we need a `PartiallyFilled` status.

orderbook/src/database/orders.rs
```diff
@@ -202,7 +202,31 @@ struct OrdersQueryRow {
 }
 
 impl OrdersQueryRow {
+    fn calculate_status(&self) -> OrderStatus {
+        match self.kind {
+            DbOrderKind::Buy => {
+                if !self.sum_buy.is_zero() && self.sum_buy == self.buy_amount {
+                    return OrderStatus::Fulfilled;
+                }
+            }
```
bh2smith, line 211: Do we need to be concerned with any epsilons here for classifying a partially fillable order as fully filled? Something like |x-y| < 10^9. Just wondering. Same question for sell orders.

orderbook/src/database/orders.rs
```diff
@@ -261,14 +286,85 @@ impl OrdersQueryRow {
...
+    use chrono::{Duration, NaiveDateTime};
     use futures::StreamExt;
     use num_bigint::BigUint;
     use primitive_types::U256;
     use shared::event_handling::EventIndex;
     use sqlx::Executor;
     use std::collections::HashSet;
 
+    #[test]
+    fn order_status() {
```
bh2smith, line 298: This test could probably include an open, closed and expired partially fillable order. Starting to make me think that we might want the status to be a tuple

(ValidityStatus, FillStatus)

Where 

Validity is any of (open, expired) and FillStatus is any of (not filled, partially filled or fully filled)

From this you can also derive stuff like closed.

bh2smith (changes requested):
Seems alright, but I think there are a few cases missing in the test. All of my concerns lie with the logic for partially fillable orders. At this moment the fallback is not terrible... partially filled order that is still open and not canceled is "Open" and when it is expired its expired. I just think that some relevant information for the front end is lost. 

I know that the FE doesn't currently support partial fulfillment, but if we don't incorporate this now, it will be forgotten and eventually be reported as a bug.

model/src/order.rs
alfetopito: > Seems alright, but I think there are a few cases missing in the test. All of my concerns lie with the logic for partially fillable orders. At this moment the fallback is not terrible... partially filled order that is still open and not canceled is "Open" and when it is expired its expired. I just think that some relevant information for the front end is lost.
> 
> I know that the FE doesn't currently support partial fulfillment, but if we don't incorporate this now, it will be forgotten and eventually be reported as a bug.

Nick raised the same point and I have replied in the comment https://github.com/gnosis/gp-v2-services/pull/714#discussion_r650116518

We what are your thoughts based on that?

alfetopito:
> Seems alright, but I think there are a few cases missing in the test. All of my concerns lie with the logic for partially fillable orders. At this moment the fallback is not terrible... partially filled order that is still open and not canceled is "Open" and when it is expired its expired. I just think that some relevant information for the front end is lost.
> 
> I know that the FE doesn't currently support partial fulfillment, but if we don't incorporate this now, it will be forgotten and eventually be reported as a bug.

Nick raised the same point and I replied on this comment https://github.com/gnosis/gp-v2-services/pull/714#discussion_r650116518

What are your thoughts on that?

orderbook/src/database/orders.rs
alfetopito: I did just that in the explorer https://github.com/gnosis/gp-ui/blob/develop/src/utils/operator.ts#L21

But here, I didn't because I'm following what's used in the DB filter https://github.com/gnosis/gp-v2-services/blob/main/orderbook/src/database/orders.rs#L173

Have nothing against adding the epsilon here as well.

bh2smith: Let's see what others think and whether or not we are even going to include partially filled. I do think though, that if you separated the logic of filledness and openness into a tuple, that the id statements (I.e. the code) will wind up a lot cleaner and easier to maintain for future devs. I myself have gotten into some pretty messy code when it comes to having a lot of independent statuses all in one go.

model/src/order.rs
```diff
@@ -33,6 +33,7 @@ pub struct Order {
     pub order_meta_data: OrderMetaData,
     #[serde(flatten)]
     pub order_creation: OrderCreation,
+    pub status: OrderStatus,
```
fleupold, line 36: should this be part of the `OrderMetaData` struct?

orderbook/src/database/orders.rs
fleupold: I agree that fill status and validity are kind of orthogonal (not even sure we need to indicate fill status as an enum, since we already expose the filled amount which should make it trivial to identify if an order is not, partially or fully filled).

I think it might still make sense to have the ValidityStatus also include something like the Fulfilled suggested here, because that is another way of how an order can become invalid (smart contract would complain with AlreadyFilled).

fleupold (approved):
Looks good to me. I can also get behind the suggested status. Just wondering if it should be part of the order metadata instead of the plain struct (for consistency with other meta information).

vkgnosis (approved):

model/src/order.rs
nlordell: There is a notable distinction that this is a "computed" field as opposed to one stored directly in the db. One thing I don't like about putting it in the metadata is that it is extremely easy to create invalid `OrderMetadata` instances with the current API:
```rust
OrderMetaData {
    invalidated: true,
    status: OrderStatus::Open,
}
```

This problem already exists with the current model (you can have `OrderMetaData` that is invalid for its `OrderCreation`).

Just wanted to bring this up, I don't think this is something that we should necessarily solve in this PR.

nlordell (approved):
Nice! Welcome to the dark side!

model/src/order.rs
alfetopito: So, I have the same concern as Nick.
I rather not have it in the ordermetada nor ordercreation structs as this is a calculated field.
I would rather have a third struct like `DerivedFields` or whatever to distinguish between what goes in the db and what doesn't.

What are your thoughts on this?

model/src/order.rs
alfetopito: > I suppose partially filled falls back to a status of Open. 

Yes. It's `Open` as long as it's not fully filled.

> What about partially filled orders that are no longer open? I think we need a PartiallyFilled status.

You can take how this is handled on the explorer as an example:
https://github.com/gnosis/gp-ui/blob/ad8223cc234b5f2592176cd0d2b39afdb86e6d4c/src/utils/operator.ts#L249-L251

The order status is independent of the fill status, and there's some duplication on the flags `partialFilled` and `fullyFilled`.

Also, the interface might want to focus on the `PartiallyFilled` status, or it might want to know what's the order status.
Because we can have and order as `Open`, `Cancelled` and `Expired` being `PartiallyFilled`.

orderbook/src/database/orders.rs
alfetopito: From my view, the order has 4 states, where 1 is transient (`Open`) and 3 are final (`Fulfilled`, `Cancelled`, `Expired`).
Or you can call the last 3 `Closed` states, for example:

```rust

enum ClosedStates {
    Fulfilled,
    Cancelled,
    Expired,
}

enum OrderState {
     Open,
     Closed(ClosedStates),
}
```

To keep it simple though, I still prefer the flat enum. Maybe would make more sense if we had more `Open` states, like `Pending` or something, but not as is.

Orthogonal to that, there's the fill state, which as Ben mentioned are `Not filled`, `Partially filled` and `Fully filled`.
`Not filled` is implied, as this is the initial state for all orders.
`Fully filled` only applies to the `Closed(Fulfilled)` state.
`Partially filled` can apply to `Open`, `Cancelled` and `Expired`.

Summarizing, I think we can achieve what Ben suggests by keeping the states as I propose, and adding a boolean flag to the Order model called: `PartiallyFilled`

What do you think @bh2smith ?

orderbook/src/database/orders.rs
alfetopito: Addressed on https://github.com/gnosis/gp-v2-services/pull/715/commits/9a1c82999a287eba9f3aa9b09503babbdc85322c

orderbook/src/database/orders.rs
alfetopito: Side note, for the test specifically, I've added a few more covering orders with partial fills https://github.com/gnosis/gp-v2-services/pull/715/commits/ff7c6336586882d495ad1307749098217bffbb52

orderbook/src/database/orders.rs
```diff
@@ -201,8 +201,51 @@ struct OrdersQueryRow {
     signing_scheme: DbSigningScheme,
 }
 
+/// Amount which is the minimum to consider an order fully filled.
+/// original amount / FILLED_PERCENTAGE_DENOMINATOR => minimum amount
+/// Fully filled order => amount - executed_amount < minimum_amount
+/// The higher this number, the smaller the amount to consider an order filled.
+const FILLED_PERCENTAGE_DENOMINATOR: u32 = 10_000; // 10_000 == 0.01 %
```
alfetopito, line 208: Don't fill like this is the right place for this. Anywhere better to put it?

orderbook/src/database/orders.rs
```diff
@@ -201,8 +201,51 @@ struct OrdersQueryRow {
...
+/// the order as fully filled
+fn is_fully_filled(amount: &BigDecimal, executed_amount: &BigDecimal) -> bool {
+    if executed_amount.is_zero() {
+        return false;
+    }
+    let negligible_amount = amount / FILLED_PERCENTAGE_DENOMINATOR;
+    let difference = amount - executed_amount;
+
+    difference < negligible_amount
+}
```
alfetopito, line 221: Also not very happy with this being inside the database folder.
Feels more like an Order model than SQL related. Any suggestions?

model/src/order.rs
vkgnosis: In the http api this does not matter because all the fields get flattened anyway. So this would be purely from the rust perspective. Originally the idea behind metadata was to contain everything that isn't order creation. Metadata already has fields that are not stored directly as values in the db like the balance.
Either way you can still create invalid combined orders so I don't feel moving the fields changes much in that regard. If we wanted to prevent that then it shouldn't be a field and instead be calculated by users of the api.
Anyway, I don't have a problem with a new DerivedFields struct either so if it's fine to do it that way if more like prefer it.

orderbook/src/database/orders.rs
vkgnosis: I am not sure about this. I don't find it more intuitive that there is an epsilon for partially fillable orders. You could also say that if it was filled once it is filled or it is only filled if the amount is the real full amount. Or we could add a new status PartiallyFilled that is set until it is fully Filled. I would stay with one of those solutions until we get actual experience with partially fillable orders and how they would be displayed in the frontend etc.

orderbook/src/database/orders.rs
vkgnosis: I see what you mean but I also think it's practical to keep this here for now. I feel the function makes the most sense somewhere in the orderbook crate though.

bh2smith (approved):

model/src/order.rs
fleupold: Invalidate is only needed for backwards compatibility, isn't it? Technically it's superseded by OrderStatus. In a way, because order status can be derived from the fields on the order already we create this data redundancy which also allows for inconsistent state (I assume the reason we do it here is that we don't have to do it on the client multiple times)?

Generally I'm not sure we should derive many things in the backend (and maybe instead offer a core library on the ts side that can derive things from the original json) so I would also vote for keeping this part of the metadata.

orderbook/src/database/orders.rs
fleupold: I'm not a fan of having the backend be opinionated of when an order should be considered fully filled (even when it's not). If the order was in a "final" state I would expect it to be impossible to match it further (we might even filter solvable orders based on that). However, for very large orders even a small percentage may still be fillable, e.g. when price moves significantly, so I find it problematic to hard code a FILLED_PERCENTAGE_DENOMINATOR and would rather only treat orders as fully filled when they actually are.

model/src/order.rs
alfetopito: Makes sense. I agree and think there's room for improvement separating the fields by "concern", but then it wouldn't fit in this PR.
For now, I'll move it to OrderMetaData.

orderbook/src/database/orders.rs
alfetopito: Based on @vkgnosis  and @fleupold (https://github.com/gnosis/gp-v2-services/pull/715#discussion_r651671530) comments, I'll revert the change and only set it to filled when 100% filled.

Regarding the partially filled status, see my concerns with a dedicated status and a proposal on https://github.com/gnosis/gp-v2-services/pull/715#discussion_r651163943

orderbook/src/database/orders.rs
alfetopito: Fair enough, will revert the change.

orderbook/src/database/orders.rs
alfetopito: Removed this logic altogether

orderbook/src/database/orders.rs
alfetopito: As discussed over slack, will keep it without a dedicated flag for the partial fill for now.

PR #716, opened 2021-06-11
694/cancel order on invalid status

### LINKED ISSUE #694: Return API error when cancelling a finalized order.

alfetopito:
# Summary

Fixes #694 

Depends on #715 

Adds 3 new error types for order cancellation based on order status.

### Test Plan
Unit tests

### DIFF
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -588,7 +588,7 @@ components:
       properties:
         errorType:
           type: string
-          enum: [InvalidSignature, WrongOwner, OrderNotFound]
+          enum: [InvalidSignature, WrongOwner, OrderNotFound, OrderAlreadyCancelled, OrderFullyExecuted, OrderExpired]
         description:
           type: string
       required:
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -588,7 +588,7 @@ components:
       properties:
         errorType:
           type: string
-          enum: [InvalidSignature, WrongOwner, OrderNotFound, OrderAlreadyCancelled, OrderFullyExecuted, OrderExpired]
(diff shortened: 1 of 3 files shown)


orderbook/openapi.yml
```diff
@@ -588,7 +588,7 @@ components:
       properties:
         errorType:
           type: string
-          enum: [InvalidSignature, WrongOwner, OrderNotFound]
+          enum: [InvalidSignature, WrongOwner, OrderNotFound, OrderAlreadyCancelled, OrderFullyExecuted, OrderExpired]
```
bh2smith, line 591: ubernit: I think `AlreadyCancelled` can live without the term Order in front, but the other two (for some reason) make sense to have... I'm weird.

bh2smith (approved):
This looks good, but I think the inclusion of partial orders might wind up changing things a bit here. Keep in mind that this won't entirely solve the issues we are hoping for with inflight order cancellation, but it is already much better than we had before. 

Just to elaborate, some of these statuses will depend on our event listener which waits for trade events and updates some of the fields used in the status calculation. 

In any it is an improvement! Nice work.

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FW0Vz7zDK7vgMBc2fpw%2Fgiphy.gif&token=very_nice" data-gitmeme-token="very_nice"></a>

vkgnosis (approved):

orderbook/openapi.yml
alfetopito: I don't mind either way. Will update.

PR #710, opened 2021-06-10
Naive Benchmark for Balancer Sync

bh2smith:
This simple data analysis was collected from the code currently in #696 (event receiving from WeightedPoolFactory with pool info fetching processed sequentially). Note that the performance is not entirely self contained as this was merely a regular run of the orderbook API which is simultaneously responsible for other tasks. To run this exact same simulation again, checkout this branch and run

```sh
cargo run --bin orderbook -- --node-url YOUR-NODE-URL --log-filter shared=info
```

After performing some naive benchmarks, I see no reason to believe that our data fetching is or will ever be a bottle neck even when it is not batched or parallelized. Of course we will eventually have to account for the fact that a cold start will become impractical. In fact, from what we see here (of course this is only data from a product which is not yet live), there would be no improvement gained from batched our requests. Only when several pools are deployed within the same 100 (25 minutes w. 15 second block times) blocks can we expect to see any marginal performance gain from batching requests.

I have found from a single simulation contained in the data logs below that fetching token data for 2 pools and syncing events for about 500 blocks takes < 0.4 seconds (including all other processes that the orderbook API is running in parallel). Only when several pools are deployed within the same 100 blocks (25 minutes w. 15 second block times) can we expect to see any marginal performance gain from batching requests. 

Say, for example that 100 pools are deployed in 25 minutes... then it is expected to take 20 seconds to process and store these events (but of course this would happen without notice at different times during that 25 minute interval).

## Summary of finding from logs:

Spans over:
    1.5 months (April 21 to June 9) - 316_781 blocks

Total time to sync all pools: 8 minutes.

## Points of interest:
- There is only 2 occurrences of > 1 pool created over a period of 1000 block (and these are the real time gaps).
- The real time gap here is not with the pool data fetching, but merely with event sync
      - Between 11 and 12 (128_443 blocks) we have a time gap of 2 minutes and no pools created
      - Between 1 and 3: total time gap of 0.4 seconds to check 500 blocks and fetch the pool info for 2 new pools.

## Comparison with Uniswap V2:

- There are 41_928 pools total. (dune query: `select count(*) from uniswap_v2."Factory_evt_PairCreated"`)
- Total processing time in terms of data fetching (plus 500 block sync) 41_928 * 0.2 (seconds) = 2.3 hours
- For a cold start, this is not ideal!
- However, we can easily handle 5000 pools in 16 minutes.
By this point, I would imagine that we can read/write to an from file to to avoid such issues.

[...]

PR #717, opened 2021-06-12
Balancer Event Handler for WeightedPool2TokenFactory

bh2smith:
### Disclaimer
Not yet satisfied with the amount of code duplication and looking for some assistance. I believe the reason for the duplication stems from [this issue](https://github.com/gnosis/ethcontract-rs/issues/513) in ethcontract-rs, but I may be wrong. Maybe we could have some sort of `enum ContractRouter` that can distinguish between the two different factories and somehow reduce this redundancy...

### Change Summary

- Vendoring `WeightedPool2TokensFactory` contract artifacts
- Updating `BalancerEventUpdater` to stream events from both factories and store all pools accordingly. 

### Test Plan
You can try this out yourself in another branch (cf #710) by running
```sh
cargo run --bin orderbook -- --node-url YOUR-NODE-URL --log-filter shared=info
```

Results are as follows:

```
2021-06-12T07:33:36.139Z  Last updated block by type (0, 12363041) - loaded 1 balancer pools
2021-06-12T07:33:37.111Z  Last updated block by type (0, 12363659) - loaded 2 balancer pools
2021-06-12T07:33:41.937Z  Last updated block by type (0, 12367800) - loaded 3 balancer pools
2021-06-12T07:33:46.087Z  Last updated block by type (0, 12368069) - loaded 5 balancer pools
2021-06-12T07:33:48.539Z  Last updated block by type (0, 12369384) - loaded 10 balancer pools
2021-06-12T07:33:50.667Z  Last updated block by type (0, 12369886) - loaded 23 balancer pools
2021-06-12T07:33:51.912Z  Last updated block by type (0, 12370108) - loaded 32 balancer pools
2021-06-12T07:34:01.746Z  Last updated block by type (0, 12377374) - loaded 33 balancer pools
2021-06-12T07:35:16.734Z  Last updated block by type (0, 12437421) - loaded 34 balancer pools
2021-06-12T07:35:27.039Z  Last updated block by type (0, 12445005) - loaded 35 balancer pools
2021-06-12T07:35:54.637Z  Last updated block by type (0, 12454935) - loaded 36 balancer pools
2021-06-12T07:37:18.496Z  Last updated block by type (0, 12504909) - loaded 37 balancer pools
2021-06-12T07:37:24.432Z  Last updated block by type (0, 12510588) - loaded 39 balancer pools
2021-06-12T07:38:19.137Z  Last updated block by type (0, 12549331) - loaded 41 balancer pools
2021-06-12T07:40:12.793Z  Last updated block by type (0, 12616747) - loaded 44 balancer pools
2021-06-12T07:40:16.305Z  Last updated block by type (0, 12617821) - loaded 47 balancer pools
2021-06-12T07:40:32.042Z  Last updated block by type (12285749, 12617821) - loaded 48 balancer pools
2021-06-12T07:40:36.559Z  Last updated block by type (12286276, 12617821) - loaded 50 balancer pools
2021-06-12T07:40:50.000Z  Last updated block by type (12297008, 12617821) - loaded 51 balancer pools
2021-06-12T07:40:51.431Z  Last updated block by type (12299382, 12617821) - loaded 52 balancer pools
2021-06-12T07:40:57.932Z  Last updated block by type (12303593, 12617821) - loaded 53 balancer pools
2021-06-12T07:40:58.779Z  Last updated block by type (12304403, 12617821) - loaded 54 balancer pools
2021-06-12T07:41:05.838Z  Last updated block by type (12318074, 12617821) - loaded 55 balancer pools
2021-06-12T07:41:16.111Z  Last updated block by type (12331542, 12617821) - loaded 56 balancer pools
2021-06-12T07:41:16.547Z  Last updated block by type (12331765, 12617821) - loaded 57 balancer pools
2021-06-12T07:41:56.808Z  Last updated block by type (12376911, 12617821) - loaded 58 balancer pools
2021-06-12T07:44:46.556Z  Last updated block by type (12505354, 12617821) - loaded 59 balancer pools
2021-06-12T07:44:50.384Z  Last updated block by type (12509636, 12617821) - loaded 60 balancer pools
2021-06-12T07:46:04.959Z  Last updated block by type (12563369, 12617821) - loaded 61 balancer pools
2021-06-12T07:46:13.894Z  Last updated block by type (12567967, 12617821) - loaded 62 balancer pools
2021-06-12T07:46:37.792Z  Last updated block by type (12595481, 12617821) - loaded 63 balancer pools
2021-06-12T07:46:38.046Z  Last updated block by type (12595681, 12617821) - loaded 64 balancer pools
2021-06-12T07:46:41.543Z  Last updated block by type (12602530, 12617821) - loaded 67 balancer pools
```


shared/src/balancer/event_handler.rs
```diff
@@ -285,15 +426,32 @@ impl_event_retrieving! {
...
+impl_event_retrieving! {
+    pub BalancerV2WeightedPool2TokensFactoryContract for balancer_v2_weighted_pool_2_tokens_factory
+}
+
 #[async_trait::async_trait]
 impl Maintaining for BalancerEventUpdater {
     async fn run_maintenance(&self) -> Result<()> {
-        self.0.run_maintenance().await
+        self.two_token_pool_updater.run_maintenance().await?;
+        self.weighted_pool_updater.run_maintenance().await
```
bh2smith, line 437: This is responsible for deleting all the original pools and starting over again. Need to separate last updated in a way that it knows not to delete the other pools.

shared/src/balancer/event_handler.rs
```diff
@@ -32,27 +38,36 @@ pub struct PoolCreated {
     pub pool_address: H160,
 }
 
-#[derive(Clone, Debug, Default, Eq, PartialEq, Hash)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+pub enum PoolType {
+    WeightedGeneral,
```
fleupold, line 43: What is the difference in how the services will handle a general vs a weighted two token pool? Asking, because it's not yet clear to me why we need to differentiate between the two (I assume they have different methods for swapping of fetching balances?)

fleupold (commented):
Would it make sense to have two even handler classes (one for WeightedTwoToken pools and another one for WeightedGeneral pools)? This way we might make the class generic on the Event type (to reduce code duplication) and only implement the conversion from `WeightedTwoTokenEvent` to `RegisteredWeightedPool` and `WeightedGeneralEvent` to `RegisteredWeightedPool`. This way the event handler itself wouldn't have to take the event type as an argument any more (and do custom logic in the deletion/insertion of events)

For the consumer that wants to get all RegisteredWeightedPools, we could have an aggregator fetching all `RegisteredWeightedPool` from the two event processing classes.

Just a thought, will leave it up to you and @nlordell to decide on a design. It might also be worth while thinking ahead of how different pools will eventually map to Liquidity types that we pass in to the solver. Maybe that influences the choice of data structures?

shared/src/balancer/event_handler.rs
bh2smith: In this particular case, both types are the same as (far as I know). However, we can be certain that the other pool types will need to be differentiated for, at least, exactly the reason you mentioned (i.e. possible different transaction encoding and pool arithmetic). 

In this PR, I partially kept things separate like this mostly in preparation for what it to come. However, it is very likely that both `WeightedPool` and `WeightedPool2Tokens` don't actually need to be separated.

fleupold: How is the pool arithmetic different between general weighted pools and weighted two token pools? For swapping, I was under the impression that all we need is the poolId and the desired amounts.

bh2smith: I believe I said they are not different at all. 

> In this particular case, both types are the same

> WeightedPool and WeightedPool2Tokens don't actually need to be separated.

The arithmetic is likely different for pools of **other** types.

shared/src/balancer/event_handler.rs
```diff
@@ -151,11 +166,16 @@ impl PoolRegistry {
             .collect()
     }
 
-    async fn insert_events(&mut self, events: Vec<(EventIndex, PoolCreated)>) -> Result<()> {
+    async fn insert_events(
+        &mut self,
+        events: Vec<(EventIndex, PoolCreated)>,
```
nlordell, line 171: One way to reduce duplicated code would be to define a common type that both contract events turn into. Here a simple `(u64, H160)` would suffice.

nlordell (commented):
One thing I didn't think of was the having multiple contracts act on the same `EventStoring` causes issues when replacing events. I like @fleupold's suggestions.

shared/src/balancer/event_handler.rs
bh2smith: Yes, certainly we will want to do this. However, we have decided that at the particular moment, we will remove the pool type since these two pools don't require it and only to reintroduce this when it is absolutely necessary. With this in mind, I will sketch up a description of how each are different/handled internally when the time comes.

shared/src/balancer/event_handler.rs
```diff
@@ -32,27 +38,36 @@ pub struct PoolCreated {
...
 }
 
-#[derive(Clone, Debug, Default, Eq, PartialEq, Hash)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+pub enum PoolType {
+    WeightedGeneral,
+    WeightedTwoToken,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
```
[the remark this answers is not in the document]
bh2smith: These hashes are all used because we have `RegisteredWeightedPool` as a value in a HashMap.

shared/src/balancer/event_handler.rs
nlordell: From a cursory glance at the Vault code, they behave the same from the caller's point of view. The difference is only done to enable internal optimiations (2Token vs MinSwapInfo pools).

nlordell: I think ultimately we would want something like:
```
enum BalancerPool {
  Weighted(WeightedPool),
  Stable(StablePool),
  // ...
}
```

shared/src/balancer/event_handler.rs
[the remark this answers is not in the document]
nlordell: `Hash` is only needed if it used as a key, not a value.

shared/src/balancer/event_handler.rs
nlordell: I like the `PoolCreated` struct - I must have missed it.

shared/src/balancer/event_handler.rs
```diff
@@ -32,27 +38,36 @@ pub struct PoolCreated {
     pub pool_address: H160,
 }
 
-#[derive(Clone, Debug, Default, Eq, PartialEq, Hash)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
```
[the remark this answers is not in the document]
bh2smith:
```
error[E0277]: the trait bound `RegisteredWeightedPool: std::hash::Hash` is not satisfied
   --> shared/src/balancer/event_handler.rs:133:14
    |
133 |             .collect::<HashSet<RegisteredWeightedPool>>()
    |              ^^^^^^^ the trait `std::hash::Hash` is not implemented for `RegisteredWeightedPool`
    |
    = note: required because of the requirements on the impl of `FromIterator<RegisteredWeightedPool>` for `std::collections::HashSet<RegisteredWeightedPool>`
```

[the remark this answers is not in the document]
bh2smith: I was able to remove it from one place, but not `RegisteredWeightedPool`

shared/src/balancer/event_handler.rs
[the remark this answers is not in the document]
bh2smith:
```
error[E0277]: the trait bound `RegisteredWeightedPool: std::hash::Hash` is not satisfied
   --> shared/src/balancer/event_handler.rs:133:14
    |
133 |             .collect::<HashSet<RegisteredWeightedPool>>()
    |              ^^^^^^^ the trait `std::hash::Hash` is not implemented for `RegisteredWeightedPool`
    |
    = note: required because of the requirements on the impl of `FromIterator<RegisteredWeightedPool>` for `std::collections::HashSet<RegisteredWeightedPool>`

```

bh2smith:
I have reworked the implementation here having one `DynPoolRegistry` for each event handler and then a method to fetch the merged version of the two. Unfortunately at this moment we are stuck with passing the entire `BalancerEventUpdater` struct onto PoolFetching. I really hope there is a better way to extract the "Database" from the event updater.

shared/src/balancer/event_handler.rs
```diff
@@ -318,29 +378,64 @@ impl_event_retrieving! {
...
+        registry_1.merge(registry_2)
+    }
+}
+
 #[async_trait::async_trait]
 impl Maintaining for BalancerEventUpdater {
     async fn run_maintenance(&self) -> Result<()> {
-        self.0.run_maintenance().await
+        self.two_token_pool_updater.run_maintenance().await?;
+        self.weighted_pool_updater.run_maintenance().await
```
bh2smith, line 397: These could now be run together with `futures::future::join_all` although they are quite big tasks... going to try out both in #710

shared/src/balancer/event_handler.rs
```diff
@@ -183,6 +177,43 @@ impl PoolRegistry {
...
+    pub fn new(data_fetcher: Box<dyn PoolDataFetching>) -> Self {
+        DynPoolRegistry {
+            pools_by_token: Default::default(),
+            pools: Default::default(),
+            data_fetcher,
+        }
+    }
+
+    pub fn merge(&self, other: &DynPoolRegistry) -> PoolRegistry {
+        tracing::debug!("Merging Balancer Pool Registries");
```
bh2smith, line 204: I included this log too see how often we are merging (since it requires cloning each registry to build) and I think the pool fetcher will request this every time... Could be a problem, but I haven't managed to come up with a better way of exposing the merged Registry to the pool fetcher on demand.

bh2smith: One idea is too keep track of `last_updated`, but this value can only be derived from that of each registry as either the min or the max. Of course this value wont ever be up to date with the current block number so even if we give the pool fetcher info about last_updated, it would always try to build a new one.

shared/src/balancer/event_handler.rs
bh2smith: There are only two (ish) pool types `WeightedPool` (including `TwoTokens`) and `StablePool`.

The transaction encoding is likely the same for both since we perform swaps via the Vault contract
While the arithmetic is different for each since Stable Pools use `StableMath` library while Weighted Pools uses `WeightedMath`.

This means we will need to introduce a PoolType enum (as @nlordell suggested above) down the road (i.e. when we include StablePools) that the price estimating will need to distinguish between.

shared/src/balancer/event_handler.rs
```diff
@@ -237,75 +248,120 @@ impl PoolRegistry {
...
-                    Some(meta) => meta,
-                    None => return Err(anyhow!("event without metadata")),
-                };
-                match data {
-                    ContractEvent::PoolCreated(event) => convert_pool_created(&event, &meta),
-                }
-            })
-            .collect::<Result<Vec<_>>>()
+        token_pairs: HashSet<TokenPair>,
+    ) -> HashSet<RegisteredWeightedPool> {
```
nlordell, line 264: Why a `HashSet` and not a `Vec`? Also, I don't think `Hash` makes sense for this type. For example (in pseudo-code):
```rust
let pool0 = RegisteredWeightedPool {
  pool_id: some_id,
  address: some_address,
  ...
}
let pool1 = RegisteredWeightedPool {
  pool_id: some_id,
  address: some_other_address,
  ...
}
```

What do you expect to happen when adding these pools into the same `HashSet`? Should there be two (which is what will happen now with the `derive(Hash)` implementation)? Or one (because pool IDs are unique)? I would like to avoid this kind of ambiguity.

Additionally, using a `HashSet` here makes the code imply that you expect that there are pools common to both `weighted_pool_updater` and `two_token_pool_updater` which there shouldn't be and makes this quite confusing to me.

nlordell (changes requested):
Looks great!

Just requesting changes as I believe that the use of `HashSet` for a collection of `RegisteredWeightedPool` is slightly incorrect. I think the same applies for:

https://github.com/gnosis/gp-v2-services/blob/e77938e3f83203fe4eb1c6bf54ed3ff35447a2a9/shared/src/balancer/event_handler.rs#L148-L171

Which I missed when reviewing #698 (here, the pools are already unique by pool ID so there is no need to collect them in a `HashSet` and a `Vec` makes more sense).

shared/src/balancer/event_handler.rs
bh2smith: Created an issue and linked it here to track the collected information #733

bh2smith:
Also added an issue to include some documentation of this module (but I would prefer to wait until) some of the open PRs has settled (particularly this and #722).

https://github.com/gnosis/gp-v2-services/issues/734

nlordell (approved):

shared/src/balancer/event_handler.rs
nlordell: This isn't relevant anymore right?

shared/src/balancer/event_handler.rs
nlordell: > Results appear to show that it doesn't make much of a difference either way.

If we are behind a load balancer, these requests would be handled by different nodes and would speed things up. I think its worth joining these futures. You can use the `try_join` macro to do this quite easily:
```rust
futures::try_join!(
  self.two_token_pool_updater.run_maintenance(),
  self.weighted_pool_updater.run_maintenance(),
);
```

shared/src/balancer/event_handler.rs
```diff
@@ -235,75 +246,125 @@ impl PoolRegistry {
...
+        let deployment_block_weighted_pool =
+            get_deployment_block(weighted_pool_factory.deployment_information(), &web3).await;
+        let deployment_block_two_token_pool =
+            get_deployment_block(two_token_pool_factory.deployment_information(), &web3).await;
+        let weighted_pool_updater = Mutex::new(EventHandler::new(
+            web3.clone(),
+            BalancerV2WeightedPoolFactoryContract(weighted_pool_factory),
+            PoolStorage::new(Box::new(PoolDataFetcher {
+                web3: web3.clone(),
+                token_info_fetcher: token_info_fetcher.clone(),
```
bh2smith, line 293: Is cloning the correct way to pass the token info fetcher into both event handlers? @fleupold.

nlordell: It is behind an `Arc`, so cloning is a very cheap operation (it just increases the reference count - i.e. the number of "owners" to the inner data).

It is the correct way of sharing an `Arc`.

bh2smith: Cool, now I am just waiting on @fleupold's final thoughts and I would like to ship-it.

PR #722, opened 2021-06-14
[Cleanup] Balancer Split Files and Document

bh2smith:
event handler file had become quite large with a logical separation into 4

1. event handler: itself (responsible for definition and implementation of `BalancerPoolRegistry`)
2. pool_storage: all structs/models for the module including PoolStorage (this is analogous to the database file for GPv2Settlement event listening). 
3. pool_info_fetching: this include a trait and one implementation of the trait.
4. pool_fetching: unchanged.

No logic changes, but please check that all things which were made public are indeed correct.

Note that I have also documented the module and each file here as well.

### Test Plan
existing CI.

### DIFF
--- a/shared/src/balancer/event_handler.rs
+++ b/shared/src/balancer/event_handler.rs
@@ -126,8 +126,7 @@ impl EventStoring<WeightedPoolFactoryEvent> for PoolStorage {
         events: Vec<EthContractEvent<WeightedPoolFactoryEvent>>,
         range: RangeInclusive<BlockNumber>,
     ) -> Result<()> {
-        PoolStorage::replace_events_inner(
-            self,
+        self.replace_events_inner(
             range.start().to_u64(),
             convert_weighted_pool_created(events)?,
         )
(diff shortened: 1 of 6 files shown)


vkgnosis (approved):

shared/src/balancer/event_handler.rs
```diff
@@ -355,17 +130,20 @@ impl EventStoring<WeightedPool2TokensFactoryEvent> for PoolStorage {
         events: Vec<EthContractEvent<WeightedPool2TokensFactoryEvent>>,
         range: RangeInclusive<BlockNumber>,
     ) -> Result<()> {
-        let balancer_events = convert_two_token_pool_created(events)?;
-        PoolStorage::replace_events_inner(self, range.start().to_u64(), balancer_events).await?;
-        Ok(())
+        PoolStorage::replace_events_inner(
+            self,
```
nlordell, line 158: I missed this before but this can likely be written as `self.replace_events_inner(...)`.

nlordell (approved):
Approving assuming there were no code/logic changes besides what was required for moving files around.

shared/src/balancer/event_handler.rs
bh2smith: Yea probably, will check it out.

fedgiac (approved):
The file split makes sense and the docs are exhaustive.

Note that `logexpmath` is going to be moved in a subdirectory, so it's ok to skip its documentation.

PR #723, opened 2021-06-14
Read/Write Balancer Pool Registry From/To Binary File

bh2smith:
I have left this PR independent from the existing open code introducing the TwoToken Pools so some changes will have to be made to merge the two, but the logic would (more or less) remain similar. Just wanted to get this out the and start a discussion of whether or not this is even needed. 

Most of the code introduced in data_loader has been migrated over from gp-v1-services.

### Test Plan

TODO - write tests for load/dump files


vkgnosis:
This is for persistent storage like the orderbook in gpv1, right? In that case I think we should be using postgres instead of the more improvised solution from gpv1. It is nice to only have one source of persistence and devops wasn't happy with the v1 setup.

nlordell:
Another issue is that we need this liquidity in two places - the orderbook and the solver. I think how we want to achieve this deserves a bit more discussion.

bh2smith:
>  I think we should be using postgres

Yes this is totally fine as well. I was just putting this here since it was quick to put together and see if we might consider it. May not even be necessary if we can easily cold start from TheGraph node anyway.

> devops wasn't happy with the v1 setup.

Yes I have now been hearing this from a few different team members, but I have to admit that I had not heard any of these complaints until recently. What was the big issue there.

nlordell:
One issue I see with using postgres is that we would need this data in both the orderbook API as well as the solver. So would they share a DB or have separate ones?

One appealing thing about a file is that it can just be uploaded to an S3 bucket somewhere.

fleupold:
Given Nick's implementation of The Graph, we are not going to do file based persistence, are we?

nlordell:
I think this can be closed for now, we can revisit if we run into issues with the graph. We now have a `PoolInitializing` trait that we could implement for file-based cold starts.

PR #725, opened 2021-06-14
[Balancer] Batch and fetch scaling factors for Weighted Pool Data

bh2smith:
Part of #603 - which is near completion.

For pool arithmetic, pool scaling factors must be included in the pool data. We now optimize our EVM queries by spliiting into three batched calls

1. Pool ID must be fetched before we can query the tokens.
2. Tokens and normalized weights (these are independent queries now batched as a single call). 
3. token_decimals - makes use of pre-existing `TokenInfoFetcher` which already batches together queries for retrieving decimals for a list of token addresses.

This is the final missing piece that @fedgiac will need to incorporate his pool math work.

### Test Plan
No logical changes, existing CI.

### DIFF
--- a/shared/src/balancer/event_handler.rs
+++ b/shared/src/balancer/event_handler.rs
@@ -1,5 +1,4 @@
 use crate::pool_fetching::MAX_BATCH_SIZE;
-use crate::token_info::{TokenInfoFetcher, TokenInfoFetching};
 use crate::{
     current_block::BlockRetrieving,
     event_handling::{BlockNumber, EventHandler, EventIndex, EventStoring},
@@ -12,7 +11,7 @@ use contracts::{
     balancer_v2_weighted_pool_factory::{
         self, event_data::PoolCreated as ContractPoolCreated, Event as ContractEvent,
     },
-    BalancerV2Vault, BalancerV2WeightedPool, BalancerV2WeightedPoolFactory,
+    BalancerV2Vault, BalancerV2WeightedPool, BalancerV2WeightedPoolFactory, ERC20,
 };
 use derivative::Derivative;
 use ethcontract::batch::CallBatch;
@@ -20,6 +19,7 @@ use ethcontract::common::DeploymentInformation;
 use ethcontract::{
     dyns::DynWeb3, Bytes, Event as EthContractEvent, EventMetadata, H160, H256, U256,
 };
+use futures::future::join_all;
 use itertools::Itertools;
 use mockall::*;
 use model::TokenPair;
@@ -41,7 +41,7 @@ pub struct RegisteredWeightedPool {
     pub pool_address: H160,
     pub tokens: Vec<H160>,
     pub normalized_weights: Vec<U256>,
-    pub scaling_factors: Vec<u8>,
+    pub scaling_exponents: Vec<u8>,
     block_created: u64,
 }
 
@@ -59,7 +59,7 @@ impl RegisteredWeightedPool {
             pool_address,
             tokens: pool_data.tokens,
             normalized_weights: pool_data.weights,
-            scaling_factors: pool_data.scaling_factors,
+            scaling_exponents: pool_data.scaling_exponents,
             block_created,
         });
     }
@@ -70,7 +70,7 @@ pub struct WeightedPoolData {
     pool_id: H256,
     tokens: Vec<H160>,
     weights: Vec<U256>,
-    scaling_factors: Vec<u8>,
+    scaling_exponents: Vec<u8>,
 }
 
 #[automock]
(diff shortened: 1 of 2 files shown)


shared/src/balancer/event_handler.rs
```diff
@@ -77,24 +83,43 @@ trait PoolDataFetching: Send + Sync {
...
+            .get_token_infos(&tokens)
+            .await;
+        let scaling_factors = token_decimals
+            .iter()
+            .map(|(_, info)| {
+                // Note that balancer does not support tokens with more than 18 decimals
+                // https://github.com/balancer-labs/balancer-v2-monorepo/blob/ce70f7663e0ac94b25ed60cb86faaa8199fd9e13/pkg/pool-utils/contracts/BasePool.sol#L497-L508
+                18 - info
+                    .decimals
+                    .expect("all token decimals must be available build pool data")
```
fedgiac, line 115: Maybe it's better to return an error here, otherwise you could create a token with customizable decimals, create a Balancer pool with the decimals set to 18, then increase the decimals and make the services crash.

shared/src/balancer/pool_fetching.rs
```diff
@@ -14,6 +14,7 @@ use ethcontract::{BlockId, Bytes, H160, H256, U256};
 pub struct PoolTokenState {
     pub balance: U256,
     pub weight: U256,
+    pub scaling_factor: u8,
```
fedgiac, line 17: In the contract code the scaling factor is `10^(18-decimals)`, so maybe we should call this differently, for example `scaling_exp`?

fedgiac (approved):
Looks good from my end!

shared/src/balancer/event_handler.rs
bh2smith: I thought that the expect message was the error... Perhaps I have misunderstood. I suppose expect panics when it fails.

shared/src/balancer/pool_fetching.rs
bh2smith: Yes, good idea. I will change it. I am assuming you can handle the value then as a u8 and not as the actual contract value (like 10^x)

shared/src/balancer/event_handler.rs
fedgiac: `expect` panics with that error message, but since this is something that could be used to make the services crash we could make `get_pool_data` return an error in this case instead.

vkgnosis (approved):

shared/src/balancer/event_handler.rs
```diff
@@ -77,24 +83,51 @@ trait PoolDataFetching: Send + Sync {
...
+            .batch_call(&mut batch);
+        let normalized_weights = pool_contract
+            .methods()
+            .get_normalized_weights()
+            .batch_call(&mut batch);
+        batch.execute_all(MAX_BATCH_SIZE).await;
+
+        let tokens = token_data.await?.0;
+        // This is already a batch call for a list of token decimals.
+        let mut batch = CallBatch::new(self.transport());
```
fleupold, line 105: Why is this a separate CallBatch and not reusing the old one? Also, from reading the PR description I was assuming we would use the existing `TokenInfoFetching` component here?

bh2smith: We started with using token info fetching, but it returns a hashmap which (I assume) is not guaranteed to preserve order. Rather than having to shuffle everything around, I decided not to use it. If it does preserve order, then I would be happy to use it.

Also, I use a new batch because I wanted to make sure it was empty for the next call. If this is not necessary then I woul also be happy to change this.

fleupold: I see now why we need to use a new batch. It is because the token data is part of the first call (thus the first batch needs to be executed first).

I still don't understand though, why TokenInfoFetching is not working in this case though. You can look up the tokens in order from your original list, can't you?

Generally I think we should be mindful of adding extra roundtrips (with the token fetcher we'd likely have some cache amortisation).

bh2smith: I wasn't sure if HashMaps were guaranteed to be order preserving so I avoided it. I also avoided it because, regardless of the potential for Cached Token Info fetching, the only way we were able to use it was to instantiate a new one on each call anyway. I can imagine a situation where we instantiate a cached token info fetching external to the `BalancerEventUpdater` and pass it along all the way through and down to our PoolDataFetching so that this stuff is indeed cached and we can benefit from it. 

I would be happy to try making this happen as long as we are sure that the HashMap returned is order preserving. 

Will open a PR and see how it looks.

bh2smith: This is turning out to be a super ugly change handing down and passing around a bunch of `Option<&CachedPoolFetchers>` everywhere, expected lifetime parameters and makes the PoolRegistry even harder to test (with mocking the pool fetchers). Do you really want this?

fleupold: I would pass in an `Arc<dyn TokenInfoFetching>` like we do e.g. in the `HttpSolver`. This way there shouldn't be the need for lifetimes (also no need to leak the fact that you want to use the cached version) and it should also only need to live on `BalancerPoolFetcher` which I assume will be instantiated from main (so I don't think it would be in many places).

fleupold: > I would be happy to try making this happen as long as we are sure that the HashMap returned is order preserving.

HashMaps don't preserve order. However, since you still have the order of your tokens from the input, you can look up the decimals in order from that list.

Issue #586, opened 2021-05-11
Implement Balancer weighted pools


fedgiac:
Introduce the concept of weighted pool, which is one of Balancer's liquidity sources we are going to tap from.
This task includes implementing `BaselineSolvable` for this type of pools. The equations describing how the liquidity sources beave are [here](https://balancer.fi/whitepaper.pdf#trading-formulas), while the contracts using these formulas are [here](https://github.com/balancer-labs/balancer-core-v2/blob/master/contracts/pools/weighted/WeightedPool.sol) and [here](https://github.com/balancer-labs/balancer-core-v2/blob/master/contracts/pools/weighted/WeightedMath.sol ).
Additionally, it may be necessary to implement the custom rational number implementation from the Balancer contracts.

bh2smith:
In the event that this is not available at the time of Event handling implementation, I will likely include something like the following as a starting point

```rust
pub struct WeightedPool {
    pool_address: H160,
    tokens: Vec<H160>,
    pool_id: [u8; 32]
    // TODO - Include link to this task here
}
```

PR #726, opened 2021-06-14
Add Balancer pow function

fedgiac:
Part of #586.

The Balancer contract implements a custom `pow` function that is used to compute traded amounts. This PR adds code to replicate the smart contract code defining `pow`, which will be used in future PRs.

### Test Plan
New unit tests.


shared/src/balancer/logexpmath.rs
```diff
@@ -0,0 +1,520 @@
...
+        0 => {
+            I256::from_dec_str("38877084059945950922200000000000000000000000000000000000").unwrap()
+        }
+        1 => 6_235_149_080_811_616_882_910_000_000_i128.into(),
+        _ => panic!("Constant not provided"),
+    }
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum Errors {
```
vkgnosis, line 74: Probably would be nice to use https://docs.rs/thiserror/1.0.25/thiserror/ to derive Error for this type. But depends on how the function that returns this is used so maybe it's not needed if we never bubble them up or turn them into anyhow::Error.

shared/src/balancer/logexpmath.rs
```diff
@@ -0,0 +1,520 @@
...
+    static ref MAX_NATURAL_EXPONENT: I256 = ONE_18.checked_mul(I256::from(130_i128)).unwrap();
+    static ref MIN_NATURAL_EXPONENT: I256 = ONE_18.checked_mul(I256::from(-41_i128)).unwrap();
+    static ref LN_36_LOWER_BOUND: I256 = ONE_18.checked_sub(I256::from(10_u128.pow(17))).unwrap();
+    static ref LN_36_UPPER_BOUND: I256 = ONE_18.checked_add(I256::from(10_u128.pow(17))).unwrap();
+    static ref MILD_EXPONENT_BOUND: I256 =
+        I256::from(2_u32).pow(254_u32).checked_div(*ONE_20).unwrap();
+}
+fn constant_x_20(i: u32) -> I256 {
+    match i {
+        2 => 3_200_000_000_000_000_000_000_i128,
```
vkgnosis, line 25: I see that the balancer code has x0 to x11 and a0 to a11 as constants. Maybe it would be more ergonomic to store them in a fixed size array here? like `const X_: [I256; 11] = [...]`. I'm not sure, just an idea.

Also as the balancer constants have an explanation for how they were calculated like
>  int256 constant x0 = 128000000000000000000; // 2ˆ7

We could define them as `I256::exp10(18 + 7)`. On the other hand I can see the argument for keeping this code as close to the balancer code as possible.

fedgiac: I could definitely use only two arrays for `x` and `a`. I decided to split each in two arrays (`18` and `20`) because the first two of the array are scaled compared to the other 9 by a factor of 100. This in particular means that they are used in different places in the code, and if an index is used in the wrong place then the function panics. This was helpful when writing up everything to catch off-by-one issues. On the other end, I'm also ok with using an array and being more consistent with the original Balancer code.

About computing the constants: at first I did that for the `x` constants, but if we want to stay consistent and do the same for the `a` constants I'd also need to truncate the decimal representation of `e^n` after some point, which looks overly complicated for me to this purpose. So I also kept `x` as a list for consistency. Again no strong opinion, thanks to the tests we could define everything with an explicit calculation without any issue.

shared/src/balancer/logexpmath.rs
fedgiac: Eventually all these errors will disappear into an `Option`. This enum is going to be extended to include all Balancer revert errors that we need, so that I can work with other contract functions reverting using `?`. So unless things change I'd keep it like this without an explicit error message provided by `thiserror`.

vkgnosis (approved):

shared/src/balancer/logexpmath.rs
```diff
@@ -0,0 +1,522 @@
...
+    exp(logx_times_y).map(|v| v.into_raw())
+}
+
+fn exp(mut x: I256) -> Result<I256, Errors> {
+    if !(x >= *MIN_NATURAL_EXPONENT && x <= *MAX_NATURAL_EXPONENT) {
+        return Err(Errors::InvalidExponent);
+    }
+
+    if x < I256::zero() {
+        return Ok((*ONE_18 * *ONE_18) / exp(-x)?);
```
bh2smith, line 122: the *s in here looks awkwardly placed, spaced and separated...

I have also noticed a lot of `*` before variable and not between anything... is this something I should know about rust that I do not?

shared/src/balancer/logexpmath.rs
```diff
@@ -0,0 +1,522 @@
...
+
+    series_sum * 2.into()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use regex::Regex;
+
+    // Compare computed constants with smart contract code.
```
bh2smith, line 221: Could we also link to the corresponding smart contract tests here?

bh2smith (approved):
Wowie. Isn't EVM exponentiation such a joy!

shared/src/balancer/logexpmath.rs
fedgiac: At the time I wrote this, there were no public tests to read from. Here is a short test file from 5 days ago by Balancer, but I don't think it adds much: 
https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/solidity-utils/test/LogExpMath.test.ts

shared/src/balancer/logexpmath.rs
fedgiac: My understanding of this is that `lazy_static` creates a new type for every `static ref` we define with it. This type can be dereferenced to the definition type, which is why I use the `*`.
I think it's like that because the thing defined by lazy_static is an object that takes care of initializing the actual object, and not the object itself.

bh2smith: Oh right! I completely forgot about * for dereferencing. Silly me.

shared/src/balancer/logexpmath.rs
nlordell: > so that I can work with other contract functions reverting using ?.

In this case, we probably want to use `thiserror` as `anyhow` requires errors to implement `Error` (or at least `Display`) if I'm not mistaken. Also, in general, I think its nice that error types implement `std::error::Error`.

shared/src/balancer/logexpmath.rs
```diff
@@ -0,0 +1,522 @@
...
+    static ref UNSIGNED_ONE_18: U256 = U256::try_from(*ONE_18).unwrap();
+    static ref MAX_NATURAL_EXPONENT: I256 = ONE_18.checked_mul(I256::from(130_i128)).unwrap();
+    static ref MIN_NATURAL_EXPONENT: I256 = ONE_18.checked_mul(I256::from(-41_i128)).unwrap();
+    static ref LN_36_LOWER_BOUND: I256 = ONE_18.checked_sub(I256::exp10(17)).unwrap();
+    static ref LN_36_UPPER_BOUND: I256 = ONE_18.checked_add(I256::exp10(17)).unwrap();
+    static ref MILD_EXPONENT_BOUND: U256 = (U256::one() << 254_u32)
+        .checked_div(U256::try_from(*ONE_20).unwrap())
+        .unwrap();
+}
+fn constant_x_20(i: u32) -> I256 {
```
nlordell, line 24: I wonder if there is a cleaver way to make these const-fns... (not actually suggesting the change, just mildly curious :stuck_out_tongue:).

shared/src/balancer/logexpmath.rs
```diff
@@ -0,0 +1,522 @@
...
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum Errors {
+    XOutOfBounds,
+    YOutOfBounds,
+    ProductOutOfBounds,
+    InvalidExponent,
+}
+
+pub fn pow(x: U256, y: U256) -> Result<U256, Errors> {
```
nlordell, line 82: These are technically `fixed128x18` right? I know the code doesn't use this, but it might make sense to add a type alias like:
```rust
type Fixed128x18 = U256
```

Just so that its more obvious what the "unit" of the numbers are?

nlordell (approved):
:clap:. Very dense, but not so hard to review if you compare it side by side with the Solidity source. Also, very nice tests!

shared/src/balancer/logexpmath.rs
fedgiac: Makes sense! Just to be sure since I'm not familiar with the notation, you mean `fixed256x18`, right?

nlordell: I actually meant `Ufixed128x18` which is the [default fixed point size in Solidity](https://docs.soliditylang.org/en/v0.8.5/types.html#fixed-point-numbers).

The notation `ufixedMxN` implies:
- M is the number of bits of the integer part of the number number
- N is the decimals in 10^N

It can't be `fixed256x18` as that would imply that there are 256 bits for the integer which would leave no bits for the fractional part of the number.

Note that this notation is very strange to me and very particular to Solidity. Typically, fixed point numbers are denoted as `NxF` where both N and F are the number of bits of the integer and fractional components respectively. Solidity decided that the fractional "size" for fixed point numbers should be in base 10 instead of 2 - which is a little odd.

shared/src/balancer/logexpmath.rs
fedgiac: This error struct will be extended to contain other Balancer errors, so `anyhow` should not be required. On the other hand, having a backtrace of the error would be nice, since the same error will be thrown from different functions in the code as in the balancer contract. I'll make the changes when I'll introduce more errors!

nlordell: > , so anyhow should not be required

But if the calling code wants to handle the error with `?` and `anyhow` then it would be.

> I'll make the changes when I'll introduce more errors!

Sounds good to me!

shared/src/balancer/logexpmath.rs
fedgiac: > M is the number of bits of the integer part of the number number

Then it should be `ufixed197x18`, where `197 = ceil(log2(2^256÷10^18))`, right?
In the link I read:

> M represents the number of bits taken by the type

which is why I'd say 256.

nlordell: > > M represents the number of bits taken by the type
>
> which is why I'd say 256.

Ah right! I think you might be right... I'm not sure because Solidity fixed point numbers confuse me :confounded:.

shared/Cargo.toml
```diff
@@ -37,10 +38,6 @@ tracing-subscriber = "0.2"
...
 warp = "0.3"
 web3 = { version = "0.16", default-features = false }
-num-bigint = "0.3"
-mockall = "0.9"
-thiserror = "1.0"
-itertools = "0.10.1"
 
 [dev-dependencies]
-regex = "1.5.4"
+regex = "1.5.4"
```
bh2smith, line 43: ubdernit:
```suggestion
regex = "1.5.4"

```

shared/Cargo.toml
```diff
@@ -37,10 +38,6 @@ tracing-subscriber = "0.2"
 url = "2.2"
 warp = "0.3"
 web3 = { version = "0.16", default-features = false }
-num-bigint = "0.3"
-mockall = "0.9"
-thiserror = "1.0"
```
bh2smith: I see that these were all removed, but I don't see that they were taken out from any import statements. Does this mean they were just never used? Note that I know, I am at least using mockall in the shared crate.

fedgiac: All of these actually appear in the list above. It comes from here: https://github.com/gnosis/gp-v2-services/commit/7289681feba06aa23ec7601a764edf0c6d023142
It could be that GitHub shows only the merge conflicts in the merge commit: inserting the new lines above was successful and creates no diff, while these lines are a conflict and are shown (everything from here to EOF was written my original merge base).

bh2smith:
Lets merge this baby!

PR #727, opened 2021-06-14
Add generic subgraph client

nlordell:
This PR adds a generic Subgraph client that just does GraphQL query posts.

### Test Plan

Added some unit tests for serialization and deserialization of results.

### DIFF
--- a/shared/src/subgraph.rs
+++ b/shared/src/subgraph.rs
@@ -4,7 +4,7 @@ use crate::http::default_http_client;
 use anyhow::{bail, Result};
 use lazy_static::lazy_static;
 use reqwest::{Client, IntoUrl, Url};
-use serde::{de::DeserializeOwned, ser::Serializer, Deserialize, Serialize};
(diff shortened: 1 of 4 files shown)


vkgnosis (approved):

shared/src/subgraph.rs
```diff
@@ -0,0 +1,209 @@
...
+            .await?
+            .into_result()
+    }
+}
+
+/// A GraphQL query.
+#[derive(Serialize)]
+struct Query<'a> {
+    query: &'a str,
+}
```
bh2smith, line 64: Is this something that will have more content later or will we always allow queries to be arbitrary strings? This leads me to believe that graph queries have no common structure...

bh2smith (approved):

shared/src/subgraph.rs
nlordell: I just don't want to write a GraphQL parser.

josojo (approved):

fleupold (approved):

PR #728, opened 2021-06-15
Pass target transaction confirmation time as Instant instead of Duration

vkgnosis:
The current code is always picking the gas price for target_confirm_time
which is by default 30 seconds. Even if we have already for example
waited 1 minute.
With this commit we have a target confirmation point in time and the
remaining time is calculated appropriately.

Iirc we were doing it like in this commit in gpv1 so I probably made this mistake when initially creating the gas price stream here.

I have not added a test because the stream is annoying to test as it uses tokio sleep and the current time and the logic is simple. However I think I found good way to make it more testable which I will do in another PR.

### Test Plan
still compiles

### DIFF
--- a/solver/src/settlement_submission.rs
+++ b/solver/src/settlement_submission.rs
@@ -11,7 +11,7 @@ use futures::stream::StreamExt;
 use gas_estimation::GasPriceEstimating;
 use gas_price_stream::gas_price_stream;
 use primitive_types::{H160, U256};
-use std::time::Duration;
+use std::time::{Duration, Instant};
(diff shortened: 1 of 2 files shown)


bh2smith:
> we were doing it like in this commit in gpv1 

Missing link?

bh2smith (approved):

vkgnosis:
This change is correct regardless of how it is done in v1 and we are not going to change v1 now so I didn't feel like searching for that code. Was just more context from memory.

nlordell (approved):

fleupold (approved):

PR #729, opened 2021-06-15
Test gas price stream

vkgnosis:
With tokio::time it is possible to manipulate the current time and
sleeps will automatically complete while advancing time. This way we can
test the stream just by switching to the tokio time type.

### Test Plan
new test


nlordell (approved):
Always like more tests.

fleupold (approved):
Nice

PR #730, opened 2021-06-15
Use node url from env variable for tests

vkgnosis:
We have a bunch of #ignore tests that are supposed to be run against a
real node. Instead of hardcoding the node url it is nicer to take it
from an env variable so that we don't have gnosis specific urls in the
code and so that it is easier to use a custom node.

### Test Plan
tested that env var works

### DIFF
--- a/orderbook/src/account_balances.rs
+++ b/orderbook/src/account_balances.rs
@@ -191,12 +191,12 @@ mod tests {
     use contracts::ERC20Mintable;
     use ethcontract::prelude::Account;
     use hex_literal::hex;
-    use shared::transport::create_test_transport;
+    use shared::transport::create_env_test_transport;
 
     #[tokio::test]
     #[ignore]
     async fn mainnet_can_transfer() {
-        let http = create_test_transport("http://127.0.0.1:8545");
+        let http = create_env_test_transport();
         let web3 = Web3::new(http);
         let settlement = contracts::GPv2Settlement::deployed(&web3).await.unwrap();
         let allowance = settlement.allowance_manager().call().await.unwrap();
@@ -214,7 +214,7 @@ mod tests {
     #[tokio::test]
     #[ignore]
     async fn mainnet_cannot_transfer() {
-        let http = create_test_transport("http://127.0.0.1:8545");
+        let http = create_env_test_transport();
         let web3 = Web3::new(http);
         let settlement = contracts::GPv2Settlement::deployed(&web3).await.unwrap();
         let allowance = settlement.allowance_manager().call().await.unwrap();
@@ -234,7 +234,7 @@ mod tests {
     #[tokio::test]
     #[ignore]
     async fn watch_testnet_balance() {
-        let http = create_test_transport("http://127.0.0.1:8545");
+        let http = create_env_test_transport();
         let web3 = Web3::new(http);
 
         let accounts: Vec<H160> = web3.eth().accounts().await.expect("get accounts failed");
--- a/shared/src/bad_token/trace_call.rs
+++ b/shared/src/bad_token/trace_call.rs
@@ -270,7 +270,7 @@ mod tests {
     use super::*;
     use crate::{
         amm_pair_provider::{SushiswapPairProvider, UniswapPairProvider},
-        transport::create_test_transport,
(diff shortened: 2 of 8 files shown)


orderbook/src/account_balances.rs
```diff
@@ -196,7 +196,7 @@ mod tests {
     #[tokio::test]
     #[ignore]
     async fn mainnet_can_transfer() {
-        let http = create_test_transport("http://127.0.0.1:8545");
+        let http = create_test_transport(&std::env::var("NODE_URL").unwrap());
```
nlordell, line 199: nit: why not change the `create_test_transport` to just read from `env` directly instead of reading it at every call site?
```
fn create_test_transport() {
  let url = std::env::var("NODE_URL").unwrap()
  // ...
}
```

vkgnosis: I felt that create_test_transport should just create the transport without deciding on the url. But I don't mind that way either.

nlordell: Another suggestion would be to:
```
fn create_default_test_transport() {
  create_test_transport(std::env::var("NODE_URL").unwrap_or_else("http://127.0.0.1:8545"));
}
```

solver/src/oneinch_solver.rs
```diff
@@ -264,7 +265,7 @@ mod tests {
     #[tokio::test]
     #[ignore]
     async fn returns_error_on_non_mainnet() {
-        let web3 = testutil::infura("rinkeby");
+        let web3 = Web3::new(create_test_transport(&std::env::var("NODE_URL").unwrap()));
```
nlordell, line 268: This test specifically needs to connect to a non-mainnet node in order to pass.

solver/src/oneinch_solver.rs
```diff
@@ -229,7 +230,7 @@ mod tests {
     #[tokio::test]
     #[ignore]
     async fn solve_order_on_oneinch() {
-        let web3 = testutil::infura("mainnet");
+        let web3 = Web3::new(create_test_transport(&std::env::var("NODE_URL").unwrap()));
```
nlordell, line 233: This test specifically needs to connect to a mainnet node in order to pass.

solver/src/oneinch_solver.rs
vkgnosis: I understand that but imo you should set NODE_URL appropriately as you do for all the other tests.

nlordell: But `testutil::infura()` already requires setting `INFURA_PROJECT_ID` implying that you are configuring the tests for use with Infura. My only concern with using `NODE_URL` here is that it makes it impossible to run both `solve_order_on_oneinch` and `returns_error_on_non_mainnet` in a single run.

vkgnosis: On the other hand I can't run the tests if I'm not using infura.

nlordell: Hmm, good point... What I don't like though is that `cargo test -p solver -- --ignored oneinch` to run all 1Inch solver tests (including the ignored ones) can never succeed. That being said, requiring Infura configuration for the test is also quite annoying.

Maybe `env::var::RINKEBY_NODE_URL`?

shared/src/bad_token/trace_call.rs
```diff
@@ -386,7 +386,7 @@ mod tests {
     #[ignore]
     async fn mainnet_tokens() {
         // shared::tracing::initialize("orderbook::bad_token=debug,shared::transport=debug");
-        let http = create_test_transport("https://dev-openethereum.mainnet.gnosisdev.com/");
```
bh2smith, line 389: I definitely think this is a good change - regardless of how it is implemented, because I was under the impression that this node url is not really meant to be made public.

bh2smith (approved):

nlordell (approved):
I **really** like this PR in general - I was just hoping to find a good solution to the particular of being able to run all 1Inch tests simultaneously.

vkgnosis:
I changed the oneinch test to use another env variable so it can be run in one command with the other.
Added another function that directly creates the transport from the env variable.

PR #732, opened 2021-06-15
Cached token info fetching in balancer pool data.

bh2smith:
As rightfully noted by @fleupold in https://github.com/gnosis/gp-v2-services/pull/725#discussion_r651656251

We should use a global cached token info fetcher for our token decimals. As the Balancer Event Handler is not yet used we don't actually provide a way to pass one in yet, but this will come when all is connected together. Will put together a demo for this #710 (which actually uses it).

No new tests needed here.

### DIFF
--- a/shared/src/balancer/event_handler.rs
+++ b/shared/src/balancer/event_handler.rs
@@ -118,8 +118,9 @@ impl PoolDataFetching for PoolDataFetcher {
         // https://github.com/balancer-labs/balancer-v2-monorepo/blob/ce70f7663e0ac94b25ed60cb86faaa8199fd9e13/pkg/pool-utils/contracts/BasePool.sol#L497-L508
         let scaling_exponents = ordered_decimals
             .iter()
-            .map(|decimals| 18.checked_sub(decimals).ok_or_else(|| anyhow!("token with more than 18 decimals")))
-            .collect();
+            .map(|decimals| 18u8.checked_sub(*decimals))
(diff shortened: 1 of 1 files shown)


shared/src/balancer/event_handler.rs
```diff
@@ -101,27 +107,18 @@ impl PoolDataFetching for Web3 {
...
-        let scaling_exponents = token_decimals
+            .map(|token| token_decimals.get(token).and_then(|t| t.decimals))
+            .collect::<Option<Vec<_>>>()
+            .ok_or_else(|| anyhow!("all token decimals required to build scaling factors"))?;
+        let scaling_exponents = ordered_decimals
             .iter()
-            .map(|decimals| {
-                // Note that balancer does not support tokens with more than 18 decimals
-                // https://github.com/balancer-labs/balancer-v2-monorepo/blob/ce70f7663e0ac94b25ed60cb86faaa8199fd9e13/pkg/pool-utils/contracts/BasePool.sol#L497-L508
-                18 - decimals
```
nlordell, line 123: I think the comment is still relative.
Also, I relized that this is slightly incorrect and should be:
```
18.checked_sub(decimals).ok_or_else(|| anyhow!("token with more than 18 decimals"))
```

Otherwise you could get wrapping behaviour in release mode.

nlordell (commented):
Looks good. I think the `scaling_exponents` needs a small fix.

nlordell (approved):

Issue #657, opened 2021-05-26
Bad token detector should check tokens that cannot be approved with balance.


nlordell:
Some tokens like [RARE v2](https://blockscout.com/poa/xdai/address/0x57e93BB58268dE818B42E3795c97BAD58aFCD3Fe/contracts) have rules on `approve` that prevent allowances greater than the current balance. This causes issues with out solver where an the `approve` interaction leading up to the Honeyswap `swap` interaction would fail.

This is currently not detected by the bad token detector.

fleupold:
After thinking about it more, I actually like your suggestion to check in the settlement handling logic if we can approve uint max and if not only approve the required amount more as it's not more complex and leads to better UX

PR #736, opened 2021-06-16
Deny tokens that don't allow approving more than balance

vkgnosis:
Related to https://github.com/gnosis/gp-v2-services/issues/657 . There are tokens that only allow approving current balance which causes us to fail simulations. This PR improves the status quo by deny listing these tokens instead.
A further improvement outlined in the issue is detecting this situation and only approving the balance. I was working on this too but it got more complicated than expected so I wanted to get this PR in first.

### Test Plan
adapted test and checked that a token properly gets deny listed that used to be allowed


shared/src/bad_token/trace_call.rs
```diff
@@ -369,6 +380,30 @@ mod tests {
...
+                    result: Some(Res::Call(CallResult {
+                        gas_used: 1.into(),
+                        output: Bytes(Vec::new()),
+                    })),
+                    error: None,
+                }]),
+                vm_trace: None,
+                state_diff: None,
+                transaction_hash: None,
+            },
```
bh2smith, line 406: Maybe I am missing something, but which part of this `BlockTrace` indicated that the simulation failed on infinite token approval?

bh2smith (approved):
Very nice, just one question in the test.

shared/src/bad_token/trace_call.rs
vkgnosis: That test is called handle_response_ok and as such does not test the error case. The error case test is the #ignored mainnet one.

nlordell (approved):
Nice. In the long run, we may want to allow tokens like this by marking them as "approval limited by balance" so that the generated solution can approve the exact traded amounts. We could even have the orderbook service charge a premium on the fee in those cases to offset the additional cost to the solver.

fleupold (approved):
This logic could also go in the non-trace based bad token checks, couldn't it?

vkgnosis:
It doesn't need the trace api so it could but I still feel it fits in the bad token detector better.

PR #740, opened 2021-06-17
Add Balancer fixed point numbers

fedgiac:
Implements Balancer fixed point contract math. Will be later used to compute amounts in/out when trading with Balancer.

### Test Plan
New unit tests.

### DIFF
--- a/shared/src/balancer/swap/fixed_point.rs
+++ b/shared/src/balancer/swap/fixed_point.rs
@@ -72,8 +72,6 @@ impl Debug for Bfp {
 }
 
 impl Bfp {
-    pub const MAX: Self = Self(U256::MAX);
-
     pub fn as_uint256(self) -> U256 {
         self.0
     }
@@ -84,10 +82,6 @@ impl Bfp {
 
     pub fn one() -> Self {
         *ONE
-    }
-
-    pub fn epsilon() -> Self {
-        *EPSILON
     }
 
     pub fn from_wei(num: U256) -> Self {
@@ -192,7 +186,7 @@ mod tests {
         assert_eq!(Bfp::from(40).add(2.into()).unwrap(), 42.into());
 
         assert_eq!(
-            Bfp::MAX.add(Bfp::epsilon()).unwrap_err(),
+            Bfp(U256::MAX).add(*EPSILON).unwrap_err(),
             Error::AddOverflow
         );
     }
@@ -228,8 +222,8 @@ mod tests {
         test_mul!(mul_up);
 
         let one_half = Bfp((5 * 10_u128.pow(17)).into());
-        assert_eq!(Bfp::epsilon().mul_down(one_half).unwrap(), Bfp::zero());
-        assert_eq!(Bfp::epsilon().mul_up(one_half).unwrap(), Bfp::epsilon());
+        assert_eq!(EPSILON.mul_down(one_half).unwrap(), Bfp::zero());
+        assert_eq!(EPSILON.mul_up(one_half).unwrap(), *EPSILON);
     }
 
     macro_rules! test_div {
@@ -255,8 +249,8 @@ mod tests {
         test_div!(div_down);
         test_div!(div_up);
 
-        assert_eq!(Bfp::epsilon().div_down(2.into()).unwrap(), Bfp::zero());
-        assert_eq!(Bfp::epsilon().div_up(2.into()).unwrap(), Bfp::epsilon());
+        assert_eq!(EPSILON.div_down(2.into()).unwrap(), Bfp::zero());
+        assert_eq!(EPSILON.div_up(2.into()).unwrap(), *EPSILON);
     }
 
     #[test]
@@ -275,7 +269,7 @@ mod tests {
         ); // powDown: 0
 
         assert_eq!(
-            Bfp::MAX.pow_up(Bfp::one()).unwrap_err(),
+            Bfp(U256::MAX).pow_up(Bfp::one()).unwrap_err(),
             Error::XOutOfBounds,
         );
         // note: the values were chosen to get a large value from `pow`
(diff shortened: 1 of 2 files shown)


shared/src/balancer/swap/fixed_point.rs
```diff
@@ -1,3 +1,307 @@
...
+        *ZERO
+    }
+
+    pub fn one() -> Self {
+        *ONE
+    }
+
+    pub fn epsilon() -> Self {
+        *EPSILON
+    }
```
bh2smith, line 92: If these lazy static constants are only defined to be used here, do we really gain all that much from just putting their declarations right here?

shared/src/balancer/swap/fixed_point.rs
```diff
@@ -1,3 +1,307 @@
...
+
+    pub fn from_wei(num: U256) -> Self {
+        Self(num)
+    }
```
bh2smith, line 96: I would expect the `from_wei` is at least a bit more involved than this.

bh2smith (approved):
Wow a lot of raw details before even touching the actual equations! Looking good @fedgiac - you are on a roll!

shared/src/balancer/swap/fixed_point.rs
```diff
@@ -1,3 +1,307 @@
...
+        let mut split_dot = s.split('.');
+        let units = split_dot.next().unwrap();
+        let decimals = split_dot.next().unwrap_or("0");
+        if split_dot.next().is_some() {
+            bail!("Too many decimal separators");
+        }
+        if units.is_empty() || decimals.is_empty() || decimals.len() > 18 {
+            bail!("Invalid decimal representation");
+        }
+        Ok(Bfp(U256::from_dec_str(&format!("{:0<18}", decimals))?
```
vkgnosis, line 54: Does `{:0<18}` do anything? I think it left pads the value with zeroes so that it is always 18 characters long but this will not change the represented number.

shared/src/balancer/swap/fixed_point.rs
```diff
@@ -1,3 +1,307 @@
...
+        assert_eq!(Bfp::from(50).sub(8.into()).unwrap(), 42.into());
+
+        assert_eq!(
+            Bfp::one().sub(Bfp(*ONE_18 + 1)).unwrap_err(),
+            Error::SubOverflow
+        );
+    }
+
+    macro_rules! test_mul {
+        ($fn_name:ident) => {
```
vkgnosis, line 212: Maybe a test where mul|div up|down return different values?

vkgnosis (approved):

shared/src/balancer/swap/fixed_point.rs
fedgiac: Mmh, you are right, it looks like most of these constants won't be used outside this file. Should I move them to the test section or should I just remove them and use the constants directly?

shared/src/balancer/swap/fixed_point.rs
fedgiac: This function exists because I don't want to make the constructor public. What I think could be an issue with that is the confusion that creating something like `Bfp(100)` is not going to create 100.0, but 10^-16. So to make it super clear that decimals are involved I called the only external constructor `from_wei`, just because luckily it happens that Balancer picked 18 decimals for their fixed point representation. So it's really that easy! :smile:

shared/src/balancer/swap/fixed_point.rs
fedgiac: This pads with zeros to the right, it's needed because for example the number `1.01` is parsed as `decimals = 01`, and after padding to the left we get the right number of wei for the decimal: `format!("{:0<18}", decimals) = 010000000000000000`, which becomes `10000000000000000`. I'll add a comment about it.

shared/src/balancer/swap/fixed_point.rs
fedgiac: I tested the differerent behavior [here](https://github.com/gnosis/gp-v2-services/pull/740/files#diff-becc6a22d1880dbaa103dad137a9f7b67cb8a7f5276329373edd20f459d1b0ddR231-R233) for `mul` and [here](
https://github.com/gnosis/gp-v2-services/pull/740/files#diff-becc6a22d1880dbaa103dad137a9f7b67cb8a7f5276329373edd20f459d1b0ddR259-R260) for `div`. I used `.epsilon()` because it was easier to come up with a test case, but maybe I should use a proper number?

shared/src/balancer/swap/fixed_point.rs
vkgnosis: If it pads to the right then that makes perfect sense. I misremembered how the format string works.

shared/src/balancer/swap/fixed_point.rs
```diff
@@ -1,3 +1,307 @@
...
+        Self(U256::from(num).checked_mul(*ONE_18).unwrap())
+    }
+}
+
+impl FromStr for Bfp {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut split_dot = s.split('.');
+        let units = split_dot.next().unwrap();
```
nlordell, line 46: nit: Could you add a brief comment or use `expect` to add that `split` always has at least one element? Personally I'm a fan of having `unwrap`s include enough information to make it immediately obvious why its not expected to panic.

nlordell (approved):
Wonderful! I need this for my Balancer subgraph as they use `BigDecimal`s to store weigths in the subgraph.

shared/src/balancer/swap/fixed_point.rs
bh2smith: Fine either way. Maybe to follow the pattern of what you've been doing just use the constants directly.

fedgiac: I've removed `EPSILON` and `MAX`, but the others are actually needed.

PR #741, opened 2021-06-17
Collect database query time metrics

### LINKED ISSUE #739: Instrument DB query roundtrip time
We have been seeing some sqlx warnings of the kind:

> WARN sqlx::query: /* SQLx ping */; rows: 0, elapsed: 1.293s  

which made me think that it would probably be a good idea to collect prometheus metrics for our db query execution times to see if some queries take longer than expected.

vkgnosis:
Fixes #739 

### Test Plan
run locally and see that some metrics are populated


nlordell (approved):

orderbook/src/database/events.rs
```diff
@@ -48,17 +48,21 @@ impl Database {
                 (SELECT COALESCE(MAX(block_number), 0) FROM trades), \
                 (SELECT COALESCE(MAX(block_number), 0) FROM settlements), \
                 (SELECT COALESCE(MAX(block_number), 0) FROM invalidations));";
-        let block_number: i64 = sqlx::query_scalar(QUERY)
+        let start = Instant::now();
```
taminomara, line 51: It's unfortunate that sqlx doesn't provide any way to add these metrics in one place (at least I couldn't find any).

Perhaps we can add a RAII object to avoid manually measuring time? Just a thought.

vkgnosis: I think the rust prometheus library does provide a RAII guard or we could implement our own. We have already been measuring time manually like this in a few places so I did it the same way. But I am definitely open to improving it everywhere.

fleupold: There is https://docs.rs/prometheus/0.12.0/prometheus/struct.HistogramTimer.html

IIrc, I started measuring time manually for the API requests, because we needed the `Reply` in order to know which labels we want the metric to be on (cf [here](https://github.com/gnosis/oba-services/blob/ece6ab9cc9226772a3671bdacf9af220aef95d24/orderbook/src/metrics.rs#L122)). I couldn't find a way to set labels after starting the timer so I just took the start instant manually.

fleupold (approved):
Maybe making `Database` a trait and implementing `InstrumentedDatabase` as a wrapper around the basic methods (just measuring time of `inner.method()`) would make it slightly safer to not forget adding new metrics when a new method is added?

Right now I feel like we rely on a future contributor to check the other methods or remember that we want metrics for all our queries. If they have to add a new method in 2 implementations they'd be forced to think about instrumenting the new method.

orderbook/src/database/events.rs
bh2smith: I wonder if this could be worked around by wrapping the queries in a `timed_query` and only having to start and stop the timer in one place...

In python, this could then be invoked by adding a "decorator" to each of the queries, I wonder if there is something similar here...

```

fn timed_query(query, message_string) {
    start_time
    inner_query(query)
    stop_time
    log_metrics(message_string)
}

@timed_query(message_string = "some metric table name")
fn inner_query(query) {}
```

bh2smith (approved):
Seems good although it would be really nice not to have to start and stop the timer on every db query.

vkgnosis:
Made use of the new database related traits and used the timer that prometheus provides instead of manually keeping track of time elapsed.

nlordell (approved):
Looks great!

bh2smith (approved):
Looks cool, that change to the event updater is a bit weird looking. Could you elaborate a bit on why the database is in two places there now (once inside the mutex and once outside?

Also I see all the unused start timers but I have not been able to find a stop timer anywhere.

orderbook/src/database/instrumented.rs
```diff
@@ -0,0 +1,147 @@
...
+            .database_query_histogram("cancel_order")
+            .start_timer();
+        self.inner.cancel_order(order_uid, now).await
+    }
+
+    fn orders<'a>(
+        &'a self,
+        filter: &'a super::orders::OrderFilter,
+    ) -> futures::stream::BoxStream<'a, anyhow::Result<model::order::Order>> {
+        self.inner.orders(filter)
```
taminomara, line 125: Some calls don't have a timer, is this intentional?

taminomara (approved):
Nice 👍

vkgnosis:
> Looks cool, that change to the event updater is a bit weird looking. Could you elaborate a bit on why the database is in two places there now (once inside the mutex and once outside?

Database turned into a generic type which is probably what you mean. It is only stored once like before.

> Also I see all the unused start timers but I have not been able to find a stop timer anywhere.

Timers report the time when they go out of scope (when the function returns).

taminomara:
> Timers report the time when they go out of scope (when the function returns).

FYI, this idiom is known as [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization). With it, you don't have to stop timer manually, hence you can't forget it.

PR #742, opened 2021-06-17
Add helper function to compute Balancer traded amounts

fedgiac:
This PR introduces a helper function to compute the exact amounts traded on Balancer. These functions are taken from the `WeightedMath.sol` contract.

After this PR, all is ready to implement `BaselineSolvable` for `BalancerPool`.

### Test Plan
New unit tests.


vkgnosis (approved):

nlordell (approved):
Nice!

shared/src/balancer/swap/weighted_math.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// https://github.com/balancer-labs/balancer-v2-monorepo/blob/6c9e24e22d0c46cca6dd15861d3d33da61a60b98/pkg/core/contracts/pools/weighted/WeightedMath.sol#L69-L100
+pub fn calc_out_given_in(
+    balance_in: Bfp,
+    weight_in: Bfp,
+    balance_out: Bfp,
+    weight_out: Bfp,
+    amount_in: Bfp,
+) -> Result<Bfp, Error> {
```
bh2smith, line 27: While this may not be necessary, I see there is nothing binding this method to a particular or any WeightedPool. While the method is internal to the balancer module, it still could be defined explicitly for a weighted pool. This might allow us to use attributes from the pool instance like (like weight_in/out, balance_in/out) by just providing the in/out tokens and amount_in.

Just an idea though as it would look a lot more like 

http://github.com/gnosis/gp-v2-services/blob/7c00a326640901ccdbf3414f50d7a13024adf2e5/shared/src/baseline_solver.rs#L11-L15

However, as you mentioned, we would use these to implement BaselineSolvable so I suppose it was your intention to do this next while making use of these emulated contract functions.

bh2smith (approved):
Looks good, just wondering if we might want to skip the direct contract code mirroring and define these functions directly on the Weighted Pool itself.

shared/src/balancer/swap/weighted_math.rs
nlordell: I think the idea is to use this in the `BaseLineSolvable` implementation for a weighted pool. So something like:
```rust
impl BaseLineSolvable for WeighedPool {
  fn get_amount_out(&self, out_token: H160, in_amount: U256, in_token: H160) -> Option<U256> {
    let reserve_out = self.reserives.get(&out_token)?;
    let reserve_in = self.reserves.get(&in_token)?;
    calc_out_given_in(reserve_in.balance, reserve_in.weight, resrve_out.balance, reserve_out.weight, in_amount).ok()
  }
}
```

This way the method stays nice and easy to test (and check against the Balancer SC implementation) while allowing it to eventually be tied to a weighted pool (which has some other unrelated logic associated with it - such as getting the reserves for the specified tokens).

fedgiac: Everything that Nick says is right.
I want to be as close as possible to the smart contract to make it easy to verify, since the actual implementation of `BaselineSolvable` must be different because the interface is different.

PR #745, opened 2021-06-17
Make SingleOrderSolver return Result<Option<_>>

fleupold:
Currently we cannot distinguish between benign failures to match a single order (e.g. limit price not being satisfied) and actual errors (e.g. API request failed).

This PR changes the return type of the SingleOrderSolver trait from a pure Result to a Result<Option> - Err indicating an unexpected (alert-worthy) error and Ok(None) indicating a benign lack of a solution.

### Test Plan
CI

### DIFF
--- a/solver/src/solver/oneinch_solver.rs
+++ b/solver/src/solver/oneinch_solver.rs
@@ -77,7 +77,7 @@ impl OneInchSolver {
         &self,
         order: LimitOrder,
         protocols: Option<Vec<String>>,
-    ) -> Result<Settlement> {
+    ) -> Result<Option<Settlement>> {
         debug_assert_eq!(
             order.kind,
             OrderKind::Sell,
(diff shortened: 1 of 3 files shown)


vkgnosis (approved):

taminomara (approved):

solver/src/solver/oneinch_solver.rs
```diff
@@ -113,10 +113,11 @@ impl OneInchSolver {
         tracing::debug!("querying 1Inch swap api with {:?}", query);
         let swap = self.client.get_swap(query).await?;
 
-        ensure!(
-            swap.to_token_amount >= order.buy_amount,
-            "order limit price not respected",
-        );
+        if !swap.to_token_amount >= order.buy_amount {
```
fleupold, line 116: This is a pretty bad bug - it's not the same `!(swap.to_token_amount >= order.buy_amount)`

PR #746, opened 2021-06-17
[Trivial] Fix clippy warning in balancer pool_storage

fleupold:
☝️ 

not sure why @bh2smith's IDE doesn't seem to flag this issue.

### Test Plan
Run clippy without complaints.


vkgnosis (approved):
I don't mind this change but I think this might be your clippy being out of date. If there was a clippy warning CI would have prevented the PR from being merged.

nlordell (approved):
This can also be a `rust-analyzer` thing. I don't think its `clippy` because CI didn't catch it, and neither did `cargo clippy` locally.

fleupold:
I believe this used to be a default lint in an older version of clippy: some more background https://github.com/rust-lang/rust-clippy/issues/6726

Issue #679, opened 2021-06-01
Combined fee/amount endpoint for frontend


fleupold:
To avoid multiple roundtrips to the backend and ensure a congruent fee for the given sell amount (after fee) it would be nice if the backend exposed an endpoint that allows for a given token pair, order kind and amount to return the following:

- min fee that needs to be specified with this order
- estimated amount **after the min fee has been deducted** (sell amount for buy orders, buy amount for sell orders)
- price impact - difference (in USDC value) of buy tokens vs sell tokens.

We could potentially repurpose https://protocol-rinkeby.dev.gnosisdev.com/api/#/default/get_api_v1_markets__baseToken___quoteToken___kind___amount_ for this, but this might break existing clients so we might need a new route.

nlordell:
> we might need a new route.

I think this is a good use for the version - `/api/v2/markets/{baseToken}-{quoteToken}/{kind}/{amount}`.

fleupold:
That's a good suggestion. I wonder if it is slightly weird to have different versions on the same API in use at the same time, but probably not a problem.

nlordell:
I noticed the Safe team does this for their relay service: https://safe-relay.gnosis.io/

vkgnosis:
Nothing wrong with incrementing one version imo.

vkgnosis:
This has been implemented with the exception of
> price impact - difference (in USDC value) of buy tokens vs sell tokens.

This can be added to the return object of the route in another PR.

PR #747, opened 2021-06-17
Add combined fee and price route

vkgnosis:
Part of https://github.com/gnosis/gp-v2-services/issues/679 . See that issue for motivation.

In this PR I add a naive implementation of the combined route. This is what afaik the front end does at the moment just done by the backend. After this PR we can improve the implementation to take the amounts into account like with Felix' more accurate fee route.

I would like to discuss how the route should look like before I add the openapi documentation. I have chosen the structure of the fee route which passes all the parameters through the query. This is different from the markets route which passes them in the path. I prefer the query version.

I haven't added the price impact that is mentioned in the issue because I wasn't sure how to correctly calculate it.

I chose a new route instead of a v2 of an existing route because I the new name is clearer but I don't mind changing this.

### Test Plan
new unit tests and manual test (TODO)

### DIFF
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -296,6 +296,84 @@ paths:
           description: Token non-existent or no valid price found
         500:
           description: Unexpected internal error while processing the request
+  /api/v1/feeAndQuote/sell:
+    get:
+      description: |
+        For a total available amount of sell_token returns the fee in the sell token and the
+        resulting buy amount after the fee has been deducted.
+        sellAmountBeforeFee is the total amount that is available for the order. From it the fee
+        is deducted and the buy amount is calculated.
+      parameters:
+        - name: sellToken
+          in: query
+          schema:
+            $ref: "#/components/schemas/Address"
+          required: true
+        - name: buyToken
+          in: query
+          schema:
+            $ref: "#/components/schemas/Address"
+          required: true
+        - name: sellAmountBeforeFee
+          in: query
+          schema:
+            $ref: "#/components/schemas/TokenAmount"
+          required: true
+      responses:
+        200:
+          description: ok
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/FeeAndQuoteSellResponse"
+        400:
+          description: Error with the input data.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/FeeAndQuoteError"
+        404:
+          description: Token non-existent or not connected to native token
+        500:
+          description: Unexpected internal error while processing the request
+  /api/v1/feeAndQuote/buy:
+    get:
+      description: |
+        For a target buy amount returns the total sell_amount that is needed and how much of it is
+        the fee.
+      parameters:
+        - name: sellToken
+          in: query
+          schema:
+            $ref: "#/components/schemas/Address"
+          required: true
+        - name: buyToken
+          in: query
+          schema:
+            $ref: "#/components/schemas/Address"
+          required: true
+        - name: buyAmountAfterFee
+          in: query
+          schema:
+            $ref: "#/components/schemas/TokenAmount"
+          required: true
+      responses:
+        200:
+          description: ok
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/FeeAndQuoteBuyResponse"
+        400:
+          description: Error with the input data.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/FeeAndQuoteError"
+        404:
+          description: Token non-existent or not connected to native token
+        500:
+          description: Unexpected internal error while processing the request
 components:
   schemas:
     TransactionHash:
@@ -589,6 +667,33 @@ components:
         errorType:
           type: string
           enum: [InvalidSignature, WrongOwner, OrderNotFound, AlreadyCancelled, OrderFullyExecuted, OrderExpired]
+        description:
+          type: string
+      required:
+        - errorType
+        - description
+    FeeAndQuoteSellResponse:
+      type: object
+      properties:
+        fee:
+          $ref: "#/components/schemas/FeeInformation"
+        buyAmountAfterFee:
+          description: The buy amount after deducting the fee.
+          $ref: "#/components/schemas/TokenAmount"
+    FeeAndQuoteBuyResponse:
+      type: object
+      properties:
+        fee:
+          $ref: "#/components/schemas/FeeInformation"
+        sellAmountBeforeFee:
+          description: The sell amount including the fee.
+          $ref: "#/components/schemas/TokenAmount"
+    FeeAndQuoteError:
+      type: object
+      properties:
+        errorType:
+          type: string
+          enum: ["NoLiquidity", "UnsupportedToken", "AmountIsZero", "SellAmountDoesNotCoverFee"]
         description:
           type: string
       required:
--- a/orderbook/src/api/get_fee_and_quote.rs
+++ b/orderbook/src/api/get_fee_and_quote.rs
@@ -1,0 +1,238 @@
+use crate::fee::{MinFeeCalculating, MinFeeCalculationError};
+use anyhow::Result;
+use chrono::{DateTime, Utc};
+use ethcontract::{H160, U256};
+use model::h160_hexadecimal;
+use model::{order::OrderKind, u256_decimal};
+use serde::{Deserialize, Serialize};
+use shared::{
+    conversions::{big_int_to_u256, U256Ext},
+    price_estimate::{PriceEstimating, PriceEstimationError},
+};
+use std::convert::Infallible;
+use std::sync::Arc;
+use warp::{hyper::StatusCode, reply, Filter, Rejection, Reply};
+
+#[derive(Deserialize)]
+struct Query {
+    #[serde(with = "h160_hexadecimal")]
+    sell_token: H160,
+    #[serde(with = "h160_hexadecimal")]
+    buy_token: H160,
+    // For sell orders in sell token, for buy orders in the buy token.
+    #[serde(with = "u256_decimal")]
+    in_amount_before_fees: U256,
+    kind: OrderKind,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct Response {
+    expiration_date: DateTime<Utc>,
+    #[serde(with = "u256_decimal")]
+    min_fee: U256,
+    // For sell orders in buy token, for buy orders in the sell token.
+    #[serde(with = "u256_decimal")]
+    out_amount_before_fees: U256,
+    // For buy orders this is the same as out_amount_before_fees. For sell orders this is less
+    // than out_amount_before_fees because the fee is first deducted from the sell amount before
+    // applying the price.
+    #[serde(with = "u256_decimal")]
+    out_amount_after_fees: U256,
+}
+
+#[derive(Debug)]
+enum Error {
+    NotFound,
+    UnsupportedToken(H160),
+    AmountIsZero,
+    SellAmountDoesNotCoverFee,
+    Other(anyhow::Error),
+}
+
+impl From<MinFeeCalculationError> for Error {
+    fn from(other: MinFeeCalculationError) -> Self {
+        match other {
+            MinFeeCalculationError::NotFound => Error::NotFound,
+            MinFeeCalculationError::UnsupportedToken(token) => Error::UnsupportedToken(token),
+            MinFeeCalculationError::Other(error) => Error::Other(error),
+        }
+    }
+}
+
+impl From<PriceEstimationError> for Error {
+    fn from(other: PriceEstimationError) -> Self {
+        match other {
+            PriceEstimationError::UnsupportedToken(token) => Error::UnsupportedToken(token),
+            PriceEstimationError::Other(error) => Error::Other(error),
+        }
+    }
+}
+
+fn request() -> impl Filter<Extract = (Query,), Error = Rejection> + Clone {
+    warp::path!("fee_and_quote")
+        .and(warp::get())
+        .and(warp::query::<Query>())
+}
+
+fn response(result: Result<Response, Error>) -> impl Reply {
+    match result {
+        Ok(response) => reply::with_status(reply::json(&response), StatusCode::OK),
+        Err(Error::NotFound) => reply::with_status(
+            super::error("NotFound", "Token was not found"),
+            StatusCode::NOT_FOUND,
+        ),
+        Err(Error::UnsupportedToken(token)) => reply::with_status(
+            super::error("UnsupportedToken", format!("Token address {:?}", token)),
+            StatusCode::BAD_REQUEST,
+        ),
+        Err(Error::AmountIsZero) => reply::with_status(
+            super::error(
+                "AmountIsZero",
+                "The input amount must be greater than zero.".to_string(),
+            ),
+            StatusCode::BAD_REQUEST,
+        ),
+        Err(Error::SellAmountDoesNotCoverFee) => reply::with_status(
+            super::error(
+                "SellAmountDoesNotCoverFee",
+                "The sell amount for the sell order is lower than the fee.".to_string(),
+            ),
+            StatusCode::BAD_REQUEST,
+        ),
+        Err(Error::Other(err)) => {
+            tracing::error!(?err, "get_fee_and_price error");
+            reply::with_status(super::internal_error(), StatusCode::INTERNAL_SERVER_ERROR)
+        }
+    }
+}
+
+async fn calculate(
+    fee_calculator: Arc<dyn MinFeeCalculating>,
+    price_estimator: Arc<dyn PriceEstimating>,
+    query: Query,
+) -> Result<Response, Error> {
+    if query.in_amount_before_fees.is_zero() {
+        return Err(Error::AmountIsZero);
+    }
+
+    let (min_fee, expiration_date) = fee_calculator
+        .min_fee(query.sell_token, None, None, None)
+        .await?;
+    let in_amount_after_fee = match query.kind {
+        OrderKind::Buy => query.in_amount_before_fees,
+        OrderKind::Sell => query
+            .in_amount_before_fees
+            .checked_sub(min_fee)
+            .ok_or(Error::SellAmountDoesNotCoverFee)?,
+    };
+
+    let price = price_estimator
+        .estimate_price(
+            query.sell_token,
+            query.buy_token,
+            in_amount_after_fee,
+            query.kind,
+        )
+        .await?;
+    let out_amount_before_fees = big_int_to_u256(
+        &match query.kind {
+            OrderKind::Buy => query.in_amount_before_fees.to_big_rational() * &price,
+            OrderKind::Sell => query.in_amount_before_fees.to_big_rational() / &price,
+        }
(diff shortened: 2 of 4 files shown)


orderbook/src/api/get_fee_and_price.rs
```diff
@@ -0,0 +1,186 @@
...
+    kind: OrderKind,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct Response {
+    expiration_date: DateTime<Utc>,
+    min_fee: U256,
+    #[serde(with = "serde_with::rust::display_fromstr")]
+    amount_after_fee: BigInt,
```
fleupold, line 32: This is the same as on the markets route, but isn't it a bit weird to have min_fee and amount_after_fee be different types?

orderbook/src/api/get_fee_and_price.rs
```diff
@@ -0,0 +1,186 @@
...
+    fn from(other: PriceEstimationError) -> Self {
+        match other {
+            PriceEstimationError::UnsupportedToken(token) => Error::UnsupportedToken(token),
+            PriceEstimationError::Other(error) => Error::Other(error),
+        }
+    }
+}
+
+fn request() -> impl Filter<Extract = (Query,), Error = Rejection> + Clone {
+    warp::path!("fee_and_price")
```
fleupold, line 62: nit: you are not really getting a price, but an amount (proceeds or costs) - we could also consider calling it "quote" in case maybe other relevant data may be added later (but I also like the very clear name you chose)?

orderbook/src/api/get_fee_and_price.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+}
+
+async fn calculate(
+    fee_calculator: Arc<dyn MinFeeCalculating>,
+    price_estimator: Arc<dyn PriceEstimating>,
+    query: Query,
+) -> Result<Response, Error> {
+    let (min_fee, expiration_date) = fee_calculator
+        .min_fee(query.sell_token.0, None, None, None)
```
fleupold, line 91: This should not use the legacy API but instead set buyToken, amount and kind (since it already knows it).

orderbook/src/api/get_fee_and_price.rs
```diff
@@ -0,0 +1,186 @@
...
+        .estimate_price(
+            query.sell_token.0,
+            query.buy_token.0,
+            query.amount,
+            query.kind,
+        )
+        .await?;
+    let amount_after_fee = match query.kind {
+        OrderKind::Buy => query.amount.to_big_rational() * price,
+        OrderKind::Sell => query.amount.saturating_sub(min_fee).to_big_rational() / price,
```
fleupold, line 103: Maybe to be more accurate we can do the subtraction before we estimate the price (but you also mentioned correctness is not the primary goal of this PR)?

orderbook/src/api/get_fee_and_price.rs
```diff
@@ -0,0 +1,186 @@
...
+use std::convert::Infallible;
+use std::sync::Arc;
+use warp::{hyper::StatusCode, reply, Filter, Rejection, Reply};
+
+#[derive(Deserialize)]
+struct Query {
+    sell_token: H160Wrapper,
+    buy_token: H160Wrapper,
+    #[serde(with = "u256_decimal")]
+    amount: U256,
```
fleupold, line 22: nit: maybe this could highlight that this is the `amount_before_fee`?

orderbook/src/api/get_fee_and_price.rs
```diff
@@ -0,0 +1,186 @@
...
+    kind: OrderKind,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct Response {
```
fleupold, line 28: I liked about the markets route that it tells you which token the amounts refer to (to make confusing the API's response harder). Not sure this is needed for the fee (which is always sell token) but I do think it would be nice for the amount.

orderbook/src/api/get_fee_and_price.rs
vkgnosis: This is intentional to make the implementation simpler. With the advanced implementation we need to take into account that the fee for the full sell amount is different from the new fee after deducting the full amount fee. So I want to handle that in a future PR.

bh2smith (approved):
Looks pretty good although I was not really part of the development for the two things being combined here so I'd say that @fleupold's comments seem appropriate. Once all this is complete, would we also be eliminating the possibility to query for these two things individually through the API?

vkgnosis:
Once the frontend uses the new route we should definitely remove the existing fee route as it is superseeded by this one. Not sure about the markets route yet.

orderbook/src/api/get_fee_and_price.rs
vkgnosis: For this route I think it makes more sense to not specify the token as the amount is always expressed in the other token with respect to the input amount. Adding the token only introduces ambiguity.
 I have added comments to the struct explain this.

vkgnosis:
Addressed the comments.
Added out_amount_before_fees to the response because it is likely useful for the frontend so they can pick which price they want to display.

orderbook/src/api/get_fee_and_quote.rs
```diff
@@ -0,0 +1,238 @@
...
+    price_estimate::{PriceEstimating, PriceEstimationError},
+};
+use std::convert::Infallible;
+use std::sync::Arc;
+use warp::{hyper::StatusCode, reply, Filter, Rejection, Reply};
+
+#[derive(Deserialize)]
+struct Query {
+    #[serde(with = "h160_hexadecimal")]
+    sell_token: H160,
```
anxolin, line 19: Why the fee uses `H160Wrapper` and here `H160`? 
Is the wrapper already H160 with serialize/deserialize impl?

orderbook/src/api/get_fee_and_quote.rs
```diff
@@ -0,0 +1,238 @@
...
+    // For sell orders in sell token, for buy orders in the buy token.
+    #[serde(with = "u256_decimal")]
+    in_amount_before_fees: U256,
+    kind: OrderKind,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct Response {
+    expiration_date: DateTime<Utc>,
```
anxolin, line 31: Would this be cleat that the expiration applies only to the fee? should fee and expiration be in a nested object to make this relation clearer?

orderbook/src/api/get_fee_and_quote.rs
```diff
@@ -0,0 +1,238 @@
...
+    #[serde(with = "u256_decimal")]
+    min_fee: U256,
+    // For sell orders in buy token, for buy orders in the sell token.
+    #[serde(with = "u256_decimal")]
+    out_amount_before_fees: U256,
+    // For buy orders this is the same as out_amount_before_fees. For sell orders this is less
+    // than out_amount_before_fees because the fee is first deducted from the sell amount before
+    // applying the price.
+    #[serde(with = "u256_decimal")]
+    out_amount_after_fees: U256,
```
anxolin, line 41: Nice names. Hopefully this kill any ambiguity

orderbook/src/api/get_fee_and_quote.rs
```diff
@@ -0,0 +1,238 @@
...
+}
+
+#[derive(Debug)]
+enum Error {
+    NotFound,
```
anxolin, line 46: I find weird this error. I notice it before, but never mentioned. I think the error should give a better context:
 - NotExistentToken, InssuficientLiquidity, ... (or whatever the error is modelling)

orderbook/src/api/get_fee_and_quote.rs
```diff
@@ -0,0 +1,238 @@
...
+    fn from(other: PriceEstimationError) -> Self {
+        match other {
+            PriceEstimationError::UnsupportedToken(token) => Error::UnsupportedToken(token),
+            PriceEstimationError::Other(error) => Error::Other(error),
+        }
+    }
+}
+
+fn request() -> impl Filter<Extract = (Query,), Error = Rejection> + Clone {
+    warp::path!("fee_and_price")
```
anxolin, line 73: Is this the path? 

I would literally use this one: 
image
https://protocol-rinkeby.dev.gnosisdev.com/api/#/default/get_api_v1_markets__baseToken___quoteToken___kind___amount_

It just changes the signature. But to make a gradual change, we can just create `/markets/next/...`  and after we migrate the frontends (balancer and cowswap) we kill the other two endpoints and we leave only the markers one. 

Would that make sense?

orderbook/src/api/get_fee_and_quote.rs
```diff
@@ -0,0 +1,238 @@
...
+
+#[derive(Deserialize)]
+struct Query {
+    #[serde(with = "h160_hexadecimal")]
+    sell_token: H160,
+    #[serde(with = "h160_hexadecimal")]
+    buy_token: H160,
+    // For sell orders in sell token, for buy orders in the buy token.
+    #[serde(with = "u256_decimal")]
+    in_amount_before_fees: U256,
```
anxolin, line 24: I like the name. Although for buy orders, the name might be confusing, cause it's actually after fees :)

orderbook/src/api/get_fee_and_quote.rs
anxolin: EDIT: Now im not sure 😅, will for buy orders be actually really after fees? My understanding is that is not.

anxolin (commented):
Good call! yes, we can try to adapt to this once is final, and kill the other endpoints. 

About price impact, it's a bit complex, we can leave it out, or even try to do it in the front end.
My understanding is that we estimate in USD the amounts before fees, and see the the relation between them. 

Something like:

sell orders impact = `(USD_ESTIMATION(in_amount_before_fees * price) - USD_ESTIMATION(in_amount_before_fees)) / USD_ESTIMATION(in_amount_before_fees * price)`


buy orders impact = `(USD_ESTIMATION(in_amount_before_fees) - USD_ESTIMATION(in_amount_before_fees / price) / USD_ESTIMATION(in_amount_before_fees)`


* USD price estimation, we can use some approximation like how much is 1 WETH worth of a token in some stable token. Suggesting 1 WETH cause I belive Martin was convinced of that amount, but probably we can study what's a good value, and if maybe we can set an amount in USD instead of WETH to make it simpler

orderbook/src/api/get_fee_and_quote.rs
vkgnosis: The wrapper only matter for serialization so it's not something you would see from a http standpoint. The difference is that the wrapper implement the rust trait FromStr while this serde_with only works for serialization.

orderbook/src/api/get_fee_and_quote.rs
vkgnosis: I would definitely make it clear in the swagger docs. I don't mind making a nested struct either.

orderbook/src/api/get_fee_and_quote.rs
vkgnosis: As the comment says, for buy orders the amount before and after fee is the same.
The reason is that for buy orders the fee is in the sell token but we don't know how much the user has in the sell token. We only know they want want to buy x amount in the buy token. So what is returned is the fee they need to set and the sell amount they need set, but there is nothing to deduct. Does that make sense?

orderbook/src/api/get_fee_and_quote.rs
vkgnosis: Agree this should be clearer.

orderbook/src/api/get_fee_and_quote.rs
vkgnosis: I don't have a strong opinion on the path but I like the new one a little more. The bigger change with this route is that it doesn't put the parameters in the path like the markets route. Instead it puts them in the query like the fee route. Do you have a preference there?

orderbook/src/api/get_fee_and_quote.rs
vkgnosis: I am starting to think we might want to routes to remove all ambiguity. One for selling and one for buying.

orderbook/src/api/get_fee_and_quote.rs
anxolin: I would favour the nested struct, like `struct Response { fee:  {  expiration_date: DateTime<Utc>, min_fee: U256.. } ... }`

This way, is self explanatory just by the data

orderbook/src/api/get_fee_and_quote.rs
anxolin: Yep, it does. It's hard to find infalible names when we play with fees, prices, and buy/sell, but I think is good enough

orderbook/src/api/get_fee_and_quote.rs
anxolin: I found `/markets/<base>-<quote>/sell/<amount>` and `/markets/<base>-<quote>/buy/<amount>` very declarative, but maybe is a personal preference. 

I usually like APIs to be hierarchical, like `/markets/<base>-<quote>` modeling a market, and then with the these endpoints we get a sell/buy quote. The only difference is that now we return the fee together with the quote, which imo makes sense to be part of it. 

I let other give their opinions. I won't be opposed to the new name if that is the general sentiment.

orderbook/src/api/get_fee_and_quote.rs
anxolin: If it's not too much overhead, it would be ideal yep.

orderbook/src/api/get_fee_and_price.rs
fleupold: The frontend currently uses the new API (with the full sell amount which is technically not correct but will still give a better estimate than not providing any information about the route at all). It doesn't have to be accurate as in the "verification" we still use the legacy API (effectively accepting the lowest fee that was provided over all routes/amounts for this sell token). Yet it would be nice if it was as close as possible (the same the frontend uses today)

If left like this, it would regress the amount of fee the orders are providing by default.

nlordell (approved):
Code looks good to me.

orderbook/src/api/get_fee_and_price.rs
vkgnosis: I didn't realize the verification uses the legacy api. Fixed.

orderbook/src/api/get_fee_and_quote.rs
vkgnosis: used nested struct

orderbook/src/api/get_fee_and_quote.rs
vkgnosis: fixed

vkgnosis:
I changed the api to have two separate routes `fee_and_quote/sell` and  `fee_and_quote/buy` with separate input and output structs. I feel this makes it as clear as possible how this works and resolves confusion around input and output amounts working differently depending on type of order.
After doing this I also decided to get rid of the redundant information buy_amount_before_fee, buy_amount_after_fee. The comments make it clear what the values mean and users of the api can calculate derived statistics themself. I stuck with sell_amount_before_fee and buy_amount_after fee so that the routes are symmetrical.

orderbook/src/api/get_fee_and_quote.rs
```diff
@@ -0,0 +1,238 @@
...
+use warp::{hyper::StatusCode, reply, Filter, Rejection, Reply};
+
+#[derive(Deserialize)]
+struct Query {
+    #[serde(with = "h160_hexadecimal")]
+    sell_token: H160,
+    #[serde(with = "h160_hexadecimal")]
+    buy_token: H160,
+    // For sell orders in sell token, for buy orders in the buy token.
+    #[serde(with = "u256_decimal")]
```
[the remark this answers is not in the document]
vkgnosis: serde_as is more powerful but also harder to understand. I prefer to use the simpler approach when possible and don't want to consistently use serde_as because it will make it less clear when serde_as is needed and switching away from it later potentially harder.

nlordell (approved):
Splitting them into two routes makes sense and helps avoid confusion.

orderbook/src/api/get_fee_and_quote.rs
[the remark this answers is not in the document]
nlordell: I also prefer `serde(with = ...)` (but only because its more familiar to me :stuck_out_tongue:). The comment was more understanding why we need both side by side (which can be a little confusing when looking at the codebase and trying to decide which one to use).

fleupold (approved):
Very nice and clear routing! Will you update the openapi in a separate PR?

vkgnosis:
I was waiting for consensus on the route. It seems like we're happy with this so I am going to add openapi to this PR before merging.

vkgnosis:
Added openapi docs and am testing manually right now. Any last thoughts @anxolin before merging?

orderbook/openapi.yml
```diff
@@ -296,6 +296,84 @@ paths:
           description: Token non-existent or no valid price found
         500:
           description: Unexpected internal error while processing the request
+  /api/v1/fee_and_quote/sell:
```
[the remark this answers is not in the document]
anxolin: I'm still not excited about the name, although I don't consider super important, and since there's many approvals already I assume people is happy with it, so I won't oppose

I'm approving, although leaving my thinking so it can get lost in this PR comments :) 
- why snake case? I thought we were using kebab 🍗
- I liked more how it was before, base-quote (in terms of a market), although i see for example paraswap don't use the market notation. The reason i like market notation is because for me is closer to what a user does: you are in WETH-DAI market, and you want to buy or sell WETH.
- Even if we don't like the market notation, and we want to use sell/buy token notation because it's closer to the contracts, I would prefer a simpler name `prices/sell`. For me, the fact that we return also the fee shouldn't have so much relevance to change the endpoint name. The name doesn't need to be an exaustive description of all the data. The main thing we are querying is the price.

anxolin (approved):
Good to go! Thanks for checking with us @e00E !!

orderbook/openapi.yml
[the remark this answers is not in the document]
vkgnosis: 1. We are using kebab case for the types but hasn't come up yet for routes because the rest are one word. Actually, there is solvable_orders which is snake case but I don't think we consciously thought about that. I will change this PR to use kebab case for the route too.
2. Personally I find calling things sell_amount and buy_amount more obvious than quote and base. But not sure how normal this is. I'm also not sure the market abstraction is that useful because the price and fee in one direction can be (slightly) different than in the other direction.
3. Right now we have a fee route and a markets route so I wanted this name to show this is route is different. Maybe once it is clear which routes will stay we can rename this one and think about what the best name would be. Like you said I also don't find it super important and don't want to spend too long discussing here.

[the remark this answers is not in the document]
nlordell: > Personally I find calling things sell_amount and buy_amount more obvious than quote and base

I think this is fine. We used "markets" in GPv1 for price estimation and it was a pretty large source of confusion. I think making it "sell" and "buy" might be weird to the financially inclined, but at least to me it feels less ambiguous and you can immediately populate request data without having to think too much.

> I will change this PR to use kebab case for the route too.

Personally, I prefer kebab case for URLs. I don't know why, nor do I have a good reason for it.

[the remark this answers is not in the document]
anxolin: All good 👌
Thxxx

PR #748, opened 2021-06-17
Cached Weighted Pool Fetching

bh2smith:
In order to make use of generic Cached Pool Fetching for Balancer Pools we introduce `BalancerPoolCache` as `RecentBlockCache` of Pool Ids to Weighted Pools. This will require an intermediary StaticPoolFetching from `TokenPairs` to  `RegisteredWeightedPool` which then caches the WeightedPool (with reserves) for each pool Id.

### Test Plan

Testing rearranged and new helper method to fetch pools and their Id's from the registry.


shared/src/balancer/pool_cache.rs
```diff
@@ -0,0 +1,144 @@
...
+            vault,
+            web3,
+        })
+    }
+}
+
+pub type BalancerPoolReserveCache =
+    RecentBlockCache<H256, WeightedPool, PoolReserveFetcher, Arc<dyn PoolCacheMetrics>>;
+
+impl CacheKey<WeightedPool> for H256 {
```
fleupold, line 40: Using H256 as the Key is very clever. I first thought we would use TokenPair and have to adapt the RecentBlockCache to work for surjective mappings (`for_values -> Vec<Self>` instead `for_value -> Self`), since one Pool can map to many token pairs, but this is much better.

shared/src/balancer/event_handler.rs
```diff
@@ -88,8 +88,8 @@ impl BalancerPoolRegistry {
     pub async fn get_pools_containing_token_pairs(
         &self,
         token_pairs: HashSet<TokenPair>,
-    ) -> Vec<RegisteredWeightedPool> {
-        let mut pool_set_1 = self
+    ) -> HashSet<H256> {
```
fleupold, line 91: nit: maybe this method should now be called get_pool_ids_containing_token_pairs?

shared/src/balancer/pool_storage.rs
```diff
@@ -135,27 +133,28 @@ impl PoolStorage {
...
-            .collect::<Vec<RegisteredWeightedPool>>()
+        pools_0.intersection(pools_1).copied().collect()
     }
 
     /// Given a collection of `TokenPair`, returns all pools containing at least one of the pairs.
-    pub fn pools_containing_token_pairs(
-        &self,
-        token_pairs: HashSet<TokenPair>,
-    ) -> Vec<RegisteredWeightedPool> {
+    pub fn pools_containing_token_pairs(&self, token_pairs: HashSet<TokenPair>) -> HashSet<H256> {
```
fleupold, line 140: ditto: pool_ids_...

fleupold (commented):
Looks very nice!

shared/src/balancer/pool_cache.rs
bh2smith: Glad you like it! This was my attempt at not having to change the `RecentBlockCache`!

nlordell (approved):
Looks good to me!

PR #752, opened 2021-06-17
Rust 1.53 clippy warnings

fleupold:
After updating rust to v1.53 I ran into a ton of more clippy complaints, which this PR tries to fix

### Test Plan
clippy itself


nlordell (approved):
Yay new version! And thanks :smile:.

PR #753, opened 2021-06-17
749/sell order status

### LINKED ISSUE #749: Order status for cancelled sell order is wrong
# Summary

With the recently added order status (#715), the calculation for execution of sell orders is incorrect.

See this order for example: https://protocol-rinkeby.dev.gnosisdev.com/api/v1/orders/0x75a87ef7c11baa3968d90671defadce607c1fe3772d3eab8479d7cf26b9739275b0abe214ab7875562adee331deff0fe1912fe4260ca7a07

The status is `cancelled`, but `sellAmount` (23677366631019630106934) matches  `executedSellAmountBeforeFees` (23677366631019630106934), while it does not match `executedSellAmount` (23682222438767276389686).
I've used `executedSellAmount` without considering the fees when calculating the status.

On explorer for instance I'm using the right way of calculating the executed sell amount https://protocol-explorer.dev.gnosisdev.com/rinkeby/orders/0x75a87ef7c11baa3968d90671defadce607c1fe3772d3eab8479d7cf26b9739275b0abe214ab7875562adee331deff0fe1912fe4260ca7a07 but I did not not that on the API.

alfetopito:
# Summary
Fixes #749 

Status calculation for sell orders did not take into account the executed fee

### Test Plan
Unit tests

### DIFF
--- a/orderbook/src/database/orders.rs
+++ b/orderbook/src/database/orders.rs
@@ -201,6 +201,18 @@ struct OrdersQueryRow {
     signing_scheme: DbSigningScheme,
 }
 
+fn is_sell_order_filled(
+    amount: &BigDecimal,
+    executed_amount: &BigDecimal,
+    executed_fee: &BigDecimal,
+) -> bool {
+    if executed_amount.is_zero() {
+        return false;
+    }
+    let total_amount = executed_amount + executed_fee;
+    total_amount == *amount
+}
+
 impl OrdersQueryRow {
     fn calculate_status(&self) -> OrderStatus {
         match self.kind {
@@ -210,7 +222,7 @@ impl OrdersQueryRow {
                 }
             }
             DbOrderKind::Sell => {
-                if !self.sum_sell.is_zero() && self.sum_sell == self.sell_amount {
+                if is_sell_order_filled(&self.sell_amount, &self.sum_sell, &self.sum_fee) {
                     return OrderStatus::Fulfilled;
                 }
             }
--- a/orderbook/src/database/orders.rs
+++ b/orderbook/src/database/orders.rs
@@ -344,8 +344,9 @@ mod tests {
         // Filled - sell (filled - 100%)
         let order_row = OrdersQueryRow {
             kind: DbOrderKind::Sell,
-            sell_amount: BigDecimal::from(1),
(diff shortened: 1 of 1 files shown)


orderbook/src/database/orders.rs
```diff
@@ -321,7 +333,11 @@ mod tests {
             signing_scheme: DbSigningScheme::Eip712,
         };
 
-        assert_eq!(order_row.calculate_status(), OrderStatus::Open);
+        assert_eq!(
+            order_row.calculate_status(),
+            OrderStatus::Open,
+            "Open - sell (filled - 0%)"
+        );
```
vkgnosis, line 340: The third argument to the assert! macros is the panic message that is printed when the assert fails. I don't think the message is useful here and would put this as a comment above if it is there to help readers understand what is being tested. Not a strong opinion, just feels slightly weird.

vkgnosis (approved):

orderbook/src/database/orders.rs
```diff
@@ -201,6 +201,18 @@ struct OrdersQueryRow {
     signing_scheme: DbSigningScheme,
 }
 
+fn is_sell_order_filled(
```
bh2smith, line 204: This is cool, but is there also an is buy order filled to accompany it?

bh2smith (approved):
Looks good, just one question about the new methods symmetric counterpart. Also, you don't have to manually request our review as the PR automatically notifies the backend team @gnosis/gp-services for review.

orderbook/src/database/orders.rs
alfetopito: I've added them because there are many asserts in this single test function.
When it fails, it doesn't tell me which one did, unless I'm missing something?

That's why I've added the panic message.

orderbook/src/database/orders.rs
alfetopito: I can definitely add one, and I'll now that you mention it.
Just haven't because there's no special case for buy orders.

orderbook/src/database/orders.rs
alfetopito: Well, I did miss the line number on the error, my bad.
Will revert the change

PR #754, opened 2021-06-18
Make Database a trait

vkgnosis:
For https://github.com/gnosis/gp-v2-services/pull/741 it has been suggested to make the database a trait so that we wrap it with with a version that collects the query metrics.
The trait also has the advantage that consumers of the Database can now mock it if they want to.
It has the disadvantage that we are using Arc<dyn Database> instead of just Database in these places. When we introduced the database struct I was against this but now I think it is a slight advantage again. I also don't like repeating a bunch of function signatures several times to implement Database, MinFeeStoring, etc.

### Test Plan
existing tests still work

### DIFF
--- a/orderbook/src/database.rs
+++ b/orderbook/src/database.rs
@@ -3,13 +3,68 @@ mod fees;
 mod orders;
 mod trades;
 
-use anyhow::Result;
+use anyhow::{Context, Result};
+use chrono::{DateTime, Utc};
+use contracts::gpv2_settlement::Event as ContractEvent;
+use ethcontract::Event as EthContractEvent;
+use futures::{stream::BoxStream, StreamExt};
+use model::{
+    order::{Order, OrderKind, OrderUid},
+    trade::Trade as ModelTrade,
+};
+use primitive_types::{H160, U256};
+use shared::{event_handling::EventIndex, maintenance::Maintaining};
 use sqlx::{Executor, PgPool, Row};
 use std::collections::HashMap;
 
 pub use events::*;
 pub use orders::OrderFilter;
 pub use trades::TradeFilter;
+
+use crate::fee::MinFeeStoring;
+
+#[async_trait::async_trait]
+pub trait Database: Send + Sync {
+    async fn clear(&self) -> Result<()>;
+    async fn count_rows_in_tables(&self) -> Result<HashMap<&'static str, i64>>;
+
+    async fn block_number_of_most_recent_event(&self) -> Result<u64>;
+    async fn insert_events(&self, events: Vec<(EventIndex, Event)>) -> Result<()>;
+    async fn replace_events(
+        &self,
+        delete_from_block_number: u64,
+        events: Vec<(EventIndex, Event)>,
+    ) -> Result<()>;
+    fn contract_to_db_events(
+        &self,
+        contract_events: Vec<EthContractEvent<ContractEvent>>,
+    ) -> Result<Vec<(EventIndex, Event)>>;
+
+    async fn insert_order(&self, order: &Order) -> Result<(), InsertionError>;
+    async fn cancel_order(&self, order_uid: &OrderUid, now: DateTime<Utc>) -> Result<()>;
+    fn orders<'a>(&'a self, filter: &'a OrderFilter) -> BoxStream<'a, Result<Order>>;
+
+    fn trades<'a>(&'a self, filter: &'a TradeFilter) -> BoxStream<'a, Result<ModelTrade>>;
+
+    async fn save_fee_measurement(
+        &self,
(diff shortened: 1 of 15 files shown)


orderbook/src/database.rs
```diff
@@ -4,13 +4,46 @@ mod orders;
...
+    fn contract_to_db_events(
+        &self,
+        contract_events: Vec<EthContractEvent<ContractEvent>>,
+    ) -> Result<Vec<(EventIndex, Event)>>;
+
+    async fn insert_order(&self, order: &Order) -> Result<(), InsertionError>;
+    async fn cancel_order(&self, order_uid: &OrderUid, now: DateTime<Utc>) -> Result<()>;
+    fn orders<'a>(&'a self, filter: &'a OrderFilter) -> BoxStream<'a, Result<Order>>;
+
+    fn trades<'a>(&'a self, filter: &'a TradeFilter) -> BoxStream<'a, Result<ModelTrade>>;
```
bh2smith, line 44: If these particular order and trade related methods weren't directly on the database we could also implement this trait for `BalancerPoolStorage`.

bh2smith (approved):

orderbook/src/database.rs
vkgnosis: I am not completely sure what you mean. This trait is roughly the set of all operations we want do with the postgres db. So if we were going to store balancer related information in the db then we could add methods here or there could be a new trait for these balancer operations that is implemented by `Postgres` like our existing `MinFeeStoring`.

bh2smith: All the trait methods you have declared on `Database` (except the ones I have highlighted) here are also defined for `BalancerPool` which is also a kind of in-memory "Database". Perhaps it would be worth stripping these less generic trait methods from the base Database trait and making an `OrderbookDatabase` which implements database as well as the orderbook specific methods. This way, we could also use the Database trait for Balancer `PoolStorage`. 

not necessary at all, just an idea.

orderbook/src/database.rs
```diff
@@ -3,14 +3,69 @@ mod fees;
...
+
+    async fn save_fee_measurement(
```
fleupold, line 49: Could `Database: Send + Sync + MinFeeStoring` to avoid redefining these methods? (not chosing to do so might be related to Ben's suggestion of composing the `Database` trait from smaller sub-traits, e.g. an `EventStoring` trait)

orderbook/src/database.rs
```diff
@@ -76,14 +131,129 @@ impl Database {
     }
 }
 
+#[async_trait::async_trait]
+impl Database for Postgres {
```
fleupold, line 135: ~Why the extra level of indirection and not move the implementation directly here?~

I see now, because the methods are implemented in separate modules.

fleupold (approved):
Code looks good, just wondering if it would make sense to even further compose the Database trait from smaller traits (which are each implemented in their own modules)?

nlordell (approved):

vkgnosis:
Based on feedback I have split the big trait into several specialized traits. We now have
MinFeeStoring, EventUpdating, OrderStoring, TradeRetrieving.

orderbook/src/main.rs
```diff
@@ -159,8 +160,11 @@ async fn main() {
         None
     };
 
-    let event_updater =
-        EventUpdater::new(settlement_contract.clone(), database.clone(), sync_start);
+    let event_updater = EventUpdater::new(
+        settlement_contract.clone(),
+        database.as_ref().clone(),
```
bh2smith, line 165: Did something change on the updater input parameters or is the database struct just different now?

orderbook/src/event_updater.rs
```diff
@@ -1,63 +1,21 @@
...
-};
-use std::ops::RangeInclusive;
+use ethcontract::dyns::DynWeb3;
+use shared::{event_handling::EventHandler, impl_event_retrieving, maintenance::Maintaining};
 use tokio::sync::Mutex;
 
-pub struct EventUpdater(Mutex<EventHandler<DynWeb3, GPv2SettlementContract, Database>>);
-
-#[async_trait::async_trait]
-impl EventStoring<ContractEvent> for Database {
```
bh2smith, line 19: I can't seem to find where all this went.. Is it somehow inherited from an existing trait somewhere?

bh2smith (approved):
Looks awesome!

orderbook/src/event_updater.rs
vkgnosis: I moved this implementation into database/events.rs . This is also why in your comment above we now pass a `Postgres` instance directly instead of `Arc<Postgres>`.

PR #755, opened 2021-06-18
Implement BaselineSolvable for Balancer's WeightedPool

### LINKED ISSUE #586: Implement Balancer weighted pools

fedgiac:
Closes #586.

This PR implements `BaselineSolvable` for `WeightedPool`.
To my understanding, this represents the correct computation for every type of WeightedPool specialization except for minor details that cannot be captured with this abstractions (e.g., reverting if the amount received from a trade plus the current solver balance would overflow Balancer's internal limits).

I implemented `gas_cost` as a constant copying what is done for Uniswap. I believe I'd need a lot of details on the token implementations or node access to properly compute an estimate, so if we want to do this it should be done in another PR.

I also changed all weights to be Balancer fixed point values, since this is how they are represented in the contract, rather than performing the conversion immediately before computing the results I need.

### Test Plan
New unit tests.

### DIFF
--- a/shared/src/balancer/swap.rs
+++ b/shared/src/balancer/swap.rs
@@ -1,3 +1,258 @@
+use crate::balancer::pool_storage::{PoolTokenState, WeightedPool};
+use crate::baseline_solver::BaselineSolvable;
+use crate::conversions::u256_to_big_int;
+use error::Error;
+use ethcontract::{H160, U256};
+use fixed_point::Bfp;
+use num::{BigRational, CheckedDiv};
+use weighted_math::{calc_in_given_out, calc_out_given_in};
+
 mod error;
 pub mod fixed_point;
 mod weighted_math;
+
+const BALANCER_SWAP_GAS_COST: usize = 100_000;
+
+impl PoolTokenState {
+    /// Converts the stored balance into its internal representation as a
+    /// Balancer fixed point number.
+    fn upscaled_balance(&self) -> Option<Bfp> {
+        self.upscale(self.balance)
+    }
+
+    /// Scales the input token amount to the value that is used by the Balancer
+    /// contract to execute math operations.
+    fn upscale(&self, amount: U256) -> Option<Bfp> {
+        amount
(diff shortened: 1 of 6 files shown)


shared/src/balancer/swap.rs
```diff
@@ -1,3 +1,258 @@
...
 mod weighted_math;
+
+const BALANCER_SWAP_GAS_COST: usize = 100_000;
+
+impl PoolTokenState {
+    /// Converts the stored balance into its internal representation as a
+    /// Balancer fixed point number.
+    fn upscaled_balance(&self) -> Option<Bfp> {
+        self.upscale(self.balance)
+    }
```
bh2smith, line 21: So upscale!

shared/src/balancer/swap/weighted_math.rs
```diff
@@ -2,8 +2,6 @@
 //! smart contract. The original contract code can be found at:
 //! https://github.com/balancer-labs/balancer-v2-monorepo/blob/6c9e24e22d0c46cca6dd15861d3d33da61a60b98/pkg/core/contracts/pools/weighted/WeightedMath.sol
 
-#![allow(dead_code)]
-
```
bh2smith, line 6: Party time! 🥳

bh2smith (approved):
Looks really good. We are getting closer.

vkgnosis (approved):

PR #758, opened 2021-06-21
[Easy] Cleanup Balancer Module

bh2smith:
- Rename `PoolCacheMetrics` to `WeightedPoolCacheMetrics` to avoid possibility of confusing invalid imports 
- A few useful log statements in event handler

Note that the last item may be removed upon suggestion (since its not yet used and needs some rework).

### Test Plan
no logic changes.

### DIFF
--- a/shared/src/balancer/pool_cache.rs
+++ b/shared/src/balancer/pool_cache.rs
@@ -1,6 +1,5 @@
 use crate::balancer::event_handler::BalancerPoolRegistry;
 use crate::balancer::pool_storage::RegisteredWeightedPool;
-use crate::pool_cache::PoolCacheMetrics;
 use crate::pool_fetching::{handle_contract_error, MAX_BATCH_SIZE};
 use crate::{
     balancer::pool_storage::WeightedPool,
@@ -13,11 +12,16 @@ use ethcontract::batch::CallBatch;
 use ethcontract::errors::MethodError;
 use ethcontract::{BlockId, Bytes, H160, H256, U256};
 use std::{collections::HashSet, sync::Arc};
+use crate::recent_block_cache::CacheMetrics;
 
 pub struct PoolReserveFetcher {
     pool_registry: Arc<BalancerPoolRegistry>,
     vault: BalancerV2Vault,
     web3: Web3,
+}
+
+pub trait WeightedPoolCacheMetrics: Send + Sync {
+    fn pools_fetched(&self, cache_hits: usize, cache_misses: usize);
 }
 
 impl PoolReserveFetcher {
@@ -32,7 +36,7 @@ impl PoolReserveFetcher {
 }
 
 pub type BalancerPoolReserveCache =
-    RecentBlockCache<H256, WeightedPool, PoolReserveFetcher, Arc<dyn PoolCacheMetrics>>;
+    RecentBlockCache<H256, WeightedPool, PoolReserveFetcher, Arc<dyn WeightedPoolCacheMetrics>>;
 
 impl CacheKey<WeightedPool> for H256 {
     fn first_ord() -> Self {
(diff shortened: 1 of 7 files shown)


shared/src/balancer/pool_cache.rs
```diff
@@ -13,10 +14,6 @@ use ethcontract::errors::MethodError;
 use ethcontract::{BlockId, Bytes, H160, H256, U256};
 use std::{collections::HashSet, sync::Arc};
 
-pub trait PoolCacheMetrics: Send + Sync {
```
nlordell, line 16: The idea with this extra indirection is we can have separate metrics for Balancer vs Uniswap pool fetching. It might be worth leaving this in.

Alternatively we can just make the metrics implement `CacheMetrics` directly and avoid the extra level of indirection with the `PoolCacheMetrics` trait.

shared/src/balancer/pool_storage.rs
```diff
@@ -192,6 +219,11 @@ impl PoolStorage {
                     .insert(pool_id);
             }
         }
+        tracing::info!(
+            "inserted {} balancer pool creation events. Synced to block {}",
+            num_events,
+            self.last_event_block()
```
nlordell, line 225: Is this accurate? If we sync to block X but the last pool was created at some other block Y won't this log be misleading?

bh2smith: Yes... although it does say up till which block we have pools. If this is too misleading or confusing, I would rather just remove it.

shared/src/balancer/pool_cache.rs
bh2smith: I don't really have a preference here. I would lean toward avoid extra direction, but I feel like this would be an unrelated change to this PR.

shared/src/balancer/pool_storage.rs
nlordell: It says `Synced to block` which is not true, since the latest pool we have is not necessarily on the latest block we've synced.

shared/src/balancer/pool_cache.rs
nlordell: Let's wait and see what others think, but IMO it is related to this PR as we are removing one of the traits here, so we should either remove both `PoolCacheMetrics` in order to remove one level of indirection, or keep the extra indirection and allow separate metrics to be used.

vkgnosis: I feel that it is useful to have separate metrics for separate pool fetching backends. Not very important, just slightly in favor.

vkgnosis (approved):

shared/src/balancer/pool_cache.rs
bh2smith: The reason I removed this is because there is a bit of a name conflict if we use `PoolCacheMetrics` because its easy to import the wrong one. I could also get around this by putting it back and calling it `WeightedPoolCacheMetrics`. I will do that for now.

bh2smith: Cool. then I will keep them all separated, but rename this one to `WeightedPoolCacheMetrics` so that imports are not easily confused.

shared/src/balancer/pool_storage.rs
bh2smith: Yea I see the problem. This log has been removed.

shared/src/balancer/pool_cache.rs
```diff
@@ -3,7 +3,7 @@ use crate::balancer::pool_storage::RegisteredWeightedPool;
 use crate::pool_fetching::{handle_contract_error, MAX_BATCH_SIZE};
 use crate::{
     balancer::pool_storage::WeightedPool,
-    recent_block_cache::{Block, CacheFetching, CacheKey, CacheMetrics, RecentBlockCache},
+    recent_block_cache::{Block, CacheFetching, CacheKey, RecentBlockCache},
```
nlordell, line 6: nit: This is an unnecessary change - it looks like the `use` statement was just moved to a different line.

nlordell (approved):

shared/src/balancer/pool_cache.rs
bh2smith: Yea, I just went over all the import statements and consolidated them. So there are a few more changed imports but now they are all in the most compact form throughout the module.

PR #761, opened 2021-06-21
*New HTTP Solver Model incorporating Balancer Pools

### LINKED ISSUE #769: Prepare Balancer Pool Data for HTTP Solver
This task is broken down into several steps:

- [x] Remodel the `UniswapModel` for a more general `PoolModel` (#761)
- [x] Start passing Token Addresses as H160 to the solver instead of `tN` (#763) - Merged Into #761.
- [x] Rename `AmmOrder` and all instances of it for `ConstantProductOrder`(#770)
- [x] Introduce `WeightedProductOrder` and all analogous mappings similar to `ConstantProductOrder`
- [ ] Pass relevant details through to `Solver::prepare_model`

bh2smith:
Closes #769.

This PR is the accumulation of several smaller changes that result in a complete refactoring of the HTTP solver request and response format to accommodate Balancer weighted pool amm models.

This is the accumulation of

- use H160 token addresses instead of String (#763)
- Prepare Balancer Pool Data for HTTP Solver (#769)
- Weighted Pool Model with Liquidity (#781)
- Weighted Pool Liquidity Model (#775)
- Move H160Wrapper to shared (#787)
- update response model (#833) 
- UpdatedAmmModel to new format (#808) 

To align with the request type specification in https://github.com/gnosis/gp-v2-solver/issues/147#issuecomment-864878822 and in preparation for encoding Balancer Pools we change the UniswapModel (passed on to the HTTP solver) to be more flexible for more general pool types.

Essentially, we have changed from having keys `token0, token1, balance0, balance1` to `reserves` as part of the pool "parameters" like so.

```rust
#[derive(Debug, Serialize)]
pub struct AmmModel {
    #[serde(flatten)]
    pub parameters: AmmParameters,
    pub cost: CostModel,
    pub mandatory: bool,
}

#[derive(Debug, Serialize)]
#[serde(tag = "kind")]
pub enum AmmParameters {
    ConstantProduct(ConstantProductPoolParameters),
    WeightedProduct(WeightedProductPoolParameters),
}
```

where constant product parameters are a `HashMap<Token, Balance>` and weighted product pool parameters are similar, but whose hashmap values are another struct holding the balance and weight of that token.

The response data is also generalized to incorporate the fact that a single amm (pool) may be used multiple times in a settlement meaning that we had to use a vector of execution plans.

### Test Plan
Introduced several new test for json (de)serialization of the top level struct `BatchAuctionModel` and response model `SettledBatchAuctionModel`.

A local Integration test was performed as follows

1. Run the orderbook - make sure to have the orderbook DB running (cf services readme)
```sh
cargo run --bin orderbook -- --node-url <YOUR_NODE_URL>
```
2. FrontEnd: Change API URL to local host and run
```sh
REACT_APP_API_URL_STAGING_MAINNET=http://0.0.0.0:8080/api yarn start
```

3. Run Solver: Note that Uniswap must be included in `baseline-sources` to get a price estimate for balancer pools.
```sh
cargo run -p solver -- --baseline-sources BalancerV2,Uniswap --node-url <YOUR_NODE_URL> --solvers Mip --mip-solver-url <HTTP_SOLVER_URL> --base-tokens '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2'
```


solver/src/solver/http_solver/model.rs
```diff
@@ -26,14 +26,16 @@ pub struct OrderModel {
...
-    pub token1: String,
-    pub token2: String,
-    #[serde(with = "serde_with::rust::display_fromstr")]
-    pub balance1: u128,
-    #[serde(with = "serde_with::rust::display_fromstr")]
-    pub balance2: u128,
+pub struct PoolTokenData {
     #[serde(with = "serde_with::rust::display_fromstr")]
+    pub balance: u128,
+    pub weight: String,
```
nlordell, line 32: This shouldn't be `String` - its a 256-bit integer.

solver/src/solver/http_solver/model.rs
```diff
@@ -26,14 +26,16 @@ pub struct OrderModel {
...
+}
+
+#[derive(Debug, Serialize)]
+pub struct PoolModel {
+    pub pool_type: String,
```
nlordell, line 37: I think instead of using a `String` here we should have an `enum` that serializes to and from some string representation. This should be possible with `serde`.

solver/src/solver/http_solver/model.rs
```diff
@@ -26,14 +26,16 @@ pub struct OrderModel {
...
+    pub reserves: HashMap<String, PoolTokenData>,
```
nlordell, line 38: As a meta comment, I'm also not a fan of `String` here as well as it causes ambiguity: what is the reserves mapping key? Token name? Token address?

Instead we should use concrete types (like `H160` or some `TokenId` type) to avoid type erasure.

I see that this is done in other parts of the HTTP solver model, so this is something that can also be done in a separate PR.

solver/src/solver/http_solver/model.rs
bh2smith: I spoke with @marcovc and he agreed with string. However I can make it u128 like the balance and to the `#[serde(with = "serde_with::rust::display_fromstr")]` if this is more appropriate here.

solver/src/solver/http_solver/model.rs
bh2smith: I would much rather use H160 for all tokens. As long as it matches with the `tokens` field provided in the dictionary it should be fine on the solver side anyway.

solver/src/solver/http_solver/model.rs
bh2smith: Actually in the balancer module this is a `Bfp`. Is this something we can include here or should we convert `Bfp`to `u128` when sending it on?

solver/src/solver/http_solver/model.rs
bh2smith: I have learned that the String we are using for tokens is actually just their index in an array (which is much less information than H160). Are we sure we want to change this?

bh2smith: https://github.com/gnosis/gp-v2-services/pull/763

solver/src/solver/http_solver/model.rs
nlordell: So, we can use `Bpf`, but I don't know how easy it would be to support arbitrary precision decimal numbers (**not** double precision floating point numbers) on the solver side (cc @marcovc and @twalth3r). Alternatively, we can use their `as_uint256` to match their EVM representation - whichever is easier for the solver team.

bh2smith: I was under the impression that we had been using u128 everywhere else because we couldn't adapt U256 to the solver team's needs. Perhaps we should ask with @e00E, @twalth3r  and/or @marcovc if there was a particular reason for u128 as balance and if we could just start using `U256` here. Otherwise I would prefer to leave it as u128 and convert the `Bfp`to this (if possible).

nlordell: > for u128 as balance and if we could just start using U256 here

Uniswap stores balances in `u96`, so `u128` is the smallest integer capable of representing Uniswap balances.

marcovc: I'm not sure what's going on on the rust side, but we're getting big ints (as in anything > 2^64) as strings: because of json parser limitations.

nlordell: In that case, we should be able to use `U256` with the `u256_decimal` formatter.

bh2smith: So in that case, I would then opt for going with U256 in both the balance and weight field here. I hope this is approach is a mutual feeling.

vkgnosis (approved):

vkgnosis:
We have to coordinate merging this and updating the solver in staging btw.

solver/src/solver/http_solver/model.rs
```diff
@@ -26,14 +26,22 @@ pub struct OrderModel {
...
 #[derive(Debug, Serialize)]
-pub struct UniswapModel {
-    pub token1: String,
-    pub token2: String,
-    #[serde(with = "serde_with::rust::display_fromstr")]
-    pub balance1: u128,
-    #[serde(with = "serde_with::rust::display_fromstr")]
-    pub balance2: u128,
-    #[serde(with = "serde_with::rust::display_fromstr")]
+pub struct PoolTokenData {
```
nlordell, line 29: I think we should add a test in this file so that the JSON format is documented in code.

bh2smith: Yes good idea, Although I might prefer to add this in the following PR that changes the tokens to H160. 

Note that the solver team has agreed to this proposed change along with the Token addresses. See here:

https://github.com/gnosis/gp-v2-solver/issues/147#issuecomment-865689927

bh2smith:
@vkgnosis 

> We have to coordinate merging this and updating the solver in staging btw.


Yes, there is an ongoing discussion [here](https://github.com/gnosis/gp-v2-solver/issues/147#issuecomment-865689927)

solver/src/solver/http_solver/model.rs
```diff
@@ -26,44 +28,54 @@ pub struct OrderModel {
...
+    #[serde(with = "u256_decimal")]
+    pub balance: U256,
+    #[serde(with = "u256_decimal")]
+    pub weight: U256,
+}
+
+#[derive(Debug, Serialize)]
+pub enum PoolType {
+    UniswapV2,
+}
```
nlordell, line 41: I think we can use a tagged enum so that we can model different pools that have completely different parameters. While technically a Uniswap pool is a weighted pool with weights of `1.0` - we might not necessarily be able to find a model that is compatible with all of Uniswap pools, Balancer wieghted pools, and other kinds of AMMs (like Balancer stable pools for example).

Something along the lines of:
```rust
pub struct PoolModel {
    pub parameters: PoolParameters,
    pub cost: CostModel,
    pub mandatory: bool,
}

pub enum PoolParameters {
    UniswapV2(UniswapPoolParameters),
    BalancerWeighted(BalancerWeightedPoolParameters),
}

pub struct UniswapPoolParameters {
    pub reserves: HashMap<H160, U256>,
    pub fee: BigDecimal,
}

pub struct BalancerWeightedPoolParameters {
    pub reserves: HashMap<H160, BalancerWeightedPoolTokenData>,
    pub fee: BigDecimal,
}
```

solver/src/solver/http_solver.rs
```diff
@@ -143,73 +137,87 @@ impl HttpSolver {
             .collect()
     }
 
-    fn map_orders_for_solver(&self, orders: Vec<LimitOrder>) -> HashMap<String, LimitOrder> {
+    fn map_orders_for_solver(&self, orders: Vec<LimitOrder>) -> HashMap<usize, LimitOrder> {
         orders
             .into_iter()
             .enumerate()
-            .map(|(index, order)| (index.to_string(), order))
+            .map(|(index, order)| (index, order))
```
nlordell, line 144: just noticed this `map` is unnecessary.
```suggestion
```

solver/src/solver/http_solver.rs
```diff
@@ -143,73 +137,87 @@ impl HttpSolver {
...
+        result
     }
 
-    fn map_amms_for_solver(&self, orders: Vec<AmmOrder>) -> HashMap<String, AmmOrder> {
+    fn map_amms_for_solver(&self, orders: Vec<AmmOrder>) -> HashMap<usize, AmmOrder> {
         orders
             .into_iter()
             .enumerate()
-            .map(|(index, amm)| (index.to_string(), amm))
+            .map(|(index, amm)| (index, amm))
```
nlordell, line 182:
```suggestion
```

solver/src/solver/http_solver/model.rs
bh2smith: This all seems like a good idea, but we need to keep the solver team in the loop on this so they know what to expect. @marcovc and @twalth3r

nlordell: This was the changes we discussed though right? That we would used a "tagged union" in order to represent pools so we can have different parameters/structure for different kinds of pools that don't fit into the weighted pool model.

nlordell: This will output JSON looking like:
```json
[
  {
    "kind": "UniswapV2",
    "reserves": {
      "0x0102030405060708091011121314151617181920": "100000"
    },
    "fee": "0.003",
    "cost": {
      "amount": "1337",
      "token": "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee"
    },
    "mandatory": false
  },
  {
    "kind": "BalancerWeighted",
    "reserves": {
      "0x0102030405060708091011121314151617181920": {
        "balance": "100000",
        "weight": "0.5"
      }
    },
    "fee": "0.003",
    "cost": {
      "amount": "1337",
      "token": "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee"
    },
    "mandatory": false
  }
]
```

bh2smith: I got the impression that you were suggesting to introduce a "parameters" field with contained the pools inner information.

bh2smith: I now see the "flatten"

bh2smith:
I have left this in draft state to avoid accidental merge since this requires coordinate with the solver team. However it is ready for review.

solver/src/solver/http_solver/model.rs
```diff
@@ -26,44 +28,56 @@ pub struct OrderModel {
...
 
+#[derive(Debug, Serialize)]
+#[serde(tag = "pool_type")]
+pub enum PoolParameters {
+    ConstantProduct(ConstantProductPoolParameters),
+}
+
+#[derive(Debug, Serialize)]
+pub struct ConstantProductPoolParameters {
+    pub reserves: HashMap<H160, U256Wrapper>,
```
nlordell, line 46: I'm not a fan of this wrapper type seeing as we already use `serde_with` and have a `SerializeAs` implementation for `U256`. I suggest using something like this which worked when I tried locally:

```rust
#[serde_as]
#[derive(Debug, Serialize)]
pub struct ConstantProductPoolParameters {
    pub reserves: HashMap<H160, U256>,
    pub fee: BigDecimal,
}
```

Serializing this to JSON output:
```json
{
  "reserves": {
    "0x0102030405060708091011121314151617181920": "1000000000000000000"
  },
  "fee": "0.003"
}
```

bh2smith: I am also happy to introduce this. I will also make a separate PR to remove H160Wrapper in the same way.

nlordell: > I will also make a separate PR to remove H160Wrapper in the same way.

Nice!

vkgnosis: Nice, I didn't realize this approach worked for hash maps too.

nlordell: For the brave and curious, there is a nice comprehensive guide here: https://docs.rs/serde_with/1.9.4/serde_with/guide/serde_as/index.html

solver/src/solver/http_solver/model.rs
```diff
@@ -26,44 +28,56 @@ pub struct OrderModel {
...
+    pub fee: BigDecimal,
```
nlordell, line 47: You mentioned today that this can be moved out to the common parameters in `PoolModel`.

bh2smith: Yes, although I have done it in the following PR #775. The one that actually introduces a reason for moving it.

twalth3r:
Just curious: I see there's a lot of `pool` as well as `amm` in the code, there's even a field `pool_type` in the AMM JSON struct that we receive. Is there a reason for that? In our solver prototyping code, we went with `pools` everywhere, but @bh2smith mentioned there's a preference for you to go with `amms`. Should we use one or the other exclusively?

solver/src/solver/http_solver/model.rs
```diff
@@ -1,20 +1,23 @@
...
 use std::collections::HashMap;
 
 #[derive(Debug, Default, Serialize)]
 pub struct BatchAuctionModel {
-    pub tokens: HashMap<String, TokenInfoModel>,
-    pub orders: HashMap<String, OrderModel>,
-    pub uniswaps: HashMap<String, UniswapModel>,
+    pub tokens: HashMap<H160, TokenInfoModel>,
+    pub orders: HashMap<usize, OrderModel>,
+    pub amms: HashMap<usize, PoolModel>,
```
bh2smith, line 13: @twalth3r  - this is the place where "uniswaps" was changed to amms. Each member of this HashMap is a form of pool. We could use the name "pools" here instead although, like I mentioned, at some point this may include 0x and then the term pools would no longer be valid. We will probably first introduce a `StablePool` Model as well.

However given that this hasmap consists only of `PoolModels` and they all have a `pool_type` field I think maybe its best to go with `pools` instead of `amms`. Is there any other part of the code where you see an inconsistency?

bh2smith:
@twalth3r 
> Should we use one or the other exclusively?

I assume you are referring to the `BatchAuctionModel` struct 

```rust
#[derive(Debug, Default, Serialize)]
pub struct BatchAuctionModel {
    pub tokens: HashMap<H160, TokenInfoModel>,
    pub orders: HashMap<usize, OrderModel>,
    pub amms: HashMap<usize, PoolModel>,
    pub metadata: Option<MetadataModel>,
}
```

in which case I am beginning to think that, yes, we can go with pools here instead of amms. Is there somewhere else (structurally) in the code that is inconsistent or just a few variable names that should be changed?

We should be aware that 0x integration could once again change this, but maybe those types of liquidity will remain independent from the HashMap of PoolModels.

solver/src/solver/http_solver/model.rs
twalth3r: I was under the impression that you are not using the term `pool` at all in your code and was surprised to see `PoolModel` etc (was probably expecting `AmmModel`). But I guess you have some structural reasoning for when you use AMM and when Pool.

twalth3r: In the JSON, if we call the field `amms`, then we should probably go with `amm_type` inside, or conversely for `pools`. Or maybe just `type`!?

twalth3r:
Another question - should we also discuss the return format here as well? The solver computes pool updates, so I see two options:

1. add `updates` field:
```
    "0": {
      "pool_type": "ConstantProduct",
      "reserves": {
        "0x000000000000000000000000000000000000a866": "200",
        "0x0000000000000000000000000000000000000539": "100"
      },
      "updates": {
        "0x000000000000000000000000000000000000a866": "-20",
        "0x0000000000000000000000000000000000000539": "10"
      },
```

2. change the `reserves` field:
```
    "0": {
      "pool_type": "ConstantProduct",
      "reserves": {
        "0x000000000000000000000000000000000000a866": {
            "balance": "200",
            "update": "-20"
        "0x0000000000000000000000000000000000000539": {
            "balance": 100",
            "update": "-10"
      }
```

I think 2. looks better - which do you prefer @bh2smith , @marcovc ? Or do you have any other ideas?

solver/src/solver/http_solver/model.rs
bh2smith: I think maybe we should just go with pools since I think the term PoolModels will be long lived. When the 0x integration we can cross that bridge later.

marcovc:
Since we don't care about the original reserves in the solution, and we'd like to make the files small, what about:
```
 "0": {
      "pool_type": "ConstantProduct",
      "updates": {
        "0x000000000000000000000000000000000000a866": "-20",
        "0x0000000000000000000000000000000000000539": "10"
      },
```
?

nlordell:
Personally, I prefer just returning the `updates`.

solver/src/solver/http_solver/model.rs
nlordell: I still think `pool.pool_type` is a bit redundant and slightly prefer `pool.type` or `pool.kind` if you want to get Rust-y (because `type` is a reserved keyword).

twalth3r:
Right - just wondering if there's some use-case for which we'd the pool balances together with the updates. When investigating a solution, to me it's helpful to have all information in one place. File size is not really a problem I guess, since we already only include the pools that have been traded (which should be few). Maybe I can just go with option 1. then and the driver just doesn't care about the `reserves`? That'd also be similar as what we do with orders, where we also pass the `{buy-/sell-amount}` along with the solution info.

nlordell:
> When investigating a solution, to me it's helpful to have all information in one place

In that case, feel free to keep it in the JSON output. I don't feel strongly either way.

bh2smith:
> Right - just wondering if there's some use-case for which we'd want the pool balances together with the updates. When investigating a solution, to me it's helpful to have all information in one place. File size is not really a problem I guess, since we already only include the pools that have been traded (which should be few). Maybe I can just go with option 1. then and the driver just doesn't care about the `reserves`? That'd also be similar to what we do with orders, where we also pass the `{buy-/sell-amount}` along with the solution info.

I will have to think about this a bit more because I haven't actually started changing anything involving the return values yet (since everything here only includes Uniswap refactor and nothing new). So up until this point the previous returned values have been handled. Thanks for reminding me! Will introduce this soon, but as of now I am off till Monday. Feel free to decid on something collectively and I will just do it that way!

solver/src/solver/http_solver/model.rs
bh2smith: Kind it is!

bh2smith:
Ok all that have been involved here (cc @twalth3r , @marcovc , @nlordell and maybe @fedgiac ). I have decided to go with Amm everywhere instead of Pool (So we now have `AmmModel` struct with `AmmParameters`). The tagged enum has been changed from "pool_type" to "kind". This actually works nicely because the AmmParameters enum allows us to specify if its a pool when it is and whatever else when its not.

```
pub enum AmmParameters {
    ConstantProduct(ConstantProductPoolParameters),
    WeightedProduct(WeightedProductPoolParameters),
    StablePool(StablePoolParameters),
    ZeroX(ZeroXOrder),
}
```

bh2smith:
> Since we don't care about the original reserves in the solution, and we'd like to make the files small, what about:
> 
> ```
>  "0": {
>       "pool_type": "ConstantProduct",
>       "updates": {
>         "0x000000000000000000000000000000000000a866": "-20",
>         "0x0000000000000000000000000000000000000539": "10"
>       },
> ```


From what I recall, this struct should have an execution plan as well although maybe this is something we include as well. Also, since we join on the known outgoing data we shouldn't even need to pass the `pool_type` (aka "kind"). Does anyone think we should keep this around just for good measure or would updates be sufficient here?

twalth3r:
> From what I recall, this struct should have an execution plan as well although maybe this is something we include as well.

Yes! The struct looks like this:
```
            "exec_plan": {
                "sequence": 0,
                "position": 0
            }
```

> Also, since we join on the known outgoing data we shouldn't even need to pass the pool_type (aka "kind"). Does anyone think we should keep this around just for good measure or would updates be sufficient here?

As mentioned, I'd be in favor of keeping the input fields, since it provides a good view on the solution in case we have to investigate, and space is not an issue. For example, the `kind` field may be useful for parsing the file when it comes to visualizations at some point. But you can of course everything that's not needed on your end to process a solution! Would that work for everyone? If yes, a complete example would look like this:
```
    "pools": {
        "23": {
            "kind": "ConstantProduct",
            "reserves": {
                "0x5d3a536e4d6dbd6114cc1ead35777bab948e3643": "441875522512065",
                "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2": "38844257255210538349"
            },
            "fee": "0.003",
            "cost": {
                "token": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
                "amount": "1672962666603536"
            },
            "updates": {
                "0x5d3a536e4d6dbd6114cc1ead35777bab948e3643": "502579848701",
                "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2": "-43998200058683932"
            },
            "exec_plan": {
                "sequence": 0,
                "position": 0
            }
        }
    }
```
(where `pools` might now be changed to `amms` again, if I understood correctly!?)

nlordell:
> Would that work for everyone?

Yes, we can just ignore fields we don't care about on the driver side. I definitely agree that space is not an issue, and improving your ability to debug solutions is quite important.

marcovc:
Wait, can we still do the execution plan like this when the pools are N-ary? For example, if we have two pools, one with tokens {A,B,C,D} and one with tokens {E,F}, and the execution plan is A->B, some orders, E->F, some orders, C->D, then the first pool will appear in two different positions in the execution plan. Plus we need a way to specify which tokens were traded in each position. Right?

bh2smith:
> Wait, can we still do the execution plan like this when the pools are N-ary? For example, if we have two pools, one with tokens {A,B,C,D} and one with tokens {E,F}, and the execution plan is A->B, some orders, E->F, some orders, C->D, then the first pool will appear in two different positions in the execution plan. Plus we need a way to specify which tokens were traded in each position. Right?

It is possible that the same pool could be used multiple times in a solution given that is has more than just two tokens. So I would think that we need some way to represent this. Either an array of execution plans or maybe an array of the inner fields.

nlordell:
> and the execution plan is A->B, some orders, E->F, some orders, C->D, then the first pool will appear in two different positions in the execution plan.

I think with that specific example, it should be fine. AFAIU for weighted pools, prices only depend on the reserves of the tokens being traded. An issue would arise if you had a {A,B,C} pool with trades A->C and B->C for example, where both trades affect the C reserves, and therefore affect prices used for each-other.

marcovc:
Ok, let me try another example then: One pool has tokens {A,B,C,D} and another tokens {B,C}. The execution plain is pool1:A ->

marcovc:
> An issue would arise if you had a {A,B,C} pool with trades A->C and B->C for example, where both trades affect the C reserves, and therefore affect prices used for each-other.

We will never issue such a solution (we explicitely forbid it in the solver because of the reason you mentioned)

bh2smith:
Shouldn't it be fine if the prices change an we use the same pool for the same token multiple times? We aren't guaranteeing uniform cleaning prices for amms anyway.

solver/src/solver/http_solver/model.rs
```diff
@@ -105,3 +119,119 @@ pub struct ExecutionPlanCoordinatesModel {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use maplit::hashmap;
+    use serde_json::json;
+
+    #[tokio::test]
+    async fn model_serialization() {
```
nlordell, line 130:
```suggestion
    #[test]
    fn model_serialization() {
```

nlordell:
> pool1:A -> pool1:B -> pool2:B -> pool2:C -> pool1:C -> pool1:D

Ok I see the issue. The updates only captures the final updates after all swaps and not one for each swap. Yeah, I imagine that would require a change to the output format as well.

marcovc:
> Shouldn't it be fine if the prices change an we use the same pool for the same token multiple times?

It would be fine, we forbid it in the solver just because we don't know how to allow it without complicating things a lot :)

marcovc:
@bh2smith Regarding the execution plan, Tom and I suggest we keep it at it is currently in this PR, and create another one just for handling its new format. The reason is that it needs a more radical change when multiple token pools (>2) are involved.

solver/src/solver/http_solver/settlement.rs
```diff
@@ -235,29 +232,27 @@ mod tests {
...
                 position: 0,
             }),
         };
         let settled = SettledBatchAuctionModel {
-            orders: hashmap! { "lo0".to_string() => executed_order },
-            uniswaps: hashmap! { "amm0".to_string() => updated_uniswap },
-            ref_token: "t0".to_string(),
-            prices: hashmap! { "t0".to_string() => Price(10.0), "t1".to_string() => Price(11.0) },
+            orders: hashmap! { 0 => executed_order },
+            amms: hashmap! { 0 => updated_uniswap },
```
bh2smith, line 250: Note to self: Going to have to double check/ be careful when building this map that we don't overwrite the CPOs with the WPOs since we are using indices.

nlordell (approved):

PR #762, opened 2021-06-21
Differentiate Balancer pools by factory address

nlordell:
This PR differentiates Balancer pools by factory address. This allows us to put the pools into their correct storage on cold start. Additionally, we can also create alerts when we load pools weighted from factory contracts that aren't supported by the event registry to catch when new pool factory contracts get added early.

### Test Plan

Existing unit tests continue to pass. Additionally run the `cargo test -p shared -- --ignored --nocapture balancer_subgraph_query` ignored test and check output for and see pools are split into two factory contracts, one for N-dimensional weighted pools, and the other for 2 token weighted pools.


vkgnosis (approved):

bh2smith (approved):
Looks good, just wondering about the unwrap or default which could potentially be taken care of internally here rather than let this unknown leak out into other places.

nlordell:
> just wondering about the unwrap or default which could potentially be taken care of internally here rather than let this unknown leak out into other places

So this is already done inside the method on the private data types and doesn't leak outside of the `graph_api` module.

shared/src/balancer/graph_api.rs
```diff
@@ -37,41 +39,47 @@ impl BalancerSubgraphClient {
...
             let has_next_page = page.len() == QUERY_PAGE_SIZE;
+            if let Some(last_pool) = page.last() {
+                last_id = last_pool.id;
+            }
 
-            results.reserve(page.len());
             for pool in page {
-                results.push(pool.into_registered(block_number)?);
+                pools_by_factory
+                    .entry(pool.factory.unwrap_or_default())
```
bh2smith, line 71: but default here will fallback on the zero address... How will we know what to do with this if we get it?

nlordell: My reasoning here is that when we get all the pools, any "factory buckets" that aren't one of the factories we are expecing (WeightedPoolFactory or Weighted2TokenPoolFactory) will trigger an allert - including the 0x00...00 address.

PR #763, opened 2021-06-21
*PoolModel - use H160 token addresses instead of String

bh2smith:
Partial fulfillment of #769 

Addressing the comment brought up [here](https://github.com/gnosis/gp-v2-services/pull/761#discussion_r655303506) that we shouldn't be using String for token identifiers.

There is a lot more code to change here though since the entire solver infrastructure relies on a `HashMap<String, H160>`

### DIFF
--- a/solver/src/solver/http_solver.rs
+++ b/solver/src/solver/http_solver.rs
@@ -119,20 +119,18 @@ impl HttpSolver {
 
     async fn token_models(
         &self,
-        tokens: &HashMap<String, H160>,
         token_infos: &HashMap<H160, TokenInfo>,
         price_estimates: &HashMap<H160, Result<BigRational, PriceEstimationError>>,
-    ) -> HashMap<String, TokenInfoModel> {
-        tokens
+    ) -> HashMap<H160, TokenInfoModel> {
+        token_infos
             .iter()
-            .map(|(index, address)| {
-                let token_info = token_infos[address];
+            .map(|(address, token_info)| {
                 let external_price = price_estimates[address]
                     .as_ref()
                     .ok()
                     .and_then(|price| price.to_f64());
                 (
-                    index.clone(),
+                    *address,
                     TokenInfoModel {
                         decimals: token_info.decimals.map(|d| d as u32),
                         external_price,
@@ -161,8 +159,8 @@ impl HttpSolver {
         for (index, order) in orders {
             let order_fee = self.order_fee(&order)?;
             let order = OrderModel {
-                sell_token: self.token_to_string(&order.sell_token),
-                buy_token: self.token_to_string(&order.buy_token),
+                sell_token: order.sell_token,
+                buy_token: order.buy_token,
                 sell_amount: order.sell_amount,
                 buy_amount: order.buy_amount,
                 allow_partial_fill: order.partially_fillable,
@@ -199,14 +197,14 @@ impl HttpSolver {
             .map(|(index, amm)| {
                 let mut reserves = HashMap::new();
                 reserves.insert(
-                    self.token_to_string(&amm.tokens.get().0),
+                    amm.tokens.get().0,
                     PoolTokenData {
                         balance: amm.reserves.0,
                         weight: 500_000_000_000_000_000u128,
                     },
                 );
                 reserves.insert(
-                    self.token_to_string(&amm.tokens.get().1),
+                    amm.tokens.get().1,
                     PoolTokenData {
                         balance: amm.reserves.1,
                         weight: 500_000_000_000_000_000u128,
@@ -260,9 +258,7 @@ impl HttpSolver {
         );
         let limit_orders = self.map_orders_for_solver(orders.0);
         let amm_orders = self.map_amms_for_solver(orders.1);
-        let token_models = self
-            .token_models(&tokens, &token_infos, &price_estimates)
-            .await;
+        let token_models = self.token_models(&token_infos, &price_estimates).await;
         let order_models = self.order_models(&limit_orders, gas_price).await?;
         let uniswap_models = self.amm_models(&amm_orders, gas_price).await;
         let model = BatchAuctionModel {
--- a/solver/src/solver/http_solver.rs
+++ b/solver/src/solver/http_solver.rs
@@ -92,15 +92,7 @@ impl HttpSolver {
         }
     }
 
-    // Solver api requires specifying token as strings. We use the address as a string for now.
-    // Later we could use a more meaningful name like the token symbol but we have to ensure
-    // uniqueness.
-    fn token_to_string(&self, token: &H160) -> String {
-        // Token names must start with a letter.
-        format!("t{:x}", token)
-    }
-
-    fn map_tokens_for_solver(&self, orders: &[Liquidity]) -> HashMap<String, H160> {
+    fn map_tokens_for_solver(&self, orders: &[Liquidity]) -> Vec<H160> {
         orders
             .iter()
             .flat_map(|liquidity| match liquidity {
@@ -113,7 +105,6 @@ impl HttpSolver {
             })
             .collect::<HashSet<_>>()
             .into_iter()
-            .map(|token| (self.token_to_string(&token), token))
             .collect()
     }
 
@@ -167,11 +158,11 @@ impl HttpSolver {
                 is_sell_order: matches!(order.kind, OrderKind::Sell),
                 fee: FeeModel {
                     amount: order_fee,
-                    token: self.token_to_string(&order.sell_token),
+                    token: order.sell_token,
                 },
                 cost: CostModel {
                     amount: order_cost,
-                    token: self.token_to_string(&self.native_token),
+                    token: self.native_token,
                 },
             };
             result.insert(index.clone(), order);
@@ -215,7 +206,7 @@ impl HttpSolver {
                     fee: *amm.fee.numer() as f64 / *amm.fee.denom() as f64,
                     cost: CostModel {
                         amount: uniswap_cost,
-                        token: self.token_to_string(&self.native_token),
+                        token: self.native_token,
                     },
                     mandatory: false,
                     reserves,
@@ -230,22 +221,16 @@ impl HttpSolver {
         liquidity: Vec<Liquidity>,
         gas_price: f64,
     ) -> Result<(BatchAuctionModel, SettlementContext)> {
-        // To send an instance to the solver we need to identify tokens and orders through strings.
-        // In order to map back and forth we store the original tokens, orders and the models for
-        // via the same mapping.
         let tokens = self.map_tokens_for_solver(liquidity.as_slice());
 
-        let addresses: Vec<H160> = tokens.values().into_iter().cloned().collect();
-
         let (token_infos, price_estimates) = join!(
-            self.token_info_fetcher
-                .get_token_infos(addresses.as_slice()),
+            self.token_info_fetcher.get_token_infos(tokens.as_slice()),
             self.price_estimator
-                .estimate_prices(addresses.as_slice(), addresses[0])
+                .estimate_prices(tokens.as_slice(), tokens[0])
         );
 
         let price_estimates: HashMap<H160, Result<BigRational, _>> =
-            addresses.iter().cloned().zip(price_estimates).collect();
+            tokens.iter().cloned().zip(price_estimates).collect();
 
         let mut orders = split_liquidity(liquidity);
 
@@ -270,7 +255,6 @@ impl HttpSolver {
             }),
         };
         let context = SettlementContext {
-            tokens,
             limit_orders,
             amm_orders,
         };
--- a/solver/src/solver/http_solver.rs
+++ b/solver/src/solver/http_solver.rs
@@ -424,13 +424,16 @@ mod tests {
         let url = std::env::var("GP_V2_OPTIMIZER_URL")
             .unwrap_or_else(|_| "http://localhost:8000".to_string());
 
+        let buy_token = H160::from_low_u64_be(1337);
+        let sell_token = H160::from_low_u64_be(43110);
+
         let mut mock_token_info_fetcher = MockTokenInfoFetching::new();
         mock_token_info_fetcher
             .expect_get_token_infos()
             .return_once(move |_| {
                 hashmap! {
-                    H160::zero() => TokenInfo { decimals: Some(18)},
-                    H160::from_low_u64_be(1) => TokenInfo { decimals: Some(18)},
+                    buy_token => TokenInfo { decimals: Some(18)},
+                    sell_token => TokenInfo { decimals: Some(18)},
                 }
             });
         let mock_token_info_fetcher: Arc<dyn TokenInfoFetching> = Arc::new(mock_token_info_fetcher);
@@ -457,8 +460,8 @@ mod tests {
         let base = |x: u128| x * 10u128.pow(18);
         let orders = vec![
             Liquidity::Limit(LimitOrder {
-                buy_token: H160::zero(),
-                sell_token: H160::from_low_u64_be(1),
+                buy_token,
+                sell_token,
                 buy_amount: base(1).into(),
                 sell_amount: base(2).into(),
                 kind: OrderKind::Sell,
@@ -468,7 +471,7 @@ mod tests {
                 id: "0".to_string(),
             }),
             Liquidity::Amm(AmmOrder {
-                tokens: TokenPair::new(H160::zero(), H160::from_low_u64_be(1)).unwrap(),
+                tokens: TokenPair::new(buy_token, sell_token).unwrap(),
                 reserves: (base(100), base(100)),
                 fee: Ratio::new(0, 1),
                 settlement_handling: CapturingSettlementHandler::arc(),
--- a/solver/src/solver/http_solver.rs
+++ b/solver/src/solver/http_solver.rs
@@ -113,7 +113,7 @@ impl HttpSolver {
             .collect()
     }
 
-    async fn token_models(
+    fn token_models(
         &self,
         token_infos: &HashMap<H160, TokenInfo>,
         price_estimates: &HashMap<H160, Result<BigRational, PriceEstimationError>>,
@@ -145,12 +145,12 @@ impl HttpSolver {
             .collect()
     }
 
-    async fn order_models(
+    fn order_models(
         &self,
         orders: &HashMap<usize, LimitOrder>,
         gas_price: f64,
-    ) -> Result<HashMap<usize, OrderModel>> {
-        let order_cost = self.order_cost(gas_price).await;
+    ) -> HashMap<usize, OrderModel> {
+        let order_cost = self.order_cost(gas_price);
         let mut result: HashMap<usize, OrderModel> = HashMap::new();
         for (index, order) in orders {
             let order_fee = self.order_fee(&order);
@@ -172,7 +172,7 @@ impl HttpSolver {
             };
             result.insert(*index, order);
         }
-        Ok(result)
+        result
     }
 
     fn map_amms_for_solver(&self, orders: Vec<AmmOrder>) -> HashMap<usize, AmmOrder> {
@@ -183,12 +183,12 @@ impl HttpSolver {
             .collect()
     }
 
-    async fn amm_models(
+    fn amm_models(
         &self,
         amms: &HashMap<usize, AmmOrder>,
         gas_price: f64,
     ) -> HashMap<usize, PoolModel> {
-        let uniswap_cost = self.uniswap_cost(gas_price).await;
+        let uniswap_cost = self.uniswap_cost(gas_price);
         amms.iter()
             .map(|(index, amm)| {
                 let mut reserves = HashMap::new();
@@ -249,9 +249,9 @@ impl HttpSolver {
         );
         let limit_orders = self.map_orders_for_solver(orders.0);
         let amm_orders = self.map_amms_for_solver(orders.1);
-        let token_models = self.token_models(&token_infos, &price_estimates).await;
-        let order_models = self.order_models(&limit_orders, gas_price).await?;
-        let amm_models = self.amm_models(&amm_orders, gas_price).await;
+        let token_models = self.token_models(&token_infos, &price_estimates);
+        let order_models = self.order_models(&limit_orders, gas_price);
+        let amm_models = self.amm_models(&amm_orders, gas_price);
         let model = BatchAuctionModel {
             tokens: token_models,
             orders: order_models,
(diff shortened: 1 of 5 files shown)


solver/src/solver/http_solver/model.rs
```diff
@@ -1,20 +1,21 @@
...
     pub uniswaps: HashMap<String, PoolModel>,
     pub metadata: Option<MetadataModel>,
 }
 
 #[derive(Debug, Serialize)]
 pub struct OrderModel {
-    pub sell_token: String,
-    pub buy_token: String,
+    pub sell_token: H160,
+    pub buy_token: H160,
```
vkgnosis, line 18: This is likely incorrect because it changes how this is formatted in the json. token_to_string says that the string has to start with with a letter so we need a custom serialization function that does this here too.

bh2smith: But there isn't actually a good reason why the string has to start with the letter `t` is there?

bh2smith: Anyway, this is just here in case the solver team decided they would like to start accepting tokens by address instead of some arbitrary index denoted as `t_i`. we don't have to merge it, but its ready to go and addresses the totally valid point brought up in #761.

solver/src/solver/http_solver.rs
```diff
@@ -113,26 +105,23 @@ impl HttpSolver {
             })
             .collect::<HashSet<_>>()
```
bh2smith, line 106: I wonder if some of this additional collecting could also be removed. Not entirely sure what the chaining business above is all about.

solver/src/solver/http_solver/model.rs
nlordell: We can have an opaque `TokenId(String)` type if we want to stay 100% backwards compatible. The comment in the other PR is that `String` is just textual data and doesn't give any indication as to what it represents:
- `42.to_string()`
- `"asdf"`
- `token_to_string(token)`

Are all strings, but have vastly different meaning. My point in the other PR is that we could use more expressive types to clear what the data represents.

From the other PR:
> I have learned that the String we are using for tokens is actually just their index in an array

This is a strong argument for changing the types, as from looking at the type definition you weren't able to see what the expected data was. In fact, it would be super easily to accidentally insert `token_address.to_string()` instead of `token_index.to_string()` in some code and generate a bad instance without having the compiler tell you that it was incorrect.

solver/src/solver/http_solver.rs
nlordell: `chain`-ing two iterators just causes the second to iterate after the first one is done. Its analogous to `concat` for arrays.

bh2smith: Sure but what is it that we are even trying to concatenate here. Shouldn't these be token addresses? Why would anyone ever concatenate them?

nlordell: From the code, it appears that its enumerating over all sell and buy tokens for limit orders as well as pool tokens for AMM pools, and then collecting them into a `HashSet` to generate a list of unique token addresses that can be given to the solver.

solver/src/solver/http_solver/model.rs
bh2smith: Cool. So I see now that we have a choice to make. 

`TokenId(String)` where we continue to send tokens identified as `tN` or continue with the refactoring as currently found in this PR.

My argument for the latter is that we are already making breaking changes on the solver side so this would be the most appropriate time to introduce the more natural token identifier as H160.

solver/src/solver/http_solver/model.rs
```diff
@@ -57,20 +58,20 @@ pub struct TokenInfoModel {
...
+    pub token: H160,
 }
 
 #[derive(Debug, Deserialize)]
 pub struct SettledBatchAuctionModel {
     pub orders: HashMap<String, ExecutedOrderModel>,
     pub uniswaps: HashMap<String, UpdatedUniswapModel>,
-    pub ref_token: String,
+    pub ref_token: H160,
     pub prices: HashMap<String, Price>,
```
nlordell, line 75: Shouldn't this also be:
```suggestion
    pub prices: HashMap<H160, Price>,
```

solver/src/solver/http_solver/model.rs
```diff
@@ -57,20 +58,20 @@ pub struct TokenInfoModel {
...
 pub struct FeeModel {
     pub amount: u128,
-    pub token: String,
+    pub token: H160,
 }
 
 #[derive(Debug, Deserialize)]
 pub struct SettledBatchAuctionModel {
     pub orders: HashMap<String, ExecutedOrderModel>,
     pub uniswaps: HashMap<String, UpdatedUniswapModel>,
```
nlordell, line 73: Order IDs and AMM IDs should probably not be strings either right?

bh2smith: They are just indices. I would prefer to leave this as it is not terribly harmful and these "indices" are essentially meaningless to us. The solver team wants these indices as strings. If you really want to change it, I would propose using usize since if you look at how these maps are made it is really just enumerate and use the index of the enumeration as the key here.

solver/src/solver/http_solver/model.rs
bh2smith: I am not sure what you are looking at, but this is H160 already in my code.

bh2smith: Introduced already in 99023b3

nlordell: That does not seem to be the case:
```
$ git show 99023b3:solver/src/solver/http_solver/model.rs | grep 'prices:'
    pub prices: HashMap<String, Price>,
```

solver/src/solver/http_solver/model.rs
nlordell: > They are just indices

My issue is that the type `String` gives absolutely no indication as to what is being used for the key. For example, for the orders, a UID makes perfect sense as a key, and for pools, the pool address also makes perfect sense for a key. The issue is that there is an implicit format for the key that could be better expressed by choosing a more appropriate type. For example:
```rust
let mut map = HashMap::new();
map.insert(42usize, true);
println!("{}", serde_json::to_string(&map).unwrap()) // {"42":true}
```

This means that using `HashMap<usize, _>` whould output the exact same JSON while keeping a more concrete key type.

Other options that we can explore are using the short order UID (like in the explorer) as well as the AMM address which might make the instances slightly more debug-able.

solver/src/solver/http_solver/model.rs
```diff
@@ -41,7 +42,7 @@ pub enum PoolType {
 #[derive(Debug, Serialize)]
 pub struct PoolModel {
     pub pool_type: PoolType,
-    pub reserves: HashMap<String, PoolTokenData>,
+    pub reserves: HashMap<H160, PoolTokenData>,
     pub fee: f64,
```
nlordell, line 46: I'm not sure we should use a 64-bit floating point number here because of precision issues.For the Uniswap fee example:
```rust
println!("{:.64}", 0.3); // 0.2999999999999999888977697537484345957636833190917968750000000000
println!("{:.64}", 0.003) // 0.0030000000000000000624500451351650553988292813301086425781250000
```

Maybe we should use arbitrary precision decimals (`BigDecimal`) so that it serializes as a string?

solver/src/solver/http_solver/model.rs
```diff
@@ -57,20 +58,20 @@ pub struct TokenInfoModel {
 #[derive(Debug, Serialize)]
 pub struct CostModel {
     pub amount: u128,
```
nlordell, line 60: This can run into precision problems since a fair share of parsers use IEEE double precision floating point numbers for JSON numbers. This means that for values larger than `2^53`, there will be loss of precision. This should probably use `display_fromstr` formatting.

nlordell (commented):
Added some more comments about the HTTP solver JSON format. It might make sense to make this PR a more general "HTTP solver format cleanup" PR.

solver/src/solver/http_solver/model.rs
bh2smith: Well I don't know what to tell you, but you are commenting on a line of code that has already been changed to your suggestion.


Screenshot 2021-06-22 at 11 23 33

bh2smith: Screenshot 2021-06-22 at 11 25 08

solver/src/solver/http_solver/model.rs
bh2smith: I think we are going to have a hard time getting the solver team to agree to all of this, but it is certainly worth a try. @marcovc, @twalth3r how would you feel about not using indices as keys for the orders and amms? I think the short orderUid is not a great idea because of potential collision. If the complete orderId is too much I would prefer to leave it as integer string indices and change our code internally from `String` to `usize`

solver/src/solver/http_solver/model.rs
bh2smith: This is a great idea, but the only problem is that 

```
the trait `Serialize` is not implemented for `BigDecimal`
```

solver/src/solver/http_solver/model.rs
bh2smith: Ok, for now I have changed all the outgoing dictionary keys from String to usize. We are still accepting incoming settlement information as String and parsing back to usize where appropriate for comparison. I hope this can be the happy medium we seek for having good structures on one side, but not too bulk on the other.

solver/src/solver/http_solver/model.rs
```diff
@@ -57,20 +58,20 @@ pub struct TokenInfoModel {
...
-    pub token: String,
+    pub token: H160,
 }
 
 #[derive(Debug, Serialize)]
 pub struct FeeModel {
     pub amount: u128,
```
nlordell, line 70: We can run into precision issues with this as it will encode the amount as a number which most parsers interpret as a IEEE double precision floating point number which would only be accurate to `2^53`. This should probably use `display_fromstr` formatting.

solver/src/solver/http_solver/model.rs
nlordell: AFAIK `HashMap<usize, _>` keys are always strings, and shouldn't need manual parsing.

https://github.com/akubera/bigdecimal-rs/blob/fc797c302f4947e429cc331febc66fe3c7badc5e/Cargo.toml#L16

solver/src/solver/http_solver/model.rs
nlordell: Note that there is a `serde` feature to add serialization support:

https://github.com/akubera/bigdecimal-rs/blob/fc797c302f4947e429cc331febc66fe3c7badc5e/Cargo.toml#L16

nlordell: Note that there is a `serde` feature on the crate to add serialization support:

```rust
let parsed = serde_json::from_str::<HashMap<usize, bool>>(r#"
    {
        "42": true
    }
"#);
println!("{:?}", parsed);
```

solver/src/solver/http_solver/model.rs
nlordell: Note that there is a `serde` feature on the crate to add serialization support:

```rust
let parsed = serde_json::from_str::<HashMap<usize, bool>>(r#"
    {
        "42": true
    }
"#);
println!("{:?}", parsed);
```

solver/src/solver/http_solver.rs
```diff
@@ -18,6 +19,7 @@ use shared::{
     price_estimate::{PriceEstimating, PriceEstimationError},
     token_info::{TokenInfo, TokenInfoFetching},
 };
+use std::ops::Div;
```
nlordell, line 22: This shouldn't be needed.

solver/src/solver/http_solver.rs
```diff
@@ -144,86 +135,87 @@ impl HttpSolver {
...
                     PoolTokenData {
                         balance: U256::from(amm.reserves.1),
                         weight: U256::from(500_000_000_000_000_000u128),
                     },
                 );
+
                 let uniswap = PoolModel {
                     pool_type: PoolType::UniswapV2,
-                    fee: *amm.fee.numer() as f64 / *amm.fee.denom() as f64,
+                    fee: BigDecimal::from(*amm.fee.numer()).div(BigDecimal::from(*amm.fee.denom())),
```
nlordell, line 210:
```suggestion
                    fee: BigDecimal::from(*amm.fee.numer()) / (BigDecimal::from(*amm.fee.denom())),
```

solver/src/solver/http_solver/model.rs
nlordell: In your second screenshot, line 75 is:
```
    pub prices: HashMap<String, Price>,
```

and I am suggesting it be changed to
```
    pub prices: HashMap<H160, Price>,
```

solver/src/solver/http_solver/model.rs
```diff
@@ -56,21 +59,23 @@ pub struct TokenInfoModel {
...
+    #[serde(with = "serde_with::rust::display_fromstr")]
     pub amount: u128,
-    pub token: String,
+    pub token: H160,
 }
 
 #[derive(Debug, Deserialize)]
 pub struct SettledBatchAuctionModel {
     pub orders: HashMap<String, ExecutedOrderModel>,
     pub uniswaps: HashMap<String, UpdatedUniswapModel>,
```
nlordell, line 77:
```suggestion
    pub orders: HashMap<usize, ExecutedOrderModel>,
    pub uniswaps: HashMap<usize, UpdatedUniswapModel>,
```

solver/src/solver/http_solver/model.rs
bh2smith: Or right. silly me. Actually I am not so sure about that. Since this is the `SettledBatchAuctionModel` that is returned from the solver. However, I guess if we can interpret the ref_token as H160, then we should also be able to parse the prices key.

solver/src/solver/http_solver/model.rs
bh2smith: Thats fine with me. In that case I will change all the incoming information to also be usize. Just wasn't sure about that.

nlordell: > I think we are going to have a hard time getting the solver team to agree to all of this

My point being that using `usize` here instead of `String` has no effect on the output JSON but gives us stricter typing on the Rust side. In other words, it would not require a change from on the HTTP solver side.

solver/src/solver/http_solver/model.rs
bh2smith: So I am guessing from the context that `optional = true` is our ticket to having BigDecimal serialization, that this is the preferred approach and my solution is essentially a hack.

solver/src/solver/http_solver.rs
bh2smith: My IDE said it was, but I will remove it.

solver/src/solver/http_solver/model.rs
bh2smith: This was left as is because I assumed incoming data should be left as String, but I will change it.

solver/src/solver/http_solver.rs
nlordell: It was needed because you used `a.div(b)` instead of `a / b`, but see suggestion :point_down:

solver/src/solver/http_solver/model.rs
nlordell: It should be fine. The keys should get deserialized from their string values, saving the extra work of manually deserializing.

solver/src/solver/http_solver/model.rs
bh2smith: Actually, I think the `Optional<BigDecimal>` wasn't the problem here (if that what the optional flag you are referring to is supposed to do). The problem is with `struct { field: BigDecimal }`. I will try your suggestion anyway, but I suspect this isn't a work around for my use of `#[serde(with = "serde_with::rust::display_fromstr")]` here.

bh2smith: Awesome!

nlordell: So [cargo has the concept of optional features](https://doc.rust-lang.org/cargo/reference/features.html) to conditionally include additional code and functionality. This means it should be possible to add the `bigdecimal` dependency in a way that it includes `serde` serialization support. Specifically in the `Cargo.toml`:
```toml
bigdecimal = { version = "0.2", features = ["serde"] }
```

solver/src/solver/http_solver/model.rs
```diff
@@ -56,21 +58,23 @@ pub struct TokenInfoModel {
...
+    #[serde(with = "serde_with::rust::display_fromstr")]
     pub amount: u128,
-    pub token: String,
+    pub token: H160,
 }
 
 #[derive(Debug, Serialize)]
 pub struct FeeModel {
+    #[serde(with = "serde_with::rust::display_fromstr")]
     pub amount: u128,
```
nlordell, line 69:
```suggestion
    #[serde(with = "u256_decimal")]
    pub weight: U256,
```

In theory order fees can be greater than `u128` (you can actually achieve this quite easily by having a token with extremely low value where the fee amount would be more than `2^128`).

Also note that this change does not affect the output JSON in any way - it just allows encoding a larger fee from the Rust side.

solver/src/solver/http_solver/settlement.rs
```diff
@@ -15,9 +16,9 @@ use std::{
 // To send an instance to the solver we need to identify tokens and orders through strings. This
 // struct combines the created model and a mapping of those identifiers to their original value.
 pub struct SettlementContext {
-    pub tokens: HashMap<String, H160>,
-    pub limit_orders: HashMap<String, LimitOrder>,
-    pub amm_orders: HashMap<String, AmmOrder>,
+    // pub tokens: Vec<H160>,
```
nlordell: This field can be removed right?

bh2smith: I belive it has been.

solver/src/solver/http_solver/settlement.rs
```diff
@@ -153,19 +153,16 @@ fn match_prepared_and_settled_amms(
...
     // Remove the indirection over the token string index from the solver prices.
     let solver_prices: HashMap<H160, Price> = solver_prices
         .into_iter()
-        .map(|(index, price)| {
-            let token = prepared_tokens
-                .get(&index)
-                .ok_or_else(|| anyhow!("invalid token {}", index))?;
-            Ok((*token, price))
+        .map(|(address_string, price)| {
+            let token = H160::from_str(&address_string).expect("token string not parsable.");
```
nlordell: I don't think we should panic in case of failing to parse the result from the solver. This could make a bug in the HTTP solver crash the driver.

bh2smith: This is also no longer relevant as we use H160 for this now.

solver/src/solver/http_solver/model.rs
bh2smith: I suppose we will want to change this pattern in all places where it occurs.

solver/src/solver/http_solver.rs
```diff
@@ -93,15 +97,7 @@ impl HttpSolver {
...
-    // Solver api requires specifying token as strings. We use the address as a string for now.
-    // Later we could use a more meaningful name like the token symbol but we have to ensure
-    // uniqueness.
-    fn token_to_string(&self, token: &H160) -> String {
-        // Token names must start with a letter.
-        format!("t{:x}", token)
-    }
-
-    fn map_tokens_for_solver(&self, orders: &[Liquidity]) -> HashMap<String, H160> {
+    fn map_tokens_for_solver(&self, orders: &[Liquidity]) -> Vec<H160> {
```
nlordell, line 100: nit: Instead of collecting into a `HashSet` to then collect into a `Vec`, can't this just return a `HashSet` right away?

solver/src/solver/http_solver.rs
```diff
@@ -114,28 +110,25 @@ impl HttpSolver {
             })
             .collect::<HashSet<_>>()
             .into_iter()
-            .map(|token| (self.token_to_string(&token), token))
             .collect()
```
nlordell, line 113: That :point_up: would make these 3 lines no longer necessary.

solver/src/solver/http_solver.rs
```diff
@@ -325,19 +310,17 @@ impl HttpSolver {
             .with_context(|| format!("failed to decode response json, {}", context()))
     }
 
-    async fn order_cost(&self, gas_price: f64) -> u128 {
-        gas_price as u128 * GAS_PER_ORDER
+    async fn order_cost(&self, gas_price: f64) -> U256 {
```
nlordell, line 313: These don't have to be `async`

solver/src/solver/http_solver.rs
```diff
@@ -325,19 +310,17 @@ impl HttpSolver {
...
+        U256::from_f64_lossy(gas_price) * *GAS_PER_ORDER
     }
 
-    async fn uniswap_cost(&self, gas_price: f64) -> u128 {
-        gas_price as u128 * GAS_PER_UNISWAP
+    async fn uniswap_cost(&self, gas_price: f64) -> U256 {
```
nlordell, line 317: Neither does this one.

solver/src/solver/http_solver.rs
```diff
@@ -144,86 +137,87 @@ impl HttpSolver {
...
+    fn map_amms_for_solver(&self, orders: Vec<AmmOrder>) -> HashMap<usize, AmmOrder> {
         orders
             .into_iter()
             .enumerate()
-            .map(|(index, amm)| (index.to_string(), amm))
+            .map(|(index, amm)| (index, amm))
             .collect()
     }
 
     async fn amm_models(
```
nlordell, line 186: If `uniswap_cost` is no longer `async` then I think this method doesn't have to be `async` either.

solver/src/solver/http_solver.rs
```diff
@@ -144,86 +137,87 @@ impl HttpSolver {
...
+    fn map_orders_for_solver(&self, orders: Vec<LimitOrder>) -> HashMap<usize, LimitOrder> {
         orders
             .into_iter()
             .enumerate()
-            .map(|(index, order)| (index.to_string(), order))
+            .map(|(index, order)| (index, order))
             .collect()
     }
 
     async fn order_models(
```
nlordell, line 148: If `order_cost` is no longer `async` then this method doesn't have to be `async` either.

bh2smith: It also doesn't need to return a result anymore either.

nlordell (approved):
Looks good. Just some methods that are `async` that I don't think they need to be.

solver/src/solver/http_solver.rs
bh2smith: In fact it looks like `prepare_model` no longer needs to be async or anything called from in it.

PR #771, opened 2021-06-22
[Paraswap] Get src/dest amount from priceRoute instead of bestRoute

fleupold:
For more complex execution plans best Route may contain multiple elements each of which containing a fraction of the total src/dest Amount. In the current API parsing code I used the first item of bestRoute which only works for "simple" execution plans.

Instead we need to use the src/dest amount from the more top level priceRoute field.

This addresses some of the alerts we saw in prod yesterday (`TransactionBuilderQuery result parsing failed: {"error":"Destination Amount Mismatch"}`) and was also the reason why the Paraswap solver wasn't used for our internal buyback program: https://gnosisinc.slack.com/archives/CC3C5SMB5/p1624316904013100

### Test Plan
Adjusting e2e to trade a really large amount (test not passing on current main)

### DIFF
--- a/solver/src/solver/paraswap_solver/api.rs
+++ b/solver/src/solver/paraswap_solver/api.rs
@@ -128,12 +128,6 @@ impl<'de> Deserialize<'de> for PriceResponse {
         #[derive(Deserialize)]
         #[serde(rename_all = "camelCase")]
         struct PriceRoute {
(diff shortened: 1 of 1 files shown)


nlordell (approved):

josojo (approved):
Nice!

bh2smith (approved):

PR #772, opened 2021-06-22
[Hotfix] 1Inch Limit Price logic

fleupold:
`!swap.to_token_amount >= order.buy_amount` is not the same as `!(swap.to_token_amount >= order.buy_amount)` - negation applies first to swap.to_token_amount which for an U256 negates the bits. The negated value is then compared to order.buy amount.

This is causing us to misinterpret pretty much all 1Inch price estimates as not fulfilling the limit price and is part of the reason why 1Inch wasn't used to settle our GNO buyback trade: https://gnosisinc.slack.com/archives/CC3C5SMB5/p1624316904013100

### Test Plan
For now I just added a very simple unit test for the satisfies_limit_price function. Ideally, the 1Inch solver would have tests similar to the Paraswap solver (with a mocked API and contract) that would allow to test this logic inside the unit. This change is more involved though and it's important we fix this issue now.


josojo (approved):
Nice! 

I was just about to investigate the same line. The bitwise operator is tricky

nlordell (approved):

bh2smith (approved):
Uh oh! All good now. I like that the comparison has been put into its own function now.

PR #774, opened 2021-06-22
Add component for Balancer pool initialization

nlordell:
This PR adds a new abstraction for Balancer pool initialization. This will eventually be used by the pool fetcher in order to initialize its storage and avoid indexing all Balancer pool creation events to speed up initialization.

### Test Plan

Added new unit tests to check factory filtering logic.


shared/src/balancer/pool_init.rs
```diff
@@ -0,0 +1,256 @@
...
+};
+use anyhow::{anyhow, Result};
+use contracts::{BalancerV2WeightedPool2TokensFactory, BalancerV2WeightedPoolFactory};
+use ethcontract::{Artifact, H160};
+
+#[derive(Debug, Default, PartialEq)]
+pub struct BalancerRegisteredPools {
+    pub weighted_pools: Vec<RegisteredWeightedPool>,
+    pub weighted_2token_pools: Vec<RegisteredWeightedPool>,
+}
```
bh2smith, line 18: I assume this is only a temporary struct to be parsed and converted into `PoolStorage` when this is hooked up to the rest.

shared/src/balancer/pool_init.rs
```diff
@@ -0,0 +1,256 @@
...
+/// This can be used to index all pools from events instead of relying on the
+/// Balancer subgraph for example.
+pub struct EmptyPoolInitializer;
+
+#[async_trait::async_trait]
+impl PoolInitializing for EmptyPoolInitializer {
+    async fn initialize_pools(&self) -> Result<BalancerRegisteredPools> {
+        Ok(Default::default())
+    }
+}
```
bh2smith, line 36: Doesn't it seem like overkill to make traits for this if we know we will only ever need it once? If it were going to be mocked in the test framework, then I see the point, but that doesn't seem to be the case.

bh2smith (approved):
Looks good, although I feel like some of the abstraction may be overkill for a single use component. Of course abstraction is fine and this is very good, perhaps for portability in the future.

shared/src/balancer/pool_init.rs
nlordell: This allows mocking and testing of how it interacts with the `BalancerPoolFetcher` once it gets hooked up. Additionally, it allows us to also have a `FilePoolInitializer` and `DatabasePoolInitializer` if we ever want to move away from the graph without having to refactor the calling code.

shared/src/balancer/pool_init.rs
```diff
@@ -0,0 +1,256 @@
...
+        // support (xDAI, Rinkeby and Mainnet) they are.
+        .ok_or_else(|| {
+            anyhow!(
+                "missing {} deployment for {}",
+                artifact.contract_name,
+                chain_id,
+            )
+        })?
+        .address)
+}
```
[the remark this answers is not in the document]
nlordell: Hmm, from what I can tell, there is the generated method for getting a contract instance given a `Web3` instance:
```rust
        let weighted_pool_factory = BalancerV2WeightedPoolFactory::deployed(&web3).await?;
```

and a method for getting a block number from deployment information of a contract:
```
async fn get_deployment_block(
    deployment_info: Option<DeploymentInformation>,
    web3: &Web3,
) -> Option<u64>;
```

AFAICT there is no helper method for getting a contract address from a chain id in `event_handler` module.

vkgnosis (approved):

shared/src/balancer/pool_init.rs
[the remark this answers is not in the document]
bh2smith: Right yes I was remembering her deployment block. I guess these are different so my comment doesn't matter.

PR #775, opened 2021-06-23
Weighted Pool Liquidity Model

bh2smith:
Towards #769 

We introduce the relevant WeightedPool information and include it into the Pool Model Enum. This will require a Refactoring of `ConstantProductOrder` (formerly known as `AmmOrder`) by using the same enum variant type that we use for `PoolModel`. Nice thing is that we can reuse the `PoolParamters` in both places. 

### Test Plan
New test for `BigDecimal From Bfp`. Not sure how high it can go... @fedgiac do we have a Bfp::MAX?


shared/src/balancer/swap/fixed_point.rs
```diff
@@ -36,6 +36,12 @@ impl From<usize> for Bfp {
     }
 }
 
+impl From<Bfp> for usize {
+    fn from(num: Bfp) -> Self {
+        num.as_uint256().checked_div(*ONE_18).unwrap().as_usize()
+    }
+}
```
nlordell, line 43: This conversion feels a bit wrong, as you will lose all fractional bits from the fixed point number.

bh2smith: I would love to do this, but I don't have any ideas in mind for converting it any differently than I have here....

nlordell: The issue with the code here is that a `Bfp` of `0.5` would become a `BigDecimal` of value `0.0` as you are truncating the fractional bits by dividing by `1.0` (correct me if I'm wrong @fedgiac).

fedgiac: Yep, this would truncate any decimal. Why is `usize` from `Bfp` needed anyway, considering that all weights are between 0 and 1? (Now I read the PR.)
The BigDecimal could be computed as numerator: `num.as_uint256()` and denominator: `*ONE_18`.

solver/src/solver/http_solver.rs
```diff
@@ -171,40 +174,75 @@ impl HttpSolver {
...
+                let reserves = amm
+                    .reserves
+                    .iter()
+                    .map(|(token, state)| {
+                        (
+                            *token,
+                            PoolTokenData {
+                                balance: state.balance,
+                                weight: BigDecimal::from_usize(usize::from(state.weight))
+                                    .expect("failed to convert Bfp to BigDecimal"),
```
fedgiac, line 224: I think the function `u256_to_big_int` might help here, I had the same issue [in my PR](https://github.com/gnosis/gp-v2-services/pull/755/files#diff-38988351762381d171eb4c77de97eaed54429cff6fe6fa178b166f50e9fc243fR117-R126).

shared/src/balancer/swap/fixed_point.rs
bh2smith: Yea. I'll do this right MEOW!

solver/src/solver/http_solver.rs
bh2smith: Cool, although this will now become part of another PR.

shared/src/balancer/swap/fixed_point.rs
```diff
@@ -295,4 +306,18 @@ mod tests {
...
+        assert_eq!(BigDecimal::from(Bfp::zero()), BigDecimal::zero());
+        assert_eq!(BigDecimal::from(Bfp::one()), BigDecimal::from(1));
+        assert_eq!(BigDecimal::from(Bfp::from(500)), BigDecimal::from(500));
+        assert_eq!(
+            BigDecimal::from(Bfp::from_wei(U256::MAX)),
+            BigDecimal::from_str(
+                "115792089237316195423570985008687907853269984665640564039457.584007913129639935"
+            )
+            .unwrap()
+        );
```
fedgiac, line 321: I'd like to also see a test that explicitly consider the case `1 > value > 0` to make sure that no weird rounding is happening. The last tested value is technically enough, but I think it's nicer to make the test case more explicit. You could use something like:
```
        assert_eq!(
            "0.4".parse::<Bfp>().unwrap(),
            "0.4".parse::<BigDecimal>().unwrap()
        );
```

solver/src/liquidity.rs
```diff
@@ -104,6 +107,12 @@ impl Default for LimitOrder {
     }
 }
 
+// pub struct AmmOrder<T> {
+//     pub reserves: HashMap<H160, T>,
+//     pub fee: Ratio<u32>,
+//     pub settlement_handling: Arc<dyn SettlementHandling<Self>>,
+// }
+
```
fedgiac, line 115: Is this a template for amm orders? If so, an explanation comment would be nice, on the other hand I don't see the need.

fedgiac (approved):
Looks good to me! I like how `BigInt`s are handled, even if I still don't understand why we need so many overlapping types.

If you want to make the max for `Bfp` explicit you could add `pub const MAX: Self = Self(U256::MAX);` to its implementation and access it with `Bfp::MAX`.

solver/src/liquidity.rs
bh2smith: This is an idea I was tinkering with that we could go with instead of having both weighted and constant product orders. I would actually like to go with this at some point but haven't had the time to put it together. The idea is that eventually we will use other forms of liquidity like 0x orders which aren't pools but rather just considered another form of Amm. This is closely related to what is happening in #769 and the discussion I've been having with @twalth3r there. From the perspective of the solver team only having to become compatible with one breaking change this would accommodate both. Unfortunately it relies heavily on whether we go with the term "amms" or "pools" in the "PoolModel" or possibly "AmmModel" and abandon the term pools, in favour of amms, in a few places. 

Same goes for uniswap v3 - if we are going to call them pools. 

If others have any further insight I would be happy to continue this discussion. Overall it boils down to whether or not the solver team will have to adapt to the keys of the dictionary they are receiving although I suppose they would have to adapt to these other forms of liquidity regardless.

bh2smith: I will cut it out of here for now and put it back when I have a chance.

shared/Cargo.toml
```diff
@@ -10,6 +10,7 @@ anyhow = "1.0"
 assert_approx_eq = "1.1"
 async-trait = "0.1"
 atty = "0.2"
+bigdecimal = "0.2"
```
vkgnosis, line 13: Why do we need bigdecimal here? I see that we define some conversions but where do we actually use the BigDecimal? The reason I am asking is that BigDecimal was originally only used when interfacing with postgres because that is the only big integer type that Postgres supports. I feel that in other code we should be using https://docs.rs/num-bigint/0.4.0/num_bigint/index.html big integers because that library is in a better state overall.

vkgnosis (approved):

shared/Cargo.toml
nlordell: We use `BigDecimal` to model weights, as they tend to have fractional components. Specifically, a 2-token Balancer pool follows the following equation:
```
(reserve0 - out0)^w0 * (reserve1 + in1)^w1 = K
```

Where
```
w0 + w1 = 1.0
```

For balancer, they are represented as `uint256` where the actual value it represents is `uint256(value) / 1e18`. That is, a `uint256` value of `8e17` represents `0.8` in the above equation.

Another option would be to use `num::Ratio` instead of `BigNumber`. I was looking into the `bigdecimal` issues and PRs and it looks abandoned by the developer.

nlordell: The only other issue, is that we need `BigDecimal` for serializing input to the solver as it expects decimal notation for the weights. But we can just do the conversion locally to that module.

vkgnosis: Yes that's why I didn't want to use it in other places. If you need fractional numbers then num_rational::BigRational would be better.

nlordell (commented):
Looks great! Maybe we should consider @vkgnosis's suggestion of using `BigRational` instead of `BigDecimal` internally for the baseline solver.

bh2smith:
> Looks great! Maybe we should consider @vkgnosis's suggestion of using `BigRational` instead of `BigDecimal` internally for the baseline solver.

Yea this is fine as long as we wont see a substantial difference when passing BigRational as scale to the HTTP solverinstead of BigDecimal.

nlordell:
> as scale to the HTTP solverinstead of BigDecimal

I think we still need to pass it as `BigDecimal` to the HTTP solver so that it gets serialized as a decimal string. AFAIK, `BigRational` serializes as a fraction (`1337/42`).

shared/Cargo.toml
bh2smith: BIgDecimal will be used in a follow up PR where we send WeightedPool's token weights (Bfp) to the HTTP solver via JSON. Which they expect as a decimal string.

Note that with the conversions we have internally it was not possible to use anything but `use num_bigint::BigInt as NumBigInt;`. I guess when you say other code, you mean the code that is not using `num_bigint`, in which case that is something I would be happy to fix in a separate PR.

vkgnosis:
Looks like there is no way to format a BigRational in base 10 without first converting to float (which defeats the point). That is annoying. Probably still better to use BigRational and like Nick said only convert to BigDecimal for the purpose of serialization for the solver.

bh2smith:
> Looks like there is no way to format a BigRational in base 10 without first converting to float (which defeats the point). That is annoying. Probably still better to use BigRational and like Nick said only convert to BigDecimal for the purpose of serialization for the solver.

Cool stuff. Will do.

solver/src/liquidity.rs
```diff
@@ -18,6 +20,7 @@ pub mod uniswap;
 pub enum Liquidity {
     Limit(LimitOrder),
     ConstantProduct(ConstantProductOrder),
+    WeightedProduct(WeightedProductOrder),
```
fleupold, line 23: Would a 50:50 balancer pool still be modelled as a constant product order? In this case maybe we don't need the WeightedPool type until the solver can actually handle e.g. 80:20 pools?

bh2smith: At this moment it was the plan to be putting all balancer pools into the weighted pool section since their settlement encoding is different.

nlordell: AFAIU the current plan is to pass weighted pools to the solver as is, and internally they will treat it like a Uniswap pool if the weights are the same.

bh2smith: Perhaps to indicate this better we should actually use the term balancer and uniswap pools on the inner types.

nlordell: Also, this is the abstract solver liquidity type that goes to all solvers, including the baseline solver which should be able to handle 80:20 pools (since it works by computing output amounts, for which we already have the math).

> since their settlement encoding is different

This should be handled by the `SettlementHandling` implementation, which in theory could also be made to work with Balancer pools.

That being said, I do believe it is correct here to distinguish between weighted and constant product pools.

nlordell: > Perhaps to indicate this better we should actually use the term balancer and uniswap pools on the inner types

I think the names are correct. The interaction encoding is done by the `SettlementHandling` implementation which already has [names based on the protocols they are encoding for](https://github.com/gnosis/gp-v2-services/blob/main/solver/src/liquidity/uniswap.rs).

bh2smith: That's all fine and dandy, but I am no longer sure if you are in favour of Felix's suggestion to assign 50:50 balancer pools to constant product orders. It is my impression that this was not the plan.

nlordell: I am for leaving it as `WeightedPool`.

fleupold: Just want to highlight that the fact that their settlement logic is different, doesn't mean they need to be separate liquidity sub-types. Each liquidity type has a dynamic "settlement_handling" which encodes how this specific instance gets translated into interactions. Two different `ConstantProductOrder` orders (e.g Uniswap, Sushiswap) can have different settlement_handling (one settles on Uni the other one on Sushi). Similarly, both EOA as well as 0x orders can be modelled with the LimitOrder subtype (despite very different settlement logic).

The concern about rounding makes sense. I think midterm it would be nice if this enum is not overlapping and since a `WeightedProductOrder` includes `ConstantProductOrder` they could be modelled as one subtype.

fleupold: Maybe @fedgiac knows if the Uniswap v2 math is equivalent to the Balancer v2 math (for 50:50 pools) or if there can be rounding discrepancies

bh2smith: Given that balancer uses their own floating point implementation on the contract level, I think we can be quite confident that, although the formulas may boil down to the same thing in overlapping circumstances, the arithmetic is bound to differ by even a marginal little bit on some corner case. Would also be nice to hear from Federico though!

bh2smith: One place to look for a difference is a uniswap pool with a token having more than 18 decimals. Balancer does not support tokens with this property but this would be a place to look for a clear difference in valuations. Another place that might be more relevant is just a two token pool where one token has 18 decimals and the other has 0 which would be possible in both types of pools. We could set the pools to have the exact same reserves and compute the in given out or vice versa for both. See what comes out.

bh2smith:
This PR now uses `BigRational` instead of `BigDecimal` as suggested and all signs of BigDecimal have been removed from this crate.

solver/src/liquidity.rs
```diff
@@ -140,6 +157,14 @@ impl Settleable for ConstantProductOrder {
     }
 }
 
+impl Settleable for WeightedProductOrder {
+    type Execution = AmmOrderExecution;
```
nlordell, line 161: Nice! I thought we would need a separate `Execution` type for weighted pools, but we don't. Pleasant surprise :smile:.

nlordell (approved):
Looks good!

shared/src/balancer/swap/fixed_point.rs
```diff
@@ -295,4 +309,29 @@ mod tests {
...
+            BigRational::from(Bfp::from(500)),
+            BigRational::new(BigInt::from(500), BigInt::one())
+        );
+        assert_eq!(
+            BigRational::from(Bfp::from_wei(U256::MAX)),
+            BigRational::new(
+                BigInt::from_str(&"115792089237316195423570985008687907853269984665640564039457584007913129639935").unwrap(),
+                BigInt::from(1_000_000_000_000_000_000u64)
+            )
+        );
```
nlordell, line 327: What a completionist!

solver/src/liquidity.rs
fleupold: If this is the case, then I'd question if the separation into ConstantProduct and WeightedProduct order makes sense. If we really need a different liquidity type per protocol (because of rounding details in the concrete implementations) then we should probably call the Liquidity types explicitly after their protocol. If however the math can be approximated by one another then the abstraction into types of functions (Limit, WeightedPool, ArithmeticMeanPool) makes sesne.

bh2smith: I am currently experimenting with an `AmmOrder<T>` with a Liquidity enum that looks like this:

```rust
pub struct AmmOrder<T> {
    pub reserves: HashMap<H160, T>,
    pub fee: Ratio<u32>,
    pub settlement_handling: Arc<dyn SettlementHandling<Self>>,
}
```

```rust
pub enum Liquidity {
    Limit(LimitOrder),
    Uniswap(AmmOrder<u128>),
    Balancer(AmmOrder<PoolTokenState>),
}
```

Is this something that looks a bit like what you are thinking.

I actually think this might work out very well, but does require some changes.

bh2smith: Unfortunately this is goingo to be quite an invasive change of structure given how much we rely on the `ConstantProductOrder` struct as it currently is. I don't belive it is worth the abstraction.

PR #777, opened 2021-06-23
Use pool initialization to populate pool storage

nlordell:
This PR uses the new `PoolInitializing` component in order to initially populate the pool storage on cold starts.

### Test Plan

Added unit tests to make sure the initial pools get put into storage correctly. Unfortunately, testing `BalancerPoolRegistry::new` was not really possible given the tight coupling to `Web3`. Suggestions on how to make that better are welcome, but I don't see anything immediately obvious without a larger refactor.


fleupold (approved):
LGTM, probably not necessary to test `BalancerPoolRegistry::new` because you would really just be testing `pool_initializer.initialize_pools()`

vkgnosis (approved):

shared/src/balancer/event_handler.rs
```diff
@@ -121,19 +131,6 @@ impl BalancerPoolRegistry {
     }
 }
 
-async fn get_deployment_block(
-    deployment_info: Option<DeploymentInformation>,
-    web3: &Web3,
-) -> Option<u64> {
```
bh2smith, line 127: Sad to see this go ☹️

bh2smith (approved):
Flippin' incredible!

PR #778, opened 2021-06-23
Make paraswap slippage configurable

fleupold:
May fix an alert we have seen somewhat frequently in prod:

2021-06-23T12:18:13.603Z ERROR solver::solver::single_order_solver: Inner solver error: TransactionBuilderQuery result parsing failed: {"error":"Destination Amount Mismatch"}
Caused by:
   missing field from at line 1 column 39

My assumption is that slippage is changing the destination amount too much, so that the order is no longer fillable. However, the [docs for the transaction builder query](https://developers.paraswap.network/api/build-parameters-for-transaction-builder-only) specifically state that the `destAmount` can be adjusted with slippage. My strategy would be to try setting slippage to 0 and if we see too many failed settlement we can increase it back to a value that we feel comfortable with.

In that case we might want to start treating the error above as "benign" and not alert on it.

### Test Plan
Added a unittest to verify slippage computation logic.

### DIFF
--- a/solver/src/main.rs
+++ b/solver/src/main.rs
@@ -148,6 +148,14 @@ struct Arguments {
         parse(try_from_str = shared::arguments::wei_from_gwei)
     )]
     gas_price_cap: f64,
+
+    /// The maximum gas price the solver is willing to pay in a settlement
+    #[structopt(
+        long,
+        env,
+        default_value = "0",
+    )]
+    paraswap_slippage_bps: usize,
 }
 
 #[tokio::main]
@@ -277,6 +285,7 @@ async fn main() {
         args.min_order_size_one_inch,
         args.disabled_one_inch_protocols,
         account.address(),
+        args.paraswap_slippage_bps,
     )
     .expect("failure creating solvers");
     let liquidity_collector = LiquidityCollector {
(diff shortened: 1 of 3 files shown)


solver/src/main.rs
```diff
@@ -148,6 +148,10 @@ struct Arguments {
         parse(try_from_str = shared::arguments::wei_from_gwei)
     )]
     gas_price_cap: f64,
+
+    /// The maximum gas price the solver is willing to pay in a settlement
+    #[structopt(long, env, default_value = "0")]
+    paraswap_slippage_bps: usize,
```
vkgnosis, line 154: Is this more natural as a bps, percentage or float ratio? 1 bps = 0.01% = 0.0001 (unit less).
I personally prefer the ratio but I realize it can be hard to read when there are too many zeroes and in finance maybe bps is more common.

vkgnosis (approved):

bh2smith (approved):

solver/src/main.rs
fleupold: I think bps is a quite common unit in finance. The problem with floats is that the conversion to u256 and back may be lossy (but I'm also happy to pass it as a float). Ratios are quite hard to parse from env variable aren't they?

nlordell: `BigDecimal` is also possible as it has a `FromStr` implementation (although might be overkill).

nlordell (approved):

josojo:
If this happens mostly for buy orders, maybe also this issue gives a hint: https://github.com/gnosis/gp-v2-services/issues/788

PR #780, opened 2021-06-23
Implement Balancer swap given out interaction

nlordell:
This PR adds a Balancer swap-given-out interaction and encoding.

### Test Plan

Added unit test to verify interaction encoding is being done correctly using Ethers.js.


bh2smith (approved):
Looks great. Only missing `cargo fmt` at this point. I notice that you only have `SwapGivenOut`. Are we doing the other way around as well.

nlordell:
> Are we doing the other way around as well.

No, as with UNiswap, only the "given out" variant is needed.

vkgnosis (approved):

PR #781, opened 2021-06-23
[HTTP Solver] Weighted Pool Model with Liquidity

bh2smith:
Following #761, we introduce `ConstantProductWeightedPool` into the `AmmModel` and connect this to Weighted Pool Liquidity.

### Test Plan
Updated `convert_settlement` to include new structures.

### DIFF
--- a/shared/src/sources/balancer/swap/fixed_point.rs
+++ b/shared/src/sources/balancer/swap/fixed_point.rs
@@ -34,7 +34,6 @@ lazy_static! {
     static ref MAX_POW_RELATIVE_ERROR: Bfp = Bfp(10000_usize.into());
 }
 
-#[cfg(test)]
 impl From<usize> for Bfp {
     fn from(num: usize) -> Self {
         Self(U256::from(num).checked_mul(*ONE_18).unwrap())
--- a/solver/src/solver/http_solver.rs
+++ b/solver/src/solver/http_solver.rs
@@ -1,19 +1,18 @@
-mod model;
+pub mod model;
 mod settlement;
 
 use self::{model::*, settlement::SettlementContext};
 use crate::{
-    liquidity::{ConstantProductOrder, LimitOrder, Liquidity},
+    liquidity::{ConstantProductOrder, LimitOrder, Liquidity, WeightedProductOrder},
     settlement::Settlement,
     solver::Solver,
 };
 use ::model::order::OrderKind;
 use anyhow::{ensure, Context, Result};
-use bigdecimal::BigDecimal;
 use ethcontract::U256;
 use futures::join;
 use lazy_static::lazy_static;
-use num::{BigRational, ToPrimitive};
+use num::{BigInt, BigRational, ToPrimitive};
 use primitive_types::H160;
 use reqwest::{header::HeaderValue, Client, Url};
 use shared::{
@@ -28,8 +27,12 @@ use std::{
 // Estimates from multivariate linear regression here:
 // https://docs.google.com/spreadsheets/d/13UeUQ9DA4bHlcy9-i8d4nSLlCxSfjcXpTelvXYzyJzQ/edit?usp=sharing
 lazy_static! {
-    static ref GAS_PER_ORDER: U256 = U256::from(66315);
-    static ref GAS_PER_UNISWAP: U256 = U256::from(94696);
+    static ref GAS_PER_ORDER: U256 = U256::from(66_315);
+    static ref GAS_PER_UNISWAP: U256 = U256::from(94_696);
+    // Taken from a sample of two swaps
+    // https://etherscan.io/tx/0x72d234d35fd169ef497ba0a1dc23258c96f278fb688d375d135eb012e5311009
+    // https://etherscan.io/tx/0x1c345a6da1edb2bba953685a4cf85f6a0d967ac751f8c5b518578c5fd20a7c96
+    static ref GAS_PER_BALANCER_SWAP: U256 = U256::from(120_000);
 }
 
 // TODO: exclude partially fillable orders
@@ -107,7 +110,7 @@ impl HttpSolver {
                 Liquidity::ConstantProduct(amm) => {
                     vec![amm.tokens.get().0, amm.tokens.get().1]
                 }
-                _ => vec![],
+                Liquidity::WeightedProduct(amm) => amm.reserves.keys().cloned().collect(),
             })
             .collect::<HashSet<_>>()
             .into_iter()
@@ -172,39 +175,75 @@ impl HttpSolver {
         result
     }
 
-    fn map_constant_product_orders_for_solver(
-        &self,
-        orders: Vec<ConstantProductOrder>,
-    ) -> HashMap<usize, ConstantProductOrder> {
+    fn map_amm_orders_for_solver<T>(&self, orders: Vec<T>) -> HashMap<usize, T> {
         orders.into_iter().enumerate().collect()
     }
 
     fn amm_models(
         &self,
-        amms: &HashMap<usize, ConstantProductOrder>,
+        constant_product_orders: &HashMap<usize, ConstantProductOrder>,
+        weighted_product_orders: &HashMap<usize, WeightedProductOrder>,
         gas_price: f64,
     ) -> HashMap<usize, AmmModel> {
         let uniswap_cost = self.uniswap_cost(gas_price);
-        amms.iter()
+        let mut pool_model_map = HashMap::new();
+        let constant_product_models: HashMap<_, AmmModel> = constant_product_orders
+            .iter()
             .map(|(index, amm)| {
                 let mut reserves = HashMap::new();
                 reserves.insert(amm.tokens.get().0, U256::from(amm.reserves.0));
                 reserves.insert(amm.tokens.get().1, U256::from(amm.reserves.1));
-
-                let uniswap = AmmModel {
+                let pool_model = AmmModel {
                     parameters: AmmParameters::ConstantProduct(ConstantProductPoolParameters {
                         reserves,
                     }),
-                    fee: BigDecimal::from(*amm.fee.numer()) / BigDecimal::from(*amm.fee.denom()),
+                    fee: BigRational::new(
+                        BigInt::from(*amm.fee.numer()),
+                        BigInt::from(*amm.fee.denom()),
+                    ),
                     cost: CostModel {
                         amount: uniswap_cost,
                         token: self.native_token,
                     },
                     mandatory: false,
                 };
-                (*index, uniswap)
+                (*index, pool_model)
             })
-            .collect()
+            .collect();
+        let balancer_cost = self.balancer_cost(gas_price);
+        let weighted_product_models: HashMap<_, AmmModel> = weighted_product_orders
+            .iter()
+            .map(|(index, amm)| {
+                let reserves = amm
+                    .reserves
+                    .iter()
+                    .map(|(token, state)| {
+                        (
+                            *token,
+                            PoolTokenData {
+                                balance: state.balance,
+                                weight: BigRational::from(state.weight),
+                            },
+                        )
+                    })
+                    .collect();
+                let pool_model = AmmModel {
+                    parameters: AmmParameters::WeightedProduct(WeightedProductPoolParameters {
+                        reserves,
+                    }),
+                    fee: amm.fee.clone(),
+                    cost: CostModel {
+                        amount: balancer_cost,
+                        token: self.native_token,
+                    },
+                    mandatory: false,
+                };
+                (*index + constant_product_models.len(), pool_model)
+            })
+            .collect();
+        pool_model_map.extend(constant_product_models);
+        pool_model_map.extend(weighted_product_models);
+        pool_model_map
     }
 
     async fn prepare_model(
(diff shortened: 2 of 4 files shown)


solver/src/solver/http_solver.rs
```diff
@@ -202,9 +206,47 @@ impl HttpSolver {
...
+                        reserves,
+                    }),
+                    fee: BigDecimal::from(*amm.fee.numer()) / BigDecimal::from(*amm.fee.denom()),
+                    cost: CostModel {
+                        amount: balancer_cost,
+                        token: self.native_token,
+                    },
+                    mandatory: false,
+                };
+                (*index + constant_product_models.len(), pool_model)
```
bh2smith, line 244: This indexing should be tested if not already!

solver/src/solver/http_solver.rs
```diff
@@ -202,9 +206,47 @@ impl HttpSolver {
...
+                    .iter()
+                    .map(|(token, state)| {
+                        (
+                            *token,
+                            PoolTokenData {
+                                balance: state.balance,
+                                // Bfp -> BigRational -> String -> BigDecimal
+                                weight: BigDecimal::from_str(
+                                    &BigRational::from(state.weight).to_string(),
+                                )
```
vkgnosis, line 227: I see that `weight` is BigDecimal. Is that correct? I think I started a discussion somewhere that we really only use BigDecimal for postgres and we should avoid using it in other places and use BigRational instead.

Regardless, we probably should not convert by going through string. This is bad for performance (doesn't matter that much) and probably incorrect because it depends on how BigRational's to_string implementation works.
If we do want to keep BigDecimal then let's move the conversion into a separate function that we can test. I suspect there are issues with the fractional part.

Alternatively it might be easier to implement a string conversion on Bfp directly instead of going through BigRational.

nlordell: The HTTP solver expects weights in big decimal notation.

> Alternatively it might be easier to implement a string conversion on Bfp directly instead of going through BigRational.

`Bpf` is a Balancer-specific representation of a big decimal. If the HTTP solver is to be generic WRT how it handles weighted pools, we shouldn't expect a `Bfp` weight.

vkgnosis: I think the correct phrasing is that the http solver expects a string that is the decimal representation of a rational number. This doesn't mean it has to be a BigDecimal in our code, we just need to serialize it properly. BigDecimal happens to do this but also it has a different inner representation than BigRational so the choice isn't obvious.

> Bpf is a Balancer-specific representation of a big decimal.

What I mean is that it might be easier to have `weight` be a string explicitly and then have a specialized BfP to String function that can be used here. This assumes that it is hard for go from BfP to BigDecimal but on second thought it is not. You can use the same division trick from a couple of lines above to avoid going through BigRational -> String> BigDecimal.

I also want to note that BigDecimal vs BigRational is not just about inner representation and to_string. BigRational actually stores rational numbers in perfect precision. BigDecimal does not. For example the number `1 / 3` can only be approximated by BigDecimal and it does so in the division implementation.
I feel that this is another point in favor of using BigRational to store `weight`. Of course we still need to serialize it as a decimal string which could still be implemented as lossy conversion to BigDecimal and then to_string.

nlordell: > I think the correct phrasing is that the http solver expects a string that is the decimal representation of a rational number. This doesn't mean it has to be a BigDecimal in our code, we just need to serialize it properly.

Makes sense. What I meant was that the conversion here should only be for building the data that we pass on the HTTP solver (and not how we internally represent weights), so in that regard it is acting as a glorified string formatter.

> This assumes that it is hard for go from BfP to BigDecimal but on second thought it is not. You can use the same division trick from a couple of lines above to avoid going through BigRational -> String> BigDecimal.

I agree, I also really dislike `BigRational` -> `String` -> `BigDecimal` (and, I'm not 100 sure it would work, as I was under the impression that `BigRational::to_string` produces `1/10` output, while `BigDecimal::from_str` expects `0.1` for example). I was just trying to answer one part of your comment :stuck_out_tongue:.

> I feel that this is another point in favor of using BigRational to store weight.

For some reason, I thought that we already represented weights as `BigRational`. I also agree that this should be the case.

bh2smith: Ok folks. I am glad I included my comment here in the code that we were converting `BigRational -> String / String -> BigDecimal` (note that the numerator and denominator as strings should be fine), but I am still lost for suggestions here as to what we actually want to do. 

We already had these discussions regarding how `Bfp -> BigRational` was the ideal conversion we wanted to have in the balancer Bfp module which resulted in #775. So now that we have `Bfp` weight and a conversion tool making this into a `BigRational`... what do we give to the HTTP solver here?

bh2smith: I could do `BigDecimal(rational.numerator().to_string()) / BigDecimal(rational.demominator().to_string())`

nlordell: > (note that the numerator and denominator as strings should be fine)

I do not think this is true:
```rust
println!(
    "{:?}",
    BigDecimal::from_str(&BigRational::new(1.into(), 1000.into()).to_string()),
);
```

Produces:
```
Err(ParseBigInt(ParseBigIntError { kind: InvalidDigit }))
```

bh2smith: The suggestion is to get the numerator and denominator separately as strings.

vkgnosis: The minimal change that I would like to see in this PR is to refactor the BfP -> BigDecimal conversion into a separate, testable function with unit tests.
The exact implementation details aren't that important but it is likely that converting the numerator and denominator separately is the way to go.
The other part of the discussion is whether `weights` should be BigRational instead of BigDecmial. We can do this in a separate issue to not block this PR.

bh2smith: > The minimal change that I would like to see in this PR is to refactor the BfP -> BigDecimal conversion into a separate, testable function with unit tests.
> 
> The exact implementation details aren't that important but it is likely that converting the numerator and denominator separately is the way to go.
> 
> The other part of the discussion is whether `weights` should be BigRational instead of BigDecmial. We can do this in a separate issue to not block this PR.

Yea I can do this, but I would like to point out that this is what I had already done in #775 before you suggested not to convert to BigDecimal.

vkgnosis: Thanks for finding that PR. I remembered that discussion but couldn't quickly find it. What I meant there was to always store BigRational in structs and then use one of those serde attributes like `#[serde(with = "u256_decimal")]` to do the conversion to a decimal string. Within this conversion we could go through BigDecimal so that it is just an implementation detail.

nlordell: > The other part of the discussion is whether weights should be BigRational instead of BigDecmial.

AFAIU, weights are currently `Bfp` but should be `BigRational` (because of the reasons you brought up - and we already use `BigRational` for fees for example). I think the `BigDecimal` conversion is only done for the data being sent to the HTTP solver, so its basically used for string formatting (and could just as well be `String` - although personally I prefer `BigDecimal` to avoid type erasure, although it could just as well be implemented as a `BigRational`with a `serde` formatter).

That being said I do think that weights should be `BigRational` in general and not `Bfp`.

> We can do this in a separate issue to not block this PR.

Agreed :+1:.

> The minimal change that I would like to see in this PR is to refactor the BfP -> BigDecimal conversion into a separate, testable function with unit tests.

:+1:

bh2smith: I have made a new commit applying the suggested changes. Still not super happy about the unwrap used in `weight: bfp_to_big_decimal(state.weight).unwrap()` but not unwrapping would mean changing the entire method signature just for this one conversion. Suggestions welcome.

solver/src/util.rs
```diff
@@ -20,3 +25,40 @@ macro_rules! chain {
         }
     };
 }
+
+pub fn bfp_to_big_decimal(num: Bfp) -> Result<BigDecimal> {
+    let rational = BigRational::from(num);
+    let top = BigDecimal::from_str(&rational.numer().to_string())?;
+    let bottom = BigDecimal::from_str(&rational.denom().to_string())?;
```
vkgnosis, line 32: I think there is a way to get rid of the string conversions and thus the Result. BigDecimal implements From<BigInt> and BigRational's numer and denom are BigInt so you can keep the division but go through BigInt.

```rust
let bottom = BigDecimal::from(rational.denom());
```

A potential problem is that BigDecimal is using an old version of the num crate (0.3 instead of 0.4) which makes the types incompatible.
The way to work around that is to depend on both versions of BigInt directly and convert the 0.4 BigInt into bytes with `to_bytes_le` and then create the 0.3 BigInt with `from_bytes_le`.

solver/src/solver/http_solver.rs
```diff
@@ -202,9 +206,44 @@ impl HttpSolver {
...
+                let reserves = amm
+                    .reserves
+                    .iter()
+                    .map(|(token, state)| {
+                        (
+                            *token,
+                            PoolTokenData {
+                                balance: state.balance,
+                                // Probably shouldn't unwrap here.
+                                weight: bfp_to_big_decimal(state.weight).unwrap(),
```
vkgnosis, line 225: If we believe that the function cannot error then we should do the unwrap inside of it as an implementation detail so that the signature doesn't contain Result.

nlordell: I also really liked @vkgnosis's suggestion of doing:

> was to always store `BigRational` in structs and then use one of those `serde` attributes like `#[serde(with = "u256_decimal")]`

And would prefer this over doing the type conversion directly.

solver/src/util.rs
bh2smith: I will try this out, but I believe I have already tried this and it was a conflicting BigInt crate.

bh2smith: It appears to be a completely different crate `num_bigint` that implements `BigDecimal::from`

solver/src/solver/http_solver.rs
bh2smith: OMG yes. I will do this if it actually works.

bh2smith: Unfortunately, this results in a Serialization issue:

```
error[E0308]: mismatched types
  --> solver/src/solver/http_solver/model.rs:55:17
   |
55 | #[derive(Debug, Serialize)]
   |                 ^^^^^^^^^ expected struct `U256`, found struct `Ratio`
   ```
I suppose we have to manually implement this or something?

File: solver/src/util.rs (hunk 3 above)
Comment (nlordell) [reply solver/src/util.rs:32]: Right, so what @vkgnosis was suggesting is that there are 2 `BigInt` types with the same name from different versions of the same crate. The compiler considers them different types internally (despite the same name, which can cause confusing compiler errors).

So, what you can do is:
```rust
use num::BigInt as BigInt04;
use num_bigint::BigInt as BigInt03;

let value_04: BigInt04 = // ...
let value_03 = BigInt03::from_bytes_le(value_04.to_bytes_le());
```

To convert between the types in an error-free way.

solver/src/solver/http_solver.rs
nlordell: Yes, you would have to implement your own `big_rational_decimal` formatter in the same way the `U256` one was implemented.

bh2smith: This crate containing BigRational (https://docs.rs/num/0.4.0/) doesn't appear to have a `serde_with` thingy

bh2smith: Ahh I see now, I wasn't aware we had written this for ourselves.

bh2smith: At the end of the day I am still going to have to write a string representation of this. Here is my most recent approach

```rust
    let float_rep = value.numer().to_f64().unwrap() / value.denom().to_f64().unwrap();
    serializer.serialize_str(&float_rep.to_string())
```

I mean there are so many ways to do this, but we will have to do it.

bh2smith: I will create a separate PR for this BTW.

solver/src/util.rs
bh2smith: Alrighty, we now have a new PR #813 which addresses this issue and will make the changes here sweet and smooth.

solver/src/solver/http_solver.rs
bh2smith: Resolved by #813

shared/src/sources/balancer/swap/fixed_point.rs
```diff
@@ -34,7 +34,6 @@ lazy_static! {
     static ref MAX_POW_RELATIVE_ERROR: Bfp = Bfp(10000_usize.into());
 }
 
-#[cfg(test)]
```
nlordell, line 37: I don't think this is needed right?

solver/src/solver/http_solver.rs
nlordell: Can you maybe add a comment here to clarify (so it doesn't surprise someone reading the code later?).

solver/src/solver/http_solver/model.rs
```diff
@@ -32,7 +35,8 @@ pub struct OrderModel {
 pub struct AmmModel {
     #[serde(flatten)]
     pub parameters: AmmParameters,
-    pub fee: BigDecimal,
+    #[serde(with = "ratio_as_decimal")]
```
nlordell, line 38: :tada:

nlordell (approved):
Looks good. A couple of very minor comments.

I'm not a huge fan of the duplicated constant-product / weighted-product code, but I really don't see a way around it without a much larger refactor, and think the approach here is pragmatic.

shared/src/sources/balancer/swap/fixed_point.rs
bh2smith: It was the only way I could use `Bfp::from(usize)` in a test elsewhere.

nlordell: Ah, because its needed outside of the `shared` crate. Makes sense.

solver/src/solver/http_solver.rs
bh2smith: Done!

PR #782, opened 2021-06-23
Better 1Inch Unit Tests

fleupold:
This PR makes the 1Inch solver unit testable by abstracting away the API Client (and the allowance fetching part of the smart contract interaction).

It then removes all the `ignored` e2e tests (except for one final e2e test) and replaces them with CI executable unit tests. We now have tests for

- ignoring buy orders
- respecting limit prices
- setting allowance
- disabled protocol

Let me know if you can think of more test cases that are interesting.

### Test Plan
`cargo test`

### DIFF
--- a/solver/src/solver/oneinch_solver.rs
+++ b/solver/src/solver/oneinch_solver.rs
@@ -95,6 +95,12 @@ where
         let spender = self.client.get_spender().await?;
         let sell_token = ERC20::at(&self.web3(), order.sell_token);
 
+        // Fetching allowance before making the SwapQuery so that the Swap info is as recent as possible
+        let existing_allowance = self
+            .allowance_fetcher
+            .existing_allowance(order.sell_token, spender.address)
+            .await?;
+
         let query = SwapQuery {
             from_token_address: order.sell_token,
             to_token_address: order.buy_token,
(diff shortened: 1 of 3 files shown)


solver/src/solver/oneinch_solver.rs
```diff
@@ -125,6 +129,11 @@ impl OneInchSolver {
 
         settlement.with_liquidity(&order, order.sell_amount)?;
 
+        let existing_allowance = self
```
nlordell, line 132: Is there a reason this moved here? The reason I had it near the beginning was to make the 1Inch API call as late as possible to make price movement less likely. I'm not sure if it has a significant impact however, so maybe it doesn't really matter.

nlordell (approved):
Nice tests!

solver/src/solver/oneinch_solver.rs
fleupold: I wanted to avoid doing a call to the node if the limit price is not met, happy to move it back up

nlordell: Either way, I don't think it makes much difference. The question becomes: which do we value more?
- Additional node stress when 1Inch price isn't good enough?
- Price older by a single node round trip?

I'm OK with either decision.

josojo (approved):
Nice PRs, nice tests.

I am currently crafting the same for matcha!

vkgnosis (approved):

PR #785, opened 2021-06-23
Add a mock transport

nlordell:
This PR adds a mock transport mostly generated by `mockall` with some glue code to make it implement `web3::Transport`. This will allow us write unit tests for code that depend on a `Web3` instance.

### Test Plan

Added code to be used for testing - also added some tests as a "sanity check" to make sure the mocking is working as expected.

### DIFF
--- a/shared/src/transport/mock.rs
+++ b/shared/src/transport/mock.rs
@@ -1,0 +1,173 @@
+//! Mockable Web3 transport implementation.
+
+use anyhow::{Error, Result};
+use ethcontract::{
+    futures::future::{self, Ready},
+    jsonrpc::{Call, Id, MethodCall, Params},
+    web3::{self, BatchTransport, RequestId, Transport},
+};
+use serde_json::Value;
+use std::{
+    fmt::{self, Debug, Formatter},
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc, Mutex, MutexGuard,
+    },
+};
+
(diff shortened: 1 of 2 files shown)


shared/src/transport/mock.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    #[tokio::test]
+    async fn can_mock_batch_requests() {
+        let transport = MockTransport::new();
+        transport
+            .mock()
+            .expect_execute_batch()
+            .with(eq(vec![
+                ("foo_bar".to_owned(), vec![json!(true), json!("stuff")]),
```
bh2smith, line 149: I like how you use json to show use a few different kinds of values we can use!

bh2smith (approved):
This is like Christmas. Very excited to play with this!

vkgnosis (approved):

fleupold:
Very cool that this allows reusing mockall matchers. I was wondering if web3 itself or ethcontract should also export this kind of functionality (or if it could even be its own crate), since it is so useful yet basic.

I guess one argument against is that mockall is kind of our personal choice for a mocking library.

nlordell:
> or if it could even be its own crate

I think this could be part of `web3` under a feature. They already can change the TLS implementation for the HTTPS transport and it would make sense to have an optional `MockTransport` behind a feature gate, so it can be enabled by projects that choose to use `mockall` as well (or left disabled for projects that don't).

PR #786, opened 2021-06-23
Add allowance manager component

nlordell:
This PR adds an allowance manager component that can be used to generate approval interactions based on allowances. The allowance manager acts similarly to how allowances are handled in the Uniswap liquidity/settlement handing, in that allowances are pre-fetched and then later used to generate the required interactions.

Additionally, a method for fetching a single approval was also added that can be used by things like the 1Inch and Paraswap solver. 

Currently, this component is unused, and will only get used in a follow up PR.

### Test Plan

Added unit tests for logic.

### DIFF
--- a/solver/src/interactions/allowances.rs
+++ b/solver/src/interactions/allowances.rs
@@ -98,10 +98,17 @@ impl Interaction for Approval {
     }
 }
 
-/// An allowance manager that an address.
+/// An allowance manager that retrive approval interactions for a given owner
+/// address.
 pub struct AllowanceManager {
     web3: Web3,
     owner: H160,
+}
+
+impl AllowanceManager {
+    pub fn new(web3: Web3, owner: H160) -> Self {
+        Self { web3, owner }
+    }
 }
 
 #[async_trait::async_trait]
--- a/solver/src/interactions/allowances.rs
+++ b/solver/src/interactions/allowances.rs
@@ -5,9 +5,9 @@
 use crate::{
     encoding::EncodedInteraction, interactions::Erc20ApproveInteraction, settlement::Interaction,
 };
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, bail, Result};
 use contracts::ERC20;
-use ethcontract::{batch::CallBatch, H160, U256};
+use ethcontract::{batch::CallBatch, errors::ExecutionError, H160, U256};
 use maplit::hashset;
 use shared::{dummy_contract, Web3};
 use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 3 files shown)


vkgnosis:
Something is wrong with the rebase so the PR looks a lot bigger than it is.

nlordell:
> Something is wrong with the rebase so the PR looks a lot bigger than it is.

Fixed. Sorry, I hadn't rebased.

solver/src/interactions/allowances.rs
```diff
@@ -0,0 +1,291 @@
...
+    let _ = batch.execute_all(MAX_BATCH_SIZE).await;
+
+    let mut allowances = HashMap::new();
+    for (token, allowance) in results {
+        let allowance = match allowance.await {
+            Ok(value) => value,
+            Err(err) => {
+                tracing::warn!("error retrieving allowance for token {:?}: {}", token, err);
+                continue;
+            }
```
vkgnosis, line 147: Is `continue` the right reaction to an error? This means we treat failure to get allowance the same as no allowance. If all of the request failed because of a transport error we would return an empty HashMap instead of an error.
Maybe it makes sense to return Result::Err on any non smart contract error and only continue on smart contract errrors.

vkgnosis (approved):

solver/src/interactions/allowances.rs
```diff
@@ -0,0 +1,291 @@
...
+                    amount: U256::max_value(),
+                };
+
+                approve.encode()
+            }
+        }
+    }
+}
+
+/// An allowance manager that an address.
```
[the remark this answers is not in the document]
nlordell: :see_no_evil:

solver/src/interactions/allowances.rs
nlordell: The main reason I did this was that this is how allowances were treated in the current Uniswap liquidity implementation:
https://github.com/gnosis/gp-v2-services/blob/0bca0d5a4e05fcd8ace2ea4cf5e97ad5d555285b/solver/src/liquidity/uniswap.rs#L108-L112

Additionally, this method would return an `Allowances` "cache" instance that treats missing allowances as errors, and different then tokens with allowance 0:

https://github.com/gnosis/gp-v2-services/blob/c29d89a506d768d5603e2dd146aa432fdc1347a9/solver/src/interactions/allowances.rs#L53

So I thought this was a happy medium where we could cache allowances ignoring errors (so it doesn't error out for tokens that end up not getting used at all) and then potentially produce errors when trying to make an `Approval` interaction for a token whose allowance we were unable to retrieve.

fleupold: I think a single token failure shouldn't make the entire request fail. Maybe we can return an error if the top level batch request fails?

fleupold (approved):
This looks like a very useful component. Looking forward to some red diffs when it's started being used.

solver/src/interactions/allowances.rs
nlordell: Added slightly hacky support to detect batch errors. I [created an issue in `ethcontract`](https://github.com/gnosis/ethcontract-rs) to improve batch error handling in order to make the detection more reliable.

fleupold: My suggestion for a fix would have been to replace

```rust
batch.execute_all(MAX_BATCH_SIZE).await;
```

with
```rust
batch.execute_all(MAX_BATCH_SIZE).await?;
```

If the batch itself fails the top level execute_all would fail wouldn't it?

nlordell: [`execute_all` doesn't return a result](https://docs.rs/ethcontract/0.12.2/ethcontract/batch/struct.CallBatch.html#method.execute_all) and seems to just forward errors to each call future.

PR #787, opened 2021-06-24
[Easy] Move H160Wrapper to shared

bh2smith:
Related to a change needed for #761 - we will need to put H160 in a Hashmap for Deserialization.

Only logical change is that the struct's inner H160 is now public.

### Test Plan
now new tests.


nlordell (approved):

shared/src/lib.rs
```diff
@@ -31,3 +37,15 @@ pub mod transport;
 pub mod web3_traits;
 
 pub type Web3 = web3::Web3<transport::MetricTransport<crate::http_transport::HttpTransport>>;
+
```
nlordell, line 40: We can also have a `types` module or something for the `H160Wrapper` instead of putting it in the root - but no strong feelings.

PR #790, opened 2021-06-24
Paraswap buy order slippage

### LINKED ISSUE #788: Paraswap buy orders and slippage
Currently, we have the following logic for buy orders in parawasp solver
```
// 0.1% slippage
        let dest_amount_with_slippage = price_response
            .dest_amount
            .checked_mul(999.into())
            .ok_or_else(|| anyhow!("Overflow during slippage computation"))?
            / 1000;
        let transaction_query = TransactionBuilderQuery {
            src_token: order.sell_token,
            dest_token: order.buy_token,
            src_amount: price_response.src_amount,
            dest_amount: dest_amount_with_slippage,
   ```
   Hence, to my understanding, the user might not get the specified buy amount, due to the fact that we build in the slippage. To my understanding, we should increase the src_amount by 0.001% for buy orders, instead of reducing the dest_amount. 
   Or do I miss something?

fleupold:
Fixes #788

Great catch by @josojo and the real reason for the `destAmount` mismatch error we have been seing. I think it's still worthwhile to keep slippage configurable (based on #778)  but let me know if you feel otherwise.

### Test Plan
Added an e2e test which confirms that this was actually the issue (if slippage is added on other side we get exactly the error we see).

Also added unit test to the solver that makes sure slippage is computed correctly.

### DIFF
--- a/solver/src/solver/paraswap_solver.rs
+++ b/solver/src/solver/paraswap_solver.rs
@@ -114,16 +114,31 @@ where
             return Ok(None);
         }
 
-        let dest_amount_with_slippage = price_response
-            .dest_amount
-            .checked_mul((10000 - self.slippage_bps).into())
-            .ok_or_else(|| anyhow!("Overflow during slippage computation"))?
-            / 10000;
+        let (src_amount, dest_amount) = match order.kind {
+            // Buy orders apply slippage to src amount, dest amount unchanged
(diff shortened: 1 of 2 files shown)


vkgnosis (approved):

josojo (approved):
you are fast!

PR #792, opened 2021-06-25
Adding matcha solver

josojo:
This code adds the matcha solver, similar to paraswap. Also, it refactors used components into solver_utils.rs.

Though, I have some questions and comments:
1. Why was for oneinch the pmm disabled? Do I need to do something similar for matcha?
2. The matcha API for buy orders is still a riddle to me. If we change the slippage percentage, they don't increase the sell_amounts, hence I think they are allowing the slippage on the buy order side, which makes more sense to me anyways. Though the API docs state it differently. I am in contact with the team.
In the current paraswap API, we will always sell "too many sell token" and hence continuously overpay paraswap, which is far from optimal. Having the slippage on the buy_amount seems best to me.
3. How did you guys test the solvers before? I just run it with an unregisted solver private key, and got:
```
0: method 'settle(address[],uint256[],(uint256,uint256,address,uint256,uint256,uint32,bytes32,uint256,uint256,uint256,bytes)[],(address,uint256,bytes)[][3])' failure: contract call reverted with message: Some("GPv2: not a solver")
    1: contract call reverted with message: Some("GPv2: not a solver")
2021-06-24T16:31:38.753Z  WARN solver::driver: settlement failure for: 
SettlementWithSolver {
    name: "Matcha",
    settlement: Settlement {
        encoder: SettlementEncoder {
            tokens: [
                0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2,
                0xdac17f958d2ee523a2206206994597c13d831ec7,
                0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee,
            ],
            clearing_prices: {
                0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee: 200000000000000,
                0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2: 200000000000000,
                0xdac17f958d2ee523a2206206994597c13d831ec7: 397918,
            },
            trades: [
                Trade {
                    order: Order {
                        order_meta_data: OrderMetaData {
                            creation_date: 2021-06-24T16:30:58.325347Z,
                            owner: 0x070330e027a3e4327ba69f0c8a1885df4b545837,
                            uid: 0xdac08f2f83c9bf8100b918aedfc336a7461cd8252cec7fcac20c48e92cdcdc36070330e027a3e4327ba69f0c8a1885df4b54583760d4b7f1,
                            available_balance: Some(
                                5978826,
                            ),
```
hence I am quite confident that it works nicely, though not 100%. Any better test suggestions are welcomed :) 

### Test Plan
Unit tests

### DIFF
--- a/solver/src/solver/matcha_solver.rs
+++ b/solver/src/solver/matcha_solver.rs
@@ -2,6 +2,20 @@
 //!
 //! This solver will simply use the Matcha API to get a quote for a
 //! single GPv2 order and produce a settlement directly against Matcha.
+//!
+//! Please be aware of the following subtlety for buy orders:
+//! 0x's API is adding the defined slippage on the sellAmount of a buy order
+//! and then returns the surplus in the buy amount to the user.
+//! I.e. if the user defines a 5% slippage, they will sell 5% more, and receive 5%
+//! more buy-tokens than ordered. Here is on example tx:
+//! https://dashboard.tenderly.co/gp-v2/staging/simulator/new?block=12735030&blockIndex=0&from=0xa6ddbd0de6b310819b49f680f65871bee85f517e&gas=8000000&gasPrice=0&value=0&contractAddress=0x3328f5f2cecaf00a2443082b657cedeaf70bfaef&rawFunctionInput=0x13d79a0b000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000e0000000000000000000000000000000000000000000000000000000000000014000000000000000000000000000000000000000000000000000000000000003600000000000000000000000000000000000000000000000000000000000000002000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2000000000000000000000000dac17f958d2ee523a2206206994597c13d831ec7000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000029143e200000000000000000000000000000000000000000000000000470de4df820000000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000036416d81e590ff67370e4523b9cd3257aa0a853c000000000000000000000000000000000000000000000000000000000291f64800000000000000000000000000000000000000000000000000470de4df8200000000000000000000000000000000000000000000000000000000000060dc5839000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000003dc140000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000029143e2000000000000000000000000000000000000000000000000000000000000016000000000000000000000000000000000000000000000000000000000000000410a7f27a6638cc9cdaba8266a15acef4cf7e1e1c9b9b2059391b7230b67bdfeb21f1d3aa45852f527a5040d3d7a190b92764a2c854f334b7eed579b390b85fd3f1b000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000003800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000120000000000000000000000000dac17f958d2ee523a2206206994597c13d831ec7000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000044095ea7b3000000000000000000000000def1c0ded9bec7f1a1670819833240f027b25effffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff00000000000000000000000000000000000000000000000000000000000000000000000000000000def1c0ded9bec7f1a1670819833240f027b25eff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000128d9627aa400000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000002b220e100000000000000000000000000000000000000000000000000470de4df82000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000000000000000000000dac17f958d2ee523a2206206994597c13d831ec7000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2869584cd000000000000000000000000100000000000000000000000000000000000001100000000000000000000000000000000000000000000003239e38b8a60dc53b70000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000&network=1   
+//! This behavior has the following risks: The additional sell tokens from the slippage
+//! are not provided by the user, hence the additional tokens might be not available in
+//! the settlement contract. For smaller amounts this is unlikely, as we always charge the
+//! fees also in the sell token, though, the fee's might not always be sufficient.
+//! This risk should be covered in a future PR.
+//!
+//! Sell orders are unproblematic, especially, since the positive slippage is handed back from matcha
 
 pub mod api;
 
@@ -26,7 +40,6 @@ use shared::Web3;
 use std::fmt::{self, Display, Formatter};
 
 /// Constant maximum slippage of 5 BPS (0.05%) to use for on-chain liquidity.
-/// This is half the 1inch slippage
 pub const STANDARD_MATCHA_SLIPPAGE_BPS: u16 = 5;
 
 /// A GPv2 solver that matches GP orders to direct Matcha swaps.
@@ -89,10 +102,6 @@ impl SingleOrderSolving for MatchaSolver {
                         STANDARD_MATCHA_SLIPPAGE_BPS,
                     )
                     .unwrap(),
-                    // From the api documentation:
-                    // SlippagePercentage(Optional): The maximum acceptable slippage in % of the buyToken amount if sellAmount is provided, the maximum acceptable slippage in % of the sellAmount amount if buyAmount is provided. This parameter will change over time with market conditions.
-                    // => Hence, allegedly, we don't need to build in buffers ourselves. Though the sell amount is not adjusted, if slippage is changed for buy order requests.
-                    // Todo: Continue discussion with the 0x-team
                     skip_validation: Some(true),
                 };
 
@@ -219,7 +228,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_satisfies_limit_price_for_sell_order() {
+    async fn test_satisfies_limit_price_for_orders() {
         let mut client = Box::new(MockMatchaApi::new());
         let mut allowance_fetcher = Box::new(MockAllowanceManaging::new());
 
@@ -227,22 +236,21 @@ mod tests {
         let buy_token = H160::from_low_u64_be(1);
 
         let allowance_target = shared::addr!("def1c0ded9bec7f1a1670819833240f027b25eff");
-        client.expect_get_swap().returning(move|_| {
+        client.expect_get_swap().returning(move |_| {
             Ok(SwapResponse {
-            sell_amount: U256::from_dec_str("100").unwrap(),
-             buy_amount: U256::from_dec_str("91").unwrap(),
-             allowance_target,
-            price: 0.91_f64,
-            to: shared::addr!("def1c0ded9bec7f1a1670819833240f027b25eff"),
-            data: web3::types::Bytes(hex::decode(
-                "d9627aa40000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000016345785d8a00000000000000000000000000000000000000000000000000001206e6c0056936e100000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc20000000000000000000000006810e776880c02933d47db1b9fc05908e5386b96869584cd0000000000000000000000001000000000000000000000000000000000000011000000000000000000000000000000000000000000000092415e982f60d431ba"
-            ).unwrap()),
-            value: U256::from_dec_str("0").unwrap(),
-        })});
+                sell_amount: U256::from_dec_str("100").unwrap(),
+                buy_amount: U256::from_dec_str("91").unwrap(),
+                allowance_target,
+                price: 0.91_f64,
+                to: shared::addr!("0000000000000000000000000000000000000000"),
+                data: web3::types::Bytes(hex::decode("00").unwrap()),
+                value: U256::from_dec_str("0").unwrap(),
+            })
+        });
 
         allowance_fetcher
             .expect_get_approval()
-            .times(1)
+            .times(2)
             .with(eq(sell_token), eq(allowance_target), eq(U256::from(100)))
             .returning(move |_, _, _| {
                 Ok(Approval::Approve {
@@ -256,7 +264,23 @@ mod tests {
             allowance_fetcher,
         };
 
-        let order_passing_limit = LimitOrder {
+        let buy_order_passing_limit = LimitOrder {
+            sell_token,
+            buy_token,
+            sell_amount: 101.into(),
+            buy_amount: 91.into(),
+            kind: model::order::OrderKind::Buy,
+            ..Default::default()
+        };
+        let buy_order_violating_limit = LimitOrder {
+            sell_token,
+            buy_token,
+            sell_amount: 99.into(),
+            buy_amount: 91.into(),
+            kind: model::order::OrderKind::Buy,
+            ..Default::default()
+        };
+        let sell_order_passing_limit = LimitOrder {
             sell_token,
             buy_token,
             sell_amount: 100.into(),
@@ -264,7 +288,7 @@ mod tests {
             kind: model::order::OrderKind::Sell,
             ..Default::default()
         };
-        let order_violating_limit = LimitOrder {
+        let sell_order_violating_limit = LimitOrder {
             sell_token,
             buy_token,
             sell_amount: 100.into(),
@@ -274,7 +298,7 @@ mod tests {
         };
 
         let result = solver
-            .settle_order(order_passing_limit)
+            .settle_order(sell_order_passing_limit)
             .await
             .unwrap()
             .unwrap();
@@ -285,67 +309,15 @@ mod tests {
                 buy_token => 100.into(),
             }
         );
-
-        let result = solver.settle_order(order_violating_limit).await.unwrap();
-        assert!(result.is_none());
-    }
-    #[tokio::test]
-    async fn test_satisfies_limit_price_for_buy_order() {
-        let mut client = Box::new(MockMatchaApi::new());
-        let mut allowance_fetcher = Box::new(MockAllowanceManaging::new());
-
-        let sell_token = H160::from_low_u64_be(1);
-        let buy_token = H160::from_low_u64_be(1);
-
-        let allowance_target = shared::addr!("def1c0ded9bec7f1a1670819833240f027b25eff");
-        client.expect_get_swap().returning(move|_| {
-            Ok(SwapResponse {
-            sell_amount: U256::from_dec_str("100").unwrap(),
-             buy_amount: U256::from_dec_str("91").unwrap(),
-             allowance_target,
-            price: 0.91_f64,
-            to: shared::addr!("def1c0ded9bec7f1a1670819833240f027b25eff"),
-            data: web3::types::Bytes(hex::decode(
-                "d9627aa40000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000016345785d8a00000000000000000000000000000000000000000000000000001206e6c0056936e100000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc20000000000000000000000006810e776880c02933d47db1b9fc05908e5386b96869584cd0000000000000000000000001000000000000000000000000000000000000011000000000000000000000000000000000000000000000092415e982f60d431ba"
-            ).unwrap()),
-            value: U256::from_dec_str("0").unwrap(),
-        })});
-
-        allowance_fetcher
-            .expect_get_approval()
-            .times(1)
-            .with(eq(sell_token), eq(allowance_target), eq(U256::from(100)))
-            .returning(move |_, _, _| {
-                Ok(Approval::Approve {
-                    token: sell_token,
-                    spender: allowance_target,
-                })
-            });
-
-        let solver = MatchaSolver {
-            client,
-            allowance_fetcher,
-        };
-
-        let order_passing_limit = LimitOrder {
-            sell_token,
-            buy_token,
-            sell_amount: 101.into(),
-            buy_amount: 91.into(),
-            kind: model::order::OrderKind::Buy,
-            ..Default::default()
-        };
-        let order_violating_limit = LimitOrder {
-            sell_token,
-            buy_token,
-            sell_amount: 99.into(),
-            buy_amount: 91.into(),
-            kind: model::order::OrderKind::Buy,
-            ..Default::default()
-        };
 
         let result = solver
-            .settle_order(order_passing_limit)
+            .settle_order(sell_order_violating_limit)
+            .await
+            .unwrap();
+        assert!(result.is_none());
+
+        let result = solver
+            .settle_order(buy_order_passing_limit)
             .await
             .unwrap()
             .unwrap();
@@ -357,7 +329,10 @@ mod tests {
             }
         );
 
-        let result = solver.settle_order(order_violating_limit).await.unwrap();
+        let result = solver
+            .settle_order(buy_order_violating_limit)
+            .await
+            .unwrap();
         assert!(result.is_none());
     }
 
@@ -385,13 +360,11 @@ mod tests {
         client.expect_get_swap().returning(move |_| {
             Ok(SwapResponse {
                 sell_amount: U256::from_dec_str("100").unwrap(),
-                 buy_amount: U256::from_dec_str("91").unwrap(),
-                 allowance_target ,
+                buy_amount: U256::from_dec_str("91").unwrap(),
+                allowance_target,
                 price: 13.121_002_575_170_278_f64,
-                to: shared::addr!("def1c0ded9bec7f1a1670819833240f027b25eff"),
-                data: web3::types::Bytes(hex::decode(
-                    "d9627aa40000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000016345785d8a00000000000000000000000000000000000000000000000000001206e6c0056936e100000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc20000000000000000000000006810e776880c02933d47db1b9fc05908e5386b96869584cd0000000000000000000000001000000000000000000000000000000000000011000000000000000000000000000000000000000000000092415e982f60d431ba"
-                ).unwrap()),
+                to: shared::addr!("0000000000000000000000000000000000000000"),
+                data: web3::types::Bytes(hex::decode("").unwrap()),
                 value: U256::from_dec_str("0").unwrap(),
             })
         });
--- a/solver/src/solver/matcha_solver.rs
+++ b/solver/src/solver/matcha_solver.rs
@@ -68,7 +68,7 @@ impl<F> std::fmt::Debug for MatchaSolver<F> {
 #[async_trait::async_trait]
 impl<F: AllowanceFetching> SingleOrderSolving for MatchaSolver<F> {
     async fn settle_order(&self, order: LimitOrder) -> Result<Option<Settlement>> {
-        let (swap, mut settlement) = match order.kind {
+        let swap = match order.kind {
             OrderKind::Sell => {
                 let query = SwapQuery {
                     sell_token: order.sell_token,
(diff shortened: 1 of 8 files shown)


fleupold:
> Why was for oneinch the pmm disabled? Do I need to do something similar for matcha?

We ran into weird issues with PMM1 (e.g. https://dashboard.tenderly.co/gp-v2/staging/simulator/8e1a49b8-2a8e-4d7f-b50c-83f93c5dfef1) calling a non-existent method on the closed source router. No need to do anything unless we see similar problems for match

> How did you guys test the solvers before? I just run it with an unregisted solver private key, and got:

You should be able to get the tenderly link, which would allow you to replace the sender with a real solver and simulate the transaction in that manner. Also, commenting out the submit_solution line in driver.rs and then testing locally with real orders worked for me. If they are available on rinkeby that could also be an option.

> In the current paraswap API, we will always sell "too many sell token" and hence continuously overpay paraswap, which is far from optimal.

I don't understand the concern here? How is selling too many sell tokens different from receiving too few buy tokens?

How does matcha deal with positive slippage in general?

solver/src/solver/matcha_solver/api.rs
```diff
@@ -0,0 +1,227 @@
...
+use reqwest::{Client, IntoUrl, Url};
+use serde::Deserialize;
+use shared::http::default_http_client;
+
+/// A Matcha API quote query parameters.
+///
+/// These parameters are currently incomplete, and missing parameters can be
+/// added incrementally as needed.
+#[derive(Clone, Debug, Default)]
+pub struct SwapQuery {
```
fleupold, line 22: Should we set an `affiliateAddress` just in case? So we can track our affiliation with them?

fleupold (commented):
Only looked at the API so far (it's quite a huge PR). Will continue careful review of the solver later.

solver/src/solver/matcha_solver/api.rs
josojo: This costs more gas and I don't see a direct benefit. Though it might change in the future: 
http://help.matcha.xyz/en/articles/4217019-does-matcha-have-an-affiliate-program

If I am doing it, I would use the solver's address, I guess

josojo:
> > Why was for oneinch the pmm disabled? Do I need to do something similar for matcha?
> 
> We ran into weird issues with PMM1 (e.g. https://dashboard.tenderly.co/gp-v2/staging/simulator/8e1a49b8-2a8e-4d7f-b50c-83f93c5dfef1) calling a non-existent method on the closed source router. No need to do anything unless we see similar problems for match
> 
> > How did you guys test the solvers before? I just run it with an unregisted solver private key, and got:
> 
> You should be able to get the tenderly link, which would allow you to replace the sender with a real solver and simulate the transaction in that manner. Also, commenting out the submit_solution line in driver.rs and then testing locally with real orders worked for me. If they are available on rinkeby that could also be an option.
> 
> > In the current paraswap API, we will always sell "too many sell token" and hence continuously overpay paraswap, which is far from optimal.
> 
> I don't understand the concern here? How is selling too many sell tokens different from receiving too few buy tokens?
> 
> How does matcha deal with positive slippage in general?

Great! thanks for the tips!

>I don't understand the concern here? How is selling too many sell tokens different from receiving too few buy tokens?

Because you are always providing more sell-tokens. If you allow for slippage, you just receive less, in case you really get less from AMMs

solver/src/solver/matcha_solver.rs
```diff
@@ -0,0 +1,312 @@
+//! Module containing implementation of the Matcha solver.
+//!
+//! This simple solver will simply use the Matcha API to get a quote for a
```
bh2smith, line 3: Nit: "simple" and then "simply" Should pick only one.

solver/src/solver/matcha_solver.rs
```diff
@@ -0,0 +1,312 @@
...
+    encoding::EncodedInteraction,
+    interactions::Erc20ApproveInteraction,
+    liquidity::LimitOrder,
+    settlement::{Interaction, Settlement},
+};
+use model::order::OrderKind;
+use std::fmt::{self, Display, Formatter};
+
+/// Constant maximum slippage of 5 BPS (0.05%) to use for on-chain liquidity.
+/// This is halve the 1inch slippage
```
bh2smith, line 28:
```suggestion
/// This is half the 1inch slippage
```

solver/src/solver/matcha_solver.rs
```diff
@@ -0,0 +1,312 @@
...
+                    sell_token: order.sell_token,
+                    buy_token: order.buy_token,
+                    sell_amount: None,
+                    buy_amount: Some(order.buy_amount),
+                    slippage_percentage: Slippage::basis_points(STANDARD_MATCHA_SLIPPAGE_BPS)
+                        .unwrap(),
+                    // From the api documentation:
+                    // SlippagePercentage(Optional): The maximum acceptable slippage in % of the buyToken amount if sellAmount is provided, the maximum acceptable slippage in % of the sellAmount amount if buyAmount is provided. This parameter will change over time with market conditions.
+                    // => Hence, allegedly, we don't need to build in buffers ourselves. Though the sell amount is not adjusted, if slippage is changed for buy order requests.
+                    // Todo: Continue discussion with the 0x-team
```
bh2smith, line 108: Do you mean the matcha team? I don't see how 0x is relevant here.

solver/src/solver/matcha_solver.rs
```diff
@@ -0,0 +1,312 @@
...
+                tracing::debug!("proposed Matcha swap is {:?}", swap);
+
+                ensure!(
+                    swap.sell_amount <= order.sell_amount,
+                    "order limit price not respected",
+                );
+                let settlement = Settlement::new(hashmap! {
+                    order.sell_token => swap.sell_amount,
+                    order.buy_token => swap.buy_amount,
+                });
```
bh2smith, line 123: This block seems very similar to the one above for sell orders. Although the inequality is different, I wonder if it could somehow be pulled out and only exist once.

solver/src/solver/matcha_solver.rs
```diff
@@ -0,0 +1,312 @@
...
+        })
+    }
+}
+impl<F> MatchaSolver<F> {
+    fn web3(&self) -> DynWeb3 {
+        self.settlement_contract.raw_instance().web3()
+    }
+}
+
+impl<F> std::fmt::Debug for MatchaSolver<F> {
```
bh2smith, line 62: I see that we have implemented display and debug both the same for this solver. Makes me wonder if we could just get away with one. Is there anything about using display for debug if display is defined. Like a fallback for the better one in the hierarchy.

solver/src/solver/matcha_solver/api.rs
```diff
@@ -0,0 +1,227 @@
...
+
+impl SwapQuery {
+    /// Encodes the swap query as
+    fn into_url(self, base_url: &Url) -> Url {
+        // The `Display` implementation for `H160` unfortunately does not print
+        // the full address and instead uses ellipsis (e.g. "0xeeee…eeee"). This
+        // helper just works around that.
+        fn addr2str(addr: H160) -> String {
+            format!("{:?}", addr)
+        }
```
bh2smith, line 45: Although I'm not sure where at the moment, I'm fairly certain that there is already some work around for this somewhere in the project. We should probably try and stick to using just one throughout.

bh2smith (approved):
Looks pretty good overall. Nice work. I have made a few nits, comments and suggestions.

solver/src/solver/matcha_solver.rs
josojo: 0x and matcha are the same teams. matcha is developed by 0x :)

solver/src/solver/matcha_solver.rs
josojo: Yeah, you are right. though I could not come up with a good solution. The tricky part is that the ensure in the middle of the code is different

bh2smith: Since the only difference is the ensure statement, it might be less code to match on the order type again outside this block just for the ensure statement. Or even make a helper function for the inequality checking that the ensure statement calls into. Just a thought might look a lot less repetitive and then you could have an ensure(limit price respected) which looks kinda cool.

fleupold:
> Because you are always providing more sell-tokens.

But we would also receive more buy tokens in case the slippage isn't needed, right (I mean matcha doesn't keep slippage for sell orders but pays it back for buy orders, do they)? Is your worry that we will always need a buffer in the sell token, where on the buy token side we only need a buffer if the slippage tolerance gets used?

One reason for why we are providing slippage on the sell token side e.g. for uniswap is that we know we will have some buffer in the sell token already (because of fees). But it's true that this could be a problem for very large sell orders (e.g. for $1M  0.1% ends up being a large buffer).

solver/src/solver/matcha_solver/api.rs
fleupold: > This costs more gas

Does it cost more gas outside of the 20 bytes non-zero calldata? This would be a very marginal overhead. On the other hand, we can argue that since our orders are always executed in a specific manner we can compute our affiliation with Dune later on as well.

solver/src/solver/matcha_solver/api.rs
```diff
@@ -0,0 +1,230 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    #[ignore]
+    async fn test_api_e2e() {
```
fleupold, line 143: Maybe test both sides e2e (buy and sell order?)

My paraswap implementation had an issue on the buy order side which I could have caught with such a test

solver/src/solver/solver_utils.rs
```diff
@@ -0,0 +1,86 @@
...
+
+pub fn deserialize_decimal_f64<'de, D>(deserializer: D) -> Result<f64, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let decimal_str = Cow::<str>::deserialize(deserializer)?;
+    decimal_str.parse::<f64>().map_err(D::Error::custom)
+}
+
+pub fn deserialize_decimal_u256<'de, D>(deserializer: D) -> Result<U256, D::Error>
```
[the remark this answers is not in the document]
fleupold: Still missing one occasion in order to be able to remove this method altogether?

solver/src/solver/solver_utils.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+pub fn deserialize_decimal_f64<'de, D>(deserializer: D) -> Result<f64, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let decimal_str = Cow::<str>::deserialize(deserializer)?;
```
fleupold, line 66: Why do we need Cow here?

fleupold (changes requested):
Looks good overall, great to have you back in the codebase! I believe there might be an issue with the clearing prices, but happy to be convinced otherwise by a unit test :-) 

The rest are just nits

solver/src/solver/solver_utils.rs
nlordell: Coincidence of wants of cource :stuck_out_tongue:.

Joking aside, I had in the past had issues with `visit_str` vs `visit_borrowed_str` vs `visit_string` and `deserialize_str` vs `deserialize_string` with deserialization with unexpected results, and using `Cow` (which can be either a `Borrowed` or `Owned` string) solved the issue.

However, looking at the documentation, it seems to suggest that this shouldn't be an issue - so maybe we should just stick with `str` and see if there is an issue?

solver/src/solver/matcha_solver.rs
```diff
@@ -0,0 +1,441 @@
...
+
+                tracing::debug!("querying Matcha swap api with {:?}", query);
+                let swap = self.client.get_swap(query).await?;
+                tracing::debug!("proposed Matcha swap is {:?}", swap);
+
+                if swap.buy_amount < order.buy_amount {
+                    tracing::debug!("Order limit price not respected");
+                    return Ok(None);
+                }
+                let settlement = Settlement::new(hashmap! {
```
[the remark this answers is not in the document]
nlordell: Agree, the way I rubber duck it is to say "USD is worth less than ETH so it should have a lower price value".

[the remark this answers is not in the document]
josojo: Nice catch

solver/src/solver/matcha_solver.rs
```diff
@@ -0,0 +1,441 @@
...
+            OrderKind::Buy => {
+                let query = SwapQuery {
+                    sell_token: order.sell_token,
+                    buy_token: order.buy_token,
+                    sell_amount: None,
+                    buy_amount: Some(order.buy_amount),
+                    slippage_percentage: Slippage::basis_points(STANDARD_MATCHA_SLIPPAGE_BPS)
+                        .unwrap(),
+                    // From the api documentation:
+                    // SlippagePercentage(Optional): The maximum acceptable slippage in % of the buyToken amount if sellAmount is provided, the maximum acceptable slippage in % of the sellAmount amount if buyAmount is provided. This parameter will change over time with market conditions.
```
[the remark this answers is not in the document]
josojo: >This parameter will change over time with market conditions.
Not that I know what they mean, but I think their front-end queries this API with different slippage parameters depending on market conditions.

>"maximum acceptable slippage in % of the sellAmount" so it would adjust the sell amount?
Yes,  I was expecting the same! But I tested it and they don't do it. I will dig into it a little bit more on Monday

solver/src/solver/matcha_solver/api.rs
josojo: >Does it cost more gas outside of the 20 bytes non-zero calldata? 

No, but I thought I would not use any overhead, if there is no benefit. Anyways, I can set it as you guys want...

solver/src/solver/matcha_solver/api.rs
josojo: I have the two tests that are very similar, just a little bit bigger:

-> solve_sell_order_on_matcha 
-> solve_buy_order_on_matcha

I think it is sufficient.

solver/src/solver/solver_utils.rs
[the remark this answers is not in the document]
josojo: Yeah, since you mentioned that the PR is so big, I did not want to also refactor the 1inch solver. It is used over there.

josojo:
>But we would also receive more buy tokens in case the slippage isn't needed, right (I mean matcha doesn't keep slippage for sell orders but pays it back for buy orders, do they)? Is your worry that we will always need a buffer in the sell token, where on the buy token side we only need a buffer if the slippage tolerance gets used?

My worry is that paraswap is keeping positive slippage ( at least 50%) and we are always providing a bad price (because of slippage buffer) to the interface. Hence, in the usual case, when no slippage is needed, we effectively pay them.

I think that Matcha is just allows a slippage in the buy amount for buy orders, but I am still testing it.

fleupold:
> My worry is that paraswap is keeping positive slippage

In that case, when we add the slippage on the buy side and provide a "too low" receiveAmount by default, why would they send us all the proceeds back (and not just what we asked for)? Don't they go via a router contract, in which case I'd think it is as easy for them to take slippage from the sell amount as from the buy amount.

josojo:
> In that case, when we add the slippage on the buy side and provide a "too low" receiveAmount by default, why would they send us all the proceeds back (and not just what we asked for)? Don't they go via a router contract, in which case I'd think it is as easy for them to take slippage from the sell amount as from the buy amount.

Yes, I am not saying I see a solution to the paraswap problem. I am just saying that we will overpay them many times - if no slippage is needed. Matcha deals better with that by not charging anything. I think Matcha is best integration as it hands over positive slippage and seems to deal well slippage in general.


I tested the current implementation with Felix suggested method: Simulating the proposed tenderly tx with a real solver address and seeing them working.

As a next step I will rebase this PR on Nick's PR

josojo:
I think the PR is ready for another round of review.

Regarding the matcha buyOrders: 
[Here](https://dashboard.tenderly.co/gp-v2/staging/simulator/new?block=12727448&blockIndex=0&from=0xa6ddbd0de6b310819b49f680f65871bee85f517e&gas=8000000&gasPrice=0&value=0&contractAddress=0x3328f5f2cecaf00a2443082b657cedeaf70bfaef&rawFunctionInput=0x13d79a0b000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000e00000000000000000000000000000000000000000000000000000000000000140000000000000000000000000000000000000000000000000000000000000036000000000000000000000000000000000000000000000000000000000000000020000000000000000000000001f8a626883d7724dbd59ef51cbd4bf1cf2016d13000000000000000000000000dd1ad9a21ce722c151a836373babe42c868ce9a40000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000003515857bbc28d2302000000000000000000000000000000000000000000000025f273933db57000000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000b2938bd4c7854c0709c2fefefe1284fb24133d190000000000000000000000000000000000000000000000035231fe21ae807718000000000000000000000000000000000000000000000025f273933db57000000000000000000000000000000000000000000000000000000000000060dac6460000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000002aaf8919c302140000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000003515857bbc28d230200000000000000000000000000000000000000000000000000000000000001600000000000000000000000000000000000000000000000000000000000000041b3b89376df50e9ab976cae507db8a1944cb85464e7af28bc9b6ac1be1780b9b00bb5873b3bc592fbf17ff07e5c14f50f3269df5e3422d72aabbca47c4ca08ffa1c000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000003a00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000120000000000000000000000000dd1ad9a21ce722c151a836373babe42c868ce9a4000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000044095ea7b3000000000000000000000000def1c0ded9bec7f1a1670819833240f027b25effffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff00000000000000000000000000000000000000000000000000000000000000000000000000000000def1c0ded9bec7f1a1670819833240f027b25eff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000148d9627aa400000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000037bcff5b85914318f000000000000000000000000000000000000000000000025f273933db570000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000003000000000000000000000000dd1ad9a21ce722c151a836373babe42c868ce9a4000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc20000000000000000000000001f8a626883d7724dbd59ef51cbd4bf1cf2016d13869584cd000000000000000000000000100000000000000000000000000000000000001100000000000000000000000000000000000000000000009933253ad560dac1ca0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000&network=1) is one example:
Matcha solver uses the buyAmount offchain to calculate a reasonable sell_amount, but on-chain buy orders and sell orders are treated exactly the same, as you can see from the tx - more STAK tokens are bought than needed, as positive slippage occurred - no refunds or trading back.
This means that negative slippage is also tolerated in the buy amount for buy orders, though it will not be tolerated by our smart contract, and hence, there needs to be a buffer to make the tx pass. 

This is kinda risky, as there is no guarantee that we have buffers on our settlement contract available. Given that the paraswap solver builds in buffers into the request, I think it's reasonable to have the matcha-solver without buffers, just allowed slippage. IMO, in many situations, it's good to have the matcha setup - as we and users don't overpay -, but in many others, it's better to have paraswap config. Especially once we can mine tx via flashbots, Matcha setup will probably be more valuable. 

One thought is to have the matcha solver with buffers on sellAmount and the paraswap without buffers on sellAmount for buy orders, as matcha does not charge positive slippage, but if we really want this, then I would go for it in another PR.

PS: Most other dex ags route always through their dex.ag contract. Matcha does not do that. [Sometimes](https://dashboard.tenderly.co/gp-v2/staging/simulator/ec86d31c-7063-4162-ad52-6224771635d6), I just got normal settlements against uniswap. Very cool! Saves quite some gas. Because of that, matcha solver will probably win most dex-ag competitions in our backend. Looking forward to the stats.

fleupold:
> Given that the paraswap solver builds in buffers into the request, I think it's reasonable to have the matcha-solver without buffers, just allowed slippage. 

Just for me to understand the difference between the Matcha and the Paraswap solver
- Matcha **always** sells more than it needs to (receives refund in buy token). This is good for small amounts (since we always have a buffer in sell token) but can be a problem for large orders where slippage >> fee
- Paraswap sells more than it needs to for buy orders (gets **refund in sell token**), but is fine buying less than it wants to for sell orders (refund in sell token in case of positive slippage)

So both solvers can potentially run into **problems with large buy order** (refund is paid at the end, so balance can temporarily be negative, which would cause a revert). 

As the frontend is moving to tight price estimates this may become an issue (the baseline solver may no longer match orders). Maybe we can cap the slippage tolerance at the received fee? This way we would still be able to trade a large amount if the slippage is not needed (not blocking this PR though).

Here is an example paraswap buy order with a $3.65 refund: https://etherscan.io/tx/0xb7cee8551289fb3a9e92b8c97dc4d66b78a6807747540fc66444527ef64fd325

Here is a paraswap sell order with positive slippage refunded in buy token (YFI): https://etherscan.io/tx/0x4472c8033e22de3ae4042590aa3e1285d48ad0a4ec950ef6abba0f2a2f511994

josojo:
> Just for me to understand the difference between the Matcha and the Paraswap solver
> 
> * Matcha **always** sells more than it needs to (receives refund in buy token). This is good for small amounts (since we always have a buffer in sell token) but can be a problem for large orders where slippage >> fee
> * Paraswap sells more than it needs to for buy orders (gets **refund in sell token**), but is fine buying less than it wants to for sell orders (refund in sell token in case of positive slippage)
> 
> So both solvers can potentially run into **problems with large buy order** (refund is paid at the end, so balance can temporarily be negative, which would cause a revert).

Yees, this is correct. But since we get only a 50% refund in paraswap, we will always loose 50% of the overbought amount. That sucks! Also the repayments of 3.6 USDs can easily get bigger than the gas costs. I don't consider this a good solution for buy orders from paraswap.

In this (https://dashboard.tenderly.co/gp-v2/staging/simulator/new?block=12735030&blockIndex=0&from=0xa6ddbd0de6b310819b49f680f65871bee85f517e&gas=8000000&gasPrice=0&value=0&contractAddress=0x3328f5f2cecaf00a2443082b657cedeaf70bfaef&rawFunctionInput=0x13d79a0b000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000e0000000000000000000000000000000000000000000000000000000000000014000000000000000000000000000000000000000000000000000000000000003600000000000000000000000000000000000000000000000000000000000000002000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2000000000000000000000000dac17f958d2ee523a2206206994597c13d831ec7000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000029143e200000000000000000000000000000000000000000000000000470de4df820000000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000036416d81e590ff67370e4523b9cd3257aa0a853c000000000000000000000000000000000000000000000000000000000291f64800000000000000000000000000000000000000000000000000470de4df8200000000000000000000000000000000000000000000000000000000000060dc5839000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000003dc140000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000029143e2000000000000000000000000000000000000000000000000000000000000016000000000000000000000000000000000000000000000000000000000000000410a7f27a6638cc9cdaba8266a15acef4cf7e1e1c9b9b2059391b7230b67bdfeb21f1d3aa45852f527a5040d3d7a190b92764a2c854f334b7eed579b390b85fd3f1b000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000003800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000120000000000000000000000000dac17f958d2ee523a2206206994597c13d831ec7000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000044095ea7b3000000000000000000000000def1c0ded9bec7f1a1670819833240f027b25effffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff00000000000000000000000000000000000000000000000000000000000000000000000000000000def1c0ded9bec7f1a1670819833240f027b25eff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000128d9627aa400000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000002b220e100000000000000000000000000000000000000000000000000470de4df82000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000000000000000000000dac17f958d2ee523a2206206994597c13d831ec7000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2869584cd000000000000000000000000100000000000000000000000000000000000001100000000000000000000000000000000000000000000003239e38b8a60dc53b70000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000&network=1)[tx] one can see how matcha is adding the 5 percent on top of the sell amount and then just refunds the 5% buyTokens of a buy order.

>Maybe we can cap the slippage tolerance at the received fee?
Yes, I like the suggestions. Or should we do the cap at received fee + buffers within the settlement contract?
This would allow us to settle more orders, but it could also eat into our buffers more frequently.
I will go for one suggestion in a follow-up PR.

fleupold:
> But since we get only a 50% refund in paraswap, we will always loose 50% of the overbought amount.

True! Another option would be to price slippage into the computed exchange rate we offer to the user. So instead of announcing price `p` in the solution but accepting price `p * 1.001` in the Paraswap interaction, we could announce the same price to the user (or even price in the expected 50% slippage and announce `p*1.0005`). This way we should "on average" not lose money.

>In this tx one can see how matcha is adding the 5 percent on top of the sell amount and then just refunds the 5% buyTokens of a buy order.

I have also seen simulations of the Paraswap solver (e.g. [this one](https://dashboard.tenderly.co/gp-v2/staging/simulator/26b31d6e-d733-43e7-a56b-39c5bf6e9751)) where positive slippage is not kept. However, this tends to happen for "simple" uniswap trades which would be caught by our baseline or MIP solver anyways. Will be interesting to see if Matcha behaves the same for complex routes.

solver/src/solver/matcha_solver.rs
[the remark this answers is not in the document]
fleupold: Can you clean up this comment please? I think it's such an important consideration that it should go on top of the method (rather than in one of the branches). Maybe you can re-summarize your observations (potentially with links to the simulations which I found very useful). I personally find their documentation quote horrible to understand.

solver/src/solver/matcha_solver.rs
```diff
@@ -0,0 +1,439 @@
...
+
+        let allowance_target = shared::addr!("def1c0ded9bec7f1a1670819833240f027b25eff");
+        client.expect_get_swap().returning(move|_| {
+            Ok(SwapResponse {
+            sell_amount: U256::from_dec_str("100").unwrap(),
+             buy_amount: U256::from_dec_str("91").unwrap(),
+             allowance_target,
+            price: 0.91_f64,
+            to: shared::addr!("def1c0ded9bec7f1a1670819833240f027b25eff"),
+            data: web3::types::Bytes(hex::decode(
```
fleupold, line 237: nit: to make this test easier to read, I'd maybe use default or dummy values wherever possible.

solver/src/solver/matcha_solver.rs
```diff
@@ -0,0 +1,439 @@
...
+                sell_token => 91.into(),
+                buy_token => 100.into(),
+            }
+        );
+
+        let result = solver.settle_order(order_violating_limit).await.unwrap();
+        assert!(result.is_none());
+    }
+    #[tokio::test]
+    async fn test_satisfies_limit_price_for_buy_order() {
```
fleupold, line 293: nit: I'd be for combining the two limit order tests (buy & sell) side if that leads to significantly less code.

solver/src/solver/matcha_solver/api.rs
fleupold: `{:#x}` (cf. https://doc.rust-lang.org/std/fmt/trait.LowerHex.html) is better imo, since debug print is not really guaranteed to forever print hex (it does today, but that could change, lowerHex is guaranteed to stay)

solver/src/solver/solver_utils.rs
fleupold: @josojo can you comment why cow is needed or address this suggestion?

fleupold (approved):
Looks good to me now, jus some minor nits (I think we should start carefully documenting the slippage behaviour)

solver/src/solver/solver_utils.rs
josojo: I just copied this line from the existing code in 1inch. Since I don't know in which issues nlordell was running, I don't know whether they will reoccur. Hence, I would just not change it, as it runs currently.

josojo:
Thanks for the really detailed and good feedback. Especially catching the price error! Thanks, Felix.

I am now merging and watching the logs :)

solver/src/solver/solver_utils.rs
nlordell: So I did some investigation because I was curious and:
```rust
println!("{:?}", <&str>::deserialize(json!("asdf"))); // Err(Error("invalid type: string \"asdf\", expected a borrowed string", line: 0, column: 0))
```

This is because:
> The reason serde can't deserialize into a borrowed string is that it needs to modify (unescape) the encoded json, which requires ownership.
>
> That said, it appears you can get around this by using a `Cow<'a, str>` allowing deserialization to borrowed or owned memory depending on the string itself.

PR #795, opened 2021-06-28
Make WeightedProductOrder implement BaselineSolvable

nlordell:
This PR implements `BaselineSolvable` for the new `WeightedProductOrder` so that the `Liquidity` type can be used directly by the baseline solver.

### Test Plan

Added some unit tests for the new BigRational <-> Bfp conversion logic. Otherwise, the rest was just "wiring" up implementations together and doesn't really need to be tested beyond compilation.

### DIFF
--- a/solver/src/solver/baseline_solver.rs
+++ b/solver/src/solver/baseline_solver.rs
@@ -1,22 +1,28 @@
+use crate::{
+    liquidity::{
+        AmmOrderExecution, ConstantProductOrder, LimitOrder, Liquidity, WeightedProductOrder,
+    },
+    settlement::Settlement,
+    solver::Solver,
+};
 use anyhow::Result;
 use ethcontract::{H160, U256};
 use maplit::hashmap;
 use model::TokenPair;
 use num::BigRational;
 use shared::{
+    balancer::swap::{fixed_point::Bfp, WeightedPoolRef},
     baseline_solver::{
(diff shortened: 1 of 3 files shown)


solver/src/solver/baseline_solver.rs
```diff
@@ -51,6 +56,32 @@ impl BaselineSolvable for ConstantProductOrder {
     }
 }
 
+impl BaselineSolvable for WeightedProductOrder {
```
bh2smith, line 59: Do we need `BaselineSolvable` for both `WeightedPool` and `WeightedProductOrder`? I see how we do it by converting it and using the other, but it seems a bit redundant.

bh2smith (approved):
Way cool!

solver/src/solver/baseline_solver.rs
nlordell: The idea here is that in the `orderbook` price estimates are done using `WeightedPool` directly fetched with a `WeigthedPoolFetching` instance. In the solver, however, we convert everything into a `Liquidity::WeightedProduct` and use that in the baseline solver. For this reason, both need to implement `BaselineSolvable` since both are used by the baseline solving code.

We also do this with `Pool` and `ConstantProductOrder`.

josojo (approved):
LGTM, thanks for adding quite some unit tests

vkgnosis (approved):

PR #798, opened 2021-06-28
Move baseline liquidity sources to submodule

nlordell:
In preparation for including Balancer pools into the fetched baseline liquidity, this PR just moves a bunch of Uniswap specific baselince liquidity source abstractions into a Uniswap specific module. Additionally, all baseline liquidity-source specific abstractions are now in `shared::sources` module.

In a follow up PR, I plan on merging `PoolFetching` and `WeightedPoolFetching` so that Balancer pools and Uniswap-like pools can be aggregated with the `PoolAggregator` that gets created in `main`, making the Balancer liquidity finally be part of the fetched baseline liquidity.

Additionally, I want to change `AmmPoolFetching` into something more generic so that Balancer pools can also be used by the bad token detector. Specifically, it needs to be changed to allow support multiple pools per token pair.

### Test Plan

Only changed code are `use` statements. Everything else is just code that got moved around.

Verify that the moved `WeightedPool` struct and code didn't change:
```
diff \
  <(git show pool-fetching-refactor:shared/src/sources/balancer/pool_fetching.rs | head -69 | tail -43) \
  <(git show main:shared/src/balancer/pool_storage.rs | head -83 | tail -43)
```


shared/src/sources/balancer/pool_fetching.rs
```diff
@@ -3,21 +3,70 @@
...
+    pub scaling_exponent: u8,
+}
+
+#[derive(Clone)]
+pub struct WeightedPool {
+    pub pool_id: H256,
+    pub pool_address: H160,
+    pub swap_fee_percentage: U256,
+    pub reserves: HashMap<H160, PoolTokenState>,
+}
```
bh2smith, line 40: These structs make sense to have here! Good move.

bh2smith (approved):

vkgnosis (approved):

PR #804, opened 2021-06-30
[Easy] Update or reset inflight trades based on submission result

### LINKED ISSUE #791: Orders are added to inflight trades even if settlement is not successful
https://github.com/gnosis/gp-v2-services/blob/6a500707a041eea0bc178618eb05fcf84fcdb09c/solver/src/driver.rs#L434-L442

This means we won't attempt to resettle orders in the next runloop if their execution has failed. Moreover, if the next runloop doesn't produce any solution, we won't enter the `if` statement in which `inflight_order` would be reset. Thus it is possible to completely "ignore" this order going forward.

To fix this `submit_settlement` should probably return information as to whether the settlement was successful and inflight trades should be reset unconditionally.

bh2smith:
Closes #791

As described in the related issue, inflight trades could be populated with orders even if submission fails and also not reset if the next run loop fails to find a solution. This PR makes the update of inflight trades dependent on the submission Result.

### Test Plan
Existing CI.

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -417,6 +417,8 @@ impl Driver {
         let rated_settlements = self
             .rate_settlements(settlements, &estimated_prices, gas_price_wei)
             .await;
+
+        self.inflight_trades.clear();
         if let Some(mut settlement) = rated_settlements
             .clone()
             .into_iter()
@@ -441,8 +443,6 @@ impl Driver {
                     .iter()
                     .map(|t| t.order.order_meta_data.uid)
                     .collect::<HashSet<OrderUid>>();
-            } else {
-                self.inflight_trades = HashSet::new()
             }
 
             self.report_matched_but_unsettled_orders(
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -137,7 +137,7 @@ impl Driver {
         .into_iter()
     }
 
-    async fn submit_settlement(&self, rated_settlement: RatedSettlement) {
+    async fn submit_settlement(&self, rated_settlement: RatedSettlement) -> Result<()> {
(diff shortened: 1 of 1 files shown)


solver/src/driver.rs
```diff
@@ -431,14 +433,17 @@ impl Driver {
...
+                self.inflight_trades = settlement
+                    .settlement
+                    .settlement
+                    .trades()
+                    .iter()
+                    .map(|t| t.order.order_meta_data.uid)
+                    .collect::<HashSet<OrderUid>>();
+            } else {
+                self.inflight_trades = HashSet::new()
+            }
```
vkgnosis, line 446: Instead of creating a new hash set every time we can reuse the existing one which is more efficient.
```suggestion
            self.inflight_trades.clear();
            if self.submit_settlement(settlement.clone()).await.is_ok() {
                self.inflight_trades.extend(
                    settlement
                        .settlement
                        .settlement
                        .trades()
                        .iter()
                        .map(|t| t.order.order_meta_data.uid),
                );
            }
```

vkgnosis (approved):

vkgnosis (approved):

solver/src/driver.rs
fleupold: Maybe the clear should even happen outside of the `if let Some(mut settlement) = rated_settlements` statement? This way we reset them even if there is no new solution found.

fleupold (approved):
This logic change would be nice to accompany with a unit test, if that's possible

bh2smith:
> This logic change would be nice to accompany with a unit test, if that's possible

I considered this but I don't believe it was tested in the first place, likely because it will require quite and elaborate setup, but I can give it a try.

PR #806, opened 2021-07-01
Paraswap Error handling

### LINKED ISSUE #802: [Paraswap] Handle "price changed" error
This would fix alerts like these:

2021-06-28T21:53:23.481Z ERROR solver::solver::single_order_solver: Inner solver error: TransactionBuilderQuery result parsing failed: {"error":"It seems like the rate has changed, please re-query the latest Price"}

The paraswap solver 1) queries a price estimate and 2) if the estimate satisfies the limit price requests a transaction object for this price. The ☝️ error is shown if the price moved between the two requests.

In this case we should probably just retry solving the order instead of bubbling this error up.

bh2smith:
A first step toward #802 while simultaneously giving more informative/useful logs when other paraswap errors occur (e.g. logging query only on failure like "ERROR_BUILDING_TRANSACTION")

### Test Plan

Added a new test `paraswap_response_handling` which verifies 4 of the Error types and 1 ignored test which tests the `Ok()` path. Unfortunately I am not sure at the moment how to mock and test the `reqwest_errors`

```
    // Connectivity or non-response error
    #[error("Failed on send")]
    Send(reqwest::Error),

    // Recovered Response but failed on async call of response.text()
    #[error(transparent)]
    TextFetch(reqwest::Error),
```


solver/src/driver.rs
```diff
@@ -267,11 +267,11 @@ impl Driver {
                 "{} settlement simulation failed at submission and block {}:\n{:?}",
                 settlement.name,
                 current_block_during_liquidity_fetch,
-                error_at_earlier_block
+                error_at_earlier_block,
             );
             // This is an additional debug log so that the log message doesn't get too long as
             // settlement information is recoverable through tenderly anyway.
-            tracing::warn!("settlement failure for: \n{:#?}", settlement,);
+            tracing::warn!("settlement failure for: \n{:#?}", settlement);
```
bh2smith, line 274: unrelated OCD comma changes (U.OCD.C.Cs)

solver/src/solver/paraswap_solver/api.rs
```diff
@@ -45,24 +46,76 @@ impl ParaswapApi for DefaultParaswapApi {
...
+    // Connectivity or non-response error
+    #[error("Failed on send")]
+    Send(reqwest::Error),
+
+    // Recovered Response but failed on async call of response.text()
+    #[error(transparent)]
+    TextFetch(reqwest::Error),
+
+    #[error(transparent)]
+    Other(#[from] anyhow::Error),
```
vkgnosis, line 88: If we are using thiserror do distinguish all the other errors anyway then let's also do that here

```suggestion
    #[error("unknown paraswap error: {0}")]
    UnknownParaswapError(String)

    #[error(transparent)]
   DeserializeError(#[from] serde_json::Error)
```

vkgnosis (approved):
More detailed errors look good. Does it fix the issue mentioned in the description already? It should retry if price changed, right?

bh2smith:
> More detailed errors look good. Does it fix the issue mentioned in the description already? It should retry if price changed, right?

There is another follow up PR on top of this desiccated specifically to the issue. I will update the description here.

PR #807, opened 2021-07-02
Add archerdao transaction api

vkgnosis:
This is just the low level http api and not yet used. Still working on integrating it with the settlement process.

### Test Plan
Just like non archer transactions this has to be tested by making a real transactions so there is no automated test for it. I have tested it manually and will test again when the full settlement is integrated.


solver/src/settlement_submission/archerapi.rs
```diff
@@ -0,0 +1,76 @@
...
+        raw_signed_transaction: &[u8],
+        deadline: SystemTime,
+    ) -> Result<()> {
+        let tx = format!("0x{}", hex::encode(raw_signed_transaction));
+        let deadline = deadline
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap()
+            .as_secs()
+            .to_string();
+        let body = serde_json::json!({
```
nlordell, line 35: Should we maybe log request and response bodies to make debugging easier in case we run into issues?

nlordell (approved):
:+1:

josojo (approved):
Nice!

solver/src/settlement_submission/archerapi.rs
bh2smith: I would be in favour of logging the request and response together at once but only in the case that the response is an Error.

bh2smith (approved):

solver/src/settlement_submission/archerapi.rs
vkgnosis: This code includes the response body in the error message. I am hesitant about the request body because it contains the authorization string which should be secret and the full transaction bytes are fairly long and not interesting (or are they?).

nlordell: > This code includes the response body in the error message.

Nice. For some reason I missed that. I think that's fine, so we don't have to log the response every time.

> I am hesitant about the request body because it contains the authorization string

From the code, I am under the impression it is included in the header and not the request body.

> the full transaction bytes are fairly long and not interesting (or are they?).

Hmm, that's a good question. In theory we log this in our transport implementation, so it might help when debugging issues. In my opinion, it would be nice to log the request JSON as `debug` and make this module only print info by default like with the HTTP transport in `shared::arguments::Arguments.log_filter`.

vkgnosis: You're right the authorization is only in the header. Added the debug log and set default log filter to info.

PR #808, opened 2021-07-03
[HTTP solver] UpdatedAmmModel to new format

bh2smith:
In correspondence with dex solver issue (https://github.com/gnosis/gp-v2-solver/issues/165) we adapt the incoming execution format from the HTTP solver. Rather than have HashSet<H160, i128> (that is a map of token address to signed integer executions) we have four fields (buy/sell token along with their non-negative execution amounts). This clean up some of the logic of determining which token is buy and sell based on the execution amount's signature and is overall more practical.

Note that when creating inputs and outputs from this info I noticed that it had to be intuitively backwards in order to pass the existing unit test. We should double check this and I will highlight the part in the code.


solver/src/solver/http_solver/settlement.rs
```diff
@@ -128,36 +128,11 @@ fn match_prepared_and_settled_amms(
...
-            } else {
-                return Err(anyhow!("invalid uniswap update {:?}", settled));
-            };
             Ok(ExecutedAmm {
                 order: prepared,
-                input,
-                output,
+                // This seems backwards, but its how we passed the test.
+                input: (settled.buy_token, settled.exec_buy_amount),
+                output: (settled.sell_token, settled.exec_sell_amount),
```
bh2smith, line 135: I would have expected input to be the sell related stuff here and output to be buy-related stuff. However, this is what passes the unit test.

solver/src/solver/http_solver/settlement.rs
```diff
@@ -239,7 +203,10 @@ mod tests {
             exec_sell_amount: 7.into(),
         };
         let updated_uniswap = UpdatedAmmModel {
-            updates: hashmap! { t0 => 8, t1 => -9 },
```
bh2smith, line 242: I interpreted this as t1 is sell token (because of negative 9 here). Maybe I was incorrect in my assumption.

solver/src/solver/http_solver/settlement.rs
```diff
@@ -128,36 +128,11 @@ fn match_prepared_and_settled_amms(
...
-                    (tokens.0, i128_abs_to_u256(updates.0)),
-                )
-            } else {
-                return Err(anyhow!("invalid uniswap update {:?}", settled));
-            };
             Ok(ExecutedAmm {
                 order: prepared,
-                input,
-                output,
+                // This seems backwards, but its how we passed the test.
```
bh2smith, line 133:
```suggestion
```

solver/src/solver/http_solver/settlement.rs
bh2smith: Double checked and confirmed with @marcovc.

nlordell (approved):

solver/src/solver/http_solver/settlement.rs
marcovc: This is modelled the point of view of an amm, where input=buy and output=sell.

PR #809, opened 2021-07-04
Solve with bal liquidity

nlordell:
This PR adds the Balancer V2 liquidity to the liquidity fetcher so that it is considered by the Baseline solver.

Currently, Balancer liquidity is limited to the baseline solver. There are other PRs to send the pools to the HTTP solver, and including it in price estimates will be done in a follow up PR.

One thing that you may note from this PR is that some of the pool abstractions (pool aggregator, AMM pair provider, etc.) are quite specific to Uniswap-like liquidity. Unfortunately, these are quite coupled with the `BaselineSource` type which is used to determine which on-chain liquidity we have access to, which in theory isn't limited to Uniswap-like pools. Additionally, I ran into quite a few issues for adding new liquidity, where there are a lot of places that need to have their own understanding of what "liquidity" is (baseline solver in `solver` crate, baseline solver in `shared` crate, price estimates, etc.). I plan on tackling some of this tech debt in following PRs.

### Test Plan

Start up a solver on Rinkeby using only the baseline solver and Balancer liquidity and watch it find an order going thought Balancer (note that Uniswap liquidity is still needed for price estimates):
```
cargo run -p solver -- --baseline-sources BalancerV2,Uniswap --solvers Baseline
```

Transaction for a successful baseline solver solution using Balancer liquidity: [0x0082537056dc6003c69b050323587c702c163425faca423b6c625f4236e4ef89](https://rinkeby.etherscan.io/tx/0x0082537056dc6003c69b050323587c702c163425faca423b6c625f4236e4ef89).

:warning: This actually requires #810 because of Balancer Rinkeby subgraph behaviour that I didn't account for when building the pool initialization component.


vkgnosis (approved):

orderbook/src/main.rs
```diff
@@ -186,13 +184,11 @@ async fn main() {
...
     );
 
-    let pair_providers: Vec<Arc<dyn AmmPairProvider>> = stream::iter(args.shared.baseline_sources)
-        .then(|source| {
-            let web3 = web3.clone();
-            async move { sources::pair_provider(source, chain_id, &web3).await }
-        })
-        .collect()
-        .await;
+    let pair_providers = sources::pair_providers(&args.shared.baseline_sources, chain_id, &web3)
```
bh2smith, line 187: It's pair providers still an appropriate name here (given that balancer is now incorporated as a liquidity source)?

shared/src/arguments.rs
```diff
@@ -12,7 +12,7 @@ pub struct Arguments {
     #[structopt(
         long,
         env = "LOG_FILTER",
-        default_value = "warn,orderbook=debug,solver=debug,shared=debug,shared::http_transport=info"
+        default_value = "warn,orderbook=debug,solver=debug,shared=debug,shared::transport::http=info"
```
bh2smith, line 15: Nice!

shared/src/sources.rs
```diff
@@ -10,36 +10,45 @@ use self::uniswap::{
...
+                    .await
+                    .expect("couldn't load deployed uniswap router"),
+                chain_id,
+            }),
+            BaselineSource::Sushiswap => Arc::new(SushiswapPairProvider {
+                factory: contracts::SushiswapV2Factory::deployed(web3)
+                    .await
+                    .expect("couldn't load deployed sushiswap router"),
+            }),
+            BaselineSource::BalancerV2 => continue,
```
bh2smith, line 47: Oh. It seems we ignore balancer pools here... so I guess pair provider's makes sense still.

solver/src/liquidity/balancer.rs
```diff
@@ -34,6 +43,24 @@ pub struct BalancerV2Liquidity {
...
 
 impl BalancerV2Liquidity {
+    pub async fn new(
+        web3: Web3,
+        pool_fetcher: Arc<dyn WeightedPoolFetching>,
+        base_tokens: HashSet<H160>,
+    ) -> Result<Self> {
+        let contracts = Contracts::new(&web3)
+            .await
+            .context("missing Balancer V2 contract deployement")?;
```
bh2smith, line 53: Isn't this more like "missing one of..."

bh2smith (approved):
Nice read. Sick flow bro.

orderbook/src/main.rs
nlordell: Balancer is skipped for pair providers (because as you mentioned it doesn't make sense for it). It is one of the APIs that are Uniswap specific that I want to tackle as tech debt this week:
> One thing that you may note from this PR is that some of the pool abstractions (pool aggregator, AMM pair provider, etc.) are quite specific to Uniswap-like liquidity. [...] I plan on tackling some of this tech debt in following PRs.

PR #810, opened 2021-07-04
Disable pool intiailzation for unsupported networks

nlordell:
This PR adds a `DefaultPoolInitializer` type that automatically determines based on the network whether or not to use the subgrph Balancer pool intializer implementation. This is required because the Balancer subgraph requires an archive node in order to index pools with complete token data, which is only available on Mainnet for the Graph hosted service.

This means that for networks for which the Balancer subgraph doesn't index all pool data, we will need to index all events :cry:. Luckily, this can be somewhat mitigated by using the subgraph just to query pool addresses, and then reading the required pool data on-chain with `eth_call`s. This is done in a follow up PR.

### Test Plan

Added unit tests. Additionally, this code is used in #809 to make a settlement trading against Balancer pools.

### DIFF
--- a/contracts/build.rs
+++ b/contracts/build.rs
@@ -28,9 +28,8 @@ fn main() {
             .add_deployment(
                 4,
                 addr("0xBA12222222228d8Ba445958a75a0704d566BF2C8"),
-                Some(tx(
-                    "0x5fe65a242760f7f32b582dc402a081791d57ea561474617fcd0e763c995cfec7",
-                )),
+                // <https://rinkeby.etherscan.io/tx/0x5fe65a242760f7f32b582dc402a081791d57ea561474617fcd0e763c995cfec7>
+                Some(DeploymentInformation::BlockNumber(8441702)),
             )
     });
     generate_contract_with_config("BalancerV2WeightedPoolFactory", |builder| {
--- a/shared/src/sources/balancer/pool_init.rs
+++ b/shared/src/sources/balancer/pool_init.rs
@@ -203,7 +203,7 @@ mod tests {
         let initializer = EmptyPoolInitializer(4);
         assert_eq!(
             initializer.initialize_pools().await.unwrap(),
-            BalancerRegisteredPools{
+            BalancerRegisteredPools {
                 fetched_block_number: 8441702,
                 ..Default::default()
             }
--- a/shared/src/sources/balancer/pool_init.rs
+++ b/shared/src/sources/balancer/pool_init.rs
@@ -3,13 +3,17 @@
 //! This module contains a component used to initialize Balancer pool registries
 //! with existing data in order to reduce the "cold start" time of the service.
 
+use std::sync::Arc;
+
 use crate::sources::balancer::{
-    graph_api::{BalancerSubgraphClient, RegisteredWeightedPools},
+    graph_api::{BalancerSubgraphClient, CreatedPools, RegisteredWeightedPools},
+    info_fetching::PoolInfoFetching,
     pool_storage::RegisteredWeightedPool,
 };
 use anyhow::{anyhow, Result};
 use contracts::{BalancerV2WeightedPool2TokensFactory, BalancerV2WeightedPoolFactory};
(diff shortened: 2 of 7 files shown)


shared/src/sources/balancer/pool_init.rs
```diff
@@ -27,12 +32,49 @@ pub trait PoolInitializing: Send + Sync {
...
+        const MAINNET_CHAIN_ID: u64 = 1;
+
+        Ok(match chain_id {
+            // Balancer subgraph seems to only correctly index pool info on
+            // chains where it supports archive nodes (because of the required
+            // `eth_call`s). This means we can only use the pure Subgraph
+            // initializer on Mainnet.
+            MAINNET_CHAIN_ID => {
+                DefaultPoolInitializer::Subgraph(SubgraphPoolInitializer::new(chain_id)?)
+            }
```
vkgnosis, line 65: I think this just binds chain_id to the constant and not check for equality which you probably intended.

vkgnosis (approved):

contracts/build.rs
```diff
@@ -28,9 +28,8 @@ fn main() {
             .add_deployment(
                 4,
                 addr("0xBA12222222228d8Ba445958a75a0704d566BF2C8"),
-                Some(tx(
-                    "0x5fe65a242760f7f32b582dc402a081791d57ea561474617fcd0e763c995cfec7",
-                )),
+                // <https://rinkeby.etherscan.io/tx/0x5fe65a242760f7f32b582dc402a081791d57ea561474617fcd0e763c995cfec7>
+                Some(DeploymentInformation::BlockNumber(8441702)),
```
bh2smith, line 32: Is it more convenient to put the block number with a comment linking to the tx hash or just set this thing up with the tx hash as it's deployment info and fetch the rest. I see we are saving an eth call so that's nice.

shared/src/sources/balancer/pool_init.rs
```diff
@@ -123,8 +165,29 @@ fn deployment_address(artifact: &Artifact, chain_id: u64) -> Result<H160> {
...
+                "missing deployment information for {}",
+                artifact.contract_name
+            )
+        })?;
+
+    match deployment_info {
+        DeploymentInformation::BlockNumber(block) => Ok(block),
+        DeploymentInformation::TransactionHash(tx) => {
+            bail!("missing deployment block number for {}", tx)
+        }
```
bh2smith, line 189: I see what you've done here, but I really don't think we need to bail when give a transaction hash. Are you trying to avoid making this method async? That's the only reason I can imagine not filling out this code. 

I understand that it's easy enough to just include the deployment block as part of the artifact info, but I'm not sure why we chose to fail when it's not there.

bh2smith (approved):
Looks good. A bit confused about why we explicitly chose to fail when balancer artifacts have a deployment transaction but not explicitly provided with deployment block.

nlordell:
> A bit confused about why we explicitly chose to fail when balancer artifacts have a deployment transaction but not explicitly provided with deployment block.

Its just less code. And we aren't using the transaction hash for anything else, so I thought it made sense to just require a deployment block. Additionally, it would mean that we would have to make a `Web3` request that can potentially fail, as well as test those paths. I was actually just lazy, but this can be changed to support both if needed.

contracts/build.rs
nlordell: > I see we are saving an eth call so that's nice.

That's my main motivation. It also saves us one failure branch.

shared/src/sources/balancer/pool_init.rs
nlordell: > but I'm not sure why we chose to fail when it's not there.

It's just to avoid the extra call, failure case and tests. I don't think we really need to support both. If you feel strongly I can make it work with `TransactionHash` as well in a separate PR.

shared/src/sources/balancer/pool_init.rs
nlordell: Changed this to use an `if` statement instead so there is less ambiguity.

PR #811, opened 2021-07-04
Fetch initial weighted pool data on-chain

nlordell:
This PR adds an additional pool initializer based on on-chain token data, using only pool addresses, factory and type from the subgraph. This allows the pool initialization to be done without having to index historical events since the dawn of time for networks without archive node support (which causes Balancer pools to not have all the required token data).

This currently doesn't quite work and requires a change to the Balancer subgraph balancer-labs/balancer-subgraph-v2#91

### Test Plan

Added unit test for new pool initializer. Waiting to merge pending required subgraph change. We should then be able to trade with old pools on Rinkeby without event indexing.

Manual test: run the solver connected to a Rinkeby node and see that the pool initialization works :tada:
```
$ cargo run -p solver -- --baseline-sources BalancerV2 --solvers Baseline
...
2021-07-08T19:00:56.003Z DEBUG shared::sources::balancer::pool_init: initialized registered pools BalancerRegisteredPools ...
...
```

### DIFF
--- a/shared/src/sources/balancer/event_handler.rs
+++ b/shared/src/sources/balancer/event_handler.rs
@@ -17,11 +17,10 @@ use crate::{
     impl_event_retrieving,
     maintenance::Maintaining,
     sources::balancer::{
-        info_fetching::PoolInfoFetcher,
+        info_fetching::PoolInfoFetching,
         pool_init::PoolInitializing,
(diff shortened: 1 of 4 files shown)


vkgnosis (approved):

bh2smith (approved):
kewl!

nlordell:
Balancer subgraph has been updated and now returns the data we want!

PR #812, opened 2021-07-05
[Paraswap] retry until not Price Error

### LINKED ISSUE #802: [Paraswap] Handle "price changed" error

bh2smith:
Fixes #802

Changing the paraswap solver run loop to loop on the price route and transaction building until the transaction response does not result in `PriceChange` error introduced in the previous PR #806. I have hardcoded `max_retries`here to be 2 since I believe this to be completely sufficient for our purposes. As the list of retryable errors grows, we may want to consider making this configurable via runtime arguments.

### Test Plan
New unit test for ParaswapSolver retry logic included here using the all powerful mock interfaces.


solver/src/solver/paraswap_solver.rs
```diff
@@ -60,37 +63,80 @@ impl ParaswapSolver {
 #[async_trait::async_trait]
 impl SingleOrderSolving for ParaswapSolver {
     async fn settle_order(&self, order: LimitOrder) -> Result<Option<Settlement>> {
+        let transaction: TransactionBuilderResponse;
+        let price_response: PriceResponse;
+        let amount: U256;
+        loop {
```
fleupold, line 69: Should we bound this loop?

solver/src/solver/paraswap_solver.rs
```diff
@@ -111,43 +157,51 @@ impl SingleOrderSolving for ParaswapSolver {
...
-        Ok(Some(settlement))
+        Ok(query)
     }
 
-    fn name(&self) -> &'static str {
-        "ParaSwap"
+    pub async fn paraswap_transaction_from_order(
+        &self,
+        order: &LimitOrder,
+    ) -> Result<(
```
fleupold, line 177: This return type looks very complex. Should we maybe provide a struct for this?

fleupold (commented):
Can you please add a unit test for this new functionality (the api is already easily mockable)?

Regarding the design: I like that you wrapped the transaction logic into an inner function. In fact, I would find it easiest if a complete "single run" would be its own inner function (including all the logic we had before). This function would then return its own internal `ParaswapError` type (composed of a TransactionBuilderError, GeneralError) which can implement `should_retry` or the like (maybe in the future it could also be "should_ignore" in which case the error gets converted into a `None` object and we don't get an alert).

Then, `settle_order` just becomes invoking a single run in a loop and checking in case of an error if it should retry (and later ignore the error).

But I don't want to get in the way in case this change would be slowing you down and the rest of the team is happy with the suggested approach.

solver/src/solver/paraswap_solver.rs
bh2smith: Yea great idea. I would think at most 1 or 2 retries is enough... but then we will have to spit out the same PriceChange error if its still happening.

nlordell:
I think this PR needs to be rebased.

bh2smith:
> I think this PR needs to be rebased.

All rebased an ready to rock 🎸

solver/src/solver/paraswap_solver.rs
```diff
@@ -57,40 +60,95 @@ impl ParaswapSolver {
...
+                        return Err(anyhow!("Price change error after {} retries", num_retries));
+                    }
+                    num_retries += 1;
+                    continue;
+                }
+                _ => {
+                    price_response = complete_response.price_response;
+                    amount = complete_response.amount;
+                    transaction = complete_response.transaction_result?;
+                    break;
```
vkgnosis, line 94: Instead of this you could "return" from the loop:

```rust
let (price_response, amount, transaction) = loop {
...
                    break (
                        complete_response.price_response,
                        complete_response.amount,
                        complete_response.transaction_result?,
                    );
```

solver/src/solver/paraswap_solver.rs
```diff
@@ -57,40 +60,95 @@ impl ParaswapSolver {
...
+        Ok(Some(settlement))
+    }
+
+    fn name(&self) -> &'static str {
+        "ParaSwap"
+    }
+}
+
+impl ParaswapSolver {
+    pub async fn get_price_for_order(
```
vkgnosis, line 124:
```suggestion
    async fn get_price_for_order(
```

solver/src/solver/paraswap_solver.rs
```diff
@@ -57,40 +60,95 @@ impl ParaswapSolver {
...
         let price_response = self.client.price(price_query).await?;
+        Ok((price_response, amount))
+    }
 
-        if !satisfies_limit_price(&order, &price_response) {
-            tracing::debug!("Order limit price not respected");
-            return Ok(None);
-        }
-
+    pub async fn transaction_query_from(
```
vkgnosis, line 146:
```suggestion
    fn transaction_query_from(
```

solver/src/solver/paraswap_solver.rs
```diff
@@ -111,43 +169,47 @@ impl SingleOrderSolving for ParaswapSolver {
...
-                .await?,
-        );
-        settlement.encoder.append_to_execution_plan(transaction);
-        Ok(Some(settlement))
+        Ok(query)
     }
 
-    fn name(&self) -> &'static str {
-        "ParaSwap"
+    pub async fn paraswap_transaction_from_order(
```
vkgnosis, line 186:
```suggestion
    async fn paraswap_transaction_from_order(
```

vkgnosis (approved):

solver/src/solver/paraswap_solver.rs
vkgnosis, line 186:
```suggestion
    async fn paraswap_transaction_from_order(
```

solver/src/solver/paraswap_solver.rs
```diff
@@ -57,40 +60,95 @@ impl ParaswapSolver {
...
+}
+
 #[async_trait::async_trait]
 impl SingleOrderSolving for ParaswapSolver {
     async fn settle_order(&self, order: LimitOrder) -> Result<Option<Settlement>> {
+        let transaction: TransactionBuilderResponse;
+        let price_response: PriceResponse;
+        let amount: U256;
+        let mut num_retries = 0;
+        loop {
```
nlordell, line 76: I'm not a huge fan of this unbound loop, and it could avoided by moving this bit to a helper function and use a `for _ in 0..MAX_RETRIES`.

nlordell (approved):

fleupold (changes requested):
I think this is still missing the unit tests (therefore I request changes).

I also don't feel like my suggestion towards isolating logic into a single run has been considered or discussed, but given the ✅ from other reviewer I guess it's fine).

Note, that regardless of the internal design (private single run method or leave like is), the unit tests should be unaffected as it should be testing the top level interface.

solver/src/solver/paraswap_solver.rs
```diff
@@ -111,43 +169,47 @@ impl SingleOrderSolving for ParaswapSolver {
...
+    async fn paraswap_transaction_from_order(
+        &self,
+        order: &LimitOrder,
+    ) -> Result<ParaswapCompleteResponse> {
+        let token_info = self
+            .token_info
+            .get_token_infos(&[order.sell_token, order.buy_token])
+            .await;
+        let (price_response, amount) = self.get_price_for_order(&order, &token_info).await?;
+        let transaction_query = self
```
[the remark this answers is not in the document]
nlordell: This is a good point and something that I didn't consider.

solver/src/solver/paraswap_solver.rs
bh2smith: Hopefully I can also do this in a for loop as was also suggested?

solver/src/solver/paraswap_solver.rs
bh2smith: as much as I like the suggestion to use a for loop here, I am not sure if can be used in combination with @vkgnosis's good suggestion of 

```
let (x, y, z) = loop {
  if condition {
    break (1, 2, 3);
  }
}

I feel like the for loop is a bit strange for returning an Error if the iteration does indeed reach the end. Any strong preference?

**Comment by bh2smith** [comment]
@fleupold  - when you say missing tests, are you referring to the `reqwest::Error`s that are part of the enum? At this moment I am not sure how we can test them because request doesn't provide a way to construct these by hand.

```
    // Connectivity or non-response error
    #[error("Failed on send")]
    Send(reqwest::Error),

    // Recovered Response but failed on async call of response.text()
    #[error(transparent)]
    TextFetch(reqwest::Error),
```

File: solver/src/solver/paraswap_solver.rs (hunk 7 above)
Comment (bh2smith) [reply solver/src/solver/paraswap_solver.rs]: However, now that I plan to include a few different cases where we retry, this loop and match statement is becoming a bit redundant

```rust
                Err(ParaswapResponseError::PriceChange) => {
                    if num_retries >= 2 {
                        return Err(anyhow!("Price change error after {} retries", num_retries));
                    }
                    num_retries += 1;
                    continue;
                }
                Err(ParaswapResponseError::BuildingTransaction(query)) => {
                    if num_retries >= 2 {
                        return Err(anyhow!("Error Building transaction after {} retries: {}", num_retries, query));
                    }
                    num_retries += 1;
                    continue;
                }
```

Perhaps it is best to go with the for loop inside a helper method.

bh2smith: The small problem with offloading this portion into a helper method is the subtle optimization performed between.

```
            let complete_response = self.paraswap_transaction_from_order(&order).await?;
            if !satisfies_limit_price(&order, &complete_response.price_response) {
                tracing::debug!("Order limit price not respected");
                return Ok(None);
            }
            match complete_response.transaction_result {
```

We have this limit price check that returns early with Ok(None) and this wont match the type we expect from the `build_or_retry()` method. 

We are already losing some of the optimization we had before, where we would return None settlement if the price query were off but now do the price and transaction query before checking the limit price. If we do this again (i.e. move the limit price check even further out) we will be making at most `max_retries` price and transaction requests before the limit price check and thus, lose all possibility of an early return.

This is not a huge issue as of now, but I would just like to point out what we lose from going this route.

bh2smith: Ok, my most recent commit involving the helper method with for loop has indeed made things much better and overall more robust! Thanks for the suggestion.

solver/src/solver/paraswap_solver.rs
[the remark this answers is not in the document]
bh2smith: I am not entirely sure what is expected as a proposed change to this PR. Does this mean some of our Error classes are no longer relevant?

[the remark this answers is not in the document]
nlordell: I think the suggestion is to check the limit price before building the transaction.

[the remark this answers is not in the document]
bh2smith: I have commented on this here

https://github.com/gnosis/gp-v2-services/pull/812#discussion_r668112376

The problem with this that we have now a bunch of helper methods that combine these queries and the return type of checking the limit price after price query is a `None` settlement. which doesn't really fall under the category of the combined method return types... at this moment (in the current PR) we are maximally inefficient about returning None when limit price isn't satisfied because we make both price and transaction queries as many times as necessary (given the retry limit) and only after successful build do we check the limit price.

I can imagine reworking some of the error types yet again to account for this, but given the current implementation don't see how we can "return early with None"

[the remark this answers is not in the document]
nlordell: You can move the limit price check from `settle_order` to `paraswap_transaction_from_order` in between getting the price quote and building the transaction. In pseudo code:
```
fn settle_order() {
  for 0..MAX_RETRIES {
    match try_settle_order() {
      Ok(settlement) => return Ok(settlement),
      Err(err) if is_retryable_err(err) => continue,
      Err(err) => return Err(err.into()),
  }
}

fn try_settle_order() {
  let quote = api.get_price()?;
  if !is_limit_price_satisfied(quote, order) {
    return Ok(None);
  }
  let tx = api.build_transaction()?;
  return Ok(...);
}
```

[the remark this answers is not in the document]
bh2smith: I will rework this in to something similar to your suggestion.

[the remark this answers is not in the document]
bh2smith: The issue I have with this suggestion (as the reason I didn't do it this way before) is that there are many other `anyhow` errors that are not `ParaswapResponseErrors` which can occur during your suggested `try_settle_order`. If there is some way to retain the ParaswapErrors while still dropping a bunch of ?s all over to compensate for all the other possible errors then this would surely be the way to go...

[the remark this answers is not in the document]
bh2smith: Ok, I have since overcome this issue by introducing a new error enum 

SettlementError {
  Anyhow(anyhow::Error),
  Paraswap(ParaswapResponseError),
}

along with a few conversion methods.

It is now structured much simpler and as suggested in the pseudo code. I hope this is looking more like what we are going for. 

Will try and introduce some tests tomorrow morning!

solver/src/solver/paraswap_solver.rs
```diff
@@ -57,40 +60,132 @@ impl ParaswapSolver {
     }
 }
 
+fn is_retryable_err(err: &SettlementError) -> bool {
+    match err {
+        SettlementError::Paraswap(ParaswapResponseError::PriceChange) => {
+            tracing::warn!("Paraswap Price Change");
```
fleupold, line 66: nanonit: it's uncommon to print in an `is_x` method. I think the print should happen below (I will point out where)

solver/src/solver/paraswap_solver.rs
```diff
@@ -57,40 +60,132 @@ impl ParaswapSolver {
...
-        let (amount, side) = match order.kind {
-            model::order::OrderKind::Buy => (order.buy_amount, Side::Buy),
-            model::order::OrderKind::Sell => (order.sell_amount, Side::Sell),
-        };
-        let token_infos = self
+        let max_retries = 2;
+        for _ in 0..=max_retries {
+            match self.try_settle_order(order.clone()).await {
+                Ok(settlement) => return Ok(settlement),
+                Err(err) if is_retryable_err(&err) => continue,
```
fleupold, line 110: I'd print something like:
```rust
tracing::debug!("Retrying due to: {:?}", err)
```
 here instead of in the `is_retryable_error` method

solver/src/solver/paraswap_solver.rs
```diff
@@ -57,40 +60,132 @@ impl ParaswapSolver {
     }
 }
 
+fn is_retryable_err(err: &SettlementError) -> bool {
+    match err {
+        SettlementError::Paraswap(ParaswapResponseError::PriceChange) => {
+            tracing::warn!("Paraswap Price Change");
+            true
+        }
+        SettlementError::Paraswap(ParaswapResponseError::BuildingTransaction(query)) => {
```
fleupold, line 69: Could you point me to the discussion that says this is something we should be retrying? From https://gnosisinc.slack.com/archives/C0203Q9R4JJ/p1625060019006500 I took this shouldn't happen anymore?

solver/src/solver/paraswap_solver.rs
```diff
@@ -57,40 +60,132 @@ impl ParaswapSolver {
...
+        let max_retries = 2;
+        for _ in 0..=max_retries {
+            match self.try_settle_order(order.clone()).await {
+                Ok(settlement) => return Ok(settlement),
+                Err(err) if is_retryable_err(&err) => continue,
+                Err(err) => return Err(err.into()),
+            }
+        }
+        Err(anyhow!(
+            "failed to retrieve adequate response after {} retries.",
```
fleupold, line 115: Maybe we can forward the "last error" in that case?

solver/src/solver/paraswap_solver.rs
```diff
@@ -57,40 +60,132 @@ impl ParaswapSolver {
...
+        }
+        SettlementError::Paraswap(ParaswapResponseError::BuildingTransaction(query)) => {
+            tracing::warn!("Paraswap ERROR_BUILDING_TRANSACTION from query {}", query);
+            true
+        }
+        _ => false,
+    }
+}
+
+enum SettlementError {
```
fleupold, line 77: naming nit (since this is internal): To me this enum is all kinds of `ParaswapError`, with subtypes as of right now  being `TransactionCreation(ParaswapResponseError)` or `Other(anyhow::Error)`.

An alternative way (which could still be coerced from the two subtype errors) of representing this data structure would be, but the enum is totally fine:

```rust
struct ParaswapError {
  inner: anyhow::Error,
  is_retryable: bool,
}
```

solver/src/solver/paraswap_solver.rs
bh2smith: Yes, so I have not managed to uncover any text regarding this being something we should retry. I am beginning to suspect this was a discussion I had with @nlordell somewhere that isn't visible (or written) so I will try and reconstruct the reason for the decision. 

In the thread you linked to it was concluded that they had halved the ping time from 2 seconds to 1 second and this was supposed to resolve the issue. However these errors are still occurring (about once per day). Because this was a ping related issue, we thought that retrying could be a good way to circumvent this.

solver/src/solver/paraswap_solver.rs
bh2smith: I very much like this idea although it looks like the inner would have to be the enum we have recently introduced since anyhow doesn't include `ParaswapResponseError`. Am I missing something here? Shouldn't I be putting the `is_retryable`also on the `ParaswapResponseError`?

bh2smith:
Have introduced a unit test for the Paraswap solver retry logic. This PR is ready for final review! cc @fleupold

fleupold (approved):

PR #813, opened 2021-07-05
[Easy]  (de)serialization for BigRational as Decimal

bh2smith:
In response to the suggestion made [here](https://github.com/gnosis/gp-v2-services/pull/781#discussion_r663542657)

we now implement a `ratio_as_decimal` serialization for converting `BigRational` losslessly to String for export to the HTTP solver.

Should merge before #781 

### Test Plan
New tests for serialization and deserialization (please let me know if there are some other cases I may have missed)

### DIFF
--- a/model/src/ratio_as_decimal.rs
+++ b/model/src/ratio_as_decimal.rs
@@ -82,7 +82,7 @@ fn sign_04_to_03(sign_04: Sign04) -> Sign03 {
 #[cfg(test)]
 mod tests {
     use crate::ratio_as_decimal::{deserialize, serialize};
-    use num::{BigInt, BigRational};
+    use num::{BigInt, BigRational, Zero};
     use serde_json::value::Serializer;
     use serde_json::Value;
 
@@ -100,6 +100,10 @@ mod tests {
             .unwrap(),
             Value::String("0.3333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333".to_string())
         );
+        assert_eq!(
+            serialize(&BigRational::zero(), Serializer).unwrap(),
+            Value::String("0".to_string())
+        );
     }
 
     #[test]
@@ -113,6 +117,11 @@ mod tests {
         assert_eq!(
             deserialize(Value::String("1/2".to_string())).unwrap(),
             BigRational::from_float(0.5).unwrap()
+        );
+
+        assert_eq!(
+            deserialize(Value::String("0/1".to_string())).unwrap(),
+            BigRational::zero()
         );
     }
 }
--- a/model/src/ratio_as_decimal.rs
+++ b/model/src/ratio_as_decimal.rs
@@ -104,6 +104,14 @@ mod tests {
             serialize(&BigRational::zero(), Serializer).unwrap(),
             Value::String("0".to_string())
         );
+        assert_eq!(
+            serialize(
+                &BigRational::new(BigInt::from(-1), BigInt::from(1)),
+                Serializer
+            )
+            .unwrap(),
+            Value::String("-1ca".to_string())
+        );
     }
 
     #[test]
@@ -122,6 +130,10 @@ mod tests {
         assert_eq!(
             deserialize(Value::String("0/1".to_string())).unwrap(),
             BigRational::zero()
+        );
+        assert_eq!(
+            deserialize(Value::String("-1/1".to_string())).unwrap(),
+            BigRational::new(BigInt::from(-1), BigInt::from(1))
         );
     }
 }
--- a/model/src/ratio_as_decimal.rs
+++ b/model/src/ratio_as_decimal.rs
@@ -1,9 +1,7 @@
 use bigdecimal::{BigDecimal, One};
-use num::bigint::Sign as Sign04;
-use num::BigRational;
+use num::{bigint::Sign as Sign04, BigRational};
 use num_bigint::{BigInt, Sign as Sign03};
-use serde::Deserialize;
-use serde::{de, Deserializer, Serializer};
+use serde::{de, Deserialize, Deserializer, Serializer};
 use serde_with::{DeserializeAs, SerializeAs};
 use std::borrow::Cow;
 use std::str::FromStr;
--- a/model/src/ratio_as_decimal.rs
+++ b/model/src/ratio_as_decimal.rs
@@ -57,7 +57,7 @@ where
                 sign_03_to_04(numerator_bytes.0),
                 &numerator_bytes.1,
             );
-            let ten = BigRational::new(num::bigint::BigInt::from(10), num::bigint::BigInt::from(1));
+            let ten = BigRational::new(10.into(), 1.into());
             let numerator = BigRational::new(base, num::bigint::BigInt::one());
             Ok(numerator / ten.pow(exp as i32))
         }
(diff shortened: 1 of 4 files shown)


model/src/ratio_as_decimal.rs
```diff
@@ -0,0 +1,139 @@
...
+    use crate::ratio_as_decimal::{deserialize, serialize};
+    use num::{BigInt, BigRational, Zero};
+    use serde_json::value::Serializer;
+    use serde_json::Value;
+
+    #[test]
+    fn serializer() {
+        assert_eq!(
+            serialize(&BigRational::from_float(1.2).unwrap(), Serializer).unwrap(),
+            Value::String("1.1999999999999999555910790149937383830547332763671875".to_string())
```
nlordell, line 93: nit: Shorthand for this is:
```suggestion
            json!("1.1999999999999999555910790149937383830547332763671875")
```

model/src/ratio_as_decimal.rs
```diff
@@ -0,0 +1,139 @@
...
+{
+    let top_bytes = value.numer().to_bytes_le();
+    let top = BigInt::from_bytes_le(sign_04_to_03(top_bytes.0), &top_bytes.1);
+    let bottom_bytes = value.denom().to_bytes_le();
+    let bottom = BigInt::from_bytes_le(sign_04_to_03(bottom_bytes.0), &bottom_bytes.1);
+    let decimal = BigDecimal::from(top) / BigDecimal::from(bottom);
+    serializer.serialize_str(&decimal.to_string())
+}
+
+pub fn deserialize<'de, D>(deserializer: D) -> Result<BigRational, D::Error>
```
nlordell, line 42: I'm not sure this implementation is correct, I would expect:
```
assert!(ratio_as_decimal::deserialize(json!("-1/1")).is_err());
```

Because `-1/1` is not a decimal string.

bh2smith: oh yea... weird...

model/src/ratio_as_decimal.rs
```diff
@@ -0,0 +1,139 @@
...
+            BigRational::from_str(&s).map_err(|err| {
+                de::Error::custom(format!(
+                    "failed to decode {:?} as decimal BigRational: {}",
+                    s, err
+                ))
+            })
+        }
+    }
+
+    deserializer.deserialize_str(Visitor {})
```
nlordell, line 70: I think the `Visitor` might be unnecessarily complicated, you could just:
```rust
BigRational::from_str(str::deserialize(deserializer)?).map_err(...)
```

model/src/ratio_as_decimal.rs
bh2smith: Should I do this everywhere then?

model/src/ratio_as_decimal.rs
```diff
@@ -0,0 +1,139 @@
...
+            serialize(&BigRational::from_float(1.2).unwrap(), Serializer).unwrap(),
+            Value::String("1.1999999999999999555910790149937383830547332763671875".to_string())
+        );
+        assert_eq!(
+            serialize(
+                &BigRational::new(BigInt::from(1), BigInt::from(3)),
+                Serializer
+            )
+            .unwrap(),
+            Value::String("0.3333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333333".to_string())
```
nlordell, line 101: I think the conversion should fail on non-decimal values.

model/src/ratio_as_decimal.rs
nlordell: Its a nit, so its up to you. I just find `json!("some string")` way easier to write than `Value::String("some string".to_string())`.

nlordell (commented):
Looks great! Just think the `deserialize` implementation is a little off.

model/src/ratio_as_decimal.rs
bh2smith: Unfortunately, `BigRational::from_str` only works for things like `a / b`... I think we are going to have to come up with something better here. Like find the index of the decimal. Multiply and divide by 10^{number of digits to the right}.

model/src/ratio_as_decimal.rs
bh2smith: Unfortunately, we will not be able to detect infinite decimal expansions without some prime factorization and greatest common divisors. Not sure its super appropriate to include here at the moment at least.

bh2smith:
> Looks great! Just think the `deserialize` implementation is a little off.

All fixed up although it looks incredibly hideous now. Enjoy!

model/src/ratio_as_decimal.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+pub fn deserialize<'de, D>(deserializer: D) -> Result<BigRational, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let value_decimal: Result<BigDecimal, D::Error> =
+        BigDecimal::from_str(&*Cow::<str>::deserialize(deserializer)?).map_err(|err| {
+            de::Error::custom(format!("failed to decode decimal BigDecimal: {}", err))
+        });
```
nlordell, line 50: nit: You can avoid the match using `?`:

```suggestion
        })?;
```

model/src/ratio_as_decimal.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    match value_decimal {
+        Ok(big_decimal) => {
+            let (x, exp) = big_decimal.into_bigint_and_exponent();
+            let numerator_bytes = x.to_bytes_le();
+            let base = num::bigint::BigInt::from_bytes_le(
+                sign_03_to_04(numerator_bytes.0),
+                &numerator_bytes.1,
+            );
+            let ten = BigRational::new(num::bigint::BigInt::from(10), num::bigint::BigInt::from(1));
```
nlordell, line 60: nit: this is semantically identical, but more succinct 
```suggestion
            let ten = BigRational::new(10.into(), 1.into());
```

model/src/ratio_as_decimal.rs
```diff
@@ -0,0 +1,137 @@
...
+            let numerator = BigRational::new(base, num::bigint::BigInt::one());
```
nlordell, line 61:
```suggestion
            let numerator = BigRational::new(base, 1.into());
```

model/src/ratio_as_decimal.rs
```diff
@@ -0,0 +1,137 @@
...
+            Ok(numerator / ten.pow(exp as i32))
```
nlordell, line 62: Interesting, so the exponent is the inverse of what I expected. Very sneaky!

nit: I would like to see an overflow check here:
```
exp.try_into().map_err(|err| {
    de::Error::custom(format!("decimal exponent overflow: {}", err))
})    
```

nlordell (approved):
Looks good, just a couple of nits. I'm going to try and think of a cleaver way of checking if a rational is also a decimal number, otherwise it should be good.

model/src/ratio_as_decimal.rs
nlordell: Also, if you want to be fancy, you can make this a `lazy_static`.

bh2smith: I might just.

model/src/ratio_as_decimal.rs
bh2smith: although apparently (according to my IDE) as i32 is considered "safe cast" I would be happy to make this explicit.

PR #821, opened 2021-07-06
Use full order execution for Match solver

nlordell:
This PR changes the Matcha solver to use the `full_execution_amount` when building the settlement instead of always using the sell amount. This fixes issues with unwrapping WETH for ETH buy orders where the incorrect WETH amount was being unwrapped.

### Test Plan

Added unit test to check change.


solver/src/solver/matcha_solver.rs
```diff
@@ -8,7 +8,7 @@
...
...00000000000000000000000000000000000044095ea7b3000000000000000000000000def1c0ded9bec7f1a1670819833240f027b25effffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff00000000000000000000000000000000000000000000000000000000000000000000000000000000def1c0ded9bec7f1a1670819833240f027b25eff000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000600000000000000000000000000000000000000000000000000000000000000128d9627aa400000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000002b220e100000000000000000000000000000000000000000000000000470de4df82000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000000000000000000000dac17f958d2ee523a2206206994597c13d831ec7000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2869584cd000000000000000000000000100000000000000000000000000000000000001100000000000000000000000000000000000000000000003239e38b8a60dc53b70000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000&network=1
```
nlordell, line 11: Just removed trailing whitespace.

bh2smith (approved):

vkgnosis (approved):

solver/src/solver/matcha_solver.rs
```diff
@@ -122,7 +122,7 @@ impl SingleOrderSolving for MatchaSolver {
         });
         let spender = swap.allowance_target;
 
-        settlement.with_liquidity(&order, swap.sell_amount)?;
+        settlement.with_liquidity(&order, order.full_execution_amount())?;
```
josojo, line 125: Thanks for doing the fix! Do you happen to have the link to an issue? Where can I see the instance? I do not yet fully understand the problem

The full execution amount can be smaller than the sell_amount for buy orders, as because of the slippage matcha is selling a little bit more, just to have sufficient funds, in case slippage was suffered. 

I am okay with reducing the buffer like that, but then we also need change the price here:
```
let mut settlement = Settlement::new(hashmap! {
            order.sell_token => swap.buy_amount,
            order.buy_token => order.full_execution_amount(),
        });
        ```

nlordell: The executed amount should not depend on the slippage and is independent of the price. For sell orders, it is how much of the sell amount to use, and for buy order it is how much of he buy amount to use. Since these are fill-or-kill orders, we should always execute the full amount, regardless of the trading price.

I went into mode detail [here](https://gnosisinc.slack.com/archives/C020UHS2UEL/p1625586555012800).

josojo: Ah cooL!
I was not aware of how handy the `with_liquidity` function is and that it works differently depending on the order type.  Love it!

josojo (approved):

PR #823, opened 2021-07-06
[Thought Experiment] Implement a simple service component registry

nlordell:
This PR implements adds an implementation for a simple component registry. This is a very simple way to do dependency injection all the while keeping all components in a single registry.

The main motivation here is:
- Passing down components to nested components requires piping parameters all the way down to where components are registered.
- Some top level component creation functions take a huge amount of parameters, for example [`solver::create` takes 16](https://github.com/gnosis/gp-v2-services/blob/e6d9e1e336266b8c86d8de40b2c21097ee0fc7f8/solver/src/solver.rs#L66-L83)
- Reduces friction of re-using shared components across different parts of code as they are easily accessible (`registry::get<dyn MyComponentTrait>()`). Especially if service components are created with something like `MyComponentImplementation::new(&registry)`, then adding a new field `Arc<dyn SomeSharedComponentTrait>` can be done entirely locally to the component implementation.
- Is actually fairly small amount of code without the need for `unsafe`.

There are still some TODOs:
- `async` and `Result` intiailzation
- Better ergonomics with mocks (maybe even have a way to create a lazy "Mock" service registry), although I don't entirely see it to be that useful for mocking and testing.

Let me know what y'all think. It might be complete overkill for our project :stuck_out_tongue:. This is partially inspired from the ASP.NET Core dependency injection system which I have used in the past and found to be extremely useful when building large services instead of doing manual dependency injection.

### Test Plan

Added a test demonstrating why it may useful. Also added a sample for the `BaselinePriceEstimator` for how it would be able to construct itself from a service registry.

### DIFF
--- a/shared/src/service_registry.rs
+++ b/shared/src/service_registry.rs
@@ -80,13 +80,43 @@ pub trait ServiceInitializable: Sized {
     fn init(registry: &ServiceRegistry) -> Self;
 }
 
+impl<T> ServiceInitializable for T
+where
+    T: Default + Sized,
+{
+    fn init(_: &ServiceRegistry) -> Self {
+        T::default()
+    }
+}
+
 pub trait ServiceLinkable<T: ?Sized> {
     fn as_link(self: Arc<Self>) -> Arc<T>;
+}
+
+#[macro_export]
+macro_rules! impl_service_linkable {
+    ($impl:ident => $trait:ident) => {
+        impl ServiceLinkable<dyn $trait> for $impl {
+            fn as_link(self: Arc<Self>) -> Arc<dyn $trait> {
+                self
+            }
+        }
+    };
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::{
+        bad_token::{BadTokenDetecting, MockBadTokenDetecting, TokenQuality},
+        gas_price_estimation::FakeGasPriceEstimator,
+        price_estimate::{BaselinePriceEstimator, PriceEstimating},
+        sources::uniswap::pool_fetching::{MockPoolFetching, PoolFetching},
+    };
+    use ethcontract::H160;
+    use gas_estimation::GasPriceEstimating;
+    use model::order::OrderKind;
+    use std::collections::HashSet;
 
     trait Fooable: Send + Sync {
         fn foo(&self) -> i32;
@@ -103,11 +133,7 @@ mod tests {
             Foo
         }
     }
-    impl ServiceLinkable<dyn Fooable> for Foo {
-        fn as_link(self: Arc<Self>) -> Arc<dyn Fooable> {
-            self
-        }
-    }
+    impl_service_linkable!(Foo => Fooable);
 
     struct Bar(Arc<dyn Fooable>);
     impl Bar {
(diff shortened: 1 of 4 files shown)


shared/src/service_registry.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    // This would represent the command line arguments.
+    struct Options {
+        base_tokens: HashSet<H160>,
+        native_token: H160,
+    }
+
+    impl ServiceInitializable for BaselinePriceEstimator {
+        fn init(registry: &ServiceRegistry) -> Self {
+            let options = registry.get::<Options>();
```
bh2smith, line 176: I'm a bit confused as to how the registry already has these options to get. I see only the struct was defined just above...

shared/src/service_registry.rs
```diff
@@ -0,0 +1,215 @@
...
+
+        let mut registry = ServiceRegistry::default();
+        registry
+            .add(Options {
+                base_tokens: Default::default(),
+                native_token: Default::default(),
+            })
+            .add(token_detector)
+            .link::<MockBadTokenDetecting, dyn BadTokenDetecting>()
+            .register_as::<MockPoolFetching, dyn PoolFetching>()
```
bh2smith, line 202: I'm not entirely clear on the differences between the three things here. Is register as essentially a combination of add and link?

bh2smith (commented):
So far so good. I would hope that we can get some extensive docs on this that would suffice as developer docs. This looks promising, but at the moment I wouldn't know how to use it in practise. I was partially expecting to see at least one existing file where we actually see the benefit of this by making use of it and get an impression of how much the boiler plate code is reduced. 

I will have to read over this again more carefully tomorrow (specifically the generics to see if I can get a better grasp on the distinction between add link register and register as).

shared/src/service_registry.rs
nlordell: It gets added to the registry that we create the `BaselinePriceEstimator` for:
```
        registry
            .add(Options {
                base_tokens: Default::default(),
                native_token: Default::default(),
            })
```

shared/src/service_registry.rs
nlordell: Yes, and register requires the type to implement `ServiceInitializable` so that it can create the instance on its behalf.

shared/src/service_registry.rs
bh2smith: I would understand this line more if I had seen registry dot add before registry dot get...

nlordell: The registry is a hashmap of components by type ID. So when initializing the `BaselinePriceEstimator` it expects all its dependencies to already be added to the hashmap.

In the test function, we create the hashmap (adding all the required dependencies) and then use it to initialize the `BaselinePriceEstimator`. The order is just positioning of code - they are separate functions so their position in the file does not reflect the call order.

nlordell:
> I was partially expecting to see at least one existing file where we actually see the benefit of this by making use of it and get an impression of how much the boiler plate code is reduced.

I wanted to get a general sentiment for the interest in in before doing the work. The idea would be:

```rust
// main.rs
fn main() {
    let args = Arguments::from_args();

    let mut registry = ServiceRegistry::new();
    registry
        .add(args.shared.clone())
        .add(args)
        .registry::<prometheus::Registry>()
        .register::<Metrics>()
        .link::<Metrics, dyn SolverMetrics>()
        .link::<Metrics, dyn TransportMetrics>()
        .link::<Metrics, dyn PoolCacheMetrics>()
        .register::<Web3>()
        .register::<Contracts>()
        .register::<database::Instrumented>()
        .add(Metrics::new(Default::default())
        // etc.
}
```

Then the initialization code would be done locally to each component.

vkgnosis:
Cool idea. I am not familiar with dependency injection (why is it even called that?) so not sure. In my own words to check that I understand it correctly:
In main we create a service registry that contains all shared components where a shared component can be a struct or dyn Trait. Components that other other components instead of taking them in their constructor take a service registry from which they can query them.
Has the advantages that you listed and has the disadvantage that it is a little harder too see which components are used and the dependencies between them since you effectively have a global component map that everyone can access. So for example it might be easier to accidentally create components in the wrong in order in which case we would panic because some needed component does not yet exist. Most likely that would be caught on start up so probably not a problem.

nlordell:
> Most likely that would be caught on start up so probably not a problem.

It should get caught on `register` (since we don't do lazy initialization - if we did, we would have to deal with dependency cycles for example)

fleupold:
Good idea to start thinking about DI given the increasing size/interdependency of components. Components should probably still implement their normal constructor (to make dependencies explicit and allow testing without a registry), and then implement a trivial init method getting the dependencies from the registry (In other languages I have seen the `init` implementation even auto-derived - but this might require introspection feature rust doesn't support). Since the registry is basically global state, I think we should avoid passing it around on the objects themselves (maybe we can achieve this by limiting the concurrency/lifetime of the registry so it cannot be easily "stored" on other objects?)

One solution for the ordering problem I have seen in the past is to lazily initialise the components when they are first queried (so register only creates a closure on how to initialise them). This would also allow to specify if we always want a "new" component (e.g. for the AllowanceManager) or if we want to share instances across the session (cached after first init). On the other hand, by instantiating them in `main` we would have more guarantees to fail early (rather than only crashing when a component is queried eventually)

nlordell:
> Components should probably still implement their normal constructor (to make dependencies explicit and allow testing without a registry), and then implement a trivial init method

I agree this could be useful for testing. For some of the simpler components, they can just be constructed directly in unit tests as well.

> In other languages I have seen the init implementation even auto-derived - but this might require introspection feature rust doesn't support

Proc macros!

PR #824, opened 2021-07-07
Balancer Pool Fetching

bh2smith:
Just setting the stage for our integration test and thought we might also want to initialize pools for the orderbook.

Not meant to be merged, but could eventually go into main.


bh2smith:
No longer necessary as his had already been done.

PR #825, opened 2021-07-08
Add Archer settlement flow

vkgnosis:
Adds a high level function to manage submitting a settlement through the
archer network roughly equivalent to settlement_submission::submit .
Not yet connected to the driver.

This is a draft because I am making use of an ethcontract commit that hasn't been released yet (and 270 of the changed lines are related to that).
Edit: ready for review

### Test Plan
I added a manual mainnet test. I am unsure about unit tests because almost all of the logic would have to be mocked and new traits introduced). The non io related logic feels fairly straightforward to me.

### DIFF
--- a/solver/src/interactions/block_coinbase.rs
+++ b/solver/src/interactions/block_coinbase.rs
@@ -1,0 +1,25 @@
+use crate::{encoding::EncodedInteraction, settlement::Interaction};
+use hex_literal::hex;
+use primitive_types::{H160, U256};
+
+// vk: A simple contract I made with verified code on etherscan:
+// https://etherscan.io/address/0x5c2cD95CF750B8f8A4881d96F04bf571A07042B1
+// Gas use for a full transaction when amount is 0 is 23849 and nonzero 30549.
+const MAINNET_ADDRESS: H160 = H160(hex!("5c2cd95cf750b8f8a4881d96f04bf571a07042b1"));
+const METHOD_ID: [u8; 4] = hex!("2755cd2d");
+
+#[derive(Clone, Debug)]
+pub struct PayBlockCoinbase {
+    // ether wei
+    pub amount: U256,
+}
+
+impl Interaction for PayBlockCoinbase {
+    fn encode(&self) -> Vec<EncodedInteraction> {
+        vec![(
+            MAINNET_ADDRESS,
+            self.amount,
+            ethcontract::Bytes(METHOD_ID.to_vec()),
+        )]
+    }
+}
--- a/solver/src/settlement_submission/archer_settlement.rs
+++ b/solver/src/settlement_submission/archer_settlement.rs
@@ -1,0 +1,377 @@
+// Design:
+// As in the traditional transaction submission workflow the main work in this module is checking
+// the gas price in a loop and updating the transaction when the gas price increases. This differs
+// so that we can make use of the property that archer transactions do not cost gas if they fail.
+// When we detect that the transaction would no longer succeed we stop trying to submit and return
+// so that the solver can run again.
+// In addition to simulation failure we make use of a deadline after which submission attempts also
+// stop. This allows the solver to update and improve a solution even if it hasn't yet become
+// invalid.
+// We do not know in advance which of our submitted transactions will get mined. Instead of polling
+// all of them we only check the account's nonce as an optimization. When this happens all our
+// transactions definitely become invalid (even if the transaction came for whatever reason
+// from outside) so it is only at that point that we need to check the hashes individually to the
+// find the one that got mined (if any).
+
+// Idea:
+// The current driver code is either solving or waiting for a solution to be mined.
+// When using archer we can improve on this by always solving the current order book as if there was
+// no pending transaction and continually updating the transaction we are sending to archer.
+// This is a bigger change so the code here still adheres to the traditional tx submission
+// workflow.
+
+// TODO: Make sure settlement contract always has eth to pay the miner. Could reuse buffers or just
+// manually fund it like we do with other solver accounts and make part of our monitor scripts.
+
+// TODO: Node gas estimates seem a little higher than actual. This can be seen when comparing
+// to tenderly simulations and mined transactions. This causes us to overpay somewhat.
+
+use super::{archer_api::ArcherApi, ESTIMATE_GAS_LIMIT_FACTOR};
+use crate::{interactions::block_coinbase, settlement::Settlement};
+use anyhow::{anyhow, Context, Result};
+use chrono::{DateTime, Utc};
+use contracts::GPv2Settlement;
+use ethcontract::{errors::MethodError, transaction::Transaction, Account, GasPrice};
+use futures::FutureExt;
+use gas_estimation::GasPriceEstimating;
+use primitive_types::{H256, U256};
+use shared::Web3;
+use std::{
+    sync::Arc,
+    time::{Duration, Instant, SystemTime},
+};
+use web3::types::TransactionId;
+
+pub struct ArcherSolutionSubmitter {
+    pub web3: Web3,
+    pub contract: GPv2Settlement,
+    pub account: Account,
+    pub archer_api: ArcherApi,
+    pub gas_price_estimator: Arc<dyn GasPriceEstimating>,
+    pub gas_price_cap: f64,
+}
+
+impl ArcherSolutionSubmitter {
+    /// Submit a settlement to the contract, updating the transaction with gas prices if they increase.
+    ///
+    /// Goes through the archerdao network so that failing transactions do not get mined and thus do
+    /// not cost gas.
+    ///
+    /// Returns None if the deadline is reached without a mined transaction.
+    ///
+    /// Only works on mainnet.
+    pub async fn submit(
+        &self,
+        target_confirm_time: Duration,
+        deadline: SystemTime,
+        settlement: Settlement,
+        gas_estimate: U256,
+    ) -> Result<Option<H256>> {
+        let nonce = self.nonce().await?;
+
+        tracing::info!(
+            "starting archer solution submission at nonce {} with deadline {}",
+            nonce,
+            DateTime::<Utc>::from(deadline).to_rfc3339(),
+        );
+
+        let mut transactions = Vec::new();
+        let submit_future = self.submit_with_increasing_gas_prices_until_simulation_fails(
+            deadline,
+            target_confirm_time,
+            nonce,
+            settlement,
+            gas_estimate,
+            &mut transactions,
+        );
+
+        let nonce_future = self.wait_for_nonce_to_change(nonce);
+
+        let deadline_future = tokio::time::sleep(
+            deadline
+                .duration_since(SystemTime::now())
+                .unwrap_or_else(|_| Duration::from_secs(0)),
+        );
+
+        futures::select! {
+            method_error = submit_future.fuse() => tracing::info!("stopping submission because simulation failed: {:?}", method_error),
+            new_nonce = nonce_future.fuse() => tracing::info!("stopping submission because account nonce changed to {}", new_nonce),
+            _ = deadline_future.fuse() => tracing::info!("stopping submission because deadline has been reached"),
+        };
+
+        // After stopping submission of new transactions we wait for some time to give a potentially
+        // mined previously submitted transaction time to propagate to our node.
+
+        if !transactions.is_empty() {
+            const MINED_TX_PROPAGATE_TIME: Duration = Duration::from_secs(20);
+            const MINED_TX_CHECK_INTERVAL: Duration = Duration::from_secs(5);
+            let tx_to_propagate_deadline = Instant::now() + MINED_TX_PROPAGATE_TIME;
+
+            tracing::info!(
+                "waiting up to {} seconds to see if a transaction was mined",
+                MINED_TX_PROPAGATE_TIME.as_secs()
+            );
+
+            loop {
+                if let Some(hash) = find_mined_transaction(&self.web3, &transactions).await {
+                    tracing::info!("found mined transaction {}", hash);
+                    return Ok(Some(hash));
+                }
+                if Instant::now() + MINED_TX_CHECK_INTERVAL > tx_to_propagate_deadline {
+                    break;
+                }
+                tokio::time::sleep(MINED_TX_CHECK_INTERVAL).await;
+            }
+        }
+
+        tracing::info!("did not find any mined transaction");
+        Ok(None)
+    }
+
+    async fn nonce(&self) -> Result<U256> {
+        self.web3
+            .eth()
+            .transaction_count(self.account.address(), None)
+            .await
+            .context("transaction_count")
+    }
+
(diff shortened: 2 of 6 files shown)


shared/src/sources/balancer/pool_init.rs
```diff
@@ -101,20 +101,14 @@ where
...
         // they cannot be considered the same, but for the networks that we
         // support (xDAI, Rinkeby and Mainnet) they are.
-        .ok_or_else(|| {
-            anyhow!(
-                "missing {} deployment for {}",
-                artifact.contract_name,
-                chain_id,
-            )
-        })?
+        .ok_or_else(|| anyhow!("missing {} deployment for {}", artifact.name, chain_id,))?
```
nlordell, line 111:
```suggestion
        .ok_or_else(|| anyhow!("missing {} deployment for {}", artifact.name, chain_id))?
```

solver/src/settlement_submission/archer_settlement.rs
```diff
@@ -0,0 +1,377 @@
...
+// When we detect that the transaction would no longer succeed we stop trying to submit and return
+// so that the solver can run again.
+// In addition to simulation failure we make use of a deadline after which submission attempts also
+// stop. This allows the solver to update and improve a solution even if it hasn't yet become
+// invalid.
+// We do not know in advance which of our submitted transactions will get mined. Instead of polling
+// all of them we only check the account's nonce as an optimization. When this happens all our
+// transactions definitely become invalid (even if the transaction came for whatever reason
+// from outside) so it is only at that point that we need to check the hashes individually to the
+// find the one that got mined (if any).
```
nlordell, line 14: Nice documentation!

solver/src/settlement_submission/archer_settlement.rs
```diff
@@ -0,0 +1,377 @@
+// Design:
+// As in the traditional transaction submission workflow the main work in this module is checking
+// the gas price in a loop and updating the transaction when the gas price increases. This differs
+// so that we can make use of the property that archer transactions do not cost cast if they fail.
```
nlordell, line 4:
```suggestion
// so that we can make use of the property that archer transactions do not cost gas if they fail.
```

solver/src/interactions/block_coinbase.rs
```diff
@@ -0,0 +1,25 @@
+use crate::{encoding::EncodedInteraction, settlement::Interaction};
+use hex_literal::hex;
+use primitive_types::{H160, U256};
+
+// vk: A simple contract I made with verified code on etherscan:
+// https://etherscan.io/address/0x5c2cD95CF750B8f8A4881d96F04bf571A07042B1
+// Gas use for a full transaction when amount is 0 is 23849 and nonzero 30549.
+const MAINNET_ADDRESS: H160 = H160(hex!("5c2cd95cf750b8f8a4881d96f04bf571a07042b1"));
```
nlordell, line 8: This contract requires the settlement contract to hold Eth right? So we fund the Settlement contract with gas money. This is nice as we already have ways to manage the settlement contract funds.

We can also consider sending 1wei to the tipping contract as that might decrease total gas costs (@fedgiac might know better than I do).

solver/src/settlement_submission/archer_settlement.rs
```diff
@@ -0,0 +1,377 @@
...
+// When using archer we can improve on this by always solving the current order book as if there was
+// no pending transaction and continually updating the transaction we are sending to archer.
+// This is a bigger change so the code here still adheres to the traditional tx submission
+// workflow.
+
+// TODO: Make sure settlement contract always has eth to pay the miner. Could reuse buffers or just
+// manually fund it like we do with other solver accounts and make part of our monitor scripts.
+
+// TODO: Node gas estimates seem a little higher than actual. This can be seen when comparing
+// to tenderly simulations and mined transactions. This causes us to overpay somewhat.
```
nlordell, line 27: I think this is because the tipping contract goes from 0 balance -> non-zero balance which has a higher gas costs from going from non-zero balance A -> non-zero balance B. Then, a portion of that gas may get refunded because it goes back to its initial state by the end of the transaction.

It works like this for storage, so I'm *assuming* that there may be something similar at play here, but I'm not 100% sure.

solver/src/settlement_submission/archer_settlement.rs
```diff
@@ -0,0 +1,377 @@
...
+            deadline
+                .duration_since(SystemTime::now())
+                .unwrap_or_else(|_| Duration::from_secs(0)),
+        );
+
+        futures::select! {
+            method_error = submit_future.fuse() => tracing::info!("stopping submission because simulation failed: {:?}", method_error),
+            new_nonce = nonce_future.fuse() => tracing::info!("stopping submission because account nonce changed to {}", new_nonce),
+            _ = deadline_future.fuse() => tracing::info!("stopping submission because deadline has been reached"),
+        };
```
nlordell, line 100: `futures` rocks! I find this quite easy to follow, and like the use of `select` as a way of "racing" to see which one finished first.

solver/src/settlement_submission/archer_settlement.rs
```diff
@@ -0,0 +1,377 @@
...
+        }
+    }
+
+    /// Keep submitting the settlement transaction to the archer network as gas prices changes.
+    ///
+    /// Returns when simulation of the transaction fails. This likely happens if the settlement
+    /// becomes invalid due to changing prices or the account's nonce changes.
+    ///
+    /// Potential transaction hashes are communicated back through a shared vector.
+    async fn submit_with_increasing_gas_prices_until_simulation_fails(
```
nlordell, line 175: Its a pity we can't resuse the code from the transaction submission crate.

solver/src/settlement_submission/archer_settlement.rs
```diff
@@ -0,0 +1,377 @@
...
+            let mut settlement = settlement.clone();
+            settlement
+                .encoder
+                .append_to_execution_plan(block_coinbase::PayBlockCoinbase {
+                    amount: tx_gas_cost_in_ether_wei,
+                });
+            let method = super::retry::settle_method_builder(&self.contract, settlement.into())
+                .from(self.account.clone())
+                .nonce(nonce)
+                // Wouldn't work because the function isn't payable.
```
nlordell, line 220: I would even argue that its nice that the gas comes from a contract, so that we can use accounts without funding for submitting transactions.

solver/src/settlement_submission/archer_settlement.rs
```diff
@@ -0,0 +1,377 @@
...
+
+            transactions.push(hash);
+            previous_tx = Some((gas_price, raw_signed_transaction));
+            tokio::time::sleep(UPDATE_INTERVAL).await;
+        }
+    }
+}
+
+/// From a list of potential hashes find one that was mined.
+async fn find_mined_transaction(web3: &Web3, hashes: &[H256]) -> Option<H256> {
```
nlordell, line 282: What we could do is an `eth_getLogs` for `Settlement` log with a topic filter for the solver address you are looking for. We should be able to do this with a single ETHRPC request.

I'm 100% fine with this solution as well.

nlordell (commented):
Looking real good already.

solver/src/interactions/block_coinbase.rs
fedgiac: I think changing the account balances to a nonzero values only applies to refunds, and contract ETH balances don't matter. In any case, I'm not sure how to compute the gas cost of the call opcode as computed in the whitepaper, so I was curious enough to test it and it seems that a nonzero balance doesn't change the gas cost ([contract transactions](https://rinkeby.etherscan.io/address/0x809d1c696963266ef163d3ea10d146050b3728f9)). Something that should make the contract more efficient without actually writing bytecode is having a `receive` function that defaults to sending the money to the coinbase, in this way we should save some calldata cost.

Something I wondered is if the coinbase could ever be a contract, in which case this call might fail if the receive logic is too expensive. In the yellow paper there is nothing explicitly saying it's not possible, however it only says that the balance is updated and not that any code would be executed, so I wonder if this is a way to deposit ETH into any contract even if it's not payable.

vkgnosis: Good point about coinbase being contract. I looked through the flashbots documentation again and found this https://docs.flashbots.net/flashbots-auction/searchers/advanced/coinbase-payment/#managing-payments-to-coinbaseaddress-when-it-is-a-contract :

> Miners will occasionally have a smart contract listed as their block.coinbase address. This changes the expected behavior of the making payments to block.coinbase. Specifically it costs more gas to transfer ETH to block.coinbase if it is a contract than if it is an EOA, and as such many searchers will underestimate their gas consumption and their bundles will fail for miners who use contracts instead.
To handle this edge case searchers can up their gas limit to accomodate the additional payment to miners and call block.coinbase in the following way:
> ```
> block.coinbase.call{value: _ethAmountToCoinbase}(new bytes(0));
> ```
> However, searchers should be acutely aware of the risk of reentrancy attacks, as calling coinbase in this way temporarily gives execution to a third party, and typically payments to coinbase are made after checks for profit. Moreover, searchers should be aware that supporting payments to coinbase addresses that are contracts will cause their gas consumption to go up, and as a result their bundle gas price to go down. This is a tradeoff that should be considered.

I am not sure what to do about it and how common the address being a contract is. It seems like miners should avoid making it a contract precisely because of these issues.

vkgnosis: > so I wonder if this is a way to deposit ETH into any contract even if it's not payable

Without testing it I am guessing that yes the miner block reward could be deposited this way. I wonder if there is a contract out there that is attackable by a miner this way.

nlordell: > so I wonder if this is a way to deposit ETH into any contract even if it's not payable

`SELFDESTRUCT` will always transfer ETH without invoking the contract code (so would work for non-payable contracts).

bh2smith: This discussion reminds me of an old Smart Contract Capture the Flag problem at ETHCC (2018?) where you could fund a non-payable contract with eth by sending the eth to the contract address before deployment. This means you would have to use create2 for the contract deployment to have a deterministic address, and just send the funds there before hand. 

Anyway, not useful, just reminiscing.

fedgiac (approved):
Looks good. :heavy_check_mark:

solver/src/settlement_submission/archer_settlement.rs
```diff
@@ -0,0 +1,377 @@
...
+        nonce: U256,
+        settlement: Settlement,
+        gas_estimate: U256,
+        transactions: &mut Vec<H256>,
+    ) -> MethodError {
+        const UPDATE_INTERVAL: Duration = Duration::from_secs(10);
+
+        // The amount of extra gas it costs to include the payment to block.coinbase interaction in
+        // an existing settlement.
+        let gas_estimate = gas_estimate + U256::from(22859);
```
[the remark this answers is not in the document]
vkgnosis: I don't remember exactly but I think I got this value by comparing an empty settlement to a settlement that only has this interaction. The latter cost this much more gas. Maybe I made a mistake or maybe there is some more overhead for non empty settlements that I attributed to this interaction when I shouldn't have.

solver/src/settlement_submission/archer_settlement.rs
```diff
@@ -0,0 +1,377 @@
...
+    ///
+    /// Only works on mainnet.
+    pub async fn submit(
+        &self,
+        target_confirm_time: Duration,
+        deadline: SystemTime,
+        settlement: Settlement,
+        gas_estimate: U256,
+    ) -> Result<Option<H256>> {
+        let nonce = self.nonce().await?;
```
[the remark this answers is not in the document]
vkgnosis: I don't plan on mixing them for the same settlement (but it could be interesting) so this should be fine. If we did do that I guess it would still be fine because the settling the new transaction should fail simulation and as you said not be taken by miners.

solver/src/settlement_submission/archer_settlement.rs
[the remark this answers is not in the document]
fedgiac: I see, I thought it was the standalone transaction price since it was mentioned before and was similar. The comparison approach makes a lot of sense, here is what I see:

| Interactions in settlement | Gas cost | Difference without tip | 
| ------------- | ------------- | ------------- |
| [none](https://dashboard.tenderly.co/tx/rinkeby/0xc28ccc35fbee495248adaa23fba4fdedb0c240bf6e0634a4c1ff532b81564101/gas-usage) | 45844  | - |
| [tip](https://dashboard.tenderly.co/tx/rinkeby/0x885bff5eba9b80e51e1df9d6441d909761f1cc897f233b144ad04188d6c3885a/gas-usage)  | 64226  | 18382|
| [other](https://dashboard.tenderly.co/tx/rinkeby/0x9c073dfc29495e3e4e7ce57dbc35a1a7a998e1cd1397b9feee3c045f87856166/gas-usage)  | 56075  | - |
| [other + tip](https://dashboard.tenderly.co/tx/rinkeby/0xbdee3247de2c9ea8b51e0d34dee5ce13e75b3e001840c58f058ec90eba101861/gas-usage)  | 74421 | 18346 |

So my estimate was about 3k gas off, not sure where this extra gas comes from.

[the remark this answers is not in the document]
vkgnosis: I will change it to `18346` then. Thanks for measuring.

solver/src/settlement_submission/archer_settlement.rs
```diff
@@ -0,0 +1,377 @@
...
+                .append_to_execution_plan(block_coinbase::PayBlockCoinbase {
+                    amount: tx_gas_cost_in_ether_wei,
+                });
+            let method = super::retry::settle_method_builder(&self.contract, settlement.into())
+                .from(self.account.clone())
+                .nonce(nonce)
+                // Wouldn't work because the function isn't payable.
+                // .value(tx_gas_cost_in_ether_wei)
+                .gas(U256::from_f64_lossy(gas_limit))
+                .gas_price(GasPrice::Value(U256::zero()));
```
fleupold, line 223: Note, that this will only work until August 4th. EIP 1559 requires at least the base fee to be paid by tx.origin.

fleupold (approved):
Logic looks good to me. I'm a bit worried that all these branches are only covered by a single integration test. Given that the ArcherAPI is already abstract, I wonder if - with a mockable web3 instance - we couldn't write some nice unit tests for this logic?

vkgnosis:
I want to continue integrating this and then come back to adding a test with mocks. From my pov I don't feel they will add much confidence in that this is correct and they make the code more complicated because of extra traits. If we mocked web3 on a transport level the test set up would probably be very annoying so I think an extra trait for our specific high level logic is better.

solver/src/settlement_submission/archer_settlement.rs
fleupold: It will be a valid tx, but since gasPrice is translated into `base_fee_per_gas` it will only get mined if the base fee is 0 (which likely won't be the case). The miner is not allowed to pay the base fee on behalf of the tx.origin.

PR #826, opened 2021-07-08
Filter orders that have same buy and sell token

vkgnosis:
This doesn't make sense and caused a panic because we have a code path
that is unwrapping `order_creation.token_pair()` which fails if the
tokens are the same.
It is reasonable to have this as an invariant so we filter it out on
order creation.

### Test Plan
didn't add test because the check is obvious enough


bh2smith (approved):
Good catch! 🎸 on!

nlordell (approved):

PR #829, opened 2021-07-08
Add support for disabling specific ParaSwap DEXs

nlordell:
This PR adds support for disabling specific ParaSwap DEXs when querying for a price and building a transaction with the ParaSwap solver.

This was added in order to temporarily disable `ParaSwapPool4` private market maker as we are running into issues with settling solutions when they are part of the ParaSwap route. The reason for this is that when we build the ParaSwap transaction, we ask the private market maker for a binding quote. When we end up not using that quote (because the ParaSwap route loses in the price competition) then the solver address that is making the request to build the transaction gets flagged by the PMM, who eventually charges an additional 1% on the binding quote because of the repeatedly unused requests. This additional 1% causes the settlement to fail because the price does not match the settlement price.

This is an imperfect solution, but its a temporary band-aid to alleviate the problem until we can implement a better, but much more involved, solution. The issue we have now is that we require the calldata before deciding the winning solution. This means that we have to request a binding quote from the PMM before we can determine whether or not we want to use it. A long term fix for this, would be to move the solution competition to before the the calldata is built, and then call the transaction builder ParaSwap API only if it is the winning settlement, thus preventing these kinds of issues in the future. Created #828.

Tagging @anxolin as this might affect front end price estimation, as the quoted price from ParaSwap may not actually be usable by the solver since we do not have access to some PMM liquidity in the services.

### Test Plan

Adjust the unit tests, and ran manual tests with `cargo test -p solver -- paraswap_solver --ignored`.

### DIFF
--- a/solver/src/solver/paraswap_solver/api.rs
+++ b/solver/src/solver/paraswap_solver/api.rs
@@ -132,8 +132,10 @@ pub struct PriceQuery {
     pub to_decimals: usize,
     /// amount of source token (in the smallest denomination, e.g. for ETH - 10**18)
     pub amount: U256,
-    // Type of order
+    /// Type of order
     pub side: Side,
+    /// The list of DEXs to exclude from the computed price route.
+    pub exclude_dexs: Option<Vec<String>>,
 }
 
 impl PriceQuery {
@@ -156,6 +158,12 @@ impl PriceQuery {
             .append_pair("amount", &self.amount.to_string())
             .append_pair("side", side)
             .append_pair("network", "1");
+
+        if let Some(dexs) = &self.exclude_dexs {
+            url.query_pairs_mut()
+                .append_pair("excludeDEXS", &dexs.join(","));
+        }
+
         url
     }
 }
@@ -283,6 +291,7 @@ mod tests {
             to_decimals: 18,
             amount: 135_000_000_000_000_000_000u128.into(),
             side: Side::Sell,
+            exclude_dexs: None,
         };
 
         let price_response: PriceResponse = reqwest::get(price_query.into_url())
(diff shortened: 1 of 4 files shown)


bh2smith (approved):
Nice change and even better story in the PR description. Right on!

solver/src/solver/paraswap_solver/api.rs
```diff
@@ -132,8 +132,10 @@ pub struct PriceQuery {
     pub to_decimals: usize,
     /// amount of source token (in the smallest denomination, e.g. for ETH - 10**18)
     pub amount: U256,
-    // Type of order
+    /// Type of order
     pub side: Side,
+    /// The list of DEXs to exclude from the computed price route.
+    pub exclude_dexs: Option<Vec<String>>,
```
vkgnosis, line 138: Can remove Option because empty Vec is same as None.

vkgnosis (approved):

solver/src/solver/paraswap_solver/api.rs
nlordell: I just left it in case `excludeDEXS=` was different than omitting it completely. Will change though.

vkgnosis:
Since disabling specific private market makers has the chance to make the price worse we should make sure the frontend does the same thing so that orders stay matchable. This PR itself doesn't disable any, just a note.

fleupold:
Should we merge this? This has been the most common source for Paraswap failures yesterday and today so far? I feel like we already won't be able to match people when PMM4 is involved (because they give us such a bad price) so we might as well exclude them, maybe there is another PMM that can offer a price that still satisfies the user's limit?

vkgnosis:
Rebased it and fixed merge conflict. Will set to auto merge.

fleupold:
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FKJ1f5iTl4Oo7u%2Fgiphy-downsized-medium.gif&token=thanks" data-gitmeme-token="thanks"></a>

PR #830, opened 2021-07-09
Prioritize most recent order of user selling the same token

vkgnosis:
This helps in the case where a user places an order and the price moves
so that it can no longer be matched. With the existing logic the oldest
order has priority. So if the user has no extra balance then if
they place a new order with an updated price it still won't be
considered because the first one eats up the balance.
In this commit we reverse the ordering so that newer orders are
considered first which works more intuitively in this situation.

Note that users could alternatively cancel the older order. Long
term a better solution probably resides in the frontend. It could warn
users when they have multiple orders selling the same token that
together make up more than the user's balance.

### Test Plan
adjusted unit test


nlordell (approved):

bh2smith (approved):
Makes sense! I love this new order ordering

PR #831, opened 2021-07-09
Consider pool limits in BaselineSolvable computation

nlordell:
This PR modifies the Uniswap pool `BaselineSolvable` implementation to consider pool reserve limits when computing input and output amounts.

### Test Plan

Added new tests for new checking method, modified existing tests to be in valid pool range.

### DIFF
--- a/shared/src/sources/uniswap/pool_fetching.rs
+++ b/shared/src/sources/uniswap/pool_fetching.rs
@@ -422,6 +422,8 @@ mod tests {
 
     #[test]
     fn check_final_reserve_limits() {
+        // final out reserve too low
+        assert!(check_final_reserves(0.into(), 1.into(), 1_000_000.into(), 0.into()).is_none());
         // final in reserve too high
         assert!(
             check_final_reserves(1.into(), 0.into(), *POOL_MAX_RESERVES, 1_000_000.into())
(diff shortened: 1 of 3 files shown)


shared/src/sources/uniswap/pool_fetching.rs
```diff
@@ -13,6 +13,10 @@ use std::{collections::HashSet, sync::Arc};
 pub const MAX_BATCH_SIZE: usize = 100;
 const POOL_SWAP_GAS_COST: usize = 60_000;
 
+lazy_static::lazy_static! {
+    static ref POOL_MAX_RESERVES: U256 = U256::from((1u128 << 96) - 1);
```
bh2smith, line 17: I didn't even realize pools had max reserves? Is this specific to uniswap since they use uint96 for amounts?

bh2smith (approved):
Very nice!

shared/src/sources/uniswap/pool_fetching.rs
nlordell: In Balancer they have `uint256` reserves, so a higher maximum.

bh2smith: So then I would assume this change only applies to Uniswap-Like-Pools

Seems reasonable.

vkgnosis (approved):

PR #832, opened 2021-07-09
Fix BaselineSolvable parameters for weighted pool

nlordell:
This PR fixes the parameter order for the Balancer V2 weighted pool `BaselineSolvable` implementation that was causing the panics in the Kaffeekranchen today.

I've also slightly modified the `BaselineSolvable` interface to make this mistake less likely in the future.

### Test Plan

Exiting unit tests

### DIFF
--- a/shared/src/sources/balancer/swap.rs
+++ b/shared/src/sources/balancer/swap.rs
@@ -87,7 +87,7 @@ impl BaselineSolvable for WeightedPoolRef<'_> {
         .flatten()
     }
 
-    fn get_amount_in(&self, out_token: H160, out_amount: U256, in_token: H160) -> Option<U256> {
+    fn get_amount_in(&self, in_token: H160, out_amount: U256, out_token: H160) -> Option<U256> {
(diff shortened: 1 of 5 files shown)


shared/src/sources/balancer/swap.rs
```diff
@@ -87,7 +87,7 @@ impl BaselineSolvable for WeightedPoolRef<'_> {
         .flatten()
     }
 
-    fn get_amount_in(&self, out_token: H160, out_amount: U256, in_token: H160) -> Option<U256> {
+    fn get_amount_in(&self, in_token: H160, (out_amount, out_token): (U256, H160)) -> Option<U256> {
```
fedgiac, line 90: Tricky!

fedgiac (approved):
Wow, nice work spotting the issue, I didn't even notice it in my first pass of reading the changes.
I like the new token grouping.

alfetopito (approved):

bh2smith (approved):
nice.

PR #835, opened 2021-07-09
Allow solutions with empty AMMs field

nlordell:
This PR just adds a serde attribute to allow empty `amms` field from the solver when it returns an empty solution.

### Test Plan

Added new unit test.


bh2smith (approved):
Looks good, but we may also run into a solution that has a uniswaps field and ignores it when its empty too. Although we can always wait till we see it in real life.

nlordell:
I think it ignores unknown fields, so missing or empty `uniswaps` should be fine right?

bh2smith:
Merge away

PR #836, opened 2021-07-09
HTTP solver - dont pass empty pools

bh2smith:
When testing we found that some balancer pools were empty and this made the http solver mad.

### DIFF
--- a/solver/src/solver/http_solver.rs
+++ b/solver/src/solver/http_solver.rs
@@ -278,11 +278,15 @@ impl HttpSolver {
         let weighted_product_orders = self.map_amm_orders_for_solver(orders.2);
         let token_models = self.token_models(&token_infos, &price_estimates);
         let order_models = self.order_models(&limit_orders, gas_price);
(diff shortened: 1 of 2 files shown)


solver/src/solver/http_solver.rs
```diff
@@ -278,11 +278,15 @@ impl HttpSolver {
...
-            gas_price,
-        );
+        let amm_models = self
+            .amm_models(
+                &constant_product_orders,
+                &weighted_product_orders,
+                gas_price,
+            )
+            .into_iter()
+            .filter(|(_, model)| !model.has_sufficient_reserves())
```
nlordell, line 288: I think this should be:

```suggestion
            .filter(|(_, model)| model.has_sufficient_reserves())
```

nlordell (approved):
Awesome!

For some reason I remember we already filter pools with 0 reserves somewhere in the code, but I couldn't find it. Maybe @vkgnosis knows where?

I think we can merge as is and then revisit if I'm not crazy about :point_up:

PR #842, opened 2021-07-11
[Trivial] Log matcha query errors

fleupold:
In order to get the actual error reason for alerts like: https://gnosisinc.slack.com/archives/CUD4MUCUF/p1626006814000100

instead of just something saying we got json that was not matching our struct definition (which no error does).

### Test Plan
🤷


bh2smith (approved):

vkgnosis (approved):

josojo (approved):
Neat! I will remember this

PR #843, opened 2021-07-11
HTTP solver response index shift

### LINKED ISSUE #837: Error reading HTTP solver solution when it contains Balancer pools.
When encoding the settlement for HTTP solver solutions that contain Balancer pools, it produces an `Invalid AMM` error because it cannot match a weighted pool ID with the corresponding Balancer pool. I believe this is related to different `usize => WeightedPool` maps getting used for the encoding and decoding, one with offset IDs, and one without.

Attached are sample instance and solution files.
[00_instance_raw.txt](https://github.com/gnosis/gp-v2-services/files/6793745/00_instance_raw.txt)
[05_solution_int.txt](https://github.com/gnosis/gp-v2-services/files/6793746/05_solution_int.txt)

bh2smith:
Fixes #837

We should have been accounting for the index shift both on request and response.

### Test Plan
Included regression test on the instance and solution data supplied in the issue

### DIFF
--- a/solver/src/liquidity.rs
+++ b/solver/src/liquidity.rs
@@ -117,14 +117,6 @@ pub struct ConstantProductOrder {
     pub settlement_handling: Arc<dyn SettlementHandling<Self>>,
 }
 
-impl PartialEq for ConstantProductOrder {
-    fn eq(&self, other: &Self) -> bool {
-        self.tokens.eq(&other.tokens)
-            && self.reserves.eq(&other.reserves)
-            && self.fee.eq(&other.fee)
-    }
-}
-
 impl std::fmt::Debug for ConstantProductOrder {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(f, "Constant Product AMM {:?}", self.tokens)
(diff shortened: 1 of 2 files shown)


solver/src/liquidity.rs
```diff
@@ -117,6 +117,14 @@ pub struct ConstantProductOrder {
     pub settlement_handling: Arc<dyn SettlementHandling<Self>>,
 }
 
+impl PartialEq for ConstantProductOrder {
```
fleupold, line 120: Does this mean a SushiSwap pool and a Uniswap v2 pool on the same token pair with same reserve values should be considered equal?

fleupold (commented):
Very nice unit test, and the shifting logic looks correct to me.

I don't think the `PartialEq` part is correct (but I'm also not sure it's needed given that we unshift the indices based on length and not based on equality).
If we want to build equality into the order model, I'd say it's fine to expose an opaque `id` field on the `ConstantProductOrder` and `WeightedProductOrder` subtype (similar to how we have it on `LimitOrder` (https://github.com/gnosis/oba-services/blob/4a2f0702014c24052909bfec6bf98b7ba50890f1/solver/src/liquidity.rs#L46-L47)

vkgnosis (approved):
I think the PartialEq impl should be removed because it is not used.

bh2smith:
> I think the PartialEq impl should be removed because it is not used.

It is used in the assertions in the unit test here.

vkgnosis:
I see. I think it would be better assert_eq just one of the inner values in the test in this case instead of a full custom PartialEq implementation. Because like Felix said the semantics of PartialEq aren't super clear. So we might not want to introduce this globally just for one test.

solver/src/liquidity.rs
bh2smith: I am going to remove this and compare field directly for the one unit test we needed this.

bh2smith:
> I don't think the `PartialEq` part is correct

Has been removed!

> I'd say it's fine to expose an opaque `id` field

Yes, this is something I would plan to do as a separate PR. Since this is working as expected for now, I would rather land this and tackle the IDs at a later point.

I believe this is ready for your approval!

fleupold (approved):

PR #845, opened 2021-07-12
Adds support to "Quasimodo" solver.

marcovc:
### Test Plan

I've run it locally with the solver server running locally as well.

### DIFF
--- a/solver/src/solver.rs
+++ b/solver/src/solver.rs
@@ -59,6 +59,7 @@ arg_enum! {
         OneInch,
         Paraswap,
         Matcha,
+        Quasimodo,
     }
 }
 
@@ -69,6 +70,7 @@ pub fn create(
     base_tokens: HashSet<H160>,
     native_token: H160,
     mip_solver_url: Url,
+    quasimodo_solver_url: Url,
     settlement_contract: &GPv2Settlement,
     token_info_fetcher: Arc<dyn TokenInfoFetching>,
     price_estimator: Arc<dyn PriceEstimating>,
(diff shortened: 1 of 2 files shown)


bh2smith (approved):
Is that all there is to it? Just instantiate and give it a URL? Is there no internal logic the driver need to implement in order to feed the solver with some data?

marcovc:
I guess that's all, really! It has essentially the same interface as the mip solver, so it just works. The set of parameters intersect for the important ones, for the others the defaults are enough for now.

vkgnosis (approved):

josojo:
Super nice! 
I was also able to run it e2e. I just run into a tiny issue that I documented here:
https://github.com/gnosis/gp-v2-solver-lib/issues/2

solver/src/solver.rs
```diff
@@ -111,6 +113,20 @@ pub fn create(
                 chain_id,
                 fee_discount_factor,
             )),
+            SolverType::Quasimodo => boxed(HttpSolver::new(
```
fleupold, line 116: nanonit: Instead of repeating the exact same arguments for 90% of the setup between MIP & Quasimode, you could pull this into a shared lambda or private method (so that we are guaranteed to update both solvers in case we want to change a value)

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia4.giphy.com%2Fmedia%2FNaboQwhxK3gMU%2Fgiphy-downsized-medium.gif&token=easy" data-gitmeme-token="easy"></a>

solver/src/solver.rs
marcovc: Good idea!

PR #853, opened 2021-07-13
Support deserializing EIP-1559 fee

vkgnosis:
This caused failures on rinkeby where the eip is already enabled. There
we failed deserializing the pending block because rust-web3's
transaction struct cannot handle it yet.

### Test Plan
new tests. also tested with real mainnet and rinkeby nodes.

### DIFF
--- a/solver/src/settlement_submission.rs
+++ b/solver/src/settlement_submission.rs
@@ -4,8 +4,8 @@ pub mod retry;
 
 use self::retry::{CancelSender, SettlementSender};
 use super::driver::solver_settlements::RatedSettlement;
-use crate::encoding::EncodedSettlement;
-use anyhow::{Context, Result};
+use crate::{encoding::EncodedSettlement, pending_transactions::Fee};
+use anyhow::{anyhow, Context, Result};
 use contracts::GPv2Settlement;
 use ethcontract::{dyns::DynTransport, errors::ExecutionError, Web3};
 use futures::stream::StreamExt;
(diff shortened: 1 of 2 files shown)


solver/src/settlement_submission.rs
```diff
@@ -114,8 +114,15 @@ async fn recover_gas_price_from_pending_transaction(
...
-        .find(|transaction| transaction.from == Some(*address) && transaction.nonce == nonce);
-    Ok(transaction.map(|transaction| transaction.gas_price))
+        .find(|transaction| transaction.from == *address && transaction.nonce == nonce)
+    {
+        Some(transaction) => transaction,
+        None => return Ok(None),
+    };
+    match transaction.fee {
+        Fee::Legacy { gas_price } => Ok(Some(gas_price)),
+        Fee::Eip1559 { .. } => Err(anyhow!("cannot handle eip 1559 fee")),
```
fleupold, line 126: Shouldn't we return `max_priority_fee_per_gas` here? Until we support EIP 1559 this will be equivalent to what gas price was before, and once we support it we will still have to outbid the previous priority_fee in order to incentivise the miner to replace our transaction.

fleupold (approved):
Nice, fast fix! Just a comment on what's correct to return in the EIP 1559 case.

solver/src/settlement_submission.rs
vkgnosis: I guess that would be correct too. The reason I did it this way is that this is only used for our own transactions. So assuming we only create non 1559 transactions the branch should never be taken anyway. If it is as simple as taking max priority then we might as well be more future proof.

fleupold: > assuming we only create non 1559 transactions the branch should never be taken anyway

Looking at the RPC response from the logs, you are right. Might be something we should double check across node implementations.

I read the pseudocode for `normalize_transaction` in https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md and thought that when we send a legacy transaction the resulting transaction object in the rpc response will contain what we specified as `gas_price` on  the `max_priority_fee_per_gas` field. Therefore I thought that as soon as EIP1559 is enabled this branch will be taken even when querying legacy transactions.

vkgnosis: Since it is more future proof and I don't see a drawback I have changed it to return the priority fee.

PR #854, opened 2021-07-14
Make bad token detector more generic over amm pool sources

vkgnosis:
Currently we use AmmPairProvider which is only implemented for Sushiswap
and UniswapV2. We introduce a trait to be more generic and can in the
future be implemented for more sources.

The plan is to use this to consider BalancerV2 pools as well.

### Test Plan
mainnet node test still works.

### DIFF
--- a/shared/src/bad_token/trace_call.rs
+++ b/shared/src/bad_token/trace_call.rs
@@ -6,7 +6,8 @@ use crate::{
 use anyhow::{anyhow, bail, ensure, Context, Result};
 use contracts::ERC20;
 use ethcontract::{
-    batch::CallBatch, dyns::DynTransport, transaction::TransactionBuilder, PrivateKey,
+    batch::CallBatch, dyns::DynTransport, errors::DeployError, transaction::TransactionBuilder,
+    PrivateKey,
 };
 use model::TokenPair;
 use primitive_types::{H160, U256};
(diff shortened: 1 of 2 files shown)


fleupold (approved):
Change looks good to me. The abstraction makes sense. 

The more logic we have to implement for the bad token detection, the more I wonder if we should eventually just simulate the baseline solution in the API to check if tokens are tradable?

shared/src/bad_token/trace_call.rs
```diff
@@ -16,6 +16,31 @@ use web3::{
...
+
+#[async_trait::async_trait]
+impl TokenOwnerFinding for AmmPairProviderFinder {
+    async fn find_candidate_owners(&self, token: H160) -> Result<Vec<H160>> {
+        Ok(self
+            .base_tokens
+            .iter()
+            .filter_map(|&base_token| TokenPair::new(base_token, token))
+            .map(|pair| self.inner.pair_address(&pair))
+            .collect())
```
bh2smith, line 40: I am not sure how this is finding potential owners of a token.... Oh! are we enumerating the pool addresses containing that token? Clever.

bh2smith (approved):
This is very cool, although I don't yet se the Balancer pools included in the mix. I will assume this is coming up next after first paving the way for it.

bh2smith:
> I wonder if we should eventually just simulate the baseline solution in the API to check if tokens are tradable?

I guess this suggestion would be to do this instead of token detecting, because otherwise working with just baseline would encounter the same problems we are seeing today (i.e. that it would work on paraswap or 1inch, but not with Baseline). Once we have integrated **all** the existing onchain liquidity sources, however, this may actually work. I suppose this is what you were hinting at.

vkgnosis:
> The more logic we have to implement for the bad token detection, the more I wonder if we should eventually just simulate the baseline solution in the API to check if tokens are tradable?

Good point. Not sure about the details. Will create issue to remember this.

PR #857, opened 2021-07-14
Use balancer vault as a source of tokens for bad token detection

vkgnosis:
The Vault has the balances for all pools so it makes sense to use it as a source for bad token detection.

### Test Plan
manually tested that bad token detector categorizes https://etherscan.io/token/0xfe18be6b3bd88a2d2a7f928d00292e7a9963cfc6 as good with this PR but not without it (there is no sushi and uniswap liquidity).
Note that this doesn't mean that the token can be traded. While the token itself is "good" we still cannot determine it's price in the native token (ETH) because the liquidity supported by the baseline solver is not good enough. Even with the existing balancer integration it won't be because we do not support stable pools.

### DIFF
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -18,7 +18,9 @@ use shared::{
     bad_token::{
         cache::CachingDetector,
         list_based::{ListBasedDetector, UnknownTokenStrategy},
-        trace_call::{AmmPairProviderFinder, TokenOwnerFinding, TraceCallDetector},
+        trace_call::{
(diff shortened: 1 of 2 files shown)


bh2smith (approved):
I guess the vault is an incredible source because it has custody over all tokens that exist in any balancer pool. Very cool. 

My first though was we could simplify some things by also using the settlement contract as an owner of some tradable tokens, but this doesn't prevent someone from transfering some "bad token" into either of these places. In fact it is quite common practice to mint and distribute non-transferable tokens to random addresses (as spam). I suspect the same could hold true for the balancer vault.

fleupold (approved):
Nice

PR #859, opened 2021-07-14
[Easy] index shift moved outside to become fixed constant

bh2smith:
We have recently seen some alerts involving `invalid AMM` which result from a non-constant index shift on the weighted and constant product pools. Because the shift was defined by the length of a mutable list we were getting index errors.

### Test Plan
Updated existing test to show no more index erroring.

### DIFF
--- a/solver/src/solver/http_solver/settlement.rs
+++ b/solver/src/solver/http_solver/settlement.rs
@@ -276,7 +276,7 @@ mod tests {
             fee: BigRational::new(3.into(), 1.into()),
             settlement_handling: wp_amm_handler.clone(),
         };
-        let weighted_product_orders = hashmap! { 1 => weighted_product_order };
+        let weighted_product_orders = hashmap! { 0 => weighted_product_order };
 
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):

taminomara (approved):

solver/src/solver/http_solver/settlement.rs
```diff
@@ -448,6 +449,20 @@ mod tests {
                         }
                     ]
                 },
+                "1": {
```
fleupold, line 452: was this test case failing without the one line change above? If not, can we please make it so?

fleupold (approved):

fleupold (commented):
Nice 🕵️  work!

solver/src/solver/http_solver/settlement.rs
bh2smith: It certainly was. Otherwise I wouldn't have added this.

Issue #856, opened 2021-07-14
Balancer - Filter Paused Pools.


bh2smith:
At the moment the balancer pool fetcher returns all balancer pools, but some of these could be "paused" which could result in failed settlement simulations (i.e. invalid settlements).

bh2smith:
Looking at a possible plan for implementation and I am beginning to think it might make sense to adapt the `BalancerPoolReserveCache` into a more general `BalancerPoolCache` which fetches other onchain data from the pool (i.e. not just the reserves).

bh2smith:
Looking at the docs a bit more, it seems like this won't be as simple as just fetching the paused status in our fetcher (at least not if we want to be efficient and mindful of our eth calls). The following snippets describe the process and show how info can be accessed will I will summarize here and propose a an approach for our use case.



Pausing Stipulations; https://github.com/balancer-labs/balancer-v2-monorepo/blob/6c9e24e22d0c46cca6dd15861d3d33da61a60b98/pkg/solidity-utils/contracts/helpers/TemporarilyPausable.sol#L20-L35

getPausedState() https://github.com/balancer-labs/balancer-v2-monorepo/blob/6c9e24e22d0c46cca6dd15861d3d33da61a60b98/pkg/solidity-utils/contracts/helpers/TemporarilyPausable.sol#L66-L83

setPausedState() https://github.com/balancer-labs/balancer-v2-monorepo/blob/6c9e24e22d0c46cca6dd15861d3d33da61a60b98/pkg/solidity-utils/contracts/helpers/TemporarilyPausable.sol#L85-L100

Note that there is an event emitted when a pool is paused, but this is not sufficient to determine if a pool becomes un-paused because of these pause windows. 

*It seems like we should check before querying if the pool is even still within its pause window.

## Key Observations
- We essentially want to know if a pool is active/paused or not on the fly so we can filter those which are not.
- A pool can only ever be paused within a certain time window (+ buffer period) from the point of the contract's initial deployment. After this period (even if the pool had been set to paused) it becomes un-paused and can no longer be paused again.
- The `_pauseWindowEndTime` and `_bufferPeriodEndTime` are set at the time of Pool's deployment so we cannot simply rely on the deployment block of each individual pool, but rather must know these values. In order to reduce our ethCalls, we may want to store these values on the `WeightedPool` and attempt to compute `is_active` from these values. Then we would only ever have to fetch pausedState during the pause window by calling `getPausedStatus`.

[...]

PR #860, opened 2021-07-14
[Balancer] Filter paused pools on fetch

bh2smith:
As a straightforward and simple solution to #856 we simply fetch the pause state of the pool at the same time we fetch the reserves. This can be done more efficiently in the future by taking advantage of the pool's pause window, so this doesn't exactly close the issue but temporarily mitigates it with very few code changes.

### Test Plan
I noticed when updating the current tests that we never actually test the Balancer pool's fetch method, but merely mock it and work with expected results (this can be seen from the fact that I introduced a paused pool to the updated test. 

I am not sure if this was intended (for some reason) or merely missed, so I will look into actually testing the fetch method with this PR.

_UPDATE_ Added unit test for filtering helper method.


vkgnosis (approved):

fleupold (approved):
Nice! (I'm still waiting on some help from the balancer team to test paused pools in practice)

PR #861, opened 2021-07-15
Move node based settlement submission into own file

vkgnosis:
Preparation for having a higher level struct that decides between node based and archer network based submission.

Code has only been moved, not changed.

### Test Plan
existing CI


fleupold (approved):
naming nit: Maybe (public)_mempool would be a slightly more descriptive name for this type of submission?

vkgnosis:
Renamed to that.

PR #862, opened 2021-07-15
Use native token for http solver price estimates

vkgnosis:
This makes the prices more consistent and understandable.

### Test Plan
small enough change to not warrant


marcovc (approved):
Thanks. Still doesn't explain the bogus exchange rates, right?

vkgnosis:
Not it doesn't.

bh2smith (approved):
Interesting. What was token0 before? Does this solve the issue @marcovc reported, or is this just something you noticed on the way.

marcovc:
I've added token[0] before (It doesn't really matter which value the prices are denominated in, except for convenience).

PR #865, opened 2021-07-15
Integrate archer network with driver

vkgnosis:
Adds a configuration option to enable submission with the archer
network. This is accomplished by a new struct SolutionSubmitter which
provides a uniform interface over the two types of transaction
submission.

In future PRs we could try advanced strategies like using archer by default but falling back to normal transactions if archer failed X times in a row maybe because their api is down.

### Test Plan
e2e tests should still work with the default transaction submission. i am going to do another manual test for the archer case.

### DIFF
--- a/solver/src/main.rs
+++ b/solver/src/main.rs
@@ -29,6 +29,7 @@ use solver::{
     liquidity::{balancer::BalancerV2Liquidity, uniswap::UniswapLikeLiquidity},
     liquidity_collector::LiquidityCollector,
     metrics::Metrics,
+    settlement_submission::{archer_api::ArcherApi, SolutionSubmitter, TransactionStrategy},
     solver::SolverType,
 };
 use std::{collections::HashMap, iter::FromIterator as _};
@@ -174,6 +175,21 @@ struct Arguments {
     /// price by 1% if built transactions don't actually get executed.
     #[structopt(long, env, default_value = "ParaSwapPool4", use_delimiter = true)]
     disabled_paraswap_dexs: Vec<String>,
+
+    /// The authorization for the archer api. If set the archer network will be used to submit
+    /// settlement transactions. If not set then transactions are sent out the normal way through
+    /// the node.
+    #[structopt(long, env)]
+    archer_authorization: Option<String>,
+
+    /// The maximum time we spend trying to settle a transaction through the archer network before
+    /// going to back to solving.
+    #[structopt(
+        long,
+        default_value = "60",
+        parse(try_from_str = shared::arguments::duration_from_seconds),
+    )]
+    max_archer_submission_seconds: Duration,
 }
 
 #[tokio::main]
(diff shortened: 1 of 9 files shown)


solver/src/main.rs
```diff
@@ -168,6 +169,21 @@ struct Arguments {
...
+    /// settlement transactions. If not set then transactions are sent out the normal way through
+    /// the node.
+    #[structopt(long, env)]
+    archer_authorization: Option<String>,
+
+    /// The maximum time we spend trying to settle a transaction through the archer network before
+    /// going to back to solving.
+    #[structopt(
+        long,
+        default_value = "120",
```
josojo, line 183: Maybe 30 secs is a better default, as the price can change a lot in 2 mins.

josojo (approved):
Nice!
Sure the current strategy is probably too simple. Looking forward to future strategies.

solver/src/main.rs
fleupold: Agree that 2 minutes is too long. Maybe we can aim for the same as "targetConfirmationTime"?

solver/src/main.rs
```diff
@@ -338,13 +354,27 @@ async fn main() {
         .await
         .map_err(|err| tracing::error!("Couldn't fetch market makable token list: {}", err))
         .ok();
+    let solution_submitter = SolutionSubmitter {
+        web3: web3.clone(),
+        contract: settlement_contract.clone(),
+        account,
+        gas_price_estimator: gas_price_estimator.clone(),
+        target_confirm_time: args.target_confirm_time,
```
fleupold, line 362: Are gas_price_estimator and target_confirm_time used in archer? I thought they provide an endpoint where they give us "tip" estimates?

fleupold (approved):
Code looks good to me.

solver/src/main.rs
vkgnosis: If the price changes so that the settlement becomes impossible then we cancel early anyway. This is just the maximum time we wait while the settlement would continue to simulate. For the public mempool this time is implicitly 5 minutes because that is what ethcontract sets by default.
With 30 seconds I am bit worried that there might be some delays with how the transaction gets propagated to the mining pools that make this not enough time. But I'm not sure what the best value is. Maybe something we need to test. Also wouldn't matter if we had this more advanced settlement - solution workflow that we talked about.

solver/src/main.rs
vkgnosis: They are used because we use the same gas estimator. I guess I didn't document this anywhere but when I tested their endpoint I didn't like that it was a private endpoint without documentation and that their estimates seemed way too high. According to the flash bots documentation (and this makes sense to me too) you just need to pay miners enough that they prefer your transaction over the currently lowest gas price transaction in the mining block. This is the same as for public mempool transactions so public gas price estimators should be fine.

vkgnosis:
Going to merge this so I can test on staging. We can of course still change the default config values.

PR #867, opened 2021-07-16
Adding app_data storage

josojo:
This PR adds the DB schema for storing app_data.

The app_data format was predefined [here](https://docs.google.com/document/d/1PsdWn9WPFf2BNeoJNtClyra8jNvWk_bJ-g33TbpUf8s/edit?ts=60ee0b05#): See sections Apps and Referred users. I used the CID (ipfs hash) as the unique identifier for all app_data's.

### Test Plan
Unit tests only. I also double-checked that ignored tests are already passing.

### Architecture design
I wasn't 100% sure whether the coded solution is the best way to store an AppData object containing a vector of another object: MetaData. I stored just MetaData and AppData as separate objects and then later merge them with two separate queries. Maybe there is a smarter way to do it.

###
This is only the first PR, an API endpoint for these objects will come next.

### DIFF
--- a/model/src/meta_data.rs
+++ b/model/src/meta_data.rs
@@ -1,0 +1,97 @@
+//! Contains the app_data file structures, which will also be stored in the data base and on ipfs
+
+use crate::h160_hexadecimal::{self};
+use anyhow::Result;
+use cid::multihash::{Code, MultihashDigest};
+use cid::Cid;
+use primitive_types::H160;
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+use thiserror::Error;
+
+#[derive(Error, Debug, Copy, Eq, PartialEq, Clone, Deserialize, Serialize, Hash)]
+pub enum MetaDataKind {
+    #[error("Referrer not found")]
+    Referrer,
+}
+
+#[serde_as]
+#[derive(Eq, PartialEq, Clone, Debug, Deserialize, Serialize, Hash)]
+#[serde(rename_all = "camelCase")]
+pub struct MetaData {
+    pub version: String,
+    pub kind: MetaDataKind,
+    #[serde(with = "h160_hexadecimal")]
+    pub referrer: H160,
+}
+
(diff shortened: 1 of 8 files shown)


model/src/meta_data.rs
```diff
@@ -0,0 +1,97 @@
...
+
+#[serde_as]
+#[derive(Eq, PartialEq, Clone, Debug, Deserialize, Serialize, Hash)]
+#[serde(rename_all = "camelCase")]
+pub struct AppData {
+    pub version: String,
+    pub app_code: String,
+    pub meta_data: Vec<MetaData>,
+}
+const RAW: u64 = 0x55;
```
bh2smith, line 36: What is this RAW for? Could we include a comment?

orderbook/src/database/app_data.rs
```diff
@@ -0,0 +1,280 @@
...
+    fn from(err: sqlx::Error) -> Self {
+        Self::DbError(err)
+    }
+}
+
+impl From<anyhow::Error> for InsertionError {
+    fn from(err: anyhow::Error) -> Self {
+        Self::AnyhowError(err)
+    }
+}
```
bh2smith, line 40:
```suggestion
}

```

bh2smith (approved):
All of this looks very good. Excellent work here. I only wonder why neither of these two new tables contain any information linking it to the orders table (i.e. `order_uid`). Is this planned for the future?

josojo:
>All of this looks very good. Excellent work here. I only wonder why neither of these two new tables contain any information linking it to the orders table (i.e. order_uid). Is this planned for the future?

The link will be made via the cid, which will be sent over with the order field app_data.

But I am realizing just now, that the app_data field needs to be an H256 and can not be a CID. This needs to be changed in the spec and in this implementation.
Please only continue reviewing, once I have fixed this. :)

josojo:
I had another call with Leandro. we are changing the spec and make it simpler. I will reopen another PR.

Issue #827, opened 2021-07-08
Support multiple accounts in solvers


taminomara:
The idea would be to be able to configure each solver with a separate account (i.e. private key) so that solutions for different solvers get submitted for with different accounts.

Initially, we would configure them to all use the same account, but once the Archer transaction service implementation is done, then we would be able to use many accounts (even without funds with MEV Geth bundles).

This would be super useful for our analytics, as well as start with an "informal" solver competition by using these different accounts to identify different DEXes that get used per settlement and make a sort-of DEX competition.

fleupold:
Plan sounds good! Just a small note, that as soon as EIP1559 goes live on mainnet (August 4th), accounts will still need funds despite using flashbot bundles. 0 gas transactions will no longer work after the upgrade as the `tx.origin` has to at least burn the base fee.

taminomara:
So far, I want to split solution into three parts:

1. extend solver interface so that solvers know their accounts,
2. use solver accounts when submitting settlements,
3. add a mechanism to configure each solver separately.

PR #868, opened 2021-07-16
Multiple accounts in solvers, step 2: driver honors solver's account

taminomara:
Part of #827.

Refactor driver to support different accounts for solvers. We drop solver names from most places, and replace them with unique solver IDs. Driver can look up solver by its ID and get its name, address, and other info.

### Test Plan
Unittests are in progress. Proper testing suite will require contract mocking, I will add it later.

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -1,6 +1,7 @@
 pub mod solver_settlements;
 
-use self::solver_settlements::{RatedSettlement, SettlementWithSolver};
+use self::solver_settlements::RatedSettlement;
+use crate::solver::{SolverId, Solvers};
 use crate::{
     chain,
     liquidity::Liquidity,
@@ -9,12 +10,10 @@ use crate::{
     settlement::Settlement,
     settlement_simulation,
     settlement_submission::{self, retry::is_transaction_failure},
-    solver::Solver,
 };
-use anyhow::{anyhow, Context, Error, Result};
+use anyhow::{Context, Error, Result};
 use contracts::GPv2Settlement;
 use ethcontract::errors::MethodError;
-use futures::future::join_all;
 use gas_estimation::GasPriceEstimating;
 use itertools::{Either, Itertools};
 use model::order::{OrderUid, BUY_ETH_ADDRESS};
@@ -30,14 +29,14 @@ use shared::{
 use std::{
     collections::{HashMap, HashSet},
     sync::Arc,
-    time::{Duration, Instant},
+    time::Duration,
 };
 
 pub struct Driver {
     settlement_contract: GPv2Settlement,
     liquidity_collector: LiquidityCollector,
     price_estimator: Arc<dyn PriceEstimating>,
-    solver: Vec<Box<dyn Solver>>,
+    solvers: Solvers,
     gas_price_estimator: Arc<dyn GasPriceEstimating>,
     target_confirm_time: Duration,
     settle_interval: Duration,
@@ -60,7 +59,7 @@ impl Driver {
         settlement_contract: GPv2Settlement,
         liquidity_collector: LiquidityCollector,
         price_estimator: Arc<dyn PriceEstimating>,
-        solver: Vec<Box<dyn Solver>>,
+        solvers: Solvers,
         gas_price_estimator: Arc<dyn GasPriceEstimating>,
         target_confirm_time: Duration,
         settle_interval: Duration,
@@ -80,7 +79,7 @@ impl Driver {
             settlement_contract,
             liquidity_collector,
             price_estimator,
-            solver,
+            solvers,
             gas_price_estimator,
             target_confirm_time,
             settle_interval,
@@ -109,38 +108,11 @@ impl Driver {
         }
     }
 
-    // Returns solver name and result.
-    async fn run_solvers(
-        &self,
-        liquidity: Vec<Liquidity>,
-        gas_price: f64,
-    ) -> impl Iterator<Item = (&'static str, Result<Vec<Settlement>>)> {
-        join_all(self.solver.iter().map(|solver| {
-            let liquidity = liquidity.clone();
-            let metrics = &self.metrics;
-            async move {
-                let start_time = Instant::now();
-                let result = match tokio::time::timeout(
-                    self.solver_time_limit,
-                    solver.solve(liquidity, gas_price),
-                )
-                .await
-                {
-                    Ok(inner) => inner,
-                    Err(_timeout) => Err(anyhow!("solver timed out")),
-                };
-                metrics.settlement_computed(solver.name(), start_time);
-                (solver.name(), result)
-            }
-        }))
-        .await
-        .into_iter()
-    }
-
     async fn submit_settlement(&self, rated_settlement: RatedSettlement) -> Result<()> {
-        let SettlementWithSolver { name, settlement } = rated_settlement.clone().settlement;
-        let trades = settlement.trades().to_vec();
+        let solver_id = rated_settlement.solver_id;
+        let trades = rated_settlement.settlement.trades().to_vec();
         match settlement_submission::public_mempool::submit(
+            self.solvers.account(rated_settlement.solver_id).clone(),
             &self.settlement_contract,
             self.gas_price_estimator.as_ref(),
             self.target_confirm_time,
@@ -150,10 +122,12 @@ impl Driver {
         .await
         {
             Ok(_) => {
-                trades
-                    .iter()
-                    .for_each(|trade| self.metrics.order_settled(&trade.order, name));
-                self.metrics.settlement_submitted(true, name);
+                trades.iter().for_each(|trade| {
+                    self.metrics
+                        .order_settled(&trade.order, self.solvers.name(solver_id))
+                });
+                self.metrics
+                    .settlement_submitted(true, self.solvers.name(solver_id));
                 Ok(())
             }
             Err(err) => {
@@ -168,7 +142,8 @@ impl Driver {
                 } else {
                     tracing::error!("Failed to submit settlement: {:?}", err)
                 };
-                self.metrics.settlement_submitted(false, name);
+                self.metrics
+                    .settlement_submitted(false, self.solvers.name(solver_id));
                 Err(err)
             }
         }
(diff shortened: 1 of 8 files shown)


solver/src/driver.rs
```diff
@@ -312,19 +287,21 @@ impl Driver {
                 .expect("Price of native token must be known.");
 
         futures::stream::iter(settlements)
-            .filter_map(|settlement| async {
-                let surplus = settlement.settlement.total_surplus(prices);
+            .filter_map(|settlement_with_solver_id| async {
+                let (solver_id, settlement) = settlement_with_solver_id;
```
bh2smith, line 291: any reason we don't just directly declare the tuple here?

solver/src/driver.rs
```diff
@@ -312,19 +287,21 @@ impl Driver {
...
+                let surplus = settlement.total_surplus(prices);
                 // Because of a potential fee discount, the solver fees may by themselves not be sufficient to make a solution economically viable (leading to a negative objective value)
                 // We therefore reverse apply the fee discount to simulate unsubsidized fees for ranking.
-                let unsubsidized_solver_fees = settlement.settlement.total_fees(prices) / BigRational::from_float(self.fee_discount_factor).expect("Discount factor is not a rational");
+                let unsubsidized_solver_fees = settlement.total_fees(prices) / BigRational::from_float(self.fee_discount_factor).expect("Discount factor is not a rational");
+                // TODO: can we reuse gas estimate from simulation?
```
bh2smith, line 296: At this moment I can't see why not. It would reduce our eth calls and possibly eliminate some of these

```
ERROR solver::driver: single run errored: failed to estimate gas price
Caused by:
   all gas estimators failed
```

solver/src/driver.rs
```diff
@@ -373,45 +350,59 @@ impl Driver {
...
+                liquidity,
+                gas_price_wei,
+                self.solver_time_limit,
+                Some(self.metrics.as_ref()),
+            )
             .await
-            .filter_map(solver_settlements::filter_bad_settlements)
-            .inspect(|(name, settlements)| {
-                for settlement in settlements.iter() {
-                    tracing::debug!("solver {} found solution:\n {:?}", name, settlement);
```
bh2smith, line 382: Are we removing this log then? I think we might still want to be informed when solutions are found.

solver/src/settlement_submission/public_mempool.rs
```diff
@@ -86,14 +85,6 @@ pub async fn submit(
     }
 }
 
-async fn transaction_count(contract: &GPv2Settlement) -> Result<U256> {
-    let defaults = contract.defaults();
-    let address = defaults.from.as_ref().unwrap().address();
-    let web3 = contract.raw_instance().web3();
-    let count = web3.eth().transaction_count(address, None).await?;
-    Ok(count)
-}
```
bh2smith, line 95: Looking back at this, I am not sure what we were even doing before, why would we use the settlement contract to get the transaction count of our solver address?

solver/src/solver.rs
```diff
@@ -33,26 +37,111 @@ mod solver_utils;
...
+    ///
+    /// This method is used in logging and metrics. By default, returns the type name.
     fn name(&self) -> &'static str {
         std::any::type_name::<Self>()
     }
 }
 
+/// Solver ID can be used to identify a solver within a [`Solvers`] collection.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub struct SolverId(usize);
```
bh2smith, line 64: Having an integer ID is cool and all, but I suspect we would like to see account address when logging these things. Perhaps we could implement `Display` for these solvers and include both `name` and `account`. These are the most useful values to have for debugging purposes.

solver/src/driver/solver_settlements.rs
```diff
@@ -1,79 +1,16 @@
...
-    }
-}
-
-impl SettlementWithSolver {
-    pub fn without_onchain_liquidity(&self) -> Self {
-        Self {
-            name: self.name,
-            settlement: self.settlement.without_onchain_liquidity(),
-        }
-    }
```
bh2smith, line 70: Have we removed this completely or has it been replaced somewhere I can't see at the moment? I mean do we still have the option to settle orders with the contract internal buffers?

bh2smith (approved):
Looking really good. I am excited for the switch to this! Some comments and questions left inline.

solver/src/settlement_submission/public_mempool.rs
taminomara: It's alright, we're using contract's default `from` address which is set to be the solver's address somewhere in the `main` function.

bh2smith: Yea, but it is certainly not the case that the deployer (assuming this is the contract's from address) of the settlement contract is always going to be a solver... seems very strange that we would ever have gone this direction here.

solver/src/driver/solver_settlements.rs
taminomara: Yes, I've just replaced `SettlementWithSolver` with `(SolverId, Settlement)`, so `SettlementWithSolver::without_onchain_liquidity` is now just `Settlement::without_onchain_liquidity`.

I'll add a unit test for this though.

solver/src/driver.rs
taminomara: Borrow checker refused to compile it for some reason. I suspect it may be a bug.

solver/src/settlement_submission/public_mempool.rs
taminomara: It's not the `deployer`, it's explicitly overridden [here][1]. I.e. every interaction with this contract is sent from the solver's address. AFAIK we don't even extract deployer's address from artifacts.

[1]: https://github.com/gnosis/gp-v2-services/blob/adc937bf4113694c53dc3980286835f481383ebe/solver/src/lib.rs#L27

solver/src/solver.rs
taminomara: Well, we don't really display it anywhere, we convert it to solver name instead. Although your comment gave me an idea how to implement it in a better way.

solver/src/settlement_submission/public_mempool.rs
bh2smith: Still, whatever we had here before seems awkward and unnecessarily complicated. Thanks for making it much better.

solver/src/driver.rs
vkgnosis: That error is from the gas price estimator but what is estimated here is the gas use. 
The gas use could probably be taken from the existing simulation. The problem is that the simulation at the moment uses "view methods" which don't return the gas used while the gas estimation uses the "estimate gas" rpc call which does. If the latter also checks for reverts just like view methods but in addition returns used gas then we could use it instead.

solver/src/solver.rs
```diff
@@ -33,26 +38,151 @@ mod solver_utils;
...
 
-    /// Solver's account that should be used to submit settlements.
+    /// Returns solver's account that should be used to submit settlements.
     fn account(&self) -> &Account;
 
-    /// Displayable name of the solver. Defaults to the type name.
+    /// Returns displayable name of the solver.
+    ///
+    /// This method is used in logging and metrics. By default, returns the type name.
     fn name(&self) -> &'static str {
```
vkgnosis, line 58: I think it probably makes sense to get to the extra information like name and Address through the new Solvers struct. Would we remove these methods in the future then and store Address, Name in the Solvers struct?

solver/src/driver.rs
```diff
@@ -373,45 +350,67 @@ impl Driver {
...
+                &mut settlements,
+            );
+
+            for settlement in &settlements {
+                tracing::debug!(
+                    "solver {} found solution:\n{:?}",
+                    self.solvers.name(solver_id),
+                    settlement
+                );
+            }
```
vkgnosis, line 392: I think this was printed before merging previously.

solver/src/solver.rs
taminomara: I'm not sure. These methods are just convenience shortcuts. Name and address is a property of a solver, why would we store them somewhere else.

In fact, I'm not sure that `Solvers` struct is the best approach. Having a reference to the solver instead of solver ID is more convenient. Another thing to take into account is this. If we configure solvers through config, then we may have multiple HTTP solvers with different names and URLs.

vkgnosis: We already have different HTTP solvers btw. Mip and Quasimodo.
If we do keep the information in the trait then I'm not sure if the Solvers struct is that useful. We could store solvers as Arc<dyn Solver> and directly use (Settlement, Arc< dyn Solver>) instead of the SolverId. Then we have all the information and less duplicated code.

taminomara: `Arc` is one way to go about it. Initially I didn't want to use `Arc` because we clone settlements here and there, so a lightweight identifier seemed preferable. But with current code structure (i.e. after `SettlementWithSolver` was replaced with a tuple), we can just use `&dyn Solver` everywhere.

fleupold: Maybe we can keep the SolverId struct as a value type that can be easily copiable for passing along with the settlement but make the solver expose a `solver_id` method that returns it. My worry about `&dyn Solver` would be that lifetimes may become challenging when we pass it around, but if this is not an issue then we can also do that.

fleupold (commented):
What I like about the `Solvers` struct (maybe we can think about another name) is that it takes logic/complexity out of the driver (which is way too big and poorly tested already).

We could even consider moving the filtering/merging logic into this "SolutionFinding" component (where the driver does more things such as LiquidityCollection, SolutionSubmission, inflight trade tracking) etc.

solver/src/solver.rs
nlordell: I'd also be OK with passing `Arc<dyn Solver>`s or `&dyn Solver`s around, then we can just read `account` and `name` from the solver directly.

> My worry about &dyn Solver would be that lifetimes may become challenging when we pass it around

AFAIU their lifetime is tied to `&self` that is passed in and also lives for the entire "run loop" future - so I'd imagine it wouldn't be too crazy.

That being said, the `SolverId(usize)` pattern is cool :smile:.

nlordell (commented):
Looks very good. Maybe see if its possible to pass `&dyn Solver`s around without too much hassle which would mean that we would not require a `SolverId`.

I also like the `Solvers` type, even if its just a `struct Solvers(Vec<Box<dyn Solver>>)` for the same reasons @fleupold mentioned: it moves code out of the driver module.

bh2smith (approved):

taminomara:
This PR is outdated, I'll send another (several of them, actually).

PR #869, opened 2021-07-16
[easy] Print solver type instead of http solver

josojo:
removes the custom name trait implementation, hence the name implementation will fall back to printing the solver type. This allows us to differentiate between quasimodo solver and mip solver

### DIFF
--- a/solver/src/solver/http_solver.rs
+++ b/solver/src/solver/http_solver.rs
@@ -59,6 +59,7 @@ impl SolverConfig {
 }
 
 pub struct HttpSolver {
+    name: &'static str,
     account: Account,
     base: Url,
     client: Client,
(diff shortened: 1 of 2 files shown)


vkgnosis:
Maybe it would be better to store a name string in the http solver struct and return that.

solver/src/solver/http_solver.rs
```diff
@@ -75,6 +76,7 @@ pub struct HttpSolver {
 impl HttpSolver {
     #[allow(clippy::too_many_arguments)]
     pub fn new(
+        name: &'static str,
         account: Account,
         base: Url,
         api_key: Option<String>,
```
bh2smith, line 82: General rust question:

How do we decide between when to use `String` or, in this case, `&'static str`?

vkgnosis (approved):

bh2smith (approved):

solver/src/solver/http_solver.rs
josojo: String is heap-allocated and potentially mutable. None is needed here.

PR #875, opened 2021-07-20
App Data Step 1: database schema and implementation for app_data

josojo:
This PR implements basic storage in the database of the tx app_data, as specified [here](https://docs.google.com/document/d/1PsdWn9WPFf2BNeoJNtClyra8jNvWk_bJ-g33TbpUf8s/edit?ts=60ee0b05#) It allows storing and retrieving of the data in the database.

### Test Plan
unit tests should be sufficient


vkgnosis:
We already store the existing app_data field that is part of the signed order data in the database. For the referral program this would be an ipfs hash that contains the extra information that we add a new table here for, right?
I wonder whether we actually need to have this information in the order book database. Wouldn't it be enough to extract this only when needed when we do evaluations on the orders for the referral program later?

josojo:
> We already store the existing app_data field that is part of the signed order data in the database. For the referral program this would be an ipfs hash that contains the extra information that we add a new table here for, right?

yes exactly

>I wonder whether we actually need to have this information in the order book database. Wouldn't it be enough to extract this only when needed when we do evaluations on the orders for the referral program later?

Yes, we can store it in any database. It's just very handy in the orderbook database, if we want to make queries like: give me all trades that have the referrer = 0x. The advantage of putting it in the orderbook database is that I can easily make SQL queries for that.

orderbook/src/database/app_data.rs
```diff
@@ -0,0 +1,206 @@
...
+
+impl From<anyhow::Error> for InsertionError {
+    fn from(err: anyhow::Error) -> Self {
+        Self::AnyhowError(err)
+    }
+}
+
+#[async_trait::async_trait]
+impl AppDataStoring for Postgres {
+    async fn insert_app_data(&self, app_data: &AppData) -> Result<H256, InsertionError> {
```
alfetopito, line 44: Won't the original data be stored anywhere?
Since we are storing the hash and the contents, we won't be able to recover it unless it's stored raw.

model/src/app_data.rs
```diff
@@ -0,0 +1,88 @@
...
+#[derive(Eq, PartialEq, Clone, Debug, Deserialize, Serialize, Hash)]
+#[serde(rename_all = "camelCase")]
+pub struct AppData {
+    pub version: String,
+    pub app_code: String,
+    pub meta_data: MetaData,
+}
+impl AppData {
+    pub fn sha_hash(&self) -> Result<H256> {
+        let string = serde_json::to_string(self)?;
```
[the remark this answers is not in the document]
josojo: yes, struct -> string -> sha is always deterministic, as the struct is unique.

but also the critical part: Json(string) -> struct is deterministic

orderbook/src/database/app_data.rs
josojo: We are storing the app_data struct. And since I can reconstruct the original json from the app_data(modulo ordering of parameters), I think there is no need to store the original data

model/src/app_data.rs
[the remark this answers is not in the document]
vkgnosis: I understand that it is deterministic in practice but I'm not sure if the serialization libraries guarantee this. Maybe a serde update would reorder the fields or otherwise change the serialization in a way that doesn't affect the semantics of the json.
And I guess this always has to match the frontend too? Maybe the spec should say exactly how the hash is calculated? For example we could say that instead of hashing the json string we create a new string that is the concatenation of version + app_code + referrer_as_hex or something like that.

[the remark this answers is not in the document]
josojo: >but I'm not sure if the serialization libraries guarantee this.
even if they don't guarantee it, I think that is quite unlikely that something will change. If worst comes to worst, we don't have to update the lib. 

But yeah, we could also go your other route. It just has one disadvantage, with each new version and new struct fields, we would have to change how it's being calculated.

wdyt @alfetopito ?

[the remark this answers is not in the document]
alfetopito: I would prefer to have a deterministic json serialization.

The suggestion to hash a subset of fields is limited to the fields we currently have, but we expect to have more in the future. It' would be a uneccessary maintenance burden IMO.

orderbook/src/database/app_data.rs
alfetopito: But this is under the assumption that only the referral address is relevant.
If we store the version and kind (it's another discussion what we should store), you'd have to flatten that structure too.

IMHO, a key-value store would make more sense for this sort of data (think of Redis/ElasticSearch/etc) as the doc format is easily queryable as is.

I just looked up and it can be stored and retrieved on Postgresql https://stackoverflow.com/questions/39010601/extract-key-value-from-json-objects-in-postgres https://www.postgresql.org/docs/9.3/functions-json.html

Doing that way can even reduce the need to store the other fields.

What do you think of this approach?

model/src/app_data.rs
```diff
@@ -0,0 +1,88 @@
...
+use cid::multihash::{Code, MultihashDigest};
+use primitive_types::{H160, H256};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+use std::convert::TryInto;
+
+#[serde_as]
+#[derive(Eq, PartialEq, Clone, Debug, Deserialize, Serialize, Hash)]
+#[serde(rename_all = "camelCase")]
+pub struct MetaData {
```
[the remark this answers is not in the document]
alfetopito: I replied to your question there.
I'm fine with the `kind` being gone, but I think the `version` has to stay.

orderbook/src/database/app_data.rs
josojo: I don't like, as we can not query easily all appdata that refer to a specific referrer. This is needed to build good queries later.

If we need the version or the kind, then I will also store them as field in the database

model/src/app_data.rs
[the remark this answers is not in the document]
vkgnosis: The question is what is deterministic json serialization. Afaik there is no such generally accepted standard and there are several ways you could implement the idea. The problem is that the library the frontend is using might not be doing the same thing as the one the backend is using even though both are trying to be deterministic. We would have to be sure that the implementations match but that probably involves implementing a whole json serializer. This is why I think it is easier to manually specify how the hashing works.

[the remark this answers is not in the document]
josojo: >We would have to be sure that the implementations match but that probably involves implementing a whole json serializer. 

I think that it is okay if we rely on API calls for that and have it only implemented in the backend. In the other pending PRs, the front-end can send any json and the backend would store it and send back its hash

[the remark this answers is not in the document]
vkgnosis: I'm not a fan of this approach. The way this hash is calculated feels like something that should be part of the spec and be "implementation defined". It should be possible without going through a running driver. And even that we'd still need to ensure that this doesn't change which serde_json doesn't guarantee. Alternatively there might be an approach that doesn't need this kind of hash? I'd have to think about it more.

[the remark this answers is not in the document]
josojo: >Alternatively there might be an approach that doesn't need this kind of hash?

We need to compress many data into one variable that can be put into appData of an order. I think we need some compression like sha hashes...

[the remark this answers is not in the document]
fleupold: The property ordering discussion is interesting!

Given that the API is just a temporal workaround until we can store this information on IPFS, we should maybe have it implement a similar interface as IPFS? 
That is you can "store" a blob (string) on the db and it will return you the key (IPFS hash) of what you stored. It's then up to the frontend to put that same hash on the order data.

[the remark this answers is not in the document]
josojo: I had a long call with @alfetopito and anna, and we concluded:

- we will try to minimize complexity now, as this is not an essential component
- for easy storage and easy queries, we will store most of the data just as a string blob, though we will keep the appHash, appCode, and referrer as separate fields to enable building easy queries for later.
- The appDataHash will just be calculated as the hash of the string blob stored. Even if this is results in different hashes for identical data - e.g. just the same values, but different order - we don't see a problem with that. Our queries and referral logic should still work.

Wdyt? Is there something we are missing?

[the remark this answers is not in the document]
fleupold: I like that idea. You could even reject the request to store the blob if it doesn deserialize into your well defined struct that we want to use. I think the main concern here was that the ordering of fields in the string shouldn't decide whether or not the blob is valid (as different encoding libraries might change that ordering).

database/sql/V010__create_app_data.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE app_data (
+    app_data_hash bytea PRIMARY KEY,
+    version bytea NOT NULL,
+    app_code bytea NOT NULL,
+    referrer bytea NOT NULL,
```
alfetopito, line 5: Both are option and should be nullable

anxolin: I would also add indexes for `app_code`, `referrer`

database/sql/V010__create_app_data.sql
```diff
@@ -0,0 +1,7 @@
...
+    referrer_version bytea NOT NULL
```
anxolin, line 6: Shouldn't we add the original blob to be able to reuploaded to IPFS?

model/src/app_data.rs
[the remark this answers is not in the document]
anxolin: My initial thinking is, frontend doesn't even need the API to attach the information. The only reason we are doing it is for "availability" reasons, to make sure that IPFS will always have this info.
Another reason to use the API is to make it also easier for Balancer.

However, strictly speaking, we don't need an API for just storing the IPFS, this can be indexed by the API pulling from the order. Since the API can also get the IPFS document easily and this is a simple JSON is easy to index it. 
Using a service like Pinata we can make sure the availability problem goes away and we would be more decentralized in info fetching. 

The issue with the code for me, is that you rely on generating the document yourself, which I think is constraining for these reasons: 
- Now you are the one defining the format of the document. If a client wants to add additional information, they can't. Think of a client wishing to add additional metadata that the API is still not aware, they won't be able to do that. The document was created in a flexible fashion, so we can keep indexers (like the API) backward compatible. New metadatas are just ignored. Futures versions of the API (or indexers can start to handle metadatas retroactivelly)
- Also, here you are the one defining the structure of the JSON, you will run into issues when you evolve this data format because if you "reprocess" old documents you will get different hash unless you start to add some workarounds and versioning things (I see there's no version in the metadata, so you might even have issues doing this)
- On top of this, if you don't upload the document to IPFS other indexers, like TheGraph, or any other will run into issues to get the metadata of the order

I'm happy either way, if we:
- Use the API, but we store the original file or the blob, and we hash the blob/file blindly, not a DTO on the API definition. If we go this path, I would consider a requirement that this hashing yields the same hash as if I would upload it to IPFS, so we make sure in the future we can upload all
- Not use the API for uploading the IPFS link, for that we can use a service like Pinata, and then the API only needs to index the hash from the signed order (same as all other indexers)


For example, for this document here:
```ts
{
  "version": "1.0.0",
  "appCode": "CowSwap",
  "metadata": {
    "referrer": {
      "version": "1.2.3",
      "kind": "Referrer",
      "referrer": "0x424a46612794dbb8000194937834250Dc723fFa5"
    }
  }
}
```

we would get `QmT8qCMvCtXtWpMeNwk7AeHHUnJNfv83DxDawLMEpMz694` https://gateway.pinata.cloud/ipfs/QmT8qCMvCtXtWpMeNwk7AeHHUnJNfv83DxDawLMEpMz694

anxolin (commented):
Raised some small concern, but most importantly, I'm not that worried about how we persist the Metadata-JSON, I would love to understand how we are doing the non-trivial part of counting:
- Number of Cows of your referrals
- Traded volume in USD of your referrals
- Number of trades of your referrals
- Your traded volume in USD
- Your number of trades

Also, how are we incrementing the counter of "number of referals", for example, I wouldn't cont users unless they trade, but probably we will also need some minimal trade size. 

There's other discussions important to me that are important to define like, where the USD estimation comes from, can we do it in smaller iterations, how do we handle low liquidity makets, etc.

For all this parts I would love to find:
- Pragmatic quicker solution
- Long term solution (I strongly believe investing in our infrastructure is key, most importantly USD estimations and volumes is something we should start building now even if it takes time)


For pragmatic, I'm fine with Dune, although Anna has the expectations to be able to see it in the website directly. We can do embedded, or hack things, maybe we can brainstorm, but I would be fine with some embedded dashboard with instructions on how to check you stats to start with. 

She mentioned that @josojo has a quick solution for this, so I was wondering about that solution.

model/src/app_data.rs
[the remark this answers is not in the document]
fleupold: Agree, that the API should not limit the format that can be uploaded but expect arbitrary strings. It may reject the string if it cannot extract the relevant information but should allow additional info.

The nice thing about uploading it to Pinata would be that aggregation of the referral information could theoretically be done in a completely separate service (e.g. based on the Graph trade data or trade event indexing and a IPFS client). But given we wanted this quick the easiest way seemed to do it in our API database directly.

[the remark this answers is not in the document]
anxolin: For me is fine to not upload to IPFS if we can upload it later retroactively. The point I tried to raise was that if we don't, at least we should make sure we hash the original JSON (as long as it has the right format and can extract at least the AppCode). This way, in the future when we implement the Graph, we can also upload all the old document so TheGraph can consume them. 

> The nice thing about uploading it to Pinata would be that aggregation of the referral information could theoretically be done in a completely separate service (e.g. based on the Graph trade data or trade event indexing and a IPFS client). But given we wanted this quick the easiest way seemed to do it in our API database directly.

Fine with me, we will also need to do a FK in the trades, right? 

Regarding the model, I think we will need a table with a mapping of user accounts to their referral. This is because, I believe that if we implement some affiliate logics in the backend we need to then:
- detect when is the first time a user trade
- If the trade is for a user that never traded before. Does it have referal information? If it does, we record it. Note that it shouldn't be relevant and we shouldn't record the referal associated to an user when the user traded before. It's meant to be an incentive only for new users.

josojo:
I updated the code according to this conclusion:
```
- for easy storage and easy queries, we will store most of the data just as a string blob, though we will keep the appHash, appCode, and referrer as separate fields to enable building easy queries for later.
- The appDataHash will just be calculated as the hash of the string blob stored. Even if this is results in different hashes for identical data 
- e.g. just the same values, but different order - we don't see a problem with that. Our queries and referral logic should still work.
```

and considered many other comments, like enforcing only the "version" field, but not the appData, or referral in the json format. 
@alfetopito @anxolin @e00E 

I am currently researching the best way to do all the usd volume estimates etc, but I think this will not influence this PR.

database/sql/V010__create_app_data.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE app_data (
+    app_data_hash bytea PRIMARY KEY,
```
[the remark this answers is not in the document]
josojo: I don't like IPFS_hash, as this quickly get's confused with their cid hashes

model/src/app_data.rs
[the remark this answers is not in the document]
nlordell: I'm also OK with leaving it out for now, and then having a DB migration script to add it once it becomes relevant.

model/src/app_data.rs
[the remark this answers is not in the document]
nlordell: I think we can use EIP-712 structured data hashing for generating deterministic hashes for structured data. This way we have an algorithm for generating the hash based on the data itself and not based on some canonical representation.

Additionally, we already support this kind of hashing for signing in both the code bases.

database/sql/V010__create_app_data.sql
```diff
@@ -0,0 +1,7 @@
...
+    version bytea NOT NULL,
```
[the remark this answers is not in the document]
nlordell: I'd also be OK with removing this (see my other comment :point_up:).

model/src/app_data.rs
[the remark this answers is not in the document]
josojo: @nlordell how would that work? We would first derive the hashStruct from the json object and then calculate the EIP-712 hash based on this struct? This would also result in different sha_hashes, if the hashes of two different permutations of app_data fields is calculated, right? 
But it solves the issues with the fact that just an additional empty_space already modifies the hash.

orderbook/src/database/app_data.rs
```diff
@@ -0,0 +1,238 @@
...
+                    .as_bytes(),
+            )
+            .bind(
+                app_data
+                    .metadata
+                    .unwrap_or_default()
+                    .referrer
+                    .unwrap_or_default()
+                    .referrer
+                    .as_ref(),
```
alfetopito, line 72: This is confusing :cold_sweat: 

I understand the model might be confusing too, but the referrer->referrer can look like a bug for someone else.
Maybe add a comment?

Also, what does `as_ref` do at the end?

triple `ref` :exploding_head:

orderbook/src/database/app_data.rs
```diff
@@ -0,0 +1,238 @@
...
+
+impl AppDataQueryRow {
+    fn into_app_data_blob(self) -> Result<AppDataBlob> {
+        let file_blob = String::from_utf8(self.file_blob)?;
+        Ok(AppDataBlob(file_blob))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
alfetopito, line 138: Can you add a test for inserting addData with the minimum required fields?

```json
{
  "version": "1.0.0",
  "metadata": {}
}

File: database/sql/V010__create_app_data.sql [hunk 10]
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE app_data (
+    app_data_hash bytea PRIMARY KEY,
+    app_code bytea,
+    referrer bytea,
+    file_blob bytea NOT NULL
```
alfetopito, line 5: Any reason to not use `jsonp` as the field type?

Since this is a json data field... https://stackoverflow.com/a/65271803/1272513

If we ever want to query the data, this will be useful

alfetopito (approved):
Much happier with the current state, we are all aligned now :)

Few minor comments, nothing blocking though

orderbook/src/database/app_data.rs
josojo: IF you want, we can call the referrer H160 object within the referrer object just `address` then it might be easier to read. 

In order to write it into the database as bytea, we need to hand over a slice of u8. This is what `as_ref` is doing: https://docs.rs/primitive-types/0.10.1/primitive_types/struct.H160.html

database/sql/V010__create_app_data.sql
josojo: This will require formatting the string into json.RawMessage before inserting. Hence, the original string with all its empty spaces, etc would not be retrievable. This would deviate from the "treating it as ipfs file content". But maybe this is not soo bad. Let me investigate.

nlordell: I'm OK with this, as we should always be able to recover the original IPFS file from the `app_data_hash` right (given known CID version and multi-codec)?

model/src/app_data.rs
```diff
@@ -0,0 +1,132 @@
...
+use primitive_types::{H160, H256};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+use std::convert::TryInto;
+
+#[serde_as]
+#[derive(Eq, PartialEq, Clone, Debug, Deserialize, Serialize, Hash, Default)]
+pub struct Referrer {
+    #[serde(with = "h160_hexadecimal")]
+    pub referrer: H160,
```
nlordell, line 15: maybe this can be called `address` to address @leafilipo's concern with the naming?

orderbook/src/database/app_data.rs
```diff
@@ -0,0 +1,238 @@
...
+    fn from(err: serde_json::Error) -> Self {
+        Self::ParsingStringError(err)
+    }
+}
+
+impl From<anyhow::Error> for InsertionError {
+    fn from(err: anyhow::Error) -> Self {
+        Self::AnyhowError(err)
+    }
+}
```
nlordell, line 46: you can use `thiserror` crate to derive some of this for you. Alternatively, you can just return an `anyhow::Error` unless you need to match on the specific error in the caller.

orderbook/src/database/app_data.rs
nlordell: > object just `address` then it might be easier to read.

:+1:

model/src/app_data.rs
```diff
@@ -0,0 +1,132 @@
...
+        assert_eq!(deserialized.get_app_data().unwrap(), expected);
+        let serialized = serde_json::to_value(expected).unwrap();
+        assert_eq!(serialized, value);
+    }
+
+    #[test]
+    fn deserialization_and_back_for_nearly_empty_data() {
+        let value = json!(
+        {
+            "appCode": serde_json::value::Value::Null,
```
nlordell, line 97: FOr some reason, I was under the impression you could just do:
```suggestion
            "appCode": null,
```

Might be wrong though.

nlordell (approved):

nlordell:
Looks good, just some minor suggestions.

bh2smith:
Note that, by the time this is merged, you will likely need to rename the flyaway migration file.

model/src/app_data.rs
alfetopito: Wrong tag, but I agree with the change.
cc @anxolin for visibility

database/sql/V010__create_app_data.sql
alfetopito: Hmm, yes good point.
If we do that then we are back at the deterministic serialization problem.
But also as we discussed, this might not be that big of a deal, if we have different hashed for same data as long as all versions properly tag the same user.

I'm in favor of the change.

cc @anxolin

model/src/app_data.rs
```diff
@@ -0,0 +1,132 @@
...
+pub struct Referrer {
+    #[serde(with = "h160_hexadecimal")]
+    pub address: H160,
+    pub version: String,
+}
+
+#[serde_as]
+#[derive(Eq, PartialEq, Clone, Debug, Deserialize, Serialize, Hash, Default)]
+pub struct Metadata {
+    pub referrer: Option<Referrer>,
```
fleupold, line 22: Is both metadata and referrer really independently optional? It just feels it bloats the code with a lot of unwraps.

orderbook/src/database/app_data.rs
```diff
@@ -0,0 +1,225 @@
...
+                    .as_bytes(),
+            )
+            .bind(
+                app_data
+                    .metadata
+                    .unwrap_or_default()
+                    .referrer
+                    .unwrap_or_default()
+                    .address
+                    .as_ref(),
```
fleupold, line 59: It looks like here you will insert the 0x0000 address by default. Is that intended? Shouldn't we just insert None if any of the optional fields are not set?

orderbook/src/database/app_data.rs
```diff
@@ -0,0 +1,225 @@
...
+use futures::StreamExt;
+use model::app_data::AppDataBlob;
+use primitive_types::{H160, H256};
+use std::borrow::Cow;
+use thiserror::Error;
+
+#[async_trait::async_trait]
+pub trait AppDataStoring: Send + Sync {
+    async fn insert_app_data(&self, app_data: &AppDataBlob) -> Result<H256, InsertionError>;
+    fn app_data<'a>(&'a self, filter: &'a AppDataFilter) -> BoxStream<'a, Result<AppDataBlob>>;
```
fleupold, line 13: Can't this also just be an async function that returns a Result? What is the value of returning a stream?

orderbook/src/database/app_data.rs
```diff
@@ -0,0 +1,225 @@
...
+                        return InsertionError::DuplicatedRecord(app_data_hash);
+                    }
+                }
+                InsertionError::DbError(err)
+            })?;
+        Ok(app_data_hash)
+    }
+
+    fn app_data<'a>(&'a self, filter: &'a AppDataFilter) -> BoxStream<'a, Result<AppDataBlob>> {
+        if let Some(app_data_hash) = &filter.app_data_hash {
```
fleupold, line 77: This if/else statement is maybe not the most elegant way to implement these filters - especially if we add more and start mixing and matching. Maybe check how we do it for nullable filter fields in the [order](https://github.com/gnosis/oba-services/blob/42483040f4d51f2b74f309e4b4751955345c0c9b/orderbook/src/database/orders.rs#L172-L177) table?

Also right now you would implicitly ignore the referrer field if the app_data_hash filter was set (in which case an enum would be the better filter type).

orderbook/src/database/app_data.rs
```diff
@@ -0,0 +1,225 @@
...
+            INSERT INTO app_data (\
+                app_data_hash, app_code, referrer, file_blob)\
+                VALUES ($1, $2, $3, $4);";
+        let app_data_hash = app_data_str.sha_hash()?;
+        sqlx::query(QUERY)
+            .bind(app_data_hash.clone().as_bytes())
+            .bind(
+                app_data
+                    .app_code
+                    .unwrap_or_else(|| "".to_string())
```
fleupold, line 49: We should be able to insert None here

orderbook/src/database/app_data.rs
```diff
@@ -0,0 +1,225 @@
...
+    pub app_data_hash: Option<H256>,
+    pub referrer: Option<H160>,
+}
+
+#[derive(Error, Debug)]
+pub enum InsertionError {
+    #[error("Parsing the string was not successful:")]
+    ParsingStringError(#[from] serde_json::Error),
+    #[error("Duplicated record for `{0}`:")]
+    DuplicatedRecord(H256),
```
fleupold, line 28: Not an issue, but curious if you thought about just ignoring duplicated records as idempotent?

database/sql/V012__create_app_data.sql
```diff
@@ -0,0 +1,11 @@
...
+    app_data_hash bytea PRIMARY KEY,
+    app_code bytea,
+    referrer bytea,
+    file_blob bytea NOT NULL
+);
+-- Get a specific referral.
+CREATE INDEX referrer ON app_data USING BTREE (referrer);
+
+-- Get a specific app_code.
+CREATE INDEX app_code ON app_data USING BTREE (app_code);
```
fleupold, line 11: nit: newline

fleupold (commented):
Mostly nits. My biggest worry is about inserting valid looking values (empty string, 0x address) when actually the field is nullable.

database/sql/V012__create_app_data.sql
bh2smith: Might also want to give the indexes different name than the fields themselves. At least from the docs this appears to be common practice.

orderbook/src/database/app_data.rs
bh2smith: I would think that the user would want to know if there was a duplicated record rather than just swallowing a second request, nothing doing anything and then not telling someone about it. Gives the front end the opportunity to say something like "yes we already know and have recorded your request".

fleupold: The other option would be to return 200 if the value already exist. Does the user care if they have previously stored the same referral data (would that change something about their workflow)? I'd think they want to tell the backend: "fyi here is the information attached to that app_id hash, if you already have it fine, if not store it", but I don't think they need to know which of the two is the case.

But again, this is not a big issue.

nlordell: I would actually suggest erroring with a [403 - Already Exists](https://developers.rebrandly.com/docs/403-already-exists-errors).

orderbook/src/database/app_data.rs
nlordell: Oh nice - because its nullable!

orderbook/src/database/app_data.rs
nlordell: The benefit I see is that it allows you to stream rows.

Otherwise, you could also just collect all rows into a `Vec` and make this an `async` function.

orderbook/src/database/app_data.rs
nlordell: I agree, if the column is nullable, we should probably insert null (i.e. `None`). Nice catch!

model/src/app_data.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub fn sha_hash(&self) -> Result<H256> {
+        // The following hash is the hash used by ipfs.
+        // The ipfs cid can be calculated by
+        // const RAW: u64 = 0x55;
+        // let hash = Code::Sha2_256.digest(&string.into_bytes());
+        // let cid = Cid::new_v1(RAW, hash);
+        let hash = Code::Sha2_256.digest(&self.0.clone().into_bytes());
+        let array: [u8; 32] = hash.to_bytes()[2..]
+            .try_into()
+            .map_err(|_| anyhow!("h256 has wrong length"))?;
```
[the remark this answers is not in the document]
nlordell: Yeah, you could use `expect` here to assert this never errors:
```suggestion
            .expect("sha256 hash unexpected length");
```

model/src/app_data.rs
```diff
@@ -0,0 +1,134 @@
...
+impl AppDataBlob {
+    pub fn sha_hash(&self) -> Result<H256> {
+        // The following hash is the hash used by ipfs.
+        // The ipfs cid can be calculated by
+        // const RAW: u64 = 0x55;
+        // let hash = Code::Sha2_256.digest(&string.into_bytes());
+        // let cid = Cid::new_v1(RAW, hash);
+        // In order to avoid json duplication, we are deriving the hash from the json object
+        let hash = Code::Sha2_256.digest(serde_json::ser::to_string(&self.0.clone())?.as_bytes());
+        let array: [u8; 32] = hash.to_bytes()[2..]
```
nlordell, line 49: ubernit: I found it a bit confusing that we skip the first two bytes of the digest, as in my mental model Sha2-256 digests are 32 bytes.

After some digging, it turns out that its because `digest` returns a `Multihash` and not the digest - so it includes the multihash code (0x12 for Sha2) and the length (as a variable length integer, so here just 1 byte `0x20`) followed by the actual 32-byte digest.


It turns out that `Multihash` has a [`digest`](https://docs.rs/multihash/0.14.0/multihash/struct.MultihashGeneric.html#method.digest) method that extracts just the digest from the multihash and might avoid some confusion:
```suggestion
        let array: [u8; 32] = hash.digest()
```

orderbook/src/database/app_data.rs
josojo: My plan was to usually return the STATUSCODE::CREATED, but if it already exists, I would return STATUSCODE::OK and not an error, as this is a legit usecase as Felix is elaborating

model/src/app_data.rs
josojo: Yes, this was the spec provided. This gives us the most flexibility in the future. Not sure whether all of that is needed at all, but at least this is what was specified and agreed upon

orderbook/src/database/app_data.rs
josojo: Yeah, one could go with both ways. I used streaming, as I am not aware of any challenges with that and because it was used in most other instances as well. I thought there is value in having similar interfaces. But I can change, it is not important to me at all.

fleupold: I didn't know we were already using the BoxStream for orders and trades. In that case, let's keep it as is.

model/src/app_data.rs
anxolin: For the DB it will probably make sense to have a table mapping users to users (Referal table). You add a referer there if the user included the referer in his first order.

Anyways, I understand this specific part is not related to the DB, is just a data object modeling a metadata attached to an order so looks good to me.

model/src/app_data.rs
```diff
@@ -0,0 +1,135 @@
...
+            "metadata": {
+              "referrer": {
+                "address":  "0x424a46612794dbb8000194937834250dc723ffa5",
+                "version": "0.3.4",
+              }
+            }
+        }
+        );
+        let app_data_blob = AppDataBlob(json);
+        let expected: H256 = "0xd0b33abbeb0349c0b9d8d1a9e1c38b28efbc9c48842c26697a41a463d019cb16"
```
anxolin, line 130: Shouldn't this be:

`0x474352A1F08C1FD73B7A450B8EC0B04DC164B2C131717AF15E71AB578B5C3A4F`

https://gateway.pinata.cloud/ipfs/QmT8qCMvCtXtWpMeNwk7AeHHUnJNfv83DxDawLMEpMz694

See: 
https://cid.ipfs.io/#QmT8qCMvCtXtWpMeNwk7AeHHUnJNfv83DxDawLMEpMz694

anxolin (approved):
Looks good! to me

I'm wondering what will be the code/logic when:
- u see a appCode in an order
- u see a trade for an order wich has an appCode

I assume these will come in upcomming ones.

model/src/app_data.rs
alfetopito: > only the first trade of the users that arrive with a link will get this

Actually, that comes in the order, so there can happen that an order has a referral but a trade is executed for another order without it.

fleupold: > For the DB it will probably make sense to have a table mapping users to users (Referal table). 

I was hoping we can compute most of these queries on the fly and only need to create materialised views when the performance need arises. 

Can a specific user only be referred by one address? What if I click a referral link three months after having first used CowSwap?

alfetopito: The idea so far:
1. Referral only applies to new users: someone who traded before is not qualified
2. Only the first referral address used will be considered: subsequent referrals are ignored

So essentially we'll look for referral only on the first trade of an user

database/sql/V012__create_app_data.sql
```diff
@@ -0,0 +1,11 @@
...
+    app_data_hash bytea PRIMARY KEY,
+    app_code bytea,
+    referrer bytea,
+    file_blob jsonb NOT NULL
+);
+-- Get a specific referral.
+CREATE INDEX referrer_index ON app_data USING BTREE (referrer);
+
+-- Get a specific app_code.
+CREATE INDEX app_code_index ON app_data USING BTREE (app_code);
```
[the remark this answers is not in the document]
bh2smith:
```suggestion
CREATE INDEX app_code_index ON app_data USING BTREE (app_code);

```

bh2smith (approved):
Looks good, but we will have to coordinate on the schema filename as we are also about to use `V012__` in #914. If this is near completion, I would think that PR will be merged as soon as it gets another approval. Happy to rename on my side if this is merged first, but we need to keep in mind to check whats-what at the time of merging so we don't get another gap in our versioning.

nlordell:
@josojo do you think that we can close this now that the affiliate program has its own service?

PR #880, opened 2021-07-20
Update dependencies

vkgnosis:
Combination of the dependabot PRs to save CI time.

### Test Plan
CI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,9 +14,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.15.1"
+version = "0.15.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


vkgnosis:
Turns out I can't merge without review so I guess we'll wait.

bh2smith (approved):
This is fine, but the diff shows only one crate being updated.

vkgnosis:
The other updates are in Cargo.lock which Github hides by default.

bh2smith:
Yea, but shouldn't the toml files show the updated dependencies as well? Or after they blanketed by a > or Something?

vkgnosis:
If you write 1.3 in the toml file then Cargo takes that to mean anything that is compatible with so it could be 1.3, 1.4, etc. We usually don't specify the patch version number so it also allows 1.3.1, 1.3.2 etc.

PR #881, opened 2021-07-20
Specify archer authorization and transaction strategy separately

vkgnosis:
This makes it easier to manage in the kubernetes setup.

### Test Plan
manually tested with these arguments


josojo (approved):
Nice

PR #884, opened 2021-07-22
Use gas price in settlement simulations

### LINKED ISSUE #883: Pass gas_price when simulating settlements
We had about 20 consecutive transaction failures in prod today, because we received a solution from 1Inch that simulated successfully (https://dashboard.tenderly.co/felix/gpv2/simulator/19226c6a-0756-4d6d-a7cc-66cab4f5c465) but in reality failed to execute on the same block: https://dashboard.tenderly.co/tx/mainnet/0x5b19de1d7c6870a5f6133f0bb37546bc2aacf65674294901c9f69a708e88efd5

The reason is that 0x computes its protocol fee based on the current `tx.gasPrice` (https://github.com/0xProject/protocol/blob/24397c51a8c7bf704948c8fc6874843bccd5d244/contracts/exchange-libs/contracts/src/LibFillResults.sol#L88)

While we shouldn't really get these types of routes from dex aggregators in the first place (they should also set a reasonable gas price on their simulations), it would probably be wise to defend against this kind of issue by setting a gas price when simulating transactions (https://github.com/gnosis/oba-services/blob/9b409b1d2de22fcaa63adf601c9dce61ddc51697/solver/src/settlement_simulation.rs#L34)

We already have access to the current gas price in the driver (needed to pass into the solvers for the objective criterion) so this shouldn't be a big change.

It's not super urgent, unless we see this issue coming up again.

vkgnosis:
Some contracts react to the gas price which can cause settlement
failures for transactions that simulate correctly at 0 gas price.

closes #883 

I haven't manually tested that this catches transactions like in the issue but it is a straightforward change, just forwarding the gas price we  already use in the driver.

### Test Plan
ci still passes.


solver/src/settlement_simulation.rs
```diff
@@ -113,6 +118,7 @@ mod tests {
             &web3,
             network_id.as_str(),
             Block::FixedWithTenderly(block),
+            0.0,
```
josojo, line 121: I am surprised that this works. But I guess for simulation against the test env the gas price is not relevant.

josojo (approved):
Was just about to tackle it myself. 

Yes neat! easy fix

fleupold (approved):
Thanks for fixing

PR #886, opened 2021-07-23
[config] Use same http client everywhere and enforce timeouts

vkgnosis:
reqwest::Client (the http client we use) is a heavy object that is
internally an Arc and should be reused by cloning when possible to save
resources.
This commit makes all places in the code that were creating their own
client use one that is passed in. We create that client in the
respective main functions.
As a side effect this allows us to easily set a global timeout for all
http requests. This is important as otherwise a request can potentially
take forever blocking whatever execution path it is on.
As part of this I have reduced the timeout setting to one. If it turns
out that we need more finegrained timeout in some places we can add
those back as users of the client can always manually set a timeout per
request.

Potential config change because the timeout option has been renamed.

### Test Plan
CI still works. I also ran all of the #ignored unit tests  to make sure that they still work.


shared/src/http.rs
```diff
@@ -1,20 +0,0 @@
...
-pub fn default_http_client() -> Result<Client> {
-    Ok(ClientBuilder::new()
-        .user_agent("gp-v2-services/2.0.0")
-        .default_headers(
-            vec![(ACCEPT, HeaderValue::from_static("application/json"))]
-                .into_iter()
-                .collect(),
-        )
-        .timeout(Duration::from_secs(60))
-        .https_only(true)
```
vkgnosis, line 18: I have reused the same user agent for the client we create in main. I have not reused the header and https_only because I do not believe that they are needed. We didn't discuss those in the PR that introduced this function and all of the solvers that used to use this function (paraswap, matcha, 1nch) still pass their #ignored tests that talk to the real api.

shared/src/arguments.rs
```diff
@@ -20,14 +20,13 @@ pub struct Arguments {
...
-    /// Timeout for web3 operations on the node in seconds.
+    /// Timeout for all http requests.
     #[structopt(
             long,
-            env = "NODE_TIMEOUT",
             default_value = "10",
             parse(try_from_str = duration_from_seconds),
         )]
-    pub node_timeout: Duration,
+    pub http_timeout: Duration,
```
fleupold, line 29: Maybe we should flag in the PR title that this requires a config change?

solver/src/main.rs
```diff
@@ -211,9 +202,16 @@ async fn main() {
...
     let metrics = Arc::new(Metrics::new(&registry).expect("Couldn't register metrics"));
 
-    // TODO: custom transport that allows setting timeout
-    let transport =
-        create_instrumented_transport(HttpTransport::new(args.shared.node_url), metrics.clone());
+    let client = reqwest::ClientBuilder::new()
+        .timeout(args.shared.http_timeout)
+        .user_agent("gp-v2-services/2.0.0")
+        .build()
+        .unwrap();
```
fleupold, line 209: nit: maybe move these 4 lines into shared so that we have the same config between API and solver until we need changes?

fleupold (approved):
Looks good to me! Thanks for addressing it so quickly.

shared/src/arguments.rs
vkgnosis: I don't think we set it but anyway yes I should flag it.

solver/src/main.rs
vkgnosis: changed

bh2smith (approved):

PR #890, opened 2021-07-26
Respected Execution Plan Ordering

### LINKED ISSUE #889: Respect Execution Plan for HTTP Solver when mixing Weighted and Constant Pools
The following settlement failed simulation: https://dashboard.tenderly.co/gp-v2/staging/simulator/3382be4f-81b0-42e3-b5ff-37da31926433

We can see that we first try to interact with Uniswap but the settlement contract doesn't have enough sell balance for that:

image

The execution plan of the MiP [solver's solution file](http://gnosis-europe-gpv2-solver.s3-eu-central-1.amazonaws.com/data/prod/2021/07/23/17%3A36%3A56.846593354_UTC_Ethereum_1/05_solution_int.json) shows that we should have first interacted with the weighted balancer pool and then invoked the constant product pool.

Checking the code in the HTTP solver we only respect the execution plan within each group of pools but not across pools:

https://github.com/gnosis/oba-services/blob/9b409b1d2de22fcaa63adf601c9dce61ddc51697/solver/src/solver/http_solver/settlement.rs#L162-L163

Therefore we always execute all constant product interactions first.

bh2smith:
Fixes #889

Previously we had been losing the execution plan ordering by returning two separate vectors of executed Amms. Now, however, we return a single vector of a new enum which holds both types of executed amms.

### Test Plan
Existing tests adapted to new types.

### DIFF
--- a/solver/src/solver/http_solver/settlement.rs
+++ b/solver/src/solver/http_solver/settlement.rs
@@ -53,24 +53,17 @@ impl ExecutedLimitOrder {
     }
 }
 
-#[derive(Clone, Debug)]
-struct ExecutedConstantProductAmm {
-    order: ConstantProductOrder,
+#[derive(Clone)]
+struct ExecutedAmm {
     input: (H160, U256),
     output: (H160, U256),
+    order: ExecutedOrder,
 }
 
-#[derive(Clone, Debug)]
-struct ExecutedWeightedProductAmm {
-    order: WeightedProductOrder,
-    input: (H160, U256),
-    output: (H160, U256),
-}
-
-#[derive(Clone, Debug)]
-enum ExecutedAmm {
-    CP(ExecutedConstantProductAmm),
-    WP(ExecutedWeightedProductAmm),
+#[derive(Clone)]
+enum ExecutedOrder {
+    ConstantProduct(ConstantProductOrder),
+    WeightedProduct(WeightedProductOrder),
 }
 
 impl IntermediateSettlement {
@@ -100,26 +93,16 @@ impl IntermediateSettlement {
             settlement.with_liquidity(&order.order, order.executed_amount())?;
         }
         for executed_amm in self.executed_amms.iter() {
-            // This match only serves the purpose of unwrapping two types with the same fields.
-            // Thus, we find identical code blocks in each arm.
-            match executed_amm {
-                ExecutedAmm::CP(amm) => {
-                    settlement.with_liquidity(
-                        &amm.order,
-                        AmmOrderExecution {
-                            input: amm.input,
-                            output: amm.output,
-                        },
-                    )?;
+            let execution = AmmOrderExecution {
+                input: executed_amm.input,
+                output: executed_amm.output,
+            };
+            match &executed_amm.order {
+                ExecutedOrder::ConstantProduct(liquidity) => {
+                    settlement.with_liquidity(liquidity, execution)?
                 }
-                ExecutedAmm::WP(amm) => {
-                    settlement.with_liquidity(
-                        &amm.order,
-                        AmmOrderExecution {
-                            input: amm.input,
-                            output: amm.output,
-                        },
-                    )?;
+                ExecutedOrder::WeightedProduct(liquidity) => {
+                    settlement.with_liquidity(liquidity, execution)?
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


solver/src/solver/http_solver/settlement.rs
```diff
@@ -54,18 +53,26 @@ impl ExecutedLimitOrder {
...
+struct ExecutedWeightedProductAmm {
     order: WeightedProductOrder,
     input: (H160, U256),
     output: (H160, U256),
 }
 
+#[derive(Clone, Debug)]
+enum ExecutedAmm {
+    CP(ExecutedConstantProductAmm),
+    WP(ExecutedWeightedProductAmm),
```
vkgnosis, line 73: We could move the whole struct declaration here instead of keeping it separate:

```rust
enum ExecutedAmm {
    ExecutedConstantProductAmm {
        ...
    },
    ExecutedWeightedProductAmm { ... }
}
```

Not necessary just something I noticed because we never use the structs standalone.

bh2smith: Cool, I wasn't aware of this! Will give it a try.

vkgnosis: Alternatively you could do this:

```rust
struct ExecutedAmm {
    input: (H160, U256),
    output: (H160, U256),
    order: ExecutedOrder
}

enum ExecutedOrder {
    ConstantProduct(ConstantProductOrder),
    WeightedProduct(WeightedProductOrder),
}
```

Which would help with your comment below that you need identical code in two match branches. With this struct input and output are always available.

solver/src/solver/http_solver/settlement.rs
```diff
@@ -201,12 +206,18 @@ fn match_settled_prices(
         .flat_map(|order| {
             iter::once(order.order.buy_token).chain(iter::once(order.order.sell_token))
         })
-        .chain(executed_amms.0.iter().flat_map(|amm| amm.order.tokens))
         .chain(
             executed_amms
-                .1
                 .iter()
-                .flat_map(|amm| amm.order.reserves.keys().copied().collect::<Vec<H160>>()),
```
vkgnosis, line 209: I know this was already here and isn't related to this PR but I'm wondering now if it is correct that the prices hashmap we return from this function contains all of the reserve keys? Doesn't this only need to contain the prices from limit orders because those are what the smart contract needs?

solver/src/solver/http_solver/settlement.rs
bh2smith: I had also tried this in the past, but there was something about it that wasn't working well. However I will try your second suggestion now. Since I am having trouble accessing the fields of the structs with your first suggestion.

solver/src/solver/http_solver/settlement.rs
bh2smith: Great point. I will check this out!

vkgnosis (approved):

solver/src/solver/http_solver/settlement.rs
bh2smith: So, given the function signature
```rust
fn match_settled_prices(
    executed_limit_orders: &[ExecutedLimitOrder],
    executed_amms: &[ExecutedAmm],
    solver_prices: HashMap<H160, Price>,
) -> Result<HashMap<H160, U256>> {
```

It would be possible to get the tokens from the Limit order however, I suspect we might run into issues where the execution plan routes though a token which is not included in the limit orders.

On the other hand, we shouldn't need to supply prices for these amm-only tokens, so I will create a separate issue for this and experiment with it in a following PR.

bh2smith: Issue filed here: #892

fleupold (approved):
Looks good. Can we maybe add a unit test (or adopt the existing one that is verifying the execution plan is respected) in a way that would have failed before this PR?

bh2smith:
> Can we maybe add a unit test (or adopt the existing one that is verifying the execution plan is respected) in a way that would have failed before this PR?

I believe the existing test that was adapted does indeed show that that ordering is respected (note the change of index in the unit test case and how the `wp_order` comes before the `cp_order`).

fleupold:
> I believe the existing test that was adapted does indeed show that that ordering is respected

I think it would be slightly more robust to test that mixing and matching execution plans (e.g. Weighted, Constant, Weighted, Constant) works as expected. In order to make the existing test code pass all we would have to do is process weighted pools before constant pools (but not actually fix the logic).

But not a big deal.

bh2smith:
> test that mixing and matching execution plans (e.g. Weighted, Constant, Weighted, Constant) works as expected.

I can try and come up with something for this in a follow up.

> ... all we would have to do is process weighted pools before constant pools (but not actually fix the logic).

Not sure I follow your thought process here.

fleupold:
> Not sure I follow your thought process here.

In test driven development you usually try to make the smallest possible change in order to make the failing test pass. In the current regression test the smallest change to make it pass would have been to inverse the order in which we add constant/weighted interactions to the execution plan. This would not fix the logic. As a long shot if in some later refactoring someone reverts this change and they coincidentally process the weighted pools before the constant ones no test would fail.

Not a big deal and kind of a long shot but just wanted to point out that the regression test could be stronger.

bh2smith:
I still think that including those match statements on the amm order types is a small change that would have failed with the big we just addressed.

Issue #779, opened 2021-06-23
Optimize price vector in settlement


fedgiac:
The price vector in the settlement contract only needs to contain a price for any token that appears in a trade.
Currently, we add unnecessary prices in some cases:
1. if an order is traded through some long Uniswap path, each token in the path is added to the price list.
2. if ETH is used, then the address for WETH is always included. 

The effect of this change would be a slightly smaller gas cost for the settlements.

<details><summary>Example of a real settlement using six tokens but needing only two</summary>

Calldata:
```
[...]

PR #893, opened 2021-07-26
Remove iteteration over AMM tokens

### LINKED ISSUE #892: [HTTP Solver] Gather price map from Limit Order Tokens
It was pointed it in https://github.com/gnosis/gp-v2-services/pull/890#discussion_r676572331 that we may not need to derive the keys of the price map from the amm's tokenset (espesially since balancer pools may include more than just token pairs), but rather construct this map from the Order set. This seems completely viable given that we don't even enforce uniform clearing prices for amm only tokens. 

Specifically, we should be able to derive executed tokens in the following snippet from `executed_limit_orders` rather than `executed_amms`:

https://github.com/gnosis/gp-v2-services/blob/8471398e4a4e779374fde95d0e98109f2bde5daf/solver/src/solver/http_solver/settlement.rs#L185-L220

bh2smith:
Fixes #892

Not sure yet if prices for irrelevant intermediary tokens are actually necessary to pass into a settlement, but e2e here should tell us.

### Test Plan
May need to create a new e2e test that involves an amm token which is not actually traded directly.

### DIFF
--- a/solver/src/solver/http_solver/model.rs
+++ b/solver/src/solver/http_solver/model.rs
@@ -120,7 +120,7 @@ pub struct SettledBatchAuctionModel {
 impl SettledBatchAuctionModel {
     pub fn has_execution_plan(&self) -> bool {
         // Its a bit weird that we expect all entries to contain an execution plan. Could make
-        // exec_plan not optional and assert that the vector of execution updates is non-empty
+        // execution plan required and assert that the vector of execution updates is non-empty
         self.amms
             .values()
             .flat_map(|u| &u.execution)
--- a/solver/src/solver/http_solver/settlement.rs
+++ b/solver/src/solver/http_solver/settlement.rs
@@ -8,10 +8,7 @@ use anyhow::{anyhow, ensure, Result};
 use itertools::Itertools;
 use model::order::OrderKind;
 use primitive_types::{H160, U256};
-use std::{
-    collections::{hash_map::Entry, HashMap},
-    iter,
-};
+use std::collections::{hash_map::Entry, HashMap};
 
 // To send an instance to the solver we need to identify tokens and orders through strings. This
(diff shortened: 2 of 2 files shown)


fleupold (approved):
Maybe @fedgiac can confirm but the smart contract should not require prices for intermediate hops (only for the order's buy and sell token to compute the exchange rate and calculate the transfer amounts). The transfer amounts for interactions should be part of the interaction calldata directly.

fedgiac (approved):
Confirmed, amm prices are not needed. In principle we can use different prices in AMMs than those we give in the price vector.
I suppose this PR partially fixes #779?

bh2smith:
Thanks for confirmation @fedgiac - I would say that this change is actually sufficient enough to close #779 - since I would imagine the WETH price is needed when buying ETH anyway. Would you agree, or do you think we should still tackle the last remaining bit of the issue before closing it?

fedgiac:
WETH price is not needed when buying ETH, since ETH price is specified by pricing the "token" `0xeee..eee`.

bh2smith:
> WETH price is not needed when buying ETH, since ETH price is specified by pricing the "token" `0xeee..eee`.

Do you think that perhaps this solution will also not include the WETH price since it only involves the buy and sell token of the orders themselves? Or is this some sort of automagic conversion happening under the hood?

fedgiac:
If I recall it correctly, we automatically convert all ETH orders to WETH orders and later add back the ETH token and unwrapping if needed. So I'd expect that while picking the buy token here you get WETH.

bh2smith:
> So I'd expect that while picking the buy token here you get WETH.

In that case it might not be worth our efforts to determine if the price should be weth or eth given that we would have to fetch additional info from outside the scope of this method.

PR #897, opened 2021-07-27
Implement Liveness Check

### LINKED ISSUE #885: Add health check endpoint
Today we saw that the driver run loop did not continue running for reasons we don't know yet. The pod on the whole was fine and continued to emit debug logs but the run loop did not progress.
One way to detect and automatically react to this would be to expose a health check endpoint which could record the last time the driver started a run loop. If the time is older than X we could automatically restart the pod.

fleupold:
Fixes #885 

This PR adds and implements a liveness checking trait which we can use in our kubernetes setup to monitor the health of our pods and in case of deadlocks automatically restart.

For the orderbook, we reuse our readiness check which is basically just verifying that we can fetch solvable orders.

For the solver we record the last time at which we have completed a single solver runloop. If larger than some thresholf (7 minutes for now) we fail the liveness check.

Once this PR is merged, we will need to add the liveness check in our staging environment and next week also to our prod environment.

### Test Plan
Running locally, `http://localhost:9587/liveness` for solver and `http://localhost:9586/liveness` for orderbook. Change MAX_RUNLOOP_DURATION to 3s and observe that depending on when the last runloop completed the query sometimes returns 200 and sometimes 503

### DIFF
--- a/orderbook/src/lib.rs
+++ b/orderbook/src/lib.rs
@@ -33,7 +33,7 @@ pub fn serve_task(
 ) -> JoinHandle<()> {
     let filter = api::handle_all_routes(
         database,
-        orderbook,
+        orderbook.clone(),
         fee_calculator,
         price_estimator,
         metrics,
@@ -44,7 +44,7 @@ pub fn serve_task(
 
     tracing::info!(%metrics_address, "serving metrics");
     metrics_address.set_port(DEFAULT_METRICS_PORT);
-    serve_metrics(registry, metrics_address)
+    serve_metrics(registry, orderbook, metrics_address)
 }
 
 /**
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -13,8 +13,8 @@ use model::{
 };
 use primitive_types::{H160, U256};
 use shared::{
-    bad_token::BadTokenDetecting, maintenance::Maintaining, time::now_in_epoch_seconds,
-    web3_traits::CodeFetching,
+    bad_token::BadTokenDetecting, maintenance::Maintaining, metrics::LivenessChecking,
+    time::now_in_epoch_seconds, web3_traits::CodeFetching,
 };
 use std::{
     collections::{HashMap, HashSet},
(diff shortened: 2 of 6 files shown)


bh2smith (approved):
Seems reasonable!

I tried out the test plan to see if we would get what we expect on a void instance of each (i.e. an instance that is essentially responsible for nothing and an empty DB).

orderbook/src/lib.rs
```diff
@@ -44,7 +44,7 @@ pub fn serve_task(
 
     tracing::info!(%metrics_address, "serving metrics");
     metrics_address.set_port(DEFAULT_METRICS_PORT);
-    serve_metrics(registry, metrics_address)
+    serve_metrics(registry, orderbook, metrics_address)
```
nlordell, line 47: Maybe I'm missing something obvious, but to me it looks like only the `orderbook` is used for liveliness checks.

josojo: Yeah, you are right. But the orderbook is the essential object and 'get_solvable_orders' also checks the working of the db. I think it would be sufficient. What else would you recommend to check for the healthiness of the orderbook server?

josojo (approved):
LGTM

orderbook/src/lib.rs
nlordell: From what I understood from the PR, the liveliness check should also fail if the runloop hasn't been executed in the past X amount of time.

fleupold: The API crate doesn't really have a runloop (or do you mean the maintenance job?). For the solver I implemented the trait on the main driver runloop.

nlordell: THis is what I was missing:

https://github.com/gnosis/gp-v2-services/blob/35a3dcf4558d0fb699b2e75b2fba251d2598a358/solver/src/main.rs#L419

Sorry for the confusion!

nlordell (approved):

orderbook/src/lib.rs
fleupold: I do like the idea of also enforcing `run_maintenance` does not get stuck (this might cause a lot of issues as well). But maybe this can happen in a separate PR.

Issue #900, opened 2021-07-27
[Contract Upgrade] Update Order Model and DB Read Queries


bh2smith:
Order model struct should be updated to accomodate for new fields (e.g. `order_metadata` will hold `contract_address`). From this point we can also update the READ orders query to include additional column info.

Depends on completion of both #898 and #899.

bh2smith:
May want to add stuff like this in order model

```rust
impl DbFundLocation {
    pub fn from(order_kind: FundLocation) -> Self {
        match order_kind {
            FundLocation::Owner => Self::Owner,
            FundLocation::VaultInternal => Self::VaultInternal,
            FundLocation::VaultExternal => Self::VaultExternal,
        }
    }
    fn into(self) -> FundLocation {
        match self {
            Self::Owner => FundLocation::Owner,
            Self::VaultInternal => FundLocation::VaultInternal,
            Self::VaultExternal => FundLocation::VaultExternal,
        }
    }
}
```

PR #906, opened 2021-07-27
Extend Orders Table and Update existing Records

### LINKED ISSUE #898: [Contract Upgrade] Extend Orders Table
Extend order table to include three new columns 

1. `contract_address/version` - 20 bytes, 
2. `balanceFrom` - int/enum (0, 1, 2), 
3. `balanceTo` - int/enum (0, 1, 2)

    where 
    0 = EOA balance
    1 = Balancer Vault internal
    2 = Balancer Vault external

**Note** - Should be introduced together with #899.

### LINKED ISSUE #899: [Contract Upgrade] Populate New Order Columns
Write DB migration to populate new columns with known values `(current_settlement_address, 0, 0)`. 

**Note** - Should be merged into #898.

bh2smith:
This PR addresses and closes #898 and closes #899 as follows:

1. introduce a tiny table `settlement_version` which holds `byeta` `settlement_contract` address
2. Introduce an enums BalanceTo and BalanceFrom (erc20 internal, external). We simutaneously introduce a rust enum DbBalanceTo/From
3. Alter orders table by adding three new columns (version_id, balance_from, balance_to) 
4. These columns are all filled in via default with appropriate values for all existing records. The columns are set to be NOT NULL.  and after existing records are updated, the default condition is dropped.
Note that we have also updated the raw sql insert order statement to start filling in these values as well. 

### Test Plan
No logical changes here, 

To manually check that the upgrade works, follow these steps:

1. Checkout main and run the orderbook (on a fresh or existing DB)
2. Run the Front end locally and place an order with the old schema
3. Shut down the orderbook service and checkout this branch
4. Apply the DB schema here and check that your order has been updated as expected.

### DIFF
--- a/database/sql/V011__create_settlement_version.sql
+++ b/database/sql/V011__create_settlement_version.sql
@@ -1,9 +1,0 @@
-CREATE TABLE settlement_version
-(
-    version_id INT GENERATED ALWAYS AS IDENTITY,
-    contract_address CHAR(42) NOT NULL,
-    PRIMARY KEY (version_id)
-);
-
-INSERT INTO settlement_version(contract_address)
-VALUES ('0x3328f5f2cEcAF00a2443082B657CedEAf70bfAEf');
--- a/database/sql/V011__create_settlement_version.sql
+++ b/database/sql/V011__create_settlement_version.sql
@@ -1,0 +1,9 @@
+CREATE TABLE settlement_version
+(
+    version_id INT GENERATED ALWAYS AS IDENTITY,
+    contract_address CHAR(42) NOT NULL,
+    PRIMARY KEY (version_id)
+);
+
+INSERT INTO settlement_version(contract_address)
+VALUES ('0x3328f5f2cEcAF00a2443082B657CedEAf70bfAEf');
--- a/database/sql/V011__extend_orders_for_contract_upgrade.sql
+++ b/database/sql/V011__extend_orders_for_contract_upgrade.sql
@@ -1,5 +1,13 @@
 CREATE TYPE FundLocation AS ENUM ('owner', 'vault_internal', 'vault_external');
 
+-- While we could have simply added columns, set them to not null and made the update values defaults,
+-- This would mean that we will forever have to ensure we don't accidentally insert without specifying
+-- these values explicitly. This is especially awkward for the settlement_version, since the default
+-- would be the old contract version. For this reason, we have chosen to go with the approach of
+-- 1. Add columns,
+-- 2. update old records with appropriate values,
+-- 3. Set new columns to NOT NULL
+
 ALTER TABLE orders
     ADD COLUMN settlement_version CHAR(42),
     ADD COLUMN balance_from FundLocation,
--- a/database/sql/V011__extend_orders_for_contract_upgrade.sql
+++ b/database/sql/V011__extend_orders_for_contract_upgrade.sql
@@ -7,7 +7,7 @@ CREATE TYPE BalanceTo AS ENUM ('owner', 'vault_internal');
 -- would be the old contract version. For this reason, we have chosen to go with the approach of
 -- 1. Add columns, setting them not null with default values,
 ALTER TABLE orders
-    ADD COLUMN settlement_contract CHAR(42) NOT NULL default '0x3328f5f2cEcAF00a2443082B657CedEAf70bfAEf',
+    ADD COLUMN settlement_contract bytea NOT NULL default '\x3328f5f2cEcAF00a2443082B657CedEAf70bfAEf',
     ADD COLUMN balance_from BalanceFrom NOT NULL default 'owner',
     ADD COLUMN balance_to BalanceTo NOT NULL default 'owner';
 
--- a/database/sql/V011__extend_orders_for_contract_upgrade.sql
+++ b/database/sql/V011__extend_orders_for_contract_upgrade.sql
@@ -5,22 +5,14 @@ CREATE TYPE BalanceTo AS ENUM ('owner', 'vault_internal');
 -- This would mean that we will forever have to ensure we don't accidentally insert without specifying
 -- these values explicitly. This is especially awkward for the settlement_contract, since the default
 -- would be the old contract version. For this reason, we have chosen to go with the approach of
--- 1. Add columns,
--- 2. update old records with appropriate values,
--- 3. Set new columns to NOT NULL
-
+-- 1. Add columns, setting them not null with default values,
 ALTER TABLE orders
-    ADD COLUMN settlement_contract CHAR(42),
-    ADD COLUMN balance_from BalanceFrom,
-    ADD COLUMN balance_to BalanceTo;
+    ADD COLUMN settlement_contract CHAR(42) NOT NULL default '0x3328f5f2cEcAF00a2443082B657CedEAf70bfAEf',,
+    ADD COLUMN balance_from BalanceFrom NOT NULL default 'owner',
+    ADD COLUMN balance_to BalanceTo NOT NULL default 'owner';
 
-UPDATE orders
-SET settlement_contract = '0x3328f5f2cEcAF00a2443082B657CedEAf70bfAEf',
-    balance_from = 'owner',
-    balance_to = 'owner'
-WHERE settlement_contract IS NULL;
-
+-- 2. Drop defaults
 ALTER TABLE orders
-    ALTER COLUMN settlement_contract SET NOT NULL,
-    ALTER COLUMN balance_from SET NOT NULL,
-    ALTER COLUMN balance_to SET NOT NULL;
+    ALTER COLUMN settlement_contract DROP DEFAULT,
+    ALTER COLUMN balance_from DROP DEFAULT,
+    ALTER COLUMN balance_to DROP DEFAULT;
--- a/database/sql/V011__extend_orders_for_contract_upgrade.sql
+++ b/database/sql/V011__extend_orders_for_contract_upgrade.sql
@@ -1,5 +1,5 @@
-CREATE TYPE BalanceFrom AS ENUM ('owner', 'vault_internal', 'vault_external');
-CREATE TYPE BalanceTo AS ENUM ('owner', 'vault_internal');
+CREATE TYPE BalanceFrom AS ENUM ('erc20', 'internal', 'external');
+CREATE TYPE BalanceTo AS ENUM ('erc20', 'internal');
 
 -- While we could have simply added columns, set them to not null and made the update values defaults,
 -- This would mean that we will forever have to ensure we don't accidentally insert without specifying
@@ -8,8 +8,8 @@ CREATE TYPE BalanceTo AS ENUM ('owner', 'vault_internal');
 -- 1. Add columns, setting them not null with default values,
 ALTER TABLE orders
     ADD COLUMN settlement_contract bytea NOT NULL default '\x3328f5f2cEcAF00a2443082B657CedEAf70bfAEf',
-    ADD COLUMN balance_from BalanceFrom NOT NULL default 'owner',
-    ADD COLUMN balance_to BalanceTo NOT NULL default 'owner';
+    ADD COLUMN balance_from BalanceFrom NOT NULL default 'erc20',
+    ADD COLUMN balance_to BalanceTo NOT NULL default 'erc20';
 
 -- 2. Drop defaults
 ALTER TABLE orders
--- a/database/sql/V011__extend_orders_for_contract_upgrade.sql
+++ b/database/sql/V011__extend_orders_for_contract_upgrade.sql
@@ -3,24 +3,24 @@ CREATE TYPE BalanceTo AS ENUM ('owner', 'vault_internal');
 
 -- While we could have simply added columns, set them to not null and made the update values defaults,
 -- This would mean that we will forever have to ensure we don't accidentally insert without specifying
--- these values explicitly. This is especially awkward for the settlement_version, since the default
+-- these values explicitly. This is especially awkward for the settlement_contract, since the default
 -- would be the old contract version. For this reason, we have chosen to go with the approach of
 -- 1. Add columns,
 -- 2. update old records with appropriate values,
 -- 3. Set new columns to NOT NULL
 
 ALTER TABLE orders
-    ADD COLUMN settlement_version CHAR(42),
+    ADD COLUMN settlement_contract CHAR(42),
     ADD COLUMN balance_from BalanceFrom,
     ADD COLUMN balance_to BalanceTo;
 
 UPDATE orders
-SET settlement_version = '0x3328f5f2cEcAF00a2443082B657CedEAf70bfAEf',
+SET settlement_contract = '0x3328f5f2cEcAF00a2443082B657CedEAf70bfAEf',
     balance_from = 'owner',
     balance_to = 'owner'
-WHERE settlement_version IS NULL;
+WHERE settlement_contract IS NULL;
 
 ALTER TABLE orders
-    ALTER COLUMN settlement_version SET NOT NULL,
+    ALTER COLUMN settlement_contract SET NOT NULL,
     ALTER COLUMN balance_from SET NOT NULL,
     ALTER COLUMN balance_to SET NOT NULL;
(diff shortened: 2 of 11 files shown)


database/sql/V011__create_settlement_version.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE settlement_version
```
fleupold, line 1: What's the advantage of using a separate table for this? Shouldn't contract just be a column on the order table? Also I wouldn't populate the contract address values hardcoded (except for migrations) but instead have code read the current address from the used contract artefact in the rust code and use that address. This way we depend on this string only in one place.

database/sql/V012__extend_orders_for_contract_upgrade.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TYPE FundLocation AS ENUM ('owner', 'vault_internal', 'vault_external');
+
+ALTER TABLE orders
+    ADD COLUMN version_id INT,
```
fleupold, line 4: Maybe you can use `NOT NULL WITH DEFAULT` here instead of doing 3 separate SQL statements.

orderbook/src/database/orders.rs
```diff
@@ -120,6 +129,10 @@ impl OrderStoring for Postgres {
             .bind(order.order_creation.partially_fillable)
             .bind(order.order_creation.signature.to_bytes().as_ref())
             .bind(DbSigningScheme::from(order.order_creation.signing_scheme))
+            // TODO - remove this in https://github.com/gnosis/gp-v2-services/issues/901
+            .bind(1)
```
fleupold, line 133: Here I would use our rust contract artefacts to derive the correct contract address.

fleupold (commented):
Looks good to me but I would like to discuss if we really need a separate contract version table with one column which in itself could serve as a FK.

What I like about this change is that it can be merged regardless of the other changes being finished.

Also, the test plan sounds great 😄  let's turn it into past tense instead of conditional 😉

database/sql/V011__create_settlement_version.sql
taminomara: I have similar thoughts. Having an integer version (or an enum version) and mapping it to contract address in Rust code will give us more flexibility IMO. This is where https://github.com/gnosis/ethcontract-rs/issues/545 would've been useful.

taminomara (commented):
Looks good!

database/sql/V011__create_settlement_version.sql
nlordell: One advantage of having the contract address directly in the DB is that it is implicitly included in the signature. Specifically, if we were to redeploy our contracts (with a different salt, so they get a different deterministic address), then the current orders and signatures would no longer be valid, even if the code is identical, because the contract address is used for deriving the domain separator which is included in the message that is used for signing.

I think manually maintaining versions in Rust would be more error prone than having the orderbook API programmatically include the settlement contract address that it is using when inserting new rows.

bh2smith: I am getting the impression that more of us are in favour of putting the `settlement_version`(contract address) directly in the orders table. This is totally fine by me as long as we are all comfortable with the redundant bytes (occupying ~1 GB per 50k orders).

Thumbs up 👍🏽 for eliminating the settlement version table and putting the address directly in the orders table.
Thumbs down 👎🏽 for keeping it as is.

fleupold: Just as a general comment, I think subjective polls (with thumb up/down votes) are good for naming questions, but technical decisions like this should be guided by technical arguments and not by a majority vote (and ideally we would at the end of the discussion agree what is the better solutions, because the arguments make sense).

database/sql/V012__extend_orders_for_contract_upgrade.sql
bh2smith: Did you read the description of this PR or [this comment](https://github.com/gnosis/gp-v2-services/issues/898#issuecomment-887699878) in the issue? 

I will highlight the reason I have chosen against this, but like the description says, the first commit here is an implementation with this suggestion if we would prefer to go this way.

> Option 1: Add non-null columns with default values. This would mean that we will forever have to ensure we don't accidentally insert without specifying these values explicitly (this would be especially important for the settlement version, since the default would be the old contract version. I am not a fan of the default value for the contract version being set to the old contracts

Later on we discuss Option 3 which I quite like (as a mix of the two)

> Do Option 2 for settlement_version and Option 1 for the other two columns (essentially making owner) the default value for balance to and from fields.

orderbook/src/database/orders.rs
bh2smith: This is cool... but we still need to decide if this field is going to be an address or not. We can push this forward by voting on https://github.com/gnosis/gp-v2-services/pull/906#discussion_r678190192

database/sql/V011__create_settlement_version.sql
bh2smith: True. Well, now that I see the occupancy calculation was off by a factor of 1000, I think the better option is to eliminate this table and put the address directly in the orders table.

orderbook/src/database/orders.rs
bh2smith: So, in order to make this method (`insert_order`) non-blocking on web3 calls I would propose we either store the contract address on the Postgres struct (will require changes to its new constructor in many places). Alternatively we could make some dynamic global constant that can be imported here and used. Not super keen on updating every occurrence of `Postgres::new` since there are many.

Screenshot 2021-07-28 at 07 29 31

orderbook/src/database/orders.rs
```diff
@@ -120,6 +129,11 @@ impl OrderStoring for Postgres {
             .bind(order.order_creation.partially_fillable)
             .bind(order.order_creation.signature.to_bytes().as_ref())
             .bind(DbSigningScheme::from(order.order_creation.signing_scheme))
+            // TODO - find a non-blocking way of accessing this.
+            .bind("0x3328f5f2cEcAF00a2443082B657CedEAf70bfAEf")
```
bh2smith, line 133: Only thing that remains here is to find a non-blocking way to access this. Does anyone have any clever suggestions that doesn't involve putting the settlement_address as a new field on the Postgres instance and adapting the constructor in the 20 places it is used throughout the code?

bh2smith: I have offloaded this adjustment to a separate PR where we can discuss the best approach.

https://github.com/gnosis/gp-v2-services/pull/907

orderbook/src/database/orders.rs
bh2smith: cf #907

database/sql/V012__extend_orders_for_contract_upgrade.sql
fleupold: I read it but maybe misunderstood what you meant by it. I missed that with my suggestion all future rows inserted would also contain the default value - agree that we don't want that!

Makes perfect sense then (maybe we can add a small comment in the code)?

orderbook/src/database/orders.rs
fleupold: How about putting the contract address on the order metadata struct? This way it can be populated when we also verify the signature (at which point we need access to the contract address for the domain separator).

database/sql/V012__extend_orders_for_contract_upgrade.sql
fleupold: After googling, I found an alternative with 2 queries would be to drop the default later on:

> ALTER TABLE mytable ALTER COLUMN mycolumn DROP DEFAULT;

But I'm also fine with the existing solution.

bh2smith: Will add a comment. I think the drop default solution is cool, but we would have to remember to do that at a later point.

orderbook/src/database/orders.rs
bh2smith: Yes. I have actually already started a PR addressing issue #900 which will make this trivial and no longer require #907.

database/sql/V012__extend_orders_for_contract_upgrade.sql
fleupold: Couldn't it simply happen inside this migration script instead of statement 2 & 3?

taminomara: Drop default solution has an advantage that DB won't have to check that all values are not NULL.

bh2smith: Only thing I would like to check is that the existing records are actually updated with the default values.

bh2smith: Works like a charm!

orderbook/src/database/orders.rs
```diff
@@ -56,6 +56,23 @@ impl DbOrderKind {
     }
 }
 
+#[derive(sqlx::Type)]
+#[sqlx(type_name = "BalanceFrom")]
+#[sqlx(rename_all = "snake_case")]
+pub enum BalanceFrom {
```
fleupold, line 62: Nit: maybe docs on what the subtypes mean (especially owner vs VaultExternal might not be obvious from the name)

fleupold (approved):
Looks great! (I assume the upgrade was tested locally with old records present in the db)

bh2smith:
> I assume the upgrade was tested locally with old records present in the db

Yes, I have tested locally although not with our existing staging or prod DBs, but rather one that I made by placing an order on main and then checking out this branch, applying the new schema and cross checking expectations.

orderbook/src/database/orders.rs
bh2smith: All fixed up with some very nice documentation. Thanks very much for pointing this out!

bh2smith:
Image below shows last three columns (the new ones) of an order placed before the migration and one placed after. All is good!

Screenshot 2021-07-28 at 12 59 11

taminomara (approved):

database/sql/V011__extend_orders_for_contract_upgrade.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TYPE BalanceFrom AS ENUM ('owner', 'vault_internal', 'vault_external');
+CREATE TYPE BalanceTo AS ENUM ('owner', 'vault_internal');
```
nlordell, line 2: nit: These should probably be the same as the SC enum:
```suggestion
CREATE TYPE BalanceFrom AS ENUM ('erc20', 'external', 'internal');
CREATE TYPE BalanceTo AS ENUM ('erc20', 'internal');
```

database/sql/V011__extend_orders_for_contract_upgrade.sql
```diff
@@ -0,0 +1,18 @@
...
+
+-- While we could have simply added columns, set them to not null and made the update values defaults,
+-- This would mean that we will forever have to ensure we don't accidentally insert without specifying
+-- these values explicitly. This is especially awkward for the settlement_contract, since the default
+-- would be the old contract version. For this reason, we have chosen to go with the approach of
+-- 1. Add columns, setting them not null with default values,
```
nlordell, line 8: very cleaver

orderbook/src/database/orders.rs
```diff
@@ -56,6 +57,30 @@ impl DbOrderKind {
     }
 }
 
+/// Location for which the sellAmount should be drawn upon order fulfilment
+#[derive(sqlx::Type)]
+#[sqlx(type_name = "BalanceFrom")]
+#[sqlx(rename_all = "snake_case")]
+pub enum BalanceFrom {
+    /// Order owner's ethereum address
+    Owner,
```
nlordell, line 66: A more accurate description is that the balance is taken using direct ERC20 allowances to the vault relayer contract. I find the names/description confusing as the balance always comes from the order's owner, it just comes from difference sources:
- Direct ERC20 allowances to the Vault relayer contract
- ERC20 allowances to the Vault with GPv2 relayer approval
- Internal balances to the Vault with GPv2 relayer approval

The reason I feel like "Order owner's ethereum address" is ambiguous is that it makes me think that the address holds the balance, which isn't entirely accurate, as its the ERC20 token contract that keeps track of the balances in storage for both `BalanceFrom::Owner` and `BalanceFrom::VaultExternal` (as opposed to ETH balance which is tied to the account and thus "held by the owner's ethereum address") or the Vault contract for `BalanceFrom::Internal`.

nlordell (commented):
Looks good, just some naming nits on the different balances.

database/sql/V011__extend_orders_for_contract_upgrade.sql
bh2smith: I will do this, but I don't particularly like that internal and external make no reference to the fact that these are vault related.

nlordell: I started a discussion about this here: https://github.com/gnosis/gp-v2-contracts/issues/543

We can always re-release a new version with just name changes to make it clearer and then update it here to match the new names. However, I think it makes sense for the enum values to match the values that get signed with the order.

database/sql/V011__create_settlement_version.sql
nlordell: I'd still be curious to see if there is a performance impact for including `bytea` in tables vs using a separate table with FK. I don't think it matters for this PR though and would be in favour of the simpler option (including the `settlment_contract` address directly in the order metadata table) and then circling back to this once we can actually measure the performance impact of a change. an additional 20 bytes per order seems a relatively small price to pay (50 * 20 = 1000 == 1KB, I'm not sure where the 1MB comes from).

database/sql/V011__extend_orders_for_contract_upgrade.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TYPE BalanceFrom AS ENUM ('erc20', 'internal', 'external');
+CREATE TYPE BalanceTo AS ENUM ('erc20', 'internal');
+
+-- While we could have simply added columns, set them to not null and made the update values defaults,
+-- This would mean that we will forever have to ensure we don't accidentally insert without specifying
+-- these values explicitly. This is especially awkward for the settlement_contract, since the default
+-- would be the old contract version. For this reason, we have chosen to go with the approach of
+-- 1. Add columns, setting them not null with default values,
+ALTER TABLE orders
+    ADD COLUMN settlement_contract bytea NOT NULL default '\x3328f5f2cEcAF00a2443082B657CedEAf70bfAEf',
```
nlordell, line 10: checksummed address - love it!

orderbook/src/database/orders.rs
```diff
@@ -56,6 +57,30 @@ impl DbOrderKind {
...
+#[derive(sqlx::Type)]
+#[sqlx(type_name = "DbBalanceFrom")]
+#[sqlx(rename_all = "snake_case")]
+pub enum DbBalanceFrom {
+    /// Direct ERC20 allowances to the Vault relayer contract
+    Erc20,
+    /// ERC20 allowances to the Vault with GPv2 relayer approval
+    Internal,
+    /// Internal balances to the Vault with GPv2 relayer approval
+    External,
```
nlordell, line 70: If you want to use `Vault*` in the Rust type, you *may* be able to use something like:
```suggestion
    /// ERC20 allowances to the Vault with GPv2 relayer approval
    #[sqlx(rename = "internal")]
    VaultInternal,
    /// Internal balances to the Vault with GPv2 relayer approval
    #[sqlx(rename = "external")]
    VaultExternal,
```

I agree that the names aren't great - maybe we should revive the discussion https://github.com/gnosis/gp-v2-contracts/issues/543

I'm also 100% fine as is, but just trying to make a suggestion that would improve the naming in the Rust codebase, while still keeping the database values consistent with what was signed by the user.

nlordell (approved):
Looks good!

orderbook/src/database/orders.rs
bh2smith: I think we can leave it as is for now and I have started to become used to it. The naming now suggests that the Balancer integration is so tight it actually gets the default naming convention by excluding the term Vault. Anyway, this can be changed later.

PR #907, opened 2021-07-28
update Postgres::new constructor

bh2smith:
This is my attempt at "deriving" the settlement contract address for order insertion in the least blocking way.

### Test Plan
Existing CI.


bh2smith:
Closed in favour of more natural solution #908.

PR #908, opened 2021-07-28
Contract upgrade - update order model(s)

### LINKED ISSUE #900: [Contract Upgrade] Update Order Model and DB Read Queries

bh2smith:
Closes #900

We add 

- `settlement_contract: H160` to `OrderMetaData` and use it for order insertion.
- `sell_token_balance` and `buy_token_balance` added to the `OrderCreation` struct

We also update the API spec to include buy/sellTokenBalance default values of "owner" (this is mostly so not to introduce a breaking change with the front end and could be removed when they are ready). 

### Test Plan
Existing tests were adapted to include various uses of each of the new field values. Perhaps as part of this integration we could plan to include an e2e test for orders of these types. I don't believe this to be ready for such a test yet, so I have logged an issue (https://github.com/gnosis/gp-v2-services/issues/910) for it.

### DIFF
--- a/model/src/order.rs
+++ b/model/src/order.rs
@@ -38,7 +38,12 @@ pub struct Order {
 
 impl Default for Order {
     fn default() -> Self {
-        Self::from_order_creation(OrderCreation::default(), &DomainSeparator::default()).unwrap()
+        Self::from_order_creation(
+            OrderCreation::default(),
+            &DomainSeparator::default(),
+            H160::default(),
+        )
+        .unwrap()
     }
 }
 
@@ -55,6 +60,7 @@ impl Order {
     pub fn from_order_creation(
         order_creation: OrderCreation,
         domain: &DomainSeparator,
+        settlement_contract: H160,
     ) -> Option<Self> {
         let owner = order_creation.signature.validate(
             order_creation.signing_scheme,
@@ -66,6 +72,7 @@ impl Order {
                 creation_date: chrono::offset::Utc::now(),
                 owner,
                 uid: order_creation.uid(domain, &owner),
+                settlement_contract,
                 ..Default::default()
             },
             order_creation,
@@ -190,6 +197,10 @@ pub struct OrderCreation {
     pub partially_fillable: bool,
     pub signature: Signature,
     pub signing_scheme: SigningScheme,
+    #[serde(default)]
+    pub sell_token_balance: BalanceFrom,
+    #[serde(default)]
+    pub buy_token_balance: BalanceTo,
 }
 
 #[derive(Clone, Copy, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
@@ -215,6 +226,8 @@ impl Default for OrderCreation {
             partially_fillable: Default::default(),
             signing_scheme: SigningScheme::Eip712,
             signature: Default::default(),
+            sell_token_balance: Default::default(),
+            buy_token_balance: Default::default(),
         };
         result.signature = Signature::sign(
             result.signing_scheme,
@@ -345,6 +358,8 @@ pub struct OrderMetaData {
     pub executed_fee_amount: BigUint,
     pub invalidated: bool,
     pub status: OrderStatus,
+    #[serde(with = "h160_hexadecimal")]
+    pub settlement_contract: H160,
 }
 
 impl Default for OrderMetaData {
@@ -360,6 +375,7 @@ impl Default for OrderMetaData {
             executed_fee_amount: Default::default(),
             invalidated: Default::default(),
             status: OrderStatus::Open,
+            settlement_contract: H160::default(),
         }
     }
 }
@@ -460,6 +476,42 @@ impl Default for OrderKind {
     }
 }
 
+/// Location for which the sellAmount should be drawn upon order fulfilment
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Deserialize, Serialize, Hash, enum_utils::FromStr)]
+#[enumeration(case_insensitive)]
+#[serde(rename_all = "snake_case")]
+pub enum BalanceFrom {
+    /// Direct ERC20 allowances to the Vault relayer contract
+    Erc20,
+    /// ERC20 allowances to the Vault with GPv2 relayer approval
+    Internal,
+    /// Internal balances to the Vault with GPv2 relayer approval
+    External,
+}
+
+impl Default for BalanceFrom {
+    fn default() -> Self {
+        Self::Erc20
+    }
+}
+
+/// Location for which the buyAmount should be transferred to order's receiver to upon fulfilment
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Deserialize, Serialize, Hash, enum_utils::FromStr)]
+#[enumeration(case_insensitive)]
+#[serde(rename_all = "snake_case")]
+pub enum BalanceTo {
+    /// TransferAs: ERC20 token transfer
+    Erc20,
+    /// TransferAs: Vault internal balance transfer.
+    Internal,
+}
+
+impl Default for BalanceTo {
+    fn default() -> Self {
+        Self::Erc20
+    }
+}
+
 pub fn debug_app_data(
     app_data: &[u8; 32],
     formatter: &mut std::fmt::Formatter,
@@ -511,6 +563,9 @@ mod tests {
             "signature": "0x0200000000000000000000000000000000000000000000000000000000000003040000000000000000000000000000000000000000000000000000000000000501",
             "signingScheme": "eip712",
             "status": "open",
+            "settlementContract": "0x0000000000000000000000000000000000000002",
+            "sellTokenBalance": "external",
+            "buyTokenBalance": "internal",
         });
         let expected = Order {
             order_meta_data: OrderMetaData {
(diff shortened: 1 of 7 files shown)


model/src/order.rs
```diff
@@ -345,6 +352,8 @@ pub struct OrderMetaData {
     pub executed_fee_amount: BigUint,
     pub invalidated: bool,
     pub status: OrderStatus,
+    #[serde(with = "h160_hexadecimal")]
+    pub contract_version: H160,
```
fleupold, line 356: naming nit: I'd call this field `settlement_contract` everywhere to be clear about what contract we are talking.

fleupold (approved):
Looks good to me. Might also need updating of the API docs, as we are exposing a new field on our order json.

model/src/order.rs
bh2smith: Do you mean `settlement_version` or should I go back and also change #906 to also use `settlement_contract` as well? I think I will do that. Use settlement_contract everywhere.

bh2smith:
@alfetopito - I wanted to point out the changes included to to make sure my assumptions are true. That is, I have included two new fields in the API Order creation spec with a default value of "owner". It is my intention that this should not be a breaking change from your end (i.e. not require any action from your side) and that these defaults could (maybe) be removed in the future when the front end is ready for this.

orderbook/openapi.yml
```diff
@@ -486,6 +494,12 @@ components:
           $ref: "#/components/schemas/Signature"
         signingScheme:
           $ref: "#/components/schemas/SigningScheme"
+        balanceFrom:
+          $ref: "#/components/schemas/BalanceFrom"
+          default: "owner"
+        balanceTo:
+          $ref: "#/components/schemas/BalanceTo"
+          default: "owner"
```
alfetopito, line 502: I get what this means and it's talking about the Balancer integration. 
But how does this play out with the `receiver` field?

If I set `balanceTo=owner` and `receiver=<not source(owner) address>` what happens?
What about when `balanceTo=vault_internal`?

I expect `receiver` would take precedence. I also assume `balanceTo` only applies to Balancer vaults.

Can you clarify?

model/src/order.rs
```diff
@@ -190,6 +197,8 @@ pub struct OrderCreation {
     pub partially_fillable: bool,
     pub signature: Signature,
     pub signing_scheme: SigningScheme,
+    pub balance_from: Option<BalanceFrom>,
+    pub balance_to: Option<BalanceTo>,
```
nlordell, line 201: These are not optional

model/src/order.rs
```diff
@@ -460,6 +474,42 @@ impl Default for OrderKind {
     }
 }
 
+/// Location for which the sellAmount should be drawn upon order fulfilment
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Deserialize, Serialize, Hash, enum_utils::FromStr)]
+#[enumeration(case_insensitive)]
+#[serde(rename_all = "snake_case")]
+pub enum BalanceFrom {
+    /// Order owner's ethereum address
+    Owner,
```
nlordell, line 483: I think these should match the SC enum, as that is what you will use to sign:
- `erc20`
- `external`
- `internal`

orderbook/openapi.yml
nlordell: It would send to the receiver and not the owner. This enum represents how the trade proceeds get paid out, either in ERC20 tokens, or in Balancer Vault internal balances.

bh2smith: And as an additional point, there is nothing preventing a transfer to a receiver's internal vault balance even if they have never used the balancer protocol.

model/src/order.rs
```diff
@@ -460,6 +476,42 @@ impl Default for OrderKind {
...
+        Self::Erc20
+    }
+}
+
+/// Location for which the buyAmount should be transferred to order's receiver to upon fulfilment
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Deserialize, Serialize, Hash, enum_utils::FromStr)]
+#[enumeration(case_insensitive)]
+#[serde(rename_all = "snake_case")]
+pub enum BalanceTo {
+    /// TransferAs: ERC20 token transfer
```
nlordell, line 503: I would remove the `TransferAs`.

orderbook/openapi.yml
```diff
@@ -486,6 +494,12 @@ components:
           $ref: "#/components/schemas/Signature"
         signingScheme:
           $ref: "#/components/schemas/SigningScheme"
+        sellTokenBalance:
+          $ref: "#/components/schemas/SellTokenBalance"
+          default: "owner"
```
nlordell: this value seems to be outdated.

orderbook/openapi.yml
```diff
@@ -486,6 +494,12 @@ components:
...
+        buyTokenBalance:
+          $ref: "#/components/schemas/BuyTokenBalance"
+          default: "owner"
```
nlordell: here too.

orderbook/src/database/orders.rs
```diff
@@ -194,7 +224,9 @@ impl OrderStoring for Postgres {
                 COALESCE(SUM(t.buy_amount), 0) AS sum_buy, \
                 COALESCE(SUM(t.sell_amount), 0) AS sum_sell, \
                 COALESCE(SUM(t.fee_amount), 0) AS sum_fee, \
-                (COUNT(invalidations.*) > 0 OR o.cancellation_timestamp IS NOT NULL) AS invalidated \
+                (COUNT(invalidations.*) > 0 OR o.cancellation_timestamp IS NOT NULL) AS invalidated, \
+                o.settlement_contract, o.balance_from, o.balance_to
```
nlordell, line 228: ubernit: maybe move this to line 223 so that is with the other order columns?

nlordell (commented):
Some minor comments.

model/src/order.rs
bh2smith: Whoops... I only did it in one place.

model/src/order.rs
```diff
@@ -460,6 +476,42 @@ impl Default for OrderKind {
     }
 }
 
+/// Location for which the sellAmount should be drawn upon order fulfilment
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Deserialize, Serialize, Hash, enum_utils::FromStr)]
+#[enumeration(case_insensitive)]
+#[serde(rename_all = "snake_case")]
+pub enum BalanceFrom {
```
nlordell, line 483: Just curious, why not call this `SellTokenBalance` like in the OpenAPI spec? It might remove some confusion that this doesn't specify which account the balance is taken from, but rather how the balance is taken?

I'm also OK with how it is.

nlordell (approved):

model/src/order.rs
bh2smith: Well, sure we could do this... but then your next question is going to be why not rename the DB columns names and enums also? Personally I am not a fan of the term "Buy/SellTokenBalance" because it seems like it should be a number, so I would prefer to switch it to something that I feel makes sense as soon as possible internally.

nlordell: I actually would change the column names to:
```
    ADD COLUMN sell_token_balance BalanceFrom NOT NULL default 'erc20',
    ADD COLUMN buy_token_balance BalanceTo NOT NULL default 'erc20';
```

Just because these two are the only columns that don't match the names of their corresponding fields. While its not a big deal, this can be a bit counter-intuitive to someone who joins the project and is trying to figure out "where does the `sell_token_balance` field come from? I can't find a column for it in the database?".

(I glanced over this and missed it in the other review).

bh2smith: I belive this is kinda too late now that #898 has already been merged. This would require another DB migration.

orderbook/src/database/orders.rs
```diff
@@ -325,6 +360,8 @@ impl OrdersQueryRow {
                     .map_err(|_| anyhow!("signature has wrong length"))?,
             ),
             signing_scheme: self.signing_scheme.into(),
+            sell_token_balance: self.balance_from.into(),
+            buy_token_balance: self.balance_to.into(),
```
nlordell, line 364: Related to :point_up:: it feels weird that this is the only field that doesn't have consistent naming between SC order properties, `model::Order` fields, `database::Order` fields and database column names. I just feel like it may cause confusion in the future.

nlordell:
Naming issues will be addressed separately as they require some manual DB intervention in staging.

orderbook/openapi.yml
alfetopito: Now I'm more confused as to how receiver and balanceTo interact.
Are they completely independent of each other?

bh2smith: BalanceTo will be renamed to BuyTokenBalance and it is just an enum indicating the location where the proceeds are to be transferred. receiver will always be the receiver of the proceeds, but this enum indicates where the receivers balance is to be updated. 

That is either the receivers Erc20 token balance is updated, or their internal vault balance.

alfetopito: Synced over slack, all good now

PR #911, opened 2021-07-29
 App Data Step 1: [additional optional PR]

josojo:
This is the implementation of the suggestion by Leandro to just store a JSON object instead of the bytes of the ipfs.

I actually like it, but it breaks the compatibility with IPFS. From the json object, I can not retrieve in all cases the oringal bytes of the ipfs-file. Both solutions are fine in my opinion.

Please give it a thumbs up, if you prefer this implementation and I will merge it into the original PR.


nlordell:
> From the json object, I can not retrieve in all cases the oringal bytes of the ipfs-file.

Interesting... that surprises me a little as my mental model of IPFS you could recover it for a known CID version/multicodec.

josojo:
yes, if we rely on ipfs for data availability then, we could do it sure. But I am unsure whether we want to do that from the beginning. I thought we just want to switch to ipfs on a later stage

wtdy? @alfetopito @anxolin

alfetopito (approved):
As I commented in the other PR, I'm in favor of this change

alfetopito:
> I can not retrieve in all cases the oringal bytes of the ipfs-file.

Can you expand what do you mean by that?

josojo:
> > I can not retrieve in all cases the oringal bytes of the ipfs-file.
> 
> Can you expand what do you mean by that?

Well if I do the conversion from <String received from api> -> <json format>, then newlines spaces etc, are getting lost. 
E..g.

`{ address: "0x" } `
and 
`{ 
address:   
"0x" }`
 would be the same, also their IPFS hash is different

alfetopito:
> > > I can not retrieve in all cases the oringal bytes of the ipfs-file.
> > 
> > 
> > Can you expand what do you mean by that?
> 
> Well if I do the conversion from -> , then newlines spaces etc, are getting lost.
> E..g.
> 
> `{ address: "0x" } `
> and
> 
> ```
> { 
> address:   
> "0x" }
> ```
> 
> would be the same, also their IPFS hash is different

Ah I see, makes sense.
I would expect the JSON to be "minified" anyway before uploading to IPFS/backend, so at least the empty spaces should not be a problem.

bh2smith:
As cool as it sounds, I'm not sure I'd be in favour of storing json here unless we know for sure that the json is bounded above in size. Maybe I'm old school but I think the binary encoding is much more compact so it really depends on what we intend to do with it. Are we going to read it from the DB and immediately deserialize to json every time? Or will we be passing it around via API services?

Personally I have no really preference here and wouldn't complain either way. I do have this hunch that storing bytes is somehow more practical long term, but I can't really back it by much more than a few thoughts on compactness. 

I did find this article when searching for other thoughts, but most arguments state that this is a matter of opinion with not much concrete evidence to definitively make one or the other stand out. 

https://betterprogramming.pub/use-binary-encoding-instead-of-json-dec745ec09b6

fleupold:
> Maybe I'm old school but I think the binary encoding is much more compact so it really depends on what we intend to do with it. 

Note, that just using `bytea` to encode a string blob (UTF-8) as it was before this PR doesn't give you the compactness that is mentioned in the article. We would have to binary encode the raw rust struct (so that it e.g. only uses 4 bytes for an int rather than allocating a string) using protobuf or the like. Since we are already storing the important fields in extra columns I'm not sure `bytea` vs `json` makes a big difference in terms of efficiency here.

PR #913, opened 2021-07-29
[Easy] adding affiliate link to matcha solver

josojo:
[Optional] 
0x team asked us to add the affiliate, It allows additional tracking internally, and we will also get visibility on their dashboards.
https://gnosisinc.slack.com/archives/CLMBRFKEF/p1626808750002500

### Test Plan
Unit tests only


nlordell (approved):

solver/src/solver/matcha_solver/api.rs
```diff
@@ -13,6 +13,10 @@ use reqwest::{Client, IntoUrl, Url};
 use serde::Deserialize;
 use web3::types::Bytes;
 
+// Matcha requires an address as an affiliate.
+// Hence we hand over the settlement contract address
+const AFFILIATE_ADDRESS: &str = "0x9008D19f58AAbD9eD0D60971565AA8510560ab41";
```
fleupold, line 18: This is the new (not yet used) contract's address. But if we hardcode this, then it's probably good to already use that one (we could also derive it from the artefact).

fleupold (approved):

solver/src/solver/matcha_solver/api.rs
josojo: Yes, I think hardcoding is the best solution, such that they can link us easily also in their dashboards.

bh2smith (approved):

vkgnosis (approved):

PR #914, opened 2021-07-29
[DB Migration] Fix undesired column names

bh2smith:
The recent merge of #906 and #908 introduced undesired column names to the DB while also making a gap in the flyaway DB migration script file naming convention (using V011 instead of V010 - this happened to make room for #875, which will not likely be merged before this).

Game plan for tomorrow is to revert both of the PRs mentioned above, restore staging DB from backup and then merge this as the corrected culmination of both.

Note that - the visable diff with main at this moment just shows the changes after the two branches are merged, but once they are reverted this should appear as all the code required to make the changes introduced in the above to PRs.

## Test Plan

```
- checkout main
- run postgress, apply schema, run orderbook and front end localy,
- place an order

- checkout this branch
- apply schema
- check your existing order and place another order.
```


bh2smith:
@nlordell - after some discussion with @alfetopito I think a more appropriate name for these types, fields and columns might be something like:

```
SellTokenSource & BuyTokenDest
```

WDYT?

nlordell:
I think that the field and column names should be `sell_token_balance` and `buy_token_balance` because that is what they are in the SC and what is signed in the order. It would be quite weird if the FE would sign some order JSON data, and then have to rename just these two properites to send to the API. Also, I feel like it causes confusion.

If we want to rename the order field names/enum values (`erc20`, `internal`, `external`) I think we should also do it in the SC and re-deploy. I think we should have that conversation here: https://github.com/gnosis/gp-v2-contracts/issues/543.

As for the enum type names, I like it!

vkgnosis (approved):

database/sql/V010__skipped.sql
```diff
@@ -1,2 +1,2 @@
--- V010 was accidentally skip, this empty placeholder file is just placed here
+-- V010 was accidentally skipped, this empty placeholder file is just placed here
```
nlordell, line 1: flyway will not allow this.

nlordell (changes requested):
Did you try the migration locally? When I tried editing previous migration scripts (even if it was to just to change a comment) the migration failed.

bh2smith:
> Did you try the migration locally? When I tried editing previous migration scripts (even if it was to just to change a comment) the migration failed.

I am trying to do it locally now, but something isn't working and the errors I am getting in the front end are not being propagated properly in the backend. I have a fresh DB and the interface keeps giving me an "Order not found error"

bh2smith:
@nlordell - Unfortunately the problem is that I tried to edit the V010 file (fixing a typo) and the checksum doesn't match. Loks like we will live with the typo forever.

```
ERROR: Validate failed: Migrations have failed validation
Migration checksum mismatch for migration version 010
-> Applied to database : 792685460
-> Resolved locally    : -1394922178. Either revert the changes to the migration, or run repair to update the schema history.
```

bh2smith:
@nlordell - testplan has been included above and I have run it locally as well. The problem you had was with the typo I tried to fix in V010.

nlordell:
Should the OpenAPI YAML need to be updated with the new names as well?

taminomara:
I wonder if we can just roll back staging and make a clean upgrade file without amends?

bh2smith:
> I wonder if we can just roll back staging and make a clean upgrade file without amends?

We tried this but undo is not available with the community version of flyaway.

bh2smith:
changes ready for review.

nlordell (approved):

PR #921, opened 2021-08-02
Update order flag encoding

### LINKED ISSUE #902: [Contract Upgrade] Settlement Encoding
We will need to adjust how order data is passed into the contract (i.e. don’t forget to include `order.balanceFrom`, `order.balanceTo` in the appropriate place when encoding order data into blob of bytes).

nlordell:
Fixes #902 

This PR updates the order flag encoding required for the updated SCs.

### Test Plan

<details><summary>Test cases were generated using the following code</summary>

```ts
import { OrderBalance, OrderKind, SigningScheme, TradeFlags, encodeTradeFlags, } from "./src/ts";

const permutations: TradeFlags[] = [];
for (const signingScheme of [SigningScheme.EIP712, SigningScheme.ETHSIGN, /* SigningScheme.EIP1271, SigningScheme.PRESIGN */])
for (const buyTokenBalance of [OrderBalance.ERC20, OrderBalance.INTERNAL])
for (const sellTokenBalance of [OrderBalance.ERC20, OrderBalance.EXTERNAL, OrderBalance.INTERNAL])
for (const partiallyFillable of [false, true])
for (const kind of [OrderKind.SELL, OrderKind.BUY])
  permutations.push({
    kind,
    partiallyFillable,
    sellTokenBalance,
    buyTokenBalance,
    signingScheme,
  });

function n(name: string): string {
  return name.replace(/(^.)/, (s) => s.toUpperCase());
}
function s(scheme: SigningScheme): string {
  switch (scheme) {
    case SigningScheme.EIP712: return "Eip712";
    case SigningScheme.ETHSIGN: return "EthSign";
    default: throw new Error("unsupported");
  }
}

console.log(`
    #[test]
    fn order_flag_permutations() {
        for (order, flags) in &[
            ${permutations.map((flags) => `(
                OrderCreation {
                    kind: OrderKind::${n(flags.kind)},
                    partially_fillable: ${flags.partiallyFillable},
                    sell_token_balance: BalanceFrom::${n(flags.sellTokenBalance!)},
                    buy_token_balance: BalanceTo::${n(flags.buyTokenBalance!)},
                    signing_scheme: SigningScheme::${s(flags.signingScheme)},
                    ..Default::default()
                },
                0b${encodeTradeFlags(flags).toString(2).padStart(7, "0")},
            ),`).join("\n            ")}
        ] {
            assert_eq!(order_flags(order), U256::from(*flags));
        }
    }
`);
```

</details>

A few permutations were hand-picked such that each value appears at least once. Can be executed manually with `cargo test -p solver -- order_flag_permutations`.

Note that CI is expected to fail because we are in a limbo state where part of the services are for the upgraded SCs, but they have not been enabled yet.

### DIFF
--- a/solver/src/encoding.rs
+++ b/solver/src/encoding.rs
@@ -43,20 +43,25 @@ pub fn encode_trade(
 
 fn order_flags(order: &OrderCreation) -> U256 {
     let mut result = 0u8;
+    // The kind is encoded as 1 bit in position 0.
     result |= match order.kind {
         OrderKind::Sell => 0b0,
         OrderKind::Buy => 0b1,
     };
+    // The order fill kind is encoded as 1 bit in position 1.
     result |= (order.partially_fillable as u8) << 1;
+    // The order sell token balance is encoded as 2 bits in position 2.
     result |= match order.sell_token_balance {
         BalanceFrom::Erc20 => 0b00,
         BalanceFrom::External => 0b10,
         BalanceFrom::Internal => 0b11,
     } << 2;
+    // The order buy token balance is encoded as 1 bit in position 4.
     result |= match order.buy_token_balance {
         BalanceTo::Erc20 => 0b0,
         BalanceTo::Internal => 0b1,
     } << 4;
+    // The signing scheme is encoded as a 2 bits in position 5.
     result |= match order.signing_scheme {
         SigningScheme::Eip712 => 0b00,
         SigningScheme::EthSign => 0b01,
@@ -94,6 +99,11 @@ mod tests {
                     signing_scheme: SigningScheme::Eip712,
                     ..Default::default()
                 },
+                // ......0 - sell order
+                // .....0. - fill-or-kill order
+                // ...00.. - ERC20 sell token balance
+                // ..0.... - ERC20 buy token balance
+                // 00..... - EIP-712 signing scheme
                 0b0000000,
             ),
             (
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):

solver/src/encoding.rs
```diff
@@ -44,16 +44,23 @@ pub fn encode_trade(
...
+    result |= (order.partially_fillable as u8) << 1;
+    result |= match order.sell_token_balance {
+        BalanceFrom::Erc20 => 0b00,
+        BalanceFrom::External => 0b10,
+        BalanceFrom::Internal => 0b11,
+    } << 2;
+    result |= match order.buy_token_balance {
+        BalanceTo::Erc20 => 0b0,
+        BalanceTo::Internal => 0b1,
+    } << 4;
```
bh2smith, line 59: I am a bit confused by the purpose of these shifts. Could we comment here?

solver/src/encoding.rs
```diff
@@ -70,3 +77,71 @@ pub struct EncodedSettlement {
...
+            (
+                OrderCreation {
+                    kind: OrderKind::Sell,
+                    partially_fillable: true,
+                    sell_token_balance: BalanceFrom::Erc20,
+                    buy_token_balance: BalanceTo::Internal,
+                    signing_scheme: SigningScheme::Eip712,
+                    ..Default::default()
+                },
+                0b0010010,
```
bh2smith, line 108: Also, I know you mention in the PR description how these were generated, but maybe we could include a comment on the breakdown of this value.

bh2smith (approved):

solver/src/encoding.rs
nlordell: `<< 4` is to move the 0 or 1 four bits to the left.

solver/src/encoding.rs
bh2smith: Love the comments!

PR #934, opened 2021-08-03
Retrieve buffers for instance submission

fedgiac:
Adds a new component to the HTTP solver that retrieves the ERC20 buffers available to the solver. This is useful to take advantage of a recent change to the Mip solver: the solution is checked in the solver for correctness and is not sent to the driver if it cannot be matched after considering the buffers.

For now the retrieved buffer is unused. Later, it will be included in the data sent to the various solvers.

### Test Plan

Ignored unit tests:

```
$ export NODE_URL_RINKEBY=...
$ cargo test -- solver::http_solver::buffers --ignored --nocapture
running 2 tests
Buffers: {
    0xc7ad46e0b8a400bb3c915120d284aafba8fc4735: Ok(
        58476096465432915195333071331,
    ),
    0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee: Ok(
        9,
    ),
    0xc778417e063141139fce010982780140aa0cd5ab: Ok(
        157146118900631341,
    ),
    0xbadbadbadbadbadbadbadbadbadbadbadbadbadb: Err(
        Erc20(
            MethodError {
                signature: "balanceOf(address):(uint256)",
                inner: AbiDecode(
                    InvalidName(
                        "please ensure the contract and method you're calling exist! failed to decode empty bytes. if you're using jsonrpc this is likely due to jsonrpc returning `0x` in case contract or method don't exist",
                    ),
                ),
            },
        ),
    ),
}
test solver::http_solver::buffers::test::retrieves_buffers_on_rinkeby ... ok
test solver::http_solver::buffers::test::retrieving_buffers_not_affected_by_eth ... ok
```

### DIFF
--- a/solver/src/solver/http_solver.rs
+++ b/solver/src/solver/http_solver.rs
@@ -6,6 +6,7 @@ use self::{model::*, settlement::SettlementContext};
 use crate::{
     liquidity::{ConstantProductOrder, LimitOrder, Liquidity, WeightedProductOrder},
     settlement::Settlement,
+    settlement_submission::retry::is_transaction_failure,
     solver::Solver,
 };
 use ::model::order::OrderKind;
@@ -269,12 +270,35 @@ impl HttpSolver {
     ) -> Result<(BatchAuctionModel, SettlementContext)> {
         let tokens = self.map_tokens_for_solver(liquidity.as_slice());
 
-        let (token_infos, price_estimates, _buffers) = join!(
+        let (token_infos, price_estimates, mut buffers_result) = join!(
             self.token_info_fetcher.get_token_infos(tokens.as_slice()),
             self.price_estimator
                 .estimate_prices(tokens.as_slice(), self.native_token),
             self.buffer_retriever.get_buffers(tokens.as_slice())
         );
+
+        let _buffers: HashMap<_, _> = buffers_result
+            .drain()
+            .filter_map(|(token, buffer)| match buffer {
+                Err(err) => {
+                    if is_transaction_failure(&err.inner) {
+                        tracing::debug!(
+                            "Failed to fetch buffers for token {} with transaction failure {}",
+                            token,
+                            err
+                        );
+                    } else {
+                        tracing::error!(
+                            "Failed to fetch buffers contract balance for token {} with error {}",
+                            token,
+                            err
+                        );
+                    }
+                    None
+                }
+                Ok(b) => Some((token, b)),
+            })
+            .collect();
 
         let price_estimates: HashMap<H160, Result<BigRational, _>> =
             tokens.iter().cloned().zip(price_estimates).collect();
@@ -519,8 +543,8 @@ mod tests {
             .expect_get_buffers()
             .return_once(move |_| {
                 hashmap! {
-                    buy_token => U256::from(42),
-                    sell_token => U256::from(1337),
+                    buy_token => Ok(U256::from(42)),
+                    sell_token => Ok(U256::from(1337)),
                 }
             });
         let mock_buffer_retriever: Arc<dyn BufferRetrieving> = Arc::new(mock_buffer_retriever);
--- a/solver/src/solver/http_solver.rs
+++ b/solver/src/solver/http_solver.rs
@@ -1,3 +1,4 @@
+pub mod buffers;
 pub mod model;
 mod settlement;
 
@@ -9,6 +10,7 @@ use crate::{
 };
 use ::model::order::OrderKind;
 use anyhow::{ensure, Context, Result};
+use buffers::BufferRetrieving;
 use ethcontract::{Account, U256};
 use futures::join;
 use lazy_static::lazy_static;
@@ -69,6 +71,7 @@ pub struct HttpSolver {
     native_token: H160,
     token_info_fetcher: Arc<dyn TokenInfoFetching>,
     price_estimator: Arc<dyn PriceEstimating>,
+    buffer_retriever: Arc<dyn BufferRetrieving>,
     network_id: String,
     chain_id: u64,
     fee_discount_factor: f64,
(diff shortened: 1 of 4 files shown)


solver/src/solver/http_solver/buffers.rs
```diff
@@ -0,0 +1,66 @@
...
+            web3,
+            settlement_contract,
+        }
+    }
+}
+
+#[cfg_attr(test, mockall::automock)]
+#[async_trait::async_trait]
+pub trait BufferRetrieving: Send + Sync {
+    async fn get_buffers(&self, tokens: &[H160]) -> HashMap<H160, U256>;
```
vkgnosis, line 29: I was going to say I would prefer if this returned a result instead of skipping failing balances but I see now that it is possible that for a bad erc20 token to revert on balance retrieval.
We could still do the same check here as we do in some places (don't remember where from top of my head) that it only skips smart contract errors (Revert and InvalidOpcode) but bubbles up any other error.

solver/src/solver/http_solver/buffers.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+#[async_trait::async_trait]
+impl BufferRetrieving for BufferRetriever {
+    async fn get_buffers(&self, tokens: &[H160]) -> HashMap<H160, U256> {
+        let web3 = Web3::new(self.web3.transport().clone());
```
vkgnosis, line 35:
```suggestion
```

solver/src/solver/http_solver/buffers.rs
```diff
@@ -0,0 +1,66 @@
...
+        let mut batch = CallBatch::new(self.web3.transport());
+
+        let futures = tokens
+            .iter()
+            .map(|address| {
+                let erc20 = ERC20::at(&web3, *address);
```
vkgnosis, line 41:
```suggestion
                let erc20 = ERC20::at(&self.web3, *address);
```

vkgnosis (approved):
An acceptable test imo would be to add a #ignored test that uses the mainnet contracts and some hard coded list of tokens. Then a developer can manually check the result against etherscan.

marcovc (approved):

solver/src/solver/http_solver.rs
```diff
@@ -264,10 +269,11 @@ impl HttpSolver {
...
         let tokens = self.map_tokens_for_solver(liquidity.as_slice());
 
-        let (token_infos, price_estimates) = join!(
+        let (token_infos, price_estimates, _buffers) = join!(
             self.token_info_fetcher.get_token_infos(tokens.as_slice()),
             self.price_estimator
-                .estimate_prices(tokens.as_slice(), self.native_token)
+                .estimate_prices(tokens.as_slice(), self.native_token),
+            self.buffer_retriever.get_buffers(tokens.as_slice())
         );
```
bh2smith: Why retrive/return it if we have not indention of using it?

solver/src/solver/http_solver.rs
```diff
@@ -507,6 +514,17 @@ mod tests {
...
+        let mut mock_buffer_retriever = MockBufferRetrieving::new();
+        mock_buffer_retriever
+            .expect_get_buffers()
+            .return_once(move |_| {
+                hashmap! {
+                    buy_token => U256::from(42),
+                    sell_token => U256::from(1337),
+                }
+            });
+        let mock_buffer_retriever: Arc<dyn BufferRetrieving> = Arc::new(mock_buffer_retriever);
```
bh2smith: Would it not make more sense to test get buffers as a standalone thing directly in file defining get_buffers?

bh2smith (approved):
Looks great, couple of small comments and questions. ~~I also noticed that we have to pass a list of tokens into the get buffer and that we give it the tokens being passed along to the solver. I am guessing then that it is overwritten on each run loop so my question would be what is the purpose of keeping these buffers? Just so the solver run loop doesn't have to do it? Or is it so that the http solver can also make use of them?~~

UPDATE - I see now, it is the purpose of this now is to pass the buffers along to the http solver. At first I thought this had something to do with the withdraw fees scripting for some reason.

solver/src/solver/http_solver.rs
fedgiac: This is the information that needs to be passed to the solver, so I prepared it in the place I think it's needed. It's also a nice split to have a smaller PR. @marcovc also mentioned he might handle the part where he prepares the data to be sent to the solver, so if he wants to do it it would be already there.

solver/src/solver/http_solver.rs
fedgiac: Done!

fedgiac:
Some changes from the previous reviews:
1. added unit tests
2. added logic to handle the eth address
3. directly hand over the error when retrieving the buffers

solver/src/solver/http_solver.rs
```diff
@@ -264,12 +270,36 @@ impl HttpSolver {
...
             self.price_estimator
-                .estimate_prices(tokens.as_slice(), self.native_token)
+                .estimate_prices(tokens.as_slice(), self.native_token),
+            self.buffer_retriever.get_buffers(tokens.as_slice())
         );
 
+        let _buffers: HashMap<_, _> = buffers_result
+            .drain()
+            .filter_map(|(token, buffer)| match buffer {
+                Err(BufferRetrievalError::Erc20(err)) if is_transaction_failure(&err.inner) => {
```
nlordell, line 283: Is there a reason you prefer to handle this here instead of in `BufferRetrieving::get_buffers`?

fedgiac: It was a change based on [this comment](https://github.com/gnosis/gp-v2-services/pull/934#discussion_r681803290). I personally don't mind either way where the error is handled.

nlordell (approved):

PR #935, opened 2021-08-03
[Matcha] Retryable Error Class

bh2smith:
Due to recent and more frequent [alerts](https://gnosisinc.slack.com/archives/CUD4MUCUF/p1627923838009300) with the Matcha solver

```
Matcha inner solver error: SwapQuery result parsing failed: {"reason":"Server Error"}

Aug 2 (2x)
Aug 1 (3x)
July 31 (3x)
July 30 (2x)
```

We implement an error class `MatchaResponseError` (analogous to the work done for Paraswap Error Classification). At this moment we have only encountered "ServerError` and they do not appear together in groups so it seems like a reasonable decision to retry here.

With this in mind we follow the same pattern as in the Paraswap solver, classifying response errors and retrying to a max of 2 times before making one final attempt and finally returning whatever error was last. This should, at least, reduce the noise in our alert channel.

**Note:** This required a slight refactoring of the existing `settle_order`method that broke up the large match statement into smaller pieces. If this is an issue of readability, I would be happy to perform this refactor first.

### Test Plan
Added three unit tests for the retry cycle: retry till i) succeeds ii) exceeds, iii) don't retry.

### DIFF
--- a/solver/src/solver/matcha_solver.rs
+++ b/solver/src/solver/matcha_solver.rs
@@ -508,4 +508,101 @@ mod tests {
             .unwrap();
         assert_eq!(handler.calls(), vec![4321.into()]);
     }
+
+    #[tokio::test]
+    async fn settle_order_retry_until_succeeds() {
+        let mut allowance_fetcher = Box::new(MockAllowanceManaging::new());
+        allowance_fetcher
+            .expect_get_approval()
+            .returning(|_, _, _| Ok(Approval::AllowanceSufficient));
+
+        let mut client = Box::new(MockMatchaApi::new());
+        let mut seq = Sequence::new();
+        client
+            .expect_get_swap()
+            .times(2)
+            .returning(|_| {
+                // Retryable error
+                Err(MatchaResponseError::ServerError(String::new()))
+            })
+            .in_sequence(&mut seq);
+        client
+            .expect_get_swap()
+            .times(1)
+            .returning(|_| {
+                Ok(SwapResponse {
+                    ..Default::default()
+                })
+            })
+            .in_sequence(&mut seq);
+
+        let solver = MatchaSolver {
+            account: account(),
+            client,
+            allowance_fetcher,
+        };
+
+        let order = LimitOrder {
+            ..Default::default()
+        };
+        assert!(solver.settle_order(order).await.is_ok());
+    }
+
+    #[tokio::test]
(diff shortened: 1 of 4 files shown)


solver/src/solver/matcha_solver.rs
```diff
@@ -80,54 +81,67 @@ impl MatchaSolver {
...
+    fn name(&self) -> &'static str {
+        "Matcha"
+    }
+}
+
+impl From<MatchaResponseError> for SettlementError {
+    fn from(err: MatchaResponseError) -> Self {
+        SettlementError {
+            inner: anyhow!("Matcha Response Error {:?}", err),
+            retryable: matches!(err, MatchaResponseError::ServerError(_)),
```
vkgnosis, line 112: Maybe try_settle_order could return MatchaResponseError and we have a function here is `fn is_retryable(error: &MatchaResponseError) -> bool`. I don't think the intermediate SettlementError type is that useful.

vkgnosis (approved):

solver/src/solver/matcha_solver.rs
bh2smith: I was mostly doing this to be consistent with the pattern we have for Paraswap solver. I see that is not super important to have here, but just thought that it might become more relevant to do things this way once we encounter more than just one retryable matcha error. 

I will revert this part for now until it is needed. Would you also suggest that this `SettlementError` isn't needed on the Paraswap side either?

vkgnosis: It is no big deal either way. It just felt like there no advantage in storing it in a struct when a plain function (or could be implemented on MatchaResponseError directly) suffices.
If the pattern is the same for Paraswap then we could make the same change there too.

bh2smith: Actually the reason for this goes a bit beyond just the retryable field. We also use this so that the ? operator can be used within try_settle_order. I think it is probably best to leave it as it is.

fleupold:
If we find ourselves implementing this type of retry logic in other single order solvers as well, we could also implement an abstract "retyring solver" that operates on an abstract error type which implements `should_retry`. This way we don't have to copy paste that retry loop in each solver.

bh2smith:
Yes. I had started thinking of this but wanted to see it in action at least for two before making the generalized approach!  Great idea.

Issue #819, opened 2021-07-06
[Paraswap] PriceQuery Errors


bh2smith:
In a recent PR #806 - we introduced a Paraswap Error enum for `TransactionBuilderQuery` responses to address the "price changed" error (reported in #802 and addressed in #812). However, there is also another class of response errors coming from `PriceQuery`.

In particular, we have recently seen both of

- ParaSwap inner solver error: PriceQuery result parsing failed: {"error":"Price Timeout"}
- ParaSwap inner solver error: PriceQuery result parsing failed: {"error":"computePrice Error"}

Apart from determining exactly what is causing this (the timeout it fairly clear), it could be useful to enumerate the possible errors and handle them accordingly. 

UPDATE: We haven't seen "Price Timeout" Since July 6. I will remove it from focus here.

bh2smith:
Closed by #937

PR #937, opened 2021-08-03
[Paraswap] include computePrice Error into ParaswapResponseError

bh2smith:
Addresses #819 

We have seen several errors of the form 
```
ParaSwap inner solver error: PriceQuery result parsing failed: {"error":"computePrice Error"}
```
This PR aims to handle the error in a more appropriate way than now. At this moment, its unclear still how to handle this error so this PR will sit in Draft State until we reach some kind of conclusion. 

Note that there is an open line of questions regarding this error [here](https://gnosisinc.slack.com/archives/C0203Q9R4JJ/p1626351590002200) with the Paraswap team.

As a simple suggestion, we could consider adding this to the list of retryable errors.


solver/src/solver/paraswap_solver/api.rs
```diff
@@ -29,18 +29,36 @@ pub struct DefaultParaswapApi {
...
+                                    message
+                                ))),
+                            }
+                        }
+                        Err(err) => Err(ParaswapResponseError::DeserializeError(err)),
+                    }
+                }
+                Err(text_fetch_err) => Err(ParaswapResponseError::TextFetch(text_fetch_err)),
+            },
+            Err(send_err) => Err(ParaswapResponseError::Send(send_err)),
```
nlordell, line 60: I feel like there is quite a bit of right-ward drift in this block. I would prefer to use something like `map_err` so that you can use the `?` try operator:
```rust
reqwest::get(url)
    .await
    .map_err(ParaswapResponseError::Send)?
    .text()
    .await
    .map_err(ParaswapResponseError::TextFetch)?
    // ...
```

bh2smith: Yes! I was hoping someone would come around and make this kinda suggestion!

bh2smith: Fixed this, thanks so much!

vkgnosis (approved):

solver/src/solver/paraswap_solver/api.rs
```diff
@@ -29,45 +29,70 @@ pub struct DefaultParaswapApi {
...
             .text()
-            .await?;
-        tracing::debug!("Response from Paraswap API (price): {}", text);
-
-        serde_json::from_str::<PriceResponse>(&text)
-            .context(format!("PriceQuery result parsing failed: {}", text))
+            .await
+            .map_err(ParaswapResponseError::TextFetch)?;
+        tracing::debug!("Response from Paraswap API (price): {}", response_text);
+        let raw_response = serde_json::from_str::<RawPriceResponse>(&response_text);
```
nlordell, line 43: ubernit: You can reduce the number of branches you need to handle :point_down: by mapping the error and using the `?` try operator first.
```suggestion
        let raw_response = serde_json::from_str::<RawPriceResponse>(&response_text)
            .map_err(ParaswapResponseError::DeserializeError)?;
```

Just thought it would be more consistent with the error handling in the rest of the function, but no strong feelings about this one (since there is barely any "rightward drift" here).

solver/src/solver/paraswap_solver/api.rs
```diff
@@ -29,45 +29,70 @@ pub struct DefaultParaswapApi {
...
+            .await
+            .map_err(ParaswapResponseError::TextFetch)?;
+        parse_paraswap_response_text(&response_text, &query_str)
     }
 }
 
 #[derive(Deserialize)]
 #[serde(untagged)]
-enum RawResponse {
+enum RawPriceResponse {
```
nlordell, line 77: nit: We could make a single `RawResponse` type that can be used by all routes:
```
enum RawResponse<Ok> {
    ResponseOk(Ok),
    ResponseErr { error: String },
}
```

Then, the price route would use `RawResponse<PriceResonse>` and the transaction route would use `RawResponse<TransactionBuilderResponse>`.

nlordell (approved):
Looks good! Just some minor nits.

bh2smith:
So I will merge this now as an independent contribution to error classification, but follow up with a PR that generalizes the retry logic for these (request based) solver types and include retry on, say, computePrice Error.

PR #938, opened 2021-08-03
Fix executed amount computations for the naive solver

### LINKED ISSUE #904: [Contract Upgrade] Solver Arithmetic
It was mentioned (cc @fedgiac) that there was a `math.floor` used in a place where `math.ceil` (or vice-versa) was used that should probably be considered when performing this migration. This may have something to do with arithmetic that was changed in the smart contract due to auditor's recommendation. 

TODO - This task needs some elaboration.

nlordell:
Fixes #904 

It turns out that the adjusted buy amount computation causes problems in two places that need exact executed buy amounts:
- Computing the amount of WETH to unwrap for Ether buy orders
- The naive solver

Fixing the rounding error in the naive solver is quite complicated. This PR works around the error by summing up the total amount actually needed from the pools and using that instead in the interaction instead of the amount computed for the price. Since rounding is done in favour of the traders, this means that it could be that we don't quite swap enough from Uniswap in order for the settlement to balance out. SInce this rounding error is capped at 1wei per trade, we just take this amount from the buffers by asking Uniswap pools to trade a little more than needed. Since the amounts are small and taken from the sell amount (so there should always be fees there to cover the rounding error) this shouldn't be a problem. Notice that this required the E2E tests to change to have order provide some fee amount so there are sufficient buffers for this rounding error. 

### Test Plan

CI should pass and E2E test should finally go back to working!

### DIFF
--- a/solver/src/liquidity.rs
+++ b/solver/src/liquidity.rs
@@ -11,6 +11,8 @@ use model::{order::OrderKind, TokenPair};
 use num::{rational::Ratio, BigRational};
 use primitive_types::{H160, U256};
 use shared::sources::balancer::pool_fetching::PoolTokenState;
+#[cfg(test)]
+use shared::sources::uniswap::pool_fetching::Pool;
 use std::collections::HashMap;
 use std::sync::Arc;
 use strum_macros::{AsStaticStr, EnumVariantNames};
@@ -120,6 +122,18 @@ pub struct ConstantProductOrder {
 impl std::fmt::Debug for ConstantProductOrder {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(f, "Constant Product AMM {:?}", self.tokens)
+    }
+}
+
+#[cfg(test)]
+impl From<Pool> for ConstantProductOrder {
+    fn from(pool: Pool) -> Self {
+        Self {
+            tokens: pool.tokens,
+            reserves: pool.reserves,
+            fee: pool.fee,
+            settlement_handling: tests::CapturingSettlementHandler::arc(),
+        }
     }
 }
 
--- a/solver/src/liquidity.rs
+++ b/solver/src/liquidity.rs
@@ -4,13 +4,13 @@ pub mod slippage;
 pub mod uniswap;
 
 use crate::settlement::SettlementEncoder;
-use anyhow::{ensure, Context as _, Result};
+use anyhow::Result;
 #[cfg(test)]
 use model::order::Order;
 use model::{order::OrderKind, TokenPair};
 use num::{rational::Ratio, BigRational};
 use primitive_types::{H160, U256};
-use shared::{conversions::U256Ext, sources::balancer::pool_fetching::PoolTokenState};
+use shared::sources::balancer::pool_fetching::PoolTokenState;
 use std::collections::HashMap;
 use std::sync::Arc;
 use strum_macros::{AsStaticStr, EnumVariantNames};
@@ -61,11 +61,6 @@ impl std::fmt::Debug for LimitOrder {
     }
 }
 
-pub struct Price {
-    pub sell_price: U256,
-    pub buy_price: U256,
-}
-
 impl LimitOrder {
     /// Returns the full execution amount for the specified limit order.
     pub fn full_execution_amount(&self) -> U256 {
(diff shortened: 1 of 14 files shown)


solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -175,61 +138,24 @@ pub mod tests {
         );
     }
 
-    #[test]
```
nlordell, line 178: moved to `solver/src/settlement.rs`

solver/src/solver/baseline_solver.rs
```diff
@@ -59,42 +59,76 @@ enum AmmOrder {
     WeightedProduct(WeightedProductOrder),
 }
 
+impl BaselineSolvable for ConstantProductOrder {
```
nlordell, line 62: This was done to be able to call `BaselineSolvable` methods directly on `ConstantProductOrder`s.

solver/src/liquidity.rs
```diff
@@ -123,6 +125,18 @@ impl std::fmt::Debug for ConstantProductOrder {
     }
 }
 
+#[cfg(test)]
```
bh2smith, line 128: Now I know how to implement something only for tests. Useful!

solver/src/liquidity/offchain_orderbook.rs
```diff
@@ -175,61 +138,24 @@ pub mod tests {
...
         let prices = hashmap! {
             native_token.address() => U256::from(100),
             sell_token => U256::from(200),
         };
-        let executed_buy_amount = U256::from(2 * 1337);
         let order = Order {
             order_creation: OrderCreation {
                 buy_token: BUY_ETH_ADDRESS,
                 sell_token,
+                sell_amount: 1337.into(),
```
bh2smith, line 158: I see that we insert sell amounts here where there were none before. Given that they were previously defaults of zero, how does this relate to the PRs intended purpose?

solver/src/settlement.rs
```diff
@@ -151,6 +188,20 @@ impl Settlement {
         self.encoder.trades()
     }
 
+    /// Returns an iterator of all executed trades.
+    pub fn executed_trades(&self) -> impl Iterator<Item = TradeExecution> + '_ {
```
bh2smith, line 192: Is this the place where the `import as _` is used? If so it seems its only purpose is here, what is the advantage or reason for import `U256Ext as _`?

solver/src/settlement/settlement_encoder.rs
```diff
@@ -1,6 +1,6 @@
-use super::{Interaction, Trade};
+use super::{Interaction, Trade, TradeExecution};
 use crate::{encoding::EncodedSettlement, interactions::UnwrapWethInteraction};
-use anyhow::{anyhow, bail, ensure, Result};
+use anyhow::{bail, ensure, Context as _, Result};
```
bh2smith, line 3: Similar question here, I can't see that `_` is actually used anywhere. What is the meaning of this?

solver/src/solver/baseline_solver.rs
```diff
@@ -59,42 +59,76 @@ enum AmmOrder {
...
+
 impl BaselineSolvable for Amm {
     fn get_amount_out(&self, out_token: H160, input: (U256, H160)) -> Option<U256> {
         match &self.order {
-            AmmOrder::ConstantProduct(order) => amm_to_pool(order).get_amount_out(out_token, input),
-            AmmOrder::WeightedProduct(order) => amm_to_weighted_pool(order)
-                .ok()?
-                .get_amount_out(out_token, input),
+            AmmOrder::ConstantProduct(order) => order.get_amount_out(out_token, input),
+            AmmOrder::WeightedProduct(order) => order.get_amount_out(out_token, input),
```
bh2smith, line 110: Nice!

solver/src/solver/naive_solver/multi_order_solver.rs
```diff
@@ -77,7 +78,7 @@ pub fn solve(
 /// for that pair is not available.
 ///
 fn solve_orders(
-    orders: impl Iterator<Item = LimitOrder> + Clone,
+    orders: &[LimitOrder],
```
bh2smith, line 81: Love this! Is this something we could have always done? Or do some of the changes made here make it possible?

bh2smith (approved):
Nice workaround and cleanup! Must have been tedious.

solver/src/liquidity/offchain_orderbook.rs
nlordell: Its because adding trades to the settlement encode now validates some values required for computing executed buy, sell and fee amount. Having them as 0 makes orders not tradable.

solver/src/settlement.rs
nlordell: `use Trait as _` allows you to use trait methods without importing the trait name into scope. Here the `'_` (notice the `'`) is a lifetime - specifically the lifetime of `&self`. This means I return an iterator which lives as long as the `&self` reference that was passed into the method.

It is needed because a closure that is being returned captures `&self`.

solver/src/settlement/settlement_encoder.rs
nlordell: :point_up:

solver/src/solver/naive_solver/multi_order_solver.rs
nlordell: No, we could have always done it. I did it mostly because I needed to iterate over `orders` twice and didn't want to clone a whole `Vec` to do it.

solver/src/settlement/settlement_encoder.rs
bh2smith: So now that I know this, doesn't that mean there are probably a bunch of places in the code we could be doing this that clippy doesn't complain about?

nlordell: Yeah, but its not a big deal IMO (and doesn't change what the code actually does at all). Its just a habit.

solver/src/solver/naive_solver/multi_order_solver.rs
```diff
@@ -115,33 +116,62 @@ fn solve_without_uniswap(
...
-            .with_liquidity(&order, order.full_execution_amount())
+            .with_liquidity(order, order.full_execution_amount())
             .ok()?;
     }
 
+    // Because the smart contracts round in the favour of the traders, it could
+    // be that we actually require a bit more from the Uniswap pool in order to
+    // pay out all proceeds. We move the rounding error to the sell token so that
+    // it either comes out of the fees or existing buffers. Compute that amount:
+    let uniswap_out_with_rounding = big_int_to_u256(&settlement.executed_trades().fold(
```
vkgnosis, line 147: In the description you mention the extra amount is capped at 1 wei per trade. So would an alternative implementation here simply add +- 1 to uniswap_out per trade here? Just trying to see if I understand this right.

nlordell: It would be to add `orders.len()` to the final output amount. Its not always needed though (if there is one order for example, then there won't be any rounding issues).

vkgnosis (approved):

PR #943, opened 2021-08-04
Use websocket based gasnow estimator

vkgnosis:
Better caching behavior. Won't make end users wait when gas cache is out of date.

I hard coded the config values here because I don't foresee we will need to adjust them.

### Test Plan
manually run

### DIFF
--- a/shared/src/gas_price_estimation.rs
+++ b/shared/src/gas_price_estimation.rs
@@ -1,11 +1,14 @@
 use crate::Web3;
 use anyhow::{anyhow, Context, Result};
 use gas_estimation::{
-    EthGasStation, GasNowGasStation, GasPriceEstimating, GnosisSafeGasStation,
+    EthGasStation, GasNowWebSocketGasStation, GasPriceEstimating, GnosisSafeGasStation,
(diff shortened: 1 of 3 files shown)


shared/src/gas_price_estimation.rs
```diff
@@ -57,7 +60,14 @@ pub async fn create_priority_estimator(
                 if !is_mainnet(&network_id) {
                     return Err(anyhow!("GasNow only supports mainnet"));
                 }
-                estimators.push(Box::new(GasNowGasStation::new(client.clone())))
+                let mut estimator = GasNowWebSocketGasStation::new(Duration::from_secs(30));
```
nlordell, line 63: ubernit: Can this value be a named constant just so I know what the 30 seconds are? In theory I could just look it up in the `gas-esimation` crate what it means? Something like:
```suggestion
                let max_update_age = Duration::from_secs(30);
                let mut estimator = GasNowWebSocketGasStation::new(max_update_age);
```

No strong feelings though, so feel free to ignore.

nlordell (approved):
Looks good.

I also think hardcoded values here are fine. We can make them configurable if we ever need to change them later.

PR #946, opened 2021-08-05
Forward buffers to http solver

fedgiac:
Forwards buffers to the http solvers. This will make the http solver avoid sending solutions if the amounts are too large for proceeds+buffer.

Left to do in the solver: after seeing if these changes are working as expected, we should probably change the solver to default to zero buffer if `internal_buffer` is `None`, otherwise we'll still have unsettlable solutions for tokens that come from outside the settlement itself, like for example in the middle of a path. (Currently it defaults to infinite for backward compatibility.)

### Test Plan
Try to run the Mip solver along with the driver and see the instance saved by the solver. `internal_buffer` was compatible with the balances of the tokens at the time.
```
{
  "tokens": {
    "0xc778417e063141139fce010982780140aa0cd5ab": {
      "decimals": 18,
      "alias": null,
      "normalize_priority": 1,
      "external_price": "1.0",
      "internal_buffer": "7728170422400341"
    },
    "0x4dbcdf9b62e891a7cec5a2568c3f4faf9e8abe2b": {
      "decimals": 6,
      "alias": null,
      "normalize_priority": 0,
      "external_price": "4049235.153344708",
      "internal_buffer": "301719007662"
    }
  },
...
```

### DIFF
--- a/model/src/u256_decimal.rs
+++ b/model/src/u256_decimal.rs
@@ -54,3 +54,44 @@ where
 
     deserializer.deserialize_str(Visitor {})
 }
+
+/// Converts an amount of units of an ERC20 token with the specified amount of
+/// decimals into its decimal representation as a string.
+///
+/// # Examples
+///
+/// ```rust
+/// use model::u256_decimal::format_units;
+///
+/// assert_eq!(format_units(42u64.into(), 0), "42");
+/// assert_eq!(format_units(1_337_000u64.into(), 6), "1.337000")
+/// ```
+pub fn format_units(amount: U256, decimals: usize) -> String {
+    let str_amount = amount.to_string();
+    if decimals == 0 {
+        str_amount
+    } else if str_amount.len() <= decimals {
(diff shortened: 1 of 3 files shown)


marcovc:
Just a question: why isn't "internal_buffer" an integer?

vkgnosis (commented):
It is inconsistent that some erc20 amounts are passed to the solver in atoms and some in base units (taking decimals into account). Unless I am misremembering that the other amounts are all in atoms like amounts in limit orders.
Should we change the buffers to also be in atoms?

Even if we don't use it your formatting function is useful and should be kept.

marcovc:
> Should we change the buffers to also be in atoms?

That seems more consistent to me as well!

fedgiac:
The PR should be ready on the services end. I've run into a small unrelated issue with the solver and I'd like to sort it out before merging, since this PR might allow that issue to surface.

vkgnosis (approved):

marcovc (approved):
Thanks @fedgiac !

PR #948, opened 2021-08-05
Automatically detect solvers being down and alert

vkgnosis:
This is a proof of concept of an idea I had how we can detect the kind of down time we had on prod recently automatically.
On a high level what it does is:
- Check if there have been no trades in the last X minutes
- If so then check if there are any open (solvable) orders
- If so then check if any of them is probably settlable (price is ok)
- If so then create an alert

This is an improvement over the most simple form of this idea which would just be step 1. The problem with that is that there might be no orders whatsoever which shouldn't cause an alert. Or we would have to set the timeout to multiple hours which would mean a longer delays until the alert.
Next if we only checked that a solvable order existed we would still get false positives because those orders might have a price that makes them unsettlable.
So as a final step we also verify that the price is ok. At the moment this done with a simple invocation of the matcha api. Any simple third party could be used here. I picked Matcha over Paraswap because I didn't want to have to retrieve the decimals of the tokens.
I also didn't want to depend on an ethereum node so the way trades are detected is by polling the solvable_orders endpoint instead of listening to on chain events.

The matcha code is slightly duplicated because I didn't want to move the existing code out of the solver crate and I'd like this application to be more independent of the existing code path for robustness.
We could even move this to a standalone repository so it isn't affected by potential CI build issues in this repository but I am not sure about this yet.

If we like this idea then the only thing left to do is move some constants to command line arguments and start running this on staging.

### Test Plan
manual tests and looking at the produced logs

### DIFF
--- a/alerter/src/main.rs
+++ b/alerter/src/main.rs
@@ -1,0 +1,296 @@
+// This application observes the order book api and tries to determine if the solver is down. It
+// does this by checking if no trades have been made recently and if so checking if it finds a
+// matchable order according to an external price api (matcha). If this is the case it alerts.
+
+use anyhow::{Context, Result};
+use model::{
+    order::{Order, OrderCreation, OrderKind, OrderStatus, OrderUid, BUY_ETH_ADDRESS},
+    u256_decimal,
+};
+use primitive_types::{H160, U256};
+use reqwest::Client;
+use std::time::{Duration, Instant, SystemTime};
+use structopt::StructOpt;
+use url::Url;
+
+struct OrderBookApi {
+    base: Url,
+    client: Client,
+}
+
+impl OrderBookApi {
+    pub fn new(client: Client, base_url: &str) -> Self {
+        Self {
+            base: base_url.parse().unwrap(),
+            client,
+        }
+    }
+
+    pub async fn solvable_orders(&self) -> reqwest::Result<Vec<Order>> {
+        let mut url = self.base.clone();
+        url.set_path("/api/v1/solvable_orders");
+        self.client
+            .get(url)
+            .send()
+            .await?
+            .error_for_status()?
+            .json()
+            .await
+    }
+
+    pub async fn order(&self, uid: &OrderUid) -> reqwest::Result<Order> {
+        let mut url = self.base.clone();
+        url.set_path(&format!("/api/v1/orders/{}", uid));
+        self.client
+            .get(url)
+            .send()
+            .await?
+            .error_for_status()?
+            .json()
+            .await
+    }
+}
+
+// Converts the eth placeholder address to weth. Leaves other addresses untouched.
+fn convert_eth_to_weth(token: H160) -> H160 {
+    let weth: H160 = "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2"
+        .parse()
+        .unwrap();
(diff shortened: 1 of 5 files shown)


alerter/src/main.rs
```diff
@@ -0,0 +1,248 @@
...
+
+struct MatchaApi {
+    base: Url,
+    client: Client,
+}
+
+impl MatchaApi {
+    pub fn new(client: Client) -> Self {
+        Self {
+            base: "	https://api.0x.org".parse().unwrap(),
```
bh2smith, line 57: Is the additional white space here intended?
```suggestion
            base: "https://api.0x.org".parse().unwrap(),
```

alerter/src/main.rs
```diff
@@ -0,0 +1,248 @@
...
+        Ok(())
+    }
+}
+
+#[tokio::main]
+async fn main() {
+    let time_format_string = "%Y-%m-%dT%H:%M:%S%.3fZ";
+    tracing_subscriber::fmt::fmt()
+        .with_timer(tracing_subscriber::fmt::time::ChronoUtc::with_format(
+            String::from(time_format_string),
```
bh2smith, line 212: General Question: Is there a reason you chose to cast to string only inside the constructor here rather than initializing `time_format_string` immediately as with `String::from`?

bh2smith (approved):
Seems pretty cool. I suppose this would be introduced into the main code only in a follow up PR.

alerter/src/main.rs
vkgnosis: Functionally it doesn't matter. Just slightly easier to read this way. I copied this code from shared/src/tracing.rs

fleupold (commented):
Very good initiative and a clever idea. One thing that this script would not yet catch (not sure if it's easy to add) would be it the API is down for some reason and e.g. doesn't allow placing orders.

Have you thought about recording the "number of matchable orders" with prometheus instead? This way some of the configuration logic around `min_alert_interval` etc would be outsourced to another component instead of this script.

In any case we should consider routing these types of alerts to a high signal slack channel that is easier for the oncall to follow out of normal work hours.

vkgnosis:
> the API is down

It would detect solvable_orders being down. Wouldn't detect if placing new orders stopped working. Wouldn't be too hard to try to place an order at a price it won't get matched at. I'd consider this a feature we can add later.

> Have you thought about recording the "number of matchable orders" with prometheus instead?

I hadn't thought about it. On first glance I feel that managing the alert from within this application is simpler than going through prometheus. On the other hand having it as a metric could be useful to look at over time and is in some ways more flexible because a prometheus alert query is easier to change than the source code here. I think I like keeping it as it is for now and then adding the metric and prometheus alert to see how it feels.

nlordell:
Would it be possible to have this running in a thread/fiber in the `solver` executable and then report the number of orders that are solvable but have not been traded in a while (basically the orders that would get passed to `fn alert(&Order)`)?

Otherwise, I think this is a very cool idea for monitoring health/liveliness.

vkgnosis:
It would be possible but I prefer it's own executable so they are less coupled together. Maybe there is an issue where the solver tokio runtime freezes up and if they both ran together the alerts from here would stop too.

nlordell (approved):

PR #951, opened 2021-08-09
Reject orders selling WETH for ETH

nlordell:
This PR just rejects orders that have `sell_token` as the native token (WETH on Mainnet and Rinkeby and WXDAI on xDai) with an appropriate error.

### Test Plan

Added a unit test to verify logic used in same token check.

### DIFF
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -7,7 +7,9 @@ use anyhow::Result;
 use chrono::Utc;
 use contracts::WETH9;
 use futures::TryStreamExt;
-use model::order::{BuyTokenDestination, OrderCancellation, OrderCreationPayload, SellTokenSource};
+use model::order::{
+    BuyTokenDestination, OrderCancellation, OrderCreation, OrderCreationPayload, SellTokenSource,
+};
 use model::{
     order::{Order, OrderStatus, OrderUid, BUY_ETH_ADDRESS},
     DomainSeparator,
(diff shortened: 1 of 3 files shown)


vkgnosis (approved):

orderbook/src/orderbook.rs
```diff
@@ -516,4 +530,44 @@ mod tests {
...
+        assert!(!has_same_buy_and_sell_token(
+            &OrderCreation {
+                sell_token: H160([0x01; 20]),
+                buy_token: H160([0x02; 20]),
+                ..Default::default()
+            },
+            &native_token,
+        ));
+        // Sell token set to 0xeee...eee has no special meaning, so it isn't
+        // considered buying and selling the same token.
```
fedgiac, line 563: With the same argument you can argue that `has_same_buy_and_sell_token(BUY_ETH_ADDRESS, BUY_ETH_ADDRESS)` should return false. Maybe we should reject any order that uses`BUY_ETH_ADDRESS` as the sell token in a different check, since it's almost certainly going to be user error.

fedgiac (approved):

orderbook/src/orderbook.rs
nlordell: This should already be the case though, as there should be an error retrieving the ERC20 token balance for address `0xeeee..eeee` when verifying that the user has sufficient balance to trade it.

Issue #733, opened 2021-06-15
Balancer StablePool Integration


bh2smith:
The logical progression of this task is broken down into the following PRs

1. [x] Contract Artifacts #997
2. [x] Fetch Stable Pools from the Graph #1003 
3. [x] Info Fetching #1006 
4. [x] PoolRegistry and Event Handler #1007
5. [x] Refactor some terms #1008
6. [x] Pool Fetching #998 
7. [x] Forward pools to HTTP solver #992 
8. [ ] Include Stable Pools in Baseline Solver
9. [ ] Include Balancer Pools in Price Estimation Service

Just to keep track of the collected knowledge so far. The following link contains some valuable information regarding the differences between Weighted and Stable Pools when the time for integration comes:

https://github.com/gnosis/gp-v2-services/pull/717#discussion_r651684681

To elaborate a bit more on Stable Pools and how they differ from our current `WeightedPool` struct:

Stable Pools do not have weights, but they do have an `amplificationParameter` (which is not part of the weighted pools) that is passed though the methods `_onSwapGivenIn` and `_onSwapGivenOut`. This amplification parameter changes on everyblock and can be fetched via a public view method `getAmplificationParameter`. 

Specifically, these methods call into

```
StableMath._calcOutGivenIn(currentAmp, balances, indexIn, indexOut, swapRequest.amount)
```

We note that generically `getAmplificationParameter` could be evaluated from the value passed into the constructor and the current block height, but there are some manual overrides to the value so we probably shouldn't attempt to compute its current value manually.

With this in mind I propose that 

```rust
pub struct RegisteredStablePool {
    pub pool_id: H256,
    pub pool_address: H160,
    pub tokens: Vec<H160>,
    pub scaling_exponents: Vec<u8>,
    block_created: u64,
}
````

Which is essentially the same as `RegisteredWeightedPool` without the weights.

Our `PoolRegistry` Would stay similar except that We would replace the values of the `pools` hashmap with `BalancerPool` where

```
enum BalancerPool {
  Weighted(WeightedPool),
  Stable(StablePool),
}
```

The `PoolFetching`, which is responsible for getting pool reserves, would then also need to get the current Amplification value whenever it is faced with a StablePool.

bh2smith:
I suppose this issue is outdated due to recent refactor and was essentially completed with the exception of the last task (that should have been a separate issue anyway)

PR #952, opened 2021-08-09
Balancer Stable Pool Integration - Part I

bh2smith:
First part of #733.

Implements Balancer stable pool integration without actually making use of the stable pools just yet

Note that some immediately apparent issues that have come up,

- Balancer ABI sources are no longer available in a tagged repo (cf [this slack discussion](https://gnosisinc.slack.com/archives/C01MZ4QD763/p1628260091020900) with Balancer team)
- Vault and Weighted pool factory are now fetched form a different source and contain differences (iirc - the diff is that one has bytecode and one doesn't). _This needs to be fixed separately._

In a following PR, we will also need to update the HTTP solver to pass stable pools along.

### Test Plan
Updated tests incorporating stable pools and also, a great indication that the graph api is working is to manually run the `balancer_subgraph_query` (ignored by CI) and compare with the total number of Balancer Pools on dune.

```sql
select count(*) from balancer_v2."WeightedPoolFactory_evt_PoolCreated"
select count(*) from balancer_v2."WeightedPool2TokensFactory_evt_PoolCreated"
select count(*) from balancer_v2."StablePoolFactory_evt_PoolCreated"
```


shared/src/sources/balancer/graph_api.rs
```diff
@@ -87,6 +88,88 @@ impl BalancerSubgraphClient {
...
+        } {}
+
+        Ok(RegisteredStablePools {
+            fetched_block_number: block_number,
+            pools,
+        })
+    }
+
+    /// Retrieves the list of registered stable pools from the subgraph.
+    pub async fn get_registered_pools(&self) -> Result<RegisteredPools> {
```
bh2smith, line 133: Here, I have provided two possible avenues of approach. Taking this would fetch all pools and allow us to filter after the fact. Although, since we have multiple pool initializers, this would result in fetching all pools for each pool type and excluding those we don't want. However, it does keep the code here less "duplicated".

vkgnosis:
Should this still be reviewed as is? I don't mind the large PR but want to make sure there isn't more changes or refactoring or splitting up into individual PRs planned.

bh2smith:
> Should this still be reviewed as is? I don't mind the large PR but want to make sure there isn't more changes or refactoring or splitting up into individual PRs planned.

I believe it's good to go as is. With the exception of possibly making changes according to reviews.

shared/src/sources/balancer/graph_api.rs
```diff
@@ -179,27 +245,153 @@ mod pools_query {
...
+                    } else {
+                        unreachable!("Not a stable pool!");
+                    }
                 },
             )
         }
     }
+
+    // impl Pool {
+    //     pub fn pool_type(&self) -> PoolType {
```
vkgnosis, line 307: What's with this large chunk of commented out code here and the small one above?

bh2smith: Something I was experimenting with. I wanted to leave it here for a bit to see if the commented code had any merit. I am happy to remove as its no longer necessary

shared/src/sources/balancer/pool_fetching.rs
```diff
@@ -27,18 +27,138 @@ use std::{
...
+        }
+    }
+
+    pub fn as_stable(&self) -> &StablePool {
+        if let BalancerPool::Stable(pool) = self {
+            pool
+        } else {
+            panic!("Not a stable pool!")
+        }
+    }
```
bh2smith, line 146: In a followup PR - I have changed this to `try_as_weighted` and `try_as_stable` returning `Result<_>` which I believe is more appropriate than the panic here.

shared/src/sources/balancer/pool_storage.rs
```diff
@@ -51,8 +52,69 @@ pub struct RegisteredWeightedPool {
...
+    }
+}
+
+#[cfg(test)]
+impl Default for RegisteredPool {
+    fn default() -> Self {
+        Self::Weighted(RegisteredWeightedPool::default())
+    }
+}
+
```
bh2smith, line 108:
```suggestion

#[cfg(test)]
```

shared/src/sources/balancer/pool_storage.rs
```diff
@@ -51,8 +52,69 @@ pub struct RegisteredWeightedPool {
...
+impl Default for PoolType {
```
bh2smith, line 109:
```suggestion
#[cfg(test)]
impl Default for PoolType {
```

shared/src/sources/balancer/pool_storage.rs
```diff
@@ -83,45 +145,96 @@ impl RegisteredWeightedPool {
...
     /// Used for O(1) access to all pool_ids for a given token
     pools_by_token: HashMap<H160, HashSet<H256>>,
     /// WeightedPool data for a given PoolId
-    pools: HashMap<H256, RegisteredWeightedPool>,
+    weighted_pools: HashMap<H256, RegisteredWeightedPool>,
+    /// StablePool data for a given PoolId
+    stable_pools: HashMap<H256, RegisteredStablePool>,
+    // TODO - either store weighted and stable in separate or together (but pick one!)
+    /// All static Balancer Pool details for a given PoolId
+    registered_pools: HashMap<H256, RegisteredPool>,
```
bh2smith, line 186: Still need to pick on here - but at this moment there is no need for the separation. All the code is using the mixed field.

shared/src/sources/balancer/pool_storage.rs
```diff
@@ -83,45 +145,96 @@ impl RegisteredWeightedPool {
...
+        for pool in initial_weighted_pools {
+            for token in &pool.tokens {
+                pools_by_token
+                    .entry(*token)
+                    .or_default()
+                    .insert(pool.pool_id);
+            }
+            // TODO - pick one!
+            weighted_pools.insert(pool.pool_id, pool.clone());
+            registered_pools.insert(pool.pool_id, RegisteredPool::Weighted(pool));
```
bh2smith, line 210: Ditto - only need registered pools here.

shared/src/sources/balancer/pool_storage.rs
```diff
@@ -149,8 +262,36 @@ impl PoolStorage {
             .collect()
     }
 
-    pub fn pools_for(&self, pool_ids: &HashSet<H256>) -> Vec<RegisteredWeightedPool> {
-        self.pools
+    pub fn weighted_pools_for(&self, pool_ids: &HashSet<H256>) -> Vec<RegisteredWeightedPool> {
```
bh2smith, line 265: May not need this either.

shared/src/sources/balancer/pool_storage.rs
```diff
@@ -199,10 +371,14 @@ impl PoolStorage {
     }
 
     fn delete_pools(&mut self, delete_from_block_number: u64) {
-        self.pools
+        self.weighted_pools
             .retain(|_, pool| pool.block_created < delete_from_block_number);
+        self.stable_pools
+            .retain(|_, pool| pool.block_created < delete_from_block_number);
+        self.registered_pools
+            .retain(|_, pool| pool.block_created() < delete_from_block_number);
```
bh2smith, line 379: TODO

bh2smith:
Thanks coverage bot! I will add more tests!

shared/src/sources/balancer/event_handler.rs
```diff
@@ -85,37 +104,65 @@ impl BalancerPoolRegistry {
...
+        let weighted_pool_ids: HashSet<_> =
+            weighted_pools.union(&two_token_pools).copied().collect();
+        let stable_pools = self
+            .stable_pool_updater
+            .lock()
+            .await
+            .store
             .ids_for_pools_containing_token_pairs(token_pairs);
-        pool_set_1.union(&pool_set_2).copied().collect()
+        weighted_pool_ids.union(&stable_pools).copied().collect()
```
vkgnosis, line 127: Could be written as 
```rust
        let mut pools = HashSet::new();
        pools.extend(
            self.weighted_pool_updater
                .lock()
                .await
                .store
                .ids_for_pools_containing_token_pairs(token_pairs.clone()),
        );
        pools.extend(
            self.two_token_pool_updater
                .lock()
                .await
                .store
                .ids_for_pools_containing_token_pairs(token_pairs.clone()),
        );
        pools.extend(
            self.stable_pool_updater
                .lock()
                .await
                .store
                .ids_for_pools_containing_token_pairs(token_pairs),
        );
        pools
```

shared/src/sources/balancer/event_handler.rs
```diff
@@ -85,37 +104,65 @@ impl BalancerPoolRegistry {
...
             .lock()
             .await
             .store
-            .pools_for(pool_ids);
+            .pools_for(pool_ids)
+            .into_iter()
+            .filter_map(|pool| pool.try_into_weighted().ok())
+            .collect();
         pool_set_1.extend(pool_set_2);
         pool_set_1
```
vkgnosis, line 153: Could be written as
```rust
        let pool_set_1 = self
            .weighted_pool_updater
            .lock()
            .await
            .store
            .pools_for(pool_ids);
        let pool_set_2 = self
            .two_token_pool_updater
            .lock()
            .await
            .store
            .pools_for(pool_ids);
        pool_set_1
            .into_iter()
            .chain(pool_set_2)
            .filter_map(|pool| pool.try_into_weighted().ok())
            .collect()
```

shared/src/sources/balancer/event_handler.rs
```diff
@@ -85,37 +104,65 @@ impl BalancerPoolRegistry {
...
+    pub async fn get_stable_pools(&self, pool_ids: &HashSet<H256>) -> Vec<RegisteredStablePool> {
+        self.stable_pool_updater
+            .lock()
+            .await
+            .store
+            .pools_for(pool_ids)
+            .into_iter()
+            .filter_map(|pool| pool.try_into_stable().ok())
+            .collect()
+    }
```
vkgnosis, line 165: Is it possible that the stable pool update returns pools where try_into_stable fails? On that note it might be nice if `shared/src/sources/balancer.rs`'s big comment at the start could explain the type of pools a bit because that influences the design of code a lot.
If the stable pool updater can only ever returns stable pools then this can be an unwrap or expect here. Or maybe there is a way to restructure the code so that it is clear from the types that it can only be stable pools.
(Maybe better for future PR)

shared/src/sources/balancer/graph_api.rs
```diff
@@ -43,10 +46,10 @@ impl BalancerSubgraphClient {
         )?))
     }
 
-    /// Retrieves the list of registered pools from the subgraph.
-    pub async fn get_weighted_pools(&self) -> Result<RegisteredWeightedPools> {
+    /// Retrieves the list of registered stable pools from the subgraph.
+    pub async fn get_registered_pools(&self) -> Result<RegisteredPools> {
```
vkgnosis, line 50: Comment says it returns registered stable pools but the type is RegisteredPools which internally stores
```rust
pub enum RegisteredPool {
    Weighted(RegisteredWeightedPool),
    Stable(RegisteredStablePool),
}
```
Again I am wondering if this could be represented in the types better if all of the results here must be Stable. Also should we call the function get_registered_stable_pools?

shared/src/sources/balancer/graph_api.rs
```diff
@@ -55,8 +58,8 @@ impl BalancerSubgraphClient {
         while {
             let page = self
                 .0
-                .query::<pools_query::Data>(
-                    pools_query::QUERY,
+                .query::<pools_query::Data<WeightedToken>>(
+                    pools_query::WEIGHTED_POOL_QUERY,
```
vkgnosis, line 62: I'd like to refactor the common part out:
```rust
    // We do paging by last ID instead of using `skip`. This is the
    // suggested approach to paging best performance:
    // <https://thegraph.com/docs/graphql-api#pagination>
    async fn i_dont_have_a_good_name_for_this<T: DeserializeOwned>(
        &self,
        block_number: u64,
        query: &str,
    ) -> Result<Vec<PoolData<T>>> {
        let mut result = Vec::new();
        let mut last_id = H256::default();
        while {
            let page = self
                .0
                .query::<pools_query::Data<T>>(
                    query,
                    Some(json_map! {
                        "block" => block_number,
                        "pageSize" => QUERY_PAGE_SIZE,
                        "lastId" => json!(last_id),
                    }),
                )
                .await?
                .pools;

            let has_next_page = page.len() == QUERY_PAGE_SIZE;
            if let Some(last_pool) = page.last() {
                last_id = last_pool.id;
            }
            result.extend(page);
            has_next_page
        } {}
        Ok(result)
    }

    /// Retrieves the list of registered stable pools from the subgraph.
    pub async fn get_registered_pools(&self) -> Result<RegisteredPools> {
        let block_number = self.get_safe_block().await?;

        let pools_0 = self
            .i_dont_have_a_good_name_for_this::<WeightedToken>(
                block_number,
                pools_query::WEIGHTED_POOL_QUERY,
            )
            .await?
            .into_iter()
            .map(|pool| (pool.factory, pool.into_registered_pool(block_number)));
        let pools_1 = self
            .i_dont_have_a_good_name_for_this::<StableToken>(
                block_number,
                pools_query::STABLE_POOL_QUERY,
            )
            .await?
            .into_iter()
            .map(|pool| (pool.factory, pool.into_registered_pool(block_number)));

        let mut pools_by_factory = HashMap::<H160, Vec<RegisteredPool>>::new();
        for (factory, pool) in pools_0.chain(pools_1) {
            let pool = match pool {
                Ok(pool) => pool,
                Err(err) => todo!(),
            };
            pools_by_factory
                .entry(factory.unwrap_or_default())
                .or_default()
                .push(pool);
        }

        Ok(RegisteredPools {
            fetched_block_number: block_number,
            pools_by_factory,
        })
    }
```

shared/src/sources/balancer/graph_api.rs
```diff
@@ -75,13 +78,42 @@ impl BalancerSubgraphClient {
...
+            let has_next_page = page.len() == QUERY_PAGE_SIZE;
+            if let Some(last_pool) = page.last() {
+                last_id = last_pool.id;
+            }
+
+            for pool in page {
+                pools_by_factory
+                    .entry(pool.factory.unwrap_or_default())
+                    .or_default()
+                    .push(pool.into_registered_pool(block_number)?);
```
vkgnosis, line 110: Should this use the `?` operator? That means the whole function fails if into_registered_pool fails which in turn only fails if `anyhow!("unsupported token with more than 18 decimals")`. I think it would be better to only skip this one pool in that case and completely abort the function. In my refactor suggestion above I left a `todo!` for this case. If we wanted to just skip this pool then it would become a `continue`.

shared/src/sources/balancer/graph_api.rs
bh2smith: Good eye and thanks for the suggestion. I have been trying to think of a way to do this and hadn't yet got back to it.

shared/src/sources/balancer/event_handler.rs
bh2smith: Good point. Yes technically the pools_for method on the store itself should only ever return stable pools. I will turn this into an expect and try to write a test for it. However, it is technically allowed to have a weighted pool in the store since we have abstracted this to be `RegisteredPool`. Would it make more sense to declare `PoolStorage` generically with `PoolStorage<T>` so that `store: HashMap<H256, T>` where T can be one of `RegisteredStablePool` or `RegisteredWeightedPool`?

bh2smith: I will also update the commentary to include Stable Pool related content.

shared/src/sources/balancer/graph_api.rs
bh2smith: That, I believe is just a typo in the comment. I explicitly removed those `get_specific` methods wherever possible in favour of `get_generic`

shared/src/sources/balancer/graph_api.rs
bh2smith: Realistically I think that expect would probably work best here (except for the case you pointed out - with the decimals). However, it is also unlikely for this to happen since we are always operating on known (deployed) balancer pools and they do not support tokens that would cause this error. So I would say we are safe to `expect` here.

bh2smith: In fact, I should really change this into an expect as well.

```rust
18u8.checked_sub(token.decimals).ok_or_else(|| anyhow!("unsupported token with more than 18 decimals"))
```

shared/src/sources/balancer/event_handler.rs
vkgnosis: Hard to say from the get go whether it makes sense. I think it is worth trying so that the types become clearer but maybe it turns out that expect is more practical.

shared/src/sources/balancer/graph_api.rs
vkgnosis: Usually I am hesitant to have an unwrap / expect anywhere that handles third party data (any data that isn't produced from our code) because we don't want to allow a bug in someone else's code to cause a panic in ours.
In this case if it is obvious from the smart contracts that this can't happen then it is probably ok but if it's not too much work I'd still slightly prefer to bubble up the error or error log it.

bh2smith: This isn't exactly "third party" since it is coming from a smart contract and those contracts do not allow pool creation for  tokens with more than 18 decimals.

vkgnosis: There could still be a bug in their smart contract code or even a bug in the ethereum node so that it returns a bogus value there. For example when we are parsing json from an external server even if their documentation says that the response always has an "id" field  and we can verify that in their  code that this is true I still wouldn't like to `expect` this.
More realistic in this case I could see that their contracts are upgradable in which case an upgrade could break the invariant.

bh2smith: https://github.com/balancer-labs/balancer-v2-monorepo/blob/0328ed575c1b36fb0ad61ab8ce848083543070b9/pkg/pool-utils/contracts/BasePool.sol#L476-L487

bh2smith: Ok then... I will change it back...

bh2smith: I replaced your todo with `tracing::error`, then `continue` and some comment explaining the situation with a link.

nlordell (commented):
As a general comment, this PR is very large making it hard to review. Could you maybe split it up into smaller PRs?

bh2smith:
Closing in favour of #997, #998 and #992 respectively

PR #953, opened 2021-08-09
Allow 'null' ref-token in HTTP solver settlements

nlordell:
This PR modifies the HTTP solver model to allow parsing solutions with a `null` ref-token which happens when the solver returns trivial solutions in certain cases.

This PR is trivial since we don't actually use the reference token anywhere (except for one test).

### Test Plan

Added new unit test.


vkgnosis (approved):

bh2smith (approved):
Sweet!

PR #955, opened 2021-08-10
Pass deadline into solvers

vkgnosis:
This allows solver implementations to react to the deadline. The
motivating issue for this is that in the http_solver we first do some
async work and then send the http request which also contains a timeout
that the python solver should obey. We were not taking our initial work
into account here which could cause the python solver to take too long
to respond which would make the driver complain that the solver timed
out.
By knowing the deadline we can adjust the timeout passed to python and
avoid this problem. Similarily the single order solver is smarter now
because it can enforce the timeout on the individual orders it solves
instead of making the whole settlement fail if one of the orders take
too long.
Finally we add some debug logging to the http solver so we can see which
async operation takes how long.

### Test Plan
CI.

### DIFF
--- a/solver/src/solver/http_solver.rs
+++ b/solver/src/solver/http_solver.rs
@@ -10,7 +10,7 @@ use crate::{
     solver::Solver,
 };
 use ::model::order::OrderKind;
-use anyhow::{ensure, Context, Result};
+use anyhow::{anyhow, ensure, Context, Result};
 use buffers::{BufferRetrievalError, BufferRetrieving};
 use ethcontract::{Account, U256};
 use futures::join;
@@ -19,13 +19,14 @@ use num::{BigInt, BigRational, ToPrimitive};
 use primitive_types::H160;
 use reqwest::{header::HeaderValue, Client, Url};
 use shared::{
+    measure_time,
     price_estimate::{PriceEstimating, PriceEstimationError},
     token_info::{TokenInfo, TokenInfoFetching},
 };
 use std::{
     collections::{HashMap, HashSet},
     sync::Arc,
-    time::Duration,
+    time::{Duration, Instant},
 };
 
 // Estimates from multivariate linear regression here:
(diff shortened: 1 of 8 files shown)


vkgnosis:
cc @marcovc

solver/src/solver/http_solver.rs
```diff
@@ -344,18 +347,33 @@ impl HttpSolver {
...
+            .append_pair("instance_name", &instance_name)
+            .append_pair("time_limit", &timeout.as_secs().to_string());
 
         self.config.add_to_query(&mut url);
         let query = url.query().map(ToString::to_string).unwrap_or_default();
-        let mut request = self.client.post(url).timeout(self.timeout);
+        // The default Client created in main has a short http request timeout which we might
+        // exceed. We remove it here because the solver already internally enforces the timeout and
+        // the driver code that calls us also enforces the timeout.
+        let mut request = self.client.post(url).timeout(Duration::from_secs(u64::MAX));
```
marcovc, line 376: Just curious: why not simply remove the ".timeout(...)" part?

marcovc (commented):
Looks good to me. Maybe another solution could have been to remove the "time_limit" from the solver constructor parameter list, and pass an updated "time_limit" in the `solve` function instead of the new `deadline`.

solver/src/solver/http_solver.rs
vkgnosis: If we don't specify any timeout here it will be use the globally configured default timeout on the Client. We need to set it something to overwrite the default.

nlordell (approved):
Wonderful

bh2smith (approved):
Seems great. ~~I see we have gone with deadline in favour of time limit. Is there a reason one is better than the other?~~ Never mind I see the description comment on this.

marcovc:
It seems @e00E is currently on vacations. Should we merge this one, as there are currently timeout alerts that can hopefully benefit from it?

PR #956, opened 2021-08-10
Deny Order Creation for "Banned Users"

bh2smith:
In some scenarios (e.g. DDoS attacks) we may want to prevent certain users from order placement.

This introduces exactly this functionality (configurable at runtime and defaulting to [])

### Test Plan
Could add a test that this happens upon request.

### DIFF
--- a/e2e/tests/services.rs
+++ b/e2e/tests/services.rs
@@ -195,6 +195,7 @@ impl OrderbookServices {
             bad_token_detector,
             Box::new(web3.clone()),
             WETH9::at(web3, native_token),
+            vec![],
         ));
         let maintenance = ServiceMaintenance {
             maintainers: vec![orderbook.clone(), db.clone(), event_updater],
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -669,6 +669,7 @@ components:
               TransferEthToContract,
               UnsupportedToken,
               WrongOwner,
+              BannedUser,
               SameBuyAndSellToken,
               UnsupportedBuyTokenDestination,
               UnsupportedSellTokenSource,
--- a/orderbook/src/api/create_order.rs
+++ b/orderbook/src/api/create_order.rs
@@ -37,6 +37,10 @@ pub fn create_order_response(result: Result<AddOrderResult>) -> impl Reply {
             ),
             StatusCode::BAD_REQUEST,
         ),
+        Ok(AddOrderResult::BannedUser(owner)) => (
+            super::error("BannedUser", format!("User account banned {}", owner)),
+            StatusCode::BAD_REQUEST,
+        ),
         Ok(AddOrderResult::DuplicatedOrder) => (
             super::error("DuplicatedOrder", "order already exists"),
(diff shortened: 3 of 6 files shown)


nlordell (approved):

bh2smith:
Just relaying some of @vkgnosis thoughts before merging (as he is out on vacation and wasn't able to comment because github was down).

> I think it is good to have this functionality but also that it is more appropriate for the database rather then a config option. The reason is that this could be a large list and changing the database doesn't require restarting services.
In hindsight we should also move the denied tokens into the database.
> I don't mind merging as is but if it is not urgent maybe the database route is better.

This seems reasonable and I would think that we could implement this as is and create a separate issue implementing the database constraint for these items after the fact. I was also wondering if we might want to leave the possibility open to un-ban users as well (which might be messy on the DB level) where as denied tokens would likely remain permanent.

nlordell:
> which might be messy on the DB level

Can't we just remove rows?

bh2smith:
> > which might be messy on the DB level
> 
> Can't we just remove rows?

If we create a DB constraint that certain cells can't take on a specific value, then we would have to perform a DB migration script just to re-enable some values. This doesn't seem convenient at all to me.

nlordell:
Ah OK. I didn't think about implementing it as column value constraints, I imagined that the service would just check the database to see if a user is in the banned table when adding and querying orders. Same for tokens.

bh2smith:
> Ah OK. I didn't think about implementing it as column value constraints, I imagined that the service would just check the database to see if a user is in the banned table when adding and querying orders. Same for tokens.

Actually, this seems like its also a pretty good option! Although, given that many devs have no direct access to the production DB, I don't see how we could "make changes on the fly" for convenience sake as proposed.

nlordell:
> I don't see how we could "make changes on the fly" for convenience sake as proposed.


If we want to be able to do this at runtime, I imagine we would have an `admin` API (requiring some login) that would allow adding and removing tokens and users from the ban list. (And not do it through database migrations as that would require new releases and deployments).

bh2smith:
> > I don't see how we could "make changes on the fly" for convenience sake as proposed.
> 
> If we want to be able to do this at runtime, I imagine we would have an `admin` API (requiring some login) that would allow adding and removing tokens and users from the ban list. (And not do it through database migrations as that would require new releases and deployments).

Well now this idea is starting to grow on me. In any case, if you thinks its appropriate, I would merge this as is and create an issue for the proposal to be implemented for both (denied tokens and banned users) as a separate task.

nlordell:
I agree that we can merge as is and then change to making this part of the database.

fleupold (commented):
Isn't `BannedUser` what the already existing type `Forbidden` Status Code was meant for? Also, is the `openapi.yml` updated in some other PR?

bh2smith:
> Isn't `BannedUser` what the already existing type `Forbidden` Status Code was meant for? Also, is the `openapi.yml` updated in some other PR?

Yes perhaps, although I hadn't noticed at the time.

PR #957, opened 2021-08-10
Increase gas price for simulation by maximum base gas fee increase

nlordell:
We've been seeing quite a few errors of the kind:
> err: max fee per gas less than block base fee: address 0xa6DDBD0dE6B310819b49f680F65871beE85f517e, maxFeePerGas: 39663492063 baseFee: 40425950411 (supplied gas 15000000)

My theory is that between getting a gas price estimate and executing the settlement simulation, a block is mined that changes the  base fee and causes the above failure. In order to avoid this error, we increase the gas price of the simulation by 12.5% which is the maximum the base gas fee can increase from one block to the other.

### Test Plan

CI - code path will get hit in E2E test. We should see a decrease in `max fee per gas less than block base fee` errors as well.

### DIFF
--- a/solver/src/settlement_simulation.rs
+++ b/solver/src/settlement_simulation.rs
@@ -11,6 +11,22 @@ use web3::types::{BlockId, BlockNumber};
 
 const SIMULATE_BATCH_SIZE: usize = 10;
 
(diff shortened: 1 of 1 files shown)


bh2smith (approved):
Excellent!

Issue #901, opened 2021-07-27
[Contract Upgrade] Order Placement: appropriate source for balance check.


bh2smith:
Now that an order must specify 1 of 3 balance sources (EOA, vault internal or external), we must add additional consideration for "balance sufficiency"

1. `balanceFrom = EOA`: This is the check we already have (allowance manager approval + sellAmount < balance)
2. `balanceFrom = vault internal`: <FILL ME IN @nlordell>
3. `balanceFrom = vault external`: <FILL ME IN @nlordell>

vkgnosis:
I believe we said we would consider vault internal balances with lower priority than the rest of the contracts upgrade. So while we definitely want to do it it can come after the milestone. Correct me if I am wrong.

nlordell:
I also think this isn't part of the milestone and something we will add after the initial contract update.

bh2smith:
Although the logic here may not be needed for the upgrade, I think it might be reasonable to at least include the unimplemented branches of this check in preparation for these values. That being said, however, it will be quite a long time before front end is actually posting orders with these new types. We just need to be cautious that nobody is sending order requests to our services from elsewhere.

vkgnosis:
If they send orders that need vault balances then they just won't get matched because their order won't appear as solvable. That's not a problem for us but not great experience for them. We could check this when adding an order and refuse if there is a vault balance.

nlordell:
Created https://github.com/gnosis/gp-v2-services/issues/918

PR #958, opened 2021-08-10
Added support for Balancer Vault external balances

nlordell:
Partial #901

This PR adds support in the account balance module for Vault external balances. This allows the orderbook to accept orders that use Vault ERC20 allowances. This mostly involved a lot of plumbing.

This PR is still in draft as I haven't had time to make the test run.

### Test Plan

Added new tests. Run the new tests manually connected to a Ganache instance with:
```
$ NODE_URL=http://127.0.0.1:8545 cargo test -p orderbook -- --ignored --nocapture testnet_vault_external_balance 
    Finished test [unoptimized + debuginfo] target(s) in 0.14s
     Running unittests (target/debug/deps/orderbook-b56b14be77745f86)

running 2 tests
test account_balances::tests::can_transfer_testnet_vault_external_balance ... ok
test account_balances::tests::watch_testnet_vault_external_balance ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 57 filtered out; finished in 1.58s

     Running unittests (target/debug/deps/orderbook-a86dfc4ff750ba67)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

   Doc-tests orderbook

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

### DIFF
--- a/orderbook/src/account_balances.rs
+++ b/orderbook/src/account_balances.rs
@@ -1,9 +1,13 @@
-use anyhow::Result;
-use contracts::ERC20;
+use anyhow::{bail, Result};
+use contracts::{BalancerV2Vault, ERC20};
 use ethcontract::batch::CallBatch;
-use futures::future::{join3, join_all};
+use futures::{
+    future::{self, join_all, BoxFuture},
+    FutureExt as _,
+};
+use model::order::SellTokenSource;
 use primitive_types::{H160, U256};
-use shared::Web3;
+use shared::{Web3, Web3Transport};
 use std::{collections::HashMap, sync::Mutex};
 use web3::types::{BlockId, BlockNumber, CallRequest};
 
@@ -13,14 +17,14 @@ const MAX_BATCH_SIZE: usize = 100;
 #[async_trait::async_trait]
 pub trait BalanceFetching: Send + Sync {
     // Register owner and address for balance updates in the background
-    async fn register(&self, owner: H160, token: H160);
+    async fn register(&self, owner: H160, token: H160, source: SellTokenSource);
 
     // Register multiple owner and addresses for balance updates in the background
-    async fn register_many(&self, owner_token_list: Vec<(H160, H160)>);
+    async fn register_many(&self, owner_token_list: Vec<(H160, H160, SellTokenSource)>);
 
     // Returns the latest balance available to the allowance manager for the given owner and token.
     // Should be non-blocking. Returns None if balance has never been fetched.
-    fn get_balance(&self, owner: H160, token: H160) -> Option<U256>;
+    fn get_balance(&self, owner: H160, token: H160, source: SellTokenSource) -> Option<U256>;
 
     // Called periodically to perform potential updates on registered balances
     async fn update(&self);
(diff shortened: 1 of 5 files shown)


orderbook/src/main.rs
```diff
@@ -151,6 +151,19 @@ async fn main() {
         .expect("Could not get chainId")
         .as_u64();
 
+    let vault = match chain_id {
+        1 | 4 => Some(
+            BalancerV2Vault::deployed(&web3)
```
vkgnosis, line 156: Would be slightly cleaner to inspect the deployments that ethcontract knows about directly and compare it to the chain id instead of first matching on chain id and then calling deployed which internally retrieves the chain id again. That way it is clear that we always look at the same chain id and automatically handle the case where balancer adds a new deployment on a new chain id without touching this.
Not sure if this possible with the current ethcontract api, just mentioning.

orderbook/src/account_balances.rs
```diff
@@ -45,73 +56,74 @@ pub struct Web3BalanceFetcher {
...
-            subscriptions
-                .map(|subscription| {
-                    let _ = guard.entry(subscription).or_default();
-                    subscription
-                })
-                .collect()
-        };
+            for subscription in &subscriptions {
+                guard.insert(*subscription, Default::default());
+            }
```
vkgnosis, line 92: This is a logic change. Before we were making sure there is at least the default but now we are always overwriting with the default. This means that while the fetching happens below the balance and allowance are 0 instead of retaining their previous value. Not sure if this is intended.
Either way we could add a comment here explaining which behavior we want.
Reading this now I'm not sure why we need default values in the first place. Does this prevent duplicate work when getting balances simultaneously somewhere?

orderbook/src/account_balances.rs
```diff
@@ -45,73 +56,74 @@ pub struct Web3BalanceFetcher {
...
+                    (SellTokenSource::External, Some(vault)) => vault_external_balance_query(
+                        &mut batch,
+                        vault.clone(),
+                        token,
+                        subscription.owner,
+                        self.vault_relayer,
+                    ),
+                    _ => async { SubscriptionValue::default() }.boxed(),
+                };
+                future::join(future::ready(subscription), value)
```
vkgnosis, line 115: Alternatively we could have one big async block inside of which we do the match and at the end return the tuple (subscription, value). That way no boxing is needed because the futures all have the same type.

orderbook/src/account_balances.rs
nlordell: Nice catch!

> Reading this now I'm not sure why we need default values in the first place. Does this prevent duplicate work when getting balances simultaneously somewhere?

There might be a subtle difference where we return `None` while updating the account balance where we used to return the previous account balance.

I will change this back to the original and add a comment.

orderbook/src/account_balances.rs
nlordell: Boxing felt easier as depending on the balance type we need to add different calls to the batch.

orderbook/src/main.rs
nlordell: Changed this to retrieve the network and check whether or not it is part of the deployment.

vkgnosis (approved):

orderbook/src/account_balances.rs
nlordell: Should be fixed now.

PR #964, opened 2021-08-12
[Temp] To see the various response texts

bh2smith:
Lately we have been seeing DeserializeError in the form

```
solver error: Matcha Response Error DeserializeError(Error("data did not match any variant of untagged enum RawResponse", line: 0, column: 0))
```

However, it is challenging to debug since I haven't been able to dig up the response (or at least link the many responses we get to this particular error). 

I suspect that the response in this case is actually empty, but this temporary fix should confirm that theory.

### DIFF
--- a/solver/src/solver/matcha_solver/api.rs
+++ b/solver/src/solver/matcha_solver/api.rs
@@ -135,8 +135,8 @@ pub enum MatchaResponseError {
     #[error("uncatalogued error message: {0}")]
     UnknownMatchaError(String),
 
-    #[error(transparent)]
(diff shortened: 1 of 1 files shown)


nlordell (approved):

fleupold (commented):
If we run into something like this again, we could also info/tracing log all HTTP requests and responses that go through the shared HTTP client.

bh2smith:
> If we run into something like this again, we could also info/tracing log all HTTP requests and responses that go through the shared HTTP client.

I also plan to make a generalized retry thing for all the single order solvers ASAP.

PR #968, opened 2021-08-13
*Making discount factor intuitivly accurate

bh2smith:
Rename fee discount factor to fee subsidy factor (synonomous, but encourages the necessary configuration change).

We make all calculations based on (1 - old_value) so that the term itself makes sense.

Now we have "Fee subsidy - 0 means no subsidy" 

Renamed a few variables and included validation at runtime.

### DIFF
--- a/orderbook/src/main.rs
+++ b/orderbook/src/main.rs
@@ -97,7 +97,7 @@ struct Arguments {
     #[structopt(long, env = "BANNED_USERS", use_delimiter = true)]
     pub banned_users: Vec<H160>,
 
-    /// List of token addresses that shoud be allowed regardless of whether the bad token detector
+    /// List of token addresses that should be allowed regardless of whether the bad token detector
     /// thinks they are bad. Base tokens are automatically allowed.
     #[structopt(long, env = "ALLOWED_TOKENS", use_delimiter = true)]
     pub allowed_tokens: Vec<H160>,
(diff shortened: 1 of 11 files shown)


marcovc (approved):
Thanks for this very welcome clarification! Note that there's stuff in gnosis-staging that needs to be updated as well!

nlordell:
Note this change requires a configuration change.

bh2smith:
> Note this change requires a configuration change.

Yep, renamed the PR with an * and will make corresponding config change.

josojo (approved):
thanks

orderbook/src/main.rs
```diff
@@ -284,7 +285,7 @@ async fn main() {
         gas_price_estimator,
         native_token.address(),
         database.clone(),
-        args.shared.fee_discount_factor,
+        args.shared.fee_subsidy_factor,
```
nlordell, line 288: I would have slightly preferred to compute `fee_discoun_factor = 1.0 - options.fee_subsidy_factor` and passed that around everywhere as:
- It would require less code changes
- The computation could have been validated instead of having to remember to call validate each location that the arguments are parsed (in both `orderbook/src/main.rs` and `solver/src/main.rs`).
- Would have kept math simpler in various locations (no `1 - x` everywhere).

bh2smith: That's cool too, I could change it to this in a follow up.

nlordell: I don't think its worth changing again. Just didn't get a chance to review last week and wanted to add my comments now.

PR #974, opened 2021-08-16
Add account setting for settlement simulation and public mempool submission

taminomara:
Part of #827.

We will use these settings in driver later.

### Test Plan

Waiting for contract mocks. I can't use git dependency because both `ethcontract` and `ethcontract-mock` are in the same git repo. So when I add it, there are two `ethcontract` instances: one from crates.io and another from git 😞


taminomara:
Upd: looks like I can use `patch.crates-io` to temporarily override `ethcontract` dependency. I'm writing tests for it now 😊

solver/src/settlement_simulation.rs
```diff
@@ -122,11 +122,11 @@ mod tests {
         let web3 = Web3::new(transport);
         let block = web3.eth().block_number().await.unwrap().as_u64();
         let network_id = web3.net().version().await.unwrap();
-        let mut contract = GPv2Settlement::deployed(&web3).await.unwrap();
-        contract.defaults_mut().from = Some(Account::Offline(
+        let contract = GPv2Settlement::deployed(&web3).await.unwrap();
+        let account = Account::Offline(
             PrivateKey::from_raw([1; 32]).unwrap(),
```
nlordell, line 127: Unrelated to this change, but does this test still work? I could see the case that it just fails because the gas price isn't high enough for the base fee or the offline account :point_up: doesn't have enough ETH for the gas. Could you double check that it still works?

nlordell (approved):

solver/src/settlement_simulation.rs
taminomara: Ran it on rinkeby: test passes, simulation returns an error "GPv2: not a solver".

taminomara: Same with mainnet.

vkgnosis (approved):

PR #975, opened 2021-08-16
Generalized Retry for single order solver

bh2smith:
Adding try_settle_order method to the Single order solver trait, and putting retry logic into the previously existing `settle_order` method which is implemented directly in the trait for all single order solvers.

### Test Plan
Existing CI.


solver/src/solver/oneinch_solver.rs
```diff
@@ -163,7 +164,10 @@ impl Interaction for Swap {
 
 #[async_trait::async_trait]
 impl SingleOrderSolving for OneInchSolver {
-    async fn settle_order(&self, order: LimitOrder) -> Result<Option<Settlement>> {
+    async fn try_settle_order(
```
nlordell, line 167: does this mean that the 1Inch solver now has retry logic (when it previously didn't)?

bh2smith: Yes sir. Although it remains to creat an error class with at least one retryable error.

bh2smith: So at this moment there are actually no retryable errors for 1inch so it would never actually retry.

solver/src/solver/single_order_solver.rs
```diff
@@ -1,21 +1,55 @@
...
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use futures::future;
 use rand::prelude::SliceRandom;
 
 use crate::{
     liquidity::{LimitOrder, Liquidity},
     settlement::Settlement,
-    solver::Solver,
+    solver::{solver_utils::SettlementError, Solver},
```
vkgnosis, line 10: Not directly relevant to this PR but I think this type SettlementError should move into this file because it is part of the single order solver abstraction.

solver/src/solver/paraswap_solver.rs
```diff
@@ -162,6 +152,16 @@ impl ParaswapSolver {
...
+    fn account(&self) -> &Account {
+        &self.account
+    }
+
+    fn name(&self) -> &'static str {
+        "ParaSwap"
+    }
+}
+
+impl ParaswapSolver {
```
vkgnosis, line 164: I think the ParaswapSolver should also have it's settle_order function removed because it duplicates the same function from the trait again.
Furthermore I feel we should remove the whole `fn settle_order(&self, order: LimitOrder) -> Result<Option<Settlement>>` from the trait and instead move it to `struct SingleOrderSolver`. This way it is not possible to overwrite the function in trait implementors which would have caught the ParaswapSolver part I'm commenting on.

bh2smith: This is a great suggestion, although I am running into the issue that inside the `solve` method defined in 

```
impl<I: SingleOrderSolving + Send + Sync> Solver for SingleOrderSolver<I>
```

there is "no method named `settle_order` found for reference `&SingleOrderSolver<I>`"

```
tokio::time::timeout_at(deadline.into(), self.settle_order(order))
   |                                          ^^^^^^^^^^^^ method not found in `&SingleOrderSolver<I>`
```

vkgnosis (approved):

nlordell (approved):

Issue #969, opened 2021-08-13
Pagination on trades and orders calls


alfetopito:
Add pagination to API for both trades and orders calls:

GET /api/v1/orders
GET /api/v1/trades

API should accept an optional parameter `pageSize` stating how many results should be returned at a time. Defaults to `10`.
API should accept an optional parameter `page` stating which is the current page being requested. Defaults to `0`.

## Proposal

`pageSize` must be an integer within the range (inclusive) from 1 - 100
When not set, set to a value outside of the range or an invalid value (undefined, 1.3, `gibberish`, etc), it should default to `10`

The API - to protect itself - should not allow to force-request all db values at once.

The response will probably have to change to include metadata regarding the pagination.
Instead of returning a flat list of results, it'll have to change to something like:

```json
{
  "results": 2123,
  "page": 3,
  "pageSize": 50,
  "results": [{}, ...]
}
```

**Discussion:** 
- We could allow higher `pageSize`, 100 is a conservative proposal
- We could set a higher default `pageSize`, no strong reason to keep it at 10
- The change in the response is a breaking change. Any suggestions to make the transition smooth? Deploy an extra endpoint, alter the response based on a request flag, etc.
- I based the pagination "design" in the most common pattern, but I'm not entirely sure how that would work given that the primary keys are non sequential. Instead of an integer could the page be a sort of db "cursor"?

vkgnosis:
I started looking into pagination from the DB's perspective last week. In order to make it efficient we have to use a model where for each query there is a cursor which is based on an index in the database. For example for when querying orders this is the order uid. When you request a page you say: give me the next page of data starting after order uid X. X is the last order uid you got in your previous response page.
This is efficient because this index is ordered so Postgres can quickly find the starting point of any value in the index. It might become more complicated in some places where we have extra filters or we might need an extra index. Still have to check this for all queries.
We have to figure this out first before we can design the api.

For backward compatibility we can create a /v2/ route for the first time, see how many requests we still get to /v1/ and remove them after a week or so.

alfetopito:
> When you request a page you say: give me the next page of data starting after order uid X. X is the last order uid you got in your previous response page.

Yes, exactly what I had in mind :)

Do you know already what's needed to get this working?

vkgnosis:
I am still in the process of investigating the queries I create with postgres tools to see if they do what I want them to. I'd like to create the PR to adjust the queries myself since I've been working on this already.

alfetopito:
[...]

PR #978, opened 2021-08-17
969/orders pagination

alfetopito:
Part of #969

Adding pagination for Orders query

### Test Plan
TBD, this is still WIP

### DIFF
--- a/orderbook/src/api/get_orders.rs
+++ b/orderbook/src/api/get_orders.rs
@@ -29,6 +29,8 @@ struct Query {
     include_insufficient_balance: bool,
     #[serde(default)]
     include_unsupported_tokens: bool,
+    page_size: Option<u32>,
+    page: Option<u32>,
 }
 
 impl Query {
@@ -36,18 +38,26 @@ impl Query {
         if self.owner.is_none() && self.sell_token.is_none() && self.buy_token.is_none() {
             return Err("need to set at least one of owner, sell_token, buy_token");
         }
-        let to_h160 = |option: Option<&H160Wrapper>| option.map(|wrapper| wrapper.0);
-        Ok(OrderFilter {
-            min_valid_to: self.min_valid_to,
-            owner: to_h160(self.owner.as_ref()),
-            sell_token: to_h160(self.sell_token.as_ref()),
-            buy_token: to_h160(self.buy_token.as_ref()),
-            exclude_fully_executed: !self.include_fully_executed,
-            exclude_invalidated: !self.include_invalidated,
-            exclude_insufficient_balance: !self.include_insufficient_balance,
-            exclude_unsupported_tokens: !self.include_unsupported_tokens,
-            uid: None,
-        })
+
+        return match self.page_size {
+            Some(i) if !(1..=100).contains(&i) => Err("page_size must be between 1 and 100"),
+            _ => {
+                let to_h160 = |option: Option<&H160Wrapper>| option.map(|wrapper| wrapper.0);
+                Ok(OrderFilter {
+                    min_valid_to: self.min_valid_to,
+                    owner: to_h160(self.owner.as_ref()),
+                    sell_token: to_h160(self.sell_token.as_ref()),
+                    buy_token: to_h160(self.buy_token.as_ref()),
+                    exclude_fully_executed: !self.include_fully_executed,
+                    exclude_invalidated: !self.include_invalidated,
+                    exclude_insufficient_balance: !self.include_insufficient_balance,
+                    exclude_unsupported_tokens: !self.include_unsupported_tokens,
+                    uid: None,
+                    page_size: self.page_size,
+                    current_page: self.page,
+                })
+            }
+        };
     }
 }
 
--- a/orderbook/src/database/orders.rs
+++ b/orderbook/src/database/orders.rs
@@ -250,7 +250,16 @@ impl OrderStoring for Postgres {
                 WHEN 'sell' THEN sum_sell < sell_amount \
                 WHEN 'buy' THEN sum_buy < buy_amount \
             END) AND \
-            ($7 OR NOT invalidated);";
+            ($7 OR NOT invalidated) \
+        ORDER BY unfiltered.creation_timestamp \
+        LIMIT $8 \
+        OFFSET $9;";
+
+        let page_size: u32 = filter.page_size.unwrap_or(10);
+        let current_page: u32 = match filter.current_page {
+            Some(p) => page_size * p,
+            None => 0,
+        };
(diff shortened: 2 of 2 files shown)


orderbook/src/api/get_orders.rs
```diff
@@ -29,25 +29,35 @@ struct Query {
...
-            exclude_insufficient_balance: !self.include_insufficient_balance,
-            exclude_unsupported_tokens: !self.include_unsupported_tokens,
-            uid: None,
-        })
+
+        return match self.page_size {
+            Some(i) if !(1..=100).contains(&i) => Err("page_size must be between 1 and 100"),
+            _ => {
+                let to_h160 = |option: Option<&H160Wrapper>| option.map(|wrapper| wrapper.0);
+                Ok(OrderFilter {
```
anxolin, line 46: maybe can be simplified destructuring `self`?

orderbook/src/api/get_orders.rs
```diff
@@ -29,25 +29,35 @@ struct Query {
...
-            buy_token: to_h160(self.buy_token.as_ref()),
-            exclude_fully_executed: !self.include_fully_executed,
-            exclude_invalidated: !self.include_invalidated,
-            exclude_insufficient_balance: !self.include_insufficient_balance,
-            exclude_unsupported_tokens: !self.include_unsupported_tokens,
-            uid: None,
-        })
+
+        return match self.page_size {
+            Some(i) if !(1..=100).contains(&i) => Err("page_size must be between 1 and 100"),
```
anxolin, line 43: Is this creating an array in-memory array of 100 elements for each query just to test belonging of the page size to it in order to validate? 

Why not just having a `>=1` `<=100` conditions?

orderbook/src/database/orders.rs
```diff
@@ -248,7 +250,16 @@ impl OrderStoring for Postgres {
                 WHEN 'sell' THEN sum_sell < sell_amount \
                 WHEN 'buy' THEN sum_buy < buy_amount \
             END) AND \
-            ($7 OR NOT invalidated);";
+            ($7 OR NOT invalidated) \
+        ORDER BY unfiltered.creation_timestamp \
+        LIMIT $8 \
+        OFFSET $9;";
```
anxolin, line 256: If i don't remember bad, these formatting syntax was very powerful, and would allow u to use "named placeholders" instead of numbers. I think its the same efficiency and more readable when u have so many params --> hard to spot a mistake.

anxolin (commented):
nice work!

orderbook/src/api/get_orders.rs
```diff
@@ -29,25 +29,35 @@ struct Query {
...
+                    owner: to_h160(self.owner.as_ref()),
+                    sell_token: to_h160(self.sell_token.as_ref()),
+                    buy_token: to_h160(self.buy_token.as_ref()),
+                    exclude_fully_executed: !self.include_fully_executed,
+                    exclude_invalidated: !self.include_invalidated,
+                    exclude_insufficient_balance: !self.include_insufficient_balance,
+                    exclude_unsupported_tokens: !self.include_unsupported_tokens,
+                    uid: None,
+                    page_size: self.page_size,
+                    current_page: self.page,
```
anxolin, line 57: Are u defining some defaults?

Probably I would even do the things in this orders:
- First do all validations, u validate `page_size` but not `current page` (unless this is zero based, then I guess the type will validate). 
- Construct the struct

orderbook/src/database/orders.rs
anxolin: Is the OFFSET the same as the `current_page`? 

afaik u can implement the navigation using an offset, or a a page number but is not the same. 
U would need to multiply the `page_size X (page_number -1)` to get the offset

i.e. 
page 1, 100 ==> offset 0
page 2, 100 ==> offset 100

orderbook/src/database/orders.rs
```diff
@@ -248,7 +250,16 @@ impl OrderStoring for Postgres {
                 WHEN 'sell' THEN sum_sell < sell_amount \
                 WHEN 'buy' THEN sum_buy < buy_amount \
             END) AND \
-            ($7 OR NOT invalidated);";
+            ($7 OR NOT invalidated) \
+        ORDER BY unfiltered.creation_timestamp \
+        LIMIT $8 \
```
anxolin, line 255: If i don't remember bad, these formatting syntax was very powerful, and would allow u to use "named placeholders" instead of numbers. I think its the same efficiency and more readable when u have so many params --> hard to spot a mistake.

orderbook/src/api/get_orders.rs
nlordell: Clippy complains :stu

orderbook/src/api/get_orders.rs
alfetopito: Right now I'm using hard coded defaults for page_size (10) and default page (0)
Do intend to clear it up and have proper constants, but not there yet.

orderbook/src/database/orders.rs
alfetopito: Take a look at the lines right after this
screenshot_2021-08-17_07-31-45

;)

orderbook/src/api/get_orders.rs
alfetopito: Yep, that's clippy :)
But to be honest, this way of declaring the interval looks more clear and concise to me. 

Also, the range is hard coded, I'd like to get a const for it

orderbook/src/database/orders.rs
alfetopito: Lol,  don't look at the insertion query (https://github.com/gnosis/gp-v2-services/blob/45dd1d177e00bda303cfa64c7af9680e61a2a49c/orderbook/src/database/orders.rs#L162)

orderbook/src/api/get_orders.rs
anxolin: ok, rust is too modern. I find it a bit weird, but im sure theres a reason, and i don't want to make clippy upset

orderbook/src/database/orders.rs
josojo: What happens if &8 is null? Would that still work?

alfetopito: `$8` will either be set or default to `0`, which is equivalent to not having the `OFFSET` clause.

Anyway, this might not look like this at all.
@vkgnosis is working on a different sort of pagination.

This draft is on-hold in the mean time

alfetopito:
Superseded by Valentin's changes

PR #982, opened 2021-08-17
Add presignature table

fedgiac:
Adds a table collecting all `PreSignature` events that are seen in the settlement contract.

A new entry is inserted on every new event, which includes presigning and withdrawing a presignature. To my understanding, reorgs are already taken into account by the surrounding code, and this is why I insert a row each time instead of updating an existing ones.

There is some repeating data in the table, as `owner` is already contained in `order_uid`. I allowed this because I imagined it would be easier to query all presignatures for a specific user.

Note that no data is ever removed from the table. I don't plan to implement any pruning, since I imagine this to be a fairly small table and I didn't see any pruning done in other contract tables of the database.

Possible optimization: of all entries older than `MAX_REORG_BLOCK_COUNT` that have the same order uid, only keep the latest.

### Test Plan

Existing ignored unit tests:
```
$ cargo test --locked --all-features --all-targets  --jobs 1 -- database::events::tests --ignored --test-threads 1
[...]
running 2 tests
test database::events::tests::postgres_events ... ok
test database::events::tests::postgres_repeated_event_insert_ignored ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 57 filtered out; finished in 0.42s
```


database/sql/V013__create_presignatures.sql
```diff
@@ -0,0 +1,12 @@
+-- PreSignature events from the smart contract.
+CREATE TABLE presignatures (
+    block_number bigint NOT NULL,
+    log_index bigint NOT NULL,
+    owner bytea NOT NULL,
+    order_uid bytea NOT NULL,
+    signed boolean NOT NULL,
+    PRIMARY KEY (block_number, log_index)
```
nlordell, line 8: Since presignatures are stored in `mapping(orderUid => bool)`, doesn't it make sense for the primary key to be the order UID in the table?

I guess it depends on what the expectations are for:
- User create presignature for order X
- User remove presignature for order X

In the above scenario, should the presignature row get removed? Should there be a new entry in it?

nlordell: Actually, now that I think about it a bit more, I think there would be an issue if we use the `orderUid` as the primary key, because we emit `PreSignature` events even if the value doesn't change. So
- User sets presignature for order X on block N
- User sets presignature again, for absolutely no reason, on block M with Tx T nonce O
- User has another Tx T' with nonce O in the mempool
- Reorg to block M-1 happens
  - presignature row gets deleted, even if the signature from block N is still there
- T' gets included in the canonical chain instead of T
- No presignature.

database/sql/V013__create_presignatures.sql
```diff
@@ -0,0 +1,12 @@
...
+);
+
+-- Get a specific user's presignature.
+CREATE INDEX presignature_owner ON presignatures USING HASH (owner);
```
nlordell, line 12: Should we index on order UID as well?

nlordell (approved):

database/sql/V013__create_presignatures.sql
vkgnosis: Other tables that store on chain events have this primary key because it uniquely identifies any event. This is also needed to efficiently query all events in some block range which is used for the reord which deletes all events in the last X blocks.
That doesn't mean it has to be a primary key here but it has to be at least an index (primary keys are always indexes).

fedgiac: This table keeps all presign events as independent rows. To my understanding we want to keep all events since it might be necessary to roll back in case of a reorg. This extra information is indeed not needed, the alternative could be storing everything in memory for the first `MAX_REORG_BLOCK_COUNT` and only then finalize the presignatures into the database after waiting that number of blocks. I think the latter solution is fairly more complex (requires reworking the event abstraction) but definitely nicer.
I went for the current implementation because it looked significantly easier, and it can be made (imho) comparably as efficient as the other by pruning unnecessary events in a maintenance step.

vkgnosis (approved):

josojo (approved):
Nice

orderbook/src/database/events.rs
```diff
@@ -152,6 +161,11 @@ async fn delete_events(
         .execute(sqlx::query(QUERY_SETTLEMENTS).bind(delete_from_block_number as i64))
         .await?;
 
+    const QUERY_PRESIGNATURES: &str = "DELETE FROM presignatures WHERE block_number >= $1;";
```
bh2smith, line 164: This query name seems misleading...

bh2smith (approved):

orderbook/src/database/events.rs
```diff
@@ -152,6 +161,11 @@ async fn delete_events(
         .execute(sqlx::query(QUERY_SETTLEMENTS).bind(delete_from_block_number as i64))
         .await?;
 
+    const QUERY_PRESIGNATURES: &str = "DELETE FROM presignature_events WHERE block_number >= $1;";
```
bh2smith, line 164: Could we call this constant something with the word "DELETE" in it? Maybe `DELETE_PRESIGNATURE_QUERY` or `QUERY_DELETE_PRESIGNATURES`

bh2smith (approved):

orderbook/src/database/events.rs
fedgiac: This is an implementation of the function `delete_events`, so I think it's ok if no `DELETE` is used in the variable. Note that the surrounding code also follows the syntax `QUERY_*`, so having all variables read `QUERY_DELETE_*` would be a bit redundant.

PR #984, opened 2021-08-18
Add code coverage to CI

nlordell:
This PR adds code coverage using `cargo-llvm-cov`. I chose this coverage tool because:
- It leverages the built-in LLVM coverage instrumentation from the nightly compiler
- Has incredibly easy setup :tada:

### Test Plan

See code coverage in CI :point_down:

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -53,6 +53,21 @@ jobs:
       - run: cargo run --bin deploy --features bin
         working-directory: contracts
       - run: cargo test --locked --package e2e
+  coverage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Rust
+        run: rustup toolchain install nightly --component llvm-tools-preview
+      - name: Install cargo-llvm-cov
+        run: curl -LsSf https://github.com/taiki-e/cargo-llvm-cov/releases/latest/download/cargo-llvm-cov-x86_64-unknown-linux-gnu.tar.gz | tar xzf - -C ~/.cargo/bin
+      - name: Generate code coverage
+        run: cargo llvm-cov --all-features --lcov --output-path lcov.info
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v2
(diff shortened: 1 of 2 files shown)


bh2smith (approved):
Love this! Also I would be very eager to bring this value up.

.github/workflows/pull-request.yaml
```diff
@@ -59,3 +59,18 @@ jobs:
...
+        run: rustup toolchain install nightly --component llvm-tools-preview
+      - name: Install cargo-llvm-cov
+        run: curl -LsSf https://github.com/taiki-e/cargo-llvm-cov/releases/latest/download/cargo-llvm-cov-x86_64-unknown-linux-gnu.tar.gz | tar xzf - -C ~/.cargo/bin
+      - name: Generate code coverage
+        run: cargo llvm-cov --all-features --lcov --output-path lcov.info
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v2
+        with:
+          files: lcov.info
+          fail_ci_if_error: true
```
fleupold, line 76: Maybe a bit harsh to fail on upload failure, but can readjust if necessary.

fleupold (approved):
Very nice! Let's also add the status badge to the top level README?

taminomara (approved):

.github/workflows/pull-request.yaml
nlordell: AFAIU, the `codecov` CI step is not required for merging - but if it fails to upload it will show an `X` on the status so we are aware. So the result I expected with this (although I may have misunderstood) would be that it would show the `codecov` step as failed, but the PR would still be allowed to merge as long as the other required steps (`openapi` and `rust-*` steps) are green.

vkgnosis (approved):
For reference when I was researching code coverage in gp-v1 the llvm based coverage was still a PR. I think it is the best approach because it is integrated into the compiler so I am happy to see that it is usable now.
Downside is that this requires nightly which might sometimes fail to compile our code. As long as that only causes coverage to fail and not the full CI job then that's fine.

PR #987, opened 2021-08-18
Bump ethcontract to 0.15.1

taminomara:
### Test Plan

CI


contracts/src/lib.rs
```diff
@@ -61,6 +61,20 @@ mod tests {
         }
     }
 
+    impl BatchTransport for ChainIdTransport {
```
taminomara, line 64: Ideally we can remove all these test transports in favour of mock transport. I'll look into this on Friday.

vkgnosis (approved):

PR #988, opened 2021-08-18
Add env variables to alerter

vkgnosis:
Forgot this before.

Although with our decision to make alerts based on metrics I'm not sure if / how we will be using the alerter anymore.

### Test Plan
cargo run --bin alerter -- --help
shows env variables


nlordell:
> Although with our decision to make alerts based on metrics I'm not sure if / how we will be using the alerter anymore.

I think this could still be useful as a "Trading-bot 2.0" that is closer to the repo.

taminomara (approved):

PR #989, opened 2021-08-18
Return orders and trades as Vec instead of Stream from db

vkgnosis:
While a stream is more powerful we were already always collecting into a
Vec anyway. Doing this inside of the trait allows us to collect metrics
for this queries when it didn't before.

### Test Plan
existing tests


nlordell (approved):
Looks good to me.

taminomara (approved):
Looks good to me as well.

bh2smith (approved):
Cool stuff!

PR #990, opened 2021-08-18
Switch to dyn transport

taminomara:
Make code abstract over transport type by switching to `DynTransport` from ethcontract. This change will allow us substituting transport in tests.

### Test Plan

This PR doesn't change any logic, so existing tests should be enough. The only side effect I can see is that we add additional overhead for every node call. I doubt this will affect production, but I'd still kept an eye on performance metrics after we release.

### DIFF
--- a/shared/src/lib.rs
+++ b/shared/src/lib.rs
@@ -24,6 +24,7 @@ pub mod tracing;
 pub mod transport;
 pub mod web3_traits;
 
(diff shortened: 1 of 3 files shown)


vkgnosis (approved):

nlordell (approved):
Yay! This can simplify quite a few tests too.

PR #991, opened 2021-08-18
Use full pool balance instead of base unit for bad token detector

### LINKED ISSUE #983: [Untradable Token] Fraction
The [Fraction token](https://blockscout.com/xdai/mainnet/tokens/0x2bF2ba13735160624a0fEaE98f6aC8F70885eA61/token-transfers) is not tradable on CowSwap, because no pool with enough liquidity exists.

The largest pool is [this](https://info.honeyswap.org/#/pair/0x53704af4ff9031ebd461cb498ca717a4b6bca95c) Honeyswap pool with 0.1 Fraction balance .

Our bad token detector tries however to transfer one unit of the token. The issue is that Fraction by design only has a total supply of 1. Therefore finding a pool with 1 unit balance is never going to happen.

Is there any reason for this magic number? Would we be able to reduce the amount we try to ransfer?

vkgnosis:
This helps for tokens where the base unit is very large.

Fixes #983

### Test Plan
Tested that still works on mainnet. Didn't test the token from the issue because I didn't see a gnosis xdai node that I could use remotely.

### DIFF
--- a/shared/src/bad_token/trace_call.rs
+++ b/shared/src/bad_token/trace_call.rs
@@ -93,35 +93,20 @@ impl BadTokenDetecting for TraceCallDetector {
 
 impl TraceCallDetector {
     pub async fn detect_impl(&self, token: H160) -> Result<TokenQuality> {
-        let instance = ERC20::at(&self.web3, token);
-        let decimals = match instance.decimals().call().await {
-            Ok(decimals) => decimals,
-            Err(err) => {
-                return match EthcontractErrorType::classify(&err) {
-                    EthcontractErrorType::Node => Err(err.into()),
-                    EthcontractErrorType::Contract => {
-                        Ok(TokenQuality::bad("failed to get decimals"))
-                    }
-                }
+        let (take_from, amount) = match self.find_largest_pool_owning_token(token).await? {
+            Some((address, balance)) => {
+                tracing::debug!(
+                    "testing token {:?} with pool {:?} amount {}",
+                    token,
+                    address,
+                    balance
+                );
+                (address, balance)
             }
-        };
-        let amount = match U256::from(10).checked_pow(decimals.into()) {
-            Some(amount) => {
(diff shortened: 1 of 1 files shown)


nlordell (approved):

shared/src/bad_token/trace_call.rs
```diff
@@ -93,32 +93,17 @@ impl BadTokenDetecting for TraceCallDetector {
...
+                    address,
+                    balance
+                );
+                (address, balance)
             }
-            _ => return Ok(TokenQuality::bad("no pool with enough balance")),
+            None => return Ok(TokenQuality::bad("no pool")),
         };
 
         // We transfer one unit (base on decimals) of the token from the amm pool into the
```
taminomara, line 109: Nit: this comment is outdated, we now transfer one atom.

shared/src/bad_token/trace_call.rs
```diff
@@ -93,32 +93,17 @@ impl BadTokenDetecting for TraceCallDetector {
...
-        let amount = match U256::from(10).checked_pow(decimals.into()) {
-            Some(amount) => {
-                tracing::debug!("testing token {:?} with amount {}", token, amount);
-                amount
-            }
-            None => return Ok(TokenQuality::bad("decimals too large")),
-        };
-
-        let take_from = match self.find_largest_pool_owning_token(token).await? {
-            Some((address, balance)) if balance > amount => {
```
taminomara, line 117: If I understand `find_largest_pool_owning_token` correctly, `balance` can be `0`, so we still need at least some check here.

taminomara (commented):
Much clearer and simpler, I like it!

bh2smith (approved):

shared/src/bad_token/trace_call.rs
vkgnosis: I've changed it to check for 0 balance. So it either returns Some with an address that has more than 0 or None.

taminomara (approved):

PR #992, opened 2021-08-18
Stable Pools part 7: Forward Pools to HTTP solver.

bh2smith:
Following #998, this is the third part of #733.

Introducing all the models for passing Stable pool types into the http solver. Note the last remaining todo is the route to handling Stable Pools in the Baseline solver.

### Test Plan
Local test performed that Stable pools reach the point of forwarding to http solver. see inline comment where this was performed.

Include the following print statement here https://github.com/gnosis/gp-v2-services/blob/34af96b23b05a52e0ce4a3b29c87d88a3bdd7767/solver/src/solver/http_solver.rs#L371-L372
```
println!("Stable Pool Orders {:?}", stable_pool_orders.len());
println!("{:?}", stable_pool_orders[0]);
```
and placing an order of `USDC <-> DAI` which is a known [Balancer stable pool.](https://etherscan.io/address/0x06df3b2bbb68adc8b0e302443692037ed9f91b42)

resulting in the 1 expected stable pool
```
Stable Pool Orders 1
Stable Pool AMM [0xdac17f958d2ee523a2206206994597c13d831ec7, 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48, 0x6b175474e89094c44da98b954eedeac495271d0f]
```

Steps to Reproduce:

Run the solver like this:

```sh
cargo run --bin solver --
 --node-url https://mainnet.infura.io/v3/YOUR_INFURA_KEY
 --base-tokens 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2,0x6B175474E89094C44Da98b954EedeAC495271d0F,0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48,0xdAC17F958D2ee523a2206206994597C13D831ec7,0xc00e94Cb662C3520282E6f5717214004A7f26888,0x9f8F72aA9304c8B593d555F12eF6589cC3A579A2,0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599 
 --solvers Mip,Quasimodo 
 --baseline-sources Uniswap,Sushiswap,BalancerV2 
 --amount-to-estimate-prices-with 1000000000000000000 
 --private-key YOUR_PK
```
Get an order between USDC and DAI into the system.

Have added new structs into existing tests where appropriate, but am still thinking about a nice clean approach to testing these changes end to end (such a test would require running the solver and orderbook locally, placing an order involving a token that is known to be part of an existing stable pool - there are only three so far and they all have no liquidity the last I checked, intercepting the http solver request and verifying that the stable pool is passed along).

### DIFF
--- a/solver/src/liquidity.rs
+++ b/solver/src/liquidity.rs
@@ -20,7 +20,6 @@ use strum_macros::{AsStaticStr, EnumVariantNames};
 /// Defines the different types of liquidity our solvers support
 #[derive(Clone, AsStaticStr, EnumVariantNames, Debug)]
 pub enum Liquidity {
-    Limit(LimitOrder),
     ConstantProduct(ConstantProductOrder),
     Balancer(BalancerOrder),
 }
--- a/solver/src/liquidity.rs
+++ b/solver/src/liquidity.rs
@@ -4,13 +4,13 @@ pub mod slippage;
 pub mod uniswap;
 
 use crate::settlement::SettlementEncoder;
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 #[cfg(test)]
 use model::order::Order;
 use model::{order::OrderKind, TokenPair};
 use num::{rational::Ratio, BigRational};
 use primitive_types::{H160, U256};
-use shared::sources::balancer::pool_fetching::WeightedTokenState;
+use shared::sources::balancer::pool_fetching::{TokenState, WeightedTokenState};
 #[cfg(test)]
 use shared::sources::uniswap::pool_fetching::Pool;
 use std::collections::HashMap;
@@ -22,7 +22,7 @@ use strum_macros::{AsStaticStr, EnumVariantNames};
 pub enum Liquidity {
     Limit(LimitOrder),
     ConstantProduct(ConstantProductOrder),
-    WeightedProduct(WeightedProductOrder),
+    Balancer(BalancerOrder),
 }
 
 /// A trait associating some liquidity model to how it is executed and encoded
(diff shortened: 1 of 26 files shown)


marcovc (approved):
Looks good, thanks!

solver/src/liquidity.rs
```diff
@@ -198,6 +209,50 @@ impl Settleable for WeightedProductOrder {
...
+    fn settlement_handling(&self) -> &dyn SettlementHandling<Self> {
+        &*self.settlement_handling
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum BalancerOrder {
+    Weighted(WeightedProductOrder),
+    Stable(StablePoolOrder),
+}
```
vkgnosis, line 224: Should we make these separate values in the Liquidity enum directly? I feel most code has to treat them differently so we'll end up matching first on liquidity and then again on BalancerOrder.

solver/src/solver/http_solver/settlement.rs
```diff
@@ -128,12 +134,14 @@ fn match_prepared_and_settled_orders(
 fn match_prepared_and_settled_amms(
     mut prepared_constant_product_orders: HashMap<usize, ConstantProductOrder>,
     mut prepared_weighted_product_orders: HashMap<usize, WeightedProductOrder>,
+    mut prepared_stable_pool_orders: HashMap<usize, StablePoolOrder>,
```
vkgnosis, line 137: This is starting to feel hairy with the shifts. Not sure what a better solution is yet and doesn't have to be in this PR but feels like a refactor target.

vkgnosis (approved):

solver/src/solver/http_solver/settlement.rs
bh2smith: I agree, these shifts are 🤮

solver/src/liquidity.rs
bh2smith: Yea, we could do that. I just found this cleaner for other purposes here. I believe this change will be part of the following refactor discussed here in https://github.com/gnosis/gp-v2-services/pull/998#issuecomment-912253237

solver/src/solver/http_solver.rs
```diff
@@ -334,12 +368,15 @@ impl HttpSolver {
         let limit_orders = self.map_orders_for_solver(orders);
         let constant_product_orders = self.map_amm_orders_for_solver(constant_product_liquidity);
         let weighted_product_orders = self.map_amm_orders_for_solver(weighted_product_liquidity);
+        let stable_pool_orders = self.map_amm_orders_for_solver(stable_pool_liquidity);
+
```
bh2smith, line 372: This was tested locally by putting a print statement here:

```
println!("Stable Pool Orders {:?}", stable_pool_orders.len());
```
and placing an order of `USDC <-> DAI` which is a known [Balancer stable pool.](https://etherscan.io/address/0x06df3b2bbb68adc8b0e302443692037ed9f91b42)

bh2smith:
Hopefully, now that the local test has been performed and verified to have Stable pools at the time of http solver requests, we can merge this before the weekend and have it run over the weekend.

nlordell (approved):

solver/src/liquidity.rs
nlordell: > Should we make these separate values in the Liquidity enum directly?

I think so as well.

PR #995, opened 2021-08-19
Allow buy token to be pain in internal Vault balances

nlordell:
This PR enables an order's `buyTokenBalance` field to be `internal` (i.e. Balancer V2 Vault internal balance). While I don't suspect it will really be used anytime soon it:
- Removes a code branch! (passively increase coverage :heavy_check_mark:)
- There is no technical reason to not allow it (besides the additional gas cost, however this is an larger general issue we face and IMO is a bigger problem with some of the tokens we allow).

### Test Plan

Modified E2E test to pay out in internal balances.


nlordell:
Closing as there seems to be an issue with allowances to the Vault contract.

PR #996, opened 2021-08-19
Presignature api endpoint

fedgiac:
Adds an endpoint that return all valid presignatures for a user or order uid. There isn't much in terms of logic except for the SQL query, however many unit tests have been added, which make this PR quite large. Most of the new non-test code is similar to the equivalent endpoint for `trades`. 

Orders are not supposed to be created until a valid presignature is included in the database. The resulto of the `presignature` query should help the frontend detect when a new order can be created.

Note that the index change in #982 is related to the query in this PR, so you may also want to review the [one-line change](https://github.com/gnosis/gp-v2-services/pull/982/commits/a6aa072dab5dc17ba714ec7bf36e926d1fcf59d1) there.

### Test Plan

New unit tests.

The endpoint can also be tested on Rinkeby, at the moment the only address with presignatures is `0xcB1FF0c484a2267E39e19a82A8c313E633C106C6`.


model/src/presignature.rs
```diff
@@ -0,0 +1,45 @@
...
+    use super::*;
+    use serde_json::json;
+
+    #[test]
+    fn deserialization_and_back() {
+        let value = json!(
+        {
+            "blockNumber": 1337u64,
+            "logIndex": 42u64,
+            "owner": "0x0000000000000000000000000000000000000042",
```
vkgnosis, line 28: Offtopic: H160 serializes like this? Then why did we need H160Hexadecmial? Did this change recently?

orderbook/src/api.rs
```diff
@@ -130,6 +135,11 @@ pub fn convert_get_trades_error_to_reply(err: anyhowError) -> WithStatus<Json> {
     with_status(internal_error(), StatusCode::INTERNAL_SERVER_ERROR)
 }
 
+pub fn convert_get_presignatures_error_to_reply(err: anyhowError) -> WithStatus<Json> {
+    tracing::error!(?err, "get_presignatures error");
+    with_status(internal_error(), StatusCode::INTERNAL_SERVER_ERROR)
+}
```
vkgnosis, line 141: Why in this file and not get_presignatures.rs ?

orderbook/src/database.rs
```diff
@@ -62,6 +66,19 @@ impl Postgres {
     }
 }
 
+pub trait DataRetrieving: PreSignatureRetrieving + TradeRetrieving {
+    fn as_presignature_retrieving(self: Arc<Self>) -> Arc<dyn PreSignatureRetrieving>;
+    fn as_trade_retrieving(self: Arc<Self>) -> Arc<dyn TradeRetrieving>;
+}
```
vkgnosis, line 72: Alternatively you can make handle_all_routes take `Arc<dyn TradeRetrieving>` and another `Arc<dyn PreSignatureRetrieving>` . It would just be a clone of the same arc of course but that way no intermediate trait is needed.

orderbook/src/database/presignatures.rs
```diff
@@ -0,0 +1,375 @@
...
+use model::{order::OrderUid, presignature::PreSignature};
+use std::convert::TryInto;
+
+#[async_trait::async_trait]
+pub trait PreSignatureRetrieving: Send + Sync {
+    fn presignatures<'a>(
+        &'a self,
+        filter: &'a PreSignatureFilter,
+    ) -> BoxStream<'a, Result<PreSignature>>;
+}
```
vkgnosis, line 18: I just merged in another PR a change that makes the orders and trades return Vec directly instead of Stream. We should do the same here and then it can have a metric in Instrumented too.

orderbook/src/database/presignatures.rs
```diff
@@ -0,0 +1,375 @@
...
+                t.owner, \
+                t.order_uid, \
+                t.signed \
+            FROM presignatures t \
+            WHERE \
+                ($1 IS NULL OR t.owner = $1) \
+            AND \
+                ($2 IS NULL OR t.order_uid = $2) \
+            ORDER BY \
+                t.order_uid, t.block_number DESC, t.log_index DESC;";
```
vkgnosis, line 45: Is this needed? I don't think it is a problem just wondering why. Or just to make it deterministic? On that note maybe we should not allow getting *all* pre signatures without any filter in the http api handler just like for orders and trades because it would be very expensive for the database.

model/src/presignature.rs
```diff
@@ -0,0 +1,45 @@
...
+
+#[derive(Eq, PartialEq, Clone, Debug, Default, Deserialize, Serialize, Hash)]
+#[serde(rename_all = "camelCase")]
+pub struct PreSignature {
+    pub block_number: u64,
+    pub log_index: u64,
+    pub owner: H160,
+    pub order_uid: OrderUid,
+    pub signed: bool,
+}
```
bh2smith, line 15: Maybe I don't understand the purpose of Presignature, but it is my impression that this will be used for smart contract wallets such as a multisig. With this in mind, I was under the impression that we would want to collect multiple signatures. Does this mean that this table is a many-one relation ship with the orders table? Or does each signature here correspond to a unique order in the orders table that can be combined together to make a single singed order?

bh2smith (approved):
Very cool. Seems like some (if not all) of @vkgnosis's comments could be implemented.

orderbook/src/database/presignatures.rs
fedgiac: This is important and goes together with `DISTINCT ON` at the beginning: given some uid `uid`, I only want to read the most recent entry that contains `uid`, as the latest presignature for `uid` invalidates all past ones. This explains `ORDER BY t.block_number DESC, t.log_index DESC`, the reason why there is also `t.order_uid` is that the query appears to be invalid if the `DISTINCT ON` parameter is not in the sort values. 

> we should not allow getting all pre signatures without any filter in the http api handler

This PR should already require the query to have a non-empty filter if called from the API. I left the option of retrieving everything from an internal call to this function available, should I change this?

model/src/presignature.rs
fedgiac: It is many to one (for every order there might be multiple "signatures"; however for every signature there is only one order). In practice the database query in `presignatures()` is built so that also for every order (uid) there is at most one "real" signature for an order (older "signatures" in the table are ignored). So if this table is only read using the function introduced in this PR, then the relation is effectively one to one.

model/src/presignature.rs
fedgiac: This was already the case since the first of April, [here](https://github.com/gnosis/gp-v2-services/blob/main/model/src/trade.rs) is the code on which this section is based.

orderbook/src/api.rs
fedgiac: For consistency with two functions above, `convert_get_orders_error_to_reply` and `convert_get_trades_error_to_reply`. I thought it was done to keep API-generic details like the tracing error and the status code in the same place, but I'd be ok to move all of them to their respective files.

vkgnosis: Ah I forgot about the other functions. In that case I don't mind either way.

orderbook/src/database/presignatures.rs
vkgnosis: You're right. You shouldn't change it. Imo should only be checked in the api and not in the database trait, like it already is.

model/src/presignature.rs
bh2smith: In that case my question would be that I don't see any signature being stored here. I can only guess then that when new signatures come in that the order is updated with the accumulated signatures so that, once the threshold is met, the order eventually becomes executable? 

If that is the case, do we have some threshold verification logic coming in later that will classify these orders as executable?

fedgiac: I think the name `presignatures` is confusing for the table, I'll rename it to `presignature_events` to make it clear that no actual signature is stored there.
There is no "threshold" to speak of: an order is signed if the user has sent a `setPreSignature` call to the contract with `true` and the order uid as the input values. Any future call with the same `orderUid` overwrites the previous one. Validity of the order will have to be rechecked every time that the order is used again, as a previously valid order might have become invalid. All this logic will come later.

orderbook/src/database.rs
josojo: I like the current method better, though I have to admit it looks a little bit strange. 

I like it better, as soon we might get other retrievers, like AppDataRetriever, etc, and cloning each time the Arc results in functions with a lot of different input parameters, all intended to talk to the same database.

But I don't have a strong opinion

josojo (approved):

orderbook/src/database.rs
fedgiac: I discussed this with Valentin and the plan would be to use a concrete type for `database` (the instrumented database instead of `Arc<dyn DataRetrieving>`). I tried to do this but then I realized that the e2e tests use the original database, so to make this work I'd need some `NoopMetrics` for the orderbook metrics, which do not exist yet and would be too much code for this PR.
The plan now is to keep the PR as it is and in a follow-up PR introduce `NoopMetrics` and use a concrete type, which allows us to eventually remove this extra trait without extra cost.

orderbook/openapi.yml
```diff
@@ -178,6 +178,31 @@ paths:
                 $ref: "#/components/schemas/LegacyFeeInformation"
         404:
           description: sellToken non-existent
+  /api/v1/presignatures:
+    get:
+      summary: Get available Presignatures.
```
fleupold, line 183: nit: Maybe add a bit more context on what a presignature is?

orderbook/openapi.yml
```diff
@@ -629,6 +654,33 @@ components:
           description: "Hash of the corresponding settlement transaction containing the trade (if available)."
           $ref: "#/components/schemas/TransactionHash"
           nullable: true
+      required:
+        - blockNumber
+        - logIndex
+        - orderUid
+        - owner
+        - signed
```
fleupold, line 662: naming nit: is "signed" the correct status - would maybe active/valid/tradable be better? I feel like once the order is used it's still signed.

fleupold (commented):
Feel free to ignore my naming nits.

> Orders are not supposed to be created until a valid presignature is included in the database. 

I understand that this is easier for spam protection, but has this been discussed with @anxolin and @biocom? I feel like the user experience is quite limited if we can only send the order information after the transaction is mined (the user might e.g. close the tab or safe app once the tx has been signed but before its mined which).

orderbook/openapi.yml
```diff
@@ -178,6 +178,31 @@ paths:
                 $ref: "#/components/schemas/LegacyFeeInformation"
         404:
           description: sellToken non-existent
+  /api/v1/presignatures:
```
nlordell, line 181: Maybe from an API standpoint, it doesn't makes sense to return `signed: false` signatures and just return the current "pre-signature status" for orders/users? Specifically, if the purpose of the query is to:

> The resulto of the presignature query should help the frontend detect when a new order can be created.

Then maybe an endpoint `api/v1/orders/{orderUid}/presigned` that returns either a `true` if there is a valid presignature (i.e. the presignature event for the specified order with the highest `blockNumber` has a value of `signed: true`) and a `false` in case the order UID is not presigned (i.e. there is either no presignature event or the latest event has a value of `signed: false`). Alternatively, it could just return use HTTP status codes `204` and `401` respectively instead of returning data, but this may be slightly abusing HTTP status codes :stuck_out_tongue:.

fedgiac: I see. The reason it's built like this is because then one can also see all presignatures for a user. I see however how the usefulness of this might be quite limited: the only use case I can think of is a user who wants to check all presigned orders in order to invalidate them selectively, but in this case it might be reasonable to expect that the user knows the orderUid to be invalidated..

nlordell: > The reason it's built like this is because then one can also see all presignatures for a user

Right, that is a valid usecase. I would still argue that that endpoint `api/v2/presigatures/{user}` could just return a `Vec<OrderUid>` for valid presigatures (i.e. where pre signatures for the order UID with the highest block number has a `true` value for `signed`).

Specifically. Lets say a user does:
```
gpv2_settlement.pre_sign(order_uid_a, true);
// order A expired
gpv2_settlement.pre_sign(order_uid_b, true);
gpv2_settlement.pre_sign(order_uid_b, false);
gpv2_settlement.pre_sign(order_uid_b, false);
gpv2_settlement.pre_sign(order_uid_b, false);
gpv2_settlement.pre_sign(order_uid_b, false);
gpv2_settlement.pre_sign(order_uid_c, true);
gpv2_settlement.pre_sign(order_uid_b, false);
gpv2_settlement.pre_sign(order_uid_c, true);
gpv2_settlement.pre_sign(order_uid_c, true);
```

Then I would argue that:
```
$ curl $BASE_URL/api/v1/presignatures/0x...
[
  "$order_uid_c",
]
```

Would be returned because:
- Order A expired (so the presignature is no longer valid) - although if this were included, it wouldn't be the end of the world. as this would required knowing what the timestamp on the "latest" block is.
- Order B was "unsigned".
- Order C was "unsigned" but then signed again.

anxolin:
Hi, thanks for the heads up! what about looking at this tomorrow? 

i would love to understand better the possibilities, not sure I get it 100%. 
My assumption, is you want to receive the details of the order, and ideally it would be already be presigned. So we avoid users posting unsigned orders for other users

Initially, my thinking is, if I understood this right, we should be able to send the tx to Ethereum, and do the call to the API to post a new order in any order. Then, the backend would just "assume" that in a few minutes the presign will be done. 

I don't see people spamming this, but if this is really a concern I guess we can review options.

orderbook/openapi.yml
fedgiac: I agree with B and C but I think A should also be returned nevertheless. If I call `presignatures`, I would not expect that some extra information on the order is used to change what `presigned` means. This is particularly relevant since if we don't have the order associated to the presignature uid then we won't be able to tell if the order is expired, and so it'll appear in the output regardless. So adding a new order (close to the expiration date) would mean that a presigned orders would not be presigned anymore.
The information on validity (validTo + presigned) can be captured by the `order` path.

orderbook/openapi.yml
fedgiac: I used "signed" since this is the name of the variable in the SC event. I agree it's not good naming for the API itself, but after implementing [Nick's comment](https://github.com/gnosis/gp-v2-services/pull/996#discussion_r693706919) I think the remaining uses of `signed` would be appropriate.

PR #997, opened 2021-08-20
Balancer Stable Pools part 1: contract artifacts

bh2smith:
First logical component of #733.
Adding Stable Pool contract artifacts. Some noteworthy points.

1. Balancer ABI sources are no longer available in a tagged repo (cf this slack discussion with Balancer team)
2. Vault and Weighted pool factory are now fetched form a different source and contain differences (iirc - the diff is that one has bytecode and one doesn't). This needs to be fixed separately.

### Test Plan
Added these artifacts to existing tests


contracts/src/bin/vendor.rs
```diff
@@ -33,19 +33,21 @@ fn run() -> Result<()> {
...
+            "balancer-labs/balancer-subgraph-v2/master/abis/WeightedPoolFactory.json",
         )?
         .github(
             "BalancerV2WeightedPool2TokensFactory",
             "balancer-labs/balancer-v2-monorepo/a3b570a2aa655d4c4941a67e3db6a06fbd72ef09/\
              pkg/deployments/deployed/mainnet/WeightedPool2TokensFactory.json",
         )?
+        .github(
+            "BalancerV2StablePoolFactory",
+            "balancer-labs/balancer-subgraph-v2/master/abis/StablePoolFactory.json",
```
nlordell, line 49: Can we use a tag or commit hash? While they have re-written commit history, I still think using a specific commit is important and more reproducible in general.

nlordell:
Just FYI, the bytecode appears to be available:
- [`WeightedPool(Factory)`](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/deployments/tasks/20210418-weighted-pool/build-info/WeightedPoolFactory.json)
- [`Vault`](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/deployments/tasks/20210418-vault/build-info/Vault.json)

contracts/src/bin/vendor.rs
bh2smith: I thought that I had pointed out somewhere that this repo does not have any tags or releases so that this wouldn't be possible. Perhaps the link to this discussion got lost when creating the new PR I will find it.

bh2smith:
> Just FYI, the bytecode appears to be available here:
> 
>     * [`WeightedPool(Factory)`](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/deployments/tasks/20210418-weighted-pool/build-info/WeightedPoolFactory.json)
> 
>     * [`Vault`](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/deployments/tasks/20210418-vault/build-info/Vault.json)

If this is the case, then this has been recently added as a result of my request.

contracts/src/bin/vendor.rs
nlordell: If the repo doesn't have tags, you can always use commit hashes:
https://raw.githubusercontent.com/balancer-labs/balancer-subgraph-v2/399c931706b682d22a4fa7f5f9326347c5cafc9f/abis/StablePool.json

nlordell:
> Are you sure the bytecode is there?

Yes:
```
curl -s https://raw.githubusercontent.com/balancer-labs/balancer-v2-monorepo/master/pkg/deployments/tasks/20210418-vault/build-info/Vault.json | jq '.output.contracts["contracts/vault/Vault.sol"].Vault.evm.bytecode.object'
```

> then this has been recently added as a result of my request

Looks like its been around for 2 months now (although stable pools is more recent, as it was deployed more recently): https://github.com/balancer-labs/balancer-v2-monorepo/tree/e29c13a6236bc7ea9c38f73bbb23920ae391f9eb/pkg/deployments/tasks

> are you linking me to a master branch and not a specific commit hash here?

I left figuring out the commit hash as an exercise to the PR author as it should be fairly trivial to do.

bh2smith:
> Looks like its been around for 2 months now

The way this project organizes its releases and overwrites history is incredibly unconventional. They have tagged a bunch of releases with "weighted-deployment" and all these other things that give you the impression they have what we are looking for and all the abi jsons aren't there. So yea, maybe this stuff has been on master for two months now, but not a sinlge one of those tags even has the pkg directory. 

Anyway, I hope you can understand this contract business has been like shooting at a moving target.

> I left figuring out the commit hash as an exercise to the PR author.

It is non-trivial, I've noticed.

nlordell:
> It is non-trivial, I've noticed.

It is trivial:
```
$ git clone https://github.com/balancer-labs/balancer-v2-monorepo.git
Cloning into 'balancer-v2-monorepo'...
remote: Enumerating objects: 14688, done.
remote: Counting objects: 100% (4368/4368), done.
remote: Compressing objects: 100% (1634/1634), done.
remote: Total 14688 (delta 3012), reused 3821 (delta 2683), pack-reused 10320
Receiving objects: 100% (14688/14688), 16.02 MiB | 3.76 MiB/s, done.
Resolving deltas: 100% (10364/10364), done.
$ cd balancer-v2-monorepo 
$ git log -1 --pretty='format:%H'
d430cdbb15468bc6f4c778f6268d98f36648ebf4
```

Alternatively with the UI:

image
image

nlordell:
To be clear, the requested change here is to include commit hashes in the github links for the stable pool ABIs.

Modifying the vendoring scripts to use a different file in the Balancer repo (specifically the `build-info` files) can be done separately and at a later time.

bh2smith:
> To be clear, the requested change here is to include commit hashes in the github links for the stable pool ABIs.
> 
> Modifying the vendoring scripts to use a different file in the Balancer repo (specifically the `build-info` files) can be done separately and at a later time. It was just an interesting finding that I thought we could use to fix the scripts given the repository history was re-written making the files we originally used no longer available.

I have already started looking into vendoring bytecode from etherscan, but I suppose this could be alright too... still doesn't really help when they rug-pull the commit history and make tags with completely different directory structure.

nlordell:
> they rug-pull the commit history and make tags with completely different directory structure.

I would argue that this is why we vendor these files instead of downloading them at build time, otherwise the rug pull would have broken our builds.

Once these links truly become unavailable, we can always point them to: https://raw.githubusercontent.com/gnosis/gp-v2-services/COMMIT_HASH/contracts/artifacts/BalancerV2Vault.json

bh2smith:
Fixed - cc @nlordell

contracts/src/bin/vendor.rs
```diff
@@ -46,6 +46,10 @@ fn run() -> Result<()> {
             "balancer-labs/balancer-v2-monorepo/a3b570a2aa655d4c4941a67e3db6a06fbd72ef09/\
              pkg/deployments/deployed/mainnet/WeightedPool2TokensFactory.json",
         )?
+        .github(
+            "BalancerV2StablePoolFactory",
+            "balancer-labs/balancer-subgraph-v2/2b97edd5e65aed06718ce64a69111ccdabccf048/abis/StablePoolFactory.json",
+        )?
```
nlordell, line 52: AFAIK, this is `abi_only` right? THere doesn't seem to be any bytecode in the vendored file.

bh2smith: I guess not... in that case I don't really know where to get it from. This has become quite a nuisance. Should I just find the bytecode and paste it in?

nlordell: We don't make use of the bytecode anywhere AFAIK, why not just include it in the `abi_only()` vendored contracts (like the `BalancerV2StablePoolFactory`)? My point is that right now its included as a `full()` vendored contract but doesn't need to be.

bh2smith: Alright I found it.

https://github.com/balancer-labs/balancer-v2-monorepo/blob/stable-deployment/pkg/deployments/tasks/20210624-stable-pool/abi/StablePoolFactory.json

bh2smith: But we will use it later... why not just include it now.

vkgnosis (approved):

nlordell (approved):

PR #998, opened 2021-08-20
Balancer Stable Pools part 6: Pool Fetching

bh2smith:
Following #1008 and part of #733.

Fetching both Stable and Weighted Balancer Pools.

### Test Plan
TBD

### DIFF
--- a/shared/src/sources/balancer/pool_fetching.rs
+++ b/shared/src/sources/balancer/pool_fetching.rs
@@ -165,15 +165,20 @@ impl WeightedPool {
         // We expect the weight and token indices are aligned with balances returned from EVM query.
         // If necessary we would also pass the tokens along with the query result,
         // use them and fetch the weights from the registry by token address.
-        for (i, balance) in balances.into_iter().enumerate() {
+        for (&token, balance, &scaling_exponent, &weight) in itertools::izip!(
+            &pool_data.common.tokens,
+            balances,
+            &pool_data.common.scaling_exponents,
+            &pool_data.normalized_weights
+        ) {
             reserves.insert(
-                pool_data.common.tokens[i],
+                token,
                 WeightedTokenState {
                     token_state: TokenState {
                         balance,
-                        scaling_exponent: pool_data.common.scaling_exponents[i],
+                        scaling_exponent,
                     },
-                    weight: pool_data.normalized_weights[i],
+                    weight,
                 },
             );
         }
@@ -209,12 +214,16 @@ impl StablePool {
         // We expect the weight and token indices are aligned with balances returned from EVM query.
         // If necessary we would also pass the tokens along with the query result,
         // use them and fetch the weights from the registry by token address.
-        for (i, balance) in balances.into_iter().enumerate() {
+        for (&token, balance, &scaling_exponent) in itertools::izip!(
+            &pool_data.common.tokens,
+            balances,
+            &pool_data.common.scaling_exponents,
+        ) {
             reserves.insert(
-                pool_data.tokens()[i],
+                token,
                 TokenState {
                     balance,
-                    scaling_exponent: pool_data.scaling_exponents()[i],
+                    scaling_exponent,
                 },
             );
         }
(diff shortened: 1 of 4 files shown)


nlordell:
While splitting out #997 was a good start, I still think this PR is quite big and hard to review.

> It is important to note that the only relevant logical changes made here

While I agree, I also think that this PR contains more logic changes than need to be included in a single PR and can be further divided into smaller, easier to review changes. Additionally, while the test plan provides a manual test to verify that at least event indexing is working, there are other logic changes in this PR that don't seem like they were verified. To give a concrete example of this, I think that, for example, the changes to `graph_api.rs` could be included in a smaller self-contained PR, where some new tests checking that the data returned from the Subgraph GraphQL query can be parsed (like the `decode_pools_data` test but for stable pools) as well as the new `PoolData<*> -> RegsiterdPool` conversion (especially since they can now panic). While such a PR would not be "hooked up" and used right away in the driver, it would still be merge-able and easier to review.

vkgnosis (approved):

shared/src/sources/balancer/pool_storage.rs
```diff
@@ -81,6 +83,14 @@ impl PoolEvaluating for RegisteredWeightedPool {
     fn tokens(&self) -> Vec<H160> {
         self.common.tokens.clone()
     }
+
```
nlordell, line 86: One suggestion to make implementing this and require fewer methods:
```rust
trait PoolEvaluating {
   fn properties(&self) -> &CommonPoolInfo;
}
```

This should have the token list, address and scaling exponents right?

nlordell (approved):
Logic looks good.

I still think there is quite a bit of duplicate code here that would be nice to get rid of using the `PoolEvaluating` trait. Ideally, with a refactor it would be possible to add a new pool type just by:
- Specifying the factory contract and implement the required trait with `impl_event_retrieving!` macro
- Specifying the logic to fetch the pool data on-chain (`PoolInfoFetching` could be repurposed for this)
- Create the Subgraph query

And have everything just work with the magic of generics. As you mentioned previously in this comment (https://github.com/gnosis/gp-v2-services/pull/1007#issuecomment-911478421) this was something you would do in a follow up PR, just wanted to share some initial thoughts I had about it. Approving based on the assumption of cleaning up some of the duplicated code in the `shared::sources::balancer`.

shared/src/sources/balancer/pool_storage.rs
bh2smith: Yes this is very nice. However, it is also important to note that Registered Pool (as it was before) is not quite the same as Balancer Pool that we are seeing here (while they are enums of various pool types, the PoolEvaluating will not cover both cases. We would need to rename `PoolEvaluating` to `RegisteredPoolEvaluating` and make yet another trait for doing the analogous thing for both Weighted and Stable Pools. The expected difference between these two traits is that one would have `CommonPoolInfo`(id, address, tokens, scaling_exponents) properties and the other would have (id, paused_state, swap_fee, reserves). 

Not that it should be a problem to do this, I just wanted to point this out in case you were thinking we would be able to reuse a lot.

shared/src/sources/balancer/pool_fetching.rs
```diff
@@ -81,14 +187,56 @@ impl WeightedPool {
...
+        // use them and fetch the weights from the registry by token address.
+        for (i, balance) in balances.into_iter().enumerate() {
+            reserves.insert(
+                pool_data.tokens()[i],
+                TokenState {
+                    balance,
+                    scaling_exponent: pool_data.scaling_exponents()[i],
+                },
+            );
+        }
```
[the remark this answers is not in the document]
bh2smith: This is a cool suggestion and I will surely do this. Just one question - how did you know which things to make references and which not in you zipper?

bh2smith:
> Approving based on the assumption of cleaning up some of the duplicated code in the `shared::sources::balancer`.

Is this assumption that the cleanup will be made here in this PR, or immediately after this is merged? If I perform the cleanup here, I am a bit worried it might wind up to be a larger diff than we were hoping for, but it might also be worth not ever including this duplication in the codebase at all. Who knows, maybe de-duplication will make this PR smaller...

nlordell:
> immediately after this is merged?

Follow up to stable pool support in the driver (so after #992).

PR #999, opened 2021-08-20
Reorganize in preparation for new signature type

fedgiac:
Separates the ecdsa signature type from the signature type defined in the contract. The contract signature type will later be extended to include other signature types.

Since most of the changes are only structural changes, it might be easier to review each commit in this PR separately.

Note that order cancellations will only support ecdsa signatures.

### Test Plan
Existing unit tests


bh2smith (approved):
Such Cryptography! Many Signature

image

model/src/signature.rs
```diff
@@ -0,0 +1,208 @@
...
+};
+
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Deserialize, Serialize, Hash)]
+#[serde(rename_all = "lowercase")]
+pub enum SigningScheme {
+    Eip712,
+    EthSign,
+}
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Default, Deserialize, Serialize, Hash)]
+pub struct Signature(EcdsaSignature);
```
nlordell, line 17: I assume this will become an `enum` or something once pre-sign gets introduced?

nlordell, line 17: Instead of separating these two, could this maybe be:
```
pub enum SigningScheme {
    Eip712(EcdsaSignature),
    EthSign(EcdsaSignature),
    // and eventually: PreSign(Address),
}
```

I'm not sure exactly how this would work with `serde` though...

model/src/signature.rs
```diff
@@ -0,0 +1,208 @@
...
+        D: serde::Deserializer<'de>,
+    {
+        struct Visitor {}
+        impl<'de> de::Visitor<'de> for Visitor {
+            type Value = EcdsaSignature;
+
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                write!(
+                    formatter,
+                    "the 65 ecdsa signature bytes as a hex encoded string"
```
nlordell, line 181: ubernit: maybe include the expected order of the bytes here (`r .. s .. v`) and that V is in Electrum notation (i.e. 27 or 28 and not 0 or 1).

nlordell (approved):

model/src/signature.rs
nlordell: **edit**: This [seems to be possible](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c8f1f22ba81233ede21b260f584426c9).

<details><summary>Code</summary>

```rust
extern crate serde; // 1.0.127
extern crate serde_derive; // 1.0.127
extern crate serde_json; // 1.0.66

use serde::{de, Deserialize, Deserializer};
use std::borrow::Cow;

#[derive(Debug, Deserialize)]
struct Order {
    #[serde(rename = "someProperty")]
    some_property: String,
    #[serde(flatten)]
    signature: Signature,
}

#[derive(Debug, Deserialize)]
#[serde(tag = "signingScheme", content = "signature")]
enum Signature {
    #[serde(rename = "eip712")]
    Eip712(EcdsaSignature),
    #[serde(rename = "ethsign")]
    EthSign(EcdsaSignature),
}

#[derive(Debug, Default)]
struct EcdsaSignature {
    v: u8,
    r: [u8; 32],
    s: [u8; 32],
}

impl<'de> Deserialize<'de> for EcdsaSignature {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        let value = Cow::<str>::deserialize(deserializer)?;
        let hex = value
            .strip_prefix("0x")
            .ok_or_else(|| de::Error::custom("missing '0x' prefix"))?;

        let mut signature = Self::default();

        // poor-man's hex::decode_to_slice...
        if hex.len() != 65 * 2 {
            return Err(de::Error::custom("not 65 bytes long"));
        }
        for i in 0..32 {
            signature.r[i] =
                u8::from_str_radix(&hex[i * 2..][..2], 16).map_err(de::Error::custom)?;
        }
        for i in 0..32 {
            signature.s[i] =
                u8::from_str_radix(&hex[(32 + i) * 2..][..2], 16).map_err(de::Error::custom)?;
        }
        signature.v = u8::from_str_radix(&hex[64 * 2..], 16).map_err(de::Error::custom)?;

        Ok(signature)
    }
}

fn main() {
    let _ = dbg!(serde_json::from_str::<Order>(
        r#"{
            "someProperty": "hello",
            "signingScheme": "ethsign",
            "signature": "0x111111111111111111111111111111111111111111111111111111111111111122222222222222222222222222222222222222222222222222222222222222221b"
        }"#
    ));
}
```

</details>

fedgiac: Yep!

fedgiac: I'm not completely convinced by this approach. While I think we could make it work, I think it's handy to specify the signing scheme without any signature attached to it, for example it allows to do things like:
```
OrderBuilder::default()
        .with_signing_scheme(SigningScheme::EthSign)
        .sign_with(
            &gpv2.domain_separator,
            SecretKeyRef::from(&SecretKey::from_slice(&TRADER_B_PK).unwrap()),
        )
```

In any case, eventually the new `Signature` will become exactly the enum you suggest for `SigningScheme`, so it might be that the number of instances required for `SigningScheme` will significantly decrease.

nlordell: I would imagine that this would just become:

```rust
OrderBuilder::default()
        .sign_with(
            &gpv2.domain_separator,
            SecretKeyRef::from(&SecretKey::from_slice(&TRADER_B_PK).unwrap()),
            SigningScheme::EthSign,
        )
```

Otherwise, you would be able to create an order:
```
Order {
    signing_scheme: SigningScheme::EthSign,
    signature: Signature::EthSign(my_address),
    ..order_params
}
```

Either way, I don't feel strongly about this suggestion so feel free to ignore it (hence already approved the PR :smile:). Was just kind of "thinking out loud".

nlordell: **edit**: I meant for the enum in my suggestion to be called `Signature` and not `SigningScheme`:
```rust
pub enum Signature {
    Eip712(EcdsaSignature),
    EthSign(EcdsaSignature),
    // and eventually: PreSign(Address),
}
```

fedgiac: Ah, then this is indeed how I hope `Signature` will look like! I thought you wanted to completely get rid of `SigningScheme`.

nlordell: > I thought you wanted to completely get rid of SigningScheme

I think it might still necessary, for example so you can do:

```rust
let signature: Signature = // ...
let scheme = signature.scheme();
```

But I still think `Order` can just have a `signature: Signature` field instead of `signing_scheme: SigningScheme, signature: SignatureData` fields.

PR #1000, opened 2021-08-23
Propagate solver address to solution sender

taminomara:
Part of #827.

Instead of using one global account for every settlement, we now use individual solver's account. In practce, the behaviour shouldn't change just yet, because solvers are initialized using the same global account that we used before.

### Test Plan
Existing CI tests. More unittests will be added in a separate PR.


solver/src/solver.rs
```diff
@@ -89,12 +92,12 @@ pub fn create(
...
     disabled_paraswap_dexs: Vec<String>,
     client: Client,
-) -> Result<Vec<Box<dyn Solver>>> {
+) -> Result<Solvers> {
     // Tiny helper function to help out with type inference. Otherwise, all
     // `Box::new(...)` expressions would have to be cast `as Box<dyn Solver>`.
     #[allow(clippy::unnecessary_wraps)]
-    fn boxed(solver: impl Solver + 'static) -> Result<Box<dyn Solver>> {
-        Ok(Box::new(solver))
+    fn boxed(solver: impl Solver + 'static) -> Result<Arc<dyn Solver>> {
```
nlordell, line 99: Tiny naming nit:

```suggestion
    fn shared(solver: impl Solver + 'static) -> Result<Arc<dyn Solver>> {
```

vkgnosis (approved):
I also like the cleanup of these SettlementWithSolver types in the driver.

nlordell (approved):

PR #1003, opened 2021-08-23
Balancer Stable Pool Part 2 - Fetch Stable Pools with graph api

bh2smith:
Following #997 and part of #733

We introduce `RegisteredStablePool` and enum `RegisteredPool` for the purpose of querying and parsing distinguished pool data from the graph. We refactor the get_weighted pool methods into `get_registered_pool` and test that all is good.

Note that the only logical changes introduces here are in `graph_api.rs`. However we have introduced some useful helper methods in `pool_storage.rs` that assist for converting `RegisteredPools` into their specialized parts (Stable/Weighted) (i.e. unwrapping the enum).

### Test Plan
Updated tests incorporating stable pools (including helper methods defined on new enum RegisteredPool) and also, a great indication that the graph api is working is to manually run the balancer_subgraph_query (ignored by CI) and compare with the total number of Balancer Pools on dune.

```sql
select count(*) from balancer_v2."WeightedPoolFactory_evt_PoolCreated"
select count(*) from balancer_v2."WeightedPool2TokensFactory_evt_PoolCreated"
select count(*) from balancer_v2."StablePoolFactory_evt_PoolCreated"
```


shared/src/sources/balancer/graph_api.rs
```diff
@@ -179,23 +247,59 @@ mod pools_query {
...
+                        pool.tokens.push(token.address);
+                        pool.normalized_weights.push(token.weight);
+                        pool.scaling_exponents
+                            .push(18u8.checked_sub(token.decimals).ok_or_else(|| {
+                                anyhow!("unsupported token with more than 18 decimals")
+                            })?);
+                        Ok(registered_pool)
+                    } else {
+                        unreachable!("Not a weighted pool");
+                    }
```
vkgnosis, line 269: Instead of creating the RegisteredPool enum immediately and then needing this unreachable branch we can do:
```rust
self.tokens
    .iter()
    .try_fold(
        RegisteredWeightedPool {
            pool_id: self.id,
            pool_address: self.address,
            tokens: Vec::with_capacity(token_count),
            normalized_weights: Vec::with_capacity(token_count),
            scaling_exponents: Vec::with_capacity(token_count),
            block_created: block_created_upper_bound,
        },
        |mut pool, token| {
            pool.tokens.push(token.address);
            pool.normalized_weights.push(token.weight);
            pool.scaling_exponents
                .push(18u8.checked_sub(token.decimals).ok_or_else(|| {
                    anyhow!("unsupported token with more than 18 decimals")
                })?);
            Ok(pool)
        },
    )
    .map(RegisteredPool::Weighted)
```

And the same change for stable pools below.

shared/src/sources/balancer/graph_api.rs
```diff
@@ -70,18 +75,57 @@ impl BalancerSubgraphClient {
...
+        for (factory, pool) in weighted_pools.chain(stable_pools) {
+            let pool = match pool {
+                Ok(pool) => pool,
+                Err(err) => {
+                    // Technically this should never happen and would only ever be from
+                    // a token with more than 18 decimals (not supported by balancer contracts)
+                    // https://github.com/balancer-labs/balancer-v2-monorepo/blob/deployments-latest/pkg/pool-utils/contracts/BasePool.sol#L476-L487
+                    // We, thus, log but skip this error.
+                    tracing::error!("Unexpected failure to build registered pool {}", err);
+                    continue;
```
nlordell, line 119: If this should never happen, then I would return an error here. I would argue that if this were to happen, we would want the pool initialization to fail, as it would mean that there is something incorrect about our assumptions of how the contracts/subgraph work

shared/src/sources/balancer/graph_api.rs
```diff
@@ -239,7 +344,7 @@ mod tests {
         use pools_query::*;
 
         assert_eq!(
-            serde_json::from_value::<Data>(json!({
+            serde_json::from_value::<Data<WeightedToken>>(json!({
```
nlordell, line 347: Can we have a similar test for `Data<StableToken>` with data taken from the subgraph?

shared/src/sources/balancer/graph_api.rs
vkgnosis: Ben and I discussed this in the previous bigger PR. I suggested we shouldn't fail the whole process. But I can see your argument too. Either way is fine with me.

vkgnosis (approved):

PR #1006, opened 2021-08-23
[Easy] Balancer Stable Pool part 3: Info Fetching

bh2smith:
Follows #1003 and part of #733

This simple change renames `get_pool_info` to `get_weighted_pool_info` to make room for its analogous `get_stable_pool_info` which will be used in a following PR. The `PoolInfo`struct is extended to include the `amplification_parameter` and  also has `weights`. We note that these are only relevant to specific pools as a comment and set the values to `nullish` when irrelevant.

Note that it may appear like the two methods have some overlap, but there is actually quite distinguished order of operations here when fetching info. 

### Test Plan
Could write an ignored test here, but would require and infura key. Happy to do so if recommended, but we didn't have one before.

### DIFF
--- a/shared/src/sources/balancer/info_fetching.rs
+++ b/shared/src/sources/balancer/info_fetching.rs
@@ -43,6 +43,7 @@ pub struct StablePoolInfo {
 pub struct PoolInfoFetcher {
     pub web3: Web3,
     pub token_info_fetcher: Arc<dyn TokenInfoFetching>,
+    pub vault: BalancerV2Vault,
 }
 
 #[automock]
@@ -58,12 +59,11 @@ impl PoolInfoFetching for PoolInfoFetcher {
     async fn get_weighted_pool_data(&self, pool_address: H160) -> Result<WeightedPoolInfo> {
         let mut batch = CallBatch::new(self.web3.transport());
         let pool_contract = BalancerV2WeightedPool::at(&self.web3, pool_address);
-        // Need vault and pool_id before we can fetch tokens.
-        let vault = BalancerV2Vault::deployed(&self.web3).await?;
         let pool_id = H256::from(pool_contract.methods().get_pool_id().call().await?.0);
 
         // token_data and weight calls can be batched
-        let token_data = vault
+        let token_data = self
+            .vault
             .methods()
             .get_pool_tokens(Bytes(pool_id.0))
             .batch_call(&mut batch);
@@ -107,12 +107,11 @@ impl PoolInfoFetching for PoolInfoFetcher {
     async fn get_stable_pool_data(&self, pool_address: H160) -> Result<StablePoolInfo> {
         let mut batch = CallBatch::new(self.web3.transport());
         let pool_contract = BalancerV2StablePool::at(&self.web3, pool_address);
-        // Need vault and pool_id before we can fetch tokens.
-        let vault = BalancerV2Vault::deployed(&self.web3).await?;
         let pool_id = H256::from(pool_contract.methods().get_pool_id().call().await?.0);
 
         // token_data and weight calls can be batched
-        let token_data = vault
+        let token_data = self
+            .vault
             .methods()
             .get_pool_tokens(Bytes(pool_id.0))
             .batch_call(&mut batch);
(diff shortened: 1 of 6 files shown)


vkgnosis (approved):
Why not make one struct with the shared fields and two separate structs with the common one plus extra info or an enum? Maybe it just works out better with the existing code or following PRs so still approving.

shared/src/sources/balancer/info_fetching.rs
```diff
@@ -5,17 +5,20 @@ use crate::{
...
-use contracts::{BalancerV2Vault, BalancerV2WeightedPool};
-use ethcontract::{batch::CallBatch, Bytes, H160, H256};
+use contracts::{BalancerV2StablePool, BalancerV2Vault, BalancerV2WeightedPool};
+use ethcontract::{batch::CallBatch, Bytes, H160, H256, U256};
 use mockall::*;
 use std::sync::Arc;
 
 #[derive(Clone)]
-pub struct WeightedPoolInfo {
+pub struct PoolInfo {
```
nlordell, line 14: I don't think this should be a single struct as the data is fundamentally different between the two kinds of pools. You can compose data if you want share some code:
```rust
pub struct CommonPoolInfo {
pub pool_id: H256,
    pub tokens: Vec<H160>,
    pub weights: Vec<Bfp>,
    pub scaling_exponents: Vec<u8>,
}

pub struct WeightedPoolInfo {
    common: CommonPoolInfo,
    weights: Vec<Bfp>,
}

pub struct StablePoolInfo {
    common: CommonPoolInfo,
    pub amplification_parameter: U256,
}
```

shared/src/sources/balancer/info_fetching.rs
```diff
@@ -82,11 +86,55 @@ impl PoolInfoFetching for PoolInfoFetcher {
...
+        Ok(PoolInfo {
             pool_id,
             tokens,
             weights,
             scaling_exponents,
+            amplification_parameter: Default::default(),
+        })
+    }
+
+    async fn get_stable_pool_data(&self, pool_address: H160) -> Result<PoolInfo> {
```
nlordell, line 98: Can we write a test for this using the new contract mocking feature introduced by @taminomara ?

bh2smith: Sure thing. it will take me a bit to get used to it.

bh2smith: I think that we are still going to run into the issue of having a mockable `CallBatch` and `Web3`for these methods... Is this something you were aware of when making this suggestion?

nlordell: Yes, with the latest `ethcontract` release, we should have mockable contracts and mock Web3 that should allow testing. Maybe @taminomara might be able to give you some pointers when you're back from vacation.

shared/src/sources/balancer/info_fetching.rs
```diff
@@ -83,10 +95,178 @@ impl PoolInfoFetching for PoolInfoFetcher {
...
+        let mock = Mock::new(1);
+        let web3 = mock.web3();
+
+        let vault = mock.deploy(BalancerV2Vault::raw_contract().abi.clone());
+        let stable_pool = mock.deploy(BalancerV2StablePool::raw_contract().abi.clone());
+
+        let pool_id = H256::from_low_u64_be(1);
+        let token = H160::from_low_u64_be(1);
+        stable_pool
+            .expect(BalancerV2StablePool::signatures().get_pool_id())
```
taminomara, line 170: Nit: prefer `expect_call` or `expect_transaction` to make sure the code uses `call` or `send` as intended.

P.S. maybe we need to rename `expect` to make it clearer? Something like `expect_call_or_transaction`?

shared/src/sources/balancer/info_fetching.rs
bh2smith: All tested and ready to rock!

shared/src/sources/balancer/info_fetching.rs
```diff
@@ -83,10 +95,310 @@ impl PoolInfoFetching for PoolInfoFetcher {
...
+            .map(|token| token_decimals.get(token).and_then(|t| t.decimals))
+            .collect::<Option<Vec<_>>>()
+            .ok_or_else(|| anyhow!("all token decimals required to build scaling factors"))?;
+        // Note that balancer does not support tokens with more than 18 decimals
+        // https://github.com/balancer-labs/balancer-v2-monorepo/blob/ce70f7663e0ac94b25ed60cb86faaa8199fd9e13/pkg/pool-utils/contracts/BasePool.sol#L497-L508
+        let scaling_exponents = ordered_decimals
+            .iter()
+            .map(|decimals| 18u8.checked_sub(*decimals))
+            .collect::<Option<Vec<_>>>()
+            .ok_or_else(|| anyhow!("token with more than 18 decimals"))?;
```
nlordell, line 138: nit: This code is duplicated in both `get_*_pool_data` methods and could be in its own method.

nlordell (approved):
Looks good, just a small nit around duplicated code.

PR #1007, opened 2021-08-23
Balancer Stable Pool part 4: Event Handling and Pool Registry

bh2smith:
Following #1006 and part of #733

We introduce the stable pool creation event listener and adapt the event handling and PoolRegistry to store stable pools along side weighted in the previously introduced enum `RegisteredPool`.

### Test Plan
TBD...

### DIFF
--- a/shared/src/sources/balancer/event_handler.rs
+++ b/shared/src/sources/balancer/event_handler.rs
@@ -19,15 +19,19 @@ use crate::{
     sources::balancer::{
         info_fetching::PoolInfoFetching,
         pool_init::PoolInitializing,
-        pool_storage::{PoolCreated, PoolStorage, RegisteredWeightedPool},
+        pool_storage::{
+            PoolCreated, PoolStorage, PoolType, RegisteredStablePool, RegisteredWeightedPool,
+        },
     },
     Web3,
 };
 use anyhow::{anyhow, Context, Result};
 use contracts::{
+    balancer_v2_stable_pool_factory::{self, Event as StablePoolFactoryEvent},
     balancer_v2_weighted_pool_2_tokens_factory::{self, Event as WeightedPool2TokensFactoryEvent},
     balancer_v2_weighted_pool_factory::{self, Event as WeightedPoolFactoryEvent},
-    BalancerV2WeightedPool2TokensFactory, BalancerV2WeightedPoolFactory,
+    BalancerV2StablePoolFactory, BalancerV2WeightedPool2TokensFactory,
+    BalancerV2WeightedPoolFactory,
 };
 use ethcontract::{Event as EthContractEvent, H256};
 use model::TokenPair;
@@ -44,6 +48,8 @@ pub struct BalancerPoolRegistry {
         Mutex<EventHandler<Web3, BalancerV2WeightedPoolFactoryContract, PoolStorage>>,
     two_token_pool_updater:
         Mutex<EventHandler<Web3, BalancerV2WeightedPool2TokensFactoryContract, PoolStorage>>,
+    stable_pool_updater:
+        Mutex<EventHandler<Web3, BalancerV2StablePoolFactoryContract, PoolStorage>>,
 }
 
 impl BalancerPoolRegistry {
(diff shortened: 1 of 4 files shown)


shared/src/sources/balancer/event_handler.rs
```diff
@@ -56,65 +62,119 @@ impl BalancerPoolRegistry {
...
+                .lock()
+                .await
+                .store
+                .pools_for(pool_ids)
+                .into_iter()
+                .map(|pool| {
+                    pool.try_into_weighted()
+                        .expect("weighted_pool_updater only contains weighted pools")
+                })
+                .collect::<Vec<_>>(),
```
vkgnosis, line 148: I think the intermediate vec is not necessary
```suggestion
```

shared/src/sources/balancer/event_handler.rs
```diff
@@ -56,65 +62,119 @@ impl BalancerPoolRegistry {
...
+                .lock()
+                .await
+                .store
+                .pools_for(pool_ids)
+                .into_iter()
+                .map(|pool| {
+                    pool.try_into_weighted()
+                        .expect("two_token_pool_updater only contains weighted pools")
+                })
+                .collect::<Vec<_>>(),
```
vkgnosis, line 161:
```suggestion
```

shared/src/sources/balancer/pool_storage.rs
```diff
@@ -92,6 +61,13 @@ pub struct RegisteredStablePool {
     pub block_created: u64,
 }
 
+#[derive(Copy, Debug, Deserialize, Clone, Eq, PartialEq)]
+pub enum PoolType {
+    Stable,
+    Weighted,
+    Other,
+}
```
vkgnosis, line 69: It looks like the `Other` variant is never created so is there a point in having it?

vkgnosis (approved):

shared/src/sources/balancer/event_handler.rs
```diff
@@ -44,6 +48,8 @@ pub struct BalancerPoolRegistry {
         Mutex<EventHandler<Web3, BalancerV2WeightedPoolFactoryContract, PoolStorage>>,
     two_token_pool_updater:
         Mutex<EventHandler<Web3, BalancerV2WeightedPool2TokensFactoryContract, PoolStorage>>,
+    stable_pool_updater:
+        Mutex<EventHandler<Web3, BalancerV2StablePoolFactoryContract, PoolStorage>>,
```
nlordell, line 52: Maybe it makes sense to have separate registries for weighted pools and stable pools? Not suggesting doing it, just wondering if you considered it and what the trade-offs were.

The reason I'm suggesting this is that I see a bunch of `{Weighted,Stable}Pool -> RegisteredPool -> {Weighted,Stable}Pool` conversions we might be able to avoid with separate registries.

shared/src/sources/balancer/event_handler.rs
```diff
@@ -56,65 +62,119 @@ impl BalancerPoolRegistry {
...
             two_token_pool_updater,
+            stable_pool_updater,
         })
     }
 
-    /// Retrieves `RegisteredWeightedPool`s from each Pool Store in the Registry and
+    /// Retrieves `RegisteredPool`s from each Pool Store in the Registry and
     /// returns the merged result.
     /// Primarily intended to be used by `BalancerPoolFetcher`.
     pub async fn get_pool_ids_containing_token_pairs(
```
nlordell, line 103: Will this be an issue now that a caller would request all the pool ids for the specified token pairs and wouldn't know which one of the `get_weighted_pools` or `get_stable_pools` methods to call it with returned pool IDs?

shared/src/sources/balancer/pool_storage.rs
nlordell: I agree this can be removed.

shared/src/sources/balancer/event_handler.rs
nlordell: Right, but I find the API a bit awkward:
```
let pool_ids = pools.get_pood_ids_containing_token_pairs(...);
let stable_pools = pools.get_stable_pools(pool_ids); // Note that there isn't 1 stable pool per ID
let weighted_pools = pools.get_weighted_pools(pool_ids); // Note that there isn't 1 weighted pool per ID.
```

It feels strange that we are joining stable pools and weighted pools in some places (for example `RegisteredPool`) but keeping them separate in other places (`get_stable_pools`/`get_weighted_pools`).

bh2smith: Good point. I will see if I clean this up with a separation (with hopes of removing all the try_intos). 

I suppose we will need `get_stable_pool_ids_for_token_pairs` and `get_weighted_pool_ids_for_token_pairs` which isn't all that inconvenient anyway. The reason I opted for having one in the first place is that when we introduce yet another pool type (MetaStablePool) we will have three methods of each different kind in many places.

bh2smith: I am not sure this is going to wind up being any nicer overall. We are going to have to split up the `PoolRegistry.pools_by_token` into two separate (but identical methods as well) and have a bunch of duplicated code. I will push a single commit with the changes so you can take a look.

nlordell: You could make `PoolRegistry` generic on the pool types. This would allow you to have a single method and only specify that bits that are different. This would also reduce code duplication and make adding new pools easier in the future.

bh2smith: I have attempted to make some of this stuff generic, but whenever I put a `T` in place of `RegisteredXXXPool` it still expects me to specify something that `T` is known for...

bh2smith: I guess then for all the "common" pool data, I would have to make a trait with methods accessing this and then state that `T: CommonPoolTrait`

bh2smith:
So I have worked on removing this `try_into` stuff by simply keeping the weighted and stable pools separated. This also resulted in removing the enum `RegisteredPool`entirely and does make things quite a bit nicer. 

It remains to deduplicate some of the code blocks for stable and weighted pools (hopefully) using generics.

bh2smith:
@nlordell - I see that with the refactoring of Registered Pool that this has become a bit too large. Will create an intermediary PR.

vkgnosis (approved):

nlordell (approved):
I still feel like these changes don't address the issue I had brought up here: https://github.com/gnosis/gp-v2-services/pull/1007#discussion_r699854168

However, I don't want to block merging this, and we can revist refactoring how we do Balancer liquidity once we have stable pool support.

bh2smith:
> I still feel like these changes don't address the issue I had brought up here: [#1007 (comment)](https://github.com/gnosis/gp-v2-services/pull/1007#discussion_r699854168)
> 
> However, I don't want to block merging this, and we can revist refactoring how we do Balancer liquidity once we have stable pool support.

I will make a follow up PR on this specifically.

PR #1010, opened 2021-08-24
Add config option for solver accounts

### LINKED ISSUE #827: Support multiple accounts in solvers

taminomara:
Fix #827.

### Test Plan
Manual testing (in progress)

### DIFF
--- a/solver/src/main.rs
+++ b/solver/src/main.rs
@@ -93,7 +93,7 @@ struct Arguments {
     /// Individual private keys for each solver
     #[structopt(
         long,
-        env = "SOLVER_PRIVATE_KEY",
+        env = "SOLVER_PRIVATE_KEYS",
         case_insensitive = true,
         use_delimiter = true,
         hide_env_values = true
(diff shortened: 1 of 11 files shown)


vkgnosis (approved):

bh2smith (approved):

solver/src/main.rs
```diff
@@ -90,6 +90,16 @@ struct Arguments {
     )]
     solvers: Vec<SolverType>,
 
+    /// Individual private keys for each solver
+    #[structopt(
+        long,
+        env = "SOLVER_PRIVATE_KEY",
```
nlordell, line 96: Why not?
```suggestion
        env = "SOLVER_PRIVATE_KEYS",
```

nlordell (approved):

PR #1011, opened 2021-08-24
Keep single order solving until deadline runs out

vkgnosis:
We used to fire off the solving process for N orders simultaneously with
each process retrying up to M times. Now we keep on solving the next order
until the deadline is reached without retries.
This has the advantage that the number of orders solved adapts to the
deadline so we always solve as many orders as possible.

Removed the parallelness of the previous code because it makes this
version more complex and it is nicer to the underlying solving apis. We
can reintroduce it later if we feel the need.

Finally I changed the tests of the inner single order solvers to not use
the outer SingleOrderSolving struct because it feels more appropriate to
me and I got rid of the method those tests were using.

In a future PR I would also like to add some metrics here as part of moving from alerts to metrics.

### Test Plan
new unit test


solver/src/solver/single_order_solver.rs
```diff
@@ -59,40 +53,31 @@ impl<I: SingleOrderSolving + Send + Sync + 'static> Solver for SingleOrderSolver
...
-                }
-                Ok(Ok(Some(settlement))) => Some(settlement),
-                Ok(Ok(None)) => None,
-                Ok(Err(err)) => {
-                    // It could be that the inner solver can't match an order and would
-                    // return an error for whatever reason. In that case, we want
-                    // to continue trying to solve for other orders.
-                    tracing::error!("{} inner solver error: {:?}", self.name(), err);
-                    None
+            .collect::<VecDeque<_>>();
```
vkgnosis, line 56: This is the meat of the PR.

solver/src/solver/single_order_solver.rs
```diff
@@ -104,40 +89,6 @@ impl<I: SingleOrderSolving + Send + Sync + 'static> Solver for SingleOrderSolver
...
-            errors.push(err.inner);
-            anyhow!(format!(
-                "{} Errored after {} attempts. Accumulated errors follow \n{}",
-                solver_name,
-                max_retries + 1,
-                errors
-                    .iter()
-                    .enumerate()
-                    .map(|(i, err)| format!("Attempt {}: {}", i + 1, err.to_string()))
-                    .join("\n")
```
bh2smith, line 135: What happened to our pretty accumulated error log?

bh2smith (approved):
Looks good. Sad to see the accumulated error log go.

solver/src/solver/single_order_solver.rs
vkgnosis: I removed it because it made sense when we were retrying the same order several times in a row but not with this logic. Retryable errors get warn logged and unretryable error logged.

solver/src/solver/single_order_solver.rs
```diff
@@ -59,40 +53,31 @@ impl<I: SingleOrderSolving + Send + Sync + 'static> Solver for SingleOrderSolver
...
+            while let Some(order) = orders.pop_front() {
+                match self.inner.try_settle_order(order.clone()).await {
+                    Ok(settlement) => settlements.extend(settlement),
+                    Err(err) => {
+                        let name = self.inner.name();
+                        if err.retryable {
+                            tracing::warn!("Solver {} benign error: {:?}", name, &err);
+                            orders.push_back(order);
+                        } else {
+                            tracing::error!("Solver {} hard error: {:?}", name, &err);
```
nlordell, line 71: Maybe we can increment some metrics instead of alerting with error here. Ok with doing that in a separate PR.

nlordell (approved):
Like this change!

PR #1012, opened 2021-08-24
Use global prometheus registry

taminomara:
This is the first step towards metrics refactoring discussed in [this proposal][1].

[1]: https://gnosisinc.slack.com/archives/C022S0FM1GW/p1626721489006400

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2570,6 +2570,7 @@ dependencies = [
  "model",
  "num",
  "num-bigint 0.3.2",
+ "once_cell",
  "primitive-types",
  "prometheus",
  "regex",
--- a/e2e/tests/services.rs
+++ b/e2e/tests/services.rs
@@ -217,7 +217,6 @@ impl OrderbookServices {
             fee_calculator,
             price_estimator.clone(),
             API_HOST[7..].parse().expect("Couldn't parse API address"),
-            registry,
             metrics,
         );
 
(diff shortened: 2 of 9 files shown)


vkgnosis (approved):

taminomara:
I've realised that both solver and orderbook metrics live in the same namespace. We can't use the default registry with this setup because we need to configure common prefix for metric names.

taminomara:
Ok I'll use the same branch to implement this, reopening.

nlordell (approved):

vkgnosis:
> I've realised that both solver and orderbook metrics live in the same namespace. We can't use the default registry with this setup because we need to configure common prefix for metric names.

Ah, we do for the e2e test and I would like that to continue to work. I guess there can be a global lazily initialized function orderbook_metrics and solver_metrics.

taminomara:
> Ah, we do for the e2e test and I would like that to continue to work. I guess there can be a global lazily initialized function orderbook_metrics and solver_metrics that return the respective registry.

This wouldn't work with common components. For example, transport just gets a global registry and reports its metrics. How would transport know if it should use `orderbook_metrics` or `solver_metrics`?

I'm not sure how to fix this issue. To be completely honest, I'm not sure it needs fixing. I think e2e should just spawn multiple processes, one per service. This way we're as close to the actual setup as possible, and we actually test `main` functions as well.

vkgnosis:
I think you are right that it doesn't need fixing. If you create multiple transports then it makes sense that they use the same registry. And transport metrics do not need a separate orderbook or solver prefix because we can still distinguish them on a per pod basis.

fleupold (commented):
I really like the idea of splitting the Registry and Metrics and having each subcomponent implements its own metrics (having the large metrics component implementing all these traits is painful as you said).

I'm not convinced that using global state for the registry/metrics is really necessary. I think passing a registry as part of the constructor is not so bad and would avoid setting precedence for creating and using non trivial global state in our codebase.

Just a thought, not blocking this if you think that using a global registry is fine.

PR #1014, opened 2021-08-25
Share per solve invocation data between http solver instances

vkgnosis:
Without this we make a lot of duplicate requests to the node as every
http solver instance individually fetches the same data.
In order for us to know what counts as the same solve invocation a new
parameter solve_id is added to the trait which is incremented for
every driver run loop.

### Test Plan
Ran locally and added logging statements to the mutex code to see that multiple instances correctly only fetch data once while the other instances wait.


nlordell (approved):

taminomara (approved):
I love the simplicity of it. This is definitely better than I imagined!

PR #1018, opened 2021-08-25
Implement transport metrics using the new global metrics interface

taminomara:
Note: I want to follow prometheus guidelines for metric names (I'll make a slack thread about it), thus some names were changed. I haven't removed previous metrics implementation so that we can compare old and new metrics after release, and switch our dashboards to new metrics later.

### Test Plan
Manual testing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2575,6 +2575,7 @@ dependencies = [
  "prometheus",
  "regex",
  "reqwest",
+ "scopeguard",
  "serde",
  "serde_json",
  "serde_with",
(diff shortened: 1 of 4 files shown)


nlordell (approved):
I like how the new metrics are ergonomic.

I'm not super happy about all the new singletons we will be adding (one per metrics module), as it would incur some technical debt (global state).

That being said, as we are shifting to use metrics instead of logging alerts, I think having ergonomic metrics is quite important.

vkgnosis (approved):

PR #1019, opened 2021-08-26
Do not allow 100% fee subsidy

### LINKED ISSUE #981: Specifying a fee subsidy of 1.0 causes a panic in the driver.
The command line arguments accept a fee subsidy of 1.0 but this causes a divide by 0 panic in the driver:

https://github.com/gnosis/gp-v2-services/blob/c10ce18e63f0eed673750025d0041e109f6d9960/solver/src/driver.rs#L323

bh2smith:
Closes #981

On second thought, I realized that subsidizing fees by more than 100% does not make sense because we can't just pay people on top of not charging fees. Thus, we avoid the zero division error by not allowing fee_subsidy_factor to equal 1.

### Test Plan
No changed tests.

### DIFF
--- a/shared/src/arguments.rs
+++ b/shared/src/arguments.rs
@@ -93,8 +93,8 @@ pub struct Arguments {
 impl Arguments {
     pub fn validate(&self) {
         assert!(
-            0f64 <= self.fee_subsidy_factor && self.fee_subsidy_factor <= 1f64,
-            "Fee subsidy must be in the range [0, 1]"
+            0f64 <= self.fee_subsidy_factor && self.fee_subsidy_factor < 1f64,
+            "Fee subsidy must be in the range [0, 1)"
         );
     }
 }


taminomara (approved):

nlordell (changes requested):
I think this should be reworked to incorporate what we discussed on Slack, specifically revert the `fee_subsity_factor` change and just make it a `fee_factor` that is multiplied with the fee (instead of `1 - fee_subsidy_factor`:

> What I don’t like about this change is that it doesn’t allow us to charge a fee that is larger than what our baseline solver estimates (the max fee the user pays is now is 100% if I understand correctly).
> Before, we could for the short term easily configure a fee factor of 2x or 3x in case we want to prevent spam or get a more realistic approximation of the true costs (in the absence of CoWs and when DEX aggregators win most batches even 100% fee is still far away from our average costs).
> I’d therefore prefer a name change but leave the semantics of the value as it was before.

Also, I'm not sure I understand this statement:

> On second thought, I realized that subsidizing fees by more than 100% does not make sense because we can't just pay people on top of not charging fees. Thus, we avoid the zero division error by not allowing fee_subsidy_factor to equal 1.

To me, subsidizing fees _more_ than 100% means that the `fee_subsidy_factor > 1`, not `>= 1`. Additionally, while I agree with the statement that:

> subsidizing fees by more than 100% does not make sense

I can still see a use case for having exactly 100% fee subsidy (thus, `fee_subsidy_factor == 1`).

taminomara (commented):
After Nick's comment I realised I don't have enough context to approve (or disapprove) this...

PR #1022, opened 2021-08-26
Fix gas estimation

taminomara:
This fixes a bug introduced in #1010: we didn't set the `from` account for calls to `estimate_gas`.

### Test Plan
Given that CI didn't catch the bug in #1010, I'm highly skeptical of it. My local cowswap installation is broken at the moment, perhaps someone else could test it? Or we could deploy to prestable and check it there.


fleupold (approved):
LGTM, I ran locally and produced this settlement tx: https://rinkeby.etherscan.io/tx/0x61e544e559fe64a7d145dc2a1f6a6bc7d2d732cfda12c91ac4251a1b6d74e289

PR #1024, opened 2021-08-26
Split orders sql statement into several use cases

vkgnosis:
We initially decided on having one big statement for all uses that was
customized by several filtering criteria (see OrderFilter). Now we see
that the number of use cases has grown:
- single order: no ordering, filtered by id
- solvable orders: no ordering, filtered by valid_to, executed amounts,
  invalidations
- user orders: ordered by creation date. many filters.
- orders with presignatures: i am not sure how the query will look

Cramming these all into the same query is in some cases impossible
(different orderings) and in other cases inefficient. The query ends up
very complex because it has to accommodate all use cases. We see this in
the `$N IS NULL o.x = $N` statements that are only needed for some type
of queries but not orders.
This is harder to understand for the programmer and more importantly
makes the query hard to optimize for Postgres and harder to reason about
its performance characteristics as the optimal query plan depends on the
value of the parameters [0] .

For these reasons we factor out the common part of the big query and
reassemble it in several smaller queries.

[0] https://www.postgresql.org/docs/13/sql-prepare.html

---

Before writing tests and integrating the new specialized functions I want to hear feedback on the PR. If we like it then I will do so in a follow up PR.

`concatcp!` is a macro for concatenating strings at compile time.

### Test Plan
existing tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -319,6 +319,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "const_format"
+version = "0.2.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "59c7d3aa11be45d56befebb10f4a8785fcb62aabddf5f33638efef922e505ec9"
+dependencies = [
+ "const_format_proc_macros",
+]
+
+[[package]]
+name = "const_format_proc_macros"
+version = "0.2.14"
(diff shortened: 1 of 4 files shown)


nlordell (approved):
Like the change, just some questions.

nlordell (approved):

PR #1025, opened 2021-08-26
Query order presignature status

nlordell:
This PR adds a `SignaturePending` status to order for pre-sign orders in the `OrderMetadata` for signalling that a pre-sign order is waiting for a pre-signature in the GPv2 settlement contract.

Please note that my SQL skillz are :-1:. So please scrutinize the query modification that was made.

### Test Plan

Added new unit tests for the added order status computation. Added an ignored test for the Postgres query for presignature status that can be manually run:
```
$ cargo test -p orderbook -- --ignored postgres_presignature_status --test-threads 1
    Finished test [unoptimized + debuginfo] target(s) in 0.14s
     Running unittests (target/debug/deps/orderbook-68be6fafd7ab0abc)

running 1 test
test database::orders::tests::postgres_presignature_status ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 59 filtered out; finished in 0.06s
```

### Query Rational

We use a subquery in order to get the latest presignature.

```
nlordell=# explain SELECT
  (o.signing_scheme = 'presign' AND COALESCE((
    SELECT (NOT p.signed) as unsigned
    FROM presignature_events p
    WHERE o.uid = p.order_uid
    ORDER BY p.block_number DESC, p.log_index DESC
    LIMIT 1
  ), true)) as presignature_pending
FROM orders o
GROUP BY o.uid;
                                        QUERY PLAN                                        
------------------------------------------------------------------------------------------
 HashAggregate  (cost=12.88..260.70 rows=230 width=58)
   Group Key: o.uid
   ->  Seq Scan on orders o  (cost=0.00..12.30 rows=230 width=61)
   SubPlan 1
     ->  Limit  (cost=1.06..1.06 rows=1 width=17)
           ->  Sort  (cost=1.06..1.06 rows=1 width=17)
                 Sort Key: p.block_number DESC, p.log_index DESC
                 ->  Seq Scan on presignature_events p  (cost=0.00..1.05 rows=1 width=17)
                       Filter: (o.uid = order_uid)
(9 rows)
```

This seems to be more performant than a `JOIN LEFT LATERAL`.

### DIFF
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -447,7 +447,7 @@ components:
     OrderStatus:
       description: The current order status
       type: string
-      enum: [open, fulfilled, cancelled, expired]
+      enum: [signaturePending, open, fulfilled, cancelled, expired]
     OrderCreation:
       description: Data a user provides when creating a new order.
       type: object
@@ -681,7 +681,16 @@ components:
       properties:
         errorType:
           type: string
-          enum: [InvalidSignature, WrongOwner, OrderNotFound, AlreadyCancelled, OrderFullyExecuted, OrderExpired]
+          enum:
+            [
+              InvalidSignature,
+              WrongOwner,
+              OrderNotFound,
+              AlreadyCancelled,
+              OrderFullyExecuted,
+              OrderExpired,
+              OnChainOrder,
+            ]
         description:
           type: string
       required:
--- a/orderbook/src/database/orders.rs
+++ b/orderbook/src/database/orders.rs
@@ -168,8 +168,8 @@ const ORDERS_SELECT: &str = "\
     COALESCE(SUM(t.buy_amount), 0) AS sum_buy, \
     COALESCE(SUM(t.sell_amount), 0) AS sum_sell, \
     COALESCE(SUM(t.fee_amount), 0) AS sum_fee, \
-    (COUNT(invalidations.*) > 0 OR o.cancellation_timestamp IS NOT NULL) AS invalidated \
-    (COUNT(ps.*) > 0) AS presignature_signed \
+    (COUNT(invalidations.*) > 0 OR o.cancellation_timestamp IS NOT NULL) AS invalidated, \
+    (o.signing_scheme = 'presign' AND COUNT(ps.*) = 0) AS presignature_pending \
 ";
 
 const ORDERS_FROM: &str = "\
@@ -360,7 +360,7 @@ struct OrdersQueryRow {
     settlement_contract: Vec<u8>,
     sell_token_balance: DbSellTokenSource,
     buy_token_balance: DbBuyTokenDestination,
-    presignature_signed: bool,
+    presignature_pending: bool,
 }
 
 impl OrdersQueryRow {
@@ -383,7 +383,7 @@ impl OrdersQueryRow {
         if self.valid_to < Utc::now().timestamp() {
             return OrderStatus::Expired;
         }
-        if self.signing_scheme == DbSigningScheme::PreSign && !self.presignature_signed {
+        if self.presignature_pending {
             return OrderStatus::SignaturePending;
         }
         OrderStatus::Open
@@ -502,7 +502,7 @@ mod tests {
             settlement_contract: vec![0; 20],
             sell_token_balance: DbSellTokenSource::External,
             buy_token_balance: DbBuyTokenDestination::Internal,
-            presignature_signed: false,
+            presignature_pending: false,
         };
 
         // Open - sell (filled - 0%)
@@ -524,7 +524,7 @@ mod tests {
         assert_eq!(
             OrdersQueryRow {
                 signing_scheme: DbSigningScheme::PreSign,
-                presignature_signed: true,
+                presignature_pending: false,
                 ..order_row()
             }
             .calculate_status(),
@@ -535,7 +535,7 @@ mod tests {
         assert_eq!(
             OrdersQueryRow {
                 signing_scheme: DbSigningScheme::PreSign,
-                presignature_signed: false,
+                presignature_pending: true,
                 ..order_row()
             }
             .calculate_status(),
--- a/orderbook/src/database/orders.rs
+++ b/orderbook/src/database/orders.rs
@@ -169,20 +169,19 @@ const ORDERS_SELECT: &str = "\
     COALESCE(SUM(t.sell_amount), 0) AS sum_sell, \
     COALESCE(SUM(t.fee_amount), 0) AS sum_fee, \
     (COUNT(invalidations.*) > 0 OR o.cancellation_timestamp IS NOT NULL) AS invalidated, \
-    (o.signing_scheme = 'presign' AND COUNT(ps.*) = 0) AS presignature_pending \
+    COALESCE((o.signing_scheme = 'presign' AND ( \
+        SELECT (NOT p.signed) as unsigned \
+        FROM presignature_events p \
+        WHERE o.uid = p.order_uid \
+        ORDER BY p.block_number DESC, p.log_index DESC \
+        LIMIT 1 \
+    )), true) AS presignature_pending \
 ";
 
 const ORDERS_FROM: &str = "\
     orders o \
     LEFT OUTER JOIN trades t ON o.uid = t.order_uid \
     LEFT OUTER JOIN invalidations ON o.uid = invalidations.order_uid \
-    LEFT OUTER JOIN LATERAL ( \
-        SELECT p.signed \
-        FROM presignature_events as p \
-        WHERE o.uid = p.order_uid \
-        ORDER BY p.block_number DESC, p.log_index DESC \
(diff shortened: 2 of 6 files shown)


vkgnosis:
Could we rebase this set of branches on main so that it includes the change to SQL query to make reviewing easier?

fedgiac (approved):
The SQL query looks good to me.

Maybe an extra test could be that an expired order that is pending a presignature returns "expired" (and not "pending").

orderbook/src/database/orders.rs
```diff
@@ -235,11 +235,19 @@ impl OrderStoring for Postgres {
...
                 orders o \
                 LEFT OUTER JOIN trades t ON o.uid = t.order_uid \
                 LEFT OUTER JOIN invalidations ON o.uid = invalidations.order_uid \
+                LEFT OUTER JOIN LATERAL ( \
+                    SELECT p.signed \
+                    FROM presignature_events as p \
+                    WHERE o.uid = p.order_uid \
+                    ORDER BY p.block_number DESC, p.log_index DESC \
+                    FETCH FIRST 1 ROW ONLY \
+                ) ps ON ps.signed \
```
vkgnosis, line 250: Because we only select one row from the presignature table a more natural way to write this is to use a sub query instead of joining:

```SQL
 (o.signing_scheme = 'presign' AND (
    SELECT p.signed
    FROM presignature_events p
    WHERE o.uid = p.order_uid
    ORDER BY p.block_number DESC, p.log_index DESC
    LIMIT 1
)) AS presignature_pending
```

If postgres complains that the subquery doesn't just return one row (not sure it is smart enough to understand LIMIT 1) then might have to transform it so that it uses an aggregate

```SQL
 (o.signing_scheme = 'presign' AND (
    SELECT COUNT(*)
    FROM presignature_events p
    WHERE o.uid = p.order_uid AND p.signed
    ORDER BY p.block_number DESC, p.log_index DESC
    LIMIT 1
)) AS presignature_pending
```

nlordell: Sounds good. I was under the impression that a lateral join ended up doing an equivalent thing, but I may be wrong.

nlordell:
> Could we rebase this set of branches on main so that it includes the change to SQL query to make reviewing easier?

@vkgnosis should be rebased now.

orderbook/src/database/orders.rs
vkgnosis: I think they are equivalent in what we observe but I feel the sub query is a more natural way to write it. When I tested some other JOIN LATERAL before I also saw that they weren't optimized as well.

nlordell: Ended up having to go with something like:
```sql
    COALESCE((o.signing_scheme = 'presign' AND (
        SELECT (NOT p.signed) as unsigned
        FROM presignature_events p
        WHERE o.uid = p.order_uid
        ORDER BY p.block_number DESC, p.log_index DESC
        LIMIT 1
    )), true) AS presignature_pending
```
The `COALESCE` is needed to deal with the case of no presignatures.

nlordell: Also worth noting that the `most_recent_with_orderuid` index is used :tada:.

orderbook/src/database/orders.rs
```diff
@@ -168,7 +168,14 @@ const ORDERS_SELECT: &str = "\
...
     COALESCE(SUM(t.fee_amount), 0) AS sum_fee, \
-    (COUNT(invalidations.*) > 0 OR o.cancellation_timestamp IS NOT NULL) AS invalidated \
+    (COUNT(invalidations.*) > 0 OR o.cancellation_timestamp IS NOT NULL) AS invalidated, \
+    COALESCE((o.signing_scheme = 'presign' AND ( \
+        SELECT (NOT p.signed) as unsigned \
+        FROM presignature_events p \
+        WHERE o.uid = p.order_uid \
+        ORDER BY p.block_number DESC, p.log_index DESC \
+        LIMIT 1 \
+    )), true) AS presignature_pending \
```
vkgnosis, line 178: Making new discussion so it shows up in the code view again.

The coalesce could be moved to just around the sub query because that is the only part that can be null. Could also do the COUNT thing so you don't need coalesce because with count it won't ever be null but not sure which way of doing it is cleaner.
If the coalesce is not moved only to the inner part that it isn't clear to me what happens in the case where
- signing scheme is != presign
- sub query returns null

I could see Postgres either stopping the evaluation of the AND immediately so presignature_pending is false (correct) or it might evaluate the AND with the second operand being null which turns the whole and into null which is coalesced into true (incorrect).

nlordell: I don't see how to use `COUNT` without filtering `signed = true` rows **after** doing the subquery for the most recent row (otherwise, you will just count the most recent signed presignature row, and not count whether or not the most presignature row is signed - I hope that makes sense).

I could also put `COALESCE` around the subquery only, but I expected that the `o.signing_scheme = 'presign'` would shortcircuit the subquery anyway, and `COALESCE(false, ...)` wouldn't have a performance impact.

vkgnosis: I don't know if it short circuits that's why I'm suggesting this. But even if it didn't the coalesce around the whole query just isn't needed because `o.signing_scheme = 'presign'` cannot be null so why not coalesce only the second part?

> otherwise, you will just count the most recent signed presignature row
Makes sense.

nlordell: > why not coalesce only the second part?

For sure! When I wrote:

> I could also put COALESCE around the subquery only, but I expected that the

I meant it more like

> Yes, I can put COALESCE around the subquery only. The reason I had done it this way was that I expected ...

nlordell: I quickly checked, and it looks like where the `COALESCE` is doesn't matter for the execution plan.

<details><summary>execution plan</summary>

```
nlordell=# explain SELECT
  COALESCE((o.signing_scheme = 'presign' AND (
    SELECT (NOT p.signed) as unsigned
    FROM presignature_events p
    WHERE o.uid = p.order_uid
    ORDER BY p.block_number DESC, p.log_index DESC
    LIMIT 1
  )), true) as presignature_pending
FROM orders o
GROUP BY o.uid;
                                        QUERY PLAN                                        
------------------------------------------------------------------------------------------
 HashAggregate  (cost=12.88..260.70 rows=230 width=58)
   Group Key: o.uid
   ->  Seq Scan on orders o  (cost=0.00..12.30 rows=230 width=61)
   SubPlan 1
     ->  Limit  (cost=1.06..1.06 rows=1 width=17)
           ->  Sort  (cost=1.06..1.06 rows=1 width=17)
                 Sort Key: p.block_number DESC, p.log_index DESC
                 ->  Seq Scan on presignature_events p  (cost=0.00..1.05 rows=1 width=17)
                       Filter: (o.uid = order_uid)
(9 rows)

nlordell=# explain SELECT
  (o.signing_scheme = 'presign' AND COALESCE((
    SELECT (NOT p.signed) as unsigned
    FROM presignature_events p
    WHERE o.uid = p.order_uid
    ORDER BY p.block_number DESC, p.log_index DESC
    LIMIT 1
  ), true)) as presignature_pending
FROM orders o
GROUP BY o.uid;
                                        QUERY PLAN                                        
------------------------------------------------------------------------------------------
 HashAggregate  (cost=12.88..260.70 rows=230 width=58)
   Group Key: o.uid
   ->  Seq Scan on orders o  (cost=0.00..12.30 rows=230 width=61)
   SubPlan 1
     ->  Limit  (cost=1.06..1.06 rows=1 width=17)
           ->  Sort  (cost=1.06..1.06 rows=1 width=17)
                 Sort Key: p.block_number DESC, p.log_index DESC
                 ->  Seq Scan on presignature_events p  (cost=0.00..1.05 rows=1 width=17)
                       Filter: (o.uid = order_uid)
(9 rows)
```

</details>

With this in mind, I prefer your suggestion of putting `COALESCE` only around the subquery, as it makes it clearer that it is the part of the expression that can be `NULL`.

nlordell: @vkgnosis changed to only `COALESCE` the subquery.

orderbook/openapi.yml
```diff
@@ -447,7 +447,7 @@ components:
     OrderStatus:
       description: The current order status
       type: string
-      enum: [open, fulfilled, cancelled, expired]
+      enum: [signaturePending, open, fulfilled, cancelled, expired]
```
nlordell, line 450: @anxolin @W3stside @alfetopito 

Just pinging you guys so you are aware of the new order status (for an order that was accepted by the orderbook and is waiting for the on-chain presignature).

vkgnosis (approved):

orderbook/openapi.yml
alfetopito: Thanks for letting us know.
I've created one task for the explorer https://github.com/gnosis/gp-ui/issues/628

I believe Anxo is taking care of cowswap changes.

PR #1030, opened 2021-08-27
Add parameter to filter order query on pending presignature

nlordell:
This PR adds a new `OrderFilter` parameter to filter orders that are waiting on a presignature. This will be used when fetching the list of solvable orders from the orderbook.

### Test Plan

Adjusted some existing unit tests for the new filter parameter and added a new manual Postgres test for checking filtering logic.

<details><summary><code>cargo test -p orderbook -- --ignored postgres_filter_presignature_pending --test-threads 1</code></summary>

```
   Compiling orderbook v0.1.0 (/var/home/nlordell/Developer/gp-v2-services/orderbook)
    Finished test [unoptimized + debuginfo] target(s) in 14.25s
     Running unittests (target/debug/deps/orderbook-68be6fafd7ab0abc)

running 1 test
test database::orders::tests::postgres_filter_presignature_pending ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 60 filtered out; finished in 0.15s
```

</details>

**edit**: TIL that postgres tests run in CI. This is good, as the test :point_up: will also run in CI :smile: :tada: :rocket: :robot:.

### DIFF
--- a/orderbook/src/api/get_orders.rs
+++ b/orderbook/src/api/get_orders.rs
@@ -29,6 +29,8 @@ struct Query {
     include_insufficient_balance: bool,
     #[serde(default)]
     include_unsupported_tokens: bool,
+    #[serde(default)]
+    include_signature_pending: bool,
 }
 
 impl Query {
@@ -46,6 +48,7 @@ impl Query {
             exclude_invalidated: !self.include_invalidated,
             exclude_insufficient_balance: !self.include_insufficient_balance,
             exclude_unsupported_tokens: !self.include_unsupported_tokens,
+            exclude_presignature_pending: !self.include_signature_pending,
             uid: None,
         })
     }
@@ -89,8 +92,7 @@ pub fn get_orders(
 mod tests {
     use super::*;
     use crate::api::response_body;
-    use hex_literal::hex;
-    use primitive_types::H160;
+    use shared::addr;
     use warp::test::{request, RequestBuilder};
 
     #[tokio::test]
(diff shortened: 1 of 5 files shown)


fleupold (approved):
Logic looks good!

orderbook/src/api/get_orders.rs
```diff
@@ -29,6 +29,8 @@ struct Query {
     include_insufficient_balance: bool,
     #[serde(default)]
     include_unsupported_tokens: bool,
+    #[serde(default)]
+    include_signature_pending: bool,
```
fedgiac, line 33: Nit: this parameter is only meaningful for presignatures, so I'd prefer this to be reflected in the name: `include_presignature_pending`. Especially since *pre*signature is used in a similar variable somewhere else in the PR.

fedgiac, line 33: You could add a parameter description in the Swagger.

orderbook/src/database/orders.rs
```diff
@@ -261,7 +262,8 @@ impl OrderStoring for Postgres {
                 WHEN 'sell' THEN sum_sell < sell_amount \
                 WHEN 'buy' THEN sum_buy < buy_amount \
             END) AND \
-            ($7 OR NOT invalidated);";
+            ($7 OR NOT invalidated) AND
```
fedgiac, line 265: Extremely nitty:
```suggestion
            ($7 OR NOT invalidated) AND \
```

fedgiac (approved):
Looks good! All comments are optional.

orderbook/src/api/get_orders.rs
nlordell: The naming discrepancy was international as I figured this may be used for EIP-1271 as well. That being said, now that I think about it more - we should probably use `include_presignature_pending` for now, and rename later if we end up needing something like this for EIP-1271 as well.

vkgnosis (approved):

orderbook/src/database/orders.rs
nlordell: Nice catch!

PR #1034, opened 2021-08-27
Use single order specialized query

vkgnosis:
and fixes openapi documentation saying we return an array for this route
when we were already always returning a single order.

### Test Plan
new test for db


nlordell (approved):
Looks good to me.

PR #1037, opened 2021-08-28
[Paraswap] Parse PriceQuotes that contain error+route as error

fleupold:
We are sometimes getting responses from the paraswap Price Query that contain both an error and a price route (like [this one](https://logs.gnosisdev.com/app/kibana#/discover/doc/de1f5360-a9ae-11ea-acbb-9d2e3cb124ca/logstash-2021.08.27?id=oXi8h3sBlg1NXsCMnHbu)). We currently classify those as success and are trying to convert them into a settlement tx.

This request always fails with an error that happens to be "retryable" (PriceChanged). However, since the price impact is too high our subsequent retries also run into the same issue.

This PR makes it so that we classify mixed responses as error. For now, we will still retry that error however (I will change TooMuchSlippage to not be retryable in a next PR)

### Test Plan
Added a unit test for mixed deserialization.

### DIFF
--- a/solver/src/solver/paraswap_solver/api.rs
+++ b/solver/src/solver/paraswap_solver/api.rs
@@ -51,6 +51,9 @@ impl ParaswapApi for DefaultParaswapApi {
                 "No routes found with enough liquidity" => {
                     Err(ParaswapResponseError::InsufficientLiquidity)
                 }
+                "ESTIMATED_LOSS_GREATER_THAN_MAX_IMPACT" => {
+                    Err(ParaswapResponseError::TooMuchSlippageOnQuote)
+                }
                 _ => Err(ParaswapResponseError::UnknownParaswapError(format!(
                     "uncatalogued Price Query error message {}",
                     message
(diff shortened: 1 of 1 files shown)


nlordell (approved):

bh2smith (approved):
I love that reversing the order of the enum changes the parsed result.

nlordell:
> I love that reversing the order of the annum changes the parsed result.

Maybe this may be a reason to consider making the `Deserialize` implementation more explicit?

bh2smith:
> > I love that reversing the order of the enum changes the parsed result.
> 
> Maybe this is a reason to consider making the `Deserialize` implementation more explicit?

Should we create an issue for this?

vkgnosis (approved):

vkgnosis:
Once you understand what is happening it makes sense that the order matters because serde in general will ignore unknown fields and this is an untagged enum. I would just add a comment to the enum explaining that the order is important but I don't think we need to manually implement Deserialize.

PR #1040, opened 2021-08-29
[Trivial] Remove error log on unsupported pool factory

fleupold:
The [StablePoolFactory](https://etherscan.io/address/0xc66ba2b6595d3613ccab350c886ace23866ede24#code) is not yet indexed, therefore we are logging this error on every restart causing a noisy alert.

Downgrading to WARN which will still allow us to see this when we check logs periodically

### Test Plan
CI

### DIFF
--- a/shared/src/sources/balancer/pool_init.rs
+++ b/shared/src/sources/balancer/pool_init.rs
@@ -20,6 +20,8 @@ use futures::stream::{self, StreamExt as _, TryStreamExt as _};
 use reqwest::Client;
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


fleupold:
Looks like our code coverage considers comments as test hits 😞

bh2smith (approved):

nlordell (commented):
I'm not sure I agree with this change. Specifically, the error log was there so we get notified when our Subgraph query is returning pools for factories that we weren't expecting.

For now can't we just:
```
pools.pools_by_factory.remove(STABLE_POOL_FACTORY_ADDRESS);
```

PR #1051, opened 2021-08-31
Add specialized query for querying user orders paginated

vkgnosis:
This is going to be used by a new route for the frontend.

related to https://github.com/gnosis/gp-v2-services/issues/969 .

I decided to keep the database level operations close to what happens under the hood. The http api can still provide a different interface.

I tried a lot of varations on the query but as I documented in the code this is what I found to work best taking postgres limitations into account.

### Test Plan
new unit test and looked at query analyzer

### DIFF
--- a/database/sql/V016__update_order_indexes.sql
+++ b/database/sql/V016__update_order_indexes.sql
@@ -1,0 +1,6 @@
+-- The following index already allows indexing by owner.
+DROP INDEX order_owner;
+
+-- The index can be traversed in both directions anyway but we usually fetch user orders with this
+-- ordering so using it here too.
+CREATE INDEX user_order_creation_timestamp ON orders USING BTREE (owner, creation_timestamp DESC);
--- a/orderbook/src/database/orders.rs
+++ b/orderbook/src/database/orders.rs
@@ -17,11 +17,18 @@ use std::{borrow::Cow, convert::TryInto};
 pub trait OrderStoring: Send + Sync {
     async fn insert_order(&self, order: &Order) -> Result<(), InsertionError>;
     async fn cancel_order(&self, order_uid: &OrderUid, now: DateTime<Utc>) -> Result<()>;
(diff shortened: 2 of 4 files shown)


bh2smith (approved):

database/sql/V016__update_order_indexes.sql
```diff
@@ -0,0 +1,6 @@
+-- The following index already allows indexing by owner.
+DROP INDEX order_owner;
```
nlordell, line 2: Should we also drop `order_creation_timestamp` index as well?

nlordell (approved):

orderbook/src/database/orders.rs
```diff
@@ -161,15 +168,33 @@ impl From<sqlx::Error> for InsertionError {
...
+//
+// It might feel more natural to use aggregate, joins and group by to calculate trade data and
+// invalidations. The problem with that is that it makes the query inefficient when we wish to limit
+// or order the selected orders because postgres is unable to understand the query well enough.
+// For example if we want to select orders ordered by creation timestamp on which there is an index
+// with offset and limit then the group by causes postgres to not use the index for the ordering. So
+// it will select orders, aggregate them with trades grouped by uid, sort by timestamp and only then
+// apply the limit. Instead we would like to apply the limit first and only then aggregate but I
+// could not get this to happen without writing the query using sub queries. A similar situation is
+// discussed in https://dba.stackexchange.com/questions/88988/postgres-error-column-must-appear-in-the-group-by-clause-or-be-used-in-an-aggre .
```
nlordell, line 180: Nice explanation.

database/sql/V016__update_order_indexes.sql
vkgnosis: That is going to be used in the future and is different from the index on (owner, creation_timestamp) because the latter doesn't allow you to order orders just by creation_timestamp.

PR #1052, opened 2021-08-31
Set X-Partner header on Paraswap queries

fleupold:
Context: https://gnosisinc.slack.com/archives/C0203Q9R4JJ/p1630337209009700

### Test Plan
- Install proxyman and add certificate to local keychain to intercept HTTPs requests
- export `https_proxy=<proxyman listening port>`
- run solver on mainnet, watch paraswap API requests, see header being set.

### DIFF
--- a/solver/src/solver/paraswap_solver/api.rs
+++ b/solver/src/solver/paraswap_solver/api.rs
@@ -11,6 +11,7 @@ use model::u256_decimal;
 use web3::types::Bytes;
 
 const BASE_URL: &str = "https://apiv4.paraswap.io";
+const PARTNER_HEADER_KEY: &str = "X-Partner";
 
 /// Mockable implementation of the API for unit test
 #[cfg_attr(test, mockall::automock)]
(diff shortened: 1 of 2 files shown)


solver/src/solver/paraswap_solver/api.rs
```diff
@@ -33,7 +36,11 @@ impl ParaswapApi for DefaultParaswapApi {
         let query_str = format!("{:?}", &query);
         let url = query.into_url();
         tracing::debug!("Querying Paraswap API (price) for url {}", url);
-        let response_text = reqwest::get(url)
```
fleupold, line 36: nasty leftover spot of not using our shared client (with timeout). Not a big issue since the solver futures have a timeout.

vkgnosis (approved):
Too late now but maybe the value should be treated as secret. Otherwise what prevents someone else from using the same referrer?

bh2smith (approved):

fleupold:
> Too late now but maybe the value should be treated as secret. 

Good point. I didn't do it because we use the same as the referrer (which is public on chain information). Will flag to their team and if we replace it with a proper secret, agree we should hide the value.

Issue #967, opened 2021-08-13
API service clear account balance subscription keys


nlordell:
From what I can tell, at a first glance of the code, the account balance subscription keys aren't ever cleared. This means over time we would query balances for more and more accounts.

I can't find evidence of this in the logs, however, so I may have missed something, making this a non-issue.

nlordell:
This appears to be an issue:

![image](https://user-images.githubusercontent.com/4210206/131794763-9ffab9f2-3ab8-433e-b292-61138c56e8ac.png)

vkgnosis:
Thought about this for a bit. Similar to my recent PR we really only need balances for orders that are solvable (open) right now. What we could do then is as part of maintenance every block get the whole set of solvable orders, update these balances, cache the orders and balances until the next maintenance.
This would also make us not query the database every time the solvable_orders endpoint is fetched because we would always have an up to date set of solvable orders and their balances.

nlordell:
One thing to be careful about is that we might miss orders that came after the latest block but before the next one. While, in general, we won't submit settlements for them because they aren't old enough, we might miss out on :cow: opportunities. Just something to keep in mind.

nlordell:
@vkgnosis I believe this is no longer an issue correct?

PR #1053, opened 2021-09-01
Do not track balances for orders that are not open

vkgnosis:
The field is only of interest for open orders so we shouldn't waste
resources tracking balances of historical orders.

### Test Plan
adjusted unit test

### DIFF
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -324,7 +324,7 @@ async fn filter_unsupported_tokens(
 }
 
 // Make sure the balance fetcher tracks all balances for user, sell token combinations in these
-// orders and returns said balances.
+// orders and returns said balances. Only records this for open orders.
 async fn track_and_get_balances(
     fetcher: &dyn BalanceFetching,
     orders: &[Order],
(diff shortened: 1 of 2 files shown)


orderbook/src/orderbook.rs
```diff
@@ -461,6 +464,9 @@ mod tests {
         let another_sell_token = H160::from_low_u64_be(3);
         let another_balance = 200.into();
 
+        // Should not get tracked because corresponding order isn't open.
+        let untracked_sell_token = H160::from_low_u64_be(4);
```
bh2smith, line 468: What is the meaning of using the token address "4" here?

bh2smith (approved):

orderbook/src/orderbook.rs
vkgnosis: No meaning. The test just needs a couple of distinct sell token addresses and it already uses 2 and 3 a couple of lines above.

taminomara (approved):

nlordell:
I created a relevant issue here: https://github.com/gnosis/gp-v2-services/issues/967

When I was touching the `account_balances.rs` code, it wasn't obvious to me that the subscription keys determining which balances to track were getting cleaned up correctly.

Just wanted to bring it up in case you noticed anything.

nlordell (approved):

PR #1059, opened 2021-09-01
Log solver addresses

taminomara:
Log solver addresses so that we can ensure we haven't messed up their order.

### Test Plan
Manual testing

### DIFF
--- a/solver/src/solver.rs
+++ b/solver/src/solver.rs
@@ -134,53 +134,68 @@ pub fn create(
 
     solvers
         .into_iter()
-        .map(|(account, solver_type)| match solver_type {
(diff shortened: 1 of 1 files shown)


taminomara:
Note: I am *certain* that one day we'll mess up order of solver addresses in the config variable. So I think switching to toml configs is a must.

vkgnosis (approved):
Could also take the solver addresses as a map instead of a list. Like
```
--solver-private-keys Naive=0x01,Baseline=0x02,...
```
Not saying config file isn't better, just an idea.

nlordell (approved):

PR #1061, opened 2021-09-01
Refactor price estimation to avoid boxing

### LINKED ISSUE #1041: Remove recursion from the shared price estimation component
https://github.com/gnosis/gp-v2-services/pull/1021#pullrequestreview-741520884
> I still find the price estimator quite hairy so I'm not super confident in my review but I do think it is correct. I have some ideas on how to refactor this file to get rid of async_recursion and boxing and maybe make it easier to read but I don't want to block this PR with that discussion so we can test in staging earlier.

vkgnosis:
fixes #1041

Note that the diff is pretty hard to read. I would open both versions side by side to understand what is going on.

Technically we could get rid of the estimate_gas_cost: bool parameter again but I think it might be useful in the future and is still useful in tests.

In order to take gas cost into account price estimation needs to perform
one extra estimate to get the price in the native of token of the sell
token for buy orders and the buy token for sell orders.

This order of function calls used to be (for a sell order):
```
price_helper(sell_token, buy_token gas = true)
    best_e_sell_order(sell_token, buy_token, gas = true)
        price_helper(native_token, buy_token, gas = false)
            best_e_sell_order(native_token, buy_token, gas = false)
```

Now it is:
```
price_helper(sell_token, buy_token gas = true)
    best_e_sell_order(native_token, buy_token)
    best_e_sell_order(sell_token, buy_token)
```

The extra estimate with the native token has moved directly into
price_helper.

This also got rid of boxing the path_comparison functions and I have
inlined the evalue_* functions because I found it easier to read this
way.

### Test Plan
existing tests, will also manually compare a few prices to what is running in staging

### DIFF
--- a/shared/src/price_estimate.rs
+++ b/shared/src/price_estimate.rs
@@ -9,7 +9,6 @@ use crate::{
     sources::uniswap::pool_fetching::{Pool, PoolFetching},
 };
 use anyhow::{anyhow, Result};
-use async_recursion::async_recursion;
 use ethcontract::{H160, U256};
 use futures::future::join_all;
 use gas_estimation::GasPriceEstimating;
@@ -184,6 +183,22 @@ impl PriceEstimating for BaselinePriceEstimator {
     }
 }
 
+fn amounts_to_price(
+    sell_amount: U256,
+    buy_amount: U256,
+) -> Result<BigRational, PriceEstimationError> {
+    // Due to integer rounding it is possible that we find path but the amount is 0.
+    if sell_amount.is_zero() || buy_amount.is_zero() {
+        return Err(PriceEstimationError::Other(anyhow!(
+            "zero amount leads to impossible price"
+        )));
+    }
+    Ok(BigRational::new(
+        sell_amount.to_big_int(),
+        buy_amount.to_big_int(),
+    ))
+}
+
 impl BaselinePriceEstimator {
     /// Returns the path and the price.
     async fn estimate_price_helper(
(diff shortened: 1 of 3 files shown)


nlordell (approved):

nlordell:
Nice cleanup.

shared/src/price_estimate.rs
```diff
@@ -184,6 +183,22 @@ impl PriceEstimating for BaselinePriceEstimator {
     }
 }
 
+fn amounts_to_price(
+    sell_amount: U256,
+    buy_amount: U256,
+) -> Result<BigRational, PriceEstimationError> {
+    // Due to integer rounding it is possible that we find path but the amount is 0.
+    if sell_amount.is_zero() || buy_amount.is_zero() {
```
marcovc, line 191: just curious, what's wrong with allowing sell_amount=0 ?

shared/src/price_estimate.rs
```diff
@@ -205,178 +220,146 @@ impl BaselinePriceEstimator {
...
+        let path_comparison = |sell_estimate: Estimate<U256, Pool>| {
+            if let Some(sell_token_price_in_native_token) = &sell_token_price_in_native_token {
+                let sell_amount_in_native_token =
+                    sell_estimate.value.to_big_rational() * sell_token_price_in_native_token;
+                let tx_cost_in_native_token = U256::from_f64_lossy(gas_price).to_big_rational()
+                    * BigRational::from_integer(sell_estimate.gas_cost().into());
+                -sell_amount_in_native_token - tx_cost_in_native_token
+            } else {
+                -sell_estimate.value.to_big_rational()
+            }
```
marcovc, line 343: No big deal, but notice that here we'll have a condition repeatedly evaluated inside the function, where before the condition was factored out.

marcovc (approved):
Thanks. I cannot say I prefer it, but I guess that is because I'm not used to parse rust. I also didn't understood what that async recursion meant before :) 

In case you haven't, could you please also double check with the uniswap frontend if the prices we're getting still make sense? This refactor is just a refactor, but just to make sure since a potential bug would be really hard to detect later.

Thank you!

shared/src/price_estimate.rs
vkgnosis: I guess  technically we can allow it but it doesn't make sense to me that you would have a 0 price. Currently the price is always sell / buy but we could have arbitrarily chosen buy / sell too in which case it would be division by 0 again.

shared/src/price_estimate.rs
vkgnosis: Yes. It is avoidable but makes the code a bit uglier so I did it this way.

bh2smith (approved):

nlordell:
> I also didn't understood what that async recursion meant before :)

For some context, Rust `async` functions make the compiler generate a "state-machine" type under the hood in order to implement the future. So, in pseudo-code:

```rust
async fn foo() -> i32 {
  thing();
  async_thing().await;
  other_thing();
  other_async_thing().await;
  42
}
``` 

Roughly (very very roughly...) would translate to something:
```rust
enum FooFuture {
  WaitingForAsyncThing(AsyncThingFuture),
  WaitingForOtherAsyncThing(OtherAsyncThingFuture),
}

impl Future for FooFuture {
  type Output = i32;
  fn poll(self: Pin<&mut Self>, context: &mut Context) -> Poll<Self::Output> {
    // Generated implementation here...
  }
}
```

This becomes a problem with recursion, because:
```rust
enum FooFuture {
  // ...
  WaitingForFooFuture(FooFuture),
}
```

Would be an infinitely sized struct and isn't allowed. The solution is box the future:
```rust
enum FooFuture {
  // ...
  WaitingForFooFuture(Box<FooFuture>),
}
```

And I assume that the `async_recursion` procedural macro generates the appropriate boxing required to make this work automatically.

vkgnosis:
Tested against staging and observed no changes compared to locally running this PR. Tested a few tokens through https://v2.info.uniswap.org/home which seemed fine too.

marcovc:
Thanks @nlordell  for the explanation, and @vkgnosis for double checking !

PR #1062, opened 2021-09-01
Reduce some rebuilding on CI

vkgnosis:
The different crates in our workspace activate different features in dependencies. This means that some dependencies have to be recompiled based on which crate (--package) is being compiled.
This commit avoids this by introducing one global build step first.
Sadly we still recompilations for the e2e tests because the only way to run them is through --package .
Also moved some commands into their own step so they are easier to keep track of in the github actions page output.

### Test Plan
ci still works

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -38,21 +38,21 @@ jobs:
           path: flyway-${{ env.FLYWAY_VERSION }}
           key: ${{ runner.os }}-build-flyway-${{ env.FLYWAY_VERSION }}
       - run: "[[ -d flyway-${FLYWAY_VERSION} ]] || (curl -L https://repo1.maven.org/maven2/org/flywaydb/flyway-commandline/${FLYWAY_VERSION}/flyway-commandline-${FLYWAY_VERSION}.tar.gz -o flyway-commandline-${FLYWAY_VERSION}.tar.gz && tar -xzf flyway-commandline-${FLYWAY_VERSION}.tar.gz && rm flyway-commandline-${FLYWAY_VERSION}.tar.gz)"
+      - run: cargo build --locked --all-features --tests
       - run: cargo test --locked --all-features
       - run: |
           sudo systemctl start postgresql.service
           sudo -u postgres createuser $USER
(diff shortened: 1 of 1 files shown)


bh2smith (approved):

.github/workflows/pull-request.yaml
```diff
@@ -38,21 +38,21 @@ jobs:
...
           sudo systemctl start postgresql.service
           sudo -u postgres createuser $USER
           sudo -u postgres createdb $USER
           psql -c "ALTER USER $USER PASSWORD '$FLYWAY_PASSWORD';"
-          flyway-${FLYWAY_VERSION}/flyway -url="jdbc:postgresql:///" -user=$USER -locations="filesystem:database/sql/" migrate 
-          # Postgres tests should not run in parallel because they use the same database.
-          cargo test --locked --workspace --all-features --jobs 1 postgres -- --ignored --test-threads 1
+          flyway-${FLYWAY_VERSION}/flyway -url="jdbc:postgresql:///" -user=$USER -locations="filesystem:database/sql/" migrate
+        # Postgres tests should not run in parallel because they use the same database.
+      - run: cargo test --locked --all-features --jobs 1 postgres -- --ignored --test-threads 1
```
nlordell, line 50:
```suggestion
      - run: cargo test --locked --all-features postgres -- --ignored --test-threads 1
```

Why reduce the number of jobs for building the tests? AFAIK `--jobs` does not affect the number of threads for executing tests.

nlordell (approved):
Something that may be worth keeping in mind (but I don't think applies here anymore seeing how the `cargo` invocations changed) is a ["workspace hack" crate like the Rust-lang repo uses](https://github.com/rust-lang/rust/tree/master/src/tools/rustc-workspace-hack) in order to deal with different crates needing different features in their shared dependencies?

.github/workflows/pull-request.yaml
vkgnosis: I didn't add this in this PR but yeah I think you're right that it isn't needed. Will apply the suggestion.

nlordell: Ah right - didn't notice it was from before.

PR #1063, opened 2021-09-01
Fix match condition for signature recovery

nlordell:
This PR fixes an issue introduced when adding pre-signature order cancellation error codes.

### Test Plan

Test coming in follow up PR - want to merge this to unblock FE.


alfetopito (approved):
:shrug:

bh2smith (approved):

PR #1068, opened 2021-09-02
remove limit order from liquidity

nlordell:
This PR removes `LimitOrder` from the `Liquidity` enum. This was done as a first step towards refactoring to make it easier to add new liquidity such that:
- `Liquidity` type will eventually be all the baseline liquidity that we fetch. This will hopefully allow us to simplify the whole solver liquidity vs API liquidity issue and make it easier to compute baseline solutions in the orderbook API for bad token detection, etc.
- We were grouping limit orders and AMM liquidity only to split it back up again in _every_ solver: `([LimitOrder], [Amms]) -> Liquidity -> ([LimitOrder], [Amms])`
- If we add actual AMM limit orders in the future (such as 0x orders) then we would need to distinguish them between the user orders.

### Test Plan

Mostly refactored for new split, should be caught by existing CI.

### DIFF
--- a/solver/src/solver/baseline_solver.rs
+++ b/solver/src/solver/baseline_solver.rs
@@ -32,8 +32,13 @@ pub struct BaselineSolver {
 
 #[async_trait::async_trait]
 impl Solver for BaselineSolver {
-    async fn solve(&self, Auction { liquidity, .. }: Auction) -> Result<Vec<Settlement>> {
-        Ok(self.solve(liquidity))
+    async fn solve(
+        &self,
+        Auction {
+            orders, liquidity, ..
+        }: Auction,
+    ) -> Result<Vec<Settlement>> {
+        Ok(self.solve(orders, liquidity))
     }
 
     fn account(&self) -> &Account {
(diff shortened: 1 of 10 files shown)


vkgnosis (approved):

solver/src/solver/baseline_solver.rs
```diff
@@ -141,30 +146,29 @@ impl BaselineSolver {
...
+                                tokens: order.tokens,
+                                order: AmmOrder::ConstantProduct(order),
                             });
                         }
+                        Liquidity::WeightedProduct(order) => {
+                            for tokens in order.token_pairs() {
+                                amm_map.entry(tokens).or_default().push(Amm {
+                                    tokens,
+                                    order: AmmOrder::WeightedProduct(order.clone()),
+                                });
```
bh2smith, line 166: order is cloned in one place but not the other...

solver/src/solver/http_solver.rs
```diff
@@ -123,18 +123,17 @@ impl HttpSolver {
...
-                Liquidity::ConstantProduct(amm) => {
-                    vec![amm.tokens.get().0, amm.tokens.get().1]
-                }
-                Liquidity::WeightedProduct(amm) => amm.reserves.keys().cloned().collect(),
-            })
+            .flat_map(|order| [order.sell_token, order.buy_token]);
+        let liquidity_tokens = liquidity.iter().flat_map(|liquidity| match liquidity {
+            Liquidity::ConstantProduct(amm) => amm.tokens.into_iter().collect::<Vec<_>>(),
+            Liquidity::WeightedProduct(amm) => amm.reserves.keys().cloned().collect(),
+        });
```
bh2smith, line 133: Nice.

solver/src/solver/http_solver.rs
```diff
@@ -322,18 +322,18 @@ impl HttpSolver {
...
+            &mut orders,
+            &constant_product_liquidity,
             &self.native_token,
         );
-        let limit_orders = self.map_orders_for_solver(orders.0);
-        let constant_product_orders = self.map_amm_orders_for_solver(orders.1);
-        let weighted_product_orders = self.map_amm_orders_for_solver(orders.2);
+        let limit_orders = self.map_orders_for_solver(orders);
+        let constant_product_orders = self.map_amm_orders_for_solver(constant_product_liquidity);
+        let weighted_product_orders = self.map_amm_orders_for_solver(weighted_product_liquidity);
```
bh2smith, line 336: Also much nicer.

solver/src/solver/http_solver.rs
```diff
@@ -449,28 +449,19 @@ impl HttpSolver {
...
-    (
-        limit_orders,
-        constant_product_orders,
-        weighted_product_orders,
-    )
+    (constant_product_orders, weighted_product_orders)
 }
 
+// TODO: This currently does **NOT** consider balancer pools, but definitely should.
 fn remove_orders_without_native_connection(
```
bh2smith, line 465: That is weird... how is this software already running fine on weighted pools if they aren't here? Is this related to price estimation? Should we link this TODO in #1075?

bh2smith (approved):

solver/src/solver/http_solver.rs
nlordell: I already created a separate issue in https://github.com/gnosis/gp-v2-services/issues/1071

This was probably something that slipped by in the initial Balancer V2 weighted pool integration.

solver/src/solver/baseline_solver.rs
nlordell: This was already the case before (look at line 160 that got removed).

Additionally, this is needed here. The reason is you have 1 token pair per order for constant product orders, so no `clone` required, however you have n token pairs per order for weighted products, so you need to clone `n` times. Also worth noting, clippy should report on superfluous `clone` calls, so the linter should catch unnecessary clones.

PR #1069, opened 2021-09-02
Introduce Auction struct passed to solver 

nlordell:
This PR just adds a new `Auction` type as input to the solvers instead of multiple parameters. This is used in #1068 and split into its own PR to make reviewing easier (as this change is arguably trivial to review).

### Test Plan

No logic change, so existing CI.

### DIFF
--- a/solver/src/solver.rs
+++ b/solver/src/solver.rs
@@ -1,7 +1,4 @@
-use crate::{
-    liquidity::{LimitOrder, Liquidity},
-    settlement::Settlement,
-};
+use crate::{liquidity::Liquidity, settlement::Settlement};
 use anyhow::Result;
 use baseline_solver::BaselineSolver;
 use contracts::GPv2Settlement;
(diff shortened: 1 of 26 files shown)


vkgnosis (approved):

solver/src/solver.rs
```diff
@@ -59,6 +53,46 @@ pub trait Solver: 'static {
...
+pub struct Auction {
+    /// An ID that idetifies a batch within a `Driver` isntance.
+    ///
+    /// Note that this ID is not unique across multiple instances of drivers,
+    /// in particular it cannot be used to uniquely identify batches across
+    /// service restarts.
+    pub id: u64,
+
+    /// The collection of all liquidity that the solver can use.
+    pub liquidity: Vec<Liquidity>,
```
bh2smith, line 67: Now that limit orders are separated from Liquidity, I was expecting `Auction` to have an `orders` field...

bh2smith (approved):
Just a bit concerned about what comes first #1068 or this? Since Auction contains Liquidity but not orders...

josojo (approved):

solver/src/solver.rs
nlordell: It does. I created the PRs out of order, but this one comes first.

PR #1072, opened 2021-09-02
[WIP] Switch orderbook to new metrics system

taminomara:
Note: some metric names have changed to follow prometheus naming conventions. Because of this, I want to deploy it to staging and create new grafana dashboards before releasing to prod.

### Test Plan
CI + manual testing


orderbook/src/api.rs
```diff
@@ -86,17 +70,41 @@ async fn handle_rejection(err: Rejection) -> Result<impl Reply, Infallible> {
...
+
+    let metrics = ApiMetrics::instance();
+    metrics
+        .requests_complete
+        .with_label_values(&[url, info.status().as_str()])
+        .inc();
+    metrics
+        .requests_duration_seconds
+        .with_label_values(&[url])
+        .observe(info.elapsed().as_secs_f64());
```
vkgnosis, line 84: We can't use the path as a log label because some routes have user data in the path. I think this is why we had LabelledRepy wrapper.

taminomara: Is it in the path or in the get parameters?

vkgnosis: https://protocol-rinkeby.dev.gnosisdev.com/api/ for example `/api/v1/orders/{UID}` is in the path . I would have preferred all parameters in the query but we decided that some routes would have some parameters in the path.

taminomara:
Will split this into smaller PRs.

PR #1073, opened 2021-09-02
pending_transaction support for Erigon

### LINKED ISSUE #1054: Make `pending_transactions` work with Erigon
Code pointer: https://github.com/gnosis/oba-services/blob/73cbfa407827a1b5a2c5e81481f219c77cc9062a/solver/src/pending_transactions.rs#L35-L47

Currently the method uses `eth_getBlockByNumber(pending)` to get a list of pending txs in the mempool, which - last time I checked - wasn't supported by Erigon.

Instead they told us to use the https://geth.ethereum.org/docs/rpc/ns-txpool API (supported by geth/erigon).

This task captures implementing logic to find pending transaction with this API and flip flop between the two based on node availability.

sunce86:
In order to get pending transaction from the mempool, pending_transactions() is now calling the `txpool_content` api instead of `eth_getBlockByNumber`

Both Erigon and Get have support for txpool_content.

### Test Plan
Tested against Erigon node using unit test.

### DIFF
--- a/solver/src/pending_transactions.rs
+++ b/solver/src/pending_transactions.rs
@@ -44,13 +44,17 @@ pub async fn pending_transactions(transport: &DynTransport) -> Result<Vec<Transa
     match transport
         .execute("txpool_content", Default::default())
         .await
-        .context("transport failed")
     {
         Ok(response) => {
             let txpool: TxPool = serde_json::from_value(response).context("deserialize failed")?;
 
-            Ok(txpool
+            let transactions = txpool
                 .pending
+                .into_iter()
+                .chain(txpool.queued)
+                .collect::<HashMap<String, HashMap<String, Transaction>>>();
+
+            Ok(transactions
                 .into_values()
                 .flatten()
                 .map(|value| value.1)
--- a/solver/src/pending_transactions.rs
+++ b/solver/src/pending_transactions.rs
@@ -2,8 +2,8 @@ use anyhow::{Context, Result};
 use ethcontract::transport::DynTransport;
 use primitive_types::{H160, U256};
 use serde::Deserialize;
-use web3::Transport;
 use std::collections::HashMap;
+use web3::Transport;
 
 #[derive(Debug, Deserialize)]
 struct TxPool {
@@ -46,9 +46,7 @@ pub async fn pending_transactions(transport: &DynTransport) -> Result<Vec<Transa
         .pending
         .into_values()
         .flatten()
-        .map(|value| {
-            value.1
-        })
+        .map(|value| value.1)
         .collect())
 }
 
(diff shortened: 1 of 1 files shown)


nlordell (commented):
Changes look good!

Seeing as (if I understood correctly in Slack) `txpool_content` is not supported by Infura Geth, would it possible to have two ways of getting pending transactions, preferring `txpool_content` but falling back to `get_blockByNumber("pending")` if it is not supported?

sunce86:
> Changes look good!
> 
> Seeing as (if I understood correctly in Slack) `txpool_content` is not supported by Infura Geth, would it possible to have two ways of getting pending transactions, preferring `txpool_content` but falling back to `get_blockByNumber("pending")` if it is not supported?

Added. Can you recheck please?

solver/src/pending_transactions.rs
```diff
@@ -32,18 +39,35 @@ pub enum Fee {
...
-    let params = vec!["pending".into(), true.into()];
-    let response = transport
-        .execute("eth_getBlockByNumber", params)
+    match transport
+        .execute("txpool_content", Default::default())
         .await
-        .context("transport failed")?;
-    let block: Block = serde_json::from_value(response).context("deserialize failed")?;
-    Ok(block.transactions)
+        .context("transport failed")
```
nlordell, line 47: ubernit: This `context` shouldn't be needed as the error gets ignored and dropped :point_down: anyway.

solver/src/pending_transactions.rs
```diff
@@ -32,18 +39,35 @@ pub enum Fee {
...
+    {
+        Ok(response) => {
+            let txpool: TxPool = serde_json::from_value(response).context("deserialize failed")?;
+
+            Ok(txpool
+                .pending
```
nlordell, line 53: Seeing as this is used for getting any submitted transaction from a previously executing solver, maybe it makes sense to return both `pending` and `queued` transactions? I was trying to understand exactly what the difference between these two was but it wasn't super obvious to me from the [documentation](https://geth.ethereum.org/docs/rpc/ns-txpool#txpool_content).

solver/src/pending_transactions.rs
sunce86: Fixed

solver/src/pending_transactions.rs
sunce86: The difference between pending and queued transaction from mempool is that pending transaction are ready to be included in a block, while queued transactions cannot be included in a block at the time, because for example, there might be a nonce gap between the queued transaction and last transaction sent from the same address, so they are queued for future inclusion.
Added merging of pending and queued transactions as suggested, to be sure that our transaction is not missed by any chance.

nlordell (approved):
Looks good to me!

vkgnosis (approved):

PR #1077, opened 2021-09-03
Pick default amount to estimate prices with based on network id

vkgnosis:
The previous default of 1 is problematic because it leads to bad or
failign price estimates. By having a built in default for known networks
we make running locally easier because we don't have to remember to set
this parameter.

### Test Plan
manual test

### DIFF
--- a/shared/src/arguments.rs
+++ b/shared/src/arguments.rs
@@ -92,14 +92,14 @@ pub struct Arguments {
     )]
     pub block_stream_poll_interval_seconds: Duration,
 
-    /// Default is one atom of native_token (i.e. price estimation uses spot prices).
+    /// The amount in native tokens atoms to use for price estimation. Should be reasonably large so
+    // that small pools do not influence the prices.
     #[structopt(
         long,
         env,
-        default_value = "1",
         parse(try_from_str = U256::from_dec_str)
     )]
-    pub amount_to_estimate_prices_with: U256,
+    pub amount_to_estimate_prices_with: Option<U256>,
 }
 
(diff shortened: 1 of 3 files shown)


shared/src/arguments.rs
```diff
@@ -120,3 +120,13 @@ pub fn wei_from_gwei(s: &str) -> anyhow::Result<f64> {
     let in_gwei: f64 = s.parse()?;
     Ok(in_gwei * 10e9)
 }
+
+pub fn default_amount_to_estimate_prices_with(network_id: &str) -> Option<U256> {
+    match network_id {
+        // Mainnet, Rinkeby
+        "1" | "4" => Some(10u128.pow(18).into()),
+        // Xdai
+        "100" => Some(10u128.pow(21).into()),
```
bh2smith, line 129: These amounts seem weird, you have 1*10^18 and 100*10^21 I think the 100 already make up for the +2 in the exponent.

bh2smith (approved):
~~Approving assuming the exponent is fixed.~~

Looks great!

shared/src/arguments.rs
vkgnosis: It shouldn't be multiplication unless I'm missing something. The 100 is just matching on the network id which is `100` for xdai. The reason for this amount is that 10e21 should be ~ 1000 usd on xdai.

bh2smith: Oh ha ha! Silly me. I just read it wrong.

nlordell (approved):

PR #1081, opened 2021-09-04
Add endpoint for user orders

vkgnosis:
Part of #969

As we talked about I went with limit and offset.
I decided to make the offset default to 0 and the limit default to 10. While I would like to make the parameters mandatory to make it clear to api users that this is a paginated api I also feel it is nice to have defaults when you quickly want to see a users most recent orders by hand.
The limit can be at most 1000.

### Test Plan
new unit test and short curl test for endpoint

### DIFF
--- a/orderbook/src/api/get_user_orders.rs
+++ b/orderbook/src/api/get_user_orders.rs
@@ -1,0 +1,104 @@
+use crate::{api::internal_error, orderbook::Orderbook};
+use anyhow::Result;
+use model::order::Order;
+use serde::Deserialize;
+use shared::H160Wrapper;
+use std::{convert::Infallible, sync::Arc};
+use warp::{
+    hyper::StatusCode,
+    reply::{self, with_status, Json, WithStatus},
+    Filter, Rejection, Reply,
+};
+
+#[derive(Clone, Copy, Debug, Deserialize)]
+struct Query {
+    offset: Option<u64>,
+    limit: Option<u64>,
+}
+
+fn request() -> impl Filter<Extract = (H160Wrapper, Query), Error = Rejection> + Clone {
+    warp::path!("account" / H160Wrapper / "orders")
+        .and(warp::get())
+        .and(warp::query::<Query>())
+}
+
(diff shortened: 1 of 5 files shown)


orderbook/src/api/get_user_orders.rs
```diff
@@ -0,0 +1,103 @@
...
+use warp::{
+    hyper::StatusCode,
+    reply::{self, with_status, Json, WithStatus},
+    Filter, Rejection, Reply,
+};
+
+#[derive(Clone, Copy, Debug, Deserialize)]
+struct Query {
+    offset: Option<u64>,
+    limit: Option<u64>,
```
nlordell, line 16: Maybe we should enforce `limit` to be non-zero (as fetching 0 orders doesn't really make sense):
```suggestion
    limit: Option<NonZeroU64>,
```

Just a suggestion, and not sure we actually need to, feel free to disregard.

nlordell (approved):

bh2smith (approved):

orderbook/src/api/get_user_orders.rs
```diff
@@ -0,0 +1,103 @@
...
+
+pub fn get_user_orders(
+    orderbook: Arc<Orderbook>,
+) -> impl Filter<Extract = (impl Reply,), Error = Rejection> + Clone {
+    request().and_then(move |owner: H160Wrapper, query: Query| {
+        let orderbook = orderbook.clone();
+        async move {
+            const DEFAULT_OFFSET: u64 = 0;
+            const DEFAULT_LIMIT: u64 = 10;
+            const MAX_LIMIT: u64 = 1000;
```
alfetopito, line 43: As Nick suggested, I'm also in favor of a MIN_LIMIT = 1

alfetopito (approved):

orderbook/src/api/get_user_orders.rs
vkgnosis: Added check for this. Was in fact needed because a limit 0 would be unlimited again in Postgres.

PR #1083, opened 2021-09-05
Update error message for smart contract receiving eth

fedgiac:
Now that we accept orders from a smart contract, the check `TransferEthToContract` might be triggered even if the receiver is not set. This PR slightly rephrases the error message.

### DIFF
--- a/orderbook/src/api/create_order.rs
+++ b/orderbook/src/api/create_order.rs
@@ -81,8 +81,8 @@ pub fn create_order_response(result: Result<AddOrderResult>) -> impl Reply {
         Ok(AddOrderResult::TransferEthToContract) => (
             super::error(
                 "TransferEthToContract",
-                "Setting receiver to a smart contract wallet when buying Ether \
-                 is currently not supported",
+                "Sending Ether to a smart contract wallets is currently not \
+                 supported",
             ),
             StatusCode::BAD_REQUEST,
         ),


orderbook/src/api/create_order.rs
```diff
@@ -81,8 +81,8 @@ pub fn create_order_response(result: Result<AddOrderResult>) -> impl Reply {
         Ok(AddOrderResult::TransferEthToContract) => (
             super::error(
                 "TransferEthToContract",
-                "Setting receiver to a smart contract wallet when buying Ether \
-                 is currently not supported",
+                "Sending Ether to a smart contract wallets is currently not \
+                 supported",
```
bh2smith, line 85: Is it ever going to be supported?

bh2smith (approved):

orderbook/src/api/create_order.rs
fedgiac: I think so, it's technically possible by including the SC wallet address in the access list.

nlordell: It is possible. I had a tenderly simulation with access list set where I successfully transferred Eth to Gnosis Safe somewhere. We just need to add support for EIP-2930 access lists and set the appropriately when sending Eth to SCs.

nlordell (approved):

PR #1084, opened 2021-09-06
Stable Pool Amplification Parameter Precision

bh2smith:
Although the precision is a hard coded value (in the stable pool contract), this could potentially change in the future. So we keep this value around and divide at fetching time. This changed what was U256 before to BigRational, which is parsed as decimal when forwarded to http solver.

### Test Plan
Updated tests involving stable pools.

### DIFF
--- a/shared/src/sources/balancer/info_fetching.rs
+++ b/shared/src/sources/balancer/info_fetching.rs
@@ -111,7 +111,7 @@ impl PoolInfoFetching for PoolInfoFetcher {
         let tokens = token_data.await?.0;
         let scaling_exponents = self.get_scaling_exponents(&tokens).await?;
         let amplification_data = amplification_parameter.await?;
-                let amplification_parameter = BigRational::new(
+        let amplification_parameter = BigRational::new(
             amplification_data.0.to_big_int(),
             amplification_data.2.to_big_int(),
         );
--- a/shared/src/sources/balancer/info_fetching.rs
+++ b/shared/src/sources/balancer/info_fetching.rs
@@ -1,13 +1,14 @@
 //! Responsible for conversion of a `pool_address` into `WeightedPoolInfo` which is used by the
 //! event handler to construct a `RegisteredWeightedPool`.
 use crate::{
-    sources::balancer::swap::fixed_point::Bfp, sources::uniswap::pool_fetching::MAX_BATCH_SIZE,
-    token_info::TokenInfoFetching, Web3,
+    conversions::U256Ext, sources::balancer::swap::fixed_point::Bfp,
+    sources::uniswap::pool_fetching::MAX_BATCH_SIZE, token_info::TokenInfoFetching, Web3,
 };
 use anyhow::{anyhow, Result};
 use contracts::{BalancerV2StablePool, BalancerV2Vault, BalancerV2WeightedPool};
-use ethcontract::{batch::CallBatch, Bytes, H160, H256, U256};
+use ethcontract::{batch::CallBatch, Bytes, H160, H256};
 use mockall::*;
+use num::BigRational;
 use std::sync::Arc;
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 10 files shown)


marcovc (approved):
Thanks!

shared/src/sources/balancer/info_fetching.rs
```diff
@@ -109,13 +110,16 @@ impl PoolInfoFetching for PoolInfoFetcher {
 
         let tokens = token_data.await?.0;
         let scaling_exponents = self.get_scaling_exponents(&tokens).await?;
+        let amplification_data = amplification_parameter.await?;
+        let amplification_parameter =
+            amplification_data.0.to_big_rational() / amplification_data.2.to_big_rational();
```
vkgnosis, line 115: Nit: Use the constructor with numerator and denominator directly
```suggestion
                let amplification_parameter = BigRational::new(
            amplification_data.0.to_big_int(),
            amplification_data.2.to_big_int(),
        );
```

solver/src/solver/http_solver/model.rs
```diff
@@ -94,12 +94,12 @@ pub struct WeightedProductPoolParameters {
...
 #[serde_as]
-#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct StablePoolParameters {
     #[serde_as(as = "HashMap<_, DecimalU256>")]
     pub reserves: HashMap<H160, U256>,
-    #[serde(with = "u256_decimal")]
-    pub amplification_parameter: U256,
+    #[serde(with = "ratio_as_decimal")]
+    pub amplification_parameter: BigRational,
```
vkgnosis, line 102: Should we pass numbers that are fractions as fractions to the solver instead of converting to decimal representation? For numbers like 1/3 there is no way to convert them to decimal in a lossless way. This is probably not the only place we could consider this.
Not important for this PR just an idea.

vkgnosis (approved):

solver/src/solver/http_solver/model.rs
bh2smith: Yea we should probably do this although I was trying to mimic what we are doing elsewhere for this PR.

marcovc: @vkgnosis I agree. That would already be possible with quasimodo, but would need to check if the MIP solver would handle it. Definitely a bigger change than this PR.

bh2smith:
Note that this is being merged into the revert of the revert branch forwarding stable pools to http solver.

shared/src/sources/balancer/info_fetching.rs
```diff
@@ -109,13 +110,18 @@ impl PoolInfoFetching for PoolInfoFetcher {
 
         let tokens = token_data.await?.0;
         let scaling_exponents = self.get_scaling_exponents(&tokens).await?;
+        let amplification_data = amplification_parameter.await?;
```
nlordell, line 113: Note that you can also destructure the tuple here:
```suggestion
        let (amplification_factor, _, precision) = amplification_parameter.await?;
```

Not a suggestion for change, (I'm fine how it is).

bh2smith: Actually I like this better and will include it before merge.

nlordell (approved):

shared/src/sources/balancer/pool_cache.rs
```diff
@@ -201,7 +202,10 @@ fn handle_results(results: Vec<FetchedBalancerPool>) -> Result<Vec<BalancerPool>
                             .expect("Stable pools must have this set."),
                     )? {
                         // This is the ratio of amplification_parameter / precision.
-                        Some(state) => state.0.to_big_rational() / state.2.to_big_rational(),
+                        Some((amplification_factor, _, precision)) => BigRational::new(
```
nlordell, line 205: Maybe this is something else that we can address in the refactor and allow `get_stable_pool_data` to be reused here, instead of duplicating logic for computing the amplification parameter.

bh2smith: Yes, I was actually a bit confused as to why there were two places that need to be changed here. Will add it to the list.

PR #1092, opened 2021-09-07
Remove automatic balance updating from order book

vkgnosis:
First part of moving to a permanent solvable orders + balances cache in one component.
While I expect it to be functional this PR isn't meant to be merged standalone as it removes all caching. I split it out for easier review to not have the many changes to account_balances and the new component in one PR.
This PR removes all caching functionality from the balance fetcher keeping only the raw operations.
The second PR will introduce the new component that keeps track of solvable orders and related balances.

Part of https://github.com/gnosis/gp-v2-services/issues/967

### Test Plan
existing tests but also not meant to be merged alone. ran the ignored balance fetcher tests.

### DIFF
--- a/orderbook/src/account_balances.rs
+++ b/orderbook/src/account_balances.rs
@@ -1,36 +1,35 @@
-use anyhow::{bail, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use contracts::{BalancerV2Vault, ERC20};
 use ethcontract::{batch::CallBatch, Account};
-use futures::{
-    future::{self, join_all, BoxFuture},
-    FutureExt as _,
-};
-use model::order::SellTokenSource;
+use futures::{FutureExt, StreamExt};
+use model::order::{Order, SellTokenSource};
 use primitive_types::{H160, U256};
 use shared::{Web3, Web3Transport};
-use std::{
-    collections::HashMap,
-    sync::{Arc, Mutex},
-};
+use std::future::Future;
 use web3::types::{BlockId, BlockNumber, CallRequest};
 
-const MAX_BATCH_SIZE: usize = 100;
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
+pub struct Query {
+    pub owner: H160,
+    pub token: H160,
+    pub source: SellTokenSource,
+}
+
+impl Query {
+    pub fn from_order(o: &Order) -> Self {
+        Self {
+            owner: o.order_meta_data.owner,
+            token: o.order_creation.sell_token,
+            source: o.order_creation.sell_token_balance,
+        }
+    }
+}
 
 #[cfg_attr(test, mockall::automock)]
 #[async_trait::async_trait]
 pub trait BalanceFetching: Send + Sync {
-    // Register owner and address for balance updates in the background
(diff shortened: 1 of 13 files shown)


orderbook/src/account_balances.rs
```diff
@@ -178,116 +109,73 @@ impl Web3BalanceFetcher {
...
-                tracing::warn!(
-                    "Couldn't fetch vault approval from {:?} to {:?}",
-                    owner,
-                    relayer
-                );
-                SubscriptionValue::default()
-            }
-        }
+        Ok(match approval.await.context("allowance")? {
+            true => balance.await.context("balance")?,
```
nlordell, line 138: I think this may be slightly wrong. If you have vault relayer approval, then the balance is:
`cmp::min(token.balance_of(owner), token.allowance(owner, vault))`

So I think it should be:
```
let balance = erc20_balance_query(batch, ...);
```

orderbook/src/account_balances.rs
```diff
@@ -605,35 +465,31 @@ mod tests {
...
             .mint(trader.address(), 100.into())
             .send()
             .await
             .unwrap();
-        fetcher.update().await;
-        assert_eq!(
-            fetcher.get_balance(trader.address(), token.address(), SellTokenSource::External),
-            Some(U256::zero()),
-        );
+        assert_eq!(get_balance().await, U256::zero(),);
```
nlordell, line 492: ubernit: Here and in a few more places:

```suggestion
        assert_eq!(get_balance().await, U256::zero());
```

orderbook/src/orderbook.rs
```diff
@@ -323,40 +301,20 @@ async fn filter_unsupported_tokens(
...
-        );
-        match fetcher.get_balance(key.0, key.1, key.2) {
-            Some(balance) => {
-                balances.insert(key, balance);
-            }
-            None => {
-                untracked.insert(key);
-            }
+        let query = crate::account_balances::Query::from_order(order);
+        if let Ok(balance) = fetcher.get_balances(&[query]).await[0] {
```
nlordell, line 314: Assuming this is just temporary, and eventually you will want to build a `Vec<Query>` for all eligible orders to retrieve balances for?

bh2smith (approved):
Looks good with the exception of @nlordell suggestion about the vault balances. Also, with this change as a standalone, but temporary fix, does this mean we will be actively fetching balances at solve time instead of caching? Should we expect that staging would then be temporarily inefficient until the rest comes along?

orderbook/src/account_balances.rs
vkgnosis: Good point. Will fix and see if I can find a good way to add a test for it.

orderbook/src/orderbook.rs
vkgnosis: Yes that happens in the next PR.

vkgnosis:
> does this mean we will be actively fetching balances at solve time instead of caching

This PR doesn't influence solver code. It does making fetching the solvable orders slower. I think this would still be fine but anyway I don't intend to merge this without the second PR.

nlordell:
> I think this would still be fine but anyway

I'm also not concerned if staging is a bit slower for a bit.

nlordell (approved):

PR #1093, opened 2021-09-07
Use solvable orders and balance cache in one component

vkgnosis:
Second part that adds the new component which keeps track of solvable orders and related balances.
Part of #967

### Test Plan
new unit tests, TODO manual test

### DIFF
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -2,8 +2,9 @@ use crate::{
     account_balances::BalanceFetching,
     database::orders::{InsertionError, OrderFilter, OrderStoring},
     fee::{EthAwareMinFeeCalculator, MinFeeCalculating},
+    solvable_orders::SolvableOrdersCache,
 };
-use anyhow::Result;
+use anyhow::{ensure, Result};
 use chrono::Utc;
 use contracts::WETH9;
 use model::{
@@ -18,11 +19,8 @@ use model::{
     DomainSeparator,
 };
 use primitive_types::{H160, U256};
-use shared::{
-    bad_token::BadTokenDetecting, metrics::LivenessChecking, time::now_in_epoch_seconds,
-    web3_traits::CodeFetching,
-};
-use std::{collections::HashMap, sync::Arc, time::Duration};
+use shared::{bad_token::BadTokenDetecting, metrics::LivenessChecking, web3_traits::CodeFetching};
+use std::{collections::HashSet, sync::Arc, time::Duration};
 
 #[derive(Debug, Eq, PartialEq)]
 pub enum AddOrderResult {
@@ -59,7 +57,7 @@ pub struct Orderbook {
     domain_separator: DomainSeparator,
     settlement_contract: H160,
     database: Arc<dyn OrderStoring>,
-    balance_fetcher: Box<dyn BalanceFetching>,
+    balance_fetcher: Arc<dyn BalanceFetching>,
     fee_validator: Arc<EthAwareMinFeeCalculator>,
     min_order_validity_period: Duration,
     bad_token_detector: Arc<dyn BadTokenDetecting>,
@@ -67,6 +65,7 @@ pub struct Orderbook {
     native_token: WETH9,
     banned_users: Vec<H160>,
     enable_presign_orders: bool,
+    solvable_orders: Arc<SolvableOrdersCache>,
 }
 
 impl Orderbook {
@@ -75,7 +74,7 @@ impl Orderbook {
         domain_separator: DomainSeparator,
         settlement_contract: H160,
         database: Arc<dyn OrderStoring>,
-        balance_fetcher: Box<dyn BalanceFetching>,
+        balance_fetcher: Arc<dyn BalanceFetching>,
         fee_validator: Arc<EthAwareMinFeeCalculator>,
         min_order_validity_period: Duration,
         bad_token_detector: Arc<dyn BadTokenDetecting>,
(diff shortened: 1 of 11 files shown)


orderbook/src/orderbook.rs
```diff
@@ -301,84 +310,13 @@ async fn filter_unsupported_tokens(
...
     }
 }
 
-// The order book has to make a choice for which orders to include when a user has multiple orders
-// selling the same token but not enough balance for all of them.
-// Assumes balance fetcher is already tracking all balances.
-fn solvable_orders(
-    mut orders: Vec<Order>,
-    balances: &HashMap<(H160, H160, SellTokenSource), U256>,
-) -> Vec<Order> {
```
vkgnosis, line 341: Moved into solvable_orders.rs

orderbook/src/solvable_orders.rs
```diff
@@ -0,0 +1,379 @@
...
+        }
+    }
+    let missing_queries = Vec::from_iter(missing_queries);
+    (new_balances, missing_queries)
+}
+
+// The order book has to make a choice for which orders to include when a user has multiple orders
+// selling the same token but not enough balance for all of them.
+// Assumes balance fetcher is already tracking all balances.
+fn solvable_orders(mut orders: Vec<Order>, balances: &Balances) -> Vec<Order> {
```
vkgnosis, line 159: Moved from orderbook.rs

orderbook/src/solvable_orders.rs
```diff
@@ -0,0 +1,379 @@
...
+    cache: Mutex<Inner>,
+}
+
+type Balances = HashMap<Query, U256>;
+
+struct Inner {
+    orders: Vec<Order>,
+    balances: Balances,
+    block: u64,
+    update_time: Instant,
```
bh2smith, line 41: Would it be more appropriate to use `update_block` instead of a time? I ask because I imagine we would be comparing a block number with the last update to determine if an update it necessary.

orderbook/src/solvable_orders.rs
```diff
@@ -0,0 +1,379 @@
...
+        let mut remaining_balance = match balances.get(&key) {
+            Some(balance) => *balance,
+            None => continue,
+        };
+        for order in orders {
+            // TODO: This is overly pessimistic for partially filled orders where the needed balance
+            // is lower. For partially fillable orders that cannot be fully filled because of the
+            // balance we could also give them as much balance as possible instead of skipping. For
+            // that we first need a way to communicate this to the solver. We could repurpose
+            // availableBalance for this.
```
bh2smith, line 178: Since this comment is related to the "epic" task of supporting partially fillable orders perhaps we should directly link it to the issue we have for it.

https://github.com/gnosis/gp-v2-services/issues/673

(or vice versa).

bh2smith (approved):
Nice work here! Excited to see it in action.

orderbook/src/solvable_orders.rs
vkgnosis: It uses to block whether it should requery  balances. The update time is used in the orderbook to figure out when we should enter an error state if the solvable orders haven't been updated in X seconds real time.

orderbook/src/solvable_orders.rs
vkgnosis: When I merge this I'll add a comment to the issue.

orderbook/src/solvable_orders.rs
```diff
@@ -0,0 +1,379 @@
...
+
+    /// The cache will update the solvable orders and missing balances as soon as possible.
+    pub fn request_update(&self) {
+        self.notify.notify_one();
+    }
+
+    /// Manually update solvable orders. Usually called by the background updating task.
+    pub async fn update(&self, block: u64) -> Result<()> {
+        let min_valid_to = now_in_epoch_seconds() + self.min_order_validity_period.as_secs() as u32;
+        let orders = self.database.solvable_orders(min_valid_to).await?;
```
nlordell, line 88: One other approach I could think of, instead of querying solvable orders, is to use the order `valid_to` for ignoring orders to query balances for.

This way, we could have kept things a little closer to the old model where we:
- "registered" an order for balance updates on creation (registering all solvable orders at startup).
- Only updated orders that are still valid on a new block
- On reorgs (i.e. if the time goes isn't monotonically increasing), we could just re-register all solvable orders.

orderbook/src/orderbook.rs
```diff
@@ -195,6 +196,7 @@ impl Orderbook {
             Err(InsertionError::DbError(err)) => return Err(err.into()),
             _ => (),
         }
+        self.solvable_orders.request_update();
```
nlordell, line 199: I have a question about this particular line, in combination with L173-182. Specifically, the `can_transfer` check would query the node for balances, and then the `solvable_orders` cache would re-query it when the update here is requested?

orderbook/src/solvable_orders.rs
vkgnosis: I had the idea too that we could manually add newly created orders instead of fetching from the database but I like this approach because it is stateless. So even if for an unexpected reason the set of solvable orders changes in the database we will notice.

orderbook/src/orderbook.rs
vkgnosis: can_transfer and get_balance are different checks. Can transfer simulates a full transaction from the user to the settlement contract. On the other hand get_balance only does view calls to get state. I believe we added can_transfer as an additional safety measure before we had the bad token detector so not sure if it is still needed.
In the old code it was also the case that can_transfer only happens once on order creation while later we only check get_balance. So this behavior hasn't changed. We could start caching can_transfer too but I not sure it's needed and would do that in a separate PR.

nlordell: Right - for some reason I confused the two. Thanks for clarifying!

nlordell (approved):

PR #1094, opened 2021-09-07
Refactor Balancer Part I - Split PoolFetching

bh2smith:
This is intended to be an overall refactor of the Balancer module to make it easier for integrating different liquidity sources. We start here by separating the pool fetcher for stable and weighted pools.

- [ ] CacheKey should be more generic...

In a following PR we plan to remove the unnecessary enum `BalancerOrder`, split up the pool_cache file into a sub module and document all the code to include StablePools

### Test Plan
No logic changes, existing CI.

### DIFF
--- a/shared/src/sources/balancer/pool_fetching.rs
+++ b/shared/src/sources/balancer/pool_fetching.rs
@@ -167,12 +167,12 @@ impl FetchedBalancerPools {
         tokens.extend(
             self.stable_pools
                 .iter()
-                .flat_map(|pool| pool.reserves.keys().copied().collect::<HashSet<_>>()),
+                .flat_map(|pool| pool.reserves.keys().copied()),
         );
         tokens.extend(
             self.weighted_pools
                 .iter()
-                .flat_map(|pool| pool.reserves.keys().copied().collect::<HashSet<_>>()),
+                .flat_map(|pool| pool.reserves.keys().copied()),
         );
         tokens
     }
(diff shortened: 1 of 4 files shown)


shared/src/sources/balancer/pool_fetching.rs
```diff
@@ -228,29 +138,61 @@ impl StablePool {
...
+    pub weighted_pools: Vec<WeightedPool>,
+}
+
+impl FetchedBalancerPools {
+    pub fn relevant_tokens(&self) -> HashSet<H160> {
+        let mut tokens = HashSet::new();
+        tokens.extend(
+            self.stable_pools
+                .iter()
+                .flat_map(|pool| pool.reserves.keys().copied().collect::<HashSet<_>>()),
```
vkgnosis, line 170: No need to collect into a temporary HashSet first
```suggestion
                .flat_map(|pool| pool.reserves.keys().copied()),
```
also 5 lines below

vkgnosis (approved):

shared/src/sources/balancer/pool_fetching.rs
bh2smith: Thanks! Someday I will get the hang of when this is or is not needed.

vkgnosis: The reason I *think* it is not needed is that we are collecting into a big hash set already. So any duplicates that the intermediate hash set would filter out would also be filtered out by the outer hash set.

nlordell (approved):

PR #1095, opened 2021-09-08
Pin nightly version to un-break coverage

nlordell:
This PR just pins the nightly version to September 2nd, as that was the last working version before code coverage started failing again.

We should be able to revert this change in the near future once the issue works itself out.

### Test Plan

See a green :heavy_check_mark: beside the coverage CI job.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -64,7 +64,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install Rust
-        run: rustup toolchain install nightly --component llvm-tools-preview
+        run: rustup toolchain install nightly-2021-09-02 --component llvm-tools-preview
       - name: Install cargo-llvm-cov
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):

vkgnosis:
I also see
          fail_ci_if_error: true
Maybe setting it to false is all that we need? Would be nicer to have nightly auto update and just not get coverage while its broken.

nlordell:
From what I understand, the `fail_ci_if_error: true` just makes the job fail if the there is an issue uploading the report. Right now, the job is failing because of the `Generate code coverage` step so I'm not sure if it would help.

bh2smith:
Give that a coverage report has been included here, doesn't this mean it's working...?

bh2smith (approved):
Appears to be working. At least from what I can see on mobile app.

nlordell:
It is, but I have a better fix in #1096

PR #1096, opened 2021-09-08
Cargo update

nlordell:
This PR fixes coverage! Supercedes #1095 

In doing a little bit of digging, I discovered that some nightly integer math feature changed some interface which broke nightly builds for the `num-*` crates. Specifically, `{integer}.div_ceil` now no longer takes a reference. Because of this, coverage started failing. However, since then, new versions of these crates that fix nightly builds have been released, so coverage just needed a `cargo update` to fix.

### Test Plan

CI - no code changes, just updates to transient dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,9 +14,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.15.2"
+version = "0.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


bh2smith (approved):
Looks good but why only lock files and no toml?

vkgnosis (approved):

nlordell:
> Looks good but why only lock files and no toml?

Only transitive dependencies were updated (i.e. dependencies of my dependencies). They only appear in the lock file

PR #1097, opened 2021-09-08
Make recover_gas_price_from_pending_transaction optional

fleupold:
If we want to submit solutions via the private taichi network, we cannot inspect the pending txs from the mempool (because they are private). They are using erigon and have the `txpool` RPC module disabled. This PR makes it so that failure to fetch the pending gas price doesn't lead to a failure to process the batch.

Alternatively, we could create a taichi solution submission strategy, however all the other aspects would be the same as in PublicMempool.

### Test Plan
Submitted a tx via https://api.taichi.network:10001/rpc/private

https://etherscan.io/tx/0x7373c59631328c32163db76d7336abf9563c935a7bc8115adf36706c3ce7eb37

### DIFF
--- a/solver/src/settlement_submission/public_mempool.rs
+++ b/solver/src/settlement_submission/public_mempool.rs
@@ -34,7 +34,13 @@ pub async fn submit(
         .context("failed to get transaction_count")?;
     let pending_gas_price = recover_gas_price_from_pending_transaction(&web3, &address, nonce)
         .await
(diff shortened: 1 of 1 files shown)


nlordell (approved):

vkgnosis (approved):

bh2smith (approved):
Looks good although I only see a change that returns none. Was there supposed to be some form that is actually now an option or was this already the case.

PR #1098, opened 2021-09-08
Support Erigon Batch RPC responses

fleupold:
Another requirement for using TaiChi solution submission (as their client is Erigon). I also reached out the team here: https://discord.com/channels/687972960811745322/789381905958699018/885124258739675146

Erigon seems to not confirm with the RPC standard and returns us responses in the form

> ["{\"jsonrpc\":\"2.0\",\"id\":\"1\",\"result\":\"erigon/2021.08.4/linux-amd64/go1.16.7\"}\n","{\"jsonrpc\":\"2.0\",\"id\":\"2\",\"result\":\"erigon/2021.08.4/linux-amd64/go1.16.7\"}\n"]

rather than

> [{\"jsonrpc\":\"2.0\",\"id\":\"1\",\"result\":\"erigon/2021.08.4/linux-amd64/go1.16.7\"}\n,{\"jsonrpc\":\"2.0\",\"id\":\"2\",\"result\":\"erigon/2021.08.4/linux-amd64/go1.16.7\"}\n]

(note that above elements are strings, where below they are objects). I believe the latter is correct according to the standard: https://www.jsonrpc.org/specification#batch

While fixing this in our code is likely not the best place I think it's the only practical way of getting this in quickly (and hopefully remove once erigon is fixed)

### Test Plan
Added a unit test

### DIFF
--- a/shared/src/transport/http.rs
+++ b/shared/src/transport/http.rs
@@ -5,6 +5,7 @@ use jsonrpc_core::types::{Call, Output, Request, Value};
 use prometheus::Registry;
 use reqwest::{Client, Url};
 use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
 use std::{
     collections::HashMap,
     sync::{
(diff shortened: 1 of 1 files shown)


shared/src/transport/http.rs
```diff
@@ -125,9 +126,26 @@ impl BatchTransport for HttpTransport {
...
+#[serde(untagged)]
+pub enum OutputOrString {
+    String(String),
+    Output(Output),
+}
+
+impl From<OutputOrString> for Output {
+    fn from(output_or_string: OutputOrString) -> Self {
+        match output_or_string {
+            OutputOrString::String(string) => jsonrpc_core::serde_from_str(&string).unwrap(),
```
vkgnosis, line 140: We shouldn't be unwrapping on data coming from outside. I think this would panic if the string is invalid json.

shared/src/transport/http.rs
```diff
@@ -125,9 +126,26 @@ impl BatchTransport for HttpTransport {
     }
 }
 
+/// Workaround for Erigon nodes, which encode each element of the Batch Response as a String rather than a deserializable JSON object
+#[derive(Debug, PartialEq, Clone, Deserialize, Serialize)]
+#[serde(untagged)]
+pub enum OutputOrString {
```
nlordell, line 132: nit: Does this need to be public?

shared/src/transport/http.rs
nlordell: Agreed. If you want to use conversion traits, you can also use the `TryFrom`. Although, because this is an internal type, I would just recommend a `try_into_output` member function.

nlordell: Changed to return an error and error the whole batch. The motivation behind erroring the whole batch is that it may be possible to not be able to line up message IDs with response IDs if it fails to parse string outputs for whatever reason.

nlordell (approved):

vkgnosis (approved):

PR #1100, opened 2021-09-08
Get rid of lazy statics in metrics, use new derive macro

taminomara:
Further improve metric ergonomics by switching to new derive macro.

### Test Plan
CI, manual testing

### DIFF
--- a/shared/Cargo.toml
+++ b/shared/Cargo.toml
@@ -29,6 +29,7 @@ num-bigint = "0.3"
 once_cell = "1.8.0"
 primitive-types = "0.9"
 prometheus = "0.12"
+prometheus-metric-storage = "0.2"
 reqwest = { version = "0.11", features = ["json"] }
 scopeguard = "1.1.0"
 serde = "1.0"
--- a/shared/Cargo.toml
+++ b/shared/Cargo.toml
@@ -29,7 +29,7 @@ num-bigint = "0.3"
 once_cell = "1.8.0"
 primitive-types = "0.9"
 prometheus = "0.12"
-prometheus-metric-storage = "0.2"
+prometheus-metric-storage = "0.3"
 reqwest = { version = "0.11", features = ["json"] }
 scopeguard = "1.1.0"
 serde = "1.0"
--- a/shared/src/transport/http.rs
+++ b/shared/src/transport/http.rs
@@ -1,10 +1,10 @@
-use crate::metrics::get_metrics_registry;
+use crate::metrics::get_metric_storage_registry;
 use ethcontract::jsonrpc as jsonrpc_core;
 use futures::{future::BoxFuture, FutureExt};
 use jsonrpc_core::types::{Call, Output, Request, Value};
-use prometheus::Registry;
 use reqwest::{Client, Url};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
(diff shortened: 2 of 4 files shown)


shared/Cargo.toml
```diff
@@ -29,6 +29,7 @@ num-bigint = "0.3"
 once_cell = "1.8.0"
 primitive-types = "0.9"
 prometheus = "0.12"
+prometheus-metric-storage = "0.2"
```
vkgnosis, line 32: I see this is your crate, nice.

shared/src/transport/http.rs
```diff
@@ -191,53 +206,23 @@ fn method_name(call: &Call) -> &str {
     }
 }
 
+#[derive(prometheus_metric_storage::MetricStorage, Clone, Debug)]
+#[metric(subsystem = "node_transport")]
 struct TransportMetrics {
+    /// Number of inflight RPC requests for ethereum node.
+    #[metric(labels("method"))]
```
vkgnosis, line 213: I couldn't find documentation for the MetricStorage derive in https://docs.rs/prometheus-metric-storage/0.2.0/prometheus_metric_storage/ and the metric attribute. Would be nice to have that like how other crates document macros.

taminomara: Yeah, documentation is in progress.

shared/Cargo.toml
taminomara: If you could take a look at the code, I'd appreciate it: https://github.com/taminomara/prometheus-metric-storage

taminomara: Specifically I'd like to discuss implementation of this function: https://github.com/taminomara/prometheus-metric-storage/blob/main/prometheus-metric-storage/src/lib.rs#L120

I don't know how to do this without an unsafe block.

We could change signature to return `Result<T>` (so, no reference), but then we'll have to clone a bunch of metric storages and save them in our structs. For example, [this field here](https://github.com/gnosis/gp-v2-services/pull/1100/files#diff-7624483e59540f6a621fdf505941ddab387cf0b3fd3673dceb537fddc5411380R26) will have type `TransportMetrics` instead of `&'static TransportMetrics`. And then we'll have to clone metrics [here](https://github.com/gnosis/gp-v2-services/pull/1100/files#diff-7624483e59540f6a621fdf505941ddab387cf0b3fd3673dceb537fddc5411380R131) to avoid capturing `self` into the async block. Overall, this will cause lots of small inconveniences.

shared/src/transport/http.rs
```diff
@@ -46,6 +47,14 @@ impl HttpTransport {
     }
 }
 
+impl Debug for HttpTransport {
```
nlordell, line 50: Good stuff, I think this is a more appropriate `Debug` implementation than the default one.

nlordell (approved):

shared/src/transport/http.rs
bh2smith: I actually noticed an ignored test for a debug implementation in model/second/order.rs yesterday that causes us to lose coverage there. Maybe if it were as pretty as this we could remove the ignore there.

vkgnosis (approved):
Approving after seeing the documentation PR.
Would still be nice to not have unsafe but I think your reasoning is good.

PR #1101, opened 2021-09-08
API metrics

taminomara:
Re-implement API metrics using the new metric storage

### Test Plan
CI and manual testing

### DIFF
--- a/orderbook/src/api.rs
+++ b/orderbook/src/api.rs
@@ -63,7 +63,7 @@ pub fn handle_all_routes(
             .or(get_fee_and_quote_sell.with(handle_metrics("get_fee_and_quote_sell")))
             .or(get_fee_and_quote_buy.with(handle_metrics("get_fee_and_quote_buy")))
             .or(get_user_orders.with(handle_metrics("get_user_orders")))
-            .or(post_quote.with(handle_metrics("get_user_orders")))
+            .or(post_quote.with(handle_metrics("get_user_orders"))),
     );
 
     routes_with_labels.recover(handle_rejection).with(cors)
--- a/orderbook/src/api.rs
+++ b/orderbook/src/api.rs
@@ -96,7 +96,6 @@ struct ApiMetrics {
     requests_complete: prometheus::CounterVec,
 
     /// Execution time for each API request.
-    #[metric(labels())]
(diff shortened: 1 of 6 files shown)


orderbook/src/api.rs
```diff
@@ -41,66 +36,62 @@ pub fn handle_all_routes(
...
-                .map(|reply| LabelledReply::new(reply, "get_fee_and_quote_sell")))
-            .unify()
-            .or(get_fee_and_quote_buy
-                .map(|reply| LabelledReply::new(reply, "get_fee_and_quote_buy")))
-            .unify()
-            .or(get_user_orders.map(|reply| LabelledReply::new(reply, "get_user_orders")))
-            .unify(),
+        (create_order.with(handle_metrics("create_order")))
+            .or(get_orders.with(handle_metrics("get_orders")))
+            .or(fee_info.with(handle_metrics("fee_info")))
```
bh2smith, line 52: This looks much nicer, could you elaborate on how the api metrics change is related to this. Or even why we no longer need "unify"?

orderbook/src/api.rs
```diff
@@ -41,66 +36,62 @@ pub fn handle_all_routes(
...
+#[metric(subsystem = "api", labels("endpoint"))]
+struct ApiMetrics {
+    /// Number of completed API requests.
+    #[metric(labels("status_code"))]
+    requests_complete: prometheus::CounterVec,
+
+    /// Execution time for each API request.
+    #[metric(labels())]
+    requests_duration_seconds: prometheus::Histogram,
 }
```
bh2smith, line 95: I see that this parameter `metrics: Arc<Metrics>,` was removed from several places, but don't see anything put back to replace it. So at this point I can only assume that this large green block is compensating for all of that. Could you please explain how the code here is instantiated in places where metrics was removed?

bh2smith (approved):

orderbook/src/api.rs
taminomara: So we've switched to the global registry that's available through the `get_metric_storage_registry` function. So now we don't have to explicitly pass registry around.

nlordell (approved):

nlordell:
The proc macro looks very nice to use.

vkgnosis (approved):

PR #1109, opened 2021-09-09
Improve PriceEstimation trait in preparation for more implementations

vkgnosis:
like for Paraswap.

Main points of this change:
1. Previously we calculated the out amount internally but always
   returned the price as BigRational. This made the trait less than
   flexible than it has to be when consumers need to turn the price back
   into an amount.
2. We had too many helper functions inside of the trait as default
   implementations. These make more sense either as external functions
   or as direct implementations in their callers. The new api is smaller
   and more flexible.

### Test Plan
adapted unit tests (which is about half of the changed lines). will make manual test to see that api response between this PR and staging is still same.

### DIFF
--- a/shared/src/price_estimate.rs
+++ b/shared/src/price_estimate.rs
@@ -1,8 +1,8 @@
 use crate::{
     bad_token::BadTokenDetecting,
     baseline_solver::{
-        estimate_buy_amount, estimate_sell_amount, path_candidates, token_path_to_pair_path,
-        Estimate, DEFAULT_MAX_HOPS,
+        self, estimate_buy_amount, estimate_sell_amount, path_candidates, token_path_to_pair_path,
+        DEFAULT_MAX_HOPS,
     },
     conversions::U256Ext,
     recent_block_cache::Block,
@@ -13,7 +13,7 @@ use ethcontract::{H160, U256};
 use futures::future::join_all;
 use gas_estimation::GasPriceEstimating;
 use model::{order::OrderKind, TokenPair};
-use num::{BigRational, ToPrimitive};
+use num::BigRational;
 use std::{
     collections::{HashMap, HashSet},
     sync::Arc,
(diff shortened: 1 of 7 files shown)


solver/src/solver/http_solver.rs
```diff
@@ -145,16 +145,15 @@ impl HttpSolver {
...
         token_infos
             .iter()
             .map(|(address, token_info)| {
-                let external_price = price_estimates[address]
-                    .as_ref()
-                    .ok()
-                    .and_then(|price| price.to_f64());
+                // TODO: Do we need to filter out infinity? This can happen when buy amount is 0 so
+                // sell / buy becomes infinity.
+                let external_price = price_estimates.get(address).copied();
```
vkgnosis, line 156: Question for @marcovc .

shared/src/price_estimate.rs
```diff
@@ -32,71 +32,58 @@ pub enum PriceEstimationError {
...
-        }))
-        .await
+    async fn estimates(&self, queries: &[Query]) -> Vec<Result<Estimate, PriceEstimationError>> {
+        // Naive default implementation that could be implemented more efficiently for some
+        // estimators.
+        join_all(queries.iter().map(|query| self.estimate(query))).await
     }
+
+    // Used when a price relative to the native token is needed without having a specific amount.
+    fn native_token_amount_to_estimate_prices_with(&self) -> U256;
```
vkgnosis, line 86: I was thinking about moving this out of the price estimator too because it is not its job to decide what a good native token amount is. Instead of passing it from the arguments into here it would be passed into the components that need a generic price without an amount. Not sure about this so decided to keep it for this PR.

solver/src/solver/http_solver.rs
marcovc: You mean that selling about 1 ETH of a token would return 0 of another token?

shared/src/price_estimate.rs
nlordell: I kind of agree. I feel like the native token amount is a configuration detail/implementation detail that doesn't need to be leaked in the trait.

For example a "CoinGecko" or "ParaSwap" price estimator wouldn't care about this but still need to implement this method.

nlordell (approved):
Nice change.

solver/src/solver/http_solver.rs
vkgnosis: Yes. Would be easy to filter out just not sure if we need to. I guess similar question for a 0 price even though I don't think it can happen.

marcovc: Just for my understanding: I've not seen the current code, but before this PR the price passed to the solver was simply cast from a rational to float, wasn't it? And since a rational can't have zero denominator, then why we need to make this decision now?

vkgnosis: correct

vkgnosis: I just saw that your last sentence was a question. The reason this wasn't relevant before is that an infinite price would have been filtered out previously as it cannot be represented by a BigRational. Now we do not go through Rational so we can have a float here that was divided by 0 which becomes infinity.

vkgnosis:
Added unit test fixes.

solver/src/solver/http_solver.rs
marcovc: Ok, I guess we should proceed as we were before for (Rationals with zero denominators) and just don't include that external price. This will work for the http solvers (as we don't depend on these external prices) but I guess it might fail for the ranking functionality in the driver (if there is no price for a traded token we can't score the solution).

vkgnosis: I will keep the behavior from before. But it sounds like we should address the case where we cannot create a token model or where the token model wouldn't have a price. This is an issue without this PR too because we don't remove orders without token model or price.

vkgnosis: Actually I see external_price is Optional in the model so we can continue to turn infinite prices into None.

vkgnosis:
Tested compared to staging. Always got  same results except that we are more lenient about out amount being 0 now because not all routes need to calculate a price.

PR #1110, opened 2021-09-09
Testing Model Crate Part 2

bh2smith:
Improving test coverage of signature, and trade.

Note that this PR also
- removed the #[ignore] tag on the order debug test as our report was considering this untested code.
- removes `#[derive(Hash)]` from `Trade` as it was only every used in a single unit test.
- removes `#[derive(Deserialize)]` from `SigningScheme` since it wasn't used either.

The result of this and the previous PR bring our model crate up to about 90% average coverage as seen in this screenshot.

Screenshot 2021-09-09 at 11 04 22 PM

### Test Plan
Only new tests introduced.

### DIFF
--- a/model/src/signature.rs
+++ b/model/src/signature.rs
@@ -8,7 +8,7 @@ use web3::{
     types::Recovery,
 };
 
-#[derive(Eq, PartialEq, Clone, Copy, Debug, Deserialize, Serialize, Hash)]
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Serialize, Hash)]
 #[serde(rename_all = "lowercase")]
 pub enum SigningScheme {
     Eip712,
@@ -279,6 +279,7 @@ impl<'de> Deserialize<'de> for EcdsaSignature {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use serde_json::json;
 
     #[test]
     fn presign_validates_to_account() {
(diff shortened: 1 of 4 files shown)


model/src/signature.rs
```diff
@@ -8,7 +8,7 @@ use web3::{
     types::Recovery,
 };
 
-#[derive(Eq, PartialEq, Clone, Copy, Debug, Deserialize, Serialize, Hash)]
+#[derive(Eq, PartialEq, Clone, Copy, Debug, Serialize, Hash)]
```
nlordell, line 11: I'm not sure I agree with this change. I would, instead of removing the `Deserialize` add a test for deserialization. My rational being that the `model` crate can be used by some other Rust project that wants to deserialize API responses from our orderbook service.

model/src/trade.rs
```diff
@@ -5,7 +5,7 @@ use num_bigint::BigUint;
 use primitive_types::{H160, H256};
 use serde::{Deserialize, Serialize};
 
-#[derive(Eq, PartialEq, Clone, Debug, Default, Deserialize, Serialize, Hash)]
+#[derive(Eq, PartialEq, Clone, Debug, Default, Deserialize, Serialize)]
```
nlordell, line 8: Nice. I don't think Trade should ever have been `Hash` - it doesn't really make sense.

model/src/signature.rs
bh2smith: That cool I can put it back, but shouldn't we only include this when it is actually used?

model/src/trade.rs
bh2smith: I recall you were saying this and finally figured out how to get rid of it.

model/src/signature.rs
nlordell: Yeah, I guess you're right. We can always add it back if needed.

nlordell (approved):

model/src/signature.rs
bh2smith: and its derived, so it doesn't require any effort to add it back.

PR #1119, opened 2021-09-15
Balancer Coverage Part 1

bh2smith:
Something particularly interesting here is that two ?s in weighted_math.rs were changed to expect including a proof that the error can never occur.

### DIFF
--- a/shared/src/sources/balancer/pool_storage.rs
+++ b/shared/src/sources/balancer/pool_storage.rs
@@ -512,8 +512,6 @@ mod tests {
                     },
                     normalized_weights: vec![weights[i], weights[i + 1]],
                 },
-                "failed assertion at index {}",
-                i
             );
         }
     }
@@ -579,8 +577,6 @@ mod tests {
                         block_created: i as u64 + 1
                     }
                 },
-                "failed assertion at index {}",
-                i
             );
         }
     }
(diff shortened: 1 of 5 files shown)


vkgnosis (approved):
All the coverage stuff is good. I'm less sure about the expect because:
-  there might be an error in the proof (even though i did not find one)
- it is less error prone to always do checked operations like when we refactor other code so maybe the proof no longer holds

On the other hand I do like unwrap for cases that should never happen.

nlordell (approved):

nlordell:
I also prefer `unwrap` for cases that shouldn't happen. Nice proof :smile:

bh2smith:
> I also prefer `unwrap` for cases that shouldn't happen. Nice proof 😄

So, are you guys suggesting I change the expect to unwrap (or are you using them as synonyms)?

nlordell:
> are you guys suggesting I change the expect to unwrap (or are you using them as synonyms)?

Synonyms. `expect` is `unwrap` with a message. I was agreeing that we should panic if something that should never happen does happen (instead of propagating a result).

vkgnosis:
Something with the auto update to main probably went wrong so the lock file isn't up to date. Check out your branch and do a `cargo check`. After that you should see that something in the lock file has changed. Commit the new lock file and it should work.

nlordell:
It looks like #1134 bumped `gas-estimate` crate to the tag `v0.3.3` in the lock file, and is causing issues.

vkgnosis:
Hmm. Surprised CI didn't catch that. Will make a fix.

PR #1120, opened 2021-09-16
Pass driver price estimates into solvers

### LINKED ISSUE #759: Pass price_estimates and time limit into main solver method
```rust
async fn solve(&self, orders: Vec<Liquidity>, gas_price: f64) -> Result<Vec<Settlement>>;
```

should become something like

```rust
async fn solve(&self, orders: Vec<Liquidity>, gas_price: f64, price_estimates: HashMap<H160, Ratio<BigInt>>, deadline: Instant) -> Result<Vec<Settlement>>;
```

Since price estimates (spot prices of baseline liquidity *before* the batch) are relevant for computing the objective value (in particular to normalize surplus across different tokens) it makes sense to pass them into the solver directly. At the moment the SellVolumeFiltering1InchSolver and the MIP solver requery price estimates themselves.

Moreover, since solver futures will be cancelled if not finished in time (and cause an error to be logged) it makes sense to also make the solvers aware of the concrete deadline. This way solvers that know they might time out (e.g. the 1Inch solver) can cancel their internal futures just before and swallow the time out error. It also allows for more sophisticated logic when deciding how many more orders can be solved before the timeout is hit.

vkgnosis:
See issue fixes #759 . Also useful when we have heavier price estimators like Paraswap.

### Test Plan
existing tests

### DIFF
--- a/solver/src/solver/http_solver.rs
+++ b/solver/src/solver/http_solver.rs
@@ -13,6 +13,7 @@ use crate::{
 };
 use ::model::order::OrderKind;
 use anyhow::{anyhow, ensure, Context, Result};
+use bigdecimal::ToPrimitive;
 use buffers::{BufferRetrievalError, BufferRetrieving};
 use ethcontract::{Account, U256};
 use futures::{join, lock::Mutex};
@@ -310,6 +311,7 @@ impl HttpSolver {
         mut orders: Vec<LimitOrder>,
         liquidity: Vec<Liquidity>,
         gas_price: f64,
+        price_estimates: HashMap<H160, BigRational>,
     ) -> Result<(BatchAuctionModel, SettlementContext)> {
         let tokens = self.map_tokens_for_solver(&orders, &liquidity);
 
@@ -318,6 +320,7 @@ impl HttpSolver {
             .native_token_amount_to_estimate_prices_with();
         let queries = tokens
             .iter()
+            .filter(|token| !price_estimates.contains_key(token))
             .map(|token| price_estimate::Query {
                 sell_token: self.native_token,
                 buy_token: *token,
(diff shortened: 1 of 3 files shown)


solver/src/solver/http_solver.rs
```diff
@@ -364,14 +367,19 @@ impl HttpSolver {
...
             .filter_map(|(query, estimate)| {
                 let price = estimate.ok()?.price_in_sell_token_f64(query);
                 Some((query.buy_token, price))
             })
             .collect();
+        new_price_estimates.extend(
+            price_estimates
+                .into_iter()
+                .filter_map(|(token, price)| Some((token, price.to_f64()?))),
+        );
```
vkgnosis, line 382: Nick mentioned that the http solver shouldn't need new price estimates. I think it does because:
- driver only estimates all tokens appearing in user limit orders
- http solver estimates all tokens even from liquidity
- liquidity contains base tokens that we haven't price estimated yet in the driver

nlordell: > http solver estimates all tokens even from liquidity

I imagine that the `set all tokens from liquidity` == `set all tokens in user limit orders` + `set of all base token`. So estimating prices from any token included in liquidity should not need to consider any new pools than are already there.

vkgnosis: The pools are there but the price estimate isn't. So we are going to make calls to the price estimator but the pools it uses should be cached.

nlordell: Right! I think I may have not been super clear with what I meant: the HTTP solver can just use the `Batch` input as all pools available for baseline price estimation instead of needing to rely on the pool cache. 

Its also fine to just rely on the pool cache though :stuck_out_tongue:.

nlordell (approved):

PR #1122, opened 2021-09-17
Migrate from ParaSwap v4 to v5

nlordell:
This PR migrates our ParaSwap API from v4 to v5.

This is to take advantage of the better prices and liquidity available on ParaSwap v5 (such as Balancer V2).

### Test Plan

CI, additionally, run the manual ParaSwap tests:
```
$ NODE_URL=https://mainnet.infura.io/v3/$INFURA_PROJECT_ID cargo test -- --ignored --nocapture paraswap
...
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 122 filtered out; finished in 4.83s
...
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 137 filtered out; finished in 5.93s
...
```

### DIFF
--- a/shared/src/paraswap_api.rs
+++ b/shared/src/paraswap_api.rs
@@ -274,6 +274,8 @@ pub struct TransactionBuilderQuery {
     /// The trade amount amount
     #[serde(flatten)]
     pub trade_amount: TradeAmount,
+    /// The maximum slippage in BPS.
+    pub slippage: u32,
     /// The decimals of the source token
     pub src_decimals: usize,
     /// The decimals of the destination token
@@ -293,19 +295,22 @@ pub enum TradeAmount {
         /// The source amount
         #[serde(with = "u256_decimal")]
         src_amount: U256,
-        /// The maximum slippage in BPS.
-        slippage: u32,
     },
     #[serde(rename_all = "camelCase")]
     Buy {
         /// The destination amount
         #[serde(with = "u256_decimal")]
         dest_amount: U256,
-        /// The maximum slippage in BPS.
-        slippage: u32,
     },
 }
 
+/// A helper struct to wrap a `TransactionBuilderQuery` that we get as input from
+/// the `ParaswapApi` trait.
+///
+/// This is done because the `partner` is longer specified in the headersd but
+/// instead in the POST body, but we want the API to stay mostly compatible and
+/// not require passing it in every time we build a transaction given that the
+/// API instance already knows what the `partner` value is.
 #[derive(Clone, Debug, Serialize)]
 #[serde(rename_all = "camelCase")]
 struct TransactionBuilderQueryWithPartner<'a> {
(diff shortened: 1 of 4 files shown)


solver/src/solver/paraswap_solver.rs
```diff
@@ -46,10 +47,10 @@ impl ParaswapSolver {
         web3: Web3,
         settlement_contract: GPv2Settlement,
         token_info: Arc<dyn TokenInfoFetching>,
-        slippage_bps: usize,
+        slippage_bps: u32,
```
nlordell, line 50: Changed to `u32` since in theory there are `usize` values that aren't representable in double precision floating point numbers, and many JSON parsers use those by default.

shared/src/paraswap_api.rs
```diff
@@ -269,31 +271,56 @@ pub struct TransactionBuilderQuery {
...
+#[derive(Clone, Debug, Eq, PartialEq, Serialize)]
+#[serde(untagged)]
+pub enum TradeAmount {
+    #[serde(rename_all = "camelCase")]
+    Sell {
+        /// The source amount
+        #[serde(with = "u256_decimal")]
+        src_amount: U256,
+        /// The maximum slippage in BPS.
+        slippage: u32,
```
fleupold, line 297: nit (really not important): this could have gone on the top level struct.

shared/src/paraswap_api.rs
```diff
@@ -269,31 +271,56 @@ pub struct TransactionBuilderQuery {
...
+        #[serde(with = "u256_decimal")]
+        dest_amount: U256,
+        /// The maximum slippage in BPS.
+        slippage: u32,
+    },
+}
+
+#[derive(Clone, Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct TransactionBuilderQueryWithPartner<'a> {
```
fleupold, line 311: nit: maybe small comment why you need this helper struct?

solver/src/solver.rs
```diff
@@ -139,7 +139,7 @@ pub fn create(
     fee_factor: f64,
     min_order_size_one_inch: U256,
     disabled_one_inch_protocols: Vec<String>,
-    paraswap_slippage_bps: usize,
+    paraswap_slippage_bps: u32,
```
fleupold, line 142: nanonit: `paraswap_partner_header_value` is now not a very accurate variable name anymore.

fleupold (approved):
Very nice!

🚄

bh2smith (approved):

shared/src/paraswap_api.rs
```diff
@@ -269,31 +271,56 @@ pub struct TransactionBuilderQuery {
...
+    #[serde(flatten)]
+    query: TransactionBuilderQuery,
+    partner: &'a str,
```
vkgnosis, line 314: This seems to be set but never read. Probably not intentional.

nlordell: Its included in the JSON that we POST to the ParaSwap API.

vkgnosis: Ah so its  part of the url for prices but body for transaction. Makes sense.

vkgnosis (approved):

Issue #1067, opened 2021-09-02
Separate fee factor for Balancer orders


nlordell:
Given #1066, one of the first "custom fee logics" we will want to implement will be to allow additional subsidies for orders from the Balancer UI. Specifically these orders must:
- Use the Balancer UI app ID
- Use Balancer Vault sell token balances.

PR #1125, opened 2021-09-20
* Fee scaling based on app ID

bh2smith:
Part of #1067

We introduce a new config parameter `partner-additional-fee-factors` which holds a hashmap of app data and fee factors which are applied at the time of fee validation (right before an order is written to the orderbook).

Note that all the values involved are U256, so we used some f64 lossy conversion to apply these fractional amounts

### Test Plan
Introduced new test which passes when it should and fails when it shouldn't.

To test with new config parameters run the orderbook with

```
--partner-additional-fee-factors '{"0xE9F29AE547955463ED535162AEFEE525D8D309571A2B18BC26086C8C35D781EB":0.5, "0xE4D1AB10F2C9FFE7BDD23C315B03F18CFF90888D6B2BB5022BACD46AB9CDDF24": 0.7}'
```

and find this displayed as a log at the start

Screenshot 2021-09-21 at 3 15 22 PM

This (e2e) test along with the newly added unit tests confirm new functionality is working as expected.

### DIFF
--- a/orderbook/src/fee.rs
+++ b/orderbook/src/fee.rs
@@ -8,6 +8,9 @@ use shared::{bad_token::BadTokenDetecting, price_estimate::PriceEstimating};
 use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
 
+// TODO - find out correct value and make it a config item.
+const BALANCER_APP_ID: [u8; 32] = [1u8;32];
+
 pub type Measurement = (U256, DateTime<Utc>);
 
 pub type EthAwareMinFeeCalculator = EthAdapter<MinFeeCalculator>;
@@ -43,7 +46,7 @@ pub trait MinFeeCalculating: Send + Sync {
     ) -> Result<Measurement, PriceEstimationError>;
 
     // Returns true if the fee satisfies a previous not yet expired estimate, or the fee is high enough given the current estimate.
-    async fn is_valid_fee(&self, sell_token: H160, fee: U256) -> bool;
+    async fn is_valid_fee(&self, sell_token: H160, fee: U256, app_data: [u8; 32]) -> bool;
 }
 
 #[async_trait::async_trait]
@@ -131,8 +134,10 @@ where
             .await
     }
 
-    async fn is_valid_fee(&self, sell_token: H160, fee: U256) -> bool {
-        self.calculator.is_valid_fee(sell_token, fee).await
+    async fn is_valid_fee(&self, sell_token: H160, fee: U256, app_data: [u8; 32]) -> bool {
+        self.calculator
+            .is_valid_fee(sell_token, fee, app_data)
+            .await
     }
 }
 
@@ -258,18 +263,25 @@ impl MinFeeCalculating for MinFeeCalculator {
     }
 
     // Returns true if the fee satisfies a previous not yet expired estimate, or the fee is high enough given the current estimate.
-    async fn is_valid_fee(&self, sell_token: H160, fee: U256) -> bool {
+    async fn is_valid_fee(&self, sell_token: H160, fee: U256, app_data: [u8; 32]) -> bool {
+
+        let app_based_fee_factor = match app_data {
+            BALANCER_APP_ID => U256::from(2),
+            _ => U256::one(),
+        };
+        let scaled_fee = fee.checked_mul(app_based_fee_factor).unwrap_or(U256::one());
+
         if let Ok(Some(past_fee)) = self
             .measurements
             .get_min_fee(sell_token, None, None, None, (self.now)())
             .await
         {
-            if fee >= past_fee {
+            if scaled_fee >= past_fee {
                 return true;
             }
         }
         if let Ok(current_fee) = self.compute_min_fee(sell_token, None, None, None).await {
-            return fee >= current_fee;
+            return scaled_fee >= current_fee;
         }
         false
     }
@@ -391,12 +403,12 @@ mod tests {
         let mut calculator = MockMinFeeCalculating::default();
         calculator
             .expect_is_valid_fee()
-            .withf(move |&sell_token, &fee| sell_token == token && fee == 42.into())
+            .withf(move |&sell_token, &fee, &_| sell_token == token && fee == 42.into())
             .times(1)
-            .returning(|_, _| true);
+            .returning(|_, _, _| true);
 
         let eth_aware = EthAdapter { calculator, weth };
-        assert!(eth_aware.is_valid_fee(token, 42.into()).await);
+        assert!(eth_aware.is_valid_fee(token, 42.into(), [0u8; 32]).await);
     }
 
     impl MinFeeCalculator {
@@ -447,11 +459,11 @@ mod tests {
 
         // fee is valid before expiry
         *time.lock().unwrap() = expiry - Duration::seconds(10);
-        assert!(fee_estimator.is_valid_fee(token, fee).await);
+        assert!(fee_estimator.is_valid_fee(token, fee, [0u8; 32]).await);
 
         // fee is invalid for some uncached token
         let token = H160::from_low_u64_be(2);
-        assert!(!fee_estimator.is_valid_fee(token, fee).await);
+        assert!(!fee_estimator.is_valid_fee(token, fee, [0u8; 32]).await);
     }
 
     #[tokio::test]
@@ -480,11 +492,11 @@ mod tests {
         let lower_fee = fee - U256::one();
 
         // slightly lower fee is not valid
-        assert!(!fee_estimator.is_valid_fee(token, lower_fee).await);
+        assert!(!fee_estimator.is_valid_fee(token, lower_fee, [0u8; 32]).await);
 
         // Gas price reduces, and slightly lower fee is now valid
         *gas_price.lock().unwrap() /= 2.0;
-        assert!(fee_estimator.is_valid_fee(token, lower_fee).await);
+        assert!(fee_estimator.is_valid_fee(token, lower_fee, [0u8; 32]).await);
     }
 
     #[tokio::test]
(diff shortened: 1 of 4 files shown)


orderbook/src/fee.rs
```diff
@@ -258,18 +263,24 @@ impl MinFeeCalculating for MinFeeCalculator {
     }
 
     // Returns true if the fee satisfies a previous not yet expired estimate, or the fee is high enough given the current estimate.
-    async fn is_valid_fee(&self, sell_token: H160, fee: U256) -> bool {
+    async fn is_valid_fee(&self, sell_token: H160, fee: U256, app_data: [u8; 32]) -> bool {
+        let app_based_fee_factor = match app_data {
+            BALANCER_APP_ID => U256::from(2),
+            _ => U256::one(),
+        };
+        let scaled_fee = fee.checked_mul(app_based_fee_factor).unwrap_or(fee);
```
bh2smith, line 271: This is the 🍖  of the change

nlordell (commented):
Nice! Looks like its not too complicated.

This is kind of what I had in mind modulo:
- Making the app ID and factors configurable
- Allowing more than one factor.
- Allowing for non-integer scaling

Maybe we can pass a flag, something like:
```
--partner-addtional-fee-factors '{"$BALANCER_APP_ID":0.5,"$METAMASK_APP_ID":0.7}'
```

bh2smith:
> Allowing for non-integer scaling
I should be able to do this by converting to a rational and first multiplying by numerator and then dividing by denominator. That way we don't lose too much precision and still get what we like).

nlordell:
> I should be able to do this by converting to a rational and first multiplying by numerator and then dividing by denominator. That way we don't lose too much precision and still get what we like).

I would just use `f64` and do some lossy conversion. The exact fee amount doesn't have to be extremely precise (especially if its getting reduced).

fleupold:
Looking at this change I'm wondering if each individual app id should give their factor relative to the default subsidised fee or relative to the unsubsidized fee? My worry when doing the former would be that it might be hard to change the fee subsidy for our default interfaces without affecting e.g. Balancer.

Or will the default subsidy go away and the fee endpoint will only return unsubsidized fee?

nlordell:
> My worry when doing the former would be that it might be hard to change the fee subsidy for our default interfaces without affecting e.g. Balancer.

Hmmm, interesting, I would have said the opposite. In my mind:
- Balancer FE would query fee endpoint which returns `X`, which is the default subsidized fee.
- It would create order with `feeAmount: X * BALANCER_FEE_FACTOR` (so total factor is now `fee_factor * BALANCER_FEE_FACTOR`)
- In the check now, we are checking that the order fee `> unsubsidized_fee * fee_factor * partner_fee_factor`.

This means that if we adjust the fee factor in the BE, then the scaling in the FE will "just work". Maybe I'm missing something or didn't fully understand your concern.

bh2smith:
> Or will the default subsidy go away and the fee endpoint will only return unsubsidized fee?

It was my impression that we would use the app_data from barn to specify the fee subsidy there (instead of what we do now), however, I do see that it could be a bit of a hassle and it would be nice if there were a global fee subsidy along side with subsidy based on app_data.

fleupold:
> This means that if we adjust the fee factor in the BE, then the scaling in the FE will "just work". Maybe I'm missing something or didn't fully understand your concern.

Right, I guess in this case we have to make sure that all our interfaces use an App ID for which we have a dedicated fee factor.  I wouldn't want changes to the CowSwap fee factor effect the fee Balancer UI user pay.

orderbook/src/main.rs
```diff
@@ -127,6 +127,14 @@ struct Arguments {
         parse(try_from_str = shared::arguments::duration_from_seconds),
     )]
     solvable_orders_max_update_age: Duration,
+
+    #[structopt(
```
nlordell, line 131: Can you include a doc comment here? It gets used to automatically generate the documentation in `--help`.

nlordell:
We had a quick sync discussion and:
- We want to keep the baseline/global fee subsidy (i.e. what `fee_factor` currently is).
- We want to be to able to configure happy hours **only** on CowSwap (this means that we would need to have a `partner-fee-subsidy` for CowSwap).
  - Once we introduce referrals, it would complicate things a little.
- Eventually, fee based on full order data (not just `app_data`).
  - This would allow things like increased fee subsidy for orders trading internal Vault or Settlement balances.
- Fee estimates route ALSO needs to consider order data (to avoid manually computing additional fee subsidies in the front end).

orderbook/src/fee.rs
```diff
@@ -248,18 +258,24 @@ impl MinFeeCalculating for MinFeeCalculator {
     }
 
     // Returns true if the fee satisfies a previous not yet expired estimate, or the fee is high enough given the current estimate.
-    async fn is_valid_fee(&self, sell_token: H160, fee: U256) -> bool {
+    async fn is_valid_fee(&self, sell_token: H160, fee: U256, app_data: [u8; 32]) -> bool {
+        let app_based_fee_factor = self
+            .partner_fee_factors
+            .get(&H256::from(app_data))
+            .unwrap_or(&1.0);
+        let scaled_fee = U256::from_f64_lossy(fee.to_f64_lossy() / app_based_fee_factor);
```
nlordell, line 266: One thing I'm not a huge fan of is that we apply fee factors in two different places. Once here, and once in `compute_min_fee`. Maybe we can refactor `compute_min_fee` to also take the `app-data` and do all the fee subsidy computation in one place?

bh2smith: This scaled fee is only computed once... I can make your suggestion happen anyway, but I don't see how it's being applied twice at the moment.

nlordell: Sorry, my comment wasn't super clear. I meant we apply fee factors in two different places in code (the global one, and the partner one). They are both correctly only applied once, the comment was more about how it would be nice if the fee factor computations only appear in a single place in the code, to make it easier to find and maintain (so more a code maintenance comment, not a correctness one).

However, looking at the code, this might be a problem because the app ID isn't known at the time the "saved" min fee is created.

orderbook/src/fee.rs
```diff
@@ -248,18 +258,23 @@ impl MinFeeCalculating for MinFeeCalculator {
...
+            .get(&H256::from(app_data))
+            .unwrap_or(&1.0);
+
         if let Ok(Some(past_fee)) = self
             .measurements
             .get_min_fee(sell_token, None, None, None, (self.now)())
             .await
         {
-            if fee >= past_fee {
+            if fee >= U256::from_f64_lossy(past_fee.to_f64_lossy() * app_based_fee_factor) {
```
nlordell, line 272: ubernit: Might be worth having a small helper method:
```
fn apply_fee_factor(fee: U256, factor: f64) -> U256;
```

orderbook/src/fee.rs
```diff
@@ -381,12 +397,12 @@ mod tests {
         let mut calculator = MockMinFeeCalculating::default();
         calculator
             .expect_is_valid_fee()
-            .withf(move |&sell_token, &fee| sell_token == token && fee == 42.into())
+            .withf(move |&sell_token, &fee, &_| sell_token == token && fee == 42.into())
```
nlordell, line 400:
```suggestion
            .withf(move |&sell_token, &fee, &app_data| sell_token == token && fee == 42.into() && app_data == [0; 32])
```

nlordell (approved):

orderbook/src/fee.rs
```diff
@@ -134,19 +139,23 @@ where
...
 impl MinFeeCalculator {
+    #[allow(clippy::too_many_arguments)]
     fn new(
         price_estimator: Arc<dyn PriceEstimating>,
         gas_estimator: Arc<dyn GasPriceEstimating>,
         native_token: H160,
         measurements: Arc<dyn MinFeeStoring>,
         fee_factor: f64,
         bad_token_detector: Arc<dyn BadTokenDetecting>,
+        partner_additional_fee_factors: HashMap<H256, f64>,
```
vkgnosis, line 158: I'm confused that this is H256 while app_data is [u8; 32] but we convert one into the other. I feel like the type should be the same.

orderbook/src/main.rs
```diff
@@ -127,6 +127,21 @@ struct Arguments {
...
+    /// Furthermore, a value of
+    /// - 1 means no subsidy and is the default for all app_data not contained in this list.
+    /// - 0.5 means that this project pays only 50% of the estimated fees.
+    #[structopt(
+        long,
+        env,
+        default_value = "{}",
+        parse(try_from_str = serde_json::from_str),
+    )]
+    partner_additional_fee_factors: HashMap<H256, f64>, // '{"$BALANCER_APP_ID":0.5,"$METAMASK_APP_ID":0.7}'
```
vkgnosis, line 144: Is using json in a command line argument a good way to implement this? It is a small amount of code but it feels weird to me that you would pass json. More natural I would find something like
--partner-additional-fee-factors "0x0000000:0.5,0x0000:0.3".
That is more code to implement though. You would need a custom parse function that goes from string (the raw argument) to the hashmap.
Not sure how others feel about this.

orderbook/src/fee.rs
bh2smith: I have chosen to do this because of the convenience for runtime arguments 

As an H256 app_id can be passed in as a hex string... as [u8;32] we would have to pass an array of bytes `[0, 0, 0, 0, 0... 1 ]`. If you have some suggestion on how we can parse the hex string directly into the correct format, I would be happy to apply it.

orderbook/src/main.rs
bh2smith: I had initially thought of going with a list of tuples (as you suggested), but thought it was cute/nice to go directly to the hashmap that we wanted. Will wait on others as well.

orderbook/src/fee.rs
vkgnosis: You would do it with a custom parse function or wrapper type like in the comment below. We are already using https://docs.rs/hex/0.4.3/hex/ in some places which can turn a hex string into bytes.

nlordell: `H256` is so convenient :stuck_out_tongue:. That being said, I do agree with @vkgnosis that we are slightly abusing the type here.

orderbook/src/main.rs
nlordell: We used JSON in v1 and it was OK. I agree that its not ideal, but I do slightly prefer it to writing custom deserialization code. No strong opinion either.

vkgnosis (approved):
Approving because it is ok to merge regardless of what we decide in the remaining discussions.

PR #1126, opened 2021-09-20
Forbid orders with 0 buy or sell amount

vkgnosis:
Conceptually this doesn't make sense.
Motivated by http solver erroring when this occurs.

### Test Plan
simple enough to not need new test

### DIFF
--- a/orderbook/src/orderbook.rs
+++ b/orderbook/src/orderbook.rs
@@ -39,6 +39,7 @@ pub enum AddOrderResult {
     SameBuyAndSellToken,
     UnsupportedBuyTokenDestination(BuyTokenDestination),
     UnsupportedSellTokenSource(SellTokenSource),
+    ZeroAmount,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


bh2smith (approved):

orderbook/src/orderbook.rs
```diff
@@ -126,10 +127,13 @@ impl Orderbook {
         ) {
             return Ok(AddOrderResult::UnsupportedSignature);
         }
-
         if has_same_buy_and_sell_token(&order, &self.native_token) {
             return Ok(AddOrderResult::SameBuyAndSellToken);
         }
+        if order.buy_amount.is_zero() || order.sell_amount.is_zero() {
```
nlordell, line 133: In theory, `buy_amount` can be 0 for a "true" market order. Although, for all practical purposes `1` is good enough and avoids sneaky divide by 0 issues that may come up.

nlordell (approved):

PR #1127, opened 2021-09-20
Balancer Baseline Stable Math

bh2smith:
First step toward #1074 

This PR implements the arithmetic for balancer stable pools. We essentially just mimic the relevant internal contract methods for `calculate_X_given_Y` where `{ X, Y } = { in, out }`. When unit testing, we also mimic their TS library which contains approximation of those functions using float arithmetic to validate the integer implementations are accurate.

- [x] Include links and documentation to mirrored functions.

Note that the remaining todo in the code is to remove allow(deadcode) and include these computations in baseline solver.

### Test Plan
Lots of new tests.


shared/src/sources/balancer/swap/fixed_point.rs
```diff
@@ -96,6 +96,10 @@ impl Debug for Bfp {
 }
 
 impl Bfp {
+    pub fn to_f64_lossy(self) -> f64 {
+        self.as_uint256().to_f64_lossy() / 1e18
+    }
```
vkgnosis, line 101: Not sure if a problem in practice but want to note that this introduces a lot of unnecessary floating point error because it first creates a large U256, converts it to float which is lossy and then divides by a large float which is lossy again.
A better way to do it is to get the integer amount first by dividing out 1e18 as U256 and then converting to float because that number will be much closer to 0. Then convert the rest (amount after comma) to U256 and only divide that by 1e18 in float.
Haven't fully thought this true so the right solution might be a little different but the concept should be the same.

shared/src/sources/balancer/swap/stable_math.rs
```diff
@@ -0,0 +1,485 @@
...
+use ethcontract::U256;
+use lazy_static::lazy_static;
+
+lazy_static! {
+    static ref AMP_PRECISION: U256 = U256::from(1000);
+}
+
+fn calculate_invariant(
+    amplification_parameter: U256,
+    balances: Vec<Bfp>,
```
vkgnosis, line 15: Can be `&[Bfp]` so that you don't have to clone this when calling the function.

shared/src/sources/balancer/swap/stable_math.rs
```diff
@@ -0,0 +1,485 @@
...
+            Some(invariant) => return Ok(invariant),
+        }
+    }
+
+    Err(Error::StableInvariantDidntConverge)
+}
+
+fn calc_out_given_in(
+    amplification_parameter: U256,
+    mut balances: Vec<Bfp>,
```
vkgnosis, line 80: can be `&mut [Bfp]` because we do not change the size of the vec.

shared/src/sources/balancer/swap/stable_math.rs
```diff
@@ -0,0 +1,485 @@
...
+    token_index_in: usize,
+    token_index_out: usize,
+    token_amount_in: Bfp,
+) -> Result<Bfp, Error> {
+    let invariant = calculate_invariant(amplification_parameter, balances.clone(), true)?;
+    balances[token_index_in] = balances[token_index_in].add(token_amount_in)?;
```
vkgnosis, line 86: Whenever we index into balances like this (and in other functions to) that is a potential panic because the index can be out of bounds. Do we need to check this and return an error? I'm not sure.

shared/src/sources/balancer/swap/fixed_point.rs
bh2smith: This is a great suggestion, I will do this. However, I would like to point out that this is only used inside the test cases introduced here. I should probably add `#[cfg(test)]` as a header to this method.

shared/src/sources/balancer/swap/stable_math.rs
bh2smith: Great idea - Thanks!

shared/src/sources/balancer/swap/stable_math.rs
bh2smith: Yea, thats a pretty good point, I was literally mirroring the contract code, but we should use `array.get()?`

vkgnosis: It depends on whether it is possible to get the out of bounds error. Maybe that is supposed to already have been checked by a higher layer which calls these functions in which case panic would be fine.
If we do want to check it here yes you could use array.get .

shared/src/sources/balancer/swap/fixed_point.rs
nlordell: Since its only used in tests, I would prefer adding `#[cfg(test)]` to avoid unnecessary for testing.

vkgnosis: It is 1 line versus 5 lines I think so not much extra complexity and could be useful for tests too to have lower error. But yeah no strong opinion as long as its used only in tests.

shared/src/sources/balancer/swap/stable_math.rs
nlordell: You can use a `array.get(index).ok_or(InvalidToken)`. They seem to have an error for this (although I'm not 100% sure its used for invalid indices in the pool tokens array:

https://github.com/balancer-labs/balancer-v2-monorepo/blob/a3875bdf1c5afb57092caf4d36062d906dcc2513/pkg/solidity-utils/contracts/helpers/BalancerErrors.sol#L126

bh2smith: I'm finding it bit awkward to write the get_mut and update the value. Would it be appropriate to simply check first that `token_index_in < balances.len()` or is that not rusty enough?

bh2smith: I have gone with this which I am not super happy with.

```rs
// Ensure no index error at token_index_in
let balance_token_in = balances.get(token_index_in).ok_or(Error::InvalidToken)?;
balances[token_index_in] = balance_token_in.add(token_amount_in)?;
```

shared/src/sources/balancer/swap/fixed_point.rs
bh2smith: So I tried out you suggestion here and found that the two results are identical. Please let me know if this is the approach you envisioned:

```rust
        let as_uint = self.as_uint256();
        let int_part = as_uint / *ONE_18;
        let decimal_part = as_uint - (int_part * *ONE_18);
        println!("Int {}, Dec {}", int_part, decimal_part);
        println!("OLD SOLUTION: {}", self.as_uint256().to_f64_lossy() / 1e18);
        println!("NEW SOLUTION: {}", int_part.to_f64_lossy() + decimal_part.to_f64_lossy() / 1e18);
```

vkgnosis: This is what I had in mind. It thought it would be equivalent for small numbers but I expected there to be differences for large ones.

shared/src/sources/balancer/swap/stable_math.rs
vkgnosis: Checking `index < len` somewhere first is completely fine too. That  could be more ergonomic if you are querying the same index multiple times.

shared/src/sources/balancer/swap/fixed_point.rs
vkgnosis: Which of these functions would be public? If it isn't all then it would also be fine to only check bounds in those higher level functions while the others have the invariant that all indexes are valid.

shared/src/sources/balancer/swap/stable_math.rs
vkgnosis: Which of these functions would be public? If it isn't all then it would also be fine to only check bounds in those higher level functions while the others have the invariant that all indexes are valid.

shared/src/sources/balancer/swap/stable_math.rs
```diff
@@ -0,0 +1,495 @@
...
+    // calling `_getTokenBalanceGivenInvariantAndAllOtherBalances` which doesn't alter the balances array.
+    balances[token_index_in] = balances[token_index_in]
+        .sub(token_amount_in)
+        .expect("will not underflow");
+
+    balances
+        .get(token_index_out)
+        .ok_or(Error::InvalidToken)?
+        .sub(final_balance_out)?
+        .sub(Bfp::from_wei(1.into()))
```
vkgnosis, line 106: Potential problem here is that you might update balances[token_index_in] first and then error afterwards. The function would return an error but the input has changed. Whether this is a problem depends on how it is called. Either way should be documented.

bh2smith: I am not exactly sure how I am supposed to go about doing this... I would actually prefer to simply assert in the beginning that `balances.len()` is greater than the two indices being passed in here. is it even possible to mutate the array's length? Ideally I could use getters everywhere, but this make the self addition and subtraction very challenging (to the extent that I couldn't figure out how). 

Also, not sure I understand the potential problem you have pointed out here. Note that we access `token_index_in` first and then `token_index_out` later.

shared/src/sources/balancer/swap/stable_math.rs
bh2smith: The only ones that will eventually be public are `calc_X_given_Y`.

shared/src/sources/balancer/swap/stable_math.rs
vkgnosis: What I mean is that in line 98 you modify a value in `balances`. But then line 104 could return an error. In that case the function has returned an error but it has still modified the input. That might be unexpected for someone calling it because they think no modifications take place if there is an error.

>  would actually prefer to simply assert in the beginning

Sounds fine.

shared/src/sources/balancer/swap/stable_math.rs
```diff
@@ -0,0 +1,497 @@
...
+    balances: &[Bfp],
+    invariant: U256,
+    token_index: usize,
+) -> Result<Bfp, Error> {
+    // Rounds result up overall
+    let num_tokens_usize = balances.len();
+    let num_tokens = U256::from(num_tokens_usize);
+    let amp_times_total = amplification_parameter.bmul(num_tokens)?;
+    let mut sum = balances[0].as_uint256();
+    let mut p_d = balances[0].as_uint256().bmul(num_tokens)?;
```
vkgnosis, line 152: Could panic if balances is empty. Again I don't know if this can happen in practice just wanted to note it.

vkgnosis (approved):

shared/src/sources/balancer/swap/stable_math.rs
bh2smith: I'll add it to the index verification at the top.

nlordell: I don't think we need index verification here as this is an internal function that should only get called by `calc_*_given_*` which would return an error if the balances slice is empty (since there is no `token_*_index` for which `token_*_index < bal.len() && bal.len() == 0`).

I think just adding a note that `panics if balances is empty` is fine.

bh2smith: gefixed!

nlordell (approved):
Looks good.

bh2smith:
Not sure what happened, but when merging main here I got an error in a place I didn't touch. Will check it out and see what happened.

```
failures:

---- settlement::settlement_encoder::tests::merge_fails_because_price_is_different stdout ----
thread 'settlement::settlement_encoder::tests::merge_fails_because_price_is_different' panicked at 'assertion failed: encoder0.merge(encoder1).is_err()', solver/src/settlement/settlement_encoder.rs:553:9


failures:
    settlement::settlement_encoder::tests::merge_fails_because_price_is_different
error: test failed, to rerun pass '-p solver --lib'
```

PR #1137, opened 2021-09-21
Do not duplicate work in price estimator when estimating multiple tokens

vkgnosis:
When estimating multiple tokens the current implementation duplicates
some work because it treats each estimate individually without sharing
data.
This commit shares data for the gas price, unsupported tokens and pools.
For every inner component now only make one request.
This has the side effect that the inner functions are now pure (and not async).

### Test Plan
existing tests and will manually test price estimates compared to staging

### DIFF
--- a/shared/src/price_estimate.rs
+++ b/shared/src/price_estimate.rs
@@ -7,11 +7,13 @@ use crate::{
 };
 use anyhow::{anyhow, Result};
 use ethcontract::{H160, U256};
-use futures::future::join_all;
 use gas_estimation::GasPriceEstimating;
 use model::{order::OrderKind, TokenPair};
 use num::BigRational;
-use std::{collections::HashMap, sync::Arc};
+use std::{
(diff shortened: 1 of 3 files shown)


shared/src/price_estimate.rs
```diff
@@ -869,81 +910,35 @@ mod tests {
...
             .1;
-
         assert!(out_amount_considering_gas_costs != out_amount_disregarding_gas_costs);
-
         assert!(out_amount_considering_gas_costs.to_f64_lossy() <= 1.008e19);
         assert!(out_amount_disregarding_gas_costs.to_f64_lossy() <= 1.008e19);
     }
-
-    #[tokio::test]
-    async fn estimate_price_does_not_panic_on_zero_amount() {
```
vkgnosis, line 918: Deleted this test because while updating I noticed it is no longer relevant.

nlordell (approved):
Nice refactor!

PR #1138, opened 2021-09-21
Allow using Paraswap as price estimator in api

vkgnosis:
moved configuration parameters from solver args to shared args.

### Test Plan
still compiles


nlordell (approved):
Nice! All we need is a "PriorityPriceEstimator" now to get estimates from `ParaSwap` then `Baseline` :smile:

PR #1141, opened 2021-09-21
Refactor http solver

### LINKED ISSUE #1071: Balancer pools not considered when filtering orders with fee connection for AMMs
We currently filter out orders that don't have a fee connection when passing the orderbook to the solver. However, this **only** considers fee connections with Uniswap-like AMMs, and ignores Balancer weighted pools.

While this is currently not an issue (since we don't accept orders without fee connections and the orderbook API only considers Uniswap pools) this will become a problem when we add Balancer pools to the price estimates in the orderbook API, where certain orders will not be sent to the optimization solvers if they trade tokens for which there is no UniswapV2-like liquidity.

Relevant lines: https://github.com/gnosis/gp-v2-services/blob/da836b43575c6780ad750d9e114c1ab17b59e31c/solver/src/solver/http_solver.rs#L474-L480

nlordell:
Fixed #1071

This PR refactors the `prepare_model` logic of the conversion of the solution to a settlement in the HTTP solver module.

The motivation for this refactor was to cleanup some of the complicated `shift_*` logic (which would only get more complicated as new liquidity gets added) as well as remove some of the code duplication caused by having multiple liquidity types.

Additionally, with this refactor, we now consider Balancer pools when filtering out orders without a fee connection (previously only Uniswap pools were considered), fixing #1071.

One final minor change, the batch model uses `BTreeMap`s instead of `HashMap`s. This makes it so the keys in the JSON appear in sorted (thus, deterministic) order.

### Test Plan

Adjusted unit tests for new refactor.

Run the solver connected to testing MIP and Quasimodo optimizers and make sure it finds solutions:
```
$ cargo run -p solver -- \
  --base-tokens 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2,0x6B175474E89094C44Da98b954EedeAC495271d0F,0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48,0xdAC17F958D2ee523a2206206994597C13D831ec7,0xc00e94Cb662C3520282E6f5717214004A7f26888,0x9f8F72aA9304c8B593d555F12eF6589cC3A579A2,0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599, \
  --baseline-sources Uniswap,Sushiswap,BalancerV2 \
  --mip-solver-url "$MIP_SOLVER_URL" \
  --quasimodo-solver-url "$QUASIMODO_SOLVER_URL" \
  --solvers Mip,Quasimodo \
  --transaction-strategy DryRun
...
2021-09-21T20:04:49.256Z DEBUG solver::driver: solver Mip found solution:
...
2021-09-21T20:04:49.256Z DEBUG solver::driver: solver Quasimodo found solution:
...
```

### DIFF
--- a/solver/src/solver/http_solver.rs
+++ b/solver/src/solver/http_solver.rs
@@ -4,9 +4,7 @@ mod settlement;
 
 use self::{model::*, settlement::SettlementContext};
 use crate::{
-    liquidity::{
-        ConstantProductOrder, LimitOrder, Liquidity, StablePoolOrder, WeightedProductOrder,
-    },
+    liquidity::{LimitOrder, Liquidity},
     settlement::Settlement,
     settlement_submission::retry::is_transaction_failure,
     solver::{Auction, Solver},
@@ -18,6 +16,7 @@ use buffers::{BufferRetrievalError, BufferRetrieving};
 use ethcontract::{Account, U256};
 use futures::{join, lock::Mutex};
 use lazy_static::lazy_static;
+use maplit::{btreemap, hashset};
 use num::{BigInt, BigRational};
 use primitive_types::H160;
 use reqwest::{header::HeaderValue, Client, Url};
@@ -27,16 +26,18 @@ use shared::{
     token_info::{TokenInfo, TokenInfoFetching},
 };
 use std::{
-    collections::{HashMap, HashSet},
+    collections::{BTreeMap, HashMap, HashSet},
+    iter::FromIterator as _,
     sync::Arc,
     time::{Duration, Instant},
 };
 
-// Estimates from multivariate linear regression here:
-// https://docs.google.com/spreadsheets/d/13UeUQ9DA4bHlcy9-i8d4nSLlCxSfjcXpTelvXYzyJzQ/edit?usp=sharing
 lazy_static! {
+    // Estimates from multivariate linear regression here:
+    // https://docs.google.com/spreadsheets/d/13UeUQ9DA4bHlcy9-i8d4nSLlCxSfjcXpTelvXYzyJzQ/edit?usp=sharing
     static ref GAS_PER_ORDER: U256 = U256::from(66_315);
     static ref GAS_PER_UNISWAP: U256 = U256::from(94_696);
+
     // Taken from a sample of two swaps
     // https://etherscan.io/tx/0x72d234d35fd169ef497ba0a1dc23258c96f278fb688d375d135eb012e5311009
     // https://etherscan.io/tx/0x1c345a6da1edb2bba953685a4cf85f6a0d967ac751f8c5b518578c5fd20a7c96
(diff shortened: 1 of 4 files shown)


solver/src/solver/http_solver.rs
```diff
@@ -129,195 +130,14 @@ impl HttpSolver {
         }
     }
 
-    fn map_tokens_for_solver(&self, orders: &[LimitOrder], liquidity: &[Liquidity]) -> Vec<H160> {
```
nlordell, line 132: Moved :point_down:

solver/src/solver/http_solver.rs
```diff
@@ -129,195 +130,14 @@ impl HttpSolver {
...
-        });
-
-        order_tokens
-            .chain(liquidity_tokens)
-            .collect::<HashSet<_>>()
-            .into_iter()
-            .collect()
-    }
-
-    fn token_models(
```
nlordell, line 149: Moved :point_down:

solver/src/solver/http_solver.rs
```diff
@@ -129,195 +130,14 @@ impl HttpSolver {
...
-                        external_price,
-                        normalize_priority: Some(if &self.native_token == address { 1 } else { 0 }),
-                        internal_buffer: buffers.get(address).copied(),
-                    },
-                )
-            })
-            .collect()
-    }
-
-    fn map_orders_for_solver(&self, orders: Vec<LimitOrder>) -> HashMap<usize, LimitOrder> {
```
nlordell, line 175: Removed.

solver/src/solver/http_solver.rs
```diff
@@ -129,195 +130,14 @@ impl HttpSolver {
...
-        orders.into_iter().enumerate().collect()
-    }
-
-    fn order_models(
```
nlordell, line 179: Moved :point_down: outside of `impl HttpSolver` block to make it easier to unit test.

solver/src/solver/http_solver.rs
```diff
@@ -129,195 +130,14 @@ impl HttpSolver {
...
-            result.insert(*index, order);
-        }
-        result
-    }
-
-    fn map_amm_orders_for_solver<T>(&self, orders: Vec<T>) -> HashMap<usize, T> {
-        orders.into_iter().enumerate().collect()
-    }
-
-    fn amm_models(
```
nlordell, line 214: Moved :point_down:.

solver/src/solver/http_solver.rs
```diff
@@ -484,23 +283,6 @@ impl HttpSolver {
             .with_context(|| format!("failed to decode response json, {}", context()))
     }
 
-    fn order_cost(&self, gas_price: f64) -> U256 {
```
nlordell, line 487: Moved :point_down: to member functions of `GasModel` to make it easier to unit test (specifically, its really easy to create a `GasModel` instance to test mapping of `LimitOrder`s and `Liquidity` to HTTP solver specific representations without needing a full `HttpSolver` instance).

solver/src/solver/http_solver.rs
```diff
@@ -512,43 +294,191 @@ impl HttpSolver {
     }
 }
 
-fn split_liquidity(
```
nlordell, line 515: Removed.

solver/src/solver/http_solver.rs
```diff
@@ -512,43 +294,191 @@ impl HttpSolver {
...
+                    fee: gas_model.order_fee(order),
+                    cost: gas_model.order_cost(),
+                    is_liquidity_order: order.is_liquidity_order,
+                },
+            ))
+        })
+        .collect()
+}
+
+fn amm_models(liquidity: &[Liquidity], gas_model: &GasModel) -> BTreeMap<usize, AmmModel> {
```
nlordell, line 414: :meat_on_bone: of the PR, no longer requires splitting `Liquidity` into separate collections for each liquidity kind.

solver/src/solver/http_solver.rs
```diff
@@ -512,43 +294,191 @@ impl HttpSolver {
...
+}
+
+fn compute_fee_connected_tokens(liquidity: &[Liquidity], native_token: H160) -> HashSet<H160> {
     // Find all tokens that are connected through potentially multiple amm hops to the fee.
     // TODO: Replace with a more optimal graph algorithm.
-    let mut amms = amms.iter().map(|amm| amm.tokens).collect::<HashSet<_>>();
-    let mut fee_connected_tokens = std::iter::once(*native_token).collect::<HashSet<_>>();
+    let mut pairs = liquidity
+        .iter()
+        .flat_map(|amm| amm.all_token_pairs())
```
nlordell, line 476: Fix for #1071

solver/src/solver/http_solver/settlement.rs
```diff
@@ -139,67 +125,25 @@ fn match_prepared_and_settled_orders(
 }
 
 fn match_prepared_and_settled_amms(
```
nlordell, line 127: Other side of the :meat_on_bone:, we no longer need to deal with split liquidity.

solver/src/solver/http_solver/settlement.rs
```diff
@@ -607,70 +549,37 @@ mod tests {
...
-        assert_eq!(executed_wp_order.reserves, weighted_product_order.reserves);
-        assert_eq!(executed_wp_order.fee, weighted_product_order.fee);
-        assert_eq!(
-            prepared_amms[0].input,
-            (token_c, U256::from(996570293625184642u128))
-        );
+        let prepared_amms =
+            match_prepared_and_settled_amms(liquidity, solution_response.amms).unwrap();
+
         assert_eq!(
```
nlordell, line 559: These assertions are the same as before, just in a much more succinct way.

solver/src/solver/http_solver/settlement.rs
```diff
@@ -576,8 +518,8 @@ mod tests {
                         {
                             "sell_token": "0xc778417e063141139fce010982780140aa0cd5ab",
                             "buy_token": "0xe4b9895e638f54c3bee2a3a78d6a297cc03e0353",
-                            "exec_sell_amount": "1",
-                            "exec_buy_amount": "2",
+                            "exec_sell_amount": "3",
+                            "exec_buy_amount": "4",
```
bh2smith, line 522: Any particular reason for this? Or just for fun.

solver/src/solver/http_solver.rs
```diff
@@ -368,47 +188,32 @@ impl HttpSolver {
             })
             .collect();
 
-        let mut new_price_estimates: HashMap<H160, f64> = queries
+        let mut all_price_estimates: HashMap<H160, f64> = queries
             .iter()
             .zip(new_price_estimates)
```
bh2smith, line 373: Does new price estimates still exist? Seems weird it's rebranded and the old one is still being used.

bh2smith (approved):
Amazing! Bravo. So my only question is then how is the liquidity expected to be "sorted" with this BTree map? Is it based on a particular field of amm model?

solver/src/solver/http_solver/settlement.rs
nlordell: So that all AMMs have different input/output values.

solver/src/solver/http_solver.rs
nlordell: Previously, we had `price_estimates`, which were the price estimates for all the tokens included in GPv2 orders, `new_price_estimates` were price estimates for all the tokens in the batch (so, including tokens for from AMMs but that aren't included in any orders).

The previous code would combine `price_estimates` and `new_price_estimates` into `new_price_estimates`, which I thought was not an ideal name, and instead renamed it to `all_price_estimates` (since it is the combination of all the price estimates, the ones from the batch used for the objective function, and the extra ones that the HTTP solver wants).

nlordell:
> expected to be "sorted" with this BTree map

`BTreeMap` is sorted by key. So, if you look at the instance now:
```js
{
  tokens: {
    // tokens appear sorted:
    "0x000...000": { /* ... */ },
    "0x111...111": { /* ... */ },
  },
  orders: {
    // orders sorted by index:
    "0": { /* ... */ },
    // with potentially "skipped" orders..., so no "1"-"4".
    "5": { /* ... */ },
  },
  amms: {
    // same for AMMs:
    "0": { /* ... */ },
    "1": { /* ... */ },
    "3": { /* ... */ },
  },
}
```

vkgnosis (approved):
Nice

bh2smith:
> > expected to be "sorted" with this BTree map
> 
> `BTreeMap` is sorted by key. So, if you look at the instance now:

If they are sorted by keys and the keys are integers as strings, wont we run into an issue with the ordering once we reach beyond single digits, "0", "1", "10", "2", ...?

vkgnosis:
I think the point is just to make it consistent. The actual order isn't  important.

nlordell:
> wont we run into an issue with the ordering once we reach beyond single digits, "0", "1", "10", "11",..., "2", ...?

The keys are `usize`, they are just serialized as JSON strings. The order should be sorted by `usize`, so: "0", "1", "2", ..., "10", "11"?

PR #1143, opened 2021-09-22
New fee and amount quote API

nlordell:
This PR proposes a new API enpoint for quoting fees and prices for orders in a way that more complete order data is specified. This allows the fee quoting mechanism to take app ID as well as balance sources (i.e. Vault balances for example) to be taken into account. In the future, things like the order owner can also be checked to see if they are staking some GP tokens in order to give them better fees as well.

### Test Plan

This PR just introduces a new end point and the OpenAPI specification for it. The implementation will come in a future PR.

ALso, you can test that the new route returns a 500 error with message "not implemented":
```
$ curl -s \
  -X POST \
  --data '{"from":"0x0101010101010101010101010101010101010101","sellToken":"0x0202020202020202020202020202020202020202","buyToken":"0x0303030303030303030303030303030303030303","receiver":"0x0404040404040404040404040404040404040404","kind":"buy","buyAmountAfterFee":"1337","validTo":12345678,"appData":"0x9090909090909090909090909090909090909090909090909090909090909090","partiallyFillable":false}' \
  'http://localhost:8080/api/v1/quote'
{"errorType":"InternalServerError","description":"not yet implemented"}
```

### DIFF
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -425,15 +425,6 @@ paths:
         a price estimate for the order. It returns a full order that can be used
         directly for signing, and with an included signature, passed directly to
         the order creation endpoint.
-      parameters:
-        - in: header
-          name: X-AppData
-          description: |
-            The front-end specific application data. This allows the backend to
-            differentiate the front-end used to quote the order even in the presence
-            of a referral.
-          $ref: "#/components/schemas/AppData"
-          required: false
       requestBody:
         description: The order parameters to compute a quote for.
         required: true
@@ -823,31 +814,43 @@ components:
       description: The buy or sell side when quoting an order.
       oneOf:
         - type: object
+          description: Quote a sell order given the final total sell amount including fees
           properties:
             kind:
-              description: "A sell order"
               type: string
               enum: [sell]
-            totalSellAmount:
+            sellAmountBeforeFee:
               description: |
-                The total sell amount for the quoted order. The sell amount for the
-                returned orders will have fees deducted from this value.
+                The total amount that is available for the order. From this value, the fee
+                is deducted and the buy amount is calculated.
               $ref: "#/components/schemas/TokenAmount"
           required:
             - kind
-            - totalSellAmount
+            - sellAmountBeforeFee
         - type: object
+          description: Quote a sell order given the sell amount.
           properties:
             kind:
-              description: "A buy order"
+              type: string
+              enum: [sell]
+            sellAmountAfterFee:
+              description: The sell amount for the order.
+              $ref: "#/components/schemas/TokenAmount"
+          required:
+            - kind
+            - sellAmountAfterFee
+        - type: object
+          description: Quote a buy order given an exact buy amount.
+          properties:
+            kind:
               type: string
               enum: [buy]
-            buyAmount:
+            buyAmountAfterFee:
               description: The buy amount for the order.
               $ref: "#/components/schemas/TokenAmount"
           required:
             - kind
-            - buyAmount
+            - buyAmountAfterFee
     OrderQuoteRequest:
       description: Request for generating a fee and price quote.
       allOf:
@@ -898,7 +901,8 @@ components:
         - $ref: "#/components/schemas/OrderParameters"
         - type: object
           properties:
-            - from: The expected address of the signer
+            from:
+              description: The expected address of the signer
               $ref: "#/components/schemas/Address"
           required:
             - from
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -448,9 +448,9 @@ paths:
         403:
           description: Forbidden, your account is deny-listed
         429:
-          description: Too many order placements
+          description: Too many order quotes
         500:
-          description: Error quoting an order
+          description: Unexpected error quoting an order
 components:
   schemas:
     TransactionHash:
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -852,7 +852,7 @@ components:
             - kind
             - buyAmountAfterFee
     OrderQuoteRequest:
-      description: Request for generating a fee and price quote.
+      description: Request fee and price quote.
       allOf:
         - $ref: "#/components/schemas/OrderQuoteSide"
         - type: object
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -417,6 +417,49 @@ paths:
                   $ref: "#/components/schemas/Order"
         400:
           description: Problem with parameters like limit being too large.
+  /api/v1/quote:
+    post:
+      summary: Quotes a price and fee for the specified order parameters.
+      description: |
+        This API endpoint accepts a partial order and computes the minimum fee and
+        a price estimate for the order. It returns a full order that can be used
+        directly for signing, and with an included signature, passed directly to
+        the order creation endpoint.
+      parameters:
+        - in: header
+          name: X-AppData
+          description: |
+            The front-end specific application data. This allows the backend to
+            differentiate the front-end used to quote the order even in the presence
+            of a referral.
+          $ref: "#/components/schemas/AppData"
+          required: false
+      requestBody:
+        description: The order parameters to compute a quote for.
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/OrderQuoteRequest"
+      responses:
+        200:
+          description: Quoted order.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/OrderQuote"
+        400:
+          description: Error quoting order.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/FeeAndQuoteError"
+        403:
+          description: Forbidden, your account is deny-listed
+        429:
+          description: Too many order placements
+        500:
+          description: Error quoting an order
 components:
   schemas:
     TransactionHash:
@@ -493,8 +536,8 @@ components:
       description: The current order status
       type: string
       enum: [presignaturePending, open, fulfilled, cancelled, expired]
-    OrderCreation:
-      description: Data a user provides when creating a new order.
+    OrderParameters:
+      description: Order parameters.
       type: object
       properties:
         sellToken:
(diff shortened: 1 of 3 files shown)


orderbook/openapi.yml
```diff
@@ -769,3 +819,86 @@ components:
       required:
         - errorType
         - description
+    OrderQuoteSide:
+      description: The buy or sell side when quoting an order.
+      oneOf:
```
fleupold, line 824: Should we differentiate between sell before fees, sell after fees? I could imagine a market maker that wants to match a buy order trying to sell exactly the amount that is being bought and then add the fees on top (I agree this is less likely than the one we already have but could still be useful).

orderbook/openapi.yml
```diff
@@ -417,6 +417,49 @@ paths:
...
+    post:
+      summary: Quotes a price and fee for the specified order parameters.
+      description: |
+        This API endpoint accepts a partial order and computes the minimum fee and
+        a price estimate for the order. It returns a full order that can be used
+        directly for signing, and with an included signature, passed directly to
+        the order creation endpoint.
+      parameters:
+        - in: header
+          name: X-AppData
```
fleupold, line 430: This is to avoid the IPFS hash lookup when receiving e.g. a CowSwap app data that also contains a referrer (making the AppData hash unique), right? 
I think that we have to resolve the IPFS hash anyways when verifying the order's fee upon placement and given this can be easily cached, I'm actually no longer sure we should provide the raw interface ID as an extra param.

fleupold (commented):
I like that this will give us full control to implement arbitrary fee logic based on the order data (without requiring more updates in the frontends). We just need to make sure we consolidate/clean up our various fee endpoints.

orderbook/openapi.yml
nlordell: Sounds good. We can also add this back in if we run into technical issues with IPFS. Lets leave it out for now.

nlordell:
> We just need to make sure we consolidate/clean up our various fee endpoints.

We can add a `DEPRECATED` tag to the old fee end points, and then refactor so that they ultimately route to the new endpoint under the hood.

vkgnosis (approved):
I like this api and also think we should remove the other fee related endpoints then. Would like to hear from frontend too what they think about this api.

orderbook/openapi.yml
```diff
@@ -417,6 +417,49 @@ paths:
...
+          description: |
+            The front-end specific application data. This allows the backend to
```
anxolin: nit: I wouldn't use front-end, since normally has UI connotations. Maybe `client`?

orderbook/openapi.yml
nlordell: Added. You should now be able to `POST`:
```js
{
  // bunch of order stuff...
  // Either
  kind: "sell",
  sellAmountBeforeFee: "100000",
  // Or
  kind: "sell",
  sellAmountAfterFee: "100000",
  // Or
  kind: "buy",
  buyAmountAfterFee: "100000",
}
```

The names were chosen to be consistent with the rest of the API.

orderbook/openapi.yml
anxolin: Why X prefix?

that naming convention  is usually used for header params, but not for the normal params, right?

orderbook/openapi.yml
```diff
@@ -417,6 +417,49 @@ paths:
...
           description: Problem with parameters like limit being too large.
+  /api/v1/quote:
+    post:
+      summary: Quotes a price and fee for the specified order parameters.
+      description: |
+        This API endpoint accepts a partial order and computes the minimum fee and
+        a price estimate for the order. It returns a full order that can be used
+        directly for signing, and with an included signature, passed directly to
+        the order creation endpoint.
+      parameters:
```
anxolin: I find a bit strange to mix query params and body info. I would stick to `requestBody` and add the additional info as optional fields there

orderbook/openapi.yml
```diff
@@ -769,3 +819,86 @@ components:
...
+              type: string
+              enum: [buy]
+            buyAmount:
+              description: The buy amount for the order.
+              $ref: "#/components/schemas/TokenAmount"
+          required:
+            - kind
+            - buyAmount
+    OrderQuoteRequest:
+      description: Request for generating a fee and price quote.
```
anxolin, line 855: `generating`? Maybe just `Request fee and price quote.` is less confusing

anxolin (commented):
Not sure if you are aware but we already provide the AppId using a custom header: 
image

I think this can be an "official" option too, to allow to provide custom headers. 

Happy to make it part of the params, so it's even more official as you suggest in this PR

orderbook/openapi.yml
nlordell: This was a header param. We have since removed it.

orderbook/openapi.yml
nlordell: This was a header parameter. It has been removed though.

nlordell:
> I think this can be an "official" option too, to allow to provide custom headers.

@anxolin  What we ultimately want though, is to have custom logic for determining fees based on the full order parameters and not just the app ID. So in this case, the header is not enough.

For example, we may want to check what the `sellTokenBalance` and `buyTokenBalance` are, so order that trade internal buffer balances tokenized on L2 would pay 0 fees (Martin's idea from the retreat). Another example would be to also have a fee discount based on the user address, so a user that stakes GP token gets a discount.

So the idea of the API is to pass as much of the order data as possible to the price/fee computation end point. This way we are more "future proof" and don't need to keep adding new header parameters as we consider different order parameters for determining the fee subsidy. I hope this makes sense :smile:.

orderbook/openapi.yml
```diff
@@ -769,3 +810,99 @@ components:
...
+        - $ref: "#/components/schemas/OrderQuoteSide"
+        - type: object
+          properties:
+            sellToken:
+              description: "ERC20 token to be sold"
+              $ref: "#/components/schemas/Address"
+            buyToken:
+              description: "ERC20 token to be bought"
+              $ref: "#/components/schemas/Address"
+            receiver:
```
alfetopito, line 866: Is the receiver relevant for the price quote?

bh2smith: While it may not be needed for the quote, it will be needed to construct the singable order that is meant to be returned from this endpoint.

bh2smith (approved):
Love this. Happy to implement the test once this has landed!

orderbook/openapi.yml
nlordell: We may offer discounts if the receiver address stakes GP tokens for example (and not just the owner).

bh2smith:
If this is essentially considered set in stone, then I would start implementing this feature on top of this branch. Will wait, however, on the go ahead.

orderbook/openapi.yml
alfetopito: For the receiver as well as the owner? Wow, ok

orderbook/openapi.yml
anxolin: My bad, i thought it was a parameter 👌

anxolin:
> This way we are more "future proof" and don't need to keep adding new header parameters as we consider different order parameters for determining the fee subsidy. I hope this makes sense

Sure it does 👌

orderbook/openapi.yml
```diff
@@ -417,6 +417,40 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/FeeAndQuoteError"
+        403:
+          description: Forbidden, your account is deny-listed
+        429:
+          description: Too many order placements
+        500:
+          description: Error quoting an order
```
bh2smith, line 453: I can't see the difference here between 400 and 500 errors - both are "Error quoting order". 

Also, is the pattern for `#/components/schemas/FeeAndQuoteError` defined here, because it doesn't appear to be.

bh2smith: Might be cute to encapsulate all these different specifics into the single FeeAndQuoteError which i imagine is the intention here. Something like 

```rust
pub enum PostQuoteResult {
    QuoteConstructed(OrderQuote),
    FeeAndQuoteError(FeeError),
    Forbidden,
}

pub enum FeeError {
    UnsupportedToken(H160),
    TransferEthToContract,
    SameBuyAndSellToken,
    UnsupportedBuyTokenDestination(BuyTokenDestination),
    UnsupportedSellTokenSource(SellTokenSource),
    ZeroAmount,
}
```

nlordell: `#/components/schemas/FeeAndQuoteError` already exists (ans is used by the `feeAndQuote` route). Since the errors are the same, I just reused the error name for now.

Additionally, we would probably reuse the error type from the existing `feeAndQuote` route.

bh2smith: That sounds reasonable, but what about all these other responses that seem out of place

```
        403:
          description: Forbidden, your account is deny-listed
        429:
          description: Too many order placements
        500:
          description: Error quoting an order
```

as I see it, the only things we will need here are 

```
#[derive(Serialize)]
pub enum OrderQuoteResult {
    QuoteConstructed(OrderQuote),
    FeeAndQuoteError(FeeAndQuoteError),
}
```
along with some possible rate limiting...

nlordell: > along with some possible rate limiting...

That is what the 429 is for AFIAU:
```
        429:
          description: Too many order placements # i.e you quoted too many orders and are being rate limited. We should `s/placements/quotes/` to make the error a bit clearer.
```

Additionally, `500` is used as a catch all "internal server error" - i.e. something unexpected went wrong.

Finally:
```
403:
          description: Forbidden, your account is deny-listed
```

Since the `from` is passed in, we can also check that the owner is forbidden by address (we have an "owner deny list" already).

bh2smith:
I have opened a draft PR #1155 based on this branch which essentially fills in the unimplemented code routes with the existing fee calculation logic. I know there is an issue listing some of the requirements for the new fee logic, but can't seem to find it.

nlordell:
> based on this branch which essentially fills in the unimplemented code routes with the existing fee calculation logic

I will do some final fix ups and merge this then.

orderbook/openapi.yml
nlordell: Fixed 429 description and added "unexpected" to 500 description.

bh2smith: There appears to be something missing in the description of 429 here in your comment...

```
We should `s/placements/quotes/` to make the error a bit clearer.
```

nlordell: It was changed to:
```
        429:
          description: Too many order quotes
```

anxolin:
@nlordell let us know when we this hits production, or when u want us to start using the new endpoint

Also, we should let Balancer know too, right?

nlordell:
> Also, we should let Balancer know too, right?

Its still a little off from being implemented (probably ~ 2 weeks away).

PR #1144, opened 2021-09-22
Priority price estimator

### LINKED ISSUE #1139: Use Baseline price estimation as fallback?
Like the gas estimators we might want to use the baseline as a fallback when Paraswap returns an error. On the other hand that could make price and gas estimates fluctuate a lot depending on whether Paraswap works or not for tokens we are bad at estimating.

vkgnosis:
Implements a priority price estimator that tries several inner estimators. I made this smarter than a naive implementation by only retrying unsuccessful queries and continuing to request bulk estimates.

Fixes #1139 

### Test Plan
new unit test and manual test

### DIFF
--- a/shared/src/price_estimation/paraswap.rs
+++ b/shared/src/price_estimation/paraswap.rs
@@ -1,7 +1,7 @@
 use super::{ensure_token_supported, Estimate, PriceEstimating, PriceEstimationError, Query};
 use crate::{
     bad_token::BadTokenDetecting,
-    paraswap_api::{ParaswapApi, PriceQuery, Side},
+    paraswap_api::{ParaswapApi, ParaswapResponseError, PriceQuery, Side},
     token_info::{TokenInfo, TokenInfoFetching},
 };
 use anyhow::{anyhow, Context, Result};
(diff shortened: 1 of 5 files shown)


shared/src/price_estimation/paraswap.rs
```diff
@@ -35,28 +35,33 @@ impl ParaswapPriceEstimator {
...
 
         let response = self
             .paraswap
             .price(price_query)
             .await
-            .map_err(anyhow::Error::from)
+            .map_err(|err| match err {
+                ParaswapResponseError::InsufficientLiquidity => PriceEstimationError::NoLiquidity,
+                _ => PriceEstimationError::Other(err.into()),
+            })
```
vkgnosis, line 58: This was needed to forward NoLiquidity errors

shared/src/price_estimation/paraswap.rs
```diff
@@ -35,28 +35,33 @@ impl ParaswapPriceEstimator {
...
-            OrderKind::Buy => (query.buy_token, query.sell_token, Side::Buy),
-            OrderKind::Sell => (query.sell_token, query.buy_token, Side::Sell),
-        };
         let price_query = PriceQuery {
-            src_token,
-            dest_token,
-            src_decimals: decimals(&src_token, token_infos)? as usize,
-            dest_decimals: decimals(&dest_token, token_infos)? as usize,
+            src_token: query.sell_token,
+            dest_token: query.buy_token,
```
vkgnosis, line 40: This is a drive by fix because while testing I noticed I was using src and dest token wrong in this file.

nlordell (approved):

shared/src/price_estimation/priority.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub fn new(inner: Vec<Box<dyn PriceEstimating>>) -> Self {
+        assert!(!inner.is_empty());
+        Self { inner }
+    }
+}
+
+#[async_trait::async_trait]
+impl PriceEstimating for PriorityPriceEstimator {
+    async fn estimates(&self, queries: &[Query]) -> Vec<Result<Estimate, PriceEstimationError>> {
+        let mut results = self.inner[0].estimates(queries).await;
```
nlordell, line 20: Cleaver. It only retries failed queries instead of an "all or nothing" approach. :+1:

nlordell:
This fixes #1139 right?

vkgnosis:
yes

PR #1146, opened 2021-09-22
use AppId with proper formatting for partner_fee_factors

### LINKED ISSUE #1145: Use appropriate type for partner-fee-factor argument
Based on the discussions in #1125, it is clear that the use of `H256` in the orderbook runtime argument `partner_additional_fee_factors: HashMap<H256, f64>` is somewhat of a hack (since `app_data` is actually `[u8; 32]` and not really a hash. To implement this properly, it was suggested that we intake a list of "tuples" of the form 

```sh
--partner-additional-fee-factors "0x0000000:0.5,0x0000:0.3"
```

and write a custom parser which converts this to `HashMap<[u8;32], f64>` instead of "lazily" relying on json string parsing.

bh2smith:
Closes #1145 

Instead of H256, we define a new type AppId that is used for `partner_additional_fee_factors` along with a corresponding parse method.

Note that this also give rise to a new open task (https://github.com/gnosis/gp-v2-services/issues/1147) of updating the `OrderCreation` struct to use `AppId` in its `app_data` field.


orderbook/src/main.rs
```diff
@@ -434,3 +436,19 @@ async fn check_database_connection(orderbook: &Orderbook) {
         .await
         .expect("failed to connect to database");
 }
+
+fn parse_partner_fee_factor(s: &str) -> Result<HashMap<AppId, f64>> {
```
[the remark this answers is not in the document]
bh2smith: I like the sound of a doc test.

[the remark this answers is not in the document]
nlordell: I don't think a doc-test is appropriate here, but I do agree that there should be a unit test.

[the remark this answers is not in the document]
bh2smith: Added tests. ALso tried adding a doc test but it wouldn't render or run.

```
   Doc-tests shared

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```


With this (that has now been removed).

```
///
/// # Examples
///
/// ```rust
/// let s = "0x0000000000000000000000000000000000000000000000000000000000000000:0.5,0x0101010101010101010101010101010101010101010101010101010101010101:0.7";
/// assert_eq!(parse_partner_fee_factor(s).unwrap(), hashmap! { AppId([0u8; 32]) => 0.5, AppId([1u8; 32]) => 0.7 });
/// ```
```

shared/src/lib.rs
```diff
@@ -83,3 +84,12 @@ pub fn debug_bytes(
 pub fn clone_anyhow_error(err: &anyhow::Error) -> anyhow::Error {
     anyhow::anyhow!("{:#}", err)
 }
+
+#[derive(Eq, Hash, PartialEq)]
+pub struct AppId(pub [u8; 32]);
+
+impl Debug for AppId {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
```
nlordell, line 92: Just so you know, the `<'_>` can li4kely be omitted here:
```suggestion
    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
```

nlordell (approved):
Love it! Can you also make `OrderCreation` use `AppId` instead of `[u8; 32]` (doesn't have to be in this PR).

orderbook/src/main.rs
```diff
@@ -434,3 +436,84 @@ async fn check_database_connection(orderbook: &Orderbook) {
...
+            .map(|x| x.trim())
+            .collect::<Vec<&str>>();
+        if split_pair.len() != 2 {
+            return Err(anyhow!("Invalid pair lengths"));
+        }
+        let mut key_bytes = [0u8; 32];
+        hex::decode_to_slice(
+            split_pair[0].strip_prefix("0x").unwrap_or(split_pair[0]),
+            &mut key_bytes,
+        )?;
```
nlordell, line 457: One minor suggestion, you could:
```rust
impl FromStr for AppId {
  type Err = hex::FromHexError;
  fn from_str(s: &str) -> Result<Self, Self::Err> {
        let mut key_bytes = [0u8; 32];
        hex::decode_to_slice(
            s.strip_prefix("0x").unwrap_or(s),
            &mut key_bytes,
        )?;
        Ok(Self(key_bytes))
  }
}
```

bh2smith:
> Love it! Can you also make `OrderCreation` use `AppId` instead of `[u8; 32]` (doesn't have to be in this PR).

I have already created an issue for this (mentioned in PR description as well #1147)

orderbook/src/main.rs
bh2smith: Very cool! Will do.

orderbook/src/main.rs
```diff
@@ -434,3 +436,84 @@ async fn check_database_connection(orderbook: &Orderbook) {
         .await
         .expect("failed to connect to database");
 }
+
+/// Parses a comma separated list of colon separated values representing fee factors for AppIds.
+fn parse_partner_fee_factor(s: &str) -> Result<HashMap<AppId, f64>> {
+    let mut res = HashMap::default();
+    for pair_str in s.split(',').into_iter().collect::<Vec<&str>>() {
```
vkgnosis, line 443: No need to `collect` in this line. We are only iterating so we don't need a Vec.

orderbook/src/main.rs
```diff
@@ -434,3 +436,84 @@ async fn check_database_connection(orderbook: &Orderbook) {
...
+        let split_pair = pair_str
+            .trim()
+            .split(':')
+            .into_iter()
```
vkgnosis, line 447: No need for into_iter because the result of split is already an iterator.

orderbook/src/main.rs
```diff
@@ -434,3 +436,84 @@ async fn check_database_connection(orderbook: &Orderbook) {
...
+            .map(|x| x.trim())
+            .collect::<Vec<&str>>();
```
vkgnosis, line 449: Not very important but we could also avoid allocating a vector here by keeping this as an iterator and doing `.next()` twice manually and failing if next returns none. Then call next a third time and fail if it isn't None.

vkgnosis (approved):

PR #1148, opened 2021-09-22
Still merge settlements if prices are scalable

fleupold:
Fixes issue describe [here on slack](https://gnosisinc.slack.com/archives/CC3C5SMB5/p1632135697012100?thread_ts=1632086177.005000&cid=CC3C5SMB5)

We currently don't merge settlements as soon as two settlements differ on a single token. This is overly harsh, because e.g. a settlement matching a USDC -> WETH trade and another solution settling and USDT -> WETH trade could still be merged, even if WETH had a different reference price in both solutions (the price of WETH in solution 2 could e.g. be "scaled" to the same price as in solution 1, the same scaling would then have to apply to the USDT).

This PR makes it so that we extract the scaling factor of the two price vectors (based on the first common token, 1 if no token is in common) and then try to scale all prices accordingly. I'm hoping that with the use of BigRationals I'm avoiding rounding errors (although we could likely live with a rounding error of up to 1bip due to buffers)

### Test Plan
Added a unit test

### DIFF
--- a/solver/src/settlement/settlement_encoder.rs
+++ b/solver/src/settlement/settlement_encoder.rs
@@ -4,9 +4,9 @@ use anyhow::{bail, ensure, Context as _, Result};
 use model::order::{Order, OrderKind};
 use num::{BigRational, Zero};
 use primitive_types::{H160, U256};
-use shared::conversions::U256Ext;
+use shared::conversions::{big_rational_to_u256, U256Ext};
 use std::{
-    collections::{hash_map::Entry, HashMap},
+    collections::{hash_map::Entry, HashMap, HashSet},
     iter,
     sync::Arc,
 };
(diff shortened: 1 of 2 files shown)


josojo (approved):
Nice, I had this code also already in place. :) 
Though, I have not yet pushed it, as I tried to add simulations after each merge, and only finalise the merge, if the simulation is successful. 
There are two reasons: 
- by rounding the prices, the output for the user could become a tiny bit bigger. If the order basically had zero surplus, then the merged output could fail.
- Sometimes dex-ag solutions are complicated and fail already during simulations. By merging many of them together without simulation check, on increases the likelyhood of failure.

But the drawback of simulations is that they cost extra time...

Anyways, the PR is good already as it is and we can merge in my opinion, as in most cases the merge settlements will not fail. If you guys think it makes sense to simulate during the merge, I could add it in another PR

solver/src/settlement/settlement_encoder.rs
```diff
@@ -259,13 +259,20 @@ impl SettlementEncoder {
 
     // Merge other into self so that the result contains both settlements.
     // Fails if the settlements cannot be merged for example because the same limit order is used in
-    // both or the same token has different clearing prices.
+    // both or more than one token has a different clearing prices (a single token difference is scaled)
     pub fn merge(mut self, mut other: Self) -> Result<Self> {
+        let scaling_factor = self.price_scaling_factor(&other);
         for (key, value) in other.clearing_prices {
+            let scaled_price =
+                big_rational_to_u256(&(value.to_big_rational() * scaling_factor.clone()))
```
vkgnosis, line 267: Might be able to do `* &scaling_factor` instead of clone.

vkgnosis (approved):

bh2smith (approved):
Seems reasonable. Looking forward to the day we can merge even more!

fleupold:
> If you guys think it makes sense to simulate during the merge, I could add it in another PR

I think simulating is a good idea, specifically since we don't "test" solutions for validity beforehand (and the MiP solver may have taken some time before returning, so e.g. the Paraswap transaction might be outdated).

I wouldn't be so worried about the rounding though as I think we already have some buffer for division accuracy in our solvers (I would risk to see limit price violation errors).

josojo:
One more comment:
curreently, we do:
```
solver_settlements::merge_settlements(
                self.max_merged_settlements,
                &estimated_prices,
                &mut settlements,
            );

            solver_settlements::filter_settlements_without_old_orders(
                self.min_order_age,
                &mut settlements,
            );
         ```
         in the driver. I think we should do it the other way around:
         ```
         

            solver_settlements::filter_settlements_without_old_orders(
                self.min_order_age,
                &mut settlements,
            );

solver_settlements::merge_settlements(
                self.max_merged_settlements,
                &estimated_prices,
                &mut settlements,
            );
            ```
            
            to make sure that we do not merge old orders and then just skip all orders after the merge because of one old order

fleupold:
> to make sure that we do not merge old orders and then just skip all orders after the merge because of one old order

We include solutions as long as **one** order is old. So merging before filtering increases the chances of settling larger batches.

PR #1150, opened 2021-09-23
Fix default partner fee factor and handle empty string

vkgnosis:
Made api not start up when the argument isn't specified.

### Test Plan
new unit test, manually ran


fedgiac (approved):

bh2smith (commented):
Whoops! Good catch. I had actually considered this before bed yesterday, but forgot to get it in there today...

PR #1151, opened 2021-09-23
Allow price estimator argument to be env var

vkgnosis:
Forgot to make this env before.

### Test Plan
manually ran


bh2smith (approved):
Kewl.

PR #1155, opened 2021-09-24
Fee&OrderQuote Endpoint: Implemention with legacy logic.

bh2smith:
Depends on #1169

based on the specification defined in #1143, we implement the fee calculating logic of the existing route along with the building of an Order Quote that is returned by this endpoint.

- Note that this also involved a refactor of the legacy fee endpoint to make us of the new one. For this we move the generic `response` method into the new file making it generically operate on all of `OrderQuoteResponse`, `BuyResponse` and  `SellResponse`. We had to open up a few fields in the `OrderQuote` to be public so that we could convert `OrderQuoteResponse` into the legacy `Buy/SellResponse` types. Furthermore, now, the legacy fee endpoint requires and `order_validator` and makes use of partial order validation on the Partial Order constructed from a legacy buy/sell request. Hopefully this minimal data doesn't lead to an order validation Error (or else we may have to introduce a validation override flag of some sort). 

### Test Plan

- [x] Run the api locally and make a request that returns an `OrderQuoteResponse`

With this api request

```sh
curl -X 'POST' \
  -d '{
  "sellToken": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
  "receiver": "0x9008D19f58AAbD9eD0D60971565AA8510560ab41",
  "validTo": 1632933600,
  "partiallyFillable": false,
  "sellTokenBalance": "erc20",
  "buyTokenBalance": "erc20",
  "from": "0x9008D19f58AAbD9eD0D60971565AA8510560ab41",
  "kind": "sell",
  "sellAmountBeforeFee": "99999999999999999999"
}'
```

We get back the following response
```
{
  "from": "0x9008d19f58aabd9ed0d60971565aa8510560ab41",
  "sellToken": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
  "receiver": "0x9008d19f58aabd9ed0d60971565aa8510560ab41",
  "sellAmount": "99999999999999999999",
  "buyAmount": "379816735088461143945",
  "validTo": 1632933600,
  "feeAmount": "31686440800600000",
  "kind": "sell",
  "partiallyFillable": false,
  "sellTokenBalance": "erc20",
  "buyTokenBalance": "erc20"
}
```

- [x] Run the legacy API as well to ensure it still functions properly.


orderbook/src/api/post_quote.rs
```diff
@@ -79,33 +112,210 @@ struct OrderQuote {
...
+                        OrderQuoteSide::Buy { .. } => OrderKind::Buy,
+                    },
+                    partially_fillable: self.partially_fillable,
+                    sell_token_balance: self.sell_token_balance,
+                    buy_token_balance: self.buy_token_balance,
+                })
+            }
+            OrderQuoteSide::Buy { .. } => Err(FeeAndQuoteError::Other(anyhow!(
+                "Can't build Quote for Buy Request with SellResponse"
+            ))),
```
nlordell, line 179: It feels a bit awkward to have multiple matches and require these unreachable branches. Maybe there is a way to refactor so that this isn't needed?

orderbook/src/api/get_fee_and_quote.rs
```diff
@@ -8,70 +8,104 @@ use serde::{Deserialize, Serialize};
 use shared::price_estimation::{self, PriceEstimating, PriceEstimationError};
 use std::convert::Infallible;
 use std::sync::Arc;
+use warp::reply::Json;
 use warp::{hyper::StatusCode, reply, Filter, Rejection, Reply};
 
 #[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
-struct Fee {
+pub struct Fee {
```
nlordell, line 16: I'm not a huge fan of this. Instead of making this fee route public, why not make the logic in `post_quote` public and turn these `{Sell,Buy}Query` into a `QuoteOrderQuery` and then take the resulting `OrderQuote` and convert it into a `{Sell,Buy}Response`.

orderbook/src/api/post_quote.rs
bh2smith: Yea, I would hope so. I will try harder to think of one.

orderbook/src/api/get_fee_and_quote.rs
bh2smith: Yes I would very much like to not make these things public. I will try and do so although I am not aware of any {Sell, Buy}Responses that exist already for which this would prevent making Fee stuff public.

nlordell: My point is that instead of `post_quote` API route using the `get_fee_and_quote` logic, we should move the logic into `post_quote` and make `get_fee_and_quote` use that instead.

bh2smith: Ok, but then wont I just have to make it public in a different place?

nlordell: > I  just have to make it public in a different place?

Yes, this is what I meant with "why not make the logic in `post_quote` public".

orderbook/src/api/post_quote.rs
bh2smith: Fixed and looks much better now. I had a feeling there was something off/silly about those matches.

orderbook/src/api/get_fee_and_quote.rs
bh2smith: Excellent, this is a very good idea and I am in the process of making it a reality. However, I worry that this is going to become a much bigger change than originally expected. May have to split it up...

orderbook/openapi.yml
```diff
@@ -880,18 +880,22 @@ components:
             partiallyFillable:
               description: Is this a fill-or-kill order or a partially fillable order?
               type: boolean
+              default: false
```
vkgnosis, line 883: Why default? I prefer making this explicit so that users have to set it because it can help prevent issues in the future.

vkgnosis: I also see a few `#[serde(default)]` below (added in previous PR) and have the same question there.

bh2smith: I will remove it here and swing back to the other one if you can tell me a bit more specifically which you are referring to.

bh2smith: Are you referring to these two fields?

```
            sellTokenBalance:
              $ref: "#/components/schemas/SellTokenSource"
              default: "erc20"
            buyTokenBalance:
              $ref: "#/components/schemas/BuyTokenDestination"
              default: "erc20"
```

vkgnosis: I feel like none of the order data should have defaults.

bh2smith: All good, I believe these are the only two that remain. Will remove the defaults here.

bh2smith: https://github.com/gnosis/gp-v2-services/pull/1189

orderbook/src/api/get_fee_and_quote.rs
```diff
@@ -1,26 +1,19 @@
...
-struct Fee {
-    #[serde(with = "u256_decimal")]
-    amount: U256,
-    expiration_date: DateTime<Utc>,
-}
-
 #[derive(Deserialize)]
 #[serde(rename_all = "camelCase")]
-struct SellQuery {
+pub struct SellQuery {
```
nlordell, line 16: Does this still need to be `pub`?

orderbook/src/api/get_fee_and_quote.rs
```diff
@@ -30,9 +23,20 @@ struct SellQuery {
...
+            },
+        };
+        OrderQuoteRequest::new(query.sell_token, query.buy_token, side)
+    }
+}
+
 #[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
-struct SellResponse {
+pub struct SellResponse {
```
nlordell, line 39: Same here?

orderbook/src/api/get_fee_and_quote.rs
bh2smith: prolly not... Will try and go through things that were made public and fix it.

orderbook/src/api/get_fee_and_quote.rs
```diff
@@ -232,7 +124,9 @@ pub fn get_fee_and_quote_buy(
         let price_estimator = price_estimator.clone();
         async move {
             Result::<_, Infallible>::Ok(response(
-                calculate_buy(fee_calculator, price_estimator, query).await,
+                OrderQuoteRequest::from(query)
+                    .calculate_fee(fee_calculator, price_estimator)
```
nlordell, line 128: neat!

orderbook/src/api/get_fee_and_quote.rs
```diff
@@ -53,117 +57,25 @@ struct BuyQuery {
...
+            buy_amount_after_fee: query.buy_amount_after_fee,
+        };
+        OrderQuoteRequest::new(query.sell_token, query.buy_token, side)
+    }
+}
+
 #[derive(Serialize)]
 #[serde(rename_all = "camelCase")]
-struct BuyResponse {
+pub struct BuyResponse {
```
nlordell, line 71: I think these structs (this one and `SellResponse`) aren't used at all anymore in favour of `FeeResponse`.

orderbook/src/api/post_quote.rs
```diff
@@ -1,19 +1,96 @@
...
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct FeeResponse {
+    fee: Fee,
+    side: OrderQuoteSide,
+}
+
+#[derive(Debug)]
+pub enum FeeError {
```
nlordell, line 39: Ubernit: Maybe this should be a `quote` error as its an error from the new Quote API?

orderbook/src/api/post_quote.rs
```diff
@@ -1,19 +1,96 @@
...
+    pub amount: U256,
+    pub expiration_date: DateTime<Utc>,
+}
+
+#[derive(Debug, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct FeeResponse {
+    fee: Fee,
+    side: OrderQuoteSide,
+}
```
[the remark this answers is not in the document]
bh2smith: But you told me to move it here!

orderbook/src/api/post_quote.rs
```diff
@@ -79,33 +179,261 @@ struct OrderQuote {
     buy_token_balance: BuyTokenDestination,
 }
 
+#[derive(Debug)]
+pub enum FeeAndQuoteError {
```
nlordell, line 183: Ubernit: 
```suggestion
pub enum QuoteError {
```

orderbook/src/api/post_quote.rs
```diff
@@ -79,33 +179,261 @@ struct OrderQuote {
...
+                SellAmount::BeforeFee { value: sell_amount } => {
+                    Ok((sell_amount, buy_amount_after_fee, OrderKind::Buy))
+                }
+                SellAmount::AfterFee { .. } => Err(FeeAndQuoteError::Other(anyhow!(
+                    "Currently unsupported route"
+                ))),
+            },
+            _ => Err(FeeAndQuoteError::Other(anyhow!(
+                "Can't build Quote from (QuoteRequest,FeeResponse) when QuoteSide have same type."
+            ))),
```
nlordell, line 257: nit: I'm not a huge fan of having these panics in unreachable routes when the code can be refactored to avoid this. I think it may be because the new route uses the old `calculate_fee` logic and then builds a `QuoteResponse` from that. I would suggest instead building a `QuoteResponse` directly and then converting it to either a `BuyResponse` or `SellResponse`.

orderbook/src/api/post_quote.rs
```diff
@@ -79,33 +179,261 @@ struct OrderQuote {
...
+            if let Err(err) = &result {
+                tracing::error!(?err, ?request, "post_quote error");
+            }
+            Result::<_, Infallible>::Ok(post_quote_response(result))
+        }
     })
 }
 
 #[cfg(test)]
 mod tests {
```
nlordell, line 428: Thank you for all the tests :pray:

orderbook/src/api/post_quote.rs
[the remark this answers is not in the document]
nlordell: Sorry if my suggestion wasn't super clear. What I meant is that the logic for price and fee estimation should move here, and then `fee_and_quote` API could make use of the new logic of the new route by doing:
```
# For sell queries:
SellQuery ---[ conversion ]--> OrderQuoteRequest ---[ compute price and quote ]--> OrderQuote ---[ conversion ]--> SellResponse

# For buy queries:
BuyQuery ---[ conversion ]--> OrderQuoteRequest ---[ compute price and quote ]--> OrderQuote ---[ conversion ]--> BuyResponse
```

Maybe the comment here adds a bit more context: https://github.com/gnosis/gp-v2-services/pull/1155#discussion_r721369271

orderbook/src/api/post_quote.rs
bh2smith: I believe we just removed Buy Response and Sell Response, but I can reintroduce them... I don't see how converting them will circumvent this faulty pairing. Of course, we would never pass the wrong info, but the fact remains that the QuoteRequest can be of two types that needs to match with the response type.

nlordell: My point is that in `post_quote` module we are first mapping
```
fn(OrderQuoteRequest) -> FeeResponse
```

And then
```
fn(OrderQuoteRequest, FeeResponse) -> OrderQuote
```

This is the cause of the unreachable branch here, because the inputs `OrderQuoteRequest` and `FeeResponse` have to "match". In order to not have this kind of unreachable branch, you could instead of directly implement:
```
fn(OrderQuoteRequest) -> OrderQuote
```

Then, in the `fee_and_quote` module, you need:
```
fn(SellRequest) -> SellResponse
```

And this can be achieved by first implementing:
```
fn(SellRequest) -> OrderQuoteRequest
```

Then calling the common logic implemented in the `post_quote` module for:
```
fn(OrderQuoteRequest) -> OrderQuote
```

And finally, mapping the resulting `OrderQuote` with:
```
fn(OrderQuote) -> SellResponse
```

bh2smith: Ok... I will try this out although I am not sure where in the mix the `FeeResponse` plays a role. It seems as if you are suggesting to build the `OrderQuote` without having a `FeeResponse`, but this doesn't seems reasonable and we will also wind with and intermediary between `OrderQuoteRequest` and `OrderQuote` (the FeeResponse) Whose side type will have to be matched with Request's side type...

nlordell: For example, in pseudo-code:
```
    fn calculate_quote(
        &self,
        fee_calculator: Arc<dyn MinFeeCalculating>,
    ) -> Result<OrderQuote, OrderQuoteError> {
        let (sell_amount, buy_amount, fee_amount, kind) = match self.side {
            OrderQuoteSide::Sell { sell_amount: SellAmount::BeforeFee { .. } } => { /* same as current calculate_fee */ }
            OrderQuoteSide::Sell { sell_amount: SellAmount::AfterFee { .. } } => { /* same as current calculate_fee */ }
            OrderQuoteSide::Buy { .. } => { /* same as current calculate_fee */ }
        };
        Ok(OrderQuote {
            from: self.from,
            sell_token: self.sell_token,
            buy_token: self.buy_token,
            receiver: self.receiver,
            sell_amount,
            buy_amount,
            valid_to: self.valid_to,
            app_data: self.app_data,
            fee_amount,
            kind,
            partially_fillable: self.partially_fillable,
            sell_token_balance: self.sell_token_balance,
            buy_token_balance: self.buy_token_balance,
        })
    }
```

This way, you only have a single match on the request `side`, and produce an `OrderQuote`.

You can then in `fee_and_quote` module:
```
impl From<SellQuery> for OrderQuoteQuery { ... }
impl From<OrderQuote> for FeeResult { ... }
```

And do:
```
OrderQuoteRequest::from(sell_query).calculate_quote(...).into()
```

The main argument for doing it this way, is that the current `calcualte_fee` method is very much "in the shape" of the old API (the `fee_and_quote` API). This would make future changes/improvements to the `post_quote` API harder, since you would have to keep some compatibility with the legacy `fee_and_quote` API (instead of, as I suggested, implementing the **legacy** `fee_and_quote` API by "wrapping" the new `post_quote` API).

orderbook/src/api/get_fee_and_quote.rs
```diff
@@ -1,21 +1,47 @@
...
+#[derive(Debug)]
+pub enum FeeError {
+    SellAmountDoesNotCoverFee,
+    PriceEstimate(PriceEstimationError),
+}
+
+impl FeeError {
+    pub fn to_warp_reply(&self) -> (Json, StatusCode) {
+        match self {
+            FeeError::PriceEstimate(err) => price_estimation_error_to_warp_reply(err.clone()),
```
vkgnosis, line 35: Can make it consume self  `fn to_warp_reply(self)` then we don't need to clone.

vkgnosis (approved):

orderbook/openapi.yml
```diff
@@ -905,3 +908,14 @@ components:
...
+      type: object
+      properties:
+        quote:
+          $ref: "#/components/schemas/OrderQuote"
+        expirationDate:
+          description: |
+            Expiration date of the offered fee. Order service might not accept
+            the fee after this expiration date. Encoded as ISO 8601 UTC.
+          type: string
+          example: "1985-03-10T18:35:18.814523Z"
```
nlordell, line 921: nit: missing newline at EOF.

orderbook/src/api.rs
```diff
@@ -41,12 +43,19 @@ pub fn handle_all_routes(
...
     let cancel_order = cancel_order::cancel_order(orderbook.clone());
     let get_amount_estimate = get_markets::get_amount_estimate(price_estimator.clone());
-    let get_fee_and_quote_sell =
-        get_fee_and_quote::get_fee_and_quote_sell(fee_calculator.clone(), price_estimator.clone());
-    let get_fee_and_quote_buy =
-        get_fee_and_quote::get_fee_and_quote_buy(fee_calculator, price_estimator.clone());
+    let get_fee_and_quote_sell = get_fee_and_quote::get_fee_and_quote_sell(
+        fee_calculator.clone(),
+        price_estimator.clone(),
+        order_validator.clone(),
```
nlordell, line 49: Maybe we can create a:
```
#[derive(Clone)]
pub struct OrderQuoter {
        fee_calculator: Arc<dyn MinFeeCalculating>,
        order_validator: Arc<dyn OrderValidating>,
}
```

To make passing this group around easier.

orderbook/src/api/get_fee_and_quote.rs
```diff
@@ -1,21 +1,47 @@
...
+pub struct Fee {
     #[serde(with = "u256_decimal")]
-    amount: U256,
-    expiration_date: DateTime<Utc>,
+    pub amount: U256,
+    pub expiration_date: DateTime<Utc>,
+}
+
+#[derive(Debug)]
+pub enum FeeError {
```
nlordell, line 27: This should be in the `post_quote` module seeing as its only ever created there in a private function.

orderbook/openapi.yml
bh2smith: Ooopsie!

orderbook/openapi.yml
```diff
@@ -886,22 +886,32 @@ components:
...
             - appData
             - partiallyFillable
+            - from
     OrderQuote:
       description: |
         An order quoted by the back end that can be directly signed and
         submitted to the order creation backend.
       allOf:
         - $ref: "#/components/schemas/OrderParameters"
         - type: object
```
nlordell, line 904: This can be removed, as we now use `OrderQuoteReponse`.

orderbook/openapi.yml
```diff
@@ -886,22 +886,32 @@ components:
...
-            from:
-              description: The expected address of the signer
-              $ref: "#/components/schemas/Address"
-          required:
-            - from
+    OrderQuoteResponse:
+      type: object
+      properties:
+        quote:
+          $ref: "#/components/schemas/OrderQuote"
```
nlordell, line 909:
```suggestion
          $ref: "#/components/schemas/OrderParameters"
```

orderbook/src/api.rs
```diff
@@ -70,6 +68,27 @@ pub fn handle_all_routes(
...
+        order_validator: Arc<dyn OrderValidating>,
+    ) -> Self {
+        Self {
+            fee_calculator,
+            price_estimator,
+            order_validator,
+        }
+    }
+}
+
```
nlordell, line 91: Why not put this type into the `post_quote` module seeing as that is where it gets used. This way the type is closer to the logic that uses it.

orderbook/src/api/post_quote.rs
```diff
@@ -58,54 +95,253 @@ enum SellAmount {
...
 #[serde(rename_all = "camelCase")]
-struct OrderQuote {
+pub struct OrderQuote {
     from: H160,
     sell_token: H160,
     buy_token: H160,
     receiver: Option<H160>,
     #[serde(with = "u256_decimal")]
-    sell_amount: U256,
+    pub sell_amount: U256,
```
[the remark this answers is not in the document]
nlordell: Maybe for consistency, all the fields can be `pub` then? I think its fine here seeing as `OrderQuote` is "just data" and the data is expected to be used outside of this module. By making only a few fields pub, you are "leaking details" of what other modules (like `fee_and_quote` module) use from this data into here.

nlordell (commented):
Like the new `OrderQuoter` type! Just feel like it should be in `post_quote` module (as that is where its used and the logic that uses it lives).

Some issues with the OpenAPI spec.

Otherwise looks great to me!

orderbook/src/api.rs
vkgnosis: I feel you can make all of the pub and not need pub(crate).

orderbook/src/api/post_quote.rs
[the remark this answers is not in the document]
nlordell: Is this something you think we shouldn't change?

nlordell (approved):

orderbook/src/api/post_quote.rs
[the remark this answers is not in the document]
bh2smith: I have made all field of `OrderQuote` public.

This PR = Ready for landing

PR #1156, opened 2021-09-24
Use AppId in app_data field of OrderCreation

### LINKED ISSUE #1147: Use AppId in OrderCreation.app_data field.
Following #1146, it has become clear that `OrderCreation` struct should be adapted to use `AppId` in its `app_data` field instead of a raw byte array. This will involve updating the serde (de)serialization methods as well.

bh2smith:
Closes #1147 

We change `OrderCreation.app_data` field to be of type AppId for consistency throughout the project.

### Test Plan
No logic changes - existing CI.


model/src/appdata_hexadecimal.rs
```diff
@@ -1,26 +1,27 @@
+use crate::order::AppId;
 use serde::{de, Deserializer, Serializer};
 use std::fmt;
 
-pub fn serialize<S>(value: &[u8; 32], serializer: S) -> Result<S::Ok, S::Error>
+pub fn serialize<S>(value: &AppId, serializer: S) -> Result<S::Ok, S::Error>
```
vkgnosis, line 5: Now that AppId is its own type that we control we `impl Serialize` and Deserialize on it directly without needing this module. The serialization can use FromStr and Display so we don't have to juggle with bytes anymore here.
The serde traits are a bit hairy so I'm fine merging as is too and tackling that later.

vkgnosis (approved):
LGTM. Would double check that the serialization of Order hasn't accidentally changed with this PR. Maybe we already have a test that compares raw json in model, not sure from the top of my head.

model/src/appdata_hexadecimal.rs
bh2smith: I am not sure I entirely follow what would change about this. Like delete this file completely and implement the same stuff somewhere else?

model/src/order.rs
```diff
@@ -239,7 +238,7 @@ impl Default for OrderCreation {
             sell_amount: Default::default(),
             buy_amount: Default::default(),
             valid_to: u32::MAX,
-            app_data: Default::default(),
+            app_data: AppId([0u8; 32]),
```
nlordell, line 241: ubernit: You can probably make `AppId` derive default as well.

model/src/appdata_hexadecimal.rs
nlordell: > The serde traits are a bit hairy so I'm fine merging as is too and tackling that later.

But the `serialize` and `deserialize` methods are basically the implementations for `Serialize::serialize` and `Deserialize::deserialize` right? So it would just mean wrapping these existing implementations in:
```rust
impl Serialize for AppId {
  fn serialize(...) -> ... {
    // the current implementation in this function...
  }
}
```

And do the same for Deserialize.

>  Like delete this file completely and implement the same stuff somewhere else?

It might be worth creating a new module `model/src/appid.rs` with the definition of `AppId` as well as the `FromStr`, `{Des,S}erialize`, etc.

nlordell (commented):
Looks good, I would, however, get rid of the `appdata_hexadecimal` and instead implement the `serde` traits on `AppId` directly.

model/src/order.rs
```diff
@@ -206,9 +206,8 @@ pub struct OrderCreation {
     #[serde(with = "u256_decimal")]
     pub buy_amount: U256,
     pub valid_to: u32,
-    #[derivative(Debug(format_with = "debug_app_data"))]
```
nlordell, line 209: Because we no longer need `Derivative` here anymore, I would get rid of:
```rust
#[derive(Eq, PartialEq, Clone, Derivative, Deserialize, Serialize, Hash)]
#[derivative(Debug)]
```

in favour of
```rust
#[derive(Eq, PartialEq, Clone, Debug, Deserialize, Serialize, Hash)]
```

Previously, `Derivative` was used to override the `Debug` implementation to format the `app_data` field as hexadecimal, but this is no longer needed as the `AppId` newtype implements `Debug` as formatted hexadecimal anyway.

model/src/appdata_hexadecimal.rs
bh2smith: Done!

model/src/app_id.rs
```diff
@@ -0,0 +1,99 @@
...
+        serializer.serialize_str(s)
+    }
+}
+
+impl<'de> Deserialize<'de> for AppId {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
```
nlordell, line 46: ubernit: We can use a `Cow<str>` to potentially avoid additional allocations.
```suggestion
        let s = Cow::<str>::deserialize(deserializer)?;
```

model/src/app_id.rs
```diff
@@ -0,0 +1,99 @@
...
+            AppId::from_str("0x00").unwrap_err().to_string(),
+            "Invalid string length"
+        );
+    }
+
+    #[test]
+    fn deserialize_app_id() {
+        let value = Value::String(
+            "0x0ddeb6e4a814908832cc25d11311c514e7efe6af3c9bafeb0d241129cf7f4d83".to_string(),
+        );
```
nlordell, line 93: ubernit: I find the `json!` macro more ergonomic:
```suggestion
        let value = json!("0x0ddeb6e4a814908832cc25d11311c514e7efe6af3c9bafeb0d241129cf7f4d83");
```

model/src/app_id.rs
```diff
@@ -0,0 +1,99 @@
...
+        assert!(AppId::deserialize(value).is_ok());
+        assert!(AppId::deserialize(Value::String("00".to_string())).is_err());
+        assert!(AppId::deserialize(Value::String("asdf".to_string())).is_err());
+        assert!(AppId::deserialize(Value::String("0x00".to_string())).is_err());
```
nlordell, line 97: Same here:
```suggestion
        assert!(AppId::deserialize(json!("00")).is_err());
        assert!(AppId::deserialize(json!("asdf")).is_err());
        assert!(AppId::deserialize(json!("0x00")).is_err());
```

nlordell (approved):
Looks good! Just some very minor suggestions

PR #1161, opened 2021-09-27
Adding new enforced uniform clearing price flag for quasimodo

josojo:
This Pr uses the [new flag enabled by Marco](https://github.com/gnosis/gp-v2-solver-lib/pull/29) in order to unleash quasimodo again for prod.

testplan:
```
export QUASIMODO_SOLVER_URL=https://optimizer:<pwd>@quasimodo-testnets.dev.gnosisdev.com/                                                    
cargo run -p solver -- \
  --solvers Quasimodo \
  --transaction-strategy DryRun
```

---
@marcovc while coding this, I thought that it is a little bit non-uniform that mip has that constraint by default but quasimodo does not. Should we maybe make it the default for quasimodo as well, just to keep things uniform?

Probably, you did not do that, due to the fact that we actually want to use the surplus for us and don't wanna leave it on the road due to uniform clearing prices, right?


solver/src/solver/http_solver.rs
```diff
@@ -60,6 +61,9 @@ impl SolverConfig {
             "max_nr_exec_orders",
             self.max_nr_exec_orders.to_string().as_str(),
         );
+        if self.uniform_clearing_prices_enforced_via_flag {
+            url.query_pairs_mut().append_pair("ucp_policy", "Enforced");
```
marcovc, line 65: Note that the flag is "--ucp_policy Enforce".

marcovc (approved):

marcovc:
> @marcovc while coding this, I thought that it is a little bit non-uniform that mip has that constraint by default but quasimodo does not. Should we maybe make it the default for quasimodo as well, just to keep things uniform?

Yes. The reason was enforcing UCP over all or orders is the main difference between Mip and Quasimodo. In the Mip solver enforcing UCP over all comes naturally at the expense of being restricted to uniswap AMMs, while in quasimodo is doesn't come natural (it makes it really slow).

I can for sure make the quasimodo parameter "--ucp_policy" default to "Enforce", but then when the fix for this surplus issue is merged we will need to pass "--ucp_policy EnforceForOrders" making it a non-uniform API again. 

Maybe a cleaner solution would be to have MipSolverConfig and a QuasimodoSolverConfig in the driver?

josojo:
>Maybe a cleaner solution would be to have MipSolverConfig and a QuasimodoSolverConfig in the driver?

Yeah, this PR builds already these two configures. They are just not yet configurable via kubernetes. This can be done in another PR.

solver/src/solver/http_solver.rs
```diff
@@ -52,6 +52,7 @@ lazy_static! {
 #[derive(Debug, Default)]
 pub struct SolverConfig {
     pub max_nr_exec_orders: u32,
+    pub uniform_clearing_prices_enforced_via_flag: bool,
```
bh2smith, line 55: I feel like this field name could be shortened to something like `enforce_uniform_clearing_prices`. Particularly, the that that it is a flag is inherent from the type being boolean.

bh2smith (approved):
All good from my side except the name of the field.

solver/src/solver/http_solver.rs
vkgnosis: Or use the same name from the query.

vkgnosis (approved):

solver/src/solver/http_solver.rs
nlordell: Agreed. I would either call it:
```suggestion
    pub enforce_uniform_clearing_prices: bool,
```

Or as @vkgnosis suggested:
```rust
pub struct SolverConfig {
    pub max_nr_exec_orders: u32,
    pub uniform_clearing_price_policy: UcpPolicy,
}

pub enum UcpPolicy {
    OrdersOnly,
    Enforce,
}
```

nlordell (approved):

solver/src/solver/http_solver.rs
josojo: thanks for the feedback. I changed it.

The reason I wasn't going for this is at the beginning was that after this change am calling:
```
SolverType::Mip => shared(create_http_solver(
                    account,
                    mip_solver_url.clone(),
                    "Mip",
                    SolverConfig {
                        max_nr_exec_orders: 100,
                        enforce_uniform_clearing_prices: false,
                    },
                ))
                ```
                and then it reads like the uniform clearing prices would not be enforced on the MIP solver, but they are the default for the mip solver and hence they are enforced. That is the reason I was stating : via_flag

PR #1166, opened 2021-09-27
Add support for 1559 transactions

### LINKED ISSUE #1056: Implement EIP1559 for solution submission
Currently we still submit transactions with the legacy transaction type. This task captures submitting EIP 1559 solutions on networks that support it (might require changes to some of our dependencies).

Not a super high priority right not but maybe a good starter task for someone trying to find their way through our library stack. 

Would make #1036 obsolete.

sunce86:
Previously, this code was working only with legacy gas price (single f64 value).
Now, it leverages the new gas price estimators with 1559 support. In case the gas estimators have 1559 support, then the simulation and submit of settlements will be done with 1559 gas price, if not, fallbacks to legacy gas price.

One important thing to notice: previously it was assumed that transaction sent with `x` gas price will either bi mined with `x` gas price or it will stay in transaction pool waiting to be mined or speed up. That said, `effective_gas_price` for legacy transactions is always the same as `x`.
Now, if transaction is sent with `x` estimated 1559 gas price, it can be mined with `y` 1559 gas price in case the transaction is mined in a block with different base_fee_per_gas then the block for which `x` estimated 1559 gas price is originally calculated. That said, `effective_gas_price` for 1559 transactions can be different from the `x`.

Todo: update library dependencies.

Related PRs:
https://github.com/gnosis/gp-gas-estimation/pull/14
https://github.com/gnosis/ethcontract-rs/pull/638
https://github.com/gnosis/gp-transaction-retry/pull/5
https://github.com/gnosis/gp-v2-services/issues/1056


nlordell (approved):
Looks good.

solver/src/settlement_simulation.rs
```diff
@@ -75,7 +77,7 @@ pub async fn simulate_settlements(
                 // set a gas limit so we don't get failed simulations because of insufficient
                 // solver balance for the default ~150M gas limit. Limit to around the
                 // block gas limit (since we can't fit more anyway).
-                .gas(15_000_000.into());
+                .gas(30_000_000.into());
```
sunce86, line 80: @nlordell 
Is this change ok? I assumed mainnet block size.

orderbook/src/fee.rs
```diff
@@ -168,7 +168,7 @@ impl MinFeeCalculator {
         amount: Option<U256>,
         kind: Option<OrderKind>,
     ) -> Result<U256, PriceEstimationError> {
-        let gas_price = self.gas_estimator.estimate().await?;
+        let gas_price = self.gas_estimator.estimate().await?.estimate();
```
bh2smith, line 171: Seems weird to need to `.estimate()` twice... I can only assume these are two different functions, but which is which.

shared/src/arguments.rs
```diff
@@ -49,6 +49,9 @@ pub struct Arguments {
     )]
     pub gas_estimators: Vec<GasEstimatorType>,
 
+    #[structopt(long, env)]
+    pub gas_estimator_api_key: Vec<String>,
```
bh2smith, line 53: Should this be optional? If it is not set, then we could fall back on whatever we were doing before without this api key? Or possibly have a default value that aligns with the default of `gas_estimators` above. 

Actually, now that I think of it, I would propose that `gas_estimators` be upgraded to a vector of tuples `Vec<(GasEstimatorType, ApiKey)>`. So that we don't have to assert their lengths are equal.

bh2smith (approved):
Looks like pretty good progress overall, but I would insist that the configuration parameters involving api key and gas estimator types be merged.

bh2smith (changes requested):
Changing accidental approval for changes requested in approval message.

orderbook/src/fee.rs
nlordell: Good point. I think the second estimate is to turn the estimated gas price into a legacy `gas_price` value. Maybe something like `to_legacy_gas_price()` might be more descriptive?

shared/src/arguments.rs
vkgnosis: Imo this should be` blocknative_api_key` . If `GasEstimatorType` includes `BlockNative` we will ensure in `main` that the api key is set. I prefer this over the tuple or vec because not every estimator needs an api key.

shared/src/gas_price_estimation.rs
```diff
@@ -44,12 +50,32 @@ pub async fn create_priority_estimator(
...
     let network_id = web3.net().version().await?;
     let mut estimators = Vec::<Box<dyn GasPriceEstimating>>::new();
-    for estimator_type in estimator_types {
+    for (estimator_type, estimator_api_key) in estimator_types.iter().zip(estimator_api_key.iter())
+    {
         match estimator_type {
+            GasEstimatorType::BlockNative => {
+                if !is_mainnet(&network_id) {
+                    return Err(anyhow!("BlockNative only supports mainnet"));
+                }
```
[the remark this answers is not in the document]
sunce86: Fixed

orderbook/src/fee.rs
sunce86: Changed to `effective_gas_price()`, since this term is broadly used and known in ethereum protocol and defi projects.

nlordell: `effective_gas_price` sounds good to me.

shared/src/arguments.rs
sunce86: Changed to `pub blocknative_api_key: Option<String>`

bh2smith (approved):

orderbook/src/fee.rs
```diff
@@ -168,7 +168,7 @@ impl MinFeeCalculator {
         amount: Option<U256>,
         kind: Option<OrderKind>,
     ) -> Result<U256, PriceEstimationError> {
-        let gas_price = self.gas_estimator.estimate().await?;
+        let gas_price = self.gas_estimator.estimate().await?.effective_gas_price();
```
nlordell, line 171: This is a gotcha! We want the effective gas price for fee computations and not gas caps.

nlordell (approved):

shared/Cargo.toml
```diff
@@ -9,15 +9,17 @@ license = "MIT OR Apache-2.0"
 anyhow = "1.0"
 assert_approx_eq = "1.1"
 async-trait = "0.1"
+async-recursion = "0.3"
```
vkgnosis, line 12: Unused dependency.

vkgnosis (approved):
LGTM.

shared/Cargo.toml
sunce86: Removed.

PR #1167, opened 2021-09-28
CowDexAgSolver solver

josojo:
When I looked into "better ways to merge" last Friday, I realised that actually, its better to do a two step approach:
1. first solve all orders just with paraswap
2. Identify cows and remove the cow-volume from orders
3. Solve aggregated-orders without cowvolume again using zeroEx.

I got hooked and coded up this CowDexAgSovler over the weekend.
It's already producing cool solutions that are settling: here [simulated only on tenderly](https://dashboard.tenderly.co/gp-v2/staging/simulator/48765ec3-feaf-4bca-8c1e-b0527fb2938f):
image

Unfortunately, it's taking too long to finish and clean up the code. Hence, I will keep this PR in draft mode and restart on eco-system friday and focus on affiliate program again.

### Test Plan
```
execute ignored tests
```
and listen to prod:
```
cargo run -p solver -- \                     
  --solvers CowDexAgSolver \
  --transaction-strategy DryRun
  ```
  
 PS: This solver has the weakness that if there is a ring of orders, it will just return an empty settlement. Martin has a even better plan for this case :)

### DIFF
--- a/solver/src/solver/cow_dex_ag_solver.rs
+++ b/solver/src/solver/cow_dex_ag_solver.rs
@@ -1,0 +1,412 @@
+use crate::solver::solver_utils::Slippage;
+use crate::solver::zeroex_solver::api::SwapQuery;
+use crate::solver::zeroex_solver::api::SwapResponse;
+use crate::solver::zeroex_solver::STANDARD_ZEROEX_SLIPPAGE_BPS;
+use crate::solver::{Auction, Solver};
+use crate::solver::{ParaswapSolver, ZeroExSolver};
+use crate::{liquidity::LimitOrder, settlement::Settlement};
+use anyhow::{anyhow, Result};
+use contracts::GPv2Settlement;
+use ethcontract::{Account, H160, U256};
+use futures::future::join_all;
+use reqwest::Client;
+use shared::{token_info::TokenInfoFetching, Web3};
+use std::collections::HashMap;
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by fleupold** [comment]
If I remember correct @koeppelmann was working on something similar.

File: solver/src/solver/cow_dex_ag_solver.rs [hunk 1]
```diff
@@ -0,0 +1,644 @@
...
+    // todo: Solver does not yet deal with unwrap and warps
+    async fn solve(&self, Auction { orders, .. }: Auction) -> Result<Vec<Settlement>> {
+        if orders.is_empty() {
+            return Ok(Vec::new());
+        }
+        // Randomize which orders we start with to prevent us getting stuck on bad orders.
+        // orders.shuffle(&mut rand::thread_rng())
+        let mut orders: Vec<LimitOrder> = orders.into_iter().collect();
+        // for now, lets only solve for up to 5 orders
+        orders.truncate(5);
```
nlordell, line 73: Why limit the amount of orders? Shouldn't we be limiting the amount of pairs instead?

nlordell:
Cool!

bh2smith (commented):
Very exciting. Will hold off on review until you feel it is in better shape.

josojo:
I will rewrite it as external http solver

PR #1168, opened 2021-09-28
Graceful shutdown of API

### LINKED ISSUE #576: Graceful shutdown of Orderbook API
We sometimes see errors like this in staging:

```
ERROR (pod: dev-dfusion-v2-solver-xdai-784db7785b-9s2hc):
2021-05-10T12:19:12.516Z ERROR solver::driver: single run errored: failed to get orderbook
Caused by:
   0: failed to get orderbook
   1: error sending request for url (http://dev-dfusion-v2-api-xdai.dfusion.svc.cluster.local/api/v1/solvable_orders): error trying to connect: tcp connect error: Connection refused (os error 111)
   2: error trying to connect: tcp connect error: Connection refused (os error 111)
   3: tcp connect error: Connection refused (os error 111)
   4: Connection refused (os error 111)
```

despite having set up rolling updates for our releases (cf. discussion with [DevOps](https://gnosisinc.slack.com/archives/C6Z2XNL5Q/p1620650640259400)).

The reason is likely that we don't have graceful shutdown handling in our orderbook API and therefore if the signal to terminate the pod is sent after the connection request has been routed but before it has been answered (~100ms window), we might ungracefully shutdown the inflight TCP connection.

I assume we can use [`tokio::signal`](https://docs.rs/tokio/1.5.0/tokio/signal/index.html) and [`warp::bind_with_graceful_shutdown`](https://docs.rs/warp/0.3.1/warp/struct.Server.html#method.bind_with_graceful_shutdown) to trigger a graceful shutdown when our pods get closed (in particular waiting for any in-flight request to finish).

Given we don't restart pods often in prod and this is not a serious issue (next connection attempt will work) this is not particularly high pri and maybe a "good first issue".

fleupold:
Fixes #576

We are from time to time seeing alerts for error messages like (mostly in staging):

> ERROR (pod: dev-dfusion-v2-solver-mainnet-6459f9677f-gx4zg):
2021-09-28T07:03:22.630Z ERROR solver::driver: single run errored: failed to get orderbook liquidity
Caused by:
   0: failed to get orderbook
   1: error sending request for url (http://...dfusion.svc.cluster.local/api/v1/solvable_orders): error trying to connect: tcp connect error: Connection refused (os error 111)
   2: error trying to connect: tcp connect error: Connection refused (os error 111)
   3: tcp connect error: Connection refused (os error 111)
   4: Connection refused (os error 111)

This is likely due to the API receiving a sigterm signal while a request is inflight (in which case the tcp connection may be refused). This PR makes it so that we intercept the sigterm signal and use it to gracefully shut down the server (which will finish serving in flight requests).

### Test Plan
Set a sleep of 10s in one the get endpoints. Query the path and before the 10s is up send a sigterm to the orderbook process. Before this PR the inflight request would error. Now, the shutdown will wait until the sleep is up and the request is served before tearing down the process.

### DIFF
--- a/docker/startup.sh
+++ b/docker/startup.sh
@@ -5,10 +5,10 @@
 # Run both commands in the background, so that we can receive signals and forward it to the main process (for this we capture the pid).
 # The stream-split command deliberately ignores interrupts (otherwise we may lose logs on teardown). It terminates automatically
 # once the main command (stdin) terminates.
-( "$@" & echo $! > main_pid ) | (trap '' TERM INT; regex-stream-split '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+(TRACE|DEBUG|INFO|WARN)' '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+ERROR') &
+( "$@" & echo $! > main_pid ) | regex-stream-split '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+(TRACE|DEBUG|INFO|WARN)' '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+ERROR' &
 
-pid=$(cat main_pid)
-trap 'kill -15 $pid' TERM
-trap 'kill -2 $pid' INT
+# Forward signal to main pid. Sync after killing to make sure the stdout is flushed
+trap 'kill -SIGTERM $(cat main_pid); sync' SIGTERM
+trap 'kill -SIGINT $(cat main_pid); sync' SIGINT
 
 wait
--- a/docker/startup.sh
+++ b/docker/startup.sh
@@ -2,8 +2,13 @@
 # The purpose of this script is to set a log regex to forward error logs to stderr and other logs to
 # stdout. This allows us to easily configure alerts when any message goes to stderr.
 
-# The main command is run in the background so that it receives interrupt signals
+# Run both commands in the background, so that we can receive signals and forward it to the main process (for this we capture the pid).
 # The stream-split command deliberately ignores interrupts (otherwise we may lose logs on teardown). It terminates automatically
 # once the main command (stdin) terminates.
-("$@" &) | (trap '' INT; regex-stream-split '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+(TRACE|DEBUG|INFO|WARN)' '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+ERROR')
+( "$@" & echo $! > main_pid ) | (trap '' TERM INT; regex-stream-split '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+(TRACE|DEBUG|INFO|WARN)' '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+ERROR') &
 
+pid=$(cat main_pid)
+trap 'echo foo; kill -15 $pid' TERM
+trap 'echo bar; kill -2 $pid' INT
+
+wait
--- a/docker/startup.sh
+++ b/docker/startup.sh
@@ -8,7 +8,7 @@
 ( "$@" & echo $! > main_pid ) | (trap '' TERM INT; regex-stream-split '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+(TRACE|DEBUG|INFO|WARN)' '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+ERROR') &
 
 pid=$(cat main_pid)
-trap 'echo foo; kill -15 $pid' TERM
-trap 'echo bar; kill -2 $pid' INT
+trap 'kill -15 $pid' TERM
+trap 'kill -2 $pid' INT
 
 wait
--- a/docker/startup.sh
+++ b/docker/startup.sh
@@ -1,4 +1,9 @@
 #!/bin/sh
 # The purpose of this script is to set a log regex to forward error logs to stderr and other logs to
 # stdout. This allows us to easily configure alerts when any message goes to stderr.
-"$@" | regex-stream-split "^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+(TRACE|DEBUG|INFO|WARN)" "^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+ERROR"
+
+# The main command is run in the background so that it receives interrupt signals
+# The stream-split command deliberately ignores interrupts (otherwise we may lose logs on teardown). It terminates automatically
+# once the main command (stdin) terminates.
+("$@" &) | (trap '' INT; regex-stream-split '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+(TRACE|DEBUG|INFO|WARN)' '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+ERROR')
+
--- a/e2e/tests/services.rs
+++ b/e2e/tests/services.rs
@@ -22,7 +22,7 @@ use shared::{
     Web3,
 };
 use solver::orderbook::OrderBookApi;
-use std::collections::HashMap;
+use std::{collections::HashMap, future::pending};
 use std::{num::NonZeroU64, str::FromStr, sync::Arc, time::Duration};
 
 pub const API_HOST: &str = "http://127.0.0.1:8080";
@@ -226,12 +226,13 @@ impl OrderbookServices {
         let maintenance = ServiceMaintenance {
             maintainers: vec![db.clone(), event_updater],
         };
-        orderbook::serve_task(
+        orderbook::serve_api(
             db.clone(),
             orderbook,
             fee_calculator,
             price_estimator.clone(),
             API_HOST[7..].parse().expect("Couldn't parse API address"),
+            pending::<()>(),
         );
 
         Self {
--- a/e2e/tests/services.rs
+++ b/e2e/tests/services.rs
@@ -232,7 +232,7 @@ impl OrderbookServices {
             fee_calculator,
             price_estimator.clone(),
             API_HOST[7..].parse().expect("Couldn't parse API address"),
-            pending::<()>(),
+            pending(),
         );
 
         Self {
--- a/orderbook/src/lib.rs
+++ b/orderbook/src/lib.rs
@@ -29,9 +29,7 @@ pub fn serve_api(
 ) -> JoinHandle<()> {
     let filter = api::handle_all_routes(database, orderbook, fee_calculator, price_estimator);
     tracing::info!(%address, "serving order book");
-    let (_, server) = warp::serve(filter).bind_with_graceful_shutdown(address, async {
-        shutdown_receiver.await;
-    });
+    let (_, server) = warp::serve(filter).bind_with_graceful_shutdown(address, shutdown_receiver);
     task::spawn(server)
 }
 
--- a/orderbook/src/lib.rs
+++ b/orderbook/src/lib.rs
@@ -19,13 +19,13 @@ use shared::price_estimation::PriceEstimating;
 use std::{net::SocketAddr, sync::Arc};
 use tokio::{task, task::JoinHandle};
 
-pub fn serve_api<T>(
+pub fn serve_api(
     database: Arc<dyn TradeRetrieving>,
     orderbook: Arc<Orderbook>,
     fee_calculator: Arc<EthAwareMinFeeCalculator>,
     price_estimator: Arc<dyn PriceEstimating>,
     address: SocketAddr,
-    shutdown_receiver: impl Future<Output = T> + Send + 'static,
+    shutdown_receiver: impl Future<Output = ()> + Send + 'static,
 ) -> JoinHandle<()> {
     let filter = api::handle_all_routes(database, orderbook, fee_calculator, price_estimator);
     tracing::info!(%address, "serving order book");
--- a/orderbook/src/lib.rs
+++ b/orderbook/src/lib.rs
@@ -13,30 +13,26 @@ use anyhow::{anyhow, Context as _, Result};
 use contracts::GPv2Settlement;
 use database::trades::TradeRetrieving;
 use fee::EthAwareMinFeeCalculator;
+use futures::Future;
 use model::DomainSeparator;
-use shared::{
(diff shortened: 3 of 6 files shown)


orderbook/src/lib.rs
```diff
@@ -13,30 +13,26 @@ use anyhow::{anyhow, Context as _, Result};
...
 use tokio::{task, task::JoinHandle};
 
-pub fn serve_task(
+pub fn serve_api<T>(
     database: Arc<dyn TradeRetrieving>,
     orderbook: Arc<Orderbook>,
     fee_calculator: Arc<EthAwareMinFeeCalculator>,
     price_estimator: Arc<dyn PriceEstimating>,
     address: SocketAddr,
+    shutdown_receiver: impl Future<Output = T> + Send + 'static,
```
vkgnosis, line 28: Should be hardcoded to Output=() imo because the T is never used.

orderbook/src/main.rs
```diff
@@ -415,21 +415,35 @@ async fn main() {
...
     tokio::select! {
-        result = serve_task => tracing::error!(?result, "serve task exited"),
+        result = &mut serve_api => tracing::error!(?result, "API task exited"),
         result = maintenance_task => tracing::error!(?result, "maintenance task exited"),
         result = db_metrics_task => tracing::error!(?result, "database metrics task exited"),
+        result = metrics_task => tracing::error!(?result, "metrics task exited"),
+        _ = tokio::signal::ctrl_c() => {
+            tracing::info!("Gracefully shutting down API");
+            shutdown_sender.send(()).expect("failed to send shutdown signal");
+            serve_api.await.expect("API failed during shutdown");
```
vkgnosis, line 445: Might be a good idea to have a timeout here in case for whatever reason a request takes a long time to complete (could be a problem with a client that is out of our control).

vkgnosis (approved):

solver/src/settlement/settlement_encoder.rs
```diff
@@ -548,7 +548,7 @@ pub mod tests {
     fn merge_fails_because_price_is_different() {
         let prices = hashmap! { token(1) => 1.into(), token(2) => 2.into() };
         let encoder0 = SettlementEncoder::new(prices);
-        let prices = hashmap! { token(1) => 1.into(), token(2) => 3.into() };
+        let prices = hashmap! { token(1) => 1.into(), token(2) => 4.into() };
```
bh2smith, line 551: What does this change have to do with the purpose of this PR?

bh2smith (approved):
Seems alright although I can't say I completely understand what's going on here.

solver/src/settlement/settlement_encoder.rs
fleupold: It's fixing a flaky unit test. Depending on the order we iterate over the hashset we may scale token 1 (scale factor 1, test works as expected) or scale token 2 (scale factor 3/2 which when multiplied with 1 results in 1, so the merge doesn't fail as expected).

orderbook/src/main.rs
fleupold: Kubernetes has a timeout built in to shutting services down 30s after they receive SIGTERM. But sure can add it here as well.

orderbook/src/lib.rs
fleupold: The issue is that oneshot::channel receiver returns a Result instead of () https://docs.rs/tokio/1.12.0/tokio/sync/oneshot/struct.Receiver.html#impl-Future. I guess I can somehow map that to ()

nlordell: You should be able to do the same trick you did below:
```rust
serve_api(...,  async { shutdown_receiver.await; });
```

orderbook/src/main.rs
nlordell: Agree we should timeout here so that using `Ctrl-C` on the command line when running locally can't hang for ever.

orderbook/src/lib.rs
```diff
@@ -13,30 +13,26 @@ use anyhow::{anyhow, Context as _, Result};
...
+    let filter = api::handle_all_routes(database, orderbook, fee_calculator, price_estimator);
     tracing::info!(%address, "serving order book");
-    task::spawn(warp::serve(filter).bind(address));
-
-    tracing::info!(%metrics_address, "serving metrics");
-    metrics_address.set_port(DEFAULT_METRICS_PORT);
-    serve_metrics(orderbook, metrics_address)
+    let (_, server) = warp::serve(filter).bind_with_graceful_shutdown(address, async {
+        shutdown_receiver.await;
+    });
```
nlordell, line 34: With @vkgnosis's suggestion of changing the parameter to `shutdown_receiver: impl Future<Output = ()> + Send + 'static,`, you would be able to:
```suggestion
    let (_, server) = warp::serve(filter).bind_with_graceful_shutdown(address, shutdown_receiver);
```

orderbook/src/main.rs
```diff
@@ -415,21 +415,35 @@ async fn main() {
...
 
-    let serve_task = serve_task(
+    let (shutdown_sender, shutdown_receiver) = tokio::sync::oneshot::channel();
+    let serve_api = serve_api(
         database.clone(),
         orderbook.clone(),
         fee_calculator,
         price_estimator,
         args.bind_address,
+        shutdown_receiver,
```
nlordell, line 425: In order to make `shutdown_receiver` a `impl Future<Output = ()>` you can:
```suggestion
        async { shutdown_receiver; },
```

nlordell (approved):
Looks good. How does this play with `tini` which AFAIU "wraps" our executable and has its own signal handling?

e2e/tests/services.rs
```diff
@@ -226,12 +226,13 @@ impl OrderbookServices {
...
             maintainers: vec![db.clone(), event_updater],
         };
-        orderbook::serve_task(
+        orderbook::serve_api(
             db.clone(),
             orderbook,
             fee_calculator,
             price_estimator.clone(),
             API_HOST[7..].parse().expect("Couldn't parse API address"),
+            pending::<()>(),
```
nlordell, line 235: I wonder if the explicit type parameter is still needed given that `serve_api` no longer generic over the output of the future type?
```suggestion
            pending(),
```

solver/src/settlement/settlement_encoder.rs
bh2smith: I see now why this was changed here! I hope we can land this PR soon because of this.

fleupold:
> Looks good. How does this play with tini which AFAIU "wraps" our executable and has its own signal handling?

I tested it locally and it seems like the culprit is actually that we pipe our binary into the `regex-stream-split` to split ERROR messages to stderr and the rest to stdout. 

Since this is happening in a sub-shell command it wasn't properly reacting to signals(https://unix.stackexchange.com/questions/146756/forward-sigterm-to-child-in-bash).

To test, I ran `docker run -ti -p 8080:8080 -e NODE_URL=https://staging-openethereum.mainnet.gnosisdev.com -e DB_URL="postgresql://host.docker.internal/?user="$USER"&password=" gp-v2-services orderbook` and hit ctrl-c (which wouldn't do anything).

I therefore now changed the startup script to launch the main process in the background (this way it seems to be getting interrupt signals).

@vkgnosis what was the reason again for doing the stream splitting as a wrapper shell script, rather than configuring the desired stream per log-level in our [tracing setup](https://github.com/gnosis/oba-services/blob/061c823a13bb2cffb17a9f0887a75e31f89907e8/shared/src/tracing.rs#L10) in the main binary?

solver/src/settlement/settlement_encoder.rs
fleupold: Feel free to copy the change if this is blocking another PR

vkgnosis:
This has been a while ago but I think the reason was that the tracing crate didn't have a nice way of choosing the output based on the log level and piping through a separate binary was a reusable fix. We could check if tracing can do this now or try to contribute it ourself.

fleupold:
After some more testing, I ended up adding support for SIGTERM as well (this is what docker/kubernetes send on teardown), which required some changes to `main.rs` as well as to the startup script (SIGTERM doesn't seem to be forwarded to child processes by default).

Now I can run `docker stop (docker ps -q --filter ancestor=gp-v2-services)` and will also see the graceful shutdown log message (as well as when hitting ctrl-c).

Could I get another sanity check please?

vkgnosis (approved):
LGTM. I don't feel confident judging the bash fu but I feel it's fine if you tested it.
Only thing I am worried about is that we are importing tokio::signal::unix which might break builds on windows. I don't know if anyone uses Windows (maybe @sunce86 ?) but it would be nice if the project still compiled there if it isn't too much extra code to just not do any signal handling if it isn't supported.

fleupold:
> Only thing I am worried about is that we are importing tokio::signal::unix which might break builds on windows.

My understanding of the crate was that it would build under windows, just not react to the signals as expected. I however don't have a Windows machine to test this on.

vkgnosis:
Based on https://docs.rs/tokio/1.12.0/tokio/signal/unix/index.html I don't think so. The only cross platform function is the ctrl + c handler on the signal module (one level higher).

docker/startup.sh
```diff
@@ -2,8 +2,13 @@
...
-# The main command is run in the background so that it receives interrupt signals
+# Run both commands in the background, so that we can receive signals and forward it to the main process (for this we capture the pid).
 # The stream-split command deliberately ignores interrupts (otherwise we may lose logs on teardown). It terminates automatically
 # once the main command (stdin) terminates.
-("$@" &) | (trap '' INT; regex-stream-split '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+(TRACE|DEBUG|INFO|WARN)' '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+ERROR')
+( "$@" & echo $! > main_pid ) | (trap '' TERM INT; regex-stream-split '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+(TRACE|DEBUG|INFO|WARN)' '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+ERROR') &
 
+pid=$(cat main_pid)
+trap 'echo foo; kill -15 $pid' TERM
+trap 'echo bar; kill -2 $pid' INT
```
bh2smith: Are foo and bar expected to be echoed when we terminate the process now? What ever happened to `kill -9`?

fleupold: This is outdated

fleupold:
> Based on https://docs.rs/tokio/1.12.0/tokio/signal/unix/index.html I don't think so.

You are right. We could extract the shutdown signal into its own module (which we then also import depending on the target system). However, with that I was running into a bunch of lifetime issues when trying to `pin_mut` the signal streams (someone needs to own them), so it might require a custom future implementation 😞

orderbook/src/main.rs
```diff
@@ -415,24 +415,70 @@ async fn main() {
...
+            .unwrap()
+            .recv()
+            .await;
+    };
+    futures::pin_mut!(sigint);
+    futures::pin_mut!(sigterm);
+    futures::future::select(sigterm, sigint).await;
+}
+
+#[cfg(windows)]
```
nlordell, line 476: Nice workaround.

docker/startup.sh
```diff
@@ -1,4 +1,14 @@
 #!/bin/sh
 # The purpose of this script is to set a log regex to forward error logs to stderr and other logs to
 # stdout. This allows us to easily configure alerts when any message goes to stderr.
-"$@" | regex-stream-split "^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+(TRACE|DEBUG|INFO|WARN)" "^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+ERROR"
+
+# Run both commands in the background, so that we can receive signals and forward it to the main process (for this we capture the pid).
+# The stream-split command deliberately ignores interrupts (otherwise we may lose logs on teardown). It terminates automatically
+# once the main command (stdin) terminates.
+( "$@" & echo $! > main_pid ) | (trap '' TERM INT; regex-stream-split '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+(TRACE|DEBUG|INFO|WARN)' '^\d+-\d+-\d+T\d+:\d+:\d+\.\d+Z\s+ERROR') &
```
nlordell, line 8: @fedgiac Do you mind taking a look as the bash :mage:?

fedgiac: I'm by no means an expert of bash traps but I'll try!

I think there is a race condition between `echo $! > main_pid` and `pid=$(cat main_pid)`.
This should be resolved for the most part by using `trap 'kill -15 $(cat main_pid)' TERM`. The traps can then be moved to the start, and the `&` at the end of the current first line and the `wait` can be removed.

After this I don't see anything else that could be troublesome but Bash surprised me many times. :see_no_evil:
As a nit I propose to use `kill -SIGINT`/`-SIGTERM` instead of `-2`/`-15`.

nlordell:
I ran the build and ran the new docker container and it seemed to work for me. Funny that when I ran the `startup.sh` locally it wouldn't output anything until I hit `Ctrl-C`. It might be an `ash` (Shell provided by BusyBox that is symlinked to `sh` on Alpine) vs `sh` thing.

nlordell (approved):

docker/startup.sh
fleupold: > I think there is a race condition between echo $! > main_pid and pid=$(cat main_pid).

Does `echo $! > main_pid` not block until the file is written? Will change to your suggestion anyways.

I don't think the traps can be moved though, since `sh` doesn't seem to be forwarding SIGTERM to the first command of the pipe while it's waiting on `regex-stream-split` to finish. If we spawn `regex-stream-split` as a background task it works however.

My test plan is to run `docker stop` against a running container and expect to see `Gracefully shutting down API`

fedgiac: > Does echo $! > main_pid not block until the file is written?

I don't know, but the issue is a bit different: to my understanding the syntax `command >file &` spawns a new thread, and then this new thread writes to `file`. At the same time, the thread that spawned the new thread tries to read the same file immediately afterwards.

As a minimal example of the issue I have this command that prints `before`, `after` or nothing randomly when trying multiple times.
If it doesn't work for you, you can try with other initial values rather than `i=300`.
```
sh -c 'echo before >/tmp/test; ( i=300; while (( i-- )); do :; done && echo after >/tmp/test ) & cat /tmp/test'
```

fedgiac: Anyway, how it is right now looks good to me!

PR #1169, opened 2021-09-29
Pre-Order Validation

bh2smith:
In Preparation for the new Fee and Quote endpoint (cf #1155) we shift some of the order validation into its own file (to be shared among both `Orderbook.add_order` and the `QuoteRequest` endpoint). More specifically, we move all order validations pertaining to the buy/sell tokens, order validity, balance sources and destinations, banned users, and transfer eth to contract into their own `PreOrderValidator`. Any order validations regarding balance or fee sufficiency and signature validity remain in the `Orderbook.add_order` method.

### Test Plan
This is a refactor although I have added some tests to the new method `validate_partial_order`

### DIFF
--- a/orderbook/src/api/validation.rs
+++ b/orderbook/src/api/validation.rs
@@ -1,0 +1,353 @@
+use contracts::WETH9;
+use ethcontract::H160;
+use model::order::{BuyTokenDestination, Order, SellTokenSource, BUY_ETH_ADDRESS};
+use shared::web3_traits::CodeFetching;
+use std::time::Duration;
+use warp::{http::StatusCode, reply::Json};
+
+#[derive(Debug)]
+pub enum ValidationError {
+    Forbidden,
+    InsufficientValidTo,
+    TransferEthToContract,
+    SameBuyAndSellToken,
+    UnsupportedBuyTokenDestination(BuyTokenDestination),
+    UnsupportedSellTokenSource(SellTokenSource),
+    Other(anyhow::Error),
+}
+
+impl ValidationError {
+    pub fn to_warp_reply(&self) -> (Json, StatusCode) {
+        match self {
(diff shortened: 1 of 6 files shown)


orderbook/src/api/validation.rs
```diff
@@ -0,0 +1,353 @@
...
+
+#[derive(Default)]
+pub struct PreOrderData {
+    owner: H160,
+    sell_token: H160,
+    buy_token: H160,
+    receiver: H160,
+    valid_to: u32,
+    buy_token_balance: BuyTokenDestination,
+    sell_token_balance: SellTokenSource,
```
vkgnosis, line 75: Probably want all of these `pub` too.

vkgnosis (approved):

orderbook/src/api/validation.rs
bh2smith: Yea good call, I made these all public in the next PR, but I can do it here in favour of less changes in the next one.

orderbook/src/api/validation.rs
```diff
@@ -0,0 +1,353 @@
...
+                sell_token: BUY_ETH_ADDRESS,
+                buy_token: native_token.address(),
+                ..Default::default()
+            },
+            &native_token,
+        ));
+    }
+
+    #[tokio::test]
+    async fn validate_partial_order_err() {
```
nlordell, line 205: One underrated feature of this refactor is the ability to test order validation better.

nlordell (approved):
Love it.

orderbook/src/api/validation.rs
bh2smith: Perhaps in a follow up PR, I can make a similar method for the other parts of order validation.

PR #1171, opened 2021-09-29
Remove orderbook dependency from solver

vkgnosis:
This is never used. Removing it should improve compile times a little as orderbook and solver crate now be compiled in paralle.

### Test Plan
CI


bh2smith (approved):
Awesome! From the commit history, it appears that orderbook was originally imported for gas estimation purposes. Glad to see such a nice change!

PR #1172, opened 2021-09-30
Fix panic with weighted token state scaling factor

vkgnosis:
U256::exp10 panics if result would overflow.

### Test Plan
I do have a test that constructs a full WeightedPool to trigger this but don't think it is useful to include since the code change is obvious enough.

### DIFF
--- a/shared/src/sources/balancer/swap.rs
+++ b/shared/src/sources/balancer/swap.rs
@@ -25,11 +25,15 @@ impl WeightedTokenState {
         self.upscale(self.token_state.balance)
     }
 
+    fn scaling_exponent_as_factor(&self) -> Option<U256> {
+        U256::from(10).checked_pow(self.token_state.scaling_exponent.into())
+    }
+
     /// Scales the input token amount to the value that is used by the Balancer
     /// contract to execute math operations.
     fn upscale(&self, amount: U256) -> Option<Bfp> {
         amount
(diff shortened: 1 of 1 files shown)


bh2smith (approved):
Very good find. Note that in our case this wouldn't have caused a panic since scaling_exponent is bound to be between 0 and 18, but it is always nice to avoid such things.

vkgnosis:
Good point. I wasn't sure if that a bound was enforced anyway. Followed to the code a bit but couldn't find it. One follow up could be to use a wrapper type to enforce the 18 maximum so we can unwrap. Or we could store the 10**scaling_factor value directly as U256 instead of as a u8 scaling_factor which also avoids the problem. But that only works if the raw scaling _factor isn't used.

nlordell (approved):

nlordell:
> I wasn't sure if that a bound was enforced anywhere. Followed to the code a bit but couldn't find it.

Its enforced by the SC: https://github.com/balancer-labs/balancer-v2-monorepo/blob/c18ff2686c61a8cbad72cdcfc65e9b11476fdbc3/pkg/pool-utils/contracts/BasePool.sol#L487

I do like checking here, the alternative would be to have a separate `pub type ScalingExponent(u8)` type that guarantees that its wrapped value belongs to `[0, 18]`, but that is a lot of overhead for very little benefit.

PR #1173, opened 2021-09-30
Add full_fee_amount column

### LINKED ISSUE #1162: Add order subsidy to databse
Seeing as order subsidy is now becoming dynamic, this has to be reflected in the objective value computation. The easiest way to do this is to add the fee subsidy to the order table in the database so it is part of the order metadata.

taminomara:
Fixes #1162

Step 1 of per-order fee computation. First, we save the original unsubsidized fee to the database, so we can use it later.

The database column is non-null, and set to be `0` for all existing orders. The idea here is to release a version that writes `full_fee_amount` but never reads it. After the release, we wait till all orders without `full_fee_amount` are either setteled or expired. Then we release a full version.

For now, the unsubsidized fee computation and validation logic is simple, we will change it later.

### Test Plan
Unit tests are in progress.

### DIFF
--- a/database/sql/V017__add_full_fee.sql
+++ b/database/sql/V017__add_full_fee.sql
@@ -1,0 +1,11 @@
+-- 1. Add full fee column.
+-- This column is a part of order metadata. Its values are used to settle orders;
+-- they are only relevant for non-expired orders. Therefore, we can set the default
+-- to `0`, deploy an intermediate version that stores this value but never actually
+-- reads it, then wait till all the old orders are expired.
+ALTER TABLE orders
+    ADD COLUMN full_fee_amount numeric(78,0) NOT NULL DEFAULT 0;
+
+-- 2. Drop defaults.
+ALTER TABLE orders
+    ALTER COLUMN full_fee_amount DROP DEFAULT;
--- a/orderbook/src/database/orders.rs
+++ b/orderbook/src/database/orders.rs
@@ -191,8 +191,8 @@ impl From<sqlx::Error> for InsertionError {
 // current amount of data this wouldn't be better.
 const ORDERS_SELECT: &str = "\
     o.uid, o.owner, o.creation_timestamp, o.sell_token, o.buy_token, o.sell_amount, o.buy_amount, \
-    o.valid_to, o.app_data, o.fee_amount, o.kind, o.partially_fillable, o.signature, o.receiver, \
-    o.signing_scheme, o.settlement_contract, o.sell_token_balance, o.buy_token_balance, \
+    o.valid_to, o.app_data, o.fee_amount, o.full_fee_amount, o.kind, o.partially_fillable, o.signature, \
+    o.receiver, o.signing_scheme, o.settlement_contract, o.sell_token_balance, o.buy_token_balance, \
     (SELECT COALESCE(SUM(t.buy_amount), 0) FROM trades t WHERE t.order_uid = o.uid) AS sum_buy, \
     (SELECT COALESCE(SUM(t.sell_amount), 0) FROM trades t WHERE t.order_uid = o.uid) AS sum_sell, \
     (SELECT COALESCE(SUM(t.fee_amount), 0) FROM trades t WHERE t.order_uid = o.uid) AS sum_fee, \
@@ -219,8 +219,8 @@ impl OrderStoring for Postgres {
             INSERT INTO orders (
                 uid, owner, creation_timestamp, sell_token, buy_token, receiver, sell_amount, buy_amount, \
                 valid_to, app_data, fee_amount, kind, partially_fillable, signature, signing_scheme, \
-                settlement_contract, sell_token_balance, buy_token_balance) \
-            VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18);";
+                settlement_contract, sell_token_balance, buy_token_balance, full_fee_amount) \
+            VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18, $19);";
         let receiver = order
             .order_creation
             .receiver
@@ -250,6 +250,7 @@ impl OrderStoring for Postgres {
             .bind(DbBuyTokenDestination::from(
                 order.order_creation.buy_token_balance,
             ))
+            .bind(u256_to_big_decimal(&order.order_meta_data.full_fee_amount))
             .execute(&self.pool)
             .await
             .map(|_| ())
@@ -406,6 +407,7 @@ struct OrdersQueryRow {
     valid_to: i64,
     app_data: Vec<u8>,
     fee_amount: BigDecimal,
+    full_fee_amount: BigDecimal,
     kind: DbOrderKind,
     partially_fillable: bool,
     signature: Vec<u8>,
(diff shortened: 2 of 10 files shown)


orderbook/src/database/orders.rs
```diff
@@ -490,7 +494,7 @@ impl OrdersQueryRow {
                     .map_err(|_| anyhow!("app_data is not [u8; 32]"))?,
             ),
             fee_amount: big_decimal_to_u256(&self.fee_amount)
-                .ok_or_else(|| anyhow!("buy_amount is not U256"))?,
+                .ok_or_else(|| anyhow!("fee_amount is not U256"))?,
```
nlordell, line 497: nice catch.

orderbook/src/fee.rs
```diff
@@ -263,8 +266,7 @@ impl MinFeeCalculating for MinFeeCalculator {
         Ok((min_fee, official_valid_until))
     }
 
-    // Returns true if the fee satisfies a previous not yet expired estimate, or the fee is high enough given the current estimate.
-    async fn is_valid_fee(&self, sell_token: H160, fee: U256, app_data: AppId) -> bool {
+    async fn validate_fee(&self, sell_token: H160, fee: U256, app_data: AppId) -> Result<U256, ()> {
```
nlordell, line 269: Interesting use of `Result` for invalid fees. I like it!

One idea would be to define a `struct InsufficientFeeError(U256)` so that:
- The caller knows the minimum in case its too small.
- Allows `impl From<InsufficientFeeError> for MyHighLevelError` conversion trait implementations for better `?` support. Otherwise, we would need `.map_err(|_| MyHighLevelError::InsufficienFeeError)?` in the calling context.

Not sure if my ideas are a bit overkill, just wanted to put them out there. Either way, I like the use of `Result` here over the bool.

orderbook/src/fee.rs
```diff
@@ -276,13 +278,17 @@ impl MinFeeCalculating for MinFeeCalculator {
             .await
         {
             if fee >= apply_fee_factor(past_fee, app_based_fee_factor) {
```
nlordell, line 280: I think `past_fee` already has the "base" fee factor applied. Maybe we should refactor the fee measurement to already capture completely un-subsidized fee.

Then, instead of
```rust
        let app_based_fee_factor = *self
            .partner_additional_fee_factors
            .get(&app_data)
            .unwrap_or(&1.0);
```

we could:
```rust
        let fee_factor = *self
            .partner_additional_fee_factors
            .get(&app_data)
            .unwrap_or(&1.0) * self.fee_factor;
```

This would make the `full_fee_amount` be the actual full fee amount, make changes to the "base" fee factor not affect orders (as the full amount will be stored, instead of full amount times "base" fee factor) and allow us to remove the `--fee-factor` configuration from the `solver` binary as well.

nlordell (commented):
Looks good already.

I would also refactor the min fee getting and storing to save the completely un-subsidized fee amount (so without the "base" `fee_factor`).

orderbook/src/fee.rs
taminomara: Nice catch! Yeah, sure I'll implement this.

database/sql/V017__add_full_fee.sql
```diff
@@ -0,0 +1,11 @@
+-- 1. Add full fee column.
+-- This column is a part of order metadata. Its values are used to settle orders;
+-- they are only relevant for non-expired orders. Therefore, we can set the default
+-- to `0`, deploy an intermediate version that stores this value but never actually
+-- reads it, then wait till all the old orders are expired.
+ALTER TABLE orders
+    ADD COLUMN full_fee_amount numeric(78,0) NOT NULL DEFAULT 0;
```
nlordell, line 7: For Postrges noobs like me:
https://www.postgresql.org/docs/9.1/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL

`numeric(78,0)` means a maximum of 78 (**decimal**) digits, where 0 of them are fractional. `U256::MAX.to_string().len() == 78` so it all checks out.

orderbook/src/fee.rs
```diff
@@ -260,11 +263,11 @@ impl MinFeeCalculating for MinFeeCalculator {
                 min_fee,
             )
             .await;
-        Ok((min_fee, official_valid_until))
+
+        Ok((apply_fee_factor(min_fee, self.fee_factor), official_valid_until))
```
nlordell, line 267: I think this still causes issues though right? Because now in `validate_fee`, depending on if the fee is fresh (i.e. it uses `compute_min_fee`) or not (uses `get_min_fee`) the base `fee_factor` will be applied or not.

Additionally, in the case it uses `compute_min_fee` result to validate the fee, the un-subsidized fee that is returned still has the "base" fee factor applied.

orderbook/src/fee.rs
```diff
@@ -177,6 +193,7 @@ impl MinFeeCalculator {
         }
     }
 
+    /// Computes unsubsidized min fee.
     async fn compute_min_fee(
```
nlordell, line 197: ubernit: Maybe rename the methods to avoid confusion:
- `compute_unsubsidized_min_fee`
- `get_unsubsidized_min_fee`

orderbook/src/fee.rs
```diff
@@ -213,6 +229,17 @@ impl MinFeeCalculator {
...
     }
+
+    fn calculate_fee_factor(&self, app_data: Option<AppId>) -> f64 {
+        app_data
+            .and_then(|app_data| self.partner_additional_fee_factors.get(&app_data).cloned())
+            .unwrap_or(1.0)
+            * self.fee_factor
+    }
+
+    fn apply_fee_factor(fee: U256, factor: f64) -> U256 {
```
nlordell, line 240: Just curious, why do you prefer this as a associated function vs a "free" function in the module?

nlordell (commented):
Looks good to me already! What is missing to for the PR to leave draft?

taminomara:
> Looks good to me already! What is missing to for the PR to leave draft?

I'll fix the test, then it'll be ready to merge.

orderbook/src/fee.rs
taminomara: Generally, my preference for associated functions is a habit from C++. In this case though, I thought it'll make more sense to keep all functions related to the same logic in one place.

nlordell: > I thought it'll make more sense to keep all functions related to the same logic in one place.

Interesting. In my mind, I tend to like to use modules to split up logic into groups instead of types (i.e. associated functions). Not suggesting one is better than the other, just interested in your PoV :smile:.

bh2smith (approved):
Looks great! What was the issue you were having before with the local DB stuff? had you forgotten to apply the migrations or something?

taminomara:
> What was the issue you were having before with the local DB stuff?

Wrong migration file name (needed capital `V` in `V017.sql`), plus some other things that I've missed.

database/sql/V017__add_full_fee.sql
nlordell: I'd slightly prefer 0 so we don't lose information on which orders we don't have this for, and so they don't get confused with orders that were created without any subsidy.

I have no strong opinion though.

nlordell:
I tend to agree and also slightly prefer 0's (see original comment https://github.com/gnosis/gp-v2-services/pull/1173#discussion_r721218685)

taminomara:
@nlordell do we merge it?

taminomara:
Oh no wait, I haven't renamed the functions.

taminomara:
> Oh no wait, I haven't renamed the functions.

Done

vkgnosis (approved):

nlordell:
Looks like CI is failing on `cargo fmt`.

nlordell (approved):
Shoot! Meant to approve but I couldn't because it was still a draft PR.

PR #1159, opened 2021-09-24
Fix handling of buy order edge case in baseline solver

vkgnosis:
To find the optimal path for buy orders we start at the buy token and
use get_amount_in repeatedly until we arrive at the sell token. However
when creating a settlement the path has to be traversed in the sell
direction because that is what the user has balance in.
This is problematic when an amm is not symmetrical: using the result of
get_amount_in in get_amount_out returns a (usually slightly) different
value. This can cause further amm interactions to fail.
This commit fixes this by enforcing the forward (sell) direction of the
resulting paths for buy orders. Sell orders do not need to be fixed as
they are already in the correct direction.
Additionally I have moved order limit checks into settle_order because
it felt more natural there with the forward path check.

Closes #1140

### Test Plan
Existing tests. I would have like to add a test for the panic case too but was unable to do so because the way the types are specified in the baseline solver I cannot easily pass in my own `impl BaselineSolvable` that has the behavior I want. So I would have to craft a pool that showcases this behavior by hand which was too time consuming.

PR #1175, opened 2021-09-30
Store weighted pool fee as Bfp instead of BigRational

vkgnosis:
WeightedProductOrder::fee was stored as BigRational. Now it is stored as
Bfp. This an improvement because underlying value is always a Bfp but
BigRational can store a superset of numbers so by switching to Bfp we
prevent illegal values from making their way into the value.
For example in amm_to_weighted_pool we were converting back to Bfp which
had to use try_into because it could fail. Now no failure is possible.

### Test Plan
CI

### DIFF
--- a/shared/src/baseline_solver.rs
+++ b/shared/src/baseline_solver.rs
@@ -10,6 +10,11 @@ const DEFAULT_MAX_HOPS: usize = 2;
 
 type PathCandidate = Vec<H160>;
 
+/// Note that get_amount_out and get_amount_in are not always symmetrical. That is for some AMMs it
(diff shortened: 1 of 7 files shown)


nlordell (approved):

bh2smith (approved):
Sweet! Nice find. Could you tell me how you discovered this?

vkgnosis:
I spent a long time finding a good example for https://github.com/gnosis/gp-v2-services/pull/1159 . During that I set up fuzzing and ended up following the weighted pool code around a lot. Eventually I saw the try_into when converting the BigRational to Bfp and felt it was weird that it was needed.

bh2smith:
Why no merge?

PR #1176, opened 2021-09-30
Independent logic for Partial and Full OrderValidation

bh2smith:
Similar to the Pre-Order Validation introduced in #1169, we offload some of the order validation statements contained in `Orderbook.add_order` to a specialized struct `PostOrderValidator`. This makes it possible to test and maintain Order Validation without making changes to the orderbook logic itself.

### Test Plan
Added new unit tests for all possible routes in `PostOrderValidator.validate()` along with tests for helper method `min_balance`


model/src/order.rs
```diff
@@ -69,7 +69,6 @@ impl Order {
         let owner = order_creation
             .signature
             .validate(domain, &order_creation.hash_struct())?;
-        // TODO - test this function when validate returns None.
```
bh2smith, line 72: This was never necessary because of the ? above.

orderbook/src/orderbook.rs
```diff
@@ -267,6 +212,10 @@ impl Orderbook {
         set_available_balances(orders.as_mut_slice(), &self.solvable_orders);
         Ok(orders)
     }
+
+    fn bad_token_detector(&self) -> &dyn BadTokenDetecting {
+        self.post_order_validator.bad_token_detector.as_ref()
+    }
```
bh2smith, line 218: This is why `bad_token_detector` was made into a public field inside the `PostOrderValidator`. I suspect this is kinda hacky and should fall back on just providing a  bad token detector directly to the orderbook, since it is clearly needed for more than just order validation. Its an Arc anyway, so It shouldn't be so costly to just give Orderbook its own bad token detector.

Would like to hear back from reviewers if my suspicion is on point.

vkgnosis: I prefer it if Orderbook stores its own Arc<BadTokenDetecting> and the field is private.

orderbook/src/api/validation.rs
```diff
@@ -104,41 +156,117 @@ impl PreOrderValidator {
...
             if code_size != 0 {
-                return Err(ValidationError::TransferEthToContract);
+                return Err(PreValidationError::TransferEthToContract);
+            }
+        }
+        Ok(())
+    }
+}
+
+pub struct PostOrderValidator {
```
vkgnosis, line 198: I feel it is unclear what the distinction between Pre and Post validation is in this file. Would be nice to have a doc comment either on the struct or on the validate function of both components explaining what they check for.

bh2smith: Thanks for the great suggestion. Also gave me the idea to combine the two validator structs into one. Also, now it might even make sense to call the methods "validate_partial" and just "validate" since now the validate also calls validate partial.

orderbook/src/api/order_validation.rs
```diff
@@ -0,0 +1,734 @@
...
+            native_token,
+            banned_users,
+            min_order_validity_period,
+            fee_validator,
+            bad_token_detector,
+            balance_fetcher,
+        }
+    }
+
+    pub async fn pre_validate(&self, order: PreOrderData) -> Result<(), PreValidationError> {
```
vkgnosis, line 207: Still not sure about the distinction here. Also this is public but the function is never used outside of this module. Will it be used later?

bh2smith: This method will only be called explicitly by the fee and quote which is inside this api module. I guess it should be private then? Although it's not being called directly from within this file.

bh2smith: More specifically, if this PR lands before #1155, then I would be calling it from there. I can make it private just for now and then open it up as public if/when necessary.

vkgnosis (approved):

orderbook/src/api/order_validation.rs
```diff
@@ -0,0 +1,734 @@
...
+///     This is the additional validation performed at the time of order placement
+///     (i.e. once all the required fields on an Order are provided). Specifically, verifying that
+///     - buy & sell amounts are non-zero,
+///     - order's owner matches the from field (if specified),
+///     - fee is sufficient,
+///     - buy & sell tokens passed "bad token" detection,
+///     - user has sufficient (transferable) funds to execute the order.
+///
+///     Furthermore, Post-order validation also calls pre-validate to ensure that
+///     other aspects of the order are not malformed.
```
nlordell, line 148: I worry that this description might quickly become out of date. Maybe a higher-level description on each of the methods instead of the type might make it closer to the actual code and less likely to become outdated over time.

orderbook/src/api/order_validation.rs
```diff
@@ -0,0 +1,734 @@
...
+            native_token,
+            banned_users,
+            min_order_validity_period,
+            fee_validator,
+            bad_token_detector,
+            balance_fetcher,
+        }
+    }
+
+    async fn pre_validate(&self, order: PreOrderData) -> Result<(), PreValidationError> {
```
nlordell, line 207: I would move documentation from the type here for pre-validation.

orderbook/src/api/order_validation.rs
```diff
@@ -0,0 +1,734 @@
...
+    }
+
+    pub async fn post_validate(
+        &self,
+        order: Order,
+        sender: Option<H160>,
+    ) -> Result<(), ValidationError> {
+        self.pre_validate(PreOrderData::from(order.clone()))
+            .await
+            .map_err(ValidationError::Pre)?;
```
nlordell, line 252: I find it weird that `post_validate` calls `pre_validate`. From the names I would expect you to have to call both:
```
validator.pre_validate()?;
// do stuff...
validator.post_validate()?;
```

Maybe `pre-`/`post-` prefixes can be changed (`partial-`/`full-`? Maybe something else)?

orderbook/src/api/order_validation.rs
```diff
@@ -0,0 +1,734 @@
+use crate::{account_balances::BalanceFetching, fee::MinFeeCalculating};
```
nlordell, line 1: I like that it's its own component now :smile:

orderbook/src/api/order_validation.rs
```diff
@@ -0,0 +1,734 @@
...
+use contracts::WETH9;
+use ethcontract::{H160, U256};
+use model::order::{BuyTokenDestination, Order, SellTokenSource, BUY_ETH_ADDRESS};
+use shared::{bad_token::BadTokenDetecting, web3_traits::CodeFetching};
+use std::{sync::Arc, time::Duration};
+use warp::{http::StatusCode, reply::Json};
+
+pub trait WarpReplyConverting {
+    fn to_warp_reply(&self) -> (Json, StatusCode);
+}
```
nlordell, line 11: While I think the trait here is cool, I don't see any generic code with something like `T: WarpReplayConverting` and feels a bit of overkill.

nlordell (commented):
Looks mostly good. Just some comments.

Its too bad GitHub UI doesn't catch that its a file rename, making it a bit harder to review.

orderbook/src/api/order_validation.rs
bh2smith: Yes, initially I had a call to each independently, but the only place this is done would be in `Orderbook.add_order`. I was worried that future refactors might forget that then both need to be called, so I enforced this by calling the pre inside the post. I completely agree that the name could be changed to `partial` and `full` or maybe even `partial_validate` and just `validate`.

orderbook/src/api/order_validation.rs
bh2smith: I think this will become more useful once #1176 comes into play (so I can implement this trait for a struct that is not declared inside this crate). Would that be a good reason for it?

orderbook/src/api/order_validation.rs
bh2smith: Fun fact, I had already mentioned this here https://github.com/gnosis/gp-v2-services/pull/1176#discussion_r720812626

orderbook/src/api/order_validation.rs
bh2smith: Me too!

orderbook/src/api/order_validation.rs
nlordell: Personally, I would still not use a trait in that situation (for mostly stylistic reasons), but there isn't anything wrong with doing so feel free to leave it in if you like it. I don't see a strong reason for using or not using a trait either way.

I think a more concrete reason to introduce such a trait would be the ability to get rid of all those `*_response` (for example `create_order_response`) methods and introduce a generic one:
```
pub fn as_response<T, E>(result: Result<T, E>) -> impl Reply
where
    T: serde::Serialize,
    E: WarpReplyConverting,
{
    // ...
}
```

nlordell:
> I could rename it back for now and just do the rename later if you prefer.

Its fine. It was mostly me complaining at the Github UI, which does it sometimes even when Git identifies changes as a file rename.

orderbook/src/api/order_validation.rs
nlordell: > `maybe even partial_validate and just validate.`

Works for me. @vkgnosis, from your comments, I think you also didn't find the names super clear, do you have any suggestions as well?

orderbook/src/api/order_validation.rs
nlordell: (Not suggesting this be done in this PR).

bh2smith: Cool I will defs do this (after).

bh2smith: For now I will move the trait to `api.rs` (i.e. before merging this).

bh2smith:
> > I could rename it back for now and just do the rename later if you prefer.

> Its fine. It was mostly me complaining at the Github UI, which does it sometimes even when Git identifies changes as a file rename.

Too late... already renamed back... Will change again later.

nlordell (approved):
It actually made it much easier to review! Tahnks.

orderbook/src/api/validation.rs
```diff
@@ -57,11 +62,79 @@ impl ValidationError {
     }
 }
 
-pub struct PreOrderValidator {
+#[derive(Debug)]
+pub enum PostValidationError {
```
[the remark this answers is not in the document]
nlordell: Maybe `PartialValidationError` and then you could have:
 ```
type ValidationError {
    Partial(PartialValidationError),
    InsufficientFee,
    InsufficientFunds,
    UnsupportedToken(H160),
    WrongOwner(H160),
    ZeroAmount,
    Other(anyhow::Error),
}
 ```
 
 And get rid of the `PostValidationError` altogether.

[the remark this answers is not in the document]
bh2smith: Excellent.

PR #1182, opened 2021-10-01
[Trivial] log gas price in gwei

bh2smith:
Major nit, bit I didn't like reading logs like

```
// submitting solution transaction at gas price 69123456789.123456789
```

so I made it into 
```
// submitting solution transaction at gas price 69.12 GWei
```

### DIFF
--- a/solver/src/settlement_submission/retry.rs
+++ b/solver/src/settlement_submission/retry.rs
@@ -68,7 +68,7 @@ impl<'a> TransactionSending for SettlementSender<'a> {
     async fn send(&self, gas_price: f64) -> Self::Output {
         tracing::info!(
             "submitting solution transaction at gas price {:.2} GWei",
-            gas_price / 1_000_000_000.
+            gas_price / 1e9
         );
         let mut method =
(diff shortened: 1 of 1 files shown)


solver/src/settlement_submission/retry.rs
```diff
@@ -66,7 +66,10 @@ impl<'a> TransactionSending for SettlementSender<'a> {
     type Output = SettleResult;
 
     async fn send(&self, gas_price: f64) -> Self::Output {
-        tracing::info!("submitting solution transaction at gas price {}", gas_price);
+        tracing::info!(
+            "submitting solution transaction at gas price {:.2} GWei",
+            gas_price / 1_000_000_000.
```
nlordell, line 71: ubernit:
To avoid counting 0s, I suggest:
```suggestion
            gas_price / 1e9
```

nlordell (approved):

solver/src/settlement_submission/retry.rs
bh2smith: Great. I just hope 1e9 also counts as a float.

bh2smith: Checked. It does.

PR #1184, opened 2021-10-02
Update Paraswap Solver Errors

fleupold:
We are seeing a lot of uncatalogued Paraswap errors (likely coming from the upgrade to v5). In this PR I went through all error alerts of last week and catalogued them.

I decided to make ServerBusy retryable (not sure if that's a good idea given that they are already encountering a lot of load in that case, but likely our single request won't make things worse).

Also, we should retry Send errors (as they may stem from timeouts or other intermittent network issues)

### Test Plan
Existing unit tests


bh2smith (approved):
Awesome! Thanks for doing this. Cataloging errors is cool.

nlordell (approved):

vkgnosis (approved):

PR #1187, opened 2021-10-04
[OnCall] Don't alert well catalogued (and somewhat expected) errors

fleupold:
Purpose of this PR is to control some of the most noisy alerts we are seeing (e.g. "Server too busy"). Those are well understood and catalogued yet they may still happen. While we may still want to alert if they happen more often than usual, it's currently not helpful to have these alerts in our channel.

### Test Plan
Existing CI

### DIFF
--- a/solver/src/solver/single_order_solver.rs
+++ b/solver/src/solver/single_order_solver.rs
@@ -86,6 +86,8 @@ impl<I: SingleOrderSolving + Send + Sync + 'static> Solver for SingleOrderSolver
 pub struct SettlementError {
     pub inner: anyhow::Error,
     pub retryable: bool,
+    // Whether or not this error should be logged as an error
+    pub should_alert: bool,
 }
 
 impl From<anyhow::Error> for SettlementError {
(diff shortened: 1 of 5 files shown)


solver/src/solver/single_order_solver.rs
```diff
@@ -86,13 +86,16 @@ impl<I: SingleOrderSolving + Send + Sync + 'static> Solver for SingleOrderSolver
 pub struct SettlementError {
     pub inner: anyhow::Error,
     pub retryable: bool,
+    // Whether or not this error should be logged as an error
+    pub should_alert: bool,
```
vkgnosis, line 90: I see this new filed but it is never read is it? So we set it to false but we don't actually prevent the alert at the moment.

fleupold: Good point, forgot to push the latest commit

vkgnosis (approved):

bh2smith (approved):
😄 great!

PR #1189, opened 2021-10-04
API - Remove defaults in order data fields

bh2smith:
Based on the discussion [here](https://github.com/gnosis/gp-v2-services/pull/1155#discussion_r720090959)

We remove all defaults for order request data and make the fields required by the API. This means that front ends, if not already, will have to start specifying sell and buy token source and destination.

cc @alfetopito, @W3stside  & @anxolin  (in case this affects cowswap interface).

### DIFF
--- a/orderbook/openapi.yml
+++ b/orderbook/openapi.yml
@@ -569,10 +569,8 @@ components:
           type: boolean
         sellTokenBalance:
           $ref: "#/components/schemas/SellTokenSource"
-          default: "erc20"
         buyTokenBalance:
           $ref: "#/components/schemas/BuyTokenDestination"
-          default: "erc20"
       required:
         - sellToken
         - buyToken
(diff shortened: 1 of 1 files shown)


fleupold:
This is quite a risky change. We do have a bunch of external partners (yearn, market makers) which may rely on these default values and whose integrations may break with this change. While I also prefer to avoid defaults for the API, I'm not sure this justifies a breaking change (if so we probably need to follow some process of letting people know).

bh2smith:
@fleupold - I totally understand and agree with your concern here. From my perspective, having these two fields with default values of erc20 isn't so bad anyway. I mean it is a pretty standard default to have for anyone not involved with Balancer.

vkgnosis:
I didn't know people already depended on it. Like you said I prefer in general no defaults but if this is already a problem then keeping them is fine.

PR #1192, opened 2021-10-04
Balancer: Rework AmplificationParameter

bh2smith:
Due to recent changes introduced in stable_math.rs #1127 it has been discovered that representing Stable Pool amplification parameter as BigRational will not be appropriate to accommodate both HTTP and Baseline solvers. The reason for this is that HTTP solver uses decimal representations of values (amplification factor divided by precision) while the `StableMath.sol` contract uses the product of these values (factor * precision). Because the Ratio (as we had before) gets reduced upon instantiation, it will not suffice to simply use the `numerator * denominator` (since reducing fractions results a loss of information).

For this reason, we introduce a new struct `AmplificationParameter { factor, precision }` along with two methods `as_big_rational` and `as_u256` which can be used when needed for each appropriate solver. 

It was noticed that amplification parameter was also being set at pool initialization time - which was no longer necessary once we realized it needed to be fetched and updated along with the other block-sensitive/dynamic pool info. This redundant duplication was removed here, but can also be split into an independent PR.

### Test Plan
TODO - Need to test to u256 and big_rational methods (an account for possibility of `as_big_rational` panic.


shared/src/sources/balancer/pool_fetching.rs
```diff
@@ -105,19 +106,42 @@ impl BalancerPoolEvaluating for WeightedPool {
     }
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct AmplificationParameter {
```
nlordell, line 110: Awesome.

shared/src/sources/balancer/pool_fetching.rs
```diff
@@ -105,19 +106,42 @@ impl BalancerPoolEvaluating for WeightedPool {
...
+}
+
+impl AmplificationParameter {
+    pub fn new(factor: U256, precision: U256) -> Self {
+        Self { factor, precision }
+    }
+
+    /// This is the format used to pass into smart contracts.
+    pub fn as_u256(&self) -> Option<U256> {
+        self.factor.checked_mul(self.precision)
```
nlordell, line 122: I think the format is `self.factor` not `self.factor * self.precision`.

nlordell (commented):
Thanks for cleaning this up. Some comments below.

shared/src/sources/balancer/pool_fetching.rs
bh2smith: That could very well be! Thanks.

shared/src/sources/balancer/pool_fetching.rs
```diff
@@ -105,19 +106,42 @@ impl BalancerPoolEvaluating for WeightedPool {
...
+    }
+
+    /// This is the format used to pass along to HTTP solver.
+    pub fn as_big_rational(&self) -> BigRational {
+        BigRational::new(self.factor.to_big_int(), self.precision.to_big_int())
```
bh2smith, line 127: @nlordell do you have any suggestions on how best to handle the completely unexpected scenario where the precision is 0?

shared/src/sources/balancer/pool_fetching.rs
```diff
@@ -105,19 +108,48 @@ impl BalancerPoolEvaluating for WeightedPool {
...
+
+    /// This is the format used to pass along to HTTP solver.
+    pub fn as_big_rational(&self) -> Result<BigRational> {
+        // In order to circumvent the possibility of panic from BigRational::new,
+        // we check for zero denominator and return result
+        ensure!(!self.precision.is_zero(), "Zero precision not allowed");
+        Ok(BigRational::new(
+            self.factor.to_big_int(),
+            self.precision.to_big_int(),
+        ))
```
bh2smith, line 135: I find it a bit silly/weird that we circumvent the panic, but then just use `.expect` on the result later...

shared/src/sources/balancer/pool_fetching.rs
```diff
@@ -105,19 +108,48 @@ impl BalancerPoolEvaluating for WeightedPool {
...
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct AmplificationParameter {
+    factor: U256,
+    precision: U256,
+}
+
+impl AmplificationParameter {
+    pub fn new(factor: U256, precision: U256) -> Self {
```
nlordell, line 118:
```suggestion
    pub fn new(factor: U256, precision: U256) -> Result<Self> {
        ensure!(!self.precision.is_zero(), "Zero precision not allowed");
```

shared/src/sources/balancer/pool_fetching.rs
```diff
@@ -105,19 +108,48 @@ impl BalancerPoolEvaluating for WeightedPool {
...
+    /// This is the format used to pass into smart contracts.
+    pub fn as_u256(&self) -> U256 {
+        self.factor
+    }
+
+    /// This is the format used to pass along to HTTP solver.
+    pub fn as_big_rational(&self) -> Result<BigRational> {
+        // In order to circumvent the possibility of panic from BigRational::new,
+        // we check for zero denominator and return result
+        ensure!(!self.precision.is_zero(), "Zero precision not allowed");
```
nlordell, line 131:
```suggestion
        // We can assert that the precision is non-zero as we check when constructing
        // new `AmplificationParameter` instances that this invariant holds, and we don't
        // allow modifications of `self.precision` such that it could become 0.
        debug_assert!(!self.precision.is_zero());
```

shared/src/sources/balancer/pool_fetching.rs
nlordell: Included some suggestions :point_up:

shared/src/sources/balancer/pool_fetching.rs
```diff
@@ -145,7 +177,10 @@ impl StablePool {
                 paused,
             },
             reserves,
-            amplification_parameter,
+            amplification_parameter: AmplificationParameter {
+                factor: amplification_factor,
+                precision: amplification_precision,
+            },
```
nlordell, line 183:
```suggestion
            amplification_parameter: AmplificationParameter::new(
                amplification_factor,
                amplification_precision,
            )?,
```

shared/src/sources/balancer/pool_fetching.rs
nlordell, line 131:
```suggestion
    pub fn as_big_rational(&self) -> BigRational {
        // We can assert that the precision is non-zero as we check when constructing
        // new `AmplificationParameter` instances that this invariant holds, and we don't
        // allow modifications of `self.precision` such that it could become 0.
        debug_assert!(!self.precision.is_zero());
```

vkgnosis (approved):

nlordell (approved):

PR #1194, opened 2021-10-05
Fetch Orders by Transaction Hash

### LINKED ISSUE #1191: Filter orders by transaction hash.
We should implement an additional query filter to filter orders by transaction hash.

bh2smith:
Closes #1191 

Implements a new api GET route for fetching orders  by transaction hash.

This PR gives rise to the issue of handling multiple settlements in a single block which is not yet considered in our ORM

New issue described here: https://github.com/gnosis/gp-v2-services/issues/1214

### Test Plan

Introduced ignored tokio test with manually inserted orders, trades and settlement. Verify that the query returns expected orders.


shared/src/lib.rs
```diff
@@ -53,6 +53,18 @@ impl FromStr for H160Wrapper {
     }
 }
 
+/// Wraps H160 with FromStr and Deserialize that can handle a `0x` prefix.
```
bh2smith, line 56:
```suggestion
/// Wraps H256 with FromStr and Deserialize that can handle a `0x` prefix.
```

vkgnosis:
> This is a bit tricky to test locally since it would require having a local DB which contains sufficient orders and settlements. I might propose merging to staging (since the new route won't affect the system logic) and testing the new endpoint there.

We have readonly access to the database right? You can write a #ignore unit test that executes the query with that database.

orderbook/src/database/orders.rs
```diff
@@ -321,6 +323,26 @@ impl OrderStoring for Postgres {
...
 
+    async fn orders_for_tx(&self, tx_hash: &H256) -> Result<Vec<Order>> {
+        #[rustfmt::skip]
+        const QUERY: &str = concatcp!(
+            "SELECT ", ORDERS_SELECT,
+            "FROM ", ORDERS_FROM,
+            "JOIN trades t \
+                ON t.order_uid = o.uid \
+             JOIN settlements s \
+                ON s.block_number = t.block_number \
```
vkgnosis, line 334: Should also join on log index

bh2smith: I really don't think we should be joining on log index. Each settlement and trade event have their own unique log index.

bh2smith: We need the settlements to get the TxHash (it is not part of the trades table).

orderbook/src/database/orders.rs
```diff
@@ -321,6 +323,26 @@ impl OrderStoring for Postgres {
...
+             WHERE t.tx_hash = $1 ",
```
bh2smith, line 335:
```suggestion
             WHERE s.tx_hash = $1 ",
```

bh2smith:
> We have readonly access to the database right? You can write a #ignore unit test that executes the query with that database.

I suppose I can, but wouldn't this require putting secrets in the code or something? Do we have any existing ignore tests which do this already?

vkgnosis:
You put secrets into env variables and make the test read them.

orderbook/src/database/orders.rs
vkgnosis: Forgot the tx hash is in settlements not trades. 
As for the join you're right too that log index would be wrong. My confusions stems from settlement having as the primary key (block, long_index) so I thought we were not uniquely identifying settlements. Well, we kind of still are. I think in other code we have this "find the next larger log index in the same block" logic to get the settlemen. Not sure how feasible that is for this query.

bh2smith: I'm just glad I was able to help clear things up here.

model/src/h256_hexadecimal.rs
```diff
@@ -0,0 +1,102 @@
+use primitive_types::H256;
```
nlordell, line 1: Is this module really needed? I thought that the default `Serialize` and `Deserialize` implementation did exactly this already.

shared/src/lib.rs
```diff
@@ -53,6 +53,18 @@ impl FromStr for H160Wrapper {
     }
 }
 
+/// Wraps H256 with FromStr and Deserialize that can handle a `0x` prefix.
+#[derive(Debug, Deserialize)]
+#[serde(transparent)]
+pub struct H256Wrapper(#[serde(with = "h256_hexadecimal")] pub H256);
+impl FromStr for H256Wrapper {
```
nlordell, line 60: Is this type also really necessary? From what I understand, there already exists `impl FromStr for H256` and it does exactly this.

orderbook/src/orderbook.rs
```diff
@@ -172,7 +173,7 @@ impl Orderbook {
         Ok(orders)
     }
 
-    pub async fn get_order(&self, uid: &OrderUid) -> Result<Option<Order>> {
+    pub async fn get_order_for_uid(&self, uid: &OrderUid) -> Result<Option<Order>> {
```
nlordell, line 176: uber-naming-nit: I kind of prefer just `get_order` here as it takes a single UID as a parameter.

orderbook/openapi.yml
```diff
@@ -161,6 +161,24 @@ paths:
           description: Invalid signature
         404:
           description: Order was not found
+  /api/v1/orders/{txHash}:
```
nlordell, line 164: Instead of implementing a new route like this, I think from an API perspective it would make more sense to add a new query parameter to the existing `api/v1/orders` endpoint (https://github.com/gnosis/gp-v2-services/blob/7428f278d961a51a555bd4b940ddd7a1cf5e4804/orderbook/openapi.yml#L57-L109).

One reason I'm kind of reluctant to use this endpoint is that it feels not very "REST"-ty. Specifically, here, you aren't returning some transaction data (which is what the URL would suggest to me), but rather orders that match some transaction hash. As such I would suggest either:
- `api/v1/orders?txHash=0x...`: This is what I slightly prefer, although I understand that the maybe the existing `api/v1/orders` may be getting too complex. That being said, it would allow you to do things like "query all orders that sold token X in transaction Y": `api/v1/orders?sellToken=X&txHash=Y`.
- `api/v1/transactions/{txHash}/orders`: This way the URL would clearly indicate that orders are the data that is being returned. It would also add a more natural place to return trades that happened in a specific transaction: `api/v1/transactions/{txHash}/trades`

What do others think?

Also looping in @anxolin and @alfetopito

nlordell (commented):
I tried locally, and I think both the `h256_hexadecimal` and `H256Wrapper` are unnecessary:
```
use ethereum_types::H256;

fn main() {
    assert_eq!(
        serde_json::to_string(&H256([0x42; 32])).unwrap(),
        "\"0x4242424242424242424242424242424242424242424242424242424242424242\"",
    );
    assert_eq!(
        "0x4242424242424242424242424242424242424242424242424242424242424242"
            .parse::<H256>()
            .unwrap(),
        H256([0x42; 32]),
    );

    println!("OK.");
}
```

orderbook/openapi.yml
anxolin: 100% agree with @nlordell , I would have written exatly the same (but I would've explained worse)

problem with `/api/v1/orders/{txHash}` is that u cannot narrow down a collection with the ID of a different entity.
- `/api/v1/orders/{orderId}` would be OK
- both options of nick above are correct for filtering by tx, i also slightly prefer the first one, just cause it's an optional field of the orders endpoint, so we don't create additional entities for little gain.

model/src/h256_hexadecimal.rs
bh2smith: I was just mimicing what we have done for H160Wrapper... I thought this had something to do with how it is parsed from string to and from DB...

shared/src/lib.rs
bh2smith: Could very well be... I will see if I can remove it.

orderbook/openapi.yml
alfetopito: I'm in favour of the later option `api/v1/transactions/{txHash}/orders` and `api/v1/transactions/{txHash}/trades`

IMO it's more clear and doesn't mix up with the existing /orders/ endpoint and follows the same pattern as the new user endpoints (`api/v1/account/{owner}/orders|trades`)

I don't expect to have that many results as to need pagination or server side filtering, so I see little benefit in being able to 
> query all orders that sold token X in transaction Y

orderbook/src/orderbook.rs
nlordell: No strong feelings here. If you feel like this makes the code easier to read, then leave it in.

vkgnosis (approved):

josojo (approved):
Nice, clean PR.

Regarding the feasibility of the sql query: I agree that this is the likely the best we can do right now. On the one handside, I was thinking that we should create a new table that contains that mapping: tx_hash, [orderuid], [interactions] and it should be populated at the time of the settlement. I guess it would be useful for future analytics. 
Though, on the other hand, hopefully we will get that data from a subgraph as well.

orderbook/src/api/get_orders_by_tx.rs
```diff
@@ -0,0 +1,117 @@
...
+                    order_uid: order.order_meta_data.uid,
+                    ..Default::default()
+                }),
+            )])
+            .await
+            .unwrap();
+        }
+
+        let res = db.orders_for_tx(&H256::from_low_u64_be(1)).await.unwrap();
+        assert_eq!(res, orders);
```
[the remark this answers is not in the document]
bh2smith: Yea good point. Will do.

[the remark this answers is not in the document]
vkgnosis: Fine either way.

nlordell (approved):
Looks great!

PR #1195, opened 2021-10-05
Write network ID in deployment script

nlordell:
This PR modifies the deployment script so that it doesn't rely on a hardcoded network ID. This allows us to use more test networks that don't necessarily have an option for configuring network IDs (like Hardhat).

### Test Plan

E2E test no longer uses deterministic network ID, and should continue to pass.

### DIFF
--- a/contracts/src/bin/deploy.rs
+++ b/contracts/src/bin/deploy.rs
@@ -32,6 +32,10 @@ async fn run() -> Result<()> {
     log::info!("checking connection to local test node {}", NODE_URL);
     wait_for_node(&web3).await?;
 
+    let network_id = web3.net().version().await.expect("get network ID failed");
+    write_network_id(&network_id)?;
+    log::info!("connected to test network {}", network_id);
+
     let accounts: Vec<Address> = web3.eth().accounts().await.expect("get accounts failed");
     let admin = accounts[0];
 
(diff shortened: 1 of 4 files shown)


vkgnosis (approved):

contracts/src/bin/deploy.rs
```diff
@@ -32,6 +32,10 @@ async fn run() -> Result<()> {
     log::info!("checking connection to local test node {}", NODE_URL);
     wait_for_node(&web3).await?;
 
+    let network_id = web3.net().version().await.expect("get network ID failed");
+    write_network_id(&network_id)?;
+    log::info!("connected to test network {}", network_id);
```
bh2smith, line 37: Do we really want to explicitly state "test" network here? Is this a test file? Maybe we should call this file `deploy_test.rs` if this is the case.

bh2smith (approved):
Looks good, just got a bit confused with the fact that deploy.rs is entirely meant for testing...

contracts/src/bin/deploy.rs
nlordell: This is the deploy script for deploying contracts for E2E testing. I'm not a fan of adding `_test` postscript to this file.

PR #1199, opened 2021-10-05
Revert "Add full_fee_amount column (#1173)"

taminomara:
Forgot to disable automerge. I've rebased that PR ad added some significant changes that weren't reviewd, so it wasn't ready for merge.


taminomara:
Ok I've realised that I can't merge this because sql migration is already applied on staging. Ok, we will go with post-commit review then...

PR #1203, opened 2021-10-06
Include transport identifer in logs

vkgnosis:
In order to distinguish the messages when multiple transports are in
use. This only matters for the solver at the moment when multiple
transaction submission nodes are used. There I decided to go with a
simple index into the initial argument list. Alternatively we could pick
the base url but it is unclear what part of such urls is needed or
secret.

### Test Plan
CI

### DIFF
--- a/shared/src/transport/http.rs
+++ b/shared/src/transport/http.rs
@@ -24,26 +24,29 @@ struct Inner {
     url: Url,
     id: AtomicUsize,
     metrics: &'static TransportMetrics,
+    /// Name of the transport used in logs to distinguish different transports.
+    name: String,
 }
 
 impl HttpTransport {
-    pub fn new(client: Client, url: Url) -> Self {
+    pub fn new(client: Client, url: Url, name: String) -> Self {
         Self {
             client,
             inner: Arc::new(Inner {
                 url,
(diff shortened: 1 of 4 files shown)


shared/src/transport/http.rs
```diff
@@ -24,26 +24,29 @@ struct Inner {
...
 impl HttpTransport {
-    pub fn new(client: Client, url: Url) -> Self {
+    pub fn new(client: Client, url: Url, name: String) -> Self {
         Self {
             client,
             inner: Arc::new(Inner {
                 url,
                 id: AtomicUsize::new(0),
                 metrics: TransportMetrics::instance(get_metric_storage_registry()).unwrap(),
+                name,
```
nlordell, line 39: Maybe for now, we can default name to the URL to avoid having more things to configure? Not sure if it makes the logs too noisy though.

nlordell (approved):

shared/src/transport/http.rs
vkgnosis: The current default is in the index in the argument so that there is nothing to configure. I think this is fine. Mostly wanted to explain why I rejected the other way. If we see in the future that we would really like custom names we can still make it configurable.

PR #1204, opened 2021-10-06
Use Order Full Fee Amount In Solver Objective Value

### LINKED ISSUE #1202: Use Order Full Fee Amount In Solver Objective Value
We currently compute the order's full fee amount by computing `order.fee_amount / fee_factor`. This has a couple of issues:
- It is incorrect for orders with additional partner fee subsidies
- It is incorrect in cases of fee subsidy configuration changes.

Now that we added a new `full_fee_amount` value to the order metadata, we can now use this value for computing things like the objective value as well as passing this to the solver.

taminomara:
Fixes #1202

We now store full fee amount in the database, so we don't have to calculate it based on the fee factor.

The old calculation logis is preserved for the migration period. We will remove it after the next release, once there are no active orders with missing full fee amount.

### Test Plan
CI


solver/src/driver.rs
```diff
@@ -327,9 +328,13 @@ impl Driver {
...
-                // Because of a potential fee discount, the solver fees may by themselves not be sufficient to make a solution economically viable (leading to a negative objective value)
-                // We therefore reverse apply the fee discount to simulate unsubsidized fees for ranking.
-                let unsubsidized_solver_fees = settlement.total_fees(prices) / BigRational::from_float(self.fee_factor).expect("Discount factor is not a rational");
+                let mut unsubsidized_solver_fees = settlement.total_unsubsidized_fees(prices);
+                // Old orders don't have unsubsidized fees set, so we use the old logic
+                // of restoring the unsubsidized fee from the subsidized one by reversing the fee factor.
+                // This logic will be removed in the future.
+                if unsubsidized_solver_fees.is_zero() {
+                    unsubsidized_solver_fees = settlement.total_fees(prices) / BigRational::from_float(self.fee_factor).expect("Discount factor is not a rational");
+                }
```
vkgnosis, line 337: This wouldn't be correct when one settlement contains some orders with the new field set and some without, right? But I don't think it's a problem since that should be a short time period.

taminomara: Yeah you're right. I'll add a check for this.

vkgnosis (approved):

bh2smith (approved):

PR #1215, opened 2021-10-07
Price fee_in_eth via sell order

fleupold:
More context: https://gnosisinc.slack.com/archives/C022S0FM1GW/p1633618275006100

DEX Aggregators tend to do better at sell orders than buy orders (since some DEXs don't support them). E.g. Paraswap allows selling a fixed stkAave for ETH but not buying a fixed ETH for stkAave. Since we are using Aggregators for our price estimation, using buy orders for estimating `fee_in_eth` reduces the number of trading pairs we can offer.

This PR switches the buy order to a sell order. Drawback is that we are likely more interested in the other price direction (since we want to sell the fee token later for ETH), so this estimate will be off by the spread of the pair (~0.6% per uni v2 hop). 

However, I didn't see a way of doing it otherwise without having more than 1 query (slightly worried about number of requests we are sending).

### Test Plan
Spun up an instance locally. See that I can now trade stkAave for WETH


vkgnosis (approved):

nlordell (approved):
The one thing that was nice about using buy orders is that it more accurately represents what the withdraw service would do with the fees - i.e. it would buy the native token with the sell token.

On the otherhand, if this makes us support more tokens, then its definitely worth it!

PR #1216, opened 2021-10-07
Remove explicit h160 hexadecimal code

vkgnosis:
The type already implements the proper serialization.

As we rediscovered recently H256 already serializes as we want to with the 0x hex prefix. This is the case for H160 too so we no longer need the custom code.

### Test Plan
Existing tests in model enforce that serialization and deserialization still works.


nlordell (approved):
Nice cleanup!

bh2smith (approved):

PR #1217, opened 2021-10-07
Make alerter not use Order struct from model

vkgnosis:
The motivation is that if we add a new mandatory field the staging
alerter would fail to read the prod api. This also makes the alerter
more independent of the rest of the code and allows it to detect if we
accidentally make an incompatible change.

Downside is that it is more work if we do want to breaking change the api but I feel it is reasonable to do it by hand in that case.

### Test Plan
manually ran to see it works with prod


nlordell (approved):
Fine by me!

bh2smith (approved):

PR #1220, opened 2021-10-07
* Add support for scaling fees for objective values

nlordell:
We had previously discussed using the solver's `fee_factor` parameter to scale fees when computing objective values. It was meant as a way to avoid negative objective values and promote more aggressive merging of single order settlements.

Now that we have a `full_fee_amount`, the existing `fee_factor` can no longer be used for this purpose, so this PR introduces a new `fee_objective_scaling_factor`. Instead of applying the scaling factor (as well as the compatibility logic for missing `full_fee_amount`) I changed the `LimitOrder` to now pass around the scaled unsubsidized fee with the limit order.

This adds a new `FEE_OBJECTIVE_SCALING_FACTOR` configuration value.

### Test Plan

Added unit test to check unsubsidized fee scaling logic.


solver/src/main.rs
```diff
@@ -191,6 +192,13 @@ struct Arguments {
...
     #[structopt(long, env, use_delimiter = true)]
     liquidity_order_owners: Vec<H160>,
+
+    /// Fee scaling factor for objective value. This controls the constant
+    /// factor by which order fees are multiplied with. Setting this to a value
+    /// greater than 1.0 makes settlements with negative objective values less
+    /// likely, promoting more aggressive merging of single order settlements.
+    #[structopt(long, env, default_value = "1", parse(try_from_str = shared::arguments::parse_fee_factor))]
+    pub fee_objective_scaling_factor: f64,
 }
```
bh2smith, line 202: I was under the impression this constant term used to make negative objective positive was going to be an additive constant. However, I guess this works too...

bh2smith (approved):
Seems legit.

vkgnosis:
Didn't think about this in detail but I thought we would add a factor to the objective value computation. Whereas here we are changing the input order data. Any preference? Or is it necessary to do it this way?

solver/src/settlement.rs
```diff
@@ -11,14 +12,13 @@ use primitive_types::{H160, U256};
...
 
-pub use settlement_encoder::SettlementEncoder;
-
 #[derive(Clone, Debug, Default, Eq, PartialEq)]
 pub struct Trade {
     pub order: Order,
     pub sell_token_index: usize,
     pub buy_token_index: usize,
     pub executed_amount: U256,
+    pub scaled_fee_amount: U256,
```
fleupold, line 21: I think a comment on what we mean by scaled here and in the other struct where it appears would be good.

fleupold (commented):
I think the advantage of manipulating the order vs only applying this factor in the ranking is that quasimodo and mip will also benefit from this (and try to merge more aggressively)

> Furthermore, I noticed that the fee being given to the MIP/Quasimodo solver was the original fee amount, and not the unsubsidized fee amount meaning that it wasn't getting the fee amount that was used for the objective value, putting it at a disadvantage.

The fee subsidy should have been accounted for here: https://github.com/gnosis/oba-services/blob/205ab4fefcbb1d30ee8739a9dda7894a713e76f3/solver/src/solver/http_solver.rs#L347

vkgnosis (approved):

PR #1221, opened 2021-10-08
Use native token sell order for computing min fee

nlordell:
This PR re-introduces #1215.

Additionally, it fixes the sell token amount computation, as we were using the inverted price. A unit test was added to verify everything works as expected.

### Test Plan

Added new unit test.


bh2smith (approved):
So quick! Excellent 🕵🏽 work

vkgnosis (approved):

Issue #882, opened 2021-07-21
Alternative for inflight trades


vkgnosis:
Currently we keep track of trades from a completed settlement transaction as inflight trades and exclude them from the next driver loop. We do this because we ran into issues where the driver saw the transaction as mined and thus went into the next loop while the order book did not see that block yet. It would then return these orders again which then fail to simulate on the driver.
Maybe another way to do this would be to return from the order book the block at which it last updated as part of the solvable_orders call. With that we can compare it in the driver and wait until we see that the order book has seen the block that had our settlement transaction.
This is related to https://github.com/gnosis/gp-v2-services/issues/879 because the inflight orders don't mesh well with it. But also not sure if this is needed yet.

nlordell:
This has been fixed by having the orderbook API also supply its last indexed block when querying solvable orders.

PR #1222, opened 2021-10-08
Return block number with solvable orders from database

vkgnosis:
Part of https://github.com/gnosis/gp-v2-services/issues/882

The purpose is to return the block number of the most recently observed settlement with the solvable orders in the api. Then the driver can use of this to wait until the api has seen a recently submitted settlement transaction.

This PR does not yet change the http route. I will make a v2 solvable_orders route where we return this struct instead of just a list of orders. Unless someone has a different idea. Then we can eventually deprecate the v1 route.

### Test Plan
existing tests, new postgres test

### DIFF
--- a/orderbook/src/database/orders.rs
+++ b/orderbook/src/database/orders.rs
@@ -5,7 +5,7 @@ use bigdecimal::{BigDecimal, Zero};
 use chrono::{DateTime, Utc};
 use const_format::concatcp;
 use ethcontract::H256;
-use futures::stream::TryStreamExt;
+use futures::{stream::TryStreamExt, FutureExt};
 use model::{
     app_id::AppId,
     order::{
@@ -15,6 +15,7 @@ use model::{
     signature::{Signature, SigningScheme},
 };
 use primitive_types::H160;
+use sqlx::Connection;
 use std::{borrow::Cow, convert::TryInto};
 
 #[cfg_attr(test, mockall::automock)]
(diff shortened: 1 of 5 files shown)


orderbook/src/database/orders.rs
```diff
@@ -352,12 +358,30 @@ impl OrderStoring for Postgres {
...
-            .and_then(|row: OrdersQueryRow| async move { row.into_order() })
-            .try_collect()
+        let mut connection = self.pool.acquire().await?;
+
+        connection
+            .transaction(move |transaction| {
+                async move {
+                    let orders = sqlx::query_as(QUERY)
+                        .bind(min_valid_to as i64)
+                        .fetch(&mut *transaction)
```
vkgnosis, line 368: I was struggling here because the compiler would say that in the next query below `transaction` had been moved in this line even though references shouldn't be moved and and `transaction` is a reference here. I don't fully understand the problem but it is related to `fetch` taking a generic parameter so the compiler treats the `transaction` as not being a reference. This is the fix I came up with.

orderbook/src/database/orders.rs
```diff
@@ -35,6 +36,11 @@ pub trait OrderStoring: Send + Sync {
     ) -> Result<Vec<Order>>;
 }
 
+pub struct SolvableOrders {
+    pub orders: Vec<Order>,
+    pub newest_settlement_block: u64,
+}
```
bh2smith, line 42: Wouldn't it be equivalent to return `Vec<Orders>` as before along with a u64 block number from the trait method. This way, we wouldn't need the new "single purpose struct" for the change here? Of course, its also fine the way it is I am more curious about your choice here than suggesting a change.

bh2smith (approved):

orderbook/src/database/orders.rs
```diff
@@ -35,6 +36,11 @@ pub trait OrderStoring: Send + Sync {
     ) -> Result<Vec<Order>>;
 }
 
+pub struct SolvableOrders {
+    pub orders: Vec<Order>,
+    pub newest_settlement_block: u64,
```
nlordell, line 41: naming nit: `latest_settlement_block`.

nlordell (approved):

PR #1224, opened 2021-10-08
Call quasimodo differently if there is just one single order.

marcovc:
This PR changes the ucp_policy parameter value when calling quasimodo, setting
a different value depending if the current instance has one single order, or
multiple orders.

The ucp_policy controls how Uniform Clearing Prices are enforced in the solver.
Because of the yet-to-solve surplus distribution problem, we are currently
asking quasimodo to enforce UCP over everything always (which makes it very slow).

However, when there is only one order in the problem, then there are no
surplus distribution issues, so we can safely call quasimodo with a more relaxed
and efficent set of constraints.

### Test Plan
Tested locally.


nlordell (approved):
Looks good to me! I think you are just missing a `cargo fmt`.

bh2smith (approved):
Lovely! all you need to do it `cargo fmt`

PR #1230, opened 2021-10-11
[Metrics] Surplus Reporting with outlier log.

### LINKED ISSUE #1207: Log when there is an order which would have gotten more surplus in a single solti
In order to inspect and protect against edge cases, it would be nice if we could check after the run loop if for any order in the submitted solution, there existed another solution candidate in which this order had received significantly better surplus (let's say more 0.1% difference).

We would like to investigate and understand these batches if they occur.

bh2smith:
Closes #1207

Inside driver settlement handler right around the place where we rename the method "report_matched_but_unsettled_orders" to `report_on_batch` so to include both 

- `report_matched_but_unsettled` orders along with the
- `order_surplus_report` introduced here.

Note that both of these methods create metrics by performing analysis on the winning settlement in comparison with all the other successful but not chosen settlements.

The surplus report does the following:

- aggregate information regarding winning settlement's surplus for each trade and the best surplus found for each of the touched trades over all alternative settlements.
- log a warning when settled surplus of an order is 1% worse than the same order's surplus if it had been matched by a different settlement that was not the winning solution. 
- Report a metric based on the tuple 

```py
[(order.submitted_surplus, order.best_alternative) for order in submitted_orders]
```

The report consists of histogram data `(winning_solver, best_alternative_solver, surplus_ratio_diff)` for each of the orders contained in the submitted order set (if there exists an alternative).

### Test Plan
Unfortunately the reporting method has become quite complex and no longer returns any values, making it difficult to test. I would like to simplify this method into a helper which builds the best alternative map.

and can perhaps test this.

Run the solver, visit the metrics page and see that this counter has been included.


solver/src/driver.rs
```diff
@@ -285,23 +288,78 @@ impl Driver {
...
+            .iter()
+            .map(|trade| {
+                (
+                    trade.order.order_meta_data.uid,
+                    trade
+                        .surplus(
+                            &prices[&trade.order.order_creation.sell_token],
+                            &prices[&trade.order.order_creation.buy_token],
+                        )
+                        .expect("I can't imagine submitting a trade with None surplus"),
```
vkgnosis, line 309: Following the calls to surplus around a bit I think this can happen because we have one function that turns None into 0 in Settlement::total_surplus and we might settle a settlement with 0 surplus.

solver/src/driver.rs
```diff
@@ -285,23 +288,78 @@ impl Driver {
         }
     }
 
-    // Record metric with the amount of orders that were matched but not settled in this runloop (effectively queued for the next one)
-    // Should help us to identify how much we can save by parallelizing execution.
-    fn report_matched_but_unsettled_orders(
+    // Record metric with surplus achieved in winning settlement
+    // vs that which was unrealized in other feasible solutions.
+    fn surplus_report(
         &self,
```
vkgnosis, line 294: Since this function doesn't use anything from `self` except `self.metrics` maybe we could refactor it into a pure free standing function that can be independently tested and live in another module so that the driver.rs file doesn't get even bigger. That function could return the count and then here in driver.rs we call the refactored function and set the metric.

bh2smith: Sure thing. I will give this a try.

bh2smith:
> As a test case, can we verify we would have been warned about https://etherscan.io/tx/0x8e7031670c5540e8c1bed63a95ffa52bcf496c17426cc2a89e2c9880f9ab2b65

Yes a very interesting and fun test case. I suppose I will fetch the instance, code up all the input parameters (matched and unmatched trades and prices) from the found solution logs and put them directly into a unit test!

solver/src/driver.rs
vkgnosis: I only glanced at the code so I'm not sure. I think it could happen on some overflow. It is probably a very rare case so ignoring it for this metric is fine but we shouldn't panic.

bh2smith: Right... so ignoring this case here without panicing sounds like `unwrap_or(ZERO)` to me.

solver/src/driver.rs
bh2smith: Cool, I will surely put it on my list of TODOs though, because it seems fun nonetheless.

nlordell (commented):
Code looks good. I like the new module, I would suggest measuring this differently to make it more flexible.

solver/src/analytics.rs
```diff
@@ -0,0 +1,159 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    let mut count = 0;
+    for (order_id, submitted) in submitted_surplus.iter() {
+        let best = best_surplus.get(order_id).expect("exists by construction");
+        if best > &(BigRational::new(101.into(), 100.into()) * submitted) {
```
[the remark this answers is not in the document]
bh2smith: Yes. You mean the 1% right. I completely agree. I can add this simple thing in a follow up unless anyone insists it happen here. Other wise I'll leave a todo with an issue.

solver/src/analytics.rs
```diff
@@ -0,0 +1,159 @@
...
+            count += 1;
+            tracing::warn!("Found Matched order {:?} whose submission surplus was 1% worse than in another, valid, lower ranked settlement", order_id)
+        }
+    }
+    count
```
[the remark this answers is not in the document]
bh2smith: This is a great suggestion. I will have to tinker a bit with the definition of buckets since there is a division by zero singularity when the submitted surplus is zero.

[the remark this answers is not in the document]
nlordell: Division by 0 shouldn't be possible if we do surplus difference (i.e. `settled_trade.surplus() - solved_but_not_settled_trade.surplus()`).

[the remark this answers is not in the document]
bh2smith: Yes but that difference alone is not a relative measure and won't fit well into buckets without some sort of normalization. Normalization of this difference is where the zero division comes from.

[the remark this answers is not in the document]
nlordell: Sorry, by `.surplus()` I meant `.surplus_percentage()`. Technically, division by 0 can happen only if the buy/sell amounts are 0 right? Since orders with 0 amounts aren't allowed by the API[1] it shouldn't ever happen. Either way, we can just skip orders with 0 amounts to avoid panics if the orderbook every misbehaves.

[1]: https://github.com/gnosis/gp-v2-services/blob/d52387e96d7a0ffbae62b547f0b79ad88460c2f5/orderbook/src/api/order_validation.rs#L297-L299

solver/src/analytics.rs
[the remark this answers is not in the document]
bh2smith: https://github.com/gnosis/gp-v2-services/issues/1350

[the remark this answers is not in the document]
nlordell: Maybe, since we plan on relying on metrics more than logs, we can just log all cases and not miss cases that we wanted to know about (for example 0.9% on a 10M USD trade is 9K USD and wouldn't get logged).

solver/src/driver.rs
```diff
@@ -270,15 +271,25 @@ impl Driver {
...
+        // Report surplus
+        self.metrics.unsettled_orders_with_better_surplus(
+            analytics::num_unsettled_orders_with_better_surplus(
+                submitted,
+                all.clone()
+                    .map(|rated_settlement| rated_settlement.settlement),
+            ),
+        );
+        // Report matched but not settled
+        // TODO - move all this data manipulation into analytics.rs
```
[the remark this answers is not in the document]
bh2smith: Right, yes, this is something I could do now. However, I was already pretty deep into this by the time I realized I should do it.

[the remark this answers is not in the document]
nlordell: Its fine to do it after, of course. One workflow that worked for me in the past is:
- Make some changes :keyboard: 
- Notice I need to move code into a new module.
- `git stash`
- `git checkout main && git checkout -b move-code-into-new-module`
- Move the code into the new module
- If you have previously existing commits on your separate branch:
  - `git checkout existing-changes && git rebase --onto move-code-into-new-module main`
- `git stash pop`

Then you can make a small PR with just the module changes and continue working on your larger change.

solver/src/analytics.rs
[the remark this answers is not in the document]
nlordell: I think you may have missed the comment :point_up:

solver/src/settlement.rs
```diff
@@ -280,20 +295,40 @@ fn sell_order_surplus(
...
     }
 }
 
+fn surplus_ratio(
+    sell_token_price: &BigRational,
+    buy_token_price: &BigRational,
+    sell_amount_limit: &BigRational,
+    buy_amount_limit: &BigRational,
+) -> Option<BigRational> {
+    if buy_token_price.is_zero() {
```
nlordell, line 321: I think you need to check that both `buy_token_price` and `buy_amount_limit` are not `0` or you could panic.

solver/src/settlement.rs
```diff
@@ -674,14 +707,53 @@ pub mod tests {
             Some(r(5000))
         );
 
-        // Buy Token worth twice as much as sell token. If we were willing to buy at 3:1, we will
-        // have a surplus of 10 sell tokens, worth 200 each.
+        // Buy Token worth twice as much as sell token. If we were willing to sell at 3:1, we will
+        // have a surplus of 10 buy tokens, worth 200 each.
         assert_eq!(
-            buy_order_surplus(&r(100), &r(200), &r(60), &r(20), &r(20)),
+            sell_order_surplus(&r(100), &r(200), &r(60), &r(20), &r(60)),
```
nlordell, line 713: Nice catch!

solver/src/analytics.rs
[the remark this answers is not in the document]
bh2smith: I will log whenever the best alternative is better and in #1350 we can configure when to error instead!

solver/src/driver.rs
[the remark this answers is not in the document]
bh2smith: Thanks - I will give this a try sometime.

solver/src/metrics.rs
```diff
@@ -278,6 +299,17 @@ impl SolverMetrics for Metrics {
...
     }
 
+    fn report_order_surplus(
+        &self,
+        winning_solver: &str,
+        best_alternative: &str,
+        surplus_diff: f64,
+    ) {
+        self.order_surplus_report
+            .with_label_values(&[winning_solver, best_alternative])
```
nlordell, line 309: One downside is that this will create a loooooot of permutations (`n * (n - 1)` where `n` is the number of solvers, so 42 in prod).

Maybe, we may consider not using labels at all for this metric and rely on logs when this does happen.

On the other hand it would be cool to measure things like how often solver X beat out Y but didn't get picked. However, maybe Prometheus metrics is not the right way to measure this kind of thing.

Don't have strong opinions either way, will differ to other's judgement on this.

solver/src/analytics.rs
```diff
@@ -0,0 +1,123 @@
...
+            self.ratio.to_f64().unwrap_or(f64::NAN)
+        )
+    }
+}
+
+fn get_prices(settlement: &Settlement) -> HashMap<&H160, BigRational> {
+    settlement
+        .clearing_prices()
+        .iter()
+        .map(|(token, price)| (token, price.to_big_rational()))
```
nlordell, line 36: ubernit: Since `H160` is a `Copy` type, I think it makes slightly more sense to:
```suggestion
fn get_prices(settlement: &Settlement) -> HashMap<H160, BigRational> {
    settlement
        .clearing_prices()
        .iter()
        .map(|(token, price)| (*token, price.to_big_rational()))
```

This may also simplify lifetime things down the line, and it makes things a little less awkward when getting items (you don't need `&&H160`s). For example no need to take the reference of the sell token address on line 53:
```rust
            let sell_token_price = &submitted_prices[trade.order.order_creation.sell_token];
```

solver/src/analytics.rs
```diff
@@ -0,0 +1,123 @@
...
+    for (order_id, submitted) in submitted_surplus.iter() {
+        if let Some(alternative) = best_alternative.get(order_id) {
+            metrics.report_order_surplus(
+                winning_solver,
+                alternative.solver_name,
+                (&submitted.ratio - &alternative.ratio)
+                    .to_f64()
+                    .unwrap_or_default(),
+            );
+            if alternative.absolute > submitted.absolute {
```
nlordell, line 81: I think you are comparing absolute surpluses without normalizing prices which is dangerous (since just scaling prices will have an effect on this value). For example, when computing the objective value, we scale surplus using external prices. Maybe it needs to be done here as well.

solver/src/analytics.rs
```diff
@@ -0,0 +1,123 @@
...
+                tracing::warn!("submission surplus worse than lower ranked settlement; order {:?} submitted {}, best alternative {}", order_id, submitted, alternative)
```
nlordell, line 82: You can use cool `tracing` notation to make this slightly more succinct:
```suggestion
                tracing::warn!(
                    ?order_id, %submitted, %alternative
                    "submission surplus worse than lower ranked settlement",
                )
```

(convenient shorthand where `?some_field` adds `"some_field = {:?}"` to the log message and `%some_field` adds `"some_field = {}"` to the message)

nlordell (commented):
Looks good. My only concern is that we are reporting absolute surplus differences calculated with prices used for settlements (so surpluses are in different units).

solver/src/analytics.rs
bh2smith: Why don't I just compare the ratios here then. That's what they are for.

solver/src/metrics.rs
bh2smith: Yes. It would be much simpler if we just report the diffs although then we could only use this for alerting and have to rely on the logs. I agree having this many new tiles in the dashboard would be silly and not so easy to look at, but from a data perspective it would be useful and we wouldn't need to actually use the labels in the grouping for our grafana alert.

solver/src/analytics.rs
bh2smith: I was able to make your suggested change alright, but was not able to remove the reference when accessing the price map.

solver/src/analytics.rs
bh2smith: Changed to ratio

solver/src/analytics.rs
bh2smith: Very cool. Very happy to learn this.

solver/src/analytics.rs
```diff
@@ -0,0 +1,126 @@
...
+                Entry::Occupied(mut entry) => {
+                    let value = entry.get_mut();
+                    if value.absolute < surplus.absolute {
+                        *value = surplus;
+                    }
+                }
+                Entry::Vacant(entry) => {
+                    entry.insert(surplus);
+                }
+            }
```
nlordell, line 122: ubernit: Note that I think this can be written more succinctly with:
```suggestion
            best_surplus.entry(order_id)
                .and_modify(|entry| {
                    if entry.absolute < surplus.absolute {
                        *entry = surplus.clone();
                    }
                })
                .or_insert(surplus);
```

solver/src/metrics.rs
nlordell: Sounds good. Lets get rid of the labels for now then.

solver/src/analytics.rs
```diff
@@ -0,0 +1,126 @@
...
+use std::{
+    collections::{hash_map::Entry, HashMap},
+    fmt::{Display, Formatter},
+    sync::Arc,
+};
+
+#[derive(Clone)]
+struct SurplusInfo {
+    solver_name: &'static str,
+    absolute: BigRational,
```
nlordell, line 16: I think we should get rid of this property for now since we shouldn't use it (since it isn't normalized at all).

solver/src/analytics.rs
```diff
@@ -0,0 +1,126 @@
...
+                    .unwrap_or_else(BigRational::zero),
+                ratio: trade
+                    .surplus_ratio(sell_token_price, buy_token_price)
+                    .unwrap_or_else(BigRational::zero),
+            };
+            let entry = best_surplus.entry(order_id);
+            match entry {
+                Entry::Occupied(mut entry) => {
+                    let value = entry.get_mut();
+                    if value.absolute < surplus.absolute {
```
nlordell, line 115: It looks like you are still comparing absolute surplus here instead of ratios. Maybe get rid of the `absolute` field for now to not accidentally compare them?

solver/src/analytics.rs
```diff
@@ -0,0 +1,126 @@
...
+    sync::Arc,
+};
+
+#[derive(Clone)]
+struct SurplusInfo {
+    solver_name: &'static str,
+    absolute: BigRational,
+    ratio: BigRational,
+}
+
```
nlordell, line 19: nit: I noticed that there was a subtle bug because we had two separate comparisons of `SurplusInfo` that did it in different ways. Maybe we can make a helper function?

```rust
impl SurplusInfo {
    fn is_better_than(&self, other: &Self) -> bool;
}
```

In theory, we could also use `impl PartialOrd for SurplusInfo`, but I'm not a fan of that because you get into weird things where the surplus ratio is the same but the solver name is different.

bh2smith:
Thanks Nick for your very good review. I will veritably include these proposed suggestions.

solver/src/settlement.rs
```diff
@@ -280,20 +295,37 @@ fn sell_order_surplus(
...
     }
 }
 
+fn surplus_ratio(
+    sell_token_price: &BigRational,
+    buy_token_price: &BigRational,
+    sell_amount_limit: &BigRational,
+    buy_amount_limit: &BigRational,
+) -> Option<BigRational> {
+    if buy_token_price.is_zero() || buy_token_price.is_zero() {
```
vkgnosis, line 318: Should one of those conditions be sell_token_price?

solver/src/settlement.rs
```diff
@@ -280,20 +295,37 @@ fn sell_order_surplus(
...
+        return None;
+    }
+    let res = (sell_amount_limit * sell_token_price) / (buy_amount_limit * buy_token_price)
+        - BigRational::one();
+    if res.is_negative() {
+        return None;
+    }
```
vkgnosis, line 325: What is the reasoning for minus 1 and then turning negative into None? Naively I would have assumed a ratio is just a division. Maybe just needs a comment explaining what surplus_ratio is.

solver/src/settlement.rs
vkgnosis, line 318: Just to be safe I think we should also check for the limit amounts being 0. Not sure if we currently accept that in the api but it could happen in the future.

fleupold (commented):
Looking forward tot hat metric!

It's not intuitively clear what `surplus_ratio` refers to me, maybe some more documentation could help?

solver/src/analytics.rs
```diff
@@ -0,0 +1,126 @@
...
+                alternative.solver_name,
+                (&submitted.ratio - &alternative.ratio)
+                    .to_f64()
+                    .unwrap_or_default(),
+            );
+            if alternative.ratio > submitted.ratio {
+                tracing::warn!(
+                    ?order_id, %submitted, %alternative,
+                    "submission surplus worse than lower ranked settlement",
+                );
```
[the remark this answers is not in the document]
bh2smith: Its not that it is different, but that something which is better was not chosen. If we had an actual threshold on the magnitude of how much better, I would even argue that it should be an error.

I can make it debug, but I do think such a scenario should to be a warning.

solver/src/driver.rs
```diff
@@ -478,28 +491,29 @@ impl Driver {
...
         }
-
-        if let Some((solver, mut settlement)) = rated_settlements
-            .clone()
-            .into_iter()
-            .max_by(|a, b| a.1.objective_value().cmp(&b.1.objective_value()))
-        {
+        // Reverse sort rated settlements by objective value
+        rated_settlements.sort_by(|a, b| a.1.objective_value().cmp(&b.1.objective_value()));
+        if let Some((wining_solver, mut winning_settlement)) = rated_settlements.pop() {
```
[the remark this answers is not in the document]
bh2smith: Right, yes, this is because I put the comment before I realized that pop is from the back and then never removed it. Thanks for noticing.

solver/src/settlement.rs
bh2smith: It definitely shouldn't be the same, but I don't think its sell token price, its whatever is in the denominator -- Will fix.

bh2smith: It should be fine to ensure the denominator values are zero.

solver/src/metrics.rs
vkgnosis: I read

> Lets get rid of the labels for now then.

as saying there wouldn't be labels. That is what I agreed with. I am against abusing Prometheus for use cases it wasn't made for - like high dimensional data. I think it makes sense as a single dimensional metric for alerts like you said in previous post.

solver/src/analytics.rs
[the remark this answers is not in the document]
vkgnosis: This warns for any difference, right? I don't see why that would be a warning. It does not indicate a failure because it is not surprising that different solvers have slightly different surpluses. A huge difference would could be a warning.

solver/src/analytics.rs
```diff
@@ -0,0 +1,126 @@
...
+            }
+        }
+    }
+}
+
+fn best_surplus_by_order(
+    settlements: Vec<(&'static str, &Settlement)>,
+) -> HashMap<OrderUid, SurplusInfo> {
```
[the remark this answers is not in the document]
bh2smith: I got a warning here that "Function cannot have anonoymous parameters"

solver/src/analytics.rs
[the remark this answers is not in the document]
bh2smith: Right, so this is what I mean by having an actual threshold. In any case I will leave this as a debug for now and only consider warnings/errors once we get a better idea of what a "big" difference actually means.

solver/src/metrics.rs
bh2smith: Right, although I had suggested that the grafana metrics would only ever be single dimensional since we would ignore the labels, but I also read Nicks comment wrong and thought he was agreeing with keeping them and not using them.

Will remove the labels for now, but I have a feeling they will come back in some way or another.

solver/src/analytics.rs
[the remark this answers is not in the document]
nlordell: Did you try `&[(&'static str, &Settlement)]`?

nlordell (approved):
Looks good to me. Just what I imagine may have been a typo :point_down:

solver/src/analytics.rs
```diff
@@ -0,0 +1,126 @@
...
+use num::BigRational;
+use shared::conversions::U256Ext;
+use std::{
+    collections::{hash_map::Entry, HashMap},
+    fmt::{Display, Formatter},
+    sync::Arc,
+};
+
+#[derive(Clone)]
+struct SurplusInfo {
```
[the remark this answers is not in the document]
nlordell: IMO it could just reference the documentation of `Trade::surplus_ratio` which should be there already.

crates/solver/src/metrics.rs
```diff
@@ -124,6 +126,25 @@ impl Metrics {
...
+                "settlement_surplus_report",
+                "Surplus ratio differences between winning and best settlement per order",
+            )
+            .buckets(vec![
+                -f64::INFINITY,
+                -0.1,
+                -0.01,
+                -0.005,
+                0.,
+                0.05,
```
nlordell, line 140:
```suggestion
                0.005,
```

solver/src/analytics.rs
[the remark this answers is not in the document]
bh2smith: Oh, right... silly me.

PR #1231, opened 2021-10-11
V2 solvable orders route

vkgnosis:
Adds a new route that includes the settlement block information. The route is the same except that the returned json is now an object with an extra field for the block.

Had to adjust api.rs code so that it doesn't assume all paths share the same prefix.

### Test Plan
will do manual test

### DIFF
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -131,6 +131,13 @@ impl DomainSeparator {
     }
 }
 
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SolvableOrders {
+    pub orders: Vec<order::Order>,
+    pub latest_settlement_block: u64,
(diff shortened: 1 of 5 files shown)


model/src/lib.rs
```diff
@@ -132,6 +132,13 @@ impl DomainSeparator {
     }
 }
 
+#[derive(Clone, Debug, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SolvableOrders {
```
nlordell, line 137: Curious about two identical `SolvableOrders` struct. Could we just have one?

vkgnosis: The reason there are multiple is that they serve different purposes. The one in model is for serialization, the one in SolvableOrdersCache also contains update time which model doesn't.

nlordell (approved):

bh2smith (approved):

PR #1232, opened 2021-10-11
Drop fee factor from the solvers code

taminomara:
Since full fee amount is now stored in the database, we don't have to restore it from the subsidized fee amount. Thus, we don't need fee factor in the solvers code.

See #1204.

### Test Plan
CI


vkgnosis (approved):

nlordell (commented):
Looks good! Can we also move the `fee_factor` from `shared::arguments::Arguments` to `orderbook::Arguments` since we don't need it at all in the `solver` binary anymore?

nlordell (approved):

bh2smith (approved):
Nice!

PR #1243, opened 2021-10-12
[Metrics] Matched order classification counts

### LINKED ISSUE #1233: Matched Orders Metric
The matched orders metric is not looking so good when a bunch of spam orders come into the orderbook (I am assuming they were nearly identical). 

One proposed solution was to exclude canceled orders from the metric like so

```
total_trades / (total_orders_created - total_orders_cancelled)
```

but have concluded that we should introduce a new metric for this while keeping the old one. 

Furthermore, we would like exclude liquidity provider orders in both metrics.

bh2smith:
Closes #1233 

Adding new metric along side `matched_but_unsettled`. Specifically that is `matched_but unsettled_liquidity_orders`

Note that we probably also want to keep track of settled liquidity orders as well, but in order to have any meaningful values in the our metrics tile we would need both. The value is easily acquired from the code here as the difference

`liquidity_order_ids.len() - matched_but_unsettled_liquidity_ids.len()`

which can be computed directly in grafana

### Test Plan

1. Run the solver pointed at the production orderbook

2. Navigate to the metrics API at http://localhost:9587/metrics and find logs for this new measure as 

```
# HELP gp_v2_solver_orders_matched_liquidity Counter for the number of orders for which at least one solver computed an execution which was from a known liquidity provider
# TYPE gp_v2_solver_orders_matched_liquidity counter
gp_v2_solver_orders_matched_liquidity 0
```


solver/src/driver.rs
```diff
@@ -285,23 +286,58 @@ impl Driver {
...
+            .iter()
+            .map(|order| order.order_meta_data.uid)
             .collect();
-        let matched_but_not_settled: HashSet<_> = all_matched.difference(&submitted).collect();
+        let matched_but_not_settled = all_matched_ids.difference(&submitted).copied().collect();
+        let cancelled_order_ids: HashSet<_> = all_matched_orders
+            .iter()
+            .filter_map(|order| match order.order_meta_data.invalidated {
+                true => Some(order.order_meta_data.uid),
+                false => None,
```
vkgnosis, line 319: Can this ever be true? If I understand the code right we're passing in *sovlable* orders. Those shouldn't be invalidated. Thinking this further I wonder if the driver is even the right place to collect these metrics. We have all the data in Postgres so we could write a script (or rust program) that at any time can create these metrics retroactively instead of doing them live through Prometheus like this.

bh2smith: Yea, perhaps we should just leave canceled orders out for this PR and focus on excluding the liquidity provider orders here. Your suggestion might be worth doing separately. but the purpose of this PR is really meant to focus on our grafan metric not showing bad settlement matching results (i.e. right now its showing < 40% matched) because of a bunch of spam liquidity provider orders.

nlordell (commented):
As @vkgnosis pointed out, I don't think this works because solvable orders don't include any cancelled orders already.

In theory, we already have this information in the metrics (status 200 for cancel order route == number of cancelled orders for example) so I think tracking this may not need additional code changes and just modifications to our Grafana dashboard.

vkgnosis (approved):

solver/src/driver.rs
```diff
@@ -285,23 +286,45 @@ impl Driver {
...
+            .iter()
+            .map(|order| order.order_meta_data.uid)
+            .collect();
+        let matched_but_not_settled = all_matched_ids.difference(&submitted).copied().collect();
+        let liquidity_order_ids: HashSet<_> = orders
+            .into_iter()
+            .filter_map(|order| match order.is_liquidity_order {
+                true => Some(
+                    OrderUid::from_str(&order.id).expect("LimitOrder was constructed from Order"),
+                ),
```
nlordell, line 319: This might cause unexpected panics if we add limit orders from other sources in the future (for example 0x). Might be worth to just filter out failures:
```suggestion
                true => OrderUid::from_str(&order.id).ok(),
```

nlordell (approved):
Looks good. We can now start removing liquidity orders from our matched order metric!

@bh2smith, as a follow up task, can you also make sure to account for this in the Grafana dashboard once it gets merged?

bh2smith:
It turns out this measure isn't going to be immediately useful, but does capture some insight as to single batch information when necessary. I would opt to keep this around for future use.

PR #1247, opened 2021-10-12
Internally use estimate_gas for settlement simulation at current block

vkgnosis:
Estimate gas internally simulates the transaction which tells us whether
the simulation succeeded and how much gas was used in one call.

Next I will change the driver to use this function so that we do not simulate twice.

### Test Plan
Adapted the ignored test to call the new function and ran the test.

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -7,7 +7,7 @@ use crate::{
     metrics::SolverMetrics,
     settlement::Settlement,
     settlement_simulation,
-    settlement_submission::{self, retry::is_transaction_failure, SolutionSubmitter},
+    settlement_submission::{retry::is_transaction_failure, SolutionSubmitter},
     solver::Solver,
     solver::{Auction, SettlementWithSolver, Solvers},
 };
@@ -193,12 +193,13 @@ impl Driver {
             return Ok(false);
         }
 
-        let simulations = settlement_simulation::simulate_settlements(
-            [(solver, settlement.settlement.without_onchain_liquidity())].iter(),
+        let simulations = settlement_simulation::simulate_and_estimate_gas_at_current_block(
+            std::iter::once((
+                solver.account().clone(),
+                settlement.settlement.without_onchain_liquidity(),
+            )),
             &self.settlement_contract,
             &self.web3,
-            &self.network_id,
-            settlement_simulation::Block::LatestWithoutTenderly,
             gas_price_wei,
         )
         .await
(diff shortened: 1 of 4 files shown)


vkgnosis:
Draft while I debug an issue.

solver/src/settlement_simulation.rs
```diff
@@ -85,18 +90,27 @@ pub async fn simulate_settlements(
...
-                    }
-                    Block::LatestWithoutTenderly => err,
-                }
+                Error::new(err).context(tenderly_link(block, network_id, transaction_builder))
             })
         })
         .collect())
 }
 
+fn method(
```
josojo, line 99: This function name seems quite unspecific, maybe get_settle_method() would be a better name?

josojo (approved):
Nice, LGTM

solver/src/driver.rs
```diff
@@ -230,8 +231,8 @@ impl Driver {
             .into_iter()
             .zip(simulations)
             .partition_map(|(settlement, result)| match result {
-                Ok(()) => Either::Left(settlement),
-                Err(err) => Either::Right((settlement, err)),
+                Ok(_) => Either::Left(settlement),
```
nlordell, line 234: One advantage of `Ok(())` is that if the type of the `result` changes, you would get a compiler error to update at this location. Very minor though, and I also see the argument that `Ok(_)` could more clearly communicate the intent of "it succeeded, don't care about the result".

Not requesting any changes, just rambling.

solver/src/driver.rs
```diff
@@ -246,26 +247,29 @@ impl Driver {
...
-                tracing::error!(
-                    "unable to complete simulation of settlements at earlier block {}: {:?}",
-                    current_block_during_liquidity_fetch,
-                    err
-                );
-                return;
-            }
-        };
+        let simulations =
+            match settlement_simulation::simulate_at_custom_block_and_error_with_tenderly_link(
```
nlordell, line 251: naming nit:
```suggestion
            match settlement_simulation::simulate_and_error_with_tenderly_link(
```

nlordell (approved):
:pray: Looks good!

Approving assuming stray `dbg!(_)` will get cleaned up before merging.

solver/src/driver.rs
vkgnosis: This is exactly why it's _ now. The type has changed from () to U256 for the gas estimation. Next PR will make use of the value.

solver/src/settlement_simulation.rs
```diff
@@ -124,30 +138,41 @@ mod tests {
...
             &contract,
             &web3,
+            0.0,
             network_id.as_str(),
-            Block::FixedWithTenderly(block),
+            block,
+        )
+        .await
+        .unwrap();
+        let _ = dbg!(result);
```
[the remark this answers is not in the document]
nlordell: Oh this is test code :man_facepalming:

PR #1248, opened 2021-10-13
Demote Paraswap Errors to warn! + success/failure metrics

bh2smith:
Based on the amount of noise produced by Paraswap errors in our alerts channel, we deemed it appropriate to demote these errors to warnings and instead record success/failure counts of the single order solver grouped by solver name.

Note that this change required passing `Arc<dyn SolverMetrics>` through the trait method `Solver::solve` which remains unused in many places (apart from the single order solver). However, this PR serves as a good preparation for recording other metrics to come (refined time counters on different stages of solving, etc...).

### Test Plan

~~I recall that @fleupold had done some sort of manual test for metrics at one point and provided the result data in some strange prometheus file format. I could do that if it seems appropriate.~~

1. Run the solver locally pointed at the production orderbook with just the Paraswap solver enabled

2. Connect to the metrics api and find something like this in the logs

```
# HELP gp_v2_solver_single_order_solver Success/Failure counts
# TYPE gp_v2_solver_single_order_solver counter
gp_v2_solver_single_order_solver{result="success",solver_type="ParaSwap"} 4
```


solver/src/driver.rs
```diff
@@ -119,13 +119,15 @@ impl Driver {
...
-                    };
+                let result = match tokio::time::timeout_at(
+                    auction.deadline.into(),
+                    solver.solve(auction, metrics.clone()),
+                )
+                .await
+                {
+                    Ok(inner) => inner,
+                    Err(_timeout) => Err(anyhow!("solver timed out")),
+                };
```
nlordell, line 130: A slightly more succinct version of this would be:
```suggestion
                let result = match tokio::time::timeout_at(
                    auction.deadline.into(),
                    solver.solve(auction, metrics.clone()),
                )
                .await
                .context("solver timed out")?;
```

solver/src/solver/single_order_solver.rs
```diff
@@ -89,15 +97,15 @@ pub struct SettlementError {
     pub inner: anyhow::Error,
     pub retryable: bool,
     // Whether or not this error should be logged as an error
-    pub should_alert: bool,
+    pub track_failure: bool,
```
nlordell, line 100: If we increment errors for all errors, then we can get rid of this field.

nlordell (commented):
Not a fan of passing metrics with the `Solver::solve` method, I would instead recommend passing the metrics in when constructing the solvers.

solver/src/solver.rs
```diff
@@ -47,7 +48,11 @@ pub trait Solver: 'static {
     /// order) so that they can be merged by the driver at its leisure.
     ///
     /// id identifies this instance of solving by the driver in which it invokes all solvers.
-    async fn solve(&self, auction: Auction) -> Result<Vec<Settlement>>;
+    async fn solve(
+        &self,
+        auction: Auction,
+        metrics: Arc<dyn SolverMetrics>,
```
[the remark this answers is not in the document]
vkgnosis: Yes!

solver/src/driver.rs
nlordell: I only noticed because the line got touched :stuck_out_tongue:. Feel free to make a separate PR, as you are right that it is an unrelated change. That being said I'm OK with doing a stealth fix in this PR as well. I'll leave it up to you :smile:.

solver/src/solver.rs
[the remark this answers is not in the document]
nlordell: I also don't think it fits. If we want to instrument solvers with timing metrics, we can always create a wrapping solver implementation for that:
```rust
pub struct InstrumentedSolver {
    inner: Arc<dyn Solver>,
    metrics: Arc<dyn SolverMetrics>,
}

impl Solver for InstrumentedSolver {
    fn solve(...) -> ... {
        let result = self.inner.solve(...);
        self.metrics.do_stuff();
        result
    }
}
```

I feel like otherwise, we would end up having a lot of duplicated code for instrumenting the "common bits" for all the solvers (i.e. recording metrics common to all solvers).

For this PR though, the metrics are very specific to `SingleOrderSolver`s.

nlordell (approved):
Looks good to me! Thanks for fixing this :smile:.

As a follow up task, could you add panels in our Grafana dashboard for this as well?

solver/src/driver.rs
bh2smith: I also get a warning that "result_flattening" is unstable (see issue #70142 <https://github.com/rust-lang/rust/issues/70142> for more information)

vkgnosis (approved):

PR #1250, opened 2021-10-13
Remove inflight trades in favor of waiting for api to see settlement

vkgnosis:
Because the code is simpler. Also gets rid of the driver run loop
interval because it no longer serves any purpose. This should improve
throughput.

### Test Plan
CI and I'm going to observe logs on staging after merging

### DIFF
--- a/e2e/tests/eth_integration.rs
+++ b/e2e/tests/eth_integration.rs
@@ -205,9 +205,10 @@ async fn eth_integration(web3: Web3) {
         }),
     );
     let solver = solver::solver::naive_solver(solver_account);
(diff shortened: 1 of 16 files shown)


nlordell (commented):
Code looks good. Just some questions:

Doesn't this change potentially slow down our ability to solve because we have to wait for the the block that mined the settlement to propagate not only to the solver but also the orderbook API? I guess in practice there isn't much of a difference because we are pointing to the same node.

Another thing to consider is that I thought we wanted to speed up solving by starting another driver run loop while the settlement transaction is mining as a way to increase throughput. How would this change fit in that context?

bh2smith (approved):
Everything looks good codewise, but I would like to get a better impression of why this is expected to improve throughout.

PR #1257, opened 2021-10-13
Fix pending transactions on xdai

vkgnosis:
On xdai the base_fee_per_gas field is missing. This change changes it to
be Optional so that we only error in
recover_gas_price_from_pending_transaction if the transaction is eip1559
AND the field is missing. Currently we error even though the tx is not
eip1559.

### Test Plan
have tested that still works on non xdai node but haven't verified it really fixes xdai although it should


sunce86 (approved):
If base_fee_per_gas is missing, then I would expect that eip1559 is not activated and no 1559 txs exist in the mempool. Strange.
The change itself looks good.

bh2smith (approved):

Issue #1157, opened 2021-09-24
Handle large number of settlements better


vkgnosis:
In today's testing we had a lot of orders in the system which yielded at some point ~400 settlements. Currently the driver loop simulates all of them which can take a long time even when we do it in batches.
A better solution could be to order by objective value first and only simulate the first N (one batch) settlements. Only if all fail we continue to simulate.

Relevant logs are are from 2021-09-24 14:00 UTC staging solver mainnet. We can also try testing our system with a large amount of orders or rinkeby or xdai.

vkgnosis:
Should also look at the gas estimation for settlements because isn't that a simulation internally too? So we might be simulating twice.

nlordell:
> Should also look at the gas estimation for settlements because isn't that a simulation internally too? So we might be simulating twice.

It is. AFAIK `eth_estimateGas` does something similar to an `eth_call` under the hood.

fleupold:
I assume most solution candidate came from the baseline solver. Maybe we should keep only the n most promising solution candidates from each solver?

fleupold:
I think as part of this task we should also create some more metrics around how many runloops we perform per minute (even under pressure, I'd hope our goal can be close to 2 full rounds incl. settlement). During the Kaffeekränzchen we saw some runloops take up to 5 minutes. It might also make sense to record the average time to settlement for orders more accurately ("settling takes too long" is the most prominent empirical feedback we have been getting)

Martin was suggesting, that if we refactor the codebase, maybe we can do it in a way that gets closer to the envisioned decentralized world. In that world, each solver would likely just report a single solution to the "driver" containing their best merged version with already computed objective value. The driver may the request concrete calldata for the best solution (or tell the solver to do execution themselves).

Such a refactoring may also lay the foundation for no longer asking for binding quotes from DexAggregators unless we are sure we will try to execute the quote (cf. #828).

fleupold:
> Should also look at the gas estimation for settlements because isn't that a simulation internally too? So we might be simulating twice.

Yes, and we are not batching those requests. Therefore, in todays session, while simulation was fast (~2s), it then took another 20s to pick the winning solution based on the gas estimates.

![image](https://user-images.githubusercontent.com/1200333/135630258-8fc1129c-12cd-41b8-8835-34f4c414f043.png)

PR #1258, opened 2021-10-14
Reuse gas estimation as simulation in driver

vkgnosis:
Previously:
- Simulate all settlements in batches
- Estimate gas one by one (!)

Now:
- Simulate and estimate gas in one call in batches

This improves the speed of handling many settlements significantly as we
now only make one node call per settlement and all calls are batched.

Part of #1157.

### Test Plan
CI

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -11,9 +11,9 @@ use crate::{
     solver::Solver,
     solver::{Auction, SettlementWithSolver, Solvers},
 };
-use anyhow::{anyhow, Context, Error, Result};
+use anyhow::{anyhow, Context, Result};
 use contracts::GPv2Settlement;
-use ethcontract::errors::MethodError;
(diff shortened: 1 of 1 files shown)


solver/src/driver.rs
```diff
@@ -346,43 +329,36 @@ impl Driver {
...
-                    rated_settlement.objective_value(),
-                    rated_settlement.surplus,
-                    rated_settlement.gas_estimate,
-                    rated_settlement.gas_price,
-                );
-                Some((solver, rated_settlement))
-            })
-            .collect::<Vec<_>>()
-            .await
+        let rate_settlement = |solver, settlement: Settlement, gas_estimate| {
```
nlordell, line 332: Sort of a general note - I quite like these closures that you use to organize logic to keep the `partition_map` closure leaner and easier to follow!

nlordell (approved):
Looks good to me!

PR #1259, opened 2021-10-14
[Trivial] Fix uncatalogued Paraswap Error

bh2smith:
Even though this was a well-known error, we were getting several alerts like [this](https://gnosisinc.slack.com/archives/CUD4MUCUF/p1634221398029700) one that it was uncatalogued. This is because either they changed the text of their error message or we added a word to it during refactoring.

After a bit of digging a bit into [slack alerts](https://gnosisinc.slack.com/archives/CUD4MUCUF/p1634044732096200), and [this PR](https://github.com/gnosis/gp-v2-services/pull/1184/files). It appears this text change was made in the Paraswap API on October 12, 2021.


josojo (approved):
NIce

PR #1261, opened 2021-10-15
[Trivial] Demote OrderQuoteError caused by users to WARN

bh2smith:
Our new Fee & Quote service shouldn't send alerts on what are most commonly "user errors" like seen [here](https://gnosisinc.slack.com/archives/CQZBU9T5J/p1634286286003400). Specifically, this error

```
ERROR orderbook::api::post_quote: post_quote error err=Order(Partial(SameBuyAndSellToken)) request=OrderQuoteRequest
```
which don't warrant any action of any kind. 

We can use grafana to alert us of specific categories of error responses (like server error).

### DIFF
--- a/orderbook/src/api/post_quote.rs
+++ b/orderbook/src/api/post_quote.rs
@@ -354,7 +354,16 @@ pub fn post_quote(
         async move {
             let result = quoter.calculate_quote(&request).await;
             if let Err(err) = &result {
-                tracing::error!(?err, ?request, "post_quote error");
+                match err {
+                    OrderQuoteError::Fee(FeeError::PriceEstimate(PriceEstimationError::Other(err))) => {
+                        tracing::error!(?err, ?request, "post_quote Fee error");
+                    },
+                    OrderQuoteError::Order(ValidationError::Other(err))=> {
+                        tracing::error!(?err, ?request, "post_quote Order Validation error");
+                    },
+                    _ => tracing::warn!(?err, ?request, "post_quote User error")
+                }
+
             }
             Result::<_, Infallible>::Ok(response(result))
         }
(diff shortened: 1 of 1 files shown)


orderbook/src/api/post_quote.rs
```diff
@@ -354,7 +354,17 @@ pub fn post_quote(
...
+                    OrderQuoteError::Fee(FeeError::PriceEstimate(PriceEstimationError::Other(
+                        err,
+                    ))) => {
+                        tracing::error!(?err, ?request, "post_quote Fee error");
+                    }
+                    OrderQuoteError::Order(ValidationError::Other(err)) => {
+                        tracing::error!(?err, ?request, "post_quote Order Validation error");
+                    }
+                    _ => tracing::warn!(?err, ?request, "post_quote User error"),
+                }
```
nlordell, line 367: Why not:
```suggestion
                tracing::warn!(?err, ?request, "post_quote error");
```

Are there any errors we want to alert on? From what I understand we can already API metrics to determine if we are failing to often. Specifically it should be: `gp_v2_api_api_requests_complete{exported_endpoint="post_quote",status_code="500"}` (pending https://github.com/gnosis/gp-v2-services/pull/1262/files).

bh2smith: I did this because all the "Other" errors are our own internal server errors, for which I would assume we want to be alerted about.

bh2smith: Changed it in favour of grafana.

bh2smith: Any reason remaining that has prevented you from approving?

nlordell: Just wanted an answer to my question before approving :stuck_out_tongue:.

nlordell (approved):

PR #1262, opened 2021-10-15
Fix the `POST v1/quote` metric label

nlordell:
This PR just fixes the label for recording API metrics for `POST v1/quote`.


bh2smith (approved):
That was an unfortunate mistake. Nice find.

PR #1263, opened 2021-10-15
Apply fee factor to stored computed fees and not just new ones

nlordell:
This PR applies changes the `MinFeeCalculator` logic so that it applies factors to past fees as well as new fees.

With the change of storing full fee amounts for orders, we needed to start storing the unsubsidized fee amount into the Postgres database. There was an early exit when the fee was stored in the database that didn't apply the fee factor.

Additionally, I added some more logging so that we get more details about how fees are being computed making it easier to diagnose failures in the future.

One **sneaky** change I through in was to increase the `ORDER_GAS_AMOUNT` to 200k. This was done because previously, we would not subsidize min-fees that were computed with this "legacy" method. Now, with the refactor, we subsidize this amount as well, so it was increased to offset that subsidy by a bit. I can revert this change if people are uncomfortable with it.

### Test Plan

Added a new unit test to make sure fee factor is applied in both places.

### DIFF
--- a/orderbook/src/fee.rs
+++ b/orderbook/src/fee.rs
@@ -91,7 +91,7 @@ pub trait MinFeeStoring: Send + Sync {
     ) -> Result<Option<U256>>;
 }
 
-const GAS_PER_ORDER: f64 = 200_000.0;
+const GAS_PER_ORDER: f64 = 300_000.0;
 
 // We use a longer validity internally for persistence to avoid writing a value to storage on every request
 // This way we can serve a previous estimate if the same token is queried again shortly after
@@ -211,8 +211,7 @@ impl MinFeeCalculator {
 
         let gas_amount =
             if let (Some(buy_token), Some(amount), Some(kind)) = (buy_token, amount, kind) {
-                let gas_estimate = self
-                    .price_estimator
+                self.price_estimator
                     .estimate(&price_estimation::Query {
                         sell_token,
                         buy_token,
@@ -221,32 +220,8 @@ impl MinFeeCalculator {
                     })
                     .await?
                     .gas
-                    .to_f64_lossy();
-
-                match kind {
-                    OrderKind::Sell => {
-                        tracing::debug!(
-                            "estimated {} gas for order selling {} {:?} for {:?}",
-                            gas_estimate,
-                            amount,
-                            sell_token,
-                            buy_token
-                        );
-                    }
-                    OrderKind::Buy => {
-                        tracing::debug!(
-                            "estimated {} gas for order buying {} {:?} for {:?}",
-                            gas_estimate,
-                            amount,
-                            buy_token,
-                            sell_token
-                        );
-                    }
-                }
-
-                gas_estimate
+                    .to_f64_lossy()
             } else {
-                tracing::debug!("using default {} gas per order", GAS_PER_ORDER);
                 GAS_PER_ORDER
             };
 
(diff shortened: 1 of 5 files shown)


orderbook/src/fee.rs
```diff
@@ -88,7 +91,7 @@ pub trait MinFeeStoring: Send + Sync {
     ) -> Result<Option<U256>>;
 }
 
-const GAS_PER_ORDER: f64 = 100_000.0;
+const GAS_PER_ORDER: f64 = 200_000.0;
```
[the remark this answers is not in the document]
fleupold: I think our average price per trade is more around 300k so maybe we should even use that?

orderbook/src/fee.rs
```diff
@@ -213,10 +221,35 @@ impl MinFeeCalculator {
                     })
                     .await?
                     .gas
-                    .to_f64_lossy()
+                    .to_f64_lossy();
+
+                match kind {
```
fleupold, line 226: nit: rather than logging here (I think this information could be conveyed together in the unconditional log below), I think it would be nice if we could log where the estimate comes from (e.g. if from Fallback, Baseline, Paraswap or DB)

orderbook/src/fee.rs
```diff
@@ -260,33 +299,37 @@ impl MinFeeCalculating for MinFeeCalculator {
...
+            .read_fee_measurement(sell_token, buy_token, amount, kind, official_valid_until)
             .await
         {
-            return Ok((past_fee, official_valid_until));
-        }
-
-        let min_fee = self
-            .compute_min_fee(sell_token, buy_token, amount, kind)
-            .await?;
+            past_fee
```
fleupold, line 307: ie. add a log here

fleupold (approved):
Very nice catch!

 <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia4.giphy.com%2Fmedia%2FxT8qB61onG5u3cNlo4%2Fgiphy.gif&token=catch" data-gitmeme-token="catch"></a>

orderbook/src/fee.rs
nlordell: > we could log where the estimate comes from (e.g. if from Fallback, Baseline, Paraswap or DB)

So, in order to distinguish between the Baseline and ParaSwap solver, that would require changes to the priority price estimator. For now, you will see warnings with:
```
inner price estimator failed
```

If the ParaSwap price estimator failed during the fee estimate, meaning that the Baseline price estimator was used instead.

This is not ideal, but I will implement disambiguating between the different price estimate sources in a separate PR as it would touch code unrelated to this change (https://github.com/gnosis/gp-v2-services/issues/1265).

orderbook/src/fee.rs
```diff
@@ -226,24 +234,30 @@ impl MinFeeCalculator {
         };
         let estimate = self.price_estimator.estimate(&query).await?;
         let price = estimate.price_in_sell_token_f64(&query);
+
+        tracing::debug!(
+            "computed unsubsidized fee amount of {} ETH at a price of {} ETH/{:?}",
+            fee_in_eth,
+            price,
+            sell_token,
+        );
```
bh2smith, line 243: Not sure which is the right way to go here, but this makes sense to me reading as "ETH per SELL_TOKEN". I think Rafa might say its the other way around because it makes sense to me and I am always wrong about this. What do we expect this to say if `sell_token = DAI` or `USDC`?

bh2smith (commented):
Love it. Very good find. Would I would like to know if this issue was only introduced recently with the introduction of the new `full_fee_factor` DB column r if this has always been the case.

PR #1264, opened 2021-10-15
Add metric to track gas price estimate

nlordell:
This PR adds an instrumented gas price estimator that records sampled gas prices as a metric.

In terms of Prometheus things... I used a `Gauge`. My understanding is that this allows us to set a single value that will get sample over time, allowing us to plot a nice graph of 
```
          ^         
          |  --    __
gas_price |/    \/
          |      
          +---------->
              time
```

### Test Plan

Start the orderbook and check metrics at:
```
$ curl -s http://localhost:9586/metrics | grep gas_price
# HELP gp_v2_api_gas_price Gas price estimate over time.
# TYPE gp_v2_api_gas_price gauge
gp_v2_api_gas_price 140.321906696
```

### DIFF
--- a/orderbook/src/gas_price.rs
+++ b/orderbook/src/gas_price.rs
@@ -1,0 +1,51 @@
+//! Module defining an instrumented Ethereum gas price estimator.
+//!
+//! This allows us to keep track of historic gas prices in Grafana and do things
+//! like alert when gas prices get too high as well as detect spikes and other
+//! anomalies.
+
+use anyhow::Result;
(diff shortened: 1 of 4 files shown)


bh2smith (approved):
Sexy! Awesome idea too.

vkgnosis (approved):

nlordell:
I will just have metrics for the API endpoint for now (as its more relevant for debugging user fees). I'll think about how to do it in the solver better (although your idea of having a separate task to just log fees seems good to me).

PR #1267, opened 2021-10-17
Ignore Liquidity/PMM orders in Objective evaluation

### LINKED ISSUE #1246: Ignore Liquidity Orders When computing objective criterion
Otherwise, with the fee scaling, we might prefer including them even (as they contribute to the objective value) if the user could have gotten a slightly better price on an external venue.

bh2smith:
Fixes #1246

This PR does a few things

- Revert useless metrics introduced in (#1243)
- Updates total surplus computation to exclude liquidity orders.
- We also exclude PMM orders from consideration when fetching AMM liquidity sources as well.

Note that this implementation will no longer require any changes to the HTTP solver.

### Test Plan
Introduced new unit test with a settlement containing two matched orders, we evaluate total surplus in all scenarios when one or, both or none of the orders is a liquidity order and make set-theoretic assertions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -29,9 +29,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "ahash"
-version = "0.7.4"
+version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43bb833f0bf979d8475d38fbf09ed3b8a55e1885fe93ad3f93239fc6a4f17b98"
+checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
 dependencies = [
  "getrandom",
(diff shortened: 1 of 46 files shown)


nlordell (commented):
One thing to consider, which might fit nicer into our current abstractions, would be to only have user orders in:
https://github.com/gnosis/gp-v2-services/blob/59b3f0d0eb5b3cb3a24e87d44ea69c84f7292efe/solver/src/solver.rs#L71-L72

Then, liquidity orders could (instead of having an `is_liquidity_order` flag) be added as a `Liquidity` enum variant:
https://github.com/gnosis/gp-v2-services/blob/59b3f0d0eb5b3cb3a24e87d44ea69c84f7292efe/solver/src/liquidity.rs#L28-L32

This has the added benefit that our single order solvers and baseline solver don't even try to solve PMM orders.

nlordell (commented):
Looks good to me.

Can you please update the description, I think it may be out of date.

nlordell (commented):
Looks good minus the outdated comments.

nlordell (approved):

PR #1274, opened 2021-10-20
Limit the number of settlements considered per solver

vkgnosis:
Part of https://github.com/gnosis/gp-v2-services/issues/1157 .

This is a simple low hanging fruit fix to address the problem of a solver returning so many settlements that we cannot simulate them in a reasonable time. While the issue thread outlines some smarter ideas like estimating objective value and dropping, the solution in this PR is much simpler and still effective. We can still consider the more complex solutions in the future but they might be come invalidated when refactoring to make solvers more independent.

### Test Plan
CI


nlordell (approved):
Seems OK to me, since this would just limit our single order solver settlements.

PR #1276, opened 2021-10-20
Move simulation at previous block into background task

vkgnosis:
so that the run loop isn't blocked.

### Test Plan
CI


nlordell (approved):
Nice!

PR #1278, opened 2021-10-21
Solver Timeout Metric

bh2smith:
Rather than alert on every solver timeout like this

```
[DFUSION][GP-V2-SERVICES] Error
ERROR (pod: dfusion-v2-solver-xdai-74c4fd6c66-tdsnz):
2021-10-20T18:35:57.680Z ERROR solver::driver: solver Quasimodo error: solver timed out
```

We keep a counter of the number of each solver timeout to be displayed as a chart in grafana.

### DIFF
--- a/solver/src/driver.rs
+++ b/solver/src/driver.rs
@@ -423,7 +423,8 @@ impl Driver {
             let mut settlements = match settlements {
                 Ok(settlement) => settlement,
                 Err(err) => {
-                    tracing::error!("solver {} error: {:?}", name, err);
+                    self.metrics.solver_time_out(name);
+                    tracing::warn!("solver {} error: {:?}", name, err);
                     continue;
                 }
             };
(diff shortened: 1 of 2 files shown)


solver/src/driver.rs
```diff
@@ -423,7 +423,8 @@ impl Driver {
             let mut settlements = match settlements {
                 Ok(settlement) => settlement,
                 Err(err) => {
-                    tracing::error!("solver {} error: {:?}", name, err);
+                    self.metrics.solver_time_out(name);
```
nlordell, line 426: Note that this is not just a timeout, it can also be an actual error. For example HTTP errors in the HTTP solver also hit this path.

nlordell (commented):
Note that I also have a PR: https://github.com/gnosis/gp-v2-services/pull/1279

PR #1281, opened 2021-10-21
Differentiate between different solver outcomes

### LINKED ISSUE #1280: Differentiate between more solver run outcomes
Right now we differentiate only success and failure for each solver run. Some other categories that we could differentiate with would be:
- success
- empty (i.e. no settlements).
- failure
- timeout

nlordell:
Fixes #1280 

This PR differentiates between different solver run outcomes:
- Empty settlements
- Success
- Timeout
- Failure

I'm not entirely sure how valuable this is compared to just success/failure, and wanted to know what you all think.

### Test Plan

Observe new metrics.

<details><summary>To try out all possibilities patch the solver:</summary>

```diff
diff --git a/crates/solver/src/solver/baseline_solver.rs b/crates/solver/src/solver/baseline_solver.rs
index b1ea23f..b05f047 100644
--- a/crates/solver/src/solver/baseline_solver.rs
+++ b/crates/solver/src/solver/baseline_solver.rs
@@ -29,6 +29,7 @@ impl Solver for BaselineSolver {
             orders, liquidity, ..
         }: Auction,
     ) -> Result<Vec<Settlement>> {
+        anyhow::bail!("test error"); // Baseline always fails
         Ok(self.solve_(orders, liquidity))
     }
 
diff --git a/crates/solver/src/solver/naive_solver.rs b/crates/solver/src/solver/naive_solver.rs
index c683219..a7fa60a 100644
--- a/crates/solver/src/solver/naive_solver.rs
+++ b/crates/solver/src/solver/naive_solver.rs
@@ -28,6 +28,7 @@ impl Solver for NaiveSolver {
             orders, liquidity, ..
         }: Auction,
     ) -> Result<Vec<Settlement>> {
+        futures::future::pending::<()>().await; // Naive always times out
         let uniswaps = extract_deepest_amm_liquidity(&liquidity);
         Ok(settle(orders, uniswaps).await)
     }
diff --git a/crates/solver/src/solver/paraswap_solver.rs b/crates/solver/src/solver/paraswap_solver.rs
index 87dd59f..3d21c46 100644
--- a/crates/solver/src/solver/paraswap_solver.rs
+++ b/crates/solver/src/solver/paraswap_solver.rs
@@ -84,6 +84,7 @@ impl SingleOrderSolving for ParaswapSolver {
         &self,
         order: LimitOrder,
     ) -> Result<Option<Settlement>, SettlementError> {
+        return Ok(None); // ParaSwap always returns empty settlements
         let token_info = self
             .token_info
             .get_token_infos(&[order.sell_token, order.buy_token])
```

</details>

Then run it:
```
$ cargo run -p solver -- --solvers Baseline,Naive,ParaSwap,ZeroEx
```

And see that the metrics increase:
```
# TYPE gp_v2_solver_solver_run counter
gp_v2_solver_solver_run{result="empty",solver_type="ParaSwap"} 1
gp_v2_solver_solver_run{result="failure",solver_type="BaselineSolver"} 1
gp_v2_solver_solver_run{result="success",solver_type="0x"} 1
gp_v2_solver_solver_run{result="timeout",solver_type="NaiveSolver"} 1
```

### DIFF
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -6,7 +6,7 @@ use crate::{
     in_flight_orders::InFlightOrders,
     liquidity::order_converter::OrderConverter,
     liquidity_collector::LiquidityCollector,
-    metrics::SolverMetrics,
+    metrics::{SolverMetrics, SolverRunOutcome},
     orderbook::OrderBookApi,
     settlement::Settlement,
     settlement_simulation,
@@ -14,7 +14,7 @@ use crate::{
     solver::Solver,
     solver::{Auction, SettlementWithSolver, Solvers},
 };
-use anyhow::{anyhow, Context, Result};
+use anyhow::{Context, Result};
 use contracts::GPv2Settlement;
 use ethcontract::errors::{ExecutionError, MethodError};
 use futures::future::join_all;
@@ -126,7 +126,7 @@ impl Driver {
     async fn run_solvers(
         &self,
         auction: Auction,
-    ) -> Vec<(Arc<dyn Solver>, Result<Vec<Settlement>>)> {
+    ) -> Vec<(Arc<dyn Solver>, Result<Vec<Settlement>, SolverRunError>)> {
         join_all(self.solvers.iter().map(|solver| {
             let auction = auction.clone();
             let metrics = &self.metrics;
@@ -136,8 +136,8 @@ impl Driver {
                     match tokio::time::timeout_at(auction.deadline.into(), solver.solve(auction))
                         .await
                     {
-                        Ok(inner) => inner,
-                        Err(_timeout) => Err(anyhow!("solver timed out")),
+                        Ok(inner) => inner.map_err(SolverRunError::Solving),
+                        Err(_timeout) => Err(SolverRunError::Timeout),
                     };
                 metrics.settlement_computed(solver.name(), start_time);
                 (solver.clone(), result)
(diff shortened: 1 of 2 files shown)


vkgnosis (approved):
I can see it being useful in the future and it is natural data to collect.

solver/src/driver.rs
```diff
@@ -410,19 +410,26 @@ impl Driver {
...
+                    }
+
                     self.metrics.solver_run_succeeded(name);
                     settlement
                 }
                 Err(err) => {
-                    self.metrics.solver_run_failed(name);
+                    match err {
+                        SolverRunError::Timeout => self.metrics.solver_run_timeout(name),
+                        SolverRunError::Solving(_) => self.metrics.solver_run_failed(name),
```
bh2smith, line 426: I guess there are different methods in the metrics so we don't have to give them the `SolverError` type. If this is actually an issue, we could cast the error type to string and pass it into a single metrics method.

bh2smith: This was the inner anyhow Error could at least be separated without a match.

bh2smith (approved):
Looks pretty good, but I think we could refine the SolverRunError::Solving even more by casting to string. Its also fine to count these all in a single bucket for now, but I think we could increment the counter with a single methods defined on SolverMetrics.

Here is an example of such proposal:

https://github.com/gnosis/gp-v2-services/blob/ec27fb7d5a3dc36458d3587bddccfa38aee4219d/solver/src/metrics.rs#L266-L271

In this case we could pass in solver error as an additional parameter and include it as the "result" field.

solver/src/driver.rs
vkgnosis: Labels on prometheus metrics should be used sparingly. We shouldn't use the error string as a label.

bh2smith: ok then.

bh2smith:
Wen merge?

nlordell:
> Wen merge?

Still need to fix your suggestion: https://github.com/gnosis/gp-v2-services/pull/1281#pullrequestreview-785676098

marcovc (approved):
Great! Maybe call trivial instead of empty? No strong opinion though.

nlordell:
> call trivial instead of empty?

I called it `empty` because solvers can return `[]` solutions (so "empty" set of solutions) which is different than (`[Solution::trivial()]`). They end up both incrementing the same metric (since we filter out empty settlements), but I kept the name "empty" to be closer to what we are measuring.

I think only MIP and Quasimodo solvers return trivial solutions, others return empty solutions AFAIK.

marcovc:
>I think only MIP and Quasimodo solvers return trivial solutions, others return empty solutions AFAIK.

I see, yes then all good.

Issue #1284, opened 2021-10-22
In flight trades being solved for.


nlordell:
It appears that inflight trades are being considered by solvers but shouldn't be. For example: https://logs.gnosisdev.com/app/kibana#/discover/doc/de1f5360-a9ae-11ea-acbb-9d2e3cb124ca/logstash-2021.10.22?id=oNX6pXwBIBf7G29guMB_

PR #1285, opened 2021-10-22
Return transaction receipt instead of just hash

vkgnosis:
This gives us access to the block number which will be used in the fix for https://github.com/gnosis/gp-v2-services/issues/1284 .

### Test Plan
CI


nlordell (approved):

sunce86 (approved):
Lg

PR #1288, opened 2021-10-22
Make in flight trades keep track of block number

### LINKED ISSUE #1284: In flight trades being solved for.

vkgnosis:
and refactor into standalone component.

We keep track of the block number so that in flight trades stay alive until the api sees the block of the new settlement which we verify using the v2 solvable orders route. Previously we would always delete in flight trades in the next run loop which was no longer enough with the recent run loop speed ups.

Fixes https://github.com/gnosis/gp-v2-services/issues/1284 .
Supersedes https://github.com/gnosis/gp-v2-services/pull/1250 .

### Test Plan
CI and adapted unit test for in flight trades.

### DIFF
--- a/e2e/tests/eth_integration.rs
+++ b/e2e/tests/eth_integration.rs
@@ -3,7 +3,7 @@ mod ganache;
 mod services;
 
 use crate::services::{
-    create_orderbook_api, create_orderbook_liquidity, deploy_mintable_token, to_wei, GPv2,
+    create_order_converter, create_orderbook_api, deploy_mintable_token, to_wei, GPv2,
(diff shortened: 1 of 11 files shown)


nlordell (approved):
Looks good! From what I can tell, this strategy won't work if we start solving the next batch while waiting for a transactions to confirm, but we should really just cross that bridge when we get there.

PR #1292, opened 2021-10-25
Generic convert_response for API reponse handling

bh2smith:
Based on the suggestion [here](https://github.com/gnosis/gp-v2-services/pull/1176#discussion_r721408663)

We write a generic method `convert_response` which is used in place of all the `*_response(result: Result<XXX>)` methods where appropriate.

This method expects the `Err( . )` route to implement `WarpReplyConverting`.  Additional work will have to go into generically handling the `Ok( . )` route - which is still lingering around for some reply conversions and will be addressed in a follow up PR.

### Test Plan
No logical changes - existing CI.

### DIFF
--- a/orderbook/src/api.rs
+++ b/orderbook/src/api.rs
@@ -126,10 +126,10 @@ fn internal_error() -> Json {
     })
 }
 
-pub fn convert_response<T, E>(result: Result<T, E>) -> WithStatus<Json>
+pub fn convert_json_response<T, E>(result: Result<T, E>) -> WithStatus<Json>
 where
     T: Serialize,
-    E: WarpReplyConverting,
+    E: IntoWarpReply,
 {
     match result {
         Ok(response) => with_status(warp::reply::json(&response), StatusCode::OK),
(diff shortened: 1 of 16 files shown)


vkgnosis (commented):
> which is used in place of all the

Is that the plan as what is supposed be in this PR already? Because we're only changing one method here.

vkgnosis (approved):
Seems fine to me. Although I would

nlordell (commented):
I don't think the PR implements my suggestion linked in the description.

Specifically:
- In general, for `Result<T, E>`, we `T: Serialize` because we convert it to JSON, so I don't really understand why we need the trait bound `T: WarpReplyConverting` - instead I suggest we should use `warp::reply::json(ok_value)`.
- The suggestion was about using the `E: WarpReplyConverting` for converting errors into warp replies. For this, we already have a few error types that implement this (`PartialValidationError`, `ValidationError`, `FeeError`). Furthermore, you could easily `impl WarpReplyConverting for anyhow::Error` where it returns a 500 error.

In the end, if we had any result where the `Ok` type implements `Serialize` (i.e. can be converted into JSON), then the `convert_response` would just convert into JSON for us, and then the `Error` type would just need to implement `WarpReplyConverting` and the `convert_response` could be used everywhere.

In general, my suggestion was aiming at deduplicate methods like:
```rust
pub fn response<T: Serialize>(result: Result<T, OrderQuoteError>) -> impl Reply {
    match result {
        Ok(response) => reply::with_status(reply::json(&response), StatusCode::OK),
        Err(err) => {
            let (reply, status) = err.convert_to_reply();
            reply::with_status(reply, status)
        }
    }
}
```

Where instead of using `OrderQuoteError`, you make it generic on the error type `E where E: WarpReplyConverting`. 

**edit**: Some other methods that could be de-deduplicated by making this generic on the error type:
- `orderbook::api::get_user_orders::response`
- `orderbook::api::get_trades::get_trades_response`
- `orderbook::api::get_solvable_orders::get_solvable_orders_response`
- `orderbook::api::get_solvable_orders_v2::get_solvable_orders_response`
- `orderbook::api::get_orders::get_orders_response`
- `orderbook::api::get_orders_by_tx::get_orders_by_tx_response`
- Maybe more.

vkgnosis (changes requested):

nlordell (approved):

vkgnosis (approved):

PR #1302, opened 2021-10-26
Update to `tracing-subscriber v0.3.1`

nlordell:
Supersedes #1293 

This PR updates the `tracing-subscriber` dependency to `v0.3.1`. Its a little more involved so I made a separate PR.

In particualr, time formatting has changed in the newer version to use the `time` crate, so the format string is specified differently. There is a built-in RFC 3339 which is almost what we need, except it uses a variable number of subsecond digits (based on trailing `0`s), so a custom format was needed.

### Test Plan

CI, no logic changes.

You can test that the formatting works by:
```
$ cargo run -p solver
2021-10-26T18:08:38.677Z  INFO solver: running solver with validated Arguments {
...
```

And then compare it to `main` to make sure its the same:
```
$ cargo run -p solver
2021-10-26T18:09:49.659Z  INFO solver: running solver with validated Arguments {
...
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1449,9 +1449,9 @@ checksum = "3e2e65a1a2e43cfcb47a895c4c8b10d1f4a61097f9f254f183aee60cad9c651d"
 
 [[package]]
 name = "matchers"
-version = "0.0.1"
+version = "0.1.0"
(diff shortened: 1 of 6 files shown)


bh2smith (approved):
Nice job here.

vkgnosis (approved):

PR #1306, opened 2021-10-27
Exit run loop early if there is no user order

vkgnosis:
Speeds up run loop because we can skip a bunch of needless operations
when we already know the settlement can't be useful.

### Test Plan
Did not add unit test because function is simple enough. e2e tests should still work.


nlordell (approved):
LGTM, would just like to briefly discuss an alternative position of the early exit (but would be OK with how it is now).

PR #1308, opened 2021-10-27
Fix price estimation metric counting all errors as unsuccessful

vkgnosis:
The PriceEstimationError enum has several cases that should not be
counted as errors for the purposes of this m,etric. I decided to only
count `Other` as error as this means we were not able to get a
successful response from the estimator. The other errors should all be
presented as errors to api users but are not errors in terms of the
price estimator.

### Test Plan
CI, change so small does not need unit test

### DIFF
--- a/shared/src/price_estimation/instrumented.rs
+++ b/shared/src/price_estimation/instrumented.rs
@@ -35,7 +35,8 @@ where
     ) -> Vec<anyhow::Result<Estimate, PriceEstimationError>> {
         let results = self.inner.estimates(queries).await;
         for result in &results {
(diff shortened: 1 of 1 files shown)


bh2smith (approved):
This is great, so now we only record metrics for errors that are "internal server errors" and not "user input error".

nlordell (approved):
Ah nice catch! Yes, we probably don't want to alert because users keep on trying to trade tokens without liquidity. Do you think we should have a separate "not found" result so we can track how successful price estimators are? The issue being that it will be heavily biased for the estimators lower in the priority queue since they already get hard to find prices.

PR #1311, opened 2021-10-27
Drop old transport metrics

taminomara:
Part of metrics refactoring. Removing instrumented transport because HTTP transport now exports its own metrics.

Note that I'm working in a separate branch, so these changes won't appear in `main` for a while.


nlordell (commented):
PR looks good in it of itself, but it looks like it merges into `main`.

vkgnosis:
Maybe Tamika means more PRs merging into this PR branch will follow. Either way I'm not sure if it should be reviewed right now. If not it would be nice to mark as draft so that's clear and doesn't auto request review from the team.

PR #1312, opened 2021-10-27
Drop old database metrics

taminomara:
Part of metrics refactoring. Removing instrumented database because it exports its own metrics now.


bh2smith (approved):

vkgnosis:
> because it exports its own metrics now.

Where does this happen? I don't remember this.

bh2smith:
These seems to be going stale....

taminomara:
I plan to revive it during oncall.

PR #1314, opened 2021-10-28
Log more fee information

vkgnosis:
### Test Plan
Manually ran and looked at log.

2021-10-28T11:28:03.381Z DEBUG orderbook::fee: unsubsidized fee amount sell_token=0xdac17f958d2ee523a2206206994597c13d831ec7 buy_token=Some(0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2) amount=Some(1000000) kind=Some(Sell) gas_price=140800745777 gas_amount=111000 fee_in_eth=15628882781247000 price=0.000000004126744675 fee=64496408.79371024

### DIFF
--- a/orderbook/src/fee.rs
+++ b/orderbook/src/fee.rs
@@ -234,15 +234,14 @@ impl MinFeeCalculator {
         };
         let estimate = self.price_estimator.estimate(&query).await?;
         let price = estimate.price_in_sell_token_f64(&query);
+        let fee = fee_in_eth * price;
 
         tracing::debug!(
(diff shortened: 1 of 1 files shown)


bh2smith (approved):
Personally I find the `Some(...)` kindof distracting from the log, but can't think of any appropriate way to avoid it (given that these are indeed optional fields). Just talking to myself - all is good and thanks for the more verbose logging!

fleupold (approved):
Looks good. The other thing that might be interesting is a log of which price estimation source was used.

PR #1315, opened 2021-10-28
Do not settle liquidity orders in single order solver and baseline

vkgnosis:
These solvers settle orders one by one without making use of liquidity
orders. Without this check they would treat them as user orders even
though we do not want these orders to be settled on their own.

Part of #1307 

### Test Plan
CI, existing tests make sure non liquidity orders are still being considered

### DIFF
--- a/solver/src/solver/baseline_solver.rs
+++ b/solver/src/solver/baseline_solver.rs
@@ -130,7 +130,12 @@ impl BaselineSolver {
         }
     }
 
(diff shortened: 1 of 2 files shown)


bh2smith (approved):
Looks good, just wondering if making user orders mutable is the right way to go? If it is, then I might suggest a name change to the input parameter since I consider "user_orders" to be those that already don't include liquidity. Maybe passing the filtered list is also an option.

fleupold (approved):
Good enough quick fix to avoid these unnecessary settlements (I still think #1305 can eventually make this more structure and avoid us checking this flag in a lot of places)

PR #1318, opened 2021-10-28
Make some logs less verbose

vkgnosis:
by converting them from info to debug

My motivation is to to make looking at the Kibana logs easier when excluding logs at lower level than info (debug, trace). I'd like all info logs to not be too spammy and be useful to tell from a high level what the application is doing at the moment. Then we something needs to be investigated in more detail debug logs can be included in the filter.

### Test Plan
CI


bh2smith (approved):
I like it! Although some of these weren't overly bothersome like the Balancer ones which only occurred once on startup. But I agree, they don't provide much in the way of "info" and were originally used for the purpose of debugging.

nlordell (approved):
Nice cleanup.

PR #1328, opened 2021-10-31
Format objective value related numbers in scientific notation

vkgnosis:
By default big rationals print their full numerator and denominator
making it hard to see how large they are. By printing the numbers in
exponent notation they become easier to read.

I changed the gas estimate too because even though it is a natural number it still becomes easier to read in exponent notation.

### Test Plan
CI


bh2smith (approved):
Love this change!

PR #1340, opened 2021-11-02
Allow using private 0x api

vkgnosis:
This helps us not run into rate limits and eventually use their RFQT liquidity.

I added the values we want to use to 1password.

### Test Plan
new ignored test that I ran

### DIFF
--- a/shared/src/zeroex_api.rs
+++ b/shared/src/zeroex_api.rs
@@ -6,7 +6,7 @@
 
 use crate::debug_bytes;
 use crate::solver_utils::{deserialize_decimal_f64, Slippage};
-use anyhow::Result;
+use anyhow::{Context, Result};
 use derivative::Derivative;
 use ethcontract::{H160, U256};
 use model::u256_decimal;
@@ -23,7 +23,7 @@ const AFFILIATE_ADDRESS: &str = "0x9008D19f58AAbD9eD0D60971565AA8510560ab41";
 ///
 /// These parameters are currently incomplete, and missing parameters can be
 /// added incrementally as needed.
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Copy, Debug, Default)]
 pub struct SwapQuery {
     /// Contract address of a token to sell.
     pub sell_token: H160,
@@ -35,8 +35,6 @@ pub struct SwapQuery {
     pub buy_amount: Option<U256>,
     /// Limit of price slippage you are willing to accept.
     pub slippage_percentage: Slippage,
-    /// Flag to disable checks of the required quantities.
-    pub skip_validation: Option<bool>,
 }
 
 impl SwapQuery {
@@ -58,8 +56,6 @@ impl SwapQuery {
             .append_pair("sellToken", &addr2str(self.sell_token))
             .append_pair("buyToken", &addr2str(self.buy_token))
             .append_pair("slippagePercentage", &self.slippage_percentage.to_string());
-        // I am not setting any affiliate address, in order to save gas
-        // .append_pair("affiliateAddress", "0xgp_address")
         if let Some(amount) = self.sell_amount {
             url.query_pairs_mut()
                 .append_pair("sellAmount", &amount.to_string());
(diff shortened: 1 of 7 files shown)


shared/src/zeroex_api.rs
```diff
@@ -35,8 +35,6 @@ pub struct SwapQuery {
     pub buy_amount: Option<U256>,
     /// Limit of price slippage you are willing to accept.
     pub slippage_percentage: Slippage,
-    /// Flag to disable checks of the required quantities.
-    pub skip_validation: Option<bool>,
```
vkgnosis, line 39: Removed from the struct because we always set it to false and it doesn't make sense without also having takerAddress. When we want to support takerAddress we can add it back.

shared/src/zeroex_api.rs
```diff
@@ -58,8 +56,6 @@ impl SwapQuery {
             .append_pair("sellToken", &addr2str(self.sell_token))
             .append_pair("buyToken", &addr2str(self.buy_token))
             .append_pair("slippagePercentage", &self.slippage_percentage.to_string());
-        // I am not setting any affiliate address, in order to save gas
-        // .append_pair("affiliateAddress", "0xgp_address")
```
vkgnosis, line 62: Was an outdated comment.

shared/src/zeroex_api.rs
```diff
@@ -224,69 +231,40 @@ mod tests {
...
-            skip_validation: Some(true),
         };
 
         let price_response = zeroex_client.get_swap(swap_query).await;
+        dbg!(&price_response);
         assert!(price_response.is_ok());
     }
 
-    #[test]
-    fn swap_query_serialization_0x_sell_order() {
```
vkgnosis, line 244: The github diff around here isn't great. I removed the serialization tests because they would have to be kept up to date with any change to `format_url` above. That function is simple so I felt the tests didn't provide value.

shared/src/zeroex_api.rs
nlordell: As a side not, from my quick testing, I'm not sure having `skipValidation=false` and specifying  `takerAddress` as the trader for price estimates will be possible because it requires allowance to be set by the trader on the 0x contract, so it wouldn't work for traders that have never traded on 0x for example.

I agree with the change here.

shared/src/zeroex_api.rs
```diff
@@ -68,12 +64,13 @@ impl SwapQuery {
...
-            url.query_pairs_mut()
-                .append_pair("skipValidation", &skip_validation.to_string());
-        }
         url.query_pairs_mut()
             .append_pair("affiliateAddress", AFFILIATE_ADDRESS);
+        // We do not provide a takerAddress so validation does not make sense.
+        url.query_pairs_mut().append_pair("skipValidation", "true");
+        // Ensure that we do not request binding quotes that we might be penalized for not taking.
+        url.query_pairs_mut()
+            .append_pair("intentOnFilling", "false");
```
nlordell, line 73: This is true for price estimates, but not for the single order solver right?

nlordell (commented):
Looks good to me. Just one question around the `intentOnFilling=false`. From what I understand, this should be `true` for `quote` calls that we use for the single order solver right?

shared/src/zeroex_api.rs
vkgnosis: It should not because for the single order solver we do not guarantee that the quote will actually get executed. After https://github.com/gnosis/gp-v2-services/issues/828 what we could do is use price endpoint in the single order solver with RFQT liquidity enabled and only use quote if its settlement wins.

vkgnosis: Technically `intentOnFilling` does nothing for the price endpoint anyway but I felt it was safer to unconditionally include it for now.

nlordell: I see, its the same issue we have with ParaSwap I guess. I guess we don't have access to their RFQT liquidity then?

vkgnosis: We shouldn't have access to their RFQT liquidity in the price call because otherwise we might quote a better price and gas amount than what we will get later through quote (which is definitely not using RFQT since intentOnFilling=false). Their api docs aren't completely clear whether /prices includes this liquidity or not which is another reason for always having intentOnFilling=false.

vkgnosis: Relevant docs https://0x.org/docs/api#get-swapv1price and https://0x.org/docs/guides/rfqt-in-the-0x-api .

Actually reading through this again I think we shouldn't even use the price endpoint at the moment because they say it is meant for RFQT.

vkgnosis: I'm going to make another PR to tackle this in the price estimator instead of including it in this PR.

nlordell: Also, it looks like we never qualified for the RFQT liquidity:
> In order to qualify for RFQ-T liquidity, the request must include the query parameter intentOnFilling=true (in addition to the aforementioned 0x-api-key and non-null takerAddress).

So I don't have any concerns with keeping `intentOnFilling=false` until we implement #828

nlordell (approved):

PR #1341, opened 2021-11-02
Use 0x quote endpoint for price estimations instead of price

vkgnosis:
The price endpoint is meant for use with RFQT and might include
liqudity sources that we cannot access until we support RFQT in the
driver. This could lead to quoting better prices or less gas use than we
will see when solving.

### Test Plan
CI


nlordell (approved):
Amazing find!

PR #1346, opened 2021-11-03
Significantly improve orderbook compile times

vkgnosis:
This appears to be a regression in Rust 1.56 where complicated types like what we create with Warp here take a long time to compile when having their trait bounds checked. Workaround is to box them. This has negligible performance impact.

To measure compilation time I ran

```
touch orderbook/src/api.rs && time cargo build --bin orderbook
```

multiple times. With this change I get 4s and without this change 26 s.

### Test Plan
CI

### DIFF
--- a/orderbook/src/api.rs
+++ b/orderbook/src/api.rs
@@ -142,11 +142,11 @@ where
 }
 
 pub trait IntoWarpReply {
-    fn into_warp_reply(self) -> WithStatus<Json>;
+    fn into_warp_reply(self) -> ApiReply;
 }
 
 impl IntoWarpReply for anyhowError {
-    fn into_warp_reply(self) -> WithStatus<Json> {
+    fn into_warp_reply(self) -> ApiReply {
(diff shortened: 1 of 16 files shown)


bh2smith (approved):
I'm ok with this

orderbook/src/api.rs
```diff
@@ -74,6 +76,8 @@ pub fn handle_all_routes(
     routes_with_labels.recover(handle_rejection).with(cors)
 }
 
+pub type ApiReply = warp::reply::WithStatus<warp::reply::Json>;
```
nlordell, line 79: I think there are some modules that already return `WithSTatus` explicitely. Can we maybe change those to use this type alias:

https://github.com/gnosis/gp-v2-services/blob/02eeb54727eccbff521cc08321cb21189d360533/orderbook/src/api.rs#L129

nlordell (approved):
Sounds good to me.

orderbook/src/api.rs
vkgnosis: fixed

PR #1347, opened 2021-11-04
Sunce86/flashbots tx submission

sunce86:
Added flashbots tx submission strategy (as a clone of archer with adjusted submit/cancel methods).

Next step is to add continuous simulation of settlements to strategy: TransactionStrategy::CustomNodes

Test plan:
- Manually submit a transaction on mainnet and make sure the tx is mined and the workflow is ok.


nlordell (commented):
Change looks good to me. Approving given the planned refactor (so we can start using Flashbots RPC right away).

> Next step is to add continuous simulation of settlements to strategy: TransactionStrategy::CustomNodes

It would also be nice to also refactor the `ArcherSwap` and `Flashbots` submission into a shared module and use a trait like:
```rust
trait TransactionSubmitting {
    type Id;
    async fn submit_raw_transaction(...) -> Result<Id>;
    async fn cancel_transaction(id: Id) -> Result<()>;
}
```

To factor out the parts that are specific to the network APIs but keep the shared logic the same. Seeing as the intention of this PR is to be able to use Flashbots as soon as possible, I'm OK with merging as is (with the duplicated logic) and doing the refactor in a follow up.

On a side note, before approving and merging, do you think you can include a manual test plan in order to check that submission is working? I don't mind helping out with this :smile:.

solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -0,0 +1,371 @@
...
+        );
+
+        let nonce_future = self.wait_for_nonce_to_change(nonce);
+
+        futures::select! {
+            method_error = submit_future.fuse() => tracing::info!("stopping submission because simulation failed: {:?}", method_error),
+            new_nonce = nonce_future.fuse() => tracing::info!("stopping submission because account nonce changed to {}", new_nonce),
+        };
+
+        // After stopping submission of new transactions we wait for some time to give a potentially
```
fleupold, line 96: Just curious why do we need to wait for the tx to show up on our node?

solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -0,0 +1,371 @@
...
+            };
+
+            // create transaction
+
+            let tx_gas_cost_in_ether_wei =
+                U256::from_f64_lossy(gas_price.effective_gas_price()) * gas_estimate;
+            let mut settlement = settlement.clone();
+            settlement
+                .encoder
+                .append_to_execution_plan(block_coinbase::PayBlockCoinbase {
```
fleupold, line 208: I don't think we need to do this for flashbots. We can just pay our gas as part of the normal transaction's gas_price.

solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -0,0 +1,371 @@
...
+            let method = super::retry::settle_method_builder(
+                self.contract,
+                settlement.into(),
+                self.account.clone(),
+            )
+            .nonce(nonce)
+            // Wouldn't work because the function isn't payable.
+            // .value(tx_gas_cost_in_ether_wei)
+            .gas(U256::from_f64_lossy(gas_limit))
+            .gas_price(0.0.into());
```
fleupold, line 220: Is this going to work with EIP 1559?

solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -0,0 +1,371 @@
...
+
+            // simulate transaction
+
+            if let Err(err) = method.clone().view().call().await {
+                if let Some((_, previous_tx)) = previous_tx.as_ref() {
+                    if let Err(err) = self.flashbots_api.cancel(previous_tx).await {
```
fleupold, line 226: Just as a side note, even for our regular submission strategy we should try to cancel transactions that we see are about to fail. This way we may also be able to avoid costly failing transactions like https://etherscan.io/tx/0x412cdea1e2e2e1134d5d9ca5164b80748bee3c46ffbb247bc3ef057fff1d3d1f

fleupold (commented):
I think Archer was written pre EIP-1559

Agree with Nick, would be great if you could add a test plan to the PR.

solver/src/settlement_submission/flashbots_settlement.rs
sunce86: Because there is a delay between submission of the tx and the receival of the mined block containing the tx.
I guess Archer API does not support tx status query, while Flashbots does. However, it seemed to me that it is better to rely on the network to check if tx is mined, rather then trust Flashbots API.

solver/src/settlement_submission/flashbots_settlement.rs
sunce86: Yes, because simulation is done using `eth_call` for which gas_price is optional, and if gas_price is equal to None or 0, base fee should be forced to 0.

solver/src/settlement_submission/flashbots_settlement.rs

solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -30,7 +30,26 @@ impl FlashbotsApi {
...
+                tracing::debug!(
+                    "flashbots bundle id: {}",
+                    serde_json::to_string(&bundle_id)
+                        .unwrap_or_else(|err| format!("error: {:?}", err)),
+                );
+                Ok(bundle_id.to_string())
+            }
+            Err(err) => {
+                tracing::debug!("failed to submit: {}", err);
+                Err(anyhow!("failed to submit"))
```
fleupold, line 50: Maybe also propagate the error in the message we bubble up to the caller (otherwise we need to enable debug logs to find the error message)

solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -30,7 +30,26 @@ impl FlashbotsApi {
...
         let body = response.text().await?;
         ensure!(status.is_success(), "status {}: {:?}", status, body);
-        Ok(body)
+
+        match serde_json::from_str::<jsonrpc_core::Output>(&body) {
+            Ok(body) => {
+                let bundle_id = if let jsonrpc_core::Output::Success(ref x) = body {
+                    x.result.as_str().unwrap_or_default()
+                } else {
+                    Default::default()
```
fleupold, line 39: shouldn't the failure cases here also lead to a high level failure of the submission instead of assuming an empty string?

solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -51,4 +70,23 @@ impl FlashbotsApi {
         ensure!(status.is_success(), "status {}: {:?}", status, body);
         Ok(())
     }
+
+    /// Query status of a previously submitted transaction.
+    pub async fn status(&self, bundle_id: &str) -> Result<()> {
```
fleupold, line 75: is this method used?

fleupold (approved):
Approving since the local test is working. Some minor comments inline

solver/src/settlement_submission/flashbots_api.rs
sunce86: Yes, fixed.

solver/src/settlement_submission/flashbots_api.rs
sunce86: Fixed that ugly unwraping.

solver/src/settlement_submission/flashbots_api.rs
sunce86: Not currently, however I plan to use it in the following PR related to transaction strategies refactoring.

solver/src/settlement_submission.rs
```diff
@@ -95,6 +103,24 @@ impl SolutionSubmitter {
...
+                    &self.web3,
+                    &self.contract,
+                    &account,
+                    flashbots_api,
+                    self.gas_price_estimator.as_ref(),
+                    self.gas_price_cap,
+                )?;
+                let result = submitter
+                    .submit(self.target_confirm_time, settlement, gas_estimate)
+                    .await;
```
bh2smith, line 117: I'm guessing this await is not being returned....

PR #1348, opened 2021-11-04
Add flat fee discount command line parameter

nlordell:
This PR adds a command line parameter for specifying a flat fee discount denominated in Eth. Note that the the `orderbook` executable will panic for any other value than `0` because it currently isn't supported. The idea behind this PR is just to make a small change to work up flat fee discounts and keep future PRs shorter.

### Test Plan

No real logic changes. Can check that the parameter defaults to 0 and panics for any other value than 0:
```
$ cargo run -p orderbook
... # Everything OK
$ cargo run -p orderbook -- --fee-discount 42
...
2021-11-04T13:30:05.282Z ERROR shared::tracing: thread 'main' panicked at 'non-0 fee discount currently not supported', orderbook/src/fee.rs:198:9:
thread 'main' panicked at 'non-0 fee discount currently not supported', orderbook/src/fee.rs:198:9
```

### DIFF
--- a/orderbook/src/fee.rs
+++ b/orderbook/src/fee.rs
@@ -28,7 +28,7 @@ pub struct EthAdapter<T> {
 ///
 /// Given an estimated fee for a trade, the mimimum fee required for an order is
 /// computed using the following formula:
-/// ```
+/// ```text
 /// (estimated_fee_in_eth - fee_discount) * fee_factor * (partner_additional_fee_factor || 1.0)
 /// ```
 pub struct FeeSubsidyConfiguration {
--- a/orderbook/src/fee.rs
+++ b/orderbook/src/fee.rs
@@ -24,9 +24,27 @@ pub struct EthAdapter<T> {
     weth: H160,
 }
 
+/// Fee subsidy configuration.
+///
+/// Given an estimated fee for a trade, the mimimum fee required for an order is
+/// computed using the following formula:
+/// ```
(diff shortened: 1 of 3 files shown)


orderbook/src/fee.rs
```diff
@@ -24,16 +24,31 @@ pub struct EthAdapter<T> {
     weth: H160,
 }
 
+pub struct FeeSubsidyConfiguration {
+    pub fee_discount: f64,
```
fleupold, line 28: maybe some docs on what is the difference between discount and factor (also make it bit more clear if this is a discount in eth or a discount in gas units)? Also I was wondering if this should be an enum or if we allow both to be set at the same time we should document which one goest first.

Edit: I see it's been documented on the command line argument. Not sure if it's worthwhile copying the documentation here?

fleupold (approved):
LGTM 

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia4.giphy.com%2Fmedia%2FjPGWvQM7IKGTYz67Jf%2Fgiphy.gif&token=connect" data-gitmeme-token="connect"></a>

orderbook/src/fee.rs
nlordell: I'll add documentation and copy it here.

sunce86 (approved):
lg

bh2smith (approved):

Issue #1333, opened 2021-11-01
Subsidy could be a fixed ETH amount rather than a factor


vkgnosis:
Related to this we discussed previously that the subsidy should maybe be a fixed ETH amount (rather than a factor). This way we can say we fully subsidise the GP overhead up to a certain gas price regardless of trade size.

_Originally posted by @fleupold in https://github.com/gnosis/gp-v2-services/issues/1325#issuecomment-956091300_

PR #1349, opened 2021-11-04
Store all parts the the fee amount is made of separately in DB

fleupold:
Part of #1333 (other side of #1348)

This PR changes the fee measurement table in a way that allows us to recover the amount of eth that the estimate was based on (by storing gas price as well as gas amount).

This allows us to not only apply a subsidy factor, but also a concrete subsidy in fixed eth (or even just gas units if we wanted to).

**Note about a sneaky change:** While debugging this, I noticed that we swallow DB errors in the fee estimation silently and just keep re-querying fee estimates in case of DB errors. I found this dangerous (as it's easy to commit a faulty SQL query without getting errors) and therefor changed the logic to error on DB query errors.

### Test Plan
Existing unit tests (when they are fixed) + made sure that I can locally quote and place orders (checking logs that measurements are read correctly)

### DIFF
--- a/orderbook/src/database/fees.rs
+++ b/orderbook/src/database/fees.rs
@@ -134,6 +134,17 @@ mod tests {
     use model::order::OrderKind;
     use primitive_types::H160;
 
+    // Convenience to allow using u32 in tests instead of the struct
+    impl From<u32> for UnsubsidizedFee {
+        fn from(v: u32) -> Self {
+            UnsubsidizedFee {
+                gas_amount: v as f64,
+                gas_price: 1.0,
+                sell_token_price: 1.0,
+            }
+        }
+    }
+
     #[tokio::test]
     #[ignore]
     async fn postgres_save_and_load_fee_measurements() {
--- a/orderbook/src/database/fees.rs
+++ b/orderbook/src/database/fees.rs
@@ -1,14 +1,26 @@
 use super::{orders::DbOrderKind, Postgres};
-use crate::{
-    conversions::*,
-    fee::{FeeData, MinFeeStoring},
-};
+use crate::{conversions::*, fee::{FeeData, MinFeeStoring, UnsubsidizedFee}};
 
-use anyhow::{anyhow, Context, Result};
-use bigdecimal::BigDecimal;
+use anyhow::{Context, Result};
 use chrono::{DateTime, Utc};
-use ethcontract::U256;
 use shared::maintenance::Maintaining;
+
+#[derive(sqlx::FromRow)]
+struct FeeRow {
+    pub gas_amount: f64,
+    pub gas_price: f64,
+    pub sell_token_price: f64,
+}
+
+impl FeeRow {
+    fn into_fee(self) -> UnsubsidizedFee {
+        UnsubsidizedFee {
+            gas_amount: self.gas_amount,
+            gas_price: self.gas_price,
+            sell_token_price: self.sell_token_price,
+        }
+    }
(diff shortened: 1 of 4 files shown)


orderbook/src/database/fees.rs
```diff
@@ -1,32 +1,46 @@
...
+use anyhow::{Context, Result};
 use chrono::{DateTime, Utc};
-use ethcontract::U256;
 use shared::maintenance::Maintaining;
 
+#[derive(sqlx::FromRow)]
+struct FeeRow {
+    pub gas_amount: f64,
+    pub gas_price: f64,
+    pub sell_token_price: f64,
```
nlordell, line 12: Note that these don't need to be `pub`. Fields are visible in the module (and sub-modules) already and the `FeeRow` type isn't public so can't be used outside anyway.

```suggestion
    gas_amount: f64,
    gas_price: f64,
    sell_token_price: f64,
```

orderbook/src/database/fees.rs
```diff
@@ -37,69 +51,58 @@ impl MinFeeStoring for Postgres {
...
         const QUERY: &str = "\
-            SELECT MIN(min_fee) FROM min_fee_measurements \
+            SELECT gas_amount, gas_price, sell_token_price FROM min_fee_measurements \
             WHERE
                 sell_token = $1 AND \
                 buy_token = $2 AND \
                 amount = $3 AND \
                 order_kind = $4 AND \
                 expiration_timestamp >= $5 \
+            ORDER BY gas_amount * gas_price * sell_token_price ASC \
```
nlordell, line 64: Not super familiar with `sqlx`, but is there any point in adding this:

```suggestion
            ORDER BY gas_amount * gas_price * sell_token_price ASC \
            LIMIT 1
```

If `fetch_optional` just reads the first row and then stops the query, then I don't think it makes a difference.

orderbook/src/fee.rs
```diff
@@ -257,12 +272,12 @@ impl MinFeeCalculating for MinFeeCalculator {
 
         tracing::debug!(?fee_data, ?app_data, "computing subsidized fee",);
 
-        let unsubsidized_min_fee = if let Ok(Some(past_fee)) = self
+        let unsubsidized_min_fee = if let Some(past_fee) = self
```
nlordell, line 275: I don't know if this change was intentional, but it changes the behaviour of fee estimates.

Before, if you would get a DB error reading reading past fees, then we would fall back to computing a new fee amount. Now, the whole `compute_subsidized_min_fee` call will fail. I'm fine with this behaviour change, but maybe add it to the PR descriptions?

fleupold: Right, I added this because I had an SQL error in my query before and nothing ever warned me about it. I'm ok changing this to ignoring DB errors (but we should at least print them). I felt like in other places we have made increasingly the decision to error whenever unexpected things happen (a DB being one of them). Will add to the PR description.

nlordell: To clarify, I agree with this change. Just wanted to make sure it was intentional.

orderbook/src/database/fees.rs
nlordell: Does it make a difference? It was more of a question :stuck_out_tongue:.

nlordell (approved):
Looks good to me! Your Rust isn't rust-y at all :stuck_out_tongue:.

sunce86 (approved):
lg

PR #1353, opened 2021-11-05
Gas calculation modifications

josojo:
This PR reduces gas costs in two cases:

**paraswap and zeroEx estimates:** Before the PR, the gas costs calculations contained the initialisation_gas costs of a tx 2-times, once in our SETTLEMENT_SINGLE_TRADE and in the estimate from the dex-ags.
Hence, we are switching to only the SETTLEMENT_OVERHEAD, not containing the gas costs of the intilziation.

**Baseline Estimates:** Usually, we have gas-refunds and they were not considered. I added them into the SETTLEMENT_SINGLE_TRADE estimate and hence our baseline estimator should result in lower estimates.

### Test Plan
no tests, just making sure that the logic is correct.

### DIFF
--- a/shared/src/price_estimation/gas.rs
+++ b/shared/src/price_estimation/gas.rs
@@ -1,9 +1,12 @@
 ///! constants to estimate gas use in GPv2
 
+/// gas for initialization
+pub const INITIALIZATION_COST: u64 =
+    // initial tx gas
+    32_000;
+
 /// minimum gas every settlement takes
 pub const SETTLEMENT: u64 =
-    // initial tx gas
-    32_000 +
     // isSolver
     7365;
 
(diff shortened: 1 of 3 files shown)


fleupold (approved):
Makes sense to me. I only think we need to hotfix the release if the current overhead + subsidy is causing issues. For this (until we have #1333 merged) I'd even suggest a "simple" hardcoded value of 80k.

shared/src/price_estimation/gas.rs
```diff
@@ -22,5 +25,13 @@ pub const TRADE: u64 =
 /// of the 90% most traded tokens by volume in the month of Oct. 2021.
 pub const ERC20_TRANSFER: u64 = 27_513;
 
+/// lower bound for gas refunds
+/// this number was derived from some empiric observations
+pub const GAS_REFUNDS: u64 = 13_000;
+
 /// a settlement that contains one trade
-pub const SETTLEMENT_SINGLE_TRADE: u64 = SETTLEMENT + TRADE + 2 * ERC20_TRANSFER;
+pub const SETTLEMENT_SINGLE_TRADE: u64 =
```
fleupold, line 33: nanonit: Since this is only used in a single place now, maybe it should be defined there?

nlordell (approved):

shared/src/price_estimation/gas.rs
nlordell: Personally, I'm slightly partial to having all our gas constants in one place.

PR #1354, opened 2021-11-05
Appends the chain_id to the metadata sent to the http solver.

marcovc:
To make it easier to debug things on the solver end.


solver/src/solver/http_solver.rs
```diff
@@ -219,7 +219,11 @@ impl HttpSolver {
             orders: order_models,
             amms: amm_models,
             metadata: Some(MetadataModel {
-                environment: Some(self.network_id.clone()),
+                environment: Some(format!(
+                    "{} / {}",
+                    self.network_id.clone(),
+                    self.chain_id.clone()
```
nlordell, line 225: Note that for Mainnet, Rinkeby and xDai, `network_id == chain_id` AFAIK.

marcovc: I'm getting "Ethereum" as network_id currently. Are you saying I will get "Ethereum / Ethereum" for all of those networks with this change? If so I must really revise this :)

marcovc: I recall I coded [this](https://github.com/gnosis/gp-v2-services/blob/ae3d2387cf0bc464cab38fdffb62b3f3263d7165/shared/src/network.rs) a while ago. Maybe I could make it more informative and just use network_name here ?

marcovc: I changed the code to use that function instead.

josojo (approved):

solver/src/solver/http_solver.rs
nlordell: To me, the network ID is the result of `net_version` RPC call and the chain ID is the result of the `eth_chainId` RPC call. They can be different, but for all networks we support they are the same (1, 4, and 100 respectively).

nlordell (approved):
Ahh, I see the issue `self.network_id` was not what I expected it to be. Looks good now!

marcovc:
> Ahh, I see the issue self.network_id was not what I expected it to be. Looks good now!

Thanks, yes that should have been better named to avoid this kind of confusion. Also fixed now!

vkgnosis (approved):

PR #1357, opened 2021-11-07
Sunce86/flashbots tx submission3

sunce86:
1. We suspect that we are sending too low max priority fee per gas. Hardcoded to 10gwei temporarily, before flashbots custom gas price estimator is implemented.

2. In case of tx cancellation failure, we want to skip submitting a new tx.

Todo:
1. Implement proper gas price estimator from flashbots
2. Add a deadline to avoid being stuck in case of a flashbots api fails for any reason.

Test plan
- [x] Manual tx to pass on local


bh2smith (approved):
Seems reasonable for now... wouldn't 5Gwei tip be also more than most?

josojo (approved):

PR #1358, opened 2021-11-07
Sunce86/flashbots cancellation logic changed

sunce86:
`eth_sendRawTransactions` used instead of `eth_sendRawTransaction`

Flashbots cancellation logic changed to return instead of continue.
(If we fail to send proper cancel request, then no need to try to submit further requests)

Added more detailed cancellation tracing.

Waiting for the mined tx increased from 20s to 60s because of the flashbots API generally being slower then expected.

Added total deadline of 90s for the whole submission process to prevent being stuck for the long time.

### DIFF
--- a/solver/src/settlement_submission/flashbots_api.rs
+++ b/solver/src/settlement_submission/flashbots_api.rs
@@ -19,7 +19,7 @@ impl FlashbotsApi {
         let body = serde_json::json!({
           "jsonrpc": "2.0",
           "id": 1,
-          "method": "eth_sendRawTransaction",
+          "method": "eth_sendRawTransactions",
           "params": [params],
         });
         tracing::debug!(
@@ -35,11 +35,7 @@ impl FlashbotsApi {
             Ok(body) => match body {
                 jsonrpc_core::Output::Success(body) => match body.result.as_str() {
                     Some(result) => {
-                        tracing::debug!(
-                            "flashbots bundle id: {}",
-                            serde_json::to_string(&result)
-                                .unwrap_or_else(|err| format!("error: {:?}", err)),
-                        );
+                        tracing::debug!("flashbots bundle id: {}", result);
                         Ok(result.to_string())
                     }
                     None => Err(anyhow!("result not a string")),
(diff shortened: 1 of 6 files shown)


solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -243,10 +243,8 @@ impl<'a> FlashbotsSolutionSubmitter<'a> {
                             err
                         );
 
-                        // if cancelation fails, we dont want to submit a new tx, rather contine and wait
-                        // for a nonce to change and end this function or a next cancelation to succeed.
-                        tokio::time::sleep(UPDATE_INTERVAL).await;
-                        continue;
+                        // if cancellation fails, we dont want to submit a new tx
```
fleupold, line 246: Nit: can you add more context to this comment on why this is needed (right now it just explains what the next line does)

solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -229,7 +229,7 @@ impl<'a> FlashbotsSolutionSubmitter<'a> {
                         tracing::error!("flashbots cancellation failed: {:?}", err);
                     }
                 }
-                return err;
+                return Err(anyhow!("flashbots failed simulation: {}", err));
```
fleupold, line 232: Is Ok returned in any of the branches? Otherwise wouldn't it be just enough to change the error type?

solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -82,7 +82,10 @@ impl FlashbotsApi {
                 },
                 jsonrpc_core::Output::Failure(body) => Err(anyhow!(body.error)),
             },
-            Err(err) => Err(anyhow!(err)),
+            Err(err) => {
+                tracing::error!("flashbot cancellation response: {}", body);
```
fleupold, line 86: Nit: every error severity log will lead to an alert in our slack channel. Given we only want this for debugging maybe a warning or info severity is sufficient?

fleupold (approved):
Looks good to me. The PR description could give a little bit more context on why this is needed and what problem you were seeing in staging that you are trying to fix.

josojo (approved):

solver/src/settlement_submission/flashbots_settlement.rs

solver/src/settlement_submission/flashbots_settlement.rs
sunce86: True, fixed.

solver/src/settlement_submission/flashbots_api.rs
sunce86: Changed to info.

solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -69,21 +65,17 @@ impl FlashbotsApi {
...
                         Ok(())
                     }
-                    None => Err(anyhow!("result not a string")),
+                    None => Err(anyhow!("result not a bool")),
                 },
                 jsonrpc_core::Output::Failure(body) => Err(anyhow!(body.error)),
             },
             Err(err) => {
-                tracing::error!("flashbot cancellation response: {}", body);
+                tracing::info!("flashbot cancellation response: {}", body);
```
josojo, line 78: I would make this a warn. for info, we try to keep only the short and succinct high-level overviews.

josojo (approved):
Nice

solver/src/settlement_submission/flashbots_api.rs
josojo: Ups, just saw that felix even suggested info. Let's keep it info...

PR #1361, opened 2021-11-08
Cleanup dependencies

vkgnosis:
I looked through crates where we compile two versions and changed this:

- Remove BigDecimal where it isn't needed and import types through num
  instead when possible
- Remove default warp features (multipart and websocket)

The BigInteger conversion can go away once sqlx updates their BigDecmial dependency which I have made a PR for.

### Test Plan
CI

### DIFF
--- a/model/src/ratio_as_decimal.rs
+++ b/model/src/ratio_as_decimal.rs
@@ -1,11 +1,8 @@
 use bigdecimal::BigDecimal;
-use num::{bigint::Sign as Sign04, BigRational};
-use num_bigint::{BigInt, Sign as Sign03};
+use num::{BigInt, BigRational};
 use serde::{de, Deserialize, Deserializer, Serializer};
 use serde_with::{DeserializeAs, SerializeAs};
-use std::borrow::Cow;
-use std::convert::TryInto;
-use std::str::FromStr;
+use std::{borrow::Cow, convert::TryInto, str::FromStr};
 
 pub struct DecimalBigRational;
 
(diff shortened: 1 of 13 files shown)


model/src/ratio_as_decimal.rs
```diff
@@ -61,23 +57,6 @@ where
...
-}
-
-fn sign_03_to_04(sign_03: Sign03) -> Sign04 {
-    match sign_03 {
-        Sign03::Minus => Sign04::Minus,
-        Sign03::NoSign => Sign04::NoSign,
-        Sign03::Plus => Sign04::Plus,
-    }
-}
-
```
bh2smith, line 80: So glad to be rid of this!

orderbook/src/conversions.rs
```diff
@@ -51,6 +52,30 @@ pub fn h256_from_vec(vec: Vec<u8>) -> Result<H256> {
...
+    bigdecimal::num_bigint::BigInt::new(
+        match sign {
+            Sign::Minus => bigdecimal::num_bigint::Sign::Minus,
+            Sign::NoSign => bigdecimal::num_bigint::Sign::NoSign,
+            Sign::Plus => bigdecimal::num_bigint::Sign::Plus,
+        },
+        digits,
+    )
+}
+
```
bh2smith, line 78: Oh no! Its back again here. Should the unit tests be moved over or did you decide to leave them out because they were trivial and silly?

bh2smith (approved):

orderbook/src/conversions.rs
vkgnosis: Yes I felt the tests didn't add value here.

nlordell (approved):

PR #1362, opened 2021-11-08
Flashbots gas price implemented

sunce86:
Several improvements:

1. Increased deadline for flashbots transaction submitting from 90s to 120s (with flashbots occasional slow API this should reduce number of deadline expires)
2. Implemented eth_gasFees - get gas price estimation directly from flashbots API - in case of failure, fallback to standard gas price estimator. (this should make inclusion faster)
3. Added 2s cancellation period - increase our chances of successful submit of next transaction.

Test plan
- [x] Manual tx to pass on local

### DIFF
--- a/solver/src/main.rs
+++ b/solver/src/main.rs
@@ -183,7 +183,7 @@ struct Arguments {
     /// network before going to back to solving.
     #[structopt(
         long,
-        default_value = "90",
+        default_value = "120",
         parse(try_from_str = shared::arguments::duration_from_seconds),
     )]
     max_flashbots_submission_seconds: Duration,
--- a/solver/src/settlement_submission/flashbots_api.rs
+++ b/solver/src/settlement_submission/flashbots_api.rs
@@ -1,5 +1,6 @@
 use anyhow::{anyhow, ensure, Result};
 use gas_estimation::{EstimatedGasPrice, GasPrice1559};
+use primitive_types::U256;
 use reqwest::Client;
 use serde::Deserialize;
 
@@ -13,14 +14,14 @@ pub struct FlashbotsApi {
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
 struct Eip1559 {
-    max_fee_per_gas: String,
-    max_priority_fee_per_gas: String,
+    max_fee_per_gas: U256,
+    max_priority_fee_per_gas: U256,
 }
 
 #[derive(Debug, Deserialize)]
 #[serde(rename_all = "camelCase")]
 struct FlashbotGasPrice {
-    base_fee_per_gas: String,
+    base_fee_per_gas: U256,
     default: Eip1559,
 }
 
@@ -138,42 +139,12 @@ impl FlashbotsApi {
                     match serde_json::from_value::<FlashbotGasPrice>(body.result) {
                         Ok(gas_price) => Ok(EstimatedGasPrice {
                             eip1559: Some(GasPrice1559 {
-                                base_fee_per_gas: match u64::from_str_radix(
-                                    &gas_price.base_fee_per_gas[2..],
-                                    16,
-                                ) {
-                                    Ok(base_fee) => base_fee as f64,
-                                    Err(err) => {
-                                        return Err(anyhow!(
-                                            "failed to parse base_fee_per_gas: {}",
-                                            err
-                                        ))
-                                    }
-                                },
-                                max_fee_per_gas: match u64::from_str_radix(
-                                    &gas_price.default.max_fee_per_gas[2..],
-                                    16,
-                                ) {
-                                    Ok(max_fee_per_gas) => max_fee_per_gas as f64,
-                                    Err(err) => {
-                                        return Err(anyhow!(
-                                            "failed to parse max_fee_per_gas: {}",
-                                            err
-                                        ))
-                                    }
-                                },
-                                max_priority_fee_per_gas: match u64::from_str_radix(
-                                    &gas_price.default.max_priority_fee_per_gas[2..],
-                                    16,
-                                ) {
-                                    Ok(max_priority_fee_per_gas) => max_priority_fee_per_gas as f64,
-                                    Err(err) => {
-                                        return Err(anyhow!(
-                                            "failed to parse max_priority_fee_per_gas: {}",
-                                            err
-                                        ))
-                                    }
-                                },
+                                base_fee_per_gas: gas_price.base_fee_per_gas.to_f64_lossy(),
+                                max_fee_per_gas: gas_price.default.max_fee_per_gas.to_f64_lossy(),
+                                max_priority_fee_per_gas: gas_price
+                                    .default
+                                    .max_priority_fee_per_gas
+                                    .to_f64_lossy(),
                             }),
                             ..Default::default()
                         }),
@@ -227,13 +198,12 @@ mod tests {
             Output::Success(s) => {
                 let deserialized = serde_json::from_value::<FlashbotGasPrice>(s.result).unwrap();
                 assert_eq!(
-                    u64::from_str_radix(&deserialized.default.max_fee_per_gas[2..], 16).unwrap(),
-                    115330291230
+                    deserialized.default.max_fee_per_gas,
+                    U256::from(115330291230u64)
                 );
                 assert_eq!(
-                    u64::from_str_radix(&deserialized.default.max_priority_fee_per_gas[2..], 16)
-                        .unwrap(),
-                    11533029123
+                    deserialized.default.max_priority_fee_per_gas,
+                    U256::from(11533029123u64)
                 );
             }
             Output::Failure(_) => panic!(),
--- a/solver/src/settlement_submission/flashbots_api.rs
+++ b/solver/src/settlement_submission/flashbots_api.rs
@@ -1,5 +1,6 @@
-use anyhow::{anyhow, ensure, Result};
+use anyhow::{anyhow, bail, ensure, Context, Result};
 use gas_estimation::{EstimatedGasPrice, GasPrice1559};
+use jsonrpc_core::Output;
 use primitive_types::U256;
 use reqwest::Client;
 use serde::Deserialize;
@@ -48,16 +49,16 @@ impl FlashbotsApi {
         let body = response.text().await?;
         ensure!(status.is_success(), "status {}: {:?}", status, body);
 
-        match serde_json::from_str::<jsonrpc_core::Output>(&body) {
+        match serde_json::from_str::<Output>(&body) {
             Ok(body) => match body {
-                jsonrpc_core::Output::Success(body) => match body.result.as_str() {
+                Output::Success(body) => match body.result.as_str() {
                     Some(bundle_id) => {
                         tracing::debug!("flashbots bundle id: {}", bundle_id);
                         Ok(bundle_id.to_string())
                     }
                     None => Err(anyhow!("result not a string")),
                 },
-                jsonrpc_core::Output::Failure(body) => Err(anyhow!(body.error)),
+                Output::Failure(body) => Err(anyhow!(body.error)),
             },
             Err(err) => Err(anyhow!(err)),
         }
@@ -80,9 +81,9 @@ impl FlashbotsApi {
         let body = response.text().await?;
         ensure!(status.is_success(), "status {}: {:?}", status, body);
 
-        match serde_json::from_str::<jsonrpc_core::Output>(&body) {
+        match serde_json::from_str::<Output>(&body) {
             Ok(body) => match body {
-                jsonrpc_core::Output::Success(body) => match body.result.as_bool() {
+                Output::Success(body) => match body.result.as_bool() {
                     Some(success) => {
                         tracing::debug!("flashbots cancellation request sent: {}", success);
                         match success {
@@ -92,7 +93,7 @@ impl FlashbotsApi {
                     }
                     None => Err(anyhow!("result not a bool")),
                 },
-                jsonrpc_core::Output::Failure(body) => Err(anyhow!(body.error)),
+                Output::Failure(body) => Err(anyhow!(body.error)),
             },
             Err(err) => {
                 tracing::info!("flashbot cancellation response: {}", body);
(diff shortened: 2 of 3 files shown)


bh2smith (approved):
Cool stuff!

solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -99,4 +118,125 @@ impl FlashbotsApi {
...
+                    }
+                }
+                jsonrpc_core::Output::Failure(body) => Err(anyhow!(body.error)),
+            },
+            Err(err) => {
+                tracing::info!("flashbot cancellation response: {}", body);
+                Err(anyhow!(err))
+            }
+        }
+    }
```
vkgnosis, line 190: Tried to make this easier to read by eliminating right ward drift.

```suggestion
    pub async fn gas_price(&self) -> Result<EstimatedGasPrice> {
        let body = serde_json::json!({
            "jsonrpc": "2.0",
            "id": 1,
            "method": "eth_gasFees",
            "params": [],
        });
        let response = self.client.post(URL).json(&body).send().await?;
        let status = response.status();
        let body = response.text().await?;
        ensure!(status.is_success(), "status {}: {:?}", status, body);

        let output = serde_json::from_str::<jsonrpc_core::Output>(&body)
            .context(format!("body: {}", body))?;
        let success = match output {
            jsonrpc_core::Output::Success(success) => success,
            jsonrpc_core::Output::Failure(body) => anyhow::bail!(body.error),
        };
        let gas_price = serde_json::from_value::<FlashbotGasPrice>(success.result)
            .context("failed to deserialize FlashbotGasPrice")?;

        Ok(EstimatedGasPrice {
            eip1559: Some(GasPrice1559 {
                base_fee_per_gas: match u64::from_str_radix(&gas_price.base_fee_per_gas[2..], 16) {
                    Ok(base_fee) => base_fee as f64,
                    Err(err) => return Err(anyhow!("failed to parse base_fee_per_gas: {}", err)),
                },
                max_fee_per_gas: match u64::from_str_radix(
                    &gas_price.default.max_fee_per_gas[2..],
                    16,
                ) {
                    Ok(max_fee_per_gas) => max_fee_per_gas as f64,
                    Err(err) => return Err(anyhow!("failed to parse max_fee_per_gas: {}", err)),
                },
                max_priority_fee_per_gas: match u64::from_str_radix(
                    &gas_price.default.max_priority_fee_per_gas[2..],
                    16,
                ) {
                    Ok(max_priority_fee_per_gas) => max_priority_fee_per_gas as f64,
                    Err(err) => {
                        return Err(anyhow!("failed to parse max_priority_fee_per_gas: {}", err))
                    }
                },
            }),
            ..Default::default()
        })
    }
```

solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -99,4 +118,125 @@ impl FlashbotsApi {
...
+                                        return Err(anyhow!(
+                                            "failed to parse base_fee_per_gas: {}",
+                                            err
+                                        ))
+                                    }
+                                },
+                                max_fee_per_gas: match u64::from_str_radix(
+                                    &gas_price.default.max_fee_per_gas[2..],
+                                    16,
+                                ) {
```
vkgnosis, line 156: Parsing u64 from a hex string starting with 0x is repeated several times so would be nice to have that as it's own function

solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -8,6 +10,20 @@ pub struct FlashbotsApi {
     client: Client,
 }
 
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct Eip1559 {
+    max_fee_per_gas: String,
+    max_priority_fee_per_gas: String,
+}
```
vkgnosis, line 18: Would be nice to replace the manual parsing later with anootations here. We have similar code in crate `model` where we parse U256 using `#[serde(with = "u256_decimal")]`. I don't mind too much because this is a private type that we immediately convert into something else so not needed for this PR but wanted to mention that this is usually nicer.

solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -99,4 +118,125 @@ impl FlashbotsApi {
         ensure!(status.is_success(), "status {}: {:?}", status, body);
         Ok(())
     }
+
+    /// Query gas_price for the current network state (simplest one for Flashbots)
+    pub async fn gas_price(&self) -> Result<EstimatedGasPrice> {
```
vkgnosis, line 123: I'm wondering if this should return `Result<GasPrice1559>` because based on the code below that is always the case instead of the full EstimatedGasPrice struct. But not sure if that makes code using this function harder.

solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -8,6 +10,20 @@ pub struct FlashbotsApi {
     client: Client,
 }
 
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct Eip1559 {
+    max_fee_per_gas: String,
+    max_priority_fee_per_gas: String,
```
nlordell, line 17: Also agree with @vkgnosis:
```suggestion
    #[serde(with = "model::u256_decimal")]
    max_fee_per_gas: U256,
    #[serde(with = "model::u256_decimal")]
    max_priority_fee_per_gas: U256,
```

solver/src/settlement_submission/flashbots_api.rs
sunce86: Not bad idea, however I would leave it as is, since the caller works with EstimatedGasPrice, and also, it will be easier to refactor the submitting strategies later.

solver/src/settlement_submission/flashbots_api.rs
nlordell: If we apply the suggestion :point_up: to use `U256` instead of `String` for the `FlashBotGasPrice` and `Eip1559` types, then we can use `to_f64_lossy()` here to parse `0x...` integers into `f64`s.

solver/src/main.rs
```diff
@@ -183,7 +183,7 @@ struct Arguments {
     /// network before going to back to solving.
     #[structopt(
         long,
-        default_value = "90",
+        default_value = "120",
```
nlordell, line 186: Not that I disagree with the change, just wondering the rational behind it.

solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -8,6 +10,20 @@ pub struct FlashbotsApi {
...
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct FlashbotGasPrice {
+    base_fee_per_gas: String,
```
nlordell, line 23: Same here.

solver/src/main.rs
nlordell: :see_no_evil: Its in the description: 
> Increased deadline for flashbots transaction submitting from 90s to 120s (with flashbots occasional slow API this should reduce number of deadline expires)

solver/src/settlement_submission/flashbots_api.rs
sunce86: With this, I have to write a custom deserializator for these structures right?

solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -199,17 +200,27 @@ impl<'a> FlashbotsSolutionSubmitter<'a> {
...
-                }
+            let gas_price = match self.flashbots_api.gas_price().await {
+                Ok(gas_price) => gas_price,
                 Err(err) => {
-                    tracing::error!("gas estimation failed: {:?}", err);
-                    tokio::time::sleep(UPDATE_INTERVAL).await;
-                    continue;
+                    tracing::warn!("flashbots gas price estimator failed: {:?}", err);
+
+                    // fallback to standard gas price estimators, with hardcoded max_priority_fee if less than 10gwei
```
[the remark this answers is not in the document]
sunce86: With Flashbots API current low reliability, I believe we need a second option in case the flashbot estimator fails to respond.

solver/src/settlement_submission/flashbots_api.rs
nlordell: No, we have existing `serde(with = ...)` modules that you can use.

vkgnosis: Our existing U256 deserialization works on decimal strings but here we have hex so they won't fit. I just mentioned to have a code example but I don't know whether the actual type should be U256 or u64 here.

solver/src/main.rs
sunce86: I noticed that we had several deadline expirations and right after our tx was mined. This is not that important change, we can leave it or remove it (since we can change it as being a config parameter), but I felt that 120 is more adequate.

solver/src/settlement_submission/flashbots_api.rs
nlordell: Oh yeah, brain fart. However, note that `serde_json::from_str::<U256>("\"0x42\"").is_ok()` is the default behaviour for `U256` (de)serialization. So it should "just work".

solver/src/settlement_submission/flashbots_api.rs
sunce86: Obsolete with the latest commit.

solver/src/settlement_submission/flashbots_api.rs
sunce86: Used `to_f64_lossy()`

solver/src/settlement_submission/flashbots_api.rs
sunce86: It "just work"-ed :) fixed to plain U256

solver/src/settlement_submission/flashbots_api.rs
sunce86: Fixed.

solver/src/settlement_submission/flashbots_api.rs
sunce86: Fixed.

solver/src/main.rs
nlordell: No its fine, I just missed it in the description. My bad!

nlordell (approved):
LGTM.

solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -99,4 +119,94 @@ impl FlashbotsApi {
...
+                        Err(err) => Err(anyhow!("result not a FlashbotGasPrice: {}", err)),
+                    }
+                }
+                jsonrpc_core::Output::Failure(body) => Err(anyhow!(body.error)),
+            },
+            Err(err) => {
+                tracing::info!("flashbot cancellation response: {}", body);
+                Err(anyhow!(err))
+            }
+        }
```
nlordell, line 160: ubernit: I agree with @vkgnosis that we can still get rid of some "rightward drift" here. Maybe something like:
```suggestion
        let gas_price = serde_json::from_str::<jsonrpc_core::Output>(&body)
            .with_context(|err| {
                tracing::info!("flashbot cancellation response: {}", body);
                anyhow!("invalid Flashbots RPC response: {}", err)
            })
            .and_then(|output| match output {
                jsonrpc_core::Output::Success(body) => serde_json::from_value::<FlashbotGasPrice>(body.result)
                    .context("result not a FlashbotGasPrice"),
                jsonrpc_core::Output::Failure(body) => bail!("Flashbots RPC error: {}", body.error),
            })?;
        Ok(EstimatedGasPrice {
            eip1559: Some(GasPrice1559 {
                base_fee_per_gas: gas_price.base_fee_per_gas.to_f64_lossy(),
                max_fee_per_gas: gas_price.default.max_fee_per_gas.to_f64_lossy(),
                max_priority_fee_per_gas: gas_price
                    .default
                    .max_priority_fee_per_gas
                    .to_f64_lossy(),
            }),
            ..Default::default()
        })
```

vkgnosis: Yes, I still think the readability can be improved with early returns using `?` and `context` more. Also should convert the tracing log into context. I'm fine with merging either way but would be nice as a follow up.

vkgnosis (approved):
.

solver/src/main.rs
fleupold: It would be great if - once we clean this up - we can unify this strategy with Archer and our default submission strategy (this might mean to change this to number of blocks rather than time)

solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -199,17 +200,27 @@ impl<'a> FlashbotsSolutionSubmitter<'a> {
...
+                    match self.gas_price(gas_limit, time_limit).await {
+                        Ok(mut gas_price) => {
+                            if let Some(ref mut eip1559) = gas_price.eip1559 {
+                                if eip1559.max_priority_fee_per_gas < 10_000_000_000.0 {
+                                    eip1559.max_priority_fee_per_gas = 10_000_000_000.0;
```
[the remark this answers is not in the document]
fleupold: maybe log as well what the suggested price was?

fleupold (commented):
LGTM

solver/src/settlement_submission/flashbots_settlement.rs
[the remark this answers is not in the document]
sunce86: @fleupold we have log for gas_price here: https://github.com/gnosis/gp-v2-services/blob/70d3727f5e34e56cbdff3a02ce33d2e6308d758c/solver/src/settlement_submission/flashbots_settlement.rs#L272

Do you think we need additional log at this place to log what is replaced?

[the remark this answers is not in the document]

solver/src/settlement_submission/flashbots_api.rs
sunce86: Oh ok, sorry, I thought the previous comment on readability was about that cumbersome `u64::from_str_radix` calls.
Anyway, this suggestion taken.

PR #1363, opened 2021-11-08
Makes instance_name a valid file name again.

marcovc:
(Which was the case until a recent commit)

### DIFF
--- a/solver/src/solver/http_solver.rs
+++ b/solver/src/solver/http_solver.rs
@@ -297,8 +297,13 @@ impl HttpSolver {
             self.network_name,
             self.chain_id
         )
-        .replace(" ", "_")
-        .replace("/", "_")
+        .chars()
+        .map(|x| match x {
+            ' ' => '_',
+            '/' => '_',
+            _ => x,
+        })
+        .collect()
     }
 }
 
--- a/solver/src/solver/http_solver.rs
+++ b/solver/src/solver/http_solver.rs
@@ -293,10 +293,12 @@ impl HttpSolver {
         let now = chrono::Utc::now();
         format!(
             "{}_{}_{}",
-            now.to_string().replace(" ", "_"),
+            now.to_string(),
             self.network_name,
             self.chain_id
         )
+        .replace(" ", "_")
+        .replace("/", "_")
     }
 }
 


solver/src/solver/http_solver.rs
```diff
@@ -293,10 +293,12 @@ impl HttpSolver {
         let now = chrono::Utc::now();
         format!(
             "{}_{}_{}",
-            now.to_string().replace(" ", "_"),
+            now.to_string(),
             self.network_name,
             self.chain_id
         )
+        .replace(" ", "_")
+        .replace("/", "_")
```
nlordell, line 301: Note that `replace` takes a `Pattern`, so you can:
```suggestion
        .replace(&[' ', '/'], "_")
```

marcovc: Somehow that doesn't work:

vkgnosis, line 301: Instead of manual replacements I think we should have a whitelist of characters. Something like:
1. keep only ascii
2. convert to lower case
3. keep only characters in the range 0-9 and a-z and `-`, `_`,

This is more future proof because it will not break if in the future we get weird path breaking characters.

I also think this should actually be enforced in the solver instead of in Rust because only it can know what path names are valid. Also it is a potential attack vector. What if we name the instance `../../../../../../etc/passwd`? It might allow a user of the api to take over the sovler somehow.

solver/src/solver/http_solver.rs
```diff
@@ -293,10 +293,12 @@ impl HttpSolver {
         let now = chrono::Utc::now();
         format!(
             "{}_{}_{}",
-            now.to_string().replace(" ", "_"),
+            now.to_string(),
```
vkgnosis, line 296: For the time it is better to use explicit format arguments https://docs.rs/chrono/0.4.19/chrono/struct.DateTime.html#method.format instead of using the default to_string and then fixing characters.

solver/src/solver/http_solver.rs
marcovc: Thanks for the suggestions. I agree (except with the last point - let's not attack ourselves!) , but since this is spamming the alerts channel a lot, would it be fine if we merge this as is and then do a proper fix later?

vkgnosis (approved):

solver/src/solver/http_solver.rs
vkgnosis: Yes, i'm fine with it.
Of course we will not attack ourself but this is an important api design decision. What if in the future we for some reason start allowing external people to solve some instances. We want it to be secure by default.

marcovc: I couldn't find a way to use replace with multiple chars but used a match statement instead. I guess it does only one pass as well.

nlordell (approved):

solver/src/solver/http_solver.rs
nlordell: Sorry, this wasn't something that blocked approval on my part, just got distracted by something else.

marcovc: You're right, this is safe only as long as we're currently naming the instances ourselves.

marcovc:
Thanks for the comments. I will get back to this with a proper fix (I agree with @vkgnosis that mangling to filesystem friendly name should happen on the process that is trying to create a file, not here). Will merge it just because there's a lot of spamming in the alerts channel currently and this should take care of that.

PR #1366, opened 2021-11-08
Move HTTP solver API to common crate

taminomara:
In preparation of using Quasimodo for price estimation, we're moving the HTTP solver API code to the common crate. We will then reuse it in orderbook.

### Test Plan
CI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2560,6 +2560,7 @@ dependencies = [
  "assert_approx_eq",
  "async-trait",
  "atty",
+ "chrono",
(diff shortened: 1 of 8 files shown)


vkgnosis:
Is it right that this ends up adding 500 lines? Maybe the deletion of one of the source files wasn't included in the PR.

taminomara:
@vkgnosis fixed

vkgnosis (approved):

nlordell (approved):

PR #1368, opened 2021-11-09
Include e2e crate in workspace default members

vkgnosis:
Draft while testing.

Advantages:
- No rebuilding needed when running e2e tests (good for CI).
- `cargo check` now includes the e2e crate so it now catches
  compilation errors where before you needed to run with --workspace

Disadvantages:
- Slightly higher compilation time by default because e2e crate is
  included. When I tested this was minor: always < 0.5 seconds after
  `touch shared/src/lib.rs` and rebuilding.

### Test Plan
CI

### DIFF
--- a/e2e/tests/e2e/local_node.rs
+++ b/e2e/tests/e2e/local_node.rs
@@ -11,7 +11,7 @@ use std::{
 use web3::{api::Namespace, helpers::CallFuture, Transport};
 
 lazy_static! {
-    static ref GANACHE_MUTEX: Mutex<()> = Mutex::new(());
+    static ref NODE_MUTEX: Mutex<()> = Mutex::new(());
 }
 
 const NODE_HOST: &str = "http://127.0.0.1:8545";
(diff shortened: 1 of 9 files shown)


nlordell:
Only concern is `cargo test` now runs e2e tests as well right?

vkgnosis:
Like the Postgres tests I made them `#ignore`d.

e2e/tests/e2e/local_node.rs
```diff
@@ -77,24 +77,24 @@ impl<T: Transport> Resetter<T> {
 }
 
 #[derive(Debug, Clone)]
-pub struct GanacheApi<T> {
+pub struct LocalNodeApi<T> {
```
nlordell, line 80: uber-naming-nit: Maybe call this `EvmApi` because of the `evm_` prefix to the RPC calls to be consistent with the other RPC namespaces (like `Eth`, `Net`, etc.).

nlordell (approved):

PR #1369, opened 2021-11-09
Disable doctests

vkgnosis:
They are slow to compile because each is its own binary. See https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html .

We only have [one](https://github.com/gnosis/gp-v2-services/blob/32408a704e644f26aab856257a3178ccd57ae25a/model/src/u256_decimal.rs#L68) doctest at the moment so this isn't a problem now but I like the change anyway to remind us to not rely on them in the future. We can still write them as a form of *documentation* but we shouldn't write them as *unit tests*.

### Test Plan
CI


nlordell (approved):
Sounds good to me.

PR #1370, opened 2021-11-09
Move crates into subdirectory

vkgnosis:
This makes the top level more organized because we are not mixing crates with other folders.

### Test Plan
CI


nlordell (approved):
Fine by me.

PR #1371, opened 2021-11-09
Move contract deployment from separate binary into e2e tests

vkgnosis:
We used to manually run the deploy binary while the local node was running. Only then could we execute the e2e tests.
Now the deployment code happens in the e2e tests so that they are truly standalone.

The motivation is that I felt that our current build.rs setup was overkill so it ended up harder to understand and more brittle than it needs to be. Now that we no longer to compile time shenanigans to keep track of locally deployed contract addresses this has improved. Another advantage is that CI no longer needs to rebuild after running deploy.

### Test Plan
CI


nlordell:
Personally, I'm OK with this change. The only issue I see is that we can no longer run GPv2 against a local node (because the code relies on `Contract::deployed()`).

Personally, I'm OK with this, but it would be nice to also have an alternative for starting the full services against a local node.

vkgnosis:
How commonly do we run against a local node when not in an e2e test? And what is the use case? Not saying there isn't one I just don't know because I have never done so. Should the state of the chain be preserved across multiple local runs (like ctr+c, edit code, run again)?

If we don't need to remember state an easy solution is to add a switch to the api so that it calls the same deployment function on startup.
If we do need to remember state we can reintroduce the previous compile time addr files but only load them at run time.

nlordell:
> How commonly do we run against a local node when not in an e2e test?

Not commonly. I have done it the past (for example with #534) but I must say its quite rare.

> If we don't need to remember state an easy solution is to add a switch to the api so that it calls the same deployment function on startup.

This is my use-case and would be 100% OK with this. I think that the address files and the build script contribute to our slow compile times and and am in favour with getting rid of it for something simpler.

vkgnosis:
Although without state I see now we cannot run api and driver on the same deployment. So maybe the second idea is better. For that I would bring the deploy binary back but make it create a single config file with the addresses of the contracts. Then api and driver get a switch "use local deployment config" which can load the file.
Since this use case is uncommon enough that I don't think it will break anyone immediately I would like to implement the fix in a separate PR as a new feature of sort. I will work on that PR next.

nlordell (approved):
LGTM

PR #1372, opened 2021-11-09
Adjust slippage for zeroEx Solver

josojo:
ZeroEx slippage is quite tight right now with 0.05% 
I think this tightness contributes to the huge costs of failed tx. Hence, I want it to be adaptable.

### Test Plan
*<Explain how you and a reviewer have/intend to test this change>*

### DIFF
--- a/crates/solver/src/main.rs
+++ b/crates/solver/src/main.rs
@@ -176,6 +176,10 @@ struct Arguments {
     #[structopt(long, env, default_value = "10")]
     paraswap_slippage_bps: u32,
 
+    /// The slippage tolerance we apply to the price quoted by Paraswap
+    #[structopt(long, env, default_value = "10")]
+    zeroex_slippage_bps: u16,
+
     /// The authorization for the archer api.
     #[structopt(long, env)]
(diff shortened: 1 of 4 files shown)


crates/solver/src/main.rs
```diff
@@ -176,6 +176,10 @@ struct Arguments {
     #[structopt(long, env, default_value = "10")]
     paraswap_slippage_bps: u32,
 
+    /// The slippage tolerance we apply to the price quoted by Paraswap
```
vkgnosis, line 179: Comment still says Paraswap

crates/solver/src/main.rs
```diff
@@ -176,6 +176,10 @@ struct Arguments {
...
+    #[structopt(long, env, default_value = "10")]
+    zeroex_slippage_bps: u16,
```
vkgnosis, line 181: I'd use same type as for the Paraswap slippage for consistency

vkgnosis (approved):
Makes sense that it is configurable.

nlordell (approved):

Issue #1342, opened 2021-11-03
Add flashbots protect API


sunce86:
In order to use flashbots protect API, custom implementation for cancelling logic needs to be added.

Plan is to create a new transaction strategy for flashbots as a clone of archer strategy, and reimplement tx submit and cancelling functions.

From there, further refactor of transaction submission logic can be made for all strategies, in order to extract common code.

nlordell:
I think at first we can implement the "general" submission logic (i.e. making sure that the settlement simulation still passes on new blocks) and then add Flashbots specific support if needed.

Note that we already have Archer (now called Eden) support (which had some custom cancellation and submission logic).

vkgnosis:
I talked about this with Dusan. I felt that we wanted Flashbots quickly so it would be better by starting out with a copy paste of the Archer logic. Then refactoring later to see what components can be merged and refactoring the existing standard node strategy.

nlordell:
OK, sounds fair. I would have done this in "reverse" by moving the simulation of settlements on new blocks to the `CustomNodes` strategy (so that we support all private transaction RPC endpoints in a limited capacity - like TaiChi) and then added custom Flashbots RPC support if needed.

gakonst:
FYI if you ever end up switching to ethers-rs, there's a Flashbots plugin for it https://github.com/onbjerg/ethers-flashbots

sunce86:
Implemented.

PR #1373, opened 2021-11-09
Flashbots added waiting for cancellation

sunce86:
Changes:

1. Reverted timeout for propagating mined transactions from 60s -> 20s (there is no point in having so long timeout now when we have a proper cancelling of the txs)
2. Added a function `cancel_and_wait()` that sends the cancel request and does manual polling of the cancellation confirmation.
This is required because sending a new transaction before the cancellation confirmation would fail.

Test plan:
- [x] Do a manual test on local
- [ ] Add more UTs


solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -26,6 +26,20 @@ struct FlashbotGasPrice {
...
+}
+
+#[derive(Deserialize, PartialEq)]
+#[serde(untagged)]
+enum Status {
+    Pending,
+    Failed,
+    Successful,
+    Cancelled,
+}
```
nlordell, line 41: nit: Manual serialization should be achievable with `serde`:
```suggestion
#[derive(Debug, Deserialize)]
struct FlashbotStatus {
    status: Status,
}

#[derive(Deserialize, PartialEq)]
enum Status {
    #[serde(rename = "PENDING_BUNDLE")]
    Pending,
    #[serde(rename = "FAILED_BUNDLE")]
    Failed,
    #[serde(rename = "SUCCESSFUL_BUNDLE")]
    Successful,
    #[serde(rename = "CANCEL_BUNDLE_SUCCESSFUL")]
    Cancelled,
}
```

crates/solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -118,7 +132,38 @@ impl FlashbotsApi {
...
+        let status = serde_json::from_str::<Output>(&body)
+            .with_context(|| {
+                tracing::info!("flashbot status response: {}", body);
+                anyhow!("invalid Flashbots RPC response")
+            })
+            .and_then(|output| match output {
+                Output::Success(body) => serde_json::from_value::<FlashbotStatus>(body.result)
+                    .context("result not a FlashbotStatus"),
+                Output::Failure(body) => bail!("Flashbots RPC error: {}", body.error),
+            })?;
```
nlordell, line 148: Might be worth putting this into something like:
```rust
fn parse_json_rpc_response<T>(body: &str) -> Result<T>
where
    T: DeserializeOwned,
{
    // ...
}
```

solver/src/settlement_submission/flashbots_api.rs
sunce86: Oh, this is nice :) Was searching for this...

sunce86: Fixed.

crates/solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -107,7 +107,7 @@ impl<'a> FlashbotsSolutionSubmitter<'a> {
         // mined previously submitted transaction time to propagate to our node.
 
         if !transactions.is_empty() {
-            const MINED_TX_PROPAGATE_TIME: Duration = Duration::from_secs(60);
+            const MINED_TX_PROPAGATE_TIME: Duration = Duration::from_secs(20);
```
fleupold, line 110: What I still don't understand is why would we want to "wait for mining" if the method errored or the deadline was reached? Why can't we immediately return in these two select cases? I get that we want to wait if the nonce changes.

crates/solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -187,6 +187,8 @@ impl<'a> FlashbotsSolutionSubmitter<'a> {
         const UPDATE_INTERVAL: Duration = Duration::from_secs(5);
         const CANCEL_PROPAGATION_TIME: Duration = Duration::from_secs(2);
         const MINIMAL_MAX_PRIORITY_FEE: f64 = 10_000_000_000.0;
+        const SUBMIT_RETRY_ATTEMPTS: usize = 5usize; // will be a strategy parameter after the refactor!
```
fleupold, line 190: Famous last words

crates/solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -289,16 +291,50 @@ impl<'a> FlashbotsSolutionSubmitter<'a> {
...
-                .submit_transaction(&raw_signed_transaction)
-                .await
-            {
-                Ok(bundle_id) => {
-                    transactions.push(hash);
-                    previous_tx = Some((gas_price, bundle_id));
+            // submit transaction
+
+            let mut retry = std::cmp::max(SUBMIT_RETRY_ATTEMPTS, 1usize);
+            loop {
```
fleupold, line 297: Can this not be a for loop?

crates/solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -249,19 +251,19 @@ impl<'a> FlashbotsSolutionSubmitter<'a> {
             if let Err(err) = method.clone().view().call().await {
                 if let Some((_, previous_tx)) = previous_tx.as_ref() {
                     if let Err(err) = self.flashbots_api.cancel(previous_tx).await {
-                        tracing::error!("flashbots cancellation request not sent: {:?}", err);
+                        tracing::debug!("flashbots cancellation request not sent: {:?}", err);
```
fleupold, line 254: nit: maybe still warn since we want to get rid of those message by deserialising the cancellation response properly, right?

crates/solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -289,16 +291,50 @@ impl<'a> FlashbotsSolutionSubmitter<'a> {
...
+                        bundle_id
+                    }
+                    Err(err) => {
+                        tracing::error!("flashbots submission failed: {:?}", err);
+                        break;
+                    }
+                };
+
+                // handle retry based on the current status of the submitted tx
+                match self.flashbots_api.is_transaction_failed(&bundle_id).await {
```
fleupold, line 323: What cases does this cover? If the transaction is no longer simulating properly we don't want to retry it do we?

fleupold (commented):
The extra retry code is becoming increasingly harder to read and it seem there are no unit tests. Is there maybe something you can extract into its own future (that can encapsulate retry) and that can be unit tested?

Maybe not right now, but I'm worried about the future maintainability of this critical piece of logic.

Not blocking ofc

nlordell:
> Is there maybe something you can extract into its own future (that can encapsulate retry) and that can be unit tested?

Agree. There is already an open issue for this as it was an initial concern when we started with this submission strategy: https://github.com/gnosis/gp-v2-services/issues/1342

crates/solver/src/settlement_submission/flashbots_settlement.rs
sunce86: I am not the original author of this piece of code, but here is one case for which this is needed:
1. We submit tx to flashbots, and we start counting 10s pause before new loop iteration.
2. In the meantime, block A gets mined (not containing our tx)
3. After some time block B is mined (containing our tx)
4. Our node receives block A.
5. Our 10s is up but our node received only block A because of the delay in block propagation. We simulate tx and it fails, we return back
6. If we don't wait another 20s to receive block B, we wont see mined tx. 

Does this make sense?

crates/solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -26,6 +26,39 @@ struct FlashbotGasPrice {
...
+    T: DeserializeOwned,
+{
+    serde_json::from_str::<Output>(body)
+        .with_context(|| {
+            tracing::info!("flashbot response: {}", body);
+            anyhow!("invalid flashbots response")
+        })
+        .and_then(|output| match output {
+            Output::Success(body) => serde_json::from_value::<T>(body.result)
+                .context("flashbots failed conversion to expected type"),
```
nlordell, line 57: if you want to be fancy you can:
```suggestion
                .with_context(|_| format!("flashbots failed conversion to expected {}", std::any::type_name::<T>())),
```
Although, I'm 100% OK with how it is now (the expected type should in theory be captured by virtue of the stack trace which `anyhow::Error`s create).

crates/solver/src/settlement_submission/flashbots_settlement.rs
nlordell: I'd also like it as `info` or `warn`.

sunce86: Currently, valid cancellation result containing false also raises error. Do we still want to trace with warn even if this case happens, or should we change the return type of `cancel` from `Result<()>` to `Result<bool>` and handle the success outside?

crates/solver/src/settlement_submission/flashbots_settlement.rs
fleupold: Yes, that makes perfect sense (I guess it depends on the type of error, we could probably detect an `OrderFilled` error, but that's advanced). Would you mind documenting that information somewhere in the code?

sunce86: Example added as a comment.

crates/solver/src/settlement_submission/flashbots_settlement.rs
sunce86: Actually yes. Refactored to for loop and moved to a separate function.

crates/solver/src/settlement_submission/flashbots_settlement.rs
sunce86: This is a bad attempt to solve the cancellation issue we are having with flashbots.
I've created a new function `cancel_and_wait` and the code should be a bit more clear and readable now, and also a bit more convenient to change to our "general submitting strategy" that we are aiming to build in, guess, future refactor PR :smile:

crates/solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -81,29 +105,34 @@ impl FlashbotsApi {
...
-                    }
-                    None => Err(anyhow!("result not a bool")),
-                },
-                Output::Failure(body) => Err(anyhow!(body.error)),
-            },
-            Err(err) => {
-                tracing::info!("flashbot cancellation response: {}", body);
-                Err(anyhow!(err))
+        let success = parse_json_rpc_response::<bool>(&body)?;
+        tracing::debug!("flashbots cancellation request sent: {}", success);
```
nlordell, line 109:
```suggestion
        tracing::debug!("flashbots cancellation for bundle {} result: {}", bundle_id, success);
```

sunce86: Added.

vkgnosis (approved):
Logic looks fine to me. Can still make code nicer in some places as suggested but that can also happen in future PRs.

crates/solver/src/settlement_submission/flashbots_api.rs
sunce86: I wanna be fancy :) Added.

crates/solver/src/settlement_submission/flashbots_settlement.rs
sunce86: Set to warn.

nlordell (approved):

crates/solver/src/settlement_submission/flashbots_api.rs
```diff
@@ -187,7 +212,7 @@ mod tests {
           },
         });
 
-        let deserialized = serde_json::from_str::<jsonrpc_core::Output>(&body.to_string()).unwrap();
```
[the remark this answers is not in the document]
sunce86: True. Fixed.

crates/solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -249,26 +256,25 @@ impl<'a> FlashbotsSolutionSubmitter<'a> {
...
+            // If gas price has increased cancel old and submit new transaction.
 
             if let Some((previous_gas_price, previous_tx)) = previous_tx.as_ref() {
                 let previous_gas_price = previous_gas_price.bump(1.125);
                 if gas_price.cap() > previous_gas_price.cap() {
-                    if let Err(err) = self.flashbots_api.cancel(previous_tx).await {
-                        tracing::error!("flashbots cancellation request not sent: {:?}", err);
+                    if let Err(err) = self.flashbots_api.cancel_and_wait(previous_tx).await {
+                        // If we are unable to cancel, lets just wait for some time and try again
+                        // Maybe the current tx will get mined in the meantime, of the cancellation will succeed a bit later
```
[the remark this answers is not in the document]
sunce86: Fixed.

crates/solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -185,7 +193,6 @@ impl<'a> FlashbotsSolutionSubmitter<'a> {
         transactions: &mut Vec<H256>,
     ) -> anyhow::Error {
         const UPDATE_INTERVAL: Duration = Duration::from_secs(5);
-        const CANCEL_PROPAGATION_TIME: Duration = Duration::from_secs(2);
         const MINIMAL_MAX_PRIORITY_FEE: f64 = 10_000_000_000.0;
 
         // The amount of extra gas it costs to include the payment to block.coinbase interaction in
```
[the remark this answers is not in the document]
sunce86: Removed.

PR #1380, opened 2021-11-10
Add metric to track the effective gas price used

nlordell:
This PR adds a new metric that plots the effective gas price of the transaction over time.

Note that this is not an ideal prometheus metric and relies on the fact that we don't settle transactions so often. In the future (especially if we start allowing solvers to mine transactions in parallel) we will likely have to change this.

For now, this will give us a nice plot of with two curves:
- gas estimates over time
- actual gas used over time

### Test Plan

Compiler.

### DIFF
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -526,7 +526,8 @@ impl Driver {
                     }
                 };
                 self.in_flight_orders.mark_settled_orders(block, orders);
-                self.metrics.transaction_gas_price(receipt.effective_gas_price);
+                self.metrics
+                    .transaction_gas_price(receipt.effective_gas_price);
             }
             self.metrics.transaction_submission(start.elapsed());
 
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -526,6 +526,7 @@ impl Driver {
                     }
                 };
                 self.in_flight_orders.mark_settled_orders(block, orders);
+                self.metrics.transaction_gas_price(receipt.effective_gas_price);
             }
             self.metrics.transaction_submission(start.elapsed());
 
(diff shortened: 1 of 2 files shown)


crates/solver/src/driver.rs
```diff
@@ -528,6 +528,8 @@ impl Driver {
                     }
                 };
                 self.in_flight_orders.mark_settled_orders(block, orders);
+                self.metrics
+                    .transaction_gas_price(receipt.effective_gas_price);
```
vkgnosis, line 532: Alternatively we could track in the solution submission strategies. The difference is that we get a gas price even when the TX fails and we get the gas price earlier. We can update the gas price every time we resubmit. On the other hand for flashbots at least we would sometimes gauge gas prices where the transactions didn't really happen but would have allowed it to happen.

vkgnosis (approved):

crates/solver/src/metrics.rs
```diff
@@ -184,6 +187,13 @@ impl Metrics {
         })?;
         registry.register(Box::new(transaction_submission.clone()))?;
 
+        let opts = Opts::new(
+            "transaction_gas_price",
```
taminomara, line 191: BTW there's a [naming convention for prometheus metrics][1], and it requires names to include units. So I guess this should be `transaction_gas_price_gwei`.

[1]: https://prometheus.io/docs/practices/naming/

taminomara (approved):
LGTM, except for naming nit.

crates/solver/src/driver.rs
nlordell: I thought about this a bit more.

Really my goal with this metric is to plot the gas estimate and what we actually use in the same plot.

Ideally, if we wanted more metrics, we could do, as you mentioned, pass in metrics to the settlement submission and also record all prices that were used as well as their "result" (like `replaced` if it was replaced because it took too long) and maybe even find a way to incorporate gas price used -> time to mine.

I'll create an issue to address this.

nlordell: https://github.com/gnosis/gp-v2-services/issues/1421

PR #1382, opened 2021-11-11
Speed up hardhat installation on CI

vkgnosis:
Installing hardhat from scratch takes ~30 seconds. With this PR we cache it instead. The cache is small so this is much faster (1s).
Also changed how flyway is cached to make it nicer (put files into separate directory, don't run install if cache action ran).

### Test Plan
CI still works

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -26,20 +26,47 @@ jobs:
       CARGO_PROFILE_DEV_DEBUG: 0
       CARGO_PROFILE_TEST_DEBUG: 0
       FLYWAY_VERSION: 7.3.1
+      FLYWAY_PATH: /home/runner/flyway
       # Flyway is connecting via TCP/IP (authentication via password), whereas psql connects via Unix-domain sockets (authentication via username)
       # We set the db password of $USER to this value which is also used by the flyway command
       FLYWAY_PASSWORD: password
+      # Extra underscore because conflicts with hardhat internally used variables.
+      HARDHAT_VERSION_: 2.6.8
+      HARDHAT_PATH_: /home/runner/hardhat
     steps:
       - uses: actions/checkout@v2
       - uses: Swatinem/rust-cache@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request.yaml
```diff
@@ -26,20 +26,47 @@ jobs:
...
+        with:
+          path: ${{ env.HARDHAT_PATH_ }}/node_modules
+          key: ${{ runner.os }}-hardhat-${{ env.HARDHAT_VERSION_ }}
+      - name: Install Hardhat
+        if: steps.cache-hardhat.outputs.cache-hit != 'true'
+        run: |
+          mkdir --parents ${{ env.HARDHAT_PATH_ }}
+          cd ${{ env.HARDHAT_PATH_ }}
+          npm install hardhat@${{ env.HARDHAT_VERSION_ }}
+          ls -al
```
nlordell, line 66: Instead of installing `hardhat` in some directly, why not use the global NPM package directory and install with `npm install -g hardhat@${{ env.HARDHAT_VERSION_ }}`? The installation directory to cache would be `npm root -g` and the executable path `$(npm root -g)/node_modules/.bin/hardhat`.

nlordell (approved):
LGTM, just suggest using the global NPM package root instead of a custom one.

.github/workflows/pull-request.yaml
vkgnosis: Thanks for the suggestion. Will try it out. I don't know much about node, npm so the current solution was trial and error. Didn't know about the global thing.

vkgnosis: I had a look. It is problematic that this directory already contains a bunch of pre installed in the CI container. We don't want to cache those but I'm not sure how to only cache hardhat. So will keep the current version for now unless someone knows how to work around that.

nlordell: Lets keep the current version then.

PR #1384, opened 2021-11-11
Treat Flashbots simulation error like method errors

fleupold:
Fixes noisy alerts whenever we cancel a flashbots tx due to it no longer succeeding to simulate

### Test Plan

Before the PR

> 2021-11-11T15:13:05.284Z ERROR solver::driver: Failed to submit 0x settlement: transaction did not get mined in time

Running this PR with against mainnet (with the adjustment of overriding miner tip with 0 to ensure we never actually get mined):

> 2021-11-11T15:51:54.062Z  WARN solver::driver: Failed to submit 0x settlement: flashbots failed simulation
Caused by:
    0: method 'settle(address[],uint256[],(uint256,uint256,address,uint256,uint256,uint32,bytes32,uint256,uint256,uint256,bytes)[],(address,uint256,bytes)[][3])' failure: contract call reverted with message: Some("GPv2: order filled")
    1: contract call reverted with message: Some("GPv2: order filled")

### DIFF
--- a/crates/solver/src/settlement_submission/flashbots_settlement.rs
+++ b/crates/solver/src/settlement_submission/flashbots_settlement.rs
@@ -15,7 +15,7 @@
 
 use super::{flashbots_api::FlashbotsApi, ESTIMATE_GAS_LIMIT_FACTOR};
 use crate::settlement::Settlement;
-use anyhow::{anyhow, ensure, Error, Context, Result};
+use anyhow::{anyhow, ensure, Context, Error, Result};
 use contracts::GPv2Settlement;
 use ethcontract::{transaction::Transaction, Account};
 use futures::FutureExt;
--- a/crates/solver/src/settlement_submission/flashbots_settlement.rs
+++ b/crates/solver/src/settlement_submission/flashbots_settlement.rs
@@ -15,7 +15,7 @@
 
 use super::{flashbots_api::FlashbotsApi, ESTIMATE_GAS_LIMIT_FACTOR};
 use crate::settlement::Settlement;
-use anyhow::{anyhow, ensure, Context, Result};
+use anyhow::{anyhow, ensure, Error, Context, Result};
 use contracts::GPv2Settlement;
 use ethcontract::{transaction::Transaction, Account};
 use futures::FutureExt;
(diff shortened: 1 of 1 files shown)


sunce86 (approved):
lg

bh2smith (approved):

crates/solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -258,7 +267,7 @@ impl<'a> FlashbotsSolutionSubmitter<'a> {
                         tracing::warn!("flashbots cancellation request not sent: {:?}", err);
                     }
                 }
-                return anyhow!("flashbots failed simulation: {}", err);
+                return anyhow::Error::new(err).context("flashbots failed simulation");
```
nlordell, line 270: ubernit:
```suggestion
                return err.context("flashbots failed simulation");
```

(requres:
```rust
use anyhow::Context as _;
```

crates/solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -97,10 +97,19 @@ impl<'a> FlashbotsSolutionSubmitter<'a> {
...
+                tracing::info!("stopping submission because simulation failed: {:?}", method_error);
+                Err(method_error)
+            },
+            new_nonce = nonce_future.fuse() => {
+                tracing::info!("stopping submission because account nonce changed to {}", new_nonce);
+                Ok(None)
+            },
+            _ = deadline_future.fuse() => {
+                tracing::info!("stopping submission because deadline has been reached");
+                Ok(None)
```
nlordell, line 111: Aren't deadlines errors in regular transaction submission?

nlordell (approved):

crates/solver/src/settlement_submission/flashbots_settlement.rs
fleupold: They are. Ok(None) will cause an error message. The only case where we don't error is if we return Err(e) and e can be downcasted to a method error.

crates/solver/src/settlement_submission/flashbots_settlement.rs
fleupold: I tried this but (even with the import) was getting:

> error[E0599]: no method named `context` found for struct `MethodError` in the current scope
   --> crates/solver/src/settlement_submission/flashbots_settlement.rs:271:28
    |
271 |                 return err.context("flashbots failed simulation");
    |                            ^^^^^^^ method not found in `MethodError`

I think I need to convert the MethodError specifically into an anyhow Error before using context.

PR #1385, opened 2021-11-12
Do not estimate extra prices for solver

### LINKED ISSUE #1383: Skip extra price estimations for http solver
https://gnosisinc.slack.com/archives/C022S0FM1GW/p1636642414014100

We sometimes take problematically long to get the extra price estimates for the http solver like https://github.com/gnosis/gp-v2-services/issues/1359 . These price estimates are in addition for the prices that are needed for the objective value. Those prices we already have but we might need extra prices for tokens that show up in AMM pools.

According to Marco these extra prices are not necessary, just good to have. It might be worth the tradeoff to not fetch them and have more solving time. We should implement a config option for this.

As part of this we should also investigate how long the other two IO tasks here take (ctr+f `measure_time`).

vkgnosis:
Closes https://github.com/gnosis/gp-v2-services/issues/1383.

Note that my solution still attempts to fetch the prices, but now with a time limit.

I also considered a more clever solution where would always give it at least as much time as the other two tasks but the code becomes more awkward with extra join and select so I prefer this solution.

Edit: Instead removed extra price estimates alltogether.

CC @marcovc 

### Test Plan
CI


nlordell (commented):
Hmm, can't we just estimate prices using the fetched AMM pools? The priced would then be defined by the AMM pools that they are in. For example:

Batch with
- Order trading `A` for `B`
- Base token `X`.
- AMMs with
  - Uniswap pools `A-X`, `B-X`, `A-B`
  - Weighted pool `A-B-C`

Can't we just determine the price of C using the price in the `A-B-C` pool that we already have data for instead of fetching pools between `C-X` as well? Especially since the quality of this price estimate is not important.

vkgnosis:
Based on discussion we had in weekly I removed the extra estimates completely.

nlordell (approved):

PR #1386, opened 2021-11-12
Add a competition based price estimator

nlordell:
This PR adds a new price estimator that computes price estimates using all configured estimators and chooses the one with the best price.

We define "best" price by the highest sell token price. This means that given a sell amount of `S` and buy amount of `B` we want maximize `S/B` - i.e. which estimator offers the best price to the user.

Additionally, we try to merge errors on a "best-effort" basis. This is done by considering them in a certain priority:
- `ZeroAmount`: The rational here being that this should only happen when estimating 0 amounts, so all estimators should, in theory, return this error for those queries. In practice, I don't think this path will be hit because the API filters out queries with 0 amounts anyway.
- `UnsupportedToken`: This means that one of the price estimators determined the token was unsupported. Again, in theory this should be returned by all price estimators. The reason I ranked this slightly higher than `NoLiquidity` error is that, in theory a token may be unsupported but by an external price estimator (0x for example) be reported as "no liquidity" because they don't consider any AMMs with this token because it isn't tradable.
- `NoLiquidity`: The last typed error.
- `Other`: Fallback error type. Note that we start "folding" the estimate results with an error of `Other("no successful price estimates")` so in the case all our estimators fail for some "internal" error, we will also get a generic message.

### Test Plan

Added a unit test.

Run manually and see price competition happening:
```
2021-11-16T11:29:13.141Z DEBUG shared::price_estimation::competition: received price estimate estimator_name=Baseline query=Query { sell_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, buy_token: 0xba100000625a3754423978a60c9317c58a424e3d, in_amount: 1000000000000000000, kind: Sell } estimate=Estimate { out_amount: 197195284678621485677, gas: 220417 }
2021-11-16T11:29:13.141Z DEBUG shared::price_estimation::competition: received price estimate estimator_name=ZeroEx query=Query { sell_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, buy_token: 0xba100000625a3754423978a60c9317c58a424e3d, in_amount: 1000000000000000000, kind: Sell } estimate=Estimate { out_amount: 197195284678621485677, gas: 217391 }
2021-11-16T11:29:13.141Z DEBUG shared::price_estimation::competition: winning price estimate query=Query { sell_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, buy_token: 0xba100000625a3754423978a60c9317c58a424e3d, in_amount: 1000000000000000000, kind: Sell } winning_estimate=EstimateData { estimator_name: "ZeroEx", estimate: Estimate { out_amount: 197195284678621485677, gas: 217391 }, price: Ratio { numer: 1000000000000000000, denom: 197195284678621485677 } }
```

### DIFF
--- a/crates/shared/src/price_estimation/competition.rs
+++ b/crates/shared/src/price_estimation/competition.rs
@@ -1,6 +1,8 @@
 use super::{Estimate, PriceEstimating, PriceEstimationError, Query};
 use anyhow::{anyhow, Result};
 use futures::future;
+use num::BigRational;
+use std::cmp;
 
 /// Price estimator that pulls estimates from various sources
 /// and competes on the best price.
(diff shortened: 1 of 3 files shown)


crates/shared/src/price_estimation/competition.rs
```diff
@@ -0,0 +1,148 @@
...
+        assert!(!inner.is_empty());
+        Self { inner }
+    }
+}
+
+#[async_trait::async_trait]
+impl PriceEstimating for CompetitionPriceEstimator {
+    async fn estimates(&self, queries: &[Query]) -> Vec<Result<Estimate, PriceEstimationError>> {
+        let all_estimates =
+            future::join_all(self.inner.iter().map(|(name, estimator)| async move {
```
fleupold, line 22: I assume these are already all timeouted because they use our HTTP Client? Maybe add as a consideration (not sure if adding its own timeout here is necessary)?

crates/shared/src/price_estimation/competition.rs
```diff
@@ -0,0 +1,148 @@
...
+                            None => {
+                                tracing::warn!(
+                                    estimator_name = %name, ?query, ?estimate,
+                                    "price estimate with zero amounts",
+                                );
+                                None
+                            }
+                        }
+                    })
+                    .max_by_key(|(_, _, price)| price.clone())
```
fleupold, line 55: Reading this snippet is a great example of why rust's type system is simply awesome. Forces you to think about everything that could go wrong 🤟 

In ts it would be too easy to just have a giant try catch block around all of the logic

fleupold (approved):
Price estimators going like

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2Fl0HFjeGa807RsYLu0%2Fgiphy.gif&token=competition" data-gitmeme-token="competition"></a>

vkgnosis:
Still reading code but before that I wonder if it isn't risky to take the lowest price in case some price estimator is wrong? For example if there is an independent chance of 5% that a price estimator gives a too low price then with this logic and 3 price estimators the chance that the user sees that low price (and then order probably doesn't get settled) is a lot higher (didn't do the full math).

fleupold:
> I wonder if it isn't risky to take the lowest price in case some price estimator is wrong?

That's a fair point. I hope we can soon get concrete calldata with the estimates and simulate those (to see if the estiamtes are reasonable).

We don't yet have proper tooling to measure how often this is an issue but right now I feel we run the risk of estimating for a cheap 0x trade if in reality Paraswap or 1Inch have a better but much more expensive settlement (would be good to quote the user directly the fee for the best single order settlement).

crates/shared/src/price_estimation/competition.rs
nlordell: I'll double check. Ideally we should have a `TimeBoundPriceEstimator` for estimators that don't timeout because of IO.

nlordell:
> For example if there is an independent chance of 5% that a price estimator gives a too low price

Fair point. I think the math would be `1 - (1 - 5%)**3 ~= 14.3%`.

crates/shared/src/price_estimation/competition.rs
bh2smith, line 55: So max price is our definition of "best" then?

bh2smith (approved):
Very nice. When reading the PR description is was unclear to me what was meant by best price. After reading though it seems clear that the max available price is the best. 

I wonder if this could go wrong when relying on third party estimates. Certainly couldn't be worse than what can go wrong when taking minimum estimates.

nlordell:
> After reading though it seems clear that the max available price is the best.

I updated the description.

crates/shared/src/price_estimation/competition.rs
```diff
@@ -0,0 +1,148 @@
...
+                    .ok_or_else(|| {
+                        PriceEstimationError::Other(anyhow!("no successful price estimates"))
+                    })
```
vkgnosis, line 58: Probably fine more now but this loses information in some sense by making ever PriceEstimationError into Other. For single price estimators we differentiate for example no liquidity from Other which was supposed to be something like internal error. But here even if all estimators return NoLiquidity it will return Other. This might cause the error to be error logged when it shouldn't or to give the frontend worse information on why an error occurred. Not sure how important all of these points are.

vkgnosis (approved):

crates/shared/src/price_estimation/competition.rs
nlordell: Hmm, good point. Choosing the "correct" error out of `n` errors is a tough problem. What we can do is try and find the "common" error across price estimators. For example if we have:
- `InsufficientLiquidity + InsufficientLiquidty + Other` = `InsufficientLiquidity`
- `InsufficientLiquidyt + ZeroAmount + UnsupportedToken` = `Other`

nlordell: Addressed with: https://github.com/gnosis/gp-v2-services/pull/1388

nlordell:
Merge pending on review of https://github.com/gnosis/gp-v2-services/pull/1388

PR #1387, opened 2021-11-15
Store fee parameters for created orders

vkgnosis:
This is so we can later analyze what fee related parameters where used during order creation.

Turned some fee related numbers from U256 into f64 because we were doing the conversion inside the functions anyway and this is clearer and easier to test.

### Test Plan
CI


crates/orderbook/src/database/fees.rs
```diff
@@ -139,17 +139,6 @@ mod tests {
     use model::order::OrderKind;
     use primitive_types::H160;
 
-    // Convenience to allow using u32 in tests instead of the struct
-    impl From<u32> for UnsubsidizedFee {
```
vkgnosis, line 143: Moved to the definition of UnsubsidizedFee because I found it confusing that this was in an unrelated file.

crates/orderbook/src/database/orders.rs
```diff
@@ -220,56 +220,95 @@ const ORDERS_FROM: &str = "\
...
+        .bind(fee.sell_token_price)
+        .execute(transaction)
+        .await
+        .map(|_| ())
+        .map_err(InsertionError::DbError)
+}
+
+#[async_trait::async_trait]
+impl OrderStoring for Postgres {
+    async fn insert_order(&self, order: &Order, fee: FeeParameters) -> Result<(), InsertionError> {
```
nlordell, line 299: Any reason why we don't have `FeeParameters` added as a field to `Order` struct? Not objecting to it, just wondering if you considered it if you did why you chose against it?

nlordell (approved):

crates/orderbook/src/database/orders.rs
nlordell: Nevermind. This would only make sense if we someday intended on returning this as part of an order, which AFAIK is not the plan.

crates/orderbook/src/database/orders.rs
```diff
@@ -220,56 +220,95 @@ const ORDERS_FROM: &str = "\
...
-            .bind(u256_to_big_decimal(&order.order_meta_data.full_fee_amount))
-            .execute(&self.pool)
-            .await
-            .map(|_| ())
-            .map_err(|err| {
-                if let sqlx::Error::Database(db_err) = &err {
-                    if let Some(Cow::Borrowed("23505")) = db_err.code() {
-                        return InsertionError::DuplicatedRecord;
-                    }
+    let receiver = order
```
nlordell, line 233: Just to confirm, there are no changes besides indentation here right?

vkgnosis: That's right. I recommend ticking "ignore whitespace change" in the Github UI but sadly need to do it again for every PR.

database/sql/V019__fee_parameters.sql
```diff
@@ -0,0 +1,8 @@
+-- Associate orders with the fee parameters they were evaluated against when they were created for
+-- future debugging.
+CREATE TABLE order_fee_parameters (
```
fleupold, line 3: Not related to this PR but in the future it may also be good to store the expected buy amount at order creation time. This will allow to compute how much slippage tolerance the user set and with that allow us to track us exactly how much surplus we generated from batching/CoWs (rather than assuming the price with slippage is the basis for surplus).

I'm not sure if such things would go in a separate table or if this table should become something like "order_metadata"?

fleupold (approved):
Nice, the second step will be to record the gas amount/price used when we index the trade events, right? 
Could we in that go also start storing timestamps of trade events (currently it's only block number). That would allow us to more accurately measure the mean time to settlement for each order in hindsight.

database/sql/V019__fee_parameters.sql
nlordell: I don't think we get that as part of the order creation unfortunately :cry:. Also, Its pretty easy to fake when creating an order so may not be super reliable.

Happy to add it though.

nlordell:
> Nice, the second step will be to record the gas amount/price used when we index the trade events, right?

Right! We get TX hashes, so technically the data is there, just not convenient ATM.

database/sql/V019__fee_parameters.sql
fleupold: You mean we don't perform a price estimate when adding the order because most of the time the fee measurement is cached? Yeah maybe it would be nice to add it (could come from some cache and doesn't have to be super up to date)

nlordell: Ah, that is a good point. We can add the estimated amount with the fee measurement we are already storing.

PR #1389, opened 2021-11-15
Add quasimodo price estimator

taminomara:
Adds quasimodo (or any HTTP solver, for that matter) as a price estimator.

### Test Plan
Manual testing


crates/orderbook/src/main.rs
```diff
@@ -157,6 +160,10 @@ struct Arguments {
         parse(try_from_str = parse_partner_fee_factor),
     )]
     partner_additional_fee_factors: HashMap<AppId, f64>,
+
```
josojo, line 163: Would it make sense to put it into:     shared: shared::arguments::Arguments,?

crates/orderbook/src/main.rs
```diff
@@ -413,6 +420,30 @@ async fn main() {
...
+                    QuasimodoPriceEstimator {
+                        api: Arc::new(HttpSolverApi {
+                            name: "quasimodo-price-estimator",
+                            network_name: network_name.to_string(),
+                            chain_id,
+                            base: args.quasimodo_solver_url.clone(),
+                            client: client.clone(),
+                            config: SolverConfig {
+                                api_key: None,
+                                max_nr_exec_orders: 100,
```
josojo, line 433: I think it makes no difference, but we could set the number to 1

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,191 @@
...
+        };
+
+        let orders = BTreeMap::from([(
+            0,
+            OrderModel {
+                sell_token: query.sell_token,
+                buy_token: query.buy_token,
+                sell_amount,
+                buy_amount,
+                allow_partial_fill: false,
```
josojo, line 90: do we also have the value query allow_partial_fill

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,191 @@
...
+        let settlement = self
+            .api
+            .solve(
+                &BatchAuctionModel {
+                    tokens,
+                    orders,
+                    amms,
+                    metadata: None,
+                },
+                Instant::now() + Duration::from_secs(5),
```
josojo, line 142: I would limit it to 0.5 secs. If it really takes longer, its a weird instance and it should not block quasimodo for normal instances

josojo (commented):
looks good.

Maybe you can add a testplan to the PR? I am also wondering whether more code form the estimate function can be shared with http_solver

crates/orderbook/src/main.rs
nlordell: Also, I suggest it being `Option<Url>` since its not necessary to use `Quasimodo` price estimator.

crates/shared/src/price_estimation/quasimodo.rs
nlordell: SInce this is for a price estimate, I think `allow_partial_fill: false` is correct here.

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,191 @@
...
+                allow_partial_fill: false,
+                is_sell_order: query.kind == OrderKind::Sell,
+                fee: FeeModel {
+                    amount: *GAS_PER_ORDER * gas_cost,
+                    token: self.native_token,
+                },
+                cost: CostModel {
+                    amount: *GAS_PER_ORDER * gas_cost,
+                    token: self.native_token,
+                },
```
nlordell, line 99: Neat idea to make it so the order cost doesn't matter.

Just wanted to confirm with @marcovc that its OK to have fee in a token other than the sell token.

crates/shared/src/price_estimation/quasimodo.rs
nlordell: Maybe 1s so it matches what we see with 0x and ParaSwap estimators?

nlordell (commented):
Code looks great and easy to follow.

Could you add a bit more context to the PR description? It would be nice to document some of the limitations that you observed (like missing Balancer pools and costs being off). Also, could you update the test plan with a few more details (like what manual testing process you used so we can try it ourselves)?

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,191 @@
...
+                Instant::now() + Duration::from_secs(5),
+            )
+            .await?;
+
+        if settlement.orders.is_empty() {
+            return Err(PriceEstimationError::NoLiquidity);
+        }
+
+        let mut cost = self.extract_cost(&settlement.orders[&0].cost)?;
```
nlordell, line 150: We probably need to account for additional settlement overhead. The constants we use are in:
https://github.com/gnosis/gp-v2-services/blob/main/crates/shared/src/price_estimation/gas.rs

crates/shared/src/http_solver_api/constants.rs
```diff
@@ -0,0 +1,13 @@
...
+    // Estimates from multivariate linear regression here:
+    // https://docs.google.com/spreadsheets/d/13UeUQ9DA4bHlcy9-i8d4nSLlCxSfjcXpTelvXYzyJzQ/edit?usp=sharing
+    pub static ref GAS_PER_ORDER: U256 = U256::from(66_315);
+    pub static ref GAS_PER_UNISWAP: U256 = U256::from(94_696);
+
+    // Taken from a sample of two swaps
+    // https://etherscan.io/tx/0x72d234d35fd169ef497ba0a1dc23258c96f278fb688d375d135eb012e5311009
+    // https://etherscan.io/tx/0x1c345a6da1edb2bba953685a4cf85f6a0d967ac751f8c5b518578c5fd20a7c96
+    pub static ref GAS_PER_BALANCER_SWAP: U256 = U256::from(120_000);
+}
```
nlordell, line 13: I noticed that we have some of these gas constants (with different values) in various places (for example, some more exist [here](https://github.com/gnosis/gp-v2-services/blob/main/crates/shared/src/price_estimation/gas.rs)). Maybe this is something that you can clean up in a follow up PR?

crates/shared/src/http_solver_api/constants.rs
```diff
@@ -0,0 +1,13 @@
+use ethcontract::U256;
+
+lazy_static::lazy_static! {
+    // Estimates from multivariate linear regression here:
+    // https://docs.google.com/spreadsheets/d/13UeUQ9DA4bHlcy9-i8d4nSLlCxSfjcXpTelvXYzyJzQ/edit?usp=sharing
+    pub static ref GAS_PER_ORDER: U256 = U256::from(66_315);
```
vkgnosis, line 6: Nit: I think it is slightly nicer to use built in types like u64 and convert them with `.into()` to U256 when they are used. That way there is no lazy_static. But not sure what others prefer.

crates/shared/src/price_estimation/quasimodo.rs
marcovc: Yes, it is fine.

crates/orderbook/src/main.rs
taminomara: I'm thinking if this setting should be per request?

crates/shared/src/price_estimation/quasimodo.rs
taminomara: Which constant should I use? I think `INITIALIZATION_COST + SETTLEMENT`?

nlordell: We probably need:
```
+ INITIALIZATION_COST // Call into contract
+ SETTLEMENT // overhead for entering the `settle()` function
+ 2*ERC20_TRANSFER // transfer in and transfer out
```

crates/orderbook/src/main.rs
nlordell: Since we only have 1 order (the order we are using to quote a price estimate), we can use `1` here.

THB I don't think it makes much of a difference since we have less than the maximum.

marcovc: Yes, it makes no difference.

crates/orderbook/src/main.rs
taminomara: Ok, I'll move it.

It makes sense with how we use shared arguments right now, i.e. anything that both solver and orderbook need goes to shared. I happen to have a bit different understanding of shared, i.e. I'd say that shared is anything that any service will need, be that orderbook, solver, or something else that we may write in the future. This is an offtopic discussion though, and I don't think it's a big deal.

taminomara: Oh hm, so in `orderbook` it needs to be optinal, while in `solver` it is mandatory...

taminomara:
All review suggestions are applied. To write unit tests I'll need to turn `HTTPSolverApi` into a mockable trait, and I'd prefer to make it in a separate PR.

crates/shared/src/http_solver_api.rs
```diff
@@ -87,6 +87,7 @@ impl HttpSolverApi {
         }
         let body = serde_json::to_string(&model).context("failed to encode body")?;
         tracing::trace!("request {}", body);
+        println!("{}", body);
```
nlordell, line 90: nit: Rogue print statement.
```suggestion
        println!("{}", body);
```

crates/orderbook/src/main.rs
nlordell: I think it has the same semantics as the orderbook where it is only mandatory if `Quasimodo` is used as a solver type.

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,350 @@
...
+            return Ok(Estimate {
+                out_amount: query.in_amount,
+                gas: 0.into(),
+            });
+        }
+
+        ensure_token_supported(query.buy_token, self.bad_token_detector.as_ref()).await?;
+        ensure_token_supported(query.sell_token, self.bad_token_detector.as_ref()).await?;
+
+        let gas_cost = U256::from_f64_lossy(self.gas_info.estimate().await?.legacy);
```
nlordell, line 49: Just noticed this now, this might not work as expected for EIP-1559 estimates. I suggest converting the estimate into an effective gas price:

```suggestion
        let gas_cost = U256::from_f64_lossy(self.gas_info.estimate().await?.effective_gas_price());
```

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,350 @@
...
+        tokens.insert(query.sell_token);
+        tokens.insert(query.buy_token);
+        tokens.insert(self.native_token);
+        let tokens: Vec<_> = tokens.drain().collect();
+
+        let token_infos = self.token_info.get_token_infos(&tokens).await;
+
+        let tokens = self
+            .base_tokens
+            .tokens()
```
nlordell, line 61: This should be all tokens right?
```suggestion
        let tokens = tokens.into_iter()
```

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,350 @@
...
+            .iter()
```
nlordell, line 62: This should be all tokens right?
```suggestion
        let tokens = tokens.into_iter()
```

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,350 @@
...
+                    token: self.native_token,
+                },
+                is_liquidity_order: false,
+                mandatory: true,
+            },
+        )]);
+
+        let token_pair = TokenPair::new(query.sell_token, query.buy_token).unwrap();
+        let mut pairs = self.base_tokens.relevant_pairs([token_pair].into_iter());
+        pairs.insert(token_pair);
```
nlordell, line 109: surprised this is needed :open_mouth:

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,350 @@
...
+        if settlement.orders.is_empty() {
+            return Err(PriceEstimationError::NoLiquidity);
+        }
+
+        let mut cost = self.extract_cost(&settlement.orders[&0].cost)?;
+        cost += U256::from(
+            INITIALIZATION_COST // Call into contract
+            + SETTLEMENT // overhead for entering the `settle()` function
+            + 2 * ERC20_TRANSFER, // transfer in and transfer out
+        );
```
nlordell, line 158: I think this may be mixing units because `INITIALIZATION_COST` et al. are in gas units, while the `settlement.orders[&0].cost` has already been converted into native token (i.e. Ether).

I think these need to be added after the `/ gas_cost`.

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,350 @@
...
+    use crate::token_info::TokenInfoFetcher;
+    use crate::transport::http::HttpTransport;
+    use crate::Web3;
+    use contracts::UniswapV2Factory;
+    use ethcontract::dyns::DynTransport;
+    use model::order::OrderKind;
+    use reqwest::Client;
+    use url::Url;
+
+    // TODO: to imolement these tests, we'll need to make HTTP solver API mockable.
```
nlordell, line 221: I agree this can be done in a follow up.

crates/shared/src/price_estimation/quasimodo.rs
taminomara: This is not needed, just a left over from some experiments.

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,346 @@
...
+
+        let token_infos = self.token_info.get_token_infos(&tokens).await;
+
+        let tokens = tokens
+            .iter()
+            .map(|token| {
+                (
+                    *token,
+                    TokenInfoModel {
+                        decimals: token_infos[token].decimals,
```
vkgnosis, line 65: Can't do this because get_token_infos will not have entries in hash map on errors.

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,346 @@
...
+impl PriceEstimating for QuasimodoPriceEstimator {
+    async fn estimates(
+        &self,
+        queries: &[Query],
+    ) -> Vec<anyhow::Result<Estimate, PriceEstimationError>> {
+        let mut results = Vec::with_capacity(queries.len());
+
+        for query in queries {
+            results.push(self.estimate(query).await);
+        }
```
vkgnosis, line 191: For a future PR it would be nice to move some shared computation here from the estimate function. Just like for the Baseline price estimator we can precalculate the gas price, token infos, pools and pass them in to avoid duplicate work.

crates/shared/src/price_estimation/quasimodo.rs
taminomara: Well we always return an entry: https://github.com/gnosis/gp-v2-services/blob/153217e3a3c6d0ca3967eef0b4a0e6cddb2d8489/crates/shared/src/token_info.rs#L56-L65

But you're right, we shouldn't count on implementation details.

vkgnosis: I didn't realize it would just set them to None, I see. In that case I would also be fine with changing the doc comment to mention this.

vkgnosis (approved):

fleupold (commented):
Nice! I found this PR a bit challenging to review due to its size (maybe the extensions to the HTTP solver could have been done in a preceeding PR).

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,347 @@
...
+                        (pool.tokens.get().0, pool.reserves.0.into()),
+                        (pool.tokens.get().1, pool.reserves.1.into()),
+                    ]),
+                }),
+                fee: BigRational::from((
+                    BigInt::from(*pool.fee.numer()),
+                    BigInt::from(*pool.fee.denom()),
+                )),
+                cost: CostModel {
+                    amount: U256::from(GAS_PER_UNISWAP) * gas_cost,
```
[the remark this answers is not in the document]
taminomara: They're not in this PR. This is because we have different representations for pools in solver and orderbook. I think the plan is to unify representations, and then add support for all pools in this estimator.

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,347 @@
...
+        let tokens = tokens
+            .iter()
+            .map(|token| {
+                let info = token_infos.get(token).cloned().unwrap_or_default();
+                (
+                    *token,
+                    TokenInfoModel {
+                        decimals: info.decimals,
+                        alias: info.symbol,
+                        normalize_priority: Some(if self.native_token == query.buy_token {
```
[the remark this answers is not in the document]
taminomara: ouch 🤦‍♀️ fixed

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,347 @@
...
+    async fn estimate(&self, query: &Query) -> Result<Estimate, PriceEstimationError> {
+        if query.buy_token == query.sell_token {
+            return Ok(Estimate {
+                out_amount: query.in_amount,
+                gas: 0.into(),
+            });
+        }
+
+        ensure_token_supported(query.buy_token, self.bad_token_detector.as_ref()).await?;
+        ensure_token_supported(query.sell_token, self.bad_token_detector.as_ref()).await?;
```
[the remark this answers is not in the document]
taminomara: Yes, but then all estimators that aren't wrapped in this sanitiser will become slightly more unsafe. I honestly don't know if there's a better way to do this.

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -0,0 +1,347 @@
...
+                )
+            })
+            .collect();
+
+        let (sell_amount, buy_amount) = match query.kind {
+            OrderKind::Buy => (U256::max_value(), query.in_amount),
+            OrderKind::Sell => (query.in_amount, U256::one()),
+        };
+
+        let orders = BTreeMap::from([(
```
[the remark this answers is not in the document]
taminomara: Nice to know! I've noticed, however, that type inference works differently with these two methods. Not a big deal though.

crates/shared/src/price_estimation/quasimodo.rs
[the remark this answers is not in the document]
nlordell: I think this is a neat idea for a follow up "cleanup" PR. I do think we should keep it out of here though because it would touch unrelated price estimators.

nlordell (approved):
LGTM!

Do you mind creating some Github issues with some of the follow up suggestions (like [sharing price computations](https://github.com/gnosis/gp-v2-services/pull/1389#discussion_r752065797) and [having a `SanitizingPriceEstimator` that does some common filtering](https://github.com/gnosis/gp-v2-services/pull/1389#discussion_r752099226)) so we don't lose track of it?

PR #1393, opened 2021-11-15
Allow the http solver to pass on interaction data directly

josojo:
I know that there is actually quite some work left to define the api and a fair solver competition. But for now, we bypass all of this and just allow http_solvers to do pass custom interactions to their solutions.

Maybe this PR alleviates the pain, we can allow external solvers to participate on a trust basis, and we gain some time to define the future api and competition.

### Test Plan

clone repo:
https://github.com/gnosis/dex-cow-solver
in there run: `cargo run`

then start this repo with the current PR and run:
```
cargo run -p solver -- \
  --node-url "https://mainnet.infura.io/v3/$INFURA_KEY" \
   --quasimodo-solver-url "http://127.0.0.1:8000" \
  --solver-account 0xa6DDBD0dE6B310819b49f680F65871beE85f517e \
  --solvers Quasimodo \
  --transaction-strategy DryRun
  ```
  (I am abusing quasimodo interface to talk to the new dex-cow solver)
  Observe beautiful solutions like:
  https://dashboard.tenderly.co/gp-v2/staging/simulator/d1039d01-4dbd-4ec2-9b5c-43d747a4ced9
  (quite often trivial solutions are returned, as the solver is still quite fragile)

### DIFF
--- a/crates/shared/src/http_solver_api/model.rs
+++ b/crates/shared/src/http_solver_api/model.rs
@@ -132,6 +132,14 @@ pub struct FeeModel {
     pub token: H160,
 }
 
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+pub struct InteractionData {
+    pub target: H160,
+    pub value: U256,
+    pub call_data: Vec<u8>,
+    pub exec_plan: Option<ExecutionPlanCoordinatesModel>,
+}
+
 #[serde_as]
 #[derive(Clone, Debug, Deserialize)]
 pub struct SettledBatchAuctionModel {
@@ -141,6 +149,8 @@ pub struct SettledBatchAuctionModel {
     pub ref_token: Option<H160>,
     #[serde_as(as = "HashMap<_, DecimalU256>")]
     pub prices: HashMap<H160, U256>,
+    #[serde(default)]
+    pub interaction_data: Vec<InteractionData>,
 }
 
 impl SettledBatchAuctionModel {
--- a/crates/solver/src/solver/http_solver/settlement.rs
+++ b/crates/solver/src/solver/http_solver/settlement.rs
@@ -1,9 +1,11 @@
+use crate::encoding::EncodedInteraction;
+use crate::settlement::Interaction;
 use crate::{
     liquidity::{AmmOrderExecution, LimitOrder, Liquidity},
     settlement::Settlement,
 };
 use anyhow::{anyhow, Result};
-use itertools::Itertools;
+use ethcontract::Bytes;
 use model::order::OrderKind;
 use primitive_types::{H160, U256};
 use shared::http_solver_api::model::*;
@@ -32,11 +34,27 @@ pub fn convert_settlement(
     }
 }
 
+#[derive(Clone, Debug)]
+#[cfg_attr(test, derive(PartialEq))]
+enum Execution {
+    ExecutionAmm(ExecutedAmm),
+    ExecutionCustomInteraction(InteractionData),
+}
+
+impl Execution {
+    fn coordinates(&self) -> Option<ExecutionPlanCoordinatesModel> {
+        match self {
+            Execution::ExecutionAmm(executed_amm) => executed_amm.exec_plan.clone(),
+            Execution::ExecutionCustomInteraction(interaction) => interaction.exec_plan.clone(),
+        }
+    }
+}
+
 // An intermediate representation between SettledBatchAuctionModel and Settlement useful for doing
 // the error checking up front and then working with a more convenient representation.
 struct IntermediateSettlement {
     executed_limit_orders: Vec<ExecutedLimitOrder>,
-    executed_amms: Vec<ExecutedAmm>,
+    executions: Vec<Execution>, // executions are sorted by execution coordinate.
     prices: HashMap<H160, U256>,
 }
 
@@ -61,17 +79,30 @@ struct ExecutedAmm {
     input: (H160, U256),
     output: (H160, U256),
     order: Liquidity,
+    exec_plan: Option<ExecutionPlanCoordinatesModel>,
+}
+
+impl Interaction for InteractionData {
+    fn encode(&self) -> Vec<EncodedInteraction> {
+        vec![(self.target, self.value, Bytes(self.call_data.clone()))]
+    }
 }
 
 impl IntermediateSettlement {
     fn new(settled: SettledBatchAuctionModel, context: SettlementContext) -> Result<Self> {
         let executed_limit_orders =
             match_prepared_and_settled_orders(context.orders, settled.orders)?;
-        let executed_amms = match_prepared_and_settled_amms(context.liquidity, settled.amms)?;
+        let executions_amm = match_prepared_and_settled_amms(context.liquidity, settled.amms)?;
+        let executions_interactions = settled
+            .interaction_data
+            .into_iter()
+            .map(Execution::ExecutionCustomInteraction)
+            .collect();
+        let executions = merge_and_order_executions(executions_amm, executions_interactions)?;
         let prices = match_settled_prices(executed_limit_orders.as_slice(), settled.prices)?;
         Ok(Self {
             executed_limit_orders,
-            executed_amms,
+            executions,
             prices,
         })
     }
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: crates/shared/src/http_solver_api/model.rs [hunk 1]
```diff
@@ -139,6 +147,7 @@ pub struct SettledBatchAuctionModel {
     pub ref_token: Option<H160>,
     #[serde_as(as = "HashMap<_, DecimalU256>")]
     pub prices: HashMap<H160, U256>,
+    pub interaction_data: Option<Vec<InteractionData>>,
```
vkgnosis, line 150: Nicer to define as non optional and with `#[serde(default)]` so that it becomes empty Vec is field is missing.

crates/shared/src/http_solver_api/model.rs
```diff
@@ -130,6 +130,14 @@ pub struct FeeModel {
     pub token: H160,
 }
 
+#[serde_as]
+#[derive(Clone, Debug, Deserialize)]
+pub struct InteractionData {
+    pub target: H160,
+    pub value: U256,
+    pub call_data: Bytes<Vec<u8>>,
+}
```
fleupold, line 139: While it's of course possible to encode the entire settlement as one big `interaction_data`, I wonder if it might be nice to still allow mix and match of AMM execution and additional custom calls. For this the interactions would also have to contain an execution plan coordinate and we should also append them to the encoder in their right position (with regards to other AMM interactions).

Wdyt?

josojo: I implemented your suggestion in the next commit. Please let me know whether the interface looks okay, then I will add tests and clean it up.

I tried to implement your suggestion in a non-breaking way: The `SettledBatchAuctionModel` just go the new field interactions_data. The downside is that the "coordinate" is hard to understand, as now, we have a coordinate in amms and one in interaction_data.

 We could also do, in order to make the "coordinate" of the execution clearer:
```
enum Execution {
AMMExecution(UpdatedAmmModel),
pureExecution(InteractionData)
}

#[serde_as]
#[derive(Clone, Debug, Deserialize)]
pub struct SettledBatchAuctionModel {
    pub orders: HashMap<usize, ExecutedOrderModel>,
    #[serde(default)]
    pub executions: HashMap<usize, Execution>,
    pub ref_token: Option<H160>,
    pub prices: HashMap<H160, U256>,
}

```
But since this would introduce breaking chances and mix two completely different types, I decided against it.

crates/shared/src/http_solver_api/model.rs
```diff
@@ -130,6 +130,15 @@ pub struct FeeModel {
     pub token: H160,
 }
 
+#[serde_as]
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+pub struct InteractionData {
+    pub target: H160,
+    pub value: U256,
+    pub call_data: Bytes<Vec<u8>>,
+    pub exec_plan: Option<ExecutionPlanCoordinatesModel>,
```
fleupold, line 139: I was surprised this is optional (should be at least commented). 
The reason it is optional in the other model is because we may not always be able to derive an execution plan for once we found the hollistic solution. Does this make sense here, too? Thinking about it more long term I think we will need to force solvers to come up with an execution plan (as otherwise it's hard to convert a solution into calldata deterministically)

crates/solver/src/solver/http_solver/settlement.rs
```diff
@@ -32,11 +33,18 @@ pub fn convert_settlement(
     }
 }
 
+#[derive(Clone, Debug)]
+#[cfg_attr(test, derive(PartialEq))]
+enum Execution {
```
fleupold, line 38: I like this new type

crates/solver/src/solver/http_solver/settlement.rs
```diff
@@ -127,23 +151,47 @@ fn match_prepared_and_settled_orders(
 fn match_prepared_and_settled_amms(
     prepared_amms: Vec<Liquidity>,
     settled_amms: HashMap<usize, UpdatedAmmModel>,
-) -> Result<Vec<ExecutedAmm>> {
+) -> Result<Vec<(Option<ExecutionPlanCoordinatesModel>, Execution)>> {
```
fleupold, line 154: Rather than making this type more complex, how about implementing a getter for the coordinates on the Execution enum directly? This way you can just call `execution.coordinates()` everywhere. You could even implement `PartialOrd` on `Execution` which would make sorting even more simple (although I don't mind the explicitness and have been overly excited in the past to shoehorn logic into existing traits).

crates/solver/src/solver/http_solver/settlement.rs
```diff
@@ -127,23 +151,47 @@ fn match_prepared_and_settled_orders(
...
+                    input: (settled.buy_token, settled.exec_buy_amount),
+                    output: (settled.sell_token, settled.exec_sell_amount),
+                }),
+            ))
         })
-        .collect()
+        .collect::<Result<Vec<(Option<ExecutionPlanCoordinatesModel>, Execution)>>>()
+}
+
+fn merge_and_order_executions(
```
fleupold, line 175: Would be nice to explicitly say (in a comment) what happens with executions that don't have coordinates (do they go first or last).

fleupold (commented):
Generally I like this very much! I think we can leverage the new type a bit more to simplify the ordering logic and not map reduce so much.

crates/shared/src/http_solver_api/model.rs
fleupold: Would still like to get an answer to this comment.

crates/solver/src/solver/http_solver/settlement.rs
```diff
@@ -32,11 +33,27 @@ pub fn convert_settlement(
...
+        }
+    }
+}
+
 // An intermediate representation between SettledBatchAuctionModel and Settlement useful for doing
 // the error checking up front and then working with a more convenient representation.
 struct IntermediateSettlement {
     executed_limit_orders: Vec<ExecutedLimitOrder>,
-    executed_amms: Vec<ExecutedAmm>,
+    executions: Vec<Execution>,
```
fleupold, line 56: nit: maybe there should be a comment here saying that this is already sorted by execution coordinate (otherwise it would be nice to re-sort before actually appending it to the execution plan).

fleupold (approved):
Can't wait to see the new solver in action!

crates/shared/src/http_solver_api/model.rs
josojo: I agree that long-term we want to have this non-optional. But in the current code, I prefer to keep it the same as for the usual http solver. Otherwise, the asymmetric handling of option and non-option makes the code look quite bad.
How about: I make a follow-up PR that enforces it for both alternatives, the execution plan is non-optional. 

>The reason it is optional in the other model is because we may not always be able to derive an execution plan for once we found the hollistic solution

I even don't really fully understand the optionality in this case: If the solver can not come up with an execution plan, will the driver be able to do it? This work should definitively happen on the solvers side.

crates/shared/src/http_solver_api/model.rs
```diff
@@ -132,6 +132,15 @@ pub struct FeeModel {
     pub token: H160,
 }
 
+#[serde_as]
```
vkgnosis, line 135: Not needed?

crates/shared/src/http_solver_api/model.rs
```diff
@@ -132,6 +132,15 @@ pub struct FeeModel {
...
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+pub struct InteractionData {
+    pub target: H160,
+    pub value: U256,
+    pub call_data: Bytes<Vec<u8>>,
```
vkgnosis, line 140: The Bytes type is not right. This more of an implementation detail in ethcontract but here it matters for serialization. It can be plain Vec<u8> in which case the it would be formatted as a json list like [1,255,4,3].

vkgnosis (approved):

josojo:
Thanks a lot for the review @Felix, I love the `coordinates` suggestion. It made the code so much more aesthetic and easier to digest. Love it!

crates/shared/src/http_solver_api/model.rs
fleupold: Maybe @marcovc can weigh in here how realistic it is that the solver doesn't have an execution plan for AMMs.

No need to change if you prefer the symmetry, however I think that thanks to the `coordinates()` abstraction (which may continue to return an Option) you could just return `Some(interaction_data.exect_plan)`. I don't see how it would make the code look bad.

marcovc: Hi! We can make it non-optional (we will just return the trivial solution if for some reason we can't come up with an execution plan). The reason why is optional is that, for debugging issues in computing the execution plan, it is helpful to still see the solution to the problem.

nlordell (approved):

PR #1395, opened 2021-11-16
[Easy] Combine printing all rated settlements

fleupold:
Because this makes tailing our instance logs easier to read

### Test Plan
Before:
image

After:
>2021-11-16T17:44:31.015Z  INFO solver::driver: Rated Settlements:
Objective value for solver 0x is 9.22e17: surplus=7.47e16, gas_estimate=3.32e5, gas_price=1.18e11
Objective value for solver ParaSwap is 9.15e17: surplus=1.20e17, gas_estimate=1.67e6, gas_price=1.18e11
Objective value for solver ParaSwap is 9.03e17: surplus=7.47e16, gas_estimate=4.97e5, gas_price=1.18e11
Objective value for solver ParaSwap is 2.43e16: surplus=3.85e16, gas_estimate=3.48e5, gas_price=1.18e11
Objective value for solver BaselineSolver is 1.32e16: surplus=1.58e16, gas_estimate=2.51e5, gas_price=1.18e11
Objective value for solver 0x is 2.72e15: surplus=6.23e15, gas_estimate=2.88e5, gas_price=1.18e11
Objective value for solver ParaSwap is 1.43e15: surplus=6.23e15, gas_estimate=2.99e5, gas_price=1.18e11
Objective value for solver BaselineSolver is 1.53e14: surplus=6.23e15, gas_estimate=3.10e5, gas_price=1.18e11
Objective value for solver BaselineSolver is -8.52e15: surplus=2.27e16, gas_estimate=1.15e6, gas_price=1.18e11
Objective value for solver BaselineSolver is -3.19e16: surplus=2.37e14, gas_estimate=4.72e5, gas_price=1.18e11
Objective value for solver 0x is -3.20e16: surplus=2.37e14, gas_estimate=4.73e5, gas_price=1.18e11
Objective value for solver ParaSwap is -3.31e16: surplus=2.37e14, gas_estimate=4.82e5, gas_price=1.18e11


nlordell:
Not a strong objection, but I have a couple issues with this:
- It breaks `kubectl logs | grep`-ability (because they are split into multiple lines)
- It is now not possible to search "all objective values for Quasimodo solver" without also getting a lot of noise.

Instead, could we log the `auction.id` with this statement, so we can just filter by that with something like in KQL:
```
log: "Objective value for solver" and log: "auction_id=42"
```

fleupold:
My main reason for this change was that while I would like to see which solvers created solution in a given runloop I would like to have a way to easily collapse that information (and have it ordered by objective value already).

I understand the concern about grep-ability, but don't we already have a bunch multiline print statements? How does that work? That being said, I think it would be fine to just comma separate all values here as well.

> all objective values for Quasimodo solver

I have never found myself doing that. Maybe we can debug log each individual solution by itself if this is a query people are interested in?

nlordell (approved):
> I understand the concern about grep-ability, but don't we already have a bunch multiline print statements?

We got rid of some of the bigger offenders (batch calls come to mind). Here is a related issue from a while back: https://github.com/gnosis/gp-v2-services/issues/669.

> and have it ordered by objective value already

That is **super** convenient! I really like this aspect of it :smile:. In theory though, you can just order the settlements before logging their objective values though to get this nice feature.

> I think it would be fine to just comma separate all values here as well.

Are you concerned about it being less readable? Feels like a sub-optimal compromise. Another suggestion would be to split this into:
```
tracing::info!("winning settlement ...");
tracing::debug!("runner ups ..."); // comma separated and ordered by objective value
```

>  by itself if this is a query people are interested in?

It was just a convoluted example, the point I was trying to make is that grouping this data into a single log document removes some of the search-ability/query-power (both with `grep` and Kibana KQL) that you get from keeping them separate. Another example off the top of my head: compare objective value between 0x and ParaSwap solvers: `log: "Objective value for solver 0x" or log: "Objective value for solver ParaSwap"`.

On the other hand, I do see the appeal when following the logs, where looking at all the objective values is not super relevant (but you want to be able to peak at the list for certain batches). If this helps debugging I definitely don't want to block this getting merged.

bh2smith (approved):

fleupold:
>  In theory though, you can just order the settlements before logging their objective values though to get this nice feature.

Kibana reorders arbitrarily by timestamp so even if we order and print in order chances are that some statements end up with the exact same timestamp and therefore get messed up again in kibana.

To conclude, I'd say a comma separated log with all settlements is fine for the info log. I'm happy to add individual debug logs if you feel like it would help the use cases you described.

vkgnosis (approved):

josojo (approved):

PR #1397, opened 2021-11-17
Remove `get_spot_price` method

nlordell:
This PR removes the old `get_spot_price` method from `BaselineSolvable` since it:
- Isn't used anymore
- Isn't implemented for all kinds of liquidity anyway

Because of the magic of Git, we can always bring it back if needed.

### Test Plan

CI.


vkgnosis (approved):

josojo (approved):

bh2smith:
Does this mean I no longer need to continue my research implementing it for stable pools?

PR #1399, opened 2021-11-17
flashbots endpoint changed to https://rpc.flashbots.net

sunce86:
In this PR we change the flashbots endpoint from `https://protection.flashbots.net/v1/rpc` to `https://rpc.flashbots.net`

Lots of code is removed since its not needed for the new endpoint, where the cancellation() and gasFees() functions are not supported. Basically, on this endpoint multiple transactions can be sent from the the same sender and replacement logic is done on the flashbots side. If transaction is mined, all previous transactions will fail inclusion since nonce is changed.

Since cancellation does not exist, we need to monitor what happens in two cases:
1. Deadline is reached (our deadline is 120s while flashbots is 25blocks ~ 375s)
2. Simulation fails (our logic returns back while flashbots continue to try with last submitted transaction even if it first fails, but becomes valid again after a few blocks)

GasPrice related, our minimum is set to 3 gwei as I've noticed significant delay at 1+ gwei.

Test plan
- [x] Do a manual test on local

### DIFF
--- a/crates/solver/src/settlement_submission/flashbots_settlement.rs
+++ b/crates/solver/src/settlement_submission/flashbots_settlement.rs
@@ -202,7 +202,7 @@ impl<'a> FlashbotsSolutionSubmitter<'a> {
         transactions: &mut Vec<H256>,
     ) -> anyhow::Error {
         const UPDATE_INTERVAL: Duration = Duration::from_secs(5);
-        const MINIMAL_MAX_PRIORITY_FEE: f64 = 10_000_000_000.0;
+        const MINIMAL_MAX_PRIORITY_FEE: f64 = 3_010_000_000.0;
 
         // The amount of extra gas it costs to include the payment to block.coinbase interaction in
         // an existing settlement.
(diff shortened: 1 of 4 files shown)


crates/solver/src/settlement_submission/flashbots_settlement.rs
```diff
@@ -202,7 +202,7 @@ impl<'a> FlashbotsSolutionSubmitter<'a> {
         transactions: &mut Vec<H256>,
     ) -> anyhow::Error {
         const UPDATE_INTERVAL: Duration = Duration::from_secs(5);
-        const MINIMAL_MAX_PRIORITY_FEE: f64 = 10_000_000_000.0;
+        const MINIMAL_MAX_PRIORITY_FEE: f64 = 3_010_000_000.0;
```
nlordell, line 205: Would it be possible to make this configurable? We can even configure it to be an additional Flashbot tip in GWei:
```
--additional-flashbot-tip 3.0
```

nlordell (approved):
LGTM.

Not blocking merge on priority fee configuration.

vkgnosis (approved):

crates/solver/src/settlement_submission/flashbots_settlement.rs
vkgnosis: I think this would be good. So we don't have to rebuild if we see TXs not getting mined because of this.

sunce86: Sure, added.

PR #1400, opened 2021-11-17
Fix not deleting old events after reorg

vkgnosis:
This was discovered when a user reported [an order](https://gnosis-protocol.io/orders/0xc63389b7a520f414c28535bb67313f44db452947f2ca35e9651facf7460d6082b52bcc47404d7860e58ac67436156cb20b8ebd8b61928c2c) that shows up as settled in the explorer even though the transaction link shows it reverted.
I tracked this down with our logs and found that the transaction was originally mined in block 13621398 which got replaced later. After the replacement no new events occurred for a while which triggered a bug in our event handling code causing the reverted event to never get deleted.

### Test Plan
Can't easily test right now but maybe with a change to the traits here to make them mockable. Logic change is simple so I want to do that first because we will do it anyway and maybe then try to add the test in second PR.

### DIFF
--- a/crates/shared/src/event_handling.rs
+++ b/crates/shared/src/event_handling.rs
@@ -108,7 +108,7 @@ where
             .past_events(&range)
             .await
             .context("failed to get past events")?
-            .ready_chunks(INSERT_EVENT_BATCH_SIZE)
+            .chunks(INSERT_EVENT_BATCH_SIZE)
             .map(|chunk| chunk.into_iter().collect::<Result<Vec<_>, _>>());
         futures::pin_mut!(events);
         // We intentionally do not go with the obvious approach of deleting old events first and
@@ -118,6 +118,8 @@ where
         // 1. It ensures that we only delete if we really have new events. Otherwise if fetching new
         //    events from the node fails for whatever reason we might keep deleting events over and
         //    over without inserting new ones resulting in the database table getting cleared.
+        //    Note that we do want to delete events if the new events are empty while fetching was
+        //    successful.
         // 2. It ensures that other users of the database are unlikely to see an inconsistent state
         //    some events have been deleted but new ones not yet inserted. This is important in case
         //    for example another part of the code calculates the total executed amount of an order.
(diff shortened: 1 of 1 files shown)


nlordell (approved):

nlordell:
Nice detective work.

crates/shared/src/event_handling.rs
```diff
@@ -108,7 +108,7 @@ where
             .past_events(&range)
             .await
             .context("failed to get past events")?
-            .ready_chunks(INSERT_EVENT_BATCH_SIZE)
+            .chunks(INSERT_EVENT_BATCH_SIZE)
```
fleupold, line 111: is this change because ready_chunks may yield an empty vector before the stream is over?

crates/shared/src/event_handling.rs
```diff
@@ -133,6 +135,8 @@ where
         // in one transaction.
         let mut have_deleted_old_events = false;
         while let Some(events_chunk) = events.next().await {
+            // Early return on error is important here so that the second !have_deleted_old_events
```
fleupold, line 138: nanonit: I had a hard time finding the return statement (it's part of the ?). If it's so important maybe explitly return?

fleupold (approved):
impressive debugging!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia4.giphy.com%2Fmedia%2Fl4JyTxAx1jFvzKKaY%2Fgiphy.gif&token=sherlock" data-gitmeme-token="sherlock"></a>

crates/shared/src/event_handling.rs
vkgnosis: Oh, forgot to explain this. ready_chunk is different because it will yield chunks when the underlying stream would be pending even if the chunk is not full yet. Neither will yield empty chunks and if they would it wouldn't be an issue. I just found the ready_chunk behavior a little surprising so I moved to chunks as it is closer to what this function on normal iterator does and I think this is what we intended.

crates/shared/src/event_handling.rs
vkgnosis: Will change comment to mention the `?` but I'd like to keep it.

PR #1403, opened 2021-11-18
Separate uniswap like protocols

nlordell:
Part of a larger refactoring around baseline liquidity in the project, with the ultimate goal of enabling baseline trade simulations.

This PR refactors Uniswap-like liquidity and splits each one into its own module. The reason for the split is to ultimately end up with something like:
```rust
enum BaselinLiquidity {
    UniswapV2(uniswap_v2::Pool),
    UniswapV3(uniswap_v3::Pool),
    Honeyswap(honeyswap::Pool),
    SushiSwap(sushiswap::Pool),
    BalancerV2(balancer_v2::Pool),
    ZeroEx(zeroex::RfqOrder),
}
```

The plan is for equivalent liquidity (like UniswapV2, Honeyswap, and SushiSwap) to share types and functions between them to avoid code duplication. Something like
```rust
mod honeyswap {
    pub type Pool = uniswap_v2::Pool;
}
```

This refactor in particular also gets rid of the `AmmPairProvider` trait since it:
- was specific to Uniswap and not AMMs.
- didn't need to be a trait and was replaced with the `PairProvider` struct
- each baseline source module implements its own creation method for getting its `pair_provider` (i.e. `sushiswap::get_pair_provider(&web3)`).

Additionally, we split `UniswapV2` into `UniswapV2` and `Honeyswap`. This will allow us to quite trivially add a `Baoswap` to improve xDai liquidity moving forward.

### Test Plan

Refactoring with no logic changes, so CI and existing tests.


crates/contracts/artifacts/SushiSwapFactory.json
```diff
@@ -0,0 +1 @@
+IUniswapLikeFactory.json
```
nlordell, line 1: Note that this was previously linked to `UniswapV2Factory`. This was incorrect because it **did not** have the same bytecode, so if we started deploying SushiSwap factories in our E2E tests, the addresses retrieved by the `PairProvider` would have been wrong because the `INIT_CODE_DIGEST` would not match.

fleupold (approved):
Makes a lot of sense

> This will allow us to quite trivially add a Baoswap to improve xDai liquidity moving forward.

Awesome!

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia4.giphy.com%2Fmedia%2FQWXlB6E4YZsCYC870d%2Fgiphy.gif&token=rebuild" data-gitmeme-token="rebuild"></a>

bh2smith (approved):
❤️

vkgnosis (approved):

Issue #1253, opened 2021-10-13
API Price Estimates Can Lead To Untradable Orders


nlordell:
It is my understanding, that adding ParaSwap price estimator to the API increases the amount of tokens the API supports by taking advantage of more liquidity.

The issue, however, is that the the solution objective value is computed using baseline price estimates, therefore (at least to my understanding), it would ignore solutions that trade tokens for which the baseline price estimator cannot find a path.

This might cause some UX issues where a user is able to create an order, but we can never execute the trade because there is no price estimate for the objective value computation.

fleupold:
I think the constraints in the driver are still a little bit less strict as for the API. E.g. I noticed stkAave is still tradable (even if it wasn't until we enabled it by using Paraswap in the API). The reason seems to be that in order to price a token we only take the current spot price (regardless of pool size), whereas in the API we want to be able to trade at least 1ETH worth (which e.g. on the stkAave uni v2 pool is not possible despite it existing).

PR #1408, opened 2021-11-18
Augment price estimates with limit price if no other orders in the book

fleupold:
Hacky solution to #1253 

If we cannot estimate a price, but only have one token pair in the batch, the price estimate doesn't matter much for solution ranking (only with regards to the gas price tradeoff). It's therefore acceptable to estimate the price based on the user's limit price (average over all buy/sellAmount ratios). It allows trading tokens such as LBP pool tokens or other previously supported but later pruned tokens.

I realised that with increasing usage the chances of such an order being the only one in the batch are getting quite small (I wasn't able to run my shadow test in prod). Maybe it's actually not worth introducing this hacky workaround. Still, since the code is there, I thought I'd put it up for discussion.

The proper fix would likely be to pass down price estimates for all relevant tokens from the API in `getSolvableOrders` (or `getBatch`).

### Test Plan
Unit test + placing [this order](https://protocol-explorer.dev.gnosisdev.com/orders/0xe44b6ea6b2b1e956de18b95da0433e5f3fc5c2fb2928e3764d2911dd137423342caef7f0ee82fb0abf1ab0dcd3a093803002e70561966062) via barn which expired (due to pruning).

However, locally with the shadow solver I was able to get solutions:

> 2021-11-18T13:51:54.265Z  INFO solver::driver: got 1 orders: [Limit Order 0xe44b6ea6b2b1e956de18b95da0433e5f3fc5c2fb2928e3764d2911dd137423342caef7f0ee82fb0abf1ab0dcd3a093803002e70561966062]
2021-11-18T13:51:54.266Z  WARN solver::driver: failed to estimate price for token 0xa36f…1ab9: NoLiquidity
2021-11-18T13:51:55.325Z  INFO solver::driver: Objective value for solver 0x is 7.63e14: surplus=5.79e13, gas_estimate=5.19e5, gas_price=1.14e11
2021-11-18T13:51:55.325Z  INFO solver::driver: 1 settlements passed simulation and 0 failed
2021-11-18T13:51:55.362Z  INFO solver::driver: winning settlement by 0x: RatedSettlement { settlement: Settlement { encoder: SettlementEncoder { tokens: [0xa36fdbbae3c9d55a1d67ee5821d53b50b63a1ab9, 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2], clearing_prices: {0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2: 54799472358728689347, 0xa36fdbbae3c9d55a1d67ee5821d53b50b63a1ab9: 5002420295039784}, trades: [Trade { order: Order { order_meta_data: OrderMetaData { creation_date: 2021-11-18T13:47:10.355676Z, owner: 0x2caef7f0ee82fb0abf1ab0dcd3a093803002e705, uid: 0xe44b6ea6b2b1e956de18b95da0433e5f3fc5c2fb2928e3764d2911dd137423342caef7f0ee82fb0abf1ab0dcd3a093803002e70561966062, available_balance: Some(1650094756984114697), executed_buy_amount: 0, executed_sell_amount: 0, executed_sell_amount_before_fees: 0, executed_fee_amount: 0, invalidated: false, status: Open, settlement_contract: 0x9008d19f58aabd9ed0d60971565aa8510560ab41, full_fee_amount: 59829815113289128 }, order_creation: OrderCreation { sell_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, buy_token: 0xa36fdbbae3c9d55a1d67ee5821d53b50b63a1ab9, receiver: Some(0x2caef7f0ee82fb0abf1ab0dcd3a093803002e705), sell_amount: 5002420295039784, buy_amount: 54172727219742279077, valid_to: 1637245026, app_data: 0xe4d1ab10f2c9ffe7bdd23c315b03f18cff90888d6b2bb5022bacd46ab9cddf24, fee_amount: 44997579704960216, kind: Sell, partially_fillable: false, signature: Eip712(EcdsaSignature { r: 0x98403ba39a7a465f82132854fc3fcbc22f294b37f1d564d818ad7c15e6062aa8, s: 0x44e8eb2e78f0ff2b4687aa70360e43d7dafbeb5be3b2c9d53c036ff36790bf44, v: 28 }), sell_token_balance: Erc20, buy_token_balance: Erc20 } }, sell_token_index: 1, buy_token_index: 0, executed_amount: 5002420295039784, scaled_fee_amount: 59829815113289128, is_liquidity_order: false }], execution_plan: [AllowanceSufficient, SwapResponse { price: PriceResponse { sell_amount: 5002420295039784, buy_amount: 54799472358728689347, allowance_target: 0xdef1c0ded9bec7f1a1670819833240f027b25eff, price: 10954.591802905052, estimated_gas: 296000 }, to: 0xdef1c0ded9bec7f1a1670819833240f027b25eff, data: 0x..., value: 0 }], unwraps: [] } }, surplus: Ratio { numer: 3135242603083148892453864996181680, denom: 54172727219742279077 }, solver_fees: Ratio { numer: 59829815113289128, denom: 1 }, gas_estimate: 519244, gas_price: Ratio { numer: 466398203650241, denom: 4096 } }
2021-11-18T13:51:55.424Z  INFO solver::settlement_submission::dry_run: not submitting transaction in dry-run mode


crates/solver/src/driver.rs
```diff
@@ -630,6 +634,74 @@ fn orders_with_price_estimates(
     orders
 }
 
+// In case we cannot estimate the price for a token but there is only one pair traded in the batch the accuracy of the price estimate for ranking a solution doesn't matter that much (surplus is not competing with other tokens).
+// Therefore we can deduct a price estimate by looking at the average limit price of the orders on the single token pair
+fn augment_prices(orders: &[LimitOrder], prices: &mut HashMap<H160, BigRational>) {
+    if orders.is_empty() {
+        return;
+    }
+    let first_order = orders.first().unwrap();
```
nlordell, line 643: ubernit: A slightly more Rust-y way to do this IMO in order to avoid `unwrap` is:
```suggestion
    let first_order = match orders.first() {
        Some(order) => order,
        None => return,
    };
```

crates/solver/src/driver.rs
```diff
@@ -630,6 +634,74 @@ fn orders_with_price_estimates(
...
+    let init = (
+        (first_order.sell_token, U256::zero()),
+        (first_order.buy_token, U256::zero()),
+    );
+    let only_pair_with_amounts = orders.iter().try_fold(
+        init,
+        |(first_token_and_amount, second_token_and_amount), order| {
```
nlordell, line 650: ubernit: You can double de-structure here to avoid all the `.0` and `.1`:
```suggestion
        |((first_token, first_amount), (second_token, second_amount)), order| {
```

If you you find the line too long you can also split it:
```suggestion
        |(first_token_and_amount, second_token_and_amount), order| {
            let (first_token, first_amount) = first_token_and_amount;
            let (second_token, second_amount) = second_token_and_amount;
```

crates/solver/src/driver.rs
```diff
@@ -630,6 +634,74 @@ fn orders_with_price_estimates(
     orders
 }
 
+// In case we cannot estimate the price for a token but there is only one pair traded in the batch the accuracy of the price estimate for ranking a solution doesn't matter that much (surplus is not competing with other tokens).
```
nlordell, line 637: nit: Because this module is already super long, do you mind creating a separate modules for this? Maybe something like `auction_preprocessing` or something?

nlordell (approved):
This is incredibly cleaver! I had thought of doing this a different way which was way more complicated (skipping objective value computation under certain conditions and just comparing prices). Awesome!

crates/solver/src/driver.rs
taminomara: Or we can accept a reference to a single order for this function, and do `if let` at the callsite.

taminomara: Oh no we can't, never mind.

taminomara (approved):
The code looks clear and good =)

crates/solver/src/driver.rs
taminomara: I'm fine with either way. I agree that the driver file becomes too long, but I feel that we may need to rethink how we divide it into submodules. Right now it is a little bit messy. We put some seemingly random things into submodules while keeping others in place, so the code becomes hard to follow. This is a separate task though, of course.

nlordell: I also agree we should be careful about sprinkling things in submodules. My idea was to have a general "auction preoprocessing" module that we pass the `&mut Auction` in and do things like `augment_prices`.

Agree that maybe we can move some other chunks of code in there (so related logic doesn't get sprinkled around). For example, I think we have a `batch_reporting` module for reporting metrics post-execution, but also do some of it directly in the `driver` module.

vkgnosis (approved):

Issue #1410, opened 2021-11-18
Tests for quasimodo price estimator


taminomara:
We'll need to make HTTP solver API mockable before writing them.

PR #1415, opened 2021-11-18
Make HTTP solver API mockable

taminomara:
Preparation for #1410.

We're turning HTTP solver into a trait so that we can mock it.

### Test Plan
No logic changes were made, so existing tests should cover everything.


sunce86 (approved):
Lg

nlordell (approved):
Great!

PR #1416, opened 2021-11-19
Round up in fee estimation

vkgnosis:
And add some fee estimation debug traces.

I decided to go back to U256 instead of f64 in some interfaces so that the rounding is an implementation detail that api users do not have to think about.

### Test Plan
new unit test that highlights the issue


bh2smith (approved):

nlordell (approved):
LGTM

PR #1419, opened 2021-11-22
Surplus Reporting: fix bucket definition

bh2smith:
While attempting to understand the heat map generated from the best alternative surplus reporting, @vkgnosis and I concluded that the buckets weren't defined correctly. This PR changes the bucket definitions to the true upper and lower bounds of the calculation excluding infinities because grafana would include them anyway if necessary.

Note the additional comment that +/- 2 is indeed the max upper and lower bound of the calculation performed when taking the difference of two percentages.

Furthermore, we noticed that the reason the y-axis did not make any sense here (i.e. did not appear to correspond to the bucket definitions themselves)
Screenshot 2021-11-22 at 11 59 59 AM

If because we need to select time series buckets in the visualization:

Screenshot 2021-11-22 at 12 00 09 PM

### DIFF
--- a/crates/solver/src/metrics.rs
+++ b/crates/solver/src/metrics.rs
@@ -132,7 +132,8 @@ impl Metrics {
                 "Surplus ratio differences between winning and best settlement per order",
             )
             .buckets(vec![
-                -f64::INFINITY,
+                -2.0,  // Difference of percentages can not exceed +/- 2
+                -1.0,
                 -0.1,
                 -0.01,
                 -0.005,
@@ -140,7 +141,8 @@ impl Metrics {
                 0.005,
                 0.01,
                 0.1,
-                f64::INFINITY,
+                1.0,
+                2.0,
             ]),
         )?;
         registry.register(Box::new(order_surplus_report.clone()))?;
--- a/crates/solver/src/metrics.rs
+++ b/crates/solver/src/metrics.rs
@@ -132,17 +132,8 @@ impl Metrics {
                 "Surplus ratio differences between winning and best settlement per order",
             )
             .buckets(vec![
-                -2.0,  // Difference of percentages can not exceed +/- 2
-                -1.0,
-                -0.1,
-                -0.01,
-                -0.005,
-                0.,
-                0.005,
-                0.01,
-                0.1,
-                1.0,
-                2.0,
+                -2.0, // Difference of percentages can not exceed +/- 2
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):

nlordell:
> Difference of percentages can not exceed +/- 2

Wait, I thought we were computing the difference in surplus ratios. If that is the case, I don't think this is correct!

Imagine a user places a market order, in that case we would have a surplus of 10000% and the difference can be much higher than 2.

bh2smith:
The Surplus Ratios are the percentages more than the limit. They are percentages (i.e. numbers between 0 and 1)

nlordell:
> Although I guess technically it is possible to get 300% more than expected... so I suppose they can be more..

Exactly my point.

Consider a limit order:
- Sell 1 ETH for at least 0.000000000000000001 DAI (i.e. a "true" market order).

Right now you would get ~4000 DAI, i.e a surplus ratio of 4e21, which is `>>> 2`.

bh2smith:
Apologies for merging so early. However, I do believe these upper and lower bounds to be correct however the comment is not. Since Grafana adds its own +/ infinity.

If a surplus ratio difference winds up being beyond +/- 2 it would fall into an extreme outlier bucket.

nlordell:
[According to the docs](https://docs.rs/prometheus/0.13.0/prometheus/struct.HistogramOpts.html#structfield.buckets), it only says it adds one of the "extreme outlier" buckets:

> There is no need to add a highest bucket with +Inf bound

bh2smith:
We can discuss the bounds in this new PR #1420

bh2smith:
Yes (-inf, -2) is covered by the current bucket definition.

crates/solver/src/metrics.rs
```diff
@@ -132,15 +132,8 @@ impl Metrics {
...
-                -f64::INFINITY,
-                -0.1,
-                -0.01,
-                -0.005,
-                0.,
-                0.005,
-                0.01,
-                0.1,
-                f64::INFINITY,
+                -2.0, // Difference of percentages can not exceed +/- 2
```
fleupold, line 135: Why not? Isn't the difference between 0 and 100% more than two? I guess I don't really understand what exactly we are measuring here.

fleupold (commented):
It would be great if the metric name could be self-explanatory in what exactly it's measuring. Something like `pct_surplus_difference_between_best_and_winning`

PR #1420, opened 2021-11-22
Incorrect comment about bucket bounds

bh2smith:
Based on an issue brought up with a prematurely merged branch #1419

https://github.com/gnosis/gp-v2-services/pull/1419#issuecomment-975424217

We remove this false comment. Can also discuss what bounds we actually want to use here (cc @nlordell )

### DIFF
--- a/crates/solver/src/metrics.rs
+++ b/crates/solver/src/metrics.rs
@@ -131,9 +131,7 @@ impl Metrics {
                 "settlement_surplus_report",
                 "Surplus ratio differences between winning and best settlement per order",
             )
-            .buckets(vec![
-                -2.0, -1.0, -0.1, -0.01, -0.005, 0., 0.005, 0.01, 0.1, 1.0, 2.0,
-            ]),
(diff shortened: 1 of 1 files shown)


crates/solver/src/metrics.rs
```diff
@@ -132,8 +132,7 @@ impl Metrics {
                 "Surplus ratio differences between winning and best settlement per order",
             )
             .buckets(vec![
-                -2.0, // Difference of percentages can not exceed +/- 2
-                -1.0, -0.1, -0.01, -0.005, 0., 0.005, 0.01, 0.1, 1.0, 2.0,
+                -2.0, -1.0, -0.1, -0.01, -0.005, 0., 0.005, 0.01, 0.1, 1.0, 2.0,
```
nlordell, line 135: Honestly, I don't see much value in separating `>200% surplus difference ; 200%-100% surplus difference` and `>100% surplus difference`.

Thanks for fixing the comment.

bh2smith: I can also remove the +/- 2 (they were really only introduced when I thought they were the absolute bounds). will remove.

vkgnosis (approved):

nlordell (approved):

PR #1422, opened 2021-11-22
[Easy] Simple Metric reporting refactor

bh2smith:
Based on a TODO left in #1230 - we move the metric reporting logic of "matched but unsettled orders" to the recently introduced file `analytics.rs`. 

No logical changes made here, just moving data manipulation and computation outside of `driver.rs`. Looking for advices on iterating over the same list twice.


bh2smith:
Superseded by #1423

PR #1423, opened 2021-11-22
Log orders that have solutions but don't get settled

nlordell:
This PR adds an additional log message printing UIDs for orders that have settlements but don't end up getting settled on chain.

This is because we have an alert:
> Insufficient Settlement Throughput

On this metric, but it is hard to debug exactly what is going on since we can't actually easily read which orders are being "left behind" this way.

### Test Plan

Rust compiler (code moved to a separate module and we added a log statement, no new logic).

### DIFF
--- a/crates/solver/src/analytics.rs
+++ b/crates/solver/src/analytics.rs
@@ -1,13 +1,47 @@
-use crate::{metrics::SolverMetrics, settlement::Settlement};
+use crate::{
+    driver::solver_settlements::RatedSettlement, metrics::SolverMetrics, settlement::Settlement,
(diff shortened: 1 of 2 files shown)


vkgnosis (approved):

bh2smith (approved):

bh2smith:
Looks like this is based on my beach but directed towards main.

nlordell:
> Looks like this is based on my beach but directed towards main.

No, sorry. @taminomara and I did this independently when debugging an alert. I didn't see your PR.

bh2smith:
Oh.

PR #1430, opened 2021-11-23
Fixed issue where all Balancer pools were getting removed on startup

nlordell:
This PR addresses an issue where Balancer pools were getting removed on startup. This is because the Balancer pool store initializes with the Graph for some re-org safe block in the past pretending that all pools were added by events on that block. The event handler was replacing all events in that (this is needed because of subtle race conditions that can happen) with an empty list - causing all Balancer pools to be removed.

The fix is to track the block for which the initial pools were fetched on and just not delete events past that point. Since we use a re-org safe block to fetch these pools, there should not be any problems where a re-orgs causes these pools to no longer exist.

Additionally, this did not seem to affect Stable pools, which led me to find a separate issue where stable pools were being indexed after startup.

### Test Plan

CI still passes. Also, you can run the solver locally and see:
```
$ cargo run -p solver -- --baseline-sources BalancerV2
...
2021-11-23T12:57:30.426Z DEBUG shared::sources::balancer_v2::pool_init: initialized registered pools (block 13670942: 3 Stable, 73 Weighted & 72 TwoTokenWeighted)
...
2021-11-23T12:57:31.832Z DEBUG shared::sources::balancer_v2::pool_storage: replacing 0 events from block number 13670917
2021-11-23T12:57:31.832Z DEBUG shared::sources::balancer_v2::pool_storage: skipping deleting events from 13670917..=13670942
...
2021-11-23T12:57:32.303Z DEBUG solver::liquidity_collector: got 6 AMMs
...
```

### DIFF
--- a/crates/shared/src/sources/balancer_v2/event_handler.rs
+++ b/crates/shared/src/sources/balancer_v2/event_handler.rs
@@ -268,6 +268,7 @@ impl Maintaining for BalancerPoolRegistry {
         futures::try_join!(
             self.two_token_pool_updater.run_maintenance(),
             self.weighted_pool_updater.run_maintenance(),
+            self.stable_pool_updater.run_maintenance(),
         )?;
         Ok(())
     }
--- a/crates/shared/src/sources/balancer_v2/pool_storage.rs
+++ b/crates/shared/src/sources/balancer_v2/pool_storage.rs
@@ -166,6 +166,9 @@ pub struct PoolStorage {
     stable_pools: HashMap<H256, RegisteredStablePool>,
     #[derivative(Debug = "ignore")]
     data_fetcher: Arc<dyn PoolInfoFetching>,
+    /// The block the initial pools were fetched on. This block is considered
+    /// reorg-safe and events prior to this block do not get replaced.
+    initial_fetched_block: u64,
 }
 
 impl PoolStorage {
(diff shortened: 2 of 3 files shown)


vkgnosis (approved):

crates/shared/src/sources/balancer_v2/pool_storage.rs
```diff
@@ -178,12 +181,16 @@ impl PoolStorage {
...
-        PoolStorage {
+        let mut storage = PoolStorage {
             pools_by_token,
             weighted_pools,
             stable_pools,
             data_fetcher,
-        }
+            initial_fetched_block: 0,
+        };
+        storage.initial_fetched_block = storage.last_event_block();
```
fleupold, line 191: nit: Looks a bit like this should be part of the constructor logic (rather than relying on the caller to do that)?

fleupold (approved):
🕵️

crates/shared/src/sources/balancer_v2/pool_storage.rs
nlordell: This is in the constructor. We went from:
```
return PoolStorage { ... }
```

to:
```
let mut storage = PoolStorage { ... }
storage.initial_fetched_block = ...
return storage
```

The caller doesn't need to do anything. The reason I wrote it in this sort of icky way is to be able to reuse the `PoolStorage::last_event_block` function. This implementation detail shouldn't leak to the caller though.

crates/shared/src/sources/balancer_v2/event_handler.rs
```diff
@@ -268,6 +268,7 @@ impl Maintaining for BalancerPoolRegistry {
         futures::try_join!(
             self.two_token_pool_updater.run_maintenance(),
             self.weighted_pool_updater.run_maintenance(),
+            self.stable_pool_updater.run_maintenance(),
```
bh2smith, line 271: Noice catch!

crates/shared/src/sources/balancer_v2/pool_storage.rs
nlordell: I'm going ahead to merge this so that we can try it out in staging before using it in prod. Let me know if I misunderstood the nit and I will fix it in a follow up.

bh2smith (approved):
Excellent find here.

Issue #1407, opened 2021-11-18
Settlement submission refactor


sunce86:
Currently, we have several transaction submitting strategies, each with a different submission logic and parameters:
1. CustomNodes
2. Archer
3. Flashbots

General submission logic needs to be extracted in order to avoid code duplication and make maintenance easier.
Idea is to share common submission logic, but to have different implementation for submit/cancel functionalities. Also, smaller differences in algorithm should be handled through single parameter object containing all differences.

Refactoring will be done in steps:

- [x] Extract `Submitter` (common submission logic) from Archer and Flashbots and eliminate both.
- [x] Refactor CustomNodes to use `Submitter`
- [x] Refactor `Submitter` to support submitting settlement to multiple destinations at the same time.
- [x] Build a wrapper around `Submitter` that will implement smart decision making on type of the strategy/strategies to use for each individual settlement.

After the refactor, adding a new strategy will be equal to implementing submit/cancel functionality of newly added strategy.

fleupold:
Looks good. One thing I'd like to think about is how gas estimates work with the different strategies. Can we extract that logic in the existing GasEstimating abstraction or will different strategies (e.g Archer vs CustomNodes) need fundamentally different gas estimates.

> Refactor GeneralSubmission to support submitting settlement to multiple destinations at the same time.

Maybe it could just take different sub strategies at the same time? This way we could use Flashbots and Archer at the same time (which I believe is something we'd like to do).

> After the refactor, adding a new strategy will be equal to implementing submit/cancel functionality of newly added strategy.

Sounds good! Even for the CustomNodes we want cancellation of submission in case simulation fails, right (by resubmitting a same tx nonce with higher tip)?

nlordell:
Instead of introducing a new submission type (`GeneralSubmission`), maybe we can just reuse the existing `SolutionSubmitter`.

> One thing I'd like to think about is how gas estimates work with the different strategies

When creating the strategy, we could just:
```rust
let gas_estimator = PriorityGasEstimator::new(vec![
   StrategySpecificGasEstimator::new(),
   passed_in_gas_estimator,
]);
```

> Maybe it could just take different sub strategies at the same time?

The way I imagined it, we would have some general submission logic (i.e. the current submission loop that we use in `Archer` and `Flashbot` submission) and then pass in a `Submitting` trait that could combine/duplicate as needed:
```rust
struct SplitSubmitter {
   inner: Vec<Box<dyn Submitting>>,
}

[...]

PR #1433, opened 2021-11-25
Sunce86/merge flashbots and archer logic

sunce86:
This PR implements the point 1 from https://github.com/gnosis/gp-v2-services/issues/1407
(Extract `Submitter` (common submission logic) from Archer and Flashbots and eliminate both)

~~Merging to branch `sunce86/submission-refactor`, that serves as a container for all submission refactor changes, before merging everything to `master`.~~ Changed base to `main`.


nlordell:
I like this: 
image

I only took a cursory glance for now, but will look in more detail later!

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -13,91 +13,109 @@
...
+#[derive(Clone, Default)]
+pub struct SubmitterParams {
+    /// Desired duration to include the transaction in a block
+    pub target_confirm_time: Duration, //todo change to blocks in the following PR
+    /// Estimated gas consumption of a transaction
+    pub gas_estimate: U256,
+    /// Maximum max_fee_per_gas to pay for a transaction
+    pub gas_price_cap: f64,
+    /// Maximum duration of a single run loop
+    pub deadline: Option<SystemTime>,
```
vkgnosis, line 38: Should probably be an `Instant`.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -13,91 +13,109 @@
...
+    /// There are two ways to pay for a transaction:
+    /// 1. Set valid gas price (default)
+    /// 2. Set 0 gas price, but add additional step to settlement to pay directly to the coinbase
+    /// Contains quantity of gas needed to pay for the additional step
+    pub pay_gas_to_coinbase: Option<U256>,
```
vkgnosis, line 43: In this case I would like this configuration to be an enum to make it clear which parameters are valid together.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -13,91 +13,109 @@
...
+    /// Boost estimated gas price miner tip in order to increase the chances of a transaction being mined
+    pub additional_miner_tip: Option<f64>,
+    /// Resimulate and resend transaction on every retry_interval seconds
+    pub retry_interval: Duration,
+}
+
+#[async_trait::async_trait]
+pub trait TransactionSubmitting {
+    async fn submit_raw_transaction(&self, tx: &[u8], params: &SubmitterParams) -> Result<String>;
+    async fn cancel_transaction(&self, id: &str) -> Result<()>;
```
vkgnosis, line 53: Would like comments here to explain the semantics of these functions. If I was looking at this struct and trait in isolation without having seen the previous implementation I feel I would not know what to do here.

crates/solver/src/settlement_submission/submitter.rs
sunce86: Its not convenient to couple this with gasPrice because:
1. Gas price is unknown at the time when this is defined
2. Gas price is mutable value because of the iterating run loop.
Tried but ends with even less clean code...

crates/solver/src/settlement_submission/submitter.rs
sunce86: Changed.

crates/solver/src/settlement_submission/submitter.rs
sunce86: Added.

crates/solver/src/settlement_submission/submitter.rs
vkgnosis: What I was confused about is the comment "Set 0 gas price" . I thought this was referring to something in the struct so I suggested an enum that either stores that or pay_gas_to_coinbase. I see now that this isn't the case. I think the comment is talking about what we (this code) do with the transaction so it makes sense.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -13,91 +13,110 @@
...
+}
+
+#[async_trait::async_trait]
+pub trait TransactionSubmitting {
+    /// Submits raw signed transation to the specific network (public mempool, archer, flashbots...).
+    /// Returns transaction handle
+    async fn submit_raw_transaction(&self, tx: &[u8], params: &SubmitterParams) -> Result<String>;
+    /// Cancels already submitted transaction using the transaction handle
+    async fn cancel_transaction(&self, id: &str) -> Result<()>;
+}
```
vkgnosis, line 57: Could we use a newtype like `struct TransactionHandle(pub String)` to clarify the type here?
This goes more into nit territory now but you could even make it an associated type so implementations of the trait can pick what they want to use as a handle.

sunce86: Yes, this makes interface more readable. Changed.
However, for the associated type I am more in favor in not adding additional complexity when its not a must, also I believe we will have a String here for future strategies also.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -13,91 +13,116 @@
...
+pub struct TransactionHandle(pub String);
+
+#[async_trait::async_trait]
+pub trait TransactionSubmitting {
+    /// Submits raw signed transation to the specific network (public mempool, archer, flashbots...).
+    /// Returns transaction handle
+    async fn submit_raw_transaction(
+        &self,
+        tx: &[u8],
+        params: &SubmitterParams,
```
fleupold, line 59: SubmitterParams is a bit too powerful imo to be passed around on every submit attempt. It feels like some of the values inside are not changing (could therefore be instance state) and some values are not even relevant for each submission (e.g. deadline). 

Maybe it would make sense to introduce a different type that keeps this interface more minimal?

fleupold (commented):
> Merging to branch sunce86/submission-refactor, that serves as a container for all submission refactor changes, before merging everything to master

I'm worried about this approach. I fear it will lead to a huge PR getting merged into staging all at once which will make it hard to trace down regressions and result in a risky release. Is there a way we can incrementally merge those refactoring into main (e.g. what is this PR breaking that would prevent it from being merged in main directly?).

I really like the `TransactionSubmitting` abstraction, but I feel it does need a bit more careful documentation around what are its responsibilities and boundaries (e.g. would it be allowed to perform custom gas estimations which might be different between Archer and Flashbots, etc).
I think there are some really powerful strategies that we can compose in easy modular ways from this abstraction (e.g. Archer + Flashbots, timeouts, stateful strategies that switch between substrategies based on internal state or external signal (e.g. volatility)).

crates/solver/src/settlement_submission/submitter.rs
vkgnosis: Good point. We have to double check how Archer works today. Could even say we should delete the implementation for now and merge Flashbots + Traditional workflow first then maybe re-add archer later?

crates/solver/src/settlement_submission.rs
```diff
@@ -108,25 +108,26 @@ impl SolutionSubmitter {
...
-                        gas_estimate,
-                        *flashbots_tip,
-                    )
-                    .await;
+                let params = SubmitterParams {
+                    target_confirm_time: self.target_confirm_time,
+                    gas_estimate,
+                    gas_price_cap: self.gas_price_cap,
+                    deadline: Some(Instant::now() + *max_confirm_time),
+                    pay_gas_to_coinbase: None,
```
nlordell, line 126: Instead of making this part of the general "Submitter" can't this be an implementation detail of `ArcherApi`, where the payment to `coinbase` gets added to the interactions before encoding the calldata?

crates/solver/src/settlement_submission.rs
```diff
@@ -82,23 +80,25 @@ impl SolutionSubmitter {
...
-                        self.target_confirm_time,
-                        SystemTime::now() + *max_confirm_time,
-                        settlement,
-                        gas_estimate,
-                    )
-                    .await;
+                let params = SubmitterParams {
+                    target_confirm_time: self.target_confirm_time,
+                    gas_estimate,
+                    gas_price_cap: self.gas_price_cap,
```
nlordell, line 95: Why couldn't this be part of the `Submitter` and instead is in the `SubmitterParams`?

crates/solver/src/settlement_submission/submitter.rs
nlordell: I think `priority_fee` is largely equivalent to paying to coinbase right? At least if the `max_fee_per_gas` is large enough. I would be in favour of removing it if that is the case.

crates/solver/src/settlement_submission.rs
```diff
@@ -82,23 +80,25 @@ impl SolutionSubmitter {
...
+                    deadline: Some(Instant::now() + *max_confirm_time),
+                    pay_gas_to_coinbase: Some(U256::from(18346)),
+                    additional_miner_tip: None,
```
nlordell, line 98: Instead of implementing this as a submission parameter, couldn't we instead:
```rust
struct FlashbotsGasEstimator {
    additional_tip: f64,
    inner: Arc<dyn GasEstimating>,
}
impl GasEstimating for FlashbotsGasEstimator {
    async fn estimate_with_limits(
        &self,
        gas_limit: f64,
        time_limit: Duration,
    ) -> Result<EstimatedGasPrice> {
        let mut gas_price = self.inner.estimate_with_limits(gas_limit, time_limit).await?;
        if let Some(eip1559) = &mut gas_price.eip1559 {
            eip1559.max_fee_per_gas += self.additional_tip;
            eip1559.max_priority_fee_per_gas += self.additional_tip;
        } else {
            gas_price.legacy += self.additional_tip;
        }
        Ok(gas_price)
    }
}
```

And then when creating the submitter for `TransactionStrategy::Flashbots` we just pass in:
```rust
FlashbotsGasEstimator {
    addition_tip: flashbots_tip,
    inner: self.gas_price_estimator.clone(),
}
```

nlordell (commented):
Nice refactor!

Only comment would be that maybe it would be nice to move more implementation details out of the submission loop (such as wrapping the gas price estimator with a `FlashbotsGasEstimator` and getting rid of the `coinbase_gas`).

crates/solver/src/settlement_submission.rs
fleupold: This would require the interface to take the MethodBuilder instead of already encoded calldata, which I'd also be in favour of.

crates/solver/src/settlement_submission/submitter.rs
sunce86: With the new archer (eden) integration this field is not needed anymore.

crates/solver/src/settlement_submission/submitter.rs
sunce86: Params removed since its not needed at this moment. For the future, will remember this comment.

crates/solver/src/settlement_submission.rs
sunce86: Removed as its not needed.

crates/solver/src/settlement_submission.rs
sunce86: It could be, however I aimed here to decouple the parameters that influence the execution path of the `Submitter`. This way, `Submitter` will be the same for all strategies, while `SubmitterParams` will be different for each strategy.
Also, not completely sure what will be our needs in the next PR when we implement point 4 from https://github.com/gnosis/gp-v2-services/issues/1407

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -13,91 +13,116 @@
...
+    /// Boost estimated gas price miner tip in order to increase the chances of a transaction being mined
+    pub additional_miner_tip: Option<f64>,
+    /// Resimulate and resend transaction on every retry_interval seconds
+    pub retry_interval: Duration,
+}
+
+pub struct TransactionHandle(pub String);
+
+#[async_trait::async_trait]
+pub trait TransactionSubmitting {
```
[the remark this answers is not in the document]
sunce86: Yes, it should be fairly easy to dummy implement TransactionSubmitting.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -13,91 +13,107 @@
...
 use primitive_types::{H256, U256};
 use shared::Web3;
-use std::time::{Duration, Instant, SystemTime};
+use std::time::{Duration, Instant};
 use web3::types::TransactionReceipt;
 
-pub struct FlashbotsSolutionSubmitter<'a> {
+/// Parameters for transaction submitting
+#[derive(Clone, Default)]
+pub struct SubmitterParams {
```
fleupold, line 30: I still think this is a bit of a magic toolbox with a bunch of unrelated things in it. `gas_price_cap`, `additional_miner_tip`, `retry_interval` and `target_confirm_time` should be passed into the submitter directly in my opinion.

The only thing that is dependent on the current submission is the gas_estimate and the deadline.

fleupold (approved):
Looks good to me to test. I'm not sure we will be able to use the TransactionSubmitting trait to compose complex submission strategies (I think we might have to use an abstraction that's closer to the current Submitter interface).

But the code looks good and should get us testing with Eden 🎉

nlordell (approved):
LGTM

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -13,91 +13,107 @@
...
+    pub gas_price_cap: f64,
+    /// Maximum duration of a single run loop
+    pub deadline: Option<Instant>,
+    /// Boost estimated gas price miner tip in order to increase the chances of a transaction being mined
+    pub additional_miner_tip: Option<f64>,
+    /// Resimulate and resend transaction on every retry_interval seconds
+    pub retry_interval: Duration,
+}
+
+pub struct TransactionHandle(pub String);
```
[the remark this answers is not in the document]
nlordell: One issue with associated types is that they make trait objects harder to use (you either need to specify the associated types upfront or some up with some wrapping type to make everything object safe).

Since both use 32 byte hashes, I suggest making it an `H256` which should cover all our bases for now, and then reevaluate when we find something that doesn't fit.

crates/solver/src/settlement_submission/submitter.rs
nlordell: I slightly agree, since we create a `Submitter::new`for every transaction, the division seems rather arbitrary to me and would be in favour of just shoving everything into one struct.

Of this seperation makes more sense for future refactor ideas, I'm also fine with it as it is.

crates/solver/src/settlement_submission.rs
sunce86: Nice one! Now `additional_tip` and `gas_price_cap` are coupled with the gas price estimator, as they should.

crates/solver/src/settlement_submission/submitter.rs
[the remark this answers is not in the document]
sunce86: @fleupold 
Is the comment equal to this one: https://github.com/gnosis/gp-v2-services/pull/1433#discussion_r757435096
@nlordell 
Makes sense, changed to H256

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -13,91 +13,116 @@
...
+
+#[async_trait::async_trait]
+pub trait TransactionSubmitting {
+    /// Submits raw signed transation to the specific network (public mempool, archer, flashbots...).
+    /// Returns transaction handle
+    async fn submit_raw_transaction(
+        &self,
+        tx: &[u8],
+        params: &SubmitterParams,
+    ) -> Result<TransactionHandle>;
```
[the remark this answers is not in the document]
sunce86: I would rather leave caller to work with `TransactionHandle` instead of with specific type (in order to abstract away the type), and also I believe that the trait is more readable with `Result<TransactionHandle>` instead of `Result<H256>`.
Also, caller handles potential error by just tracing it, so I don't see any purpose of introducing specific type of error here.
Not sure if I answered the question :smiley:

crates/solver/src/settlement_submission/submitter.rs
sunce86: `SubmitterParams` is now reduced. However, I do have a feeling we will need params separated in the following refactor PRs, when we will define submitter but call it with different params for different strategies (maybe I'm wrong). Any way, I will remember this comment and revisit it later.

sunce86:
PR tested, merging so we can continue with the refactor of `CustomNodes` strategy.

PR #1435, opened 2021-11-25
Pass in conctracts instead of calling deployed locally

### LINKED ISSUE #1379: Get rid of instsances of `Contract::deployed(web3).await.unwrap()`
We should either pass the contracts in or have a global that is populated on start up. This way we only need to talk to web3 once and it allows us to use contracts that do not have their deployment information baked in.
The global would have to store optionals because not all contracts exist on all networks.

Motivated by https://github.com/gnosis/gp-v2-services/issues/1377

vkgnosis:
Fixes https://github.com/gnosis/gp-v2-services/issues/1379

I decided to go with this simple solution of passing all required contracts in and querying them from the node in main. We can  still decide on a different refactoring in the future but this already makes the code more pure and future work easier.

### Test Plan
CI, run locally

### DIFF
--- a/crates/orderbook/src/main.rs
+++ b/crates/orderbook/src/main.rs
@@ -1,5 +1,6 @@
 use anyhow::{anyhow, Context, Result};
 use contracts::{BalancerV2Vault, GPv2Settlement, WETH9};
+use ethcontract::errors::DeployError;
 use model::{
     app_id::AppId,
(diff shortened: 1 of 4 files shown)


vkgnosis:
Updated to latest master.

nlordell (approved):
LGTM

PR #1436, opened 2021-11-26
Differentiate between different settlement submission outcomes.

nlordell:
This PR adds a new `SettlementSubmissionOutcome` enum and uses it for determining the `result` value for our `settlement_submission` metric. With this, the settlement submission metrics differentiates between different outcomes:
- success: transaction was mined
- timeout: transaction did not get mined in time
- revert: transaction was submitted and mined but reverted
- failure: transaction failed to submit - nothing was mined

Additionally, we no longer print `error!` logs for submission failures but instead can setup alerts for them instead.

### Test Plan

Added unit test for error conversion. Additionally we can check the metrics are working.

<details><summary>Apply the following patch</summary>

```diff
diff --git a/crates/solver/src/settlement_submission.rs b/crates/solver/src/settlement_submission.rs
index f70341b..4d90c59 100644
--- a/crates/solver/src/settlement_submission.rs
+++ b/crates/solver/src/settlement_submission.rs
@@ -31,6 +31,8 @@ use web3::types::TransactionReceipt;
 const ESTIMATE_GAS_LIMIT_FACTOR: f64 = 1.2;
 const GAS_PRICE_REFRESH_INTERVAL: Duration = Duration::from_secs(15);
 
+static COUNTER: std::sync::atomic::AtomicUsize = std::sync::atomic::AtomicUsize::new(0);
+
 pub struct SolutionSubmitter {
     pub web3: Web3,
     pub contract: GPv2Settlement,
@@ -128,7 +130,18 @@ impl SolutionSubmitter {
                 result?.ok_or(SubmissionError::Timeout)
             }
             TransactionStrategy::DryRun => {
-                Ok(dry_run::log_settlement(account, &self.contract, settlement).await?)
+                match COUNTER.fetch_add(1, std::sync::atomic::Ordering::Relaxed) % 4 {
+                    0 => Ok(TransactionReceipt {
+                        transaction_hash: Default::default(),
+                        block_hash: Some(Default::default()),
+                        block_number: Some(Default::default()),
+                        ..Default::default()
+                    }),
+                    1 => Err(SubmissionError::Revert(None)),
+                    2 => Err(SubmissionError::Timeout),
+                    _ => Err(SubmissionError::Other(anyhow::anyhow!("foo"))),
+                }
+                //Ok(dry_run::log_settlement(account, &self.contract, settlement).await?)
             }
         }
     }
```

</details>

Run the solver in `DryRun` mode for a new runs:
```
[...]

PR #1438, opened 2021-11-27
Turn submission timeouts into warnings and add metrics for those

taminomara:
Transaction timeout error is inactionable, so we turn it into a warning. We also add more metrics for the settlement submission subsystem.

### Test Plan
Manual testing

### DIFF
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -1,6 +1,7 @@
 pub mod solver_settlements;
 
 use self::solver_settlements::RatedSettlement;
+use crate::settlement_submission::TransactionTimeoutError;
 use crate::{
     analytics, auction_preprocessing,
     in_flight_orders::InFlightOrders,
@@ -179,10 +180,11 @@ impl Driver {
                 // Since we simulate and only submit solutions when they used to pass before, there is no
                 // point in logging transaction failures in the form of race conditions as hard errors.
                 let name = solver.name();
-                if err
-                    .downcast_ref::<MethodError>()
-                    .map(|e| is_transaction_failure(&e.inner))
-                    .unwrap_or(false)
+                if err.is::<TransactionTimeoutError>()
+                    || err
+                        .downcast_ref::<MethodError>()
+                        .map(|e| is_transaction_failure(&e.inner))
+                        .unwrap_or(false)
                 {
                     tracing::warn!("Failed to submit {} settlement: {:?}", name, err)
                 } else {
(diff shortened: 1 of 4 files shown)


crates/solver/src/settlement_submission.rs
```diff
@@ -65,9 +83,13 @@ impl SolutionSubmitter {
         gas_estimate: U256,
         account: Account,
     ) -> Result<TransactionReceipt> {
+        let metrics: &SettlementSubmissionMetrics =
+            SettlementSubmissionMetrics::instance(get_metric_storage_registry()).unwrap();
+        // metrics.requests_complete.with_label_values(&["", ""]).inc();
```
fleupold, line 88: Is this supposed to stay?

crates/solver/src/driver.rs
```diff
@@ -179,10 +180,11 @@ impl Driver {
                 // Since we simulate and only submit solutions when they used to pass before, there is no
                 // point in logging transaction failures in the form of race conditions as hard errors.
                 let name = solver.name();
-                if err
-                    .downcast_ref::<MethodError>()
-                    .map(|e| is_transaction_failure(&e.inner))
-                    .unwrap_or(false)
+                if err.is::<TransactionTimeoutError>()
```
fleupold, line 183: I'm a bit worried we are splitting the logic which tokens should not be errored upon and the fact that those are reported otherwise in two very different places.

Would it be a lot of work to refactor this error into a custom type that can have a `log_severity` or `should_error` flag?

Alternatively we could the logging of such errors here instead of in the submission module below.

crates/solver/src/settlement_submission.rs
```diff
@@ -127,15 +146,63 @@ impl SolutionSubmitter {
...
+            metrics
+                .requests_complete
+                .with_label_values(&[transaction_strategy, "error"])
+                .inc();
+            Err(err)
+        }
+    }
+}
+
+fn process_simple_result(
```
fleupold, line 197: It's just an internal function, but I generally don't like naming things "simple" as it doesn't really tell you anything about what this function does (process_tx_hash vs. process_optional_tx_hash might be better). E.g.  it turns out that `NaiveSolver` wasn't a great name as it's not really explanatory.

Here specifically though, can't we just call `process_result(result.map(Some), transaction_strategy, metrics)`? This way we don't have to duplicate label names and other shared logic.

fleupold (commented):
I thought this was merged already. Since this week's release we haven't seen any 

> Failed to submit xyz settlement: transaction did not get mined in time

errors. Will take a look

fleupold:
Actually, isn't this superseded by https://github.com/gnosis/gp-v2-services/pull/1436 ?

nlordell:
@taminomara can this be closed? I think it was superseded by #1436

PR #1439, opened 2021-11-29
Refactor Balancer Graph pool fetching

nlordell:
This PR refactors the Balancer V2 Graph API logic to reduce the amount of duplicated code between weighted and stable pools and query them in a single request.

This should hopefully make adding additional pools to the Graph query easier as well as remove some code duplication.

### Test Plan

CI, unit tests adapted for new query. Also, check that we can read the pools:
```
$ cargo test -p shared -- balancer_subgraph_query --ignored --nocapture
running 1 test
### Mainnet
Retrieved 74 pools for factory at 0x8e9a…7de9
Retrieved 72 pools for factory at 0xa5bf…4ee0
Retrieved 146 total weighted pools at block 13708201
Retrieved 3 pools for factory at 0xc66b…de24
Retrieved 3 total stable pools at block 13708201
### Rinkeby
Retrieved 18 pools for factory at 0x8e9a…7de9
Retrieved 8 pools for factory at 0xa5bf…4ee0
Retrieved 26 total weighted pools at block 9726658
Retrieved 0 total stable pools at block 9726658
test sources::balancer_v2::graph_api::tests::balancer_subgraph_query ... ok
```

PR #1440, opened 2021-11-29
Refactor Balancer pool initialization

nlordell:
This PR refactors Balancer pool initialization.

Specifically, we used to need to (depending on the network) fetch some pool data on-chain (i.e. with `eth_call`s). This is no longer the case, so this PR lets us remove a bunch of code that is no longer needed in addition to speeding up initialization.

Additionally, this PR also refactors how the initialization is done to remove even more code duplication between weighted and stable pools.

This is a follow up to #1439.

### Test Plan

Adjusted unit tests to refactor. Additionally, run the manual tests:
```
$ cargo test -p shared -- balancer_subgraph_query --ignored --nocapture
...
### Mainnet
Retrieved 149 total pools at block 13708699
- 72 Weighted pools from factory 0xa5bf2ddf098bb0ef6d120c98217dd6b141c74ee0
- 3 Stable pools from factory 0xc66ba2b6595d3613ccab350c886ace23866ede24
- 74 Weighted pools from factory 0x8e9aa87e45e92bad84d5f8dd1bff34fb92637de9
### Rinkeby
Retrieved 26 total pools at block 9727120
- 18 Weighted pools from factory 0x8e9aa87e45e92bad84d5f8dd1bff34fb92637de9
- 8 Weighted pools from factory 0xa5bf2ddf098bb0ef6d120c98217dd6b141c74ee0
test sources::balancer_v2::graph_api::tests::balancer_subgraph_query ... ok
...
$ cargo test -p shared -- balancer_subgraph_initialization --ignored --nocapture
...
### Mainnet
Retrieved 149 total pools at block 13708705
- 74 weighted pools
- 72 weighted two-token pools
- 3 stable pools
### Rinkeby
Retrieved 26 total pools at block 9727124
- 18 weighted pools
- 8 weighted two-token pools
- 0 stable pools
test sources::balancer_v2::pool_init::tests::balancer_subgraph_initialization ... ok
...
```

### DIFF
--- a/crates/shared/src/sources/balancer_v2/graph_api.rs
+++ b/crates/shared/src/sources/balancer_v2/graph_api.rs
@@ -8,19 +8,17 @@
 //! - ensure that we are using the latest up-to-date pool data by using events
 //!   from the node
 
-use self::pools_query::{PoolData, PoolType};
-use crate::{
-    event_handling::MAX_REORG_BLOCK_COUNT,
-    sources::balancer_v2::pool_storage::{
-        CommonPoolData, RegisteredStablePool, RegisteredWeightedPool,
-    },
-    subgraph::SubgraphClient,
+use super::{
+    pool_storage::{CommonPoolData, RegisteredStablePool, RegisteredWeightedPool},
+    swap::fixed_point::Bfp,
 };
+use crate::{event_handling::MAX_REORG_BLOCK_COUNT, subgraph::SubgraphClient};
 use anyhow::{anyhow, bail, Result};
 use ethcontract::{H160, H256};
 use reqwest::Client;
+use serde::Deserialize;
 use serde_json::json;
-use std::collections::HashMap;
+use serde_with::{serde_as, DisplayFromStr};
 
 /// The page size when querying pools.
 #[cfg(not(test))]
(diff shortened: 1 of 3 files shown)


crates/shared/src/sources/balancer_v2/graph_api.rs
```diff
@@ -114,56 +115,80 @@ pub struct RegisteredPools {
...
+
+/// Token data for pools.
+#[serde_as]
+#[derive(Debug, Deserialize, PartialEq)]
+pub struct Token {
+    pub address: H160,
+    pub decimals: u8,
+    #[serde_as(as = "Option<DisplayFromStr>")]
+    #[serde(default)]
+    pub weight: Option<Bfp>,
```
fleupold, line 147: Would it be safe to assume even weights for pools in which this is None (may allow more consolidation of the data model)?

fleupold (approved):

crates/shared/src/sources/balancer_v2/graph_api.rs
nlordell: The risk is we may assume we have weights if they are missing. They used to be on Rinkeby for example (but because of subgraph updates they are available now).

We convert the `None` into an `Err` for a weighted pool in those cases.

bh2smith (approved):

PR #1447, opened 2021-11-29
[1/-] Introduce factory indexing module

nlordell:
This PR introduces a new `pools` module that will ultimately contain all the Balancer pool specific logic. Ultimately, the goal is to be able to add new Balancer pools by:
- Extending the `graph_api::PoolData` to include required fields
- Add a new `pools::$POOL_TYPE` module for the new factory/pool
- Implement the `FactoryIndexing` and `PoolIndexing` traits.

The `FactoryIndexing` trait was defined (although, it is still missing a lot of things that will be introduced later, such as fetching permanent pool info and fetching current pool state). Mostly this PR moves the `PoolData` -> `Registered*Pool` conversion logic to the new module.

### Test Plan

Logic moved around. There is a could extra error cases that were checked with new unit tests.

### DIFF
--- a/crates/shared/src/sources/balancer_v2/pools/stable.rs
+++ b/crates/shared/src/sources/balancer_v2/pools/stable.rs
@@ -1,0 +1,65 @@
+//! Module implementing stable pool specific indexing logic.
+
+use super::{common, FactoryIndexing, PoolIndexing};
+use crate::sources::balancer_v2::graph_api::{PoolData, PoolType};
+use anyhow::{ensure, Result};
+use contracts::BalancerV2StablePoolFactory;
+
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct PoolInfo {
+    pub common: common::PoolInfo,
+}
+
+impl PoolIndexing for PoolInfo {
+    fn from_graph_data(pool: &PoolData, block_created: u64) -> Result<Self> {
+        ensure!(
+            pool.pool_type == PoolType::Stable,
+            "cannot convert {:?} pool to stable pool",
+            pool.pool_type,
+        );
+
+        Ok(PoolInfo {
+            common: common::PoolInfo::from_graph_data(pool, block_created)?,
+        })
+    }
+
+    fn common(&self) -> &common::PoolInfo {
+        &self.common
+    }
+}
+
+#[async_trait::async_trait]
+impl FactoryIndexing for BalancerV2StablePoolFactory {
+    type PoolInfo = PoolInfo;
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::sources::balancer_v2::graph_api::Token;
+    use ethcontract::{H160, H256};
+
+    #[test]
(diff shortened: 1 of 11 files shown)


crates/shared/src/sources/balancer_v2/pools/weighted_2token.rs
```diff
@@ -0,0 +1,10 @@
+//! Module implementing two-token weighted pool specific indexing logic.
```
bh2smith, line 1: I guess we don't need to redefine pool info here since we can reuse the one from weighted? Makes me wonder if we really want to have this additional file. Could we just put the content here into weighted or is this a letter of cleanliness?

bh2smith (approved):
Really looking nice! Great job here on the refactor!

crates/shared/src/sources/balancer_v2/pools/weighted_2token.rs
nlordell: I put it in because we still need a separate contract for indexing events, and will eventually be instantiated with `BalancerWeightedPool2TokenFactory` instead of `BalancerWeighedPoolFactory`. It will probably always stay fairly empty.

crates/shared/src/sources/balancer_v2/pools/stable.rs
```diff
@@ -0,0 +1,65 @@
+//! Module implementing stable pool specific indexing logic.
+
+use super::{common, FactoryIndexing, PoolIndexing};
+use crate::sources::balancer_v2::graph_api::{PoolData, PoolType};
+use anyhow::{ensure, Result};
+use contracts::BalancerV2StablePoolFactory;
+
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct PoolInfo {
+    pub common: common::PoolInfo,
```
fleupold, line 10: If we compose here instead of aliasing is it really needed that `common::PoolInfo` is implementing `PoolIndexing` as well? There is quite a few structs with similar names

fleupold (approved):

crates/shared/src/sources/balancer_v2/pools/stable.rs
nlordell: No it isn't. Good point. I had originally aliased `stable::PoolInfo == common::PoolInfo` but moved away from it.

I will remove the `impl PoolIndexing for common::PoolInfo`.

Regarding multiple `PoolInfo` structs, this is just a question of personal preference. I tend to prefer writing:
```rust
use path::to::stable;
stable::PoolInfo

// --- OR ---

use path::to::stable::PoolInfo as StablePoolInfo;
StablePoolInfo
```

over
```rust
use path::to::stable::StablePoolInfo;
StablePoolInfo
```

Mostly because I find `stable::StablePoolInfo` looks a bit redundant. Happy to change this though, I don't have strong feelings about it.

bh2smith: I like the first option of your preferences (before the OR).

PR #1449, opened 2021-11-30
[2/-] Split info fetching for each factory

nlordell:
This PR splits each `PoolInfo` fetching into each individual pool's `FactoryIndexing` implementation.

Currently we make use of a legacy `PoolInfoFetcher` adapter to avoid changing too much in this PR.

In the future, we will be able to create a generic `PoolFetcher<FactoryIndexing>` instance that will will for every pool:
1. Fetch `common::PoolInfo` which all pools have (this is using the `common::PoolInfoFetching` interface).
2. Pass the fetched `common::PoolInfo` into the `FactoryIndexing` implementation in order to augment it into a `{weighted,stable}::PoolInfo`
3. Index the `{weighted,stable}::PoolInfo` and create lookups by token address (as we do now).

### Test Plan

Added new tests with mock contract calls.

Use manual test to index all past events:
```
$ export NODE_URL=...
$ cargo test -p shared -- balancer_indexed_pool_events_match_subgraph --nocapture --ignored
...
test sources::balancer_v2::event_handler::tests::balancer_index_pool_events ... ok
```

### DIFF
--- a/crates/shared/src/sources/balancer_v2/info_fetching.rs
+++ b/crates/shared/src/sources/balancer_v2/info_fetching.rs
@@ -8,9 +8,11 @@ use super::{
         FactoryIndexing as _,
     },
 };
+use crate::token_info::TokenInfoFetching;
 use anyhow::Result;
 use contracts::{BalancerV2StablePoolFactory, BalancerV2Vault, BalancerV2WeightedPoolFactory};
 use ethcontract::H160;
+use std::sync::Arc;
 
 /// Legacy pool info fetching implementation.
 ///
@@ -26,11 +28,12 @@ pub struct PoolInfoFetcher {
 impl PoolInfoFetcher {
     pub fn new(
         vault: BalancerV2Vault,
+        token_infos: Arc<dyn TokenInfoFetching>,
         weighted_factory: BalancerV2WeightedPoolFactory,
         stable_factory: BalancerV2StablePoolFactory,
     ) -> Self {
         Self {
-            inner: common::PoolInfoFetcher::new(vault),
+            inner: common::PoolInfoFetcher::new(vault, token_infos),
             weighted_factory,
             stable_factory,
         }
--- a/crates/shared/src/sources/balancer_v2/info_fetching.rs
+++ b/crates/shared/src/sources/balancer_v2/info_fetching.rs
@@ -1,340 +1,75 @@
-//! Responsible for conversion of a `pool_address` into `WeightedPoolInfo` which is used by the
-//! event handler to construct a `RegisteredWeightedPool`.
-use crate::{
-    sources::balancer_v2::swap::fixed_point::Bfp, token_info::TokenInfoFetching,
-    transport::MAX_BATCH_SIZE, Web3,
+//! Responsible for conversion of a `pool_address` into `*::PoolInfo` which is
+//! stored by the event handler
+
+use super::{
+    pool_storage::{RegisteredStablePool, RegisteredWeightedPool},
+    pools::{
+        common::{self, PoolInfoFetching as _},
+        FactoryIndexing as _,
+    },
 };
-use anyhow::{anyhow, Result};
-use contracts::{BalancerV2StablePool, BalancerV2Vault, BalancerV2WeightedPool};
-use ethcontract::{batch::CallBatch, Bytes, H160, H256};
-use mockall::*;
-use std::sync::Arc;
+use anyhow::Result;
+use contracts::{BalancerV2StablePoolFactory, BalancerV2Vault, BalancerV2WeightedPoolFactory};
+use ethcontract::H160;
 
-#[derive(Clone, Debug)]
-pub struct CommonPoolInfo {
-    pub id: H256,
-    pub tokens: Vec<H160>,
-    pub scaling_exponents: Vec<u8>,
-}
-
-#[derive(Clone, Debug)]
-pub struct WeightedPoolInfo {
-    pub common: CommonPoolInfo,
-    pub weights: Vec<Bfp>,
-}
-
-#[derive(Clone, Debug)]
-pub struct StablePoolInfo {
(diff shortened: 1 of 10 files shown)


crates/shared/src/sources/balancer_v2/info_fetching.rs
```diff
@@ -1,340 +1,78 @@
...
+        &self,
+        pool_address: H160,
+        block_created: u64,
+    ) -> Result<RegisteredStablePool>;
 }
 
 #[async_trait::async_trait]
 impl PoolInfoFetching for PoolInfoFetcher {
     /// Could result in ethcontract::{NodeError, MethodError or ContractError}
-    async fn get_weighted_pool_data(&self, pool_address: H160) -> Result<WeightedPoolInfo> {
```
nlordell, line 59: This logic was split into `pools/common.rs` and `pools/weighted.rs`.

crates/shared/src/sources/balancer_v2/info_fetching.rs
```diff
@@ -1,340 +1,78 @@
...
-            },
-            weights: raw_normalized_weights
-                .await?
-                .into_iter()
-                .map(Bfp::from_wei)
-                .collect(),
-        })
-    }
-
-    async fn get_stable_pool_data(&self, pool_address: H160) -> Result<StablePoolInfo> {
```
nlordell, line 92: This logic was split into `pools/common.rs` and `pools/stable.rs`.

bh2smith (approved):
Looking good.

crates/shared/src/sources/balancer_v2/pools/stable.rs
```diff
@@ -25,6 +25,10 @@ impl PoolIndexing for PoolInfo {
 #[async_trait::async_trait]
 impl FactoryIndexing for BalancerV2StablePoolFactory {
     type PoolInfo = PoolInfo;
+
+    async fn pool_info(&self, pool: common::PoolInfo) -> Result<Self::PoolInfo> {
+        Ok(PoolInfo { common: pool })
```
fleupold, line 30: Shouldn't there be a check that this pool id is in fact a StablePool to begin with? How does the main code know which stable pool factory to call `pool_info` on?

crates/shared/src/sources/balancer_v2/pools.rs
```diff
@@ -24,6 +24,9 @@ pub trait FactoryIndexing {
     /// be retrieved once. This data will be passed in when fetching the current
     /// pool state via `fetch_pool`.
     type PoolInfo: PoolIndexing;
+
+    /// Retrive the permanent pool info for the specified common pool info.
```
fleupold, line 28: Maybe document when this method is expected to return an error?

fleupold (commented):
I think I'm missing a little bit of context how this pool_info is supposed to be used. Is there a follow up PR I need to consider?

nlordell:
> I think I'm missing a little bit of context how this pool_info is supposed to be used.

Added more context to the PR description.

crates/shared/src/sources/balancer_v2/pools/stable.rs
nlordell: > Shouldn't there be a check that this pool id is in fact a StablePool to begin with?

I think maybe this method should be renamed. Its purpose is to take some `common::PoolInfo` and "augment" it into the pool-specific "permanent" data. Would `augment_pool_info` make more sense?

For example, the `weighted::PoolInfo` additionally fetches the immutable weight parameters. Stable pools don't fetch any additional data so the augmenting is trivial.

We could technically do a check via `evm_call`, but I'm not sure its worth it. This interface is intended to be used by some:
```rust
struct PoolFetcher<Factory> {
    factory: Factory,
    common_info_fetcher: Arc<dyn common::PoolInfoFetching>,
    // ...
}

impl<Factory: FactoryIndexing> PoolFetcher<Factory> {
    fn on_pool_created_event(&self, pool_created: PoolCreated) {
        let common_pool_info = self.common_info_fetcher.pool_info(pool_created.pool_address)?;
        let augmented_pool_info = self.factory.augment_pool_info(&common_pool_info)?;
        // index `augmented_pool_info`...
    }
}
```

nlordell: > How does the main code know which stable pool factory to call pool_info on?

The source of the `PoolCreated` event.

nlordell: I'm going to rename methods to:
```
common::PoolInfoFetcher::fetch_common_pool_info()
FactoryIndexing::augument_pool_info()
```

fleupold: > Would augment_pool_info make more sense?

I'm fine with that just some other ideas would be `specialize_pool` or `downcast_pool` ?

crates/shared/src/sources/balancer_v2/pools.rs
```diff
@@ -24,6 +24,16 @@ pub trait FactoryIndexing {
...
     type PoolInfo: PoolIndexing;
+
+    /// Augments the specified common pool info for this factory.
+    ///
+    /// This allows pool factories like the `WeightedPoolFactory` to add
+    /// `weights` to the common pool info, since these are declared as
+    /// `immuatble` in the smart contract and thus can never change and don't
+    /// need to be re-fetched.
+    ///
+    /// Returns an error if fetching the augmented pool data fails.
```
fleupold, line 35: nanonit: A check to verify that the given pool id in in fact of type `Self::PoolInfo` is not mandatory.

fleupold (approved):

crates/shared/src/sources/balancer_v2/pools/stable.rs
nlordell: I'm partial towards `specialize_pool`.

PR #1450, opened 2021-12-01
Add Balancer event indexing manual test

nlordell:
This PR just adds a manual test for verifying complete Balancer pool indexing via events (we don't exercise this path usually because we typically initialize by reading the Subgraph, but its useful for refactoring event indexing code to make sure everything is working).

Additionally, I noticed that the test was failing because the deployment information included the transaction hash and not the block number which we use to know which block to start indexing events on. I changed all `DeploymentInformation`s to use block numbers instead of transaction hashes.

### Test Plan

This is the test!
```
$ export NODE_URL=...
$ cargo test -p shared -- balancer_index_pool_events --nocapture --ignored
...
test sources::balancer_v2::event_handler::tests::balancer_index_pool_events ... ok
```

### DIFF
--- a/crates/shared/src/sources/balancer_v2/event_handler.rs
+++ b/crates/shared/src/sources/balancer_v2/event_handler.rs
@@ -340,7 +340,7 @@ mod tests {
 
     #[tokio::test]
     #[ignore]
-    async fn balancer_index_pool_events() {
+    async fn balancer_indexed_pool_events_match_subgraph() {
         let transport = transport::create_env_test_transport();
         let web3 = Web3::new(transport);
         let chain_id = web3.eth().chain_id().await.unwrap().as_u64();
--- a/crates/shared/src/sources/balancer_v2/event_handler.rs
+++ b/crates/shared/src/sources/balancer_v2/event_handler.rs
@@ -322,3 +322,90 @@ fn convert_stable_pool_created(
         },
     })
 }
+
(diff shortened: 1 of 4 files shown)


crates/shared/src/sources/balancer_v2/event_handler.rs
```diff
@@ -322,3 +322,90 @@ fn convert_stable_pool_created(
...
+        token_info::TokenInfoFetcher,
+        transport,
+    };
+    use contracts::BalancerV2Vault;
+    use maplit::hashset;
+    use reqwest::Client;
+
+    #[tokio::test]
+    #[ignore]
+    async fn balancer_index_pool_events() {
```
fleupold, line 343: nanonit: name could be a bit clearer that this is comparing indexed vs subgraph events.

fleupold (approved):
Really nice tests!

> Additionally, I noticed that the test was failing because the deployment information included the transaction hash and not the block number which we use to know which block to start indexing events on.

The reason we didn't notice is because we never really index by events from scratch?

nlordell:
> The reason we didn't notice is because we never really index by events from scratch?

Exactly.

crates/shared/src/sources/balancer_v2/event_handler.rs
nlordell: Fixed!

PR #1451, opened 2021-12-01
Add Balancer `BasePool` ABI

nlordell:
This PR adds the Balancer `BasePool` ABI to our project which we will use later (in #1449). Specifically, it will be used for fetching common pool data that is the same for all Balancer pools (for example, all Balancer pools implement the `getPoolId` function for querying the ID of the pool at a given address).

It allows us use every indexed pool by address (regardless if its a `Weighted(2Token)Pool`, `StablePool` or some fictional `FooBarPool`) as `BalancerV2BasePool::at(&web3, pool_address)` in order to access the common view functions.

I just split this into a separate PR because the aforementioned #1449 was a little too big to review.

### Test Plan

Compiler :guitar:.

### DIFF
--- a/crates/contracts/artifacts/BalancerV2BasePool.json
+++ b/crates/contracts/artifacts/BalancerV2BasePool.json
@@ -1,0 +1,53 @@
+{
+  "abi": [
+    {
+      "inputs": [],
+      "name": "getPausedState",
+      "outputs": [
+        {
+          "internalType": "bool",
+          "name": "paused",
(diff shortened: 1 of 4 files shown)


fleupold (approved):
Looks good. Would you be able to explain what the BasePool will eventually be used for?

nlordell:
> Would you be able to explain what the BasePool will eventually be used for?

Updated the description.

PR #1455, opened 2021-12-03
Fix timeout logic in HTTP solver client

taminomara:
This PR lowers allowed time to interact with network to 50ms, because one second is too much. Also it increases quasimodo price estimation timeout to 2s because quasimodo reserves at least one second for maintenante work.

### Test Plan
Manual testing: try estimating prices for different tokens using different timeouts.

### DIFF
--- a/crates/shared/src/http_solver_api.rs
+++ b/crates/shared/src/http_solver_api.rs
@@ -67,10 +67,10 @@ impl HttpSolverApi for DefaultHttpSolverApi {
         let timeout = deadline
             .checked_duration_since(Instant::now())
             .ok_or_else(|| anyhow!("no time left to send request"))?;
-        // The timeout we give to the solver is one second less than the deadline to make up for
-        // overhead from the network.
+        // The timeout we give to the solver is a few milliseconds less than
+        // the deadline to make up for overhead from the network.
         let solver_timeout = timeout
-            .checked_sub(Duration::from_secs(1))
+            .checked_sub(Duration::from_millis(250))
             .ok_or_else(|| anyhow!("no time left to send request"))?;
 
         let mut url = self.base.clone();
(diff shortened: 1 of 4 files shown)


crates/shared/src/http_solver_api.rs
```diff
@@ -67,10 +67,10 @@ impl HttpSolverApi for DefaultHttpSolverApi {
         let timeout = deadline
             .checked_duration_since(Instant::now())
             .ok_or_else(|| anyhow!("no time left to send request"))?;
-        // The timeout we give to the solver is one second less than the deadline to make up for
-        // overhead from the network.
+        // The timeout we give to the solver is a few milliseconds less than
+        // the deadline to make up for overhead from the network.
         let solver_timeout = timeout
-            .checked_sub(Duration::from_secs(1))
+            .checked_sub(Duration::from_millis(50))
```
nlordell, line 73: Is 50 ms enough? Maybe 100ms might be a bit safer. Either way, I'm OK with lowering it from 1s which seemed a bit high.

nlordell (approved):

crates/shared/src/http_solver_api.rs
```diff
@@ -67,10 +67,10 @@ impl HttpSolverApi for DefaultHttpSolverApi {
         let timeout = deadline
             .checked_duration_since(Instant::now())
             .ok_or_else(|| anyhow!("no time left to send request"))?;
-        // The timeout we give to the solver is one second less than the deadline to make up for
-        // overhead from the network.
+        // The timeout we give to the solver is a few milliseconds less than
+        // the deadline to make up for overhead from the network.
         let solver_timeout = timeout
-            .checked_sub(Duration::from_secs(1))
+            .checked_sub(Duration::from_millis(250))
```
fleupold, line 73: Given that solver timeout is later converted into `as_secs` I fear that this has no effect on the deadline (it will be rounded up to the original time in seconds).

nlordell: I'm not sure that is true. I think it would round down:
```
use std::time::Duration;
fn main() {
    let duration = dbg!(Duration::from_secs(2) - Duration::from_millis(250)); // = 1.75s
    dbg!(duration.as_secs()); // = 1
    dbg!(duration.as_secs_f64()); // = 1.75
}
```

I think the issue we are seeing with MIP is that it doesn't accept non-integer time limits:
```
2021-12-05T06:32:32.595Z  WARN solver::driver: solver Mip error: ... {"detail":[{"loc":["query","time_limit"],"msg":"value is not a valid integer","type":"type_error.integer"}]})
```

PR #1456, opened 2021-12-03
[3/-] Refactor pool storage to be generic

nlordell:
This PR refactors the `PoolStorage` to be generic over the kind of factory it is for.

The `PoolStorage` type is a registry of pools that were indexed by their emitted `PoolCreated` events (which is part of the `BasePoolFactory` base class that is shared across Balancer V2 pool factories).

Previously, we had a single `PoolStorage` type that we instantiated for all 3 pool factories, and each instance has its own maps of `pool_id => weighted_pool, pool_id => stable_pool`. This meant that, for example, weighted pool factories had an instance of `HashMap<H256, stable::PoolInfo>` which it never used. Additionally, adding a new pool type would require adding additional `HashMap<H256, new_pool::PoolInfo>` storage to the `PoolStorage` struct as well as duplicating a bunch of event indexing logic and tests.

Now, `PoolStorage` is generic on the Balancer pool factory that it is for, and uses the `FactoryIndexing` trait and its associated `PoolInfo` type for the pool storage `HashMap<H256, FactoryIndexing::PoolInfo>` instead of requiring `PoolStorage` to have a `HashMap` for each possible `*::PoolInfo` type.

As a result a bunch of code got removed and de-duplicated!

### Test Plan

Unit tests have been adapted to refactor. Additionally, ran a manual test re-indexing all pools since the early Cretaceous period with:
```
$ cargo test -p shared -- balancer_indexed_pool_events_match_subgraph --nocapture --ignored
test sources::balancer_v2::event_handler::tests::balancer_indexed_pool_events_match_subgraph ... ok
```


nlordell:
@fleupold hopefully, this PR will give some more context to the reasoning behind some of the changes in https://github.com/gnosis/gp-v2-services/pull/1449#pullrequestreview-821717122

crates/shared/src/sources/balancer_v2/info_fetching.rs
```diff
@@ -1,84 +0,0 @@
-//! Responsible for conversion of a `pool_address` into `*::PoolInfo` which is
```
bh2smith, line 1: Was the content of this file moved somewhere?

bh2smith (approved):
Amazing work here!

crates/shared/src/sources/balancer_v2/info_fetching.rs
nlordell: No, it was completely removed.

crates/shared/src/sources/balancer_v2/pools/common.rs
```diff
@@ -77,6 +94,59 @@ impl PoolInfoFetching for PoolInfoFetcher {
...
+        // Because of a `mockall` limitation, we **need** the future returned
+        // here to be `'static`. This requires us to clone and move `pool` into
+        // the async closure - otherwise it would only live for as long as
+        // `pool`, i.e. `'_`.
+        let pool = pool.clone();
+        async move {
+            let (paused, _, _) = paused.await?;
+            let swap_fee = Bfp::from_wei(swap_fee.await?);
+
+            let (token_addresses, balances, _) = balances.await?;
```
fleupold, line 128: Since the futures are instantiated concurrently beforehand, this is roughly as performant as joining them here isn't it (modulo that we could return an error early if the last one failed fast)?

crates/shared/src/sources/balancer_v2/pools.rs
```diff
@@ -34,10 +59,27 @@ pub trait FactoryIndexing {
...
+    /// Additionally, a block spec and a batch call context is passed in to
+    /// specify exactly the block number the state should be read for, and allow
+    /// for more optimal performance when fetching a large number of pools.
+    fn fetch_pool_state(
+        &self,
+        pool_info: Self::PoolInfo,
+        // This **needs** to be `'static` because of a `mockall` limitation
+        // where we can't use other lifetimes here.
+        // <https://github.com/asomers/mockall/issues/299>
+        pool_state: BoxFuture<'static, common::PoolState>,
```
fleupold, line 75: naming nit: common_pool_state?

fleupold (approved):
LGTM

Now the last 🧩 is who is calling `fetch_pool_state`

crates/shared/src/sources/balancer_v2/pools/common.rs
nlordell: Because these are already resolved when the `batch_call` future resolves, there should be no performance difference whether we `join` or not.

nlordell:
I just realized I accidentally pushed changes meant for a separate branch to this PR. 🙈

nlordell:
I accidentally pushed the pool fetching stuff to this branch when it was meant to be in a follow up PR (tests were missing for example).

I reverted all the `fetch_pool_state` implementation changes and will be making some more PRs to add those changes - the PR was too big.

crates/shared/src/sources/balancer_v2/event_handler.rs
```diff
@@ -156,93 +153,44 @@ impl BalancerPoolRegistry {
...
+where
+    Factory: FactoryIndexing,
+{
     async fn replace_events(
         &mut self,
-        events: Vec<EthContractEvent<WeightedPoolFactoryEvent>>,
+        events: Vec<EthContractEvent<BasePoolFactoryEvent>>,
         range: RangeInclusive<BlockNumber>,
     ) -> Result<()> {
-        self.replace_events_inner(
```
fleupold, line 170: Can this method be removed now?

I think one reason why we differentiated between replace and delete + insert was that the db transaction could be atomic so we don't end up in a state where past blocks are detailed but new blocks are not applied. Not sure how likely/relevant this is though compared to the added complexity.

crates/shared/src/sources/balancer_v2/event_handler.rs
```diff
@@ -156,93 +153,44 @@ impl BalancerPoolRegistry {
...
-impl EventStoring<StablePoolFactoryEvent> for PoolStorage {
-    async fn replace_events(
-        &mut self,
-        events: Vec<EthContractEvent<StablePoolFactoryEvent>>,
-        range: RangeInclusive<BlockNumber>,
-    ) -> Result<()> {
-        self.replace_events_inner(range.start().to_u64(), convert_stable_pool_created(events)?)
-            .await
-    }
+            self.index_pool_creation(pool_created, block_created)
```
fleupold, line 189: I can find this method in the code when checking out locally, but I don't remember having seen a PR on it. Maybe you can add the order of PRs to the description to make it easier to navigate back and forth?

crates/shared/src/sources/balancer_v2/pool_storage.rs
```diff
@@ -54,260 +50,195 @@ pub fn common_pool(seed: u8) -> CommonPoolData {
...
+                .or_default()
+                .insert(pool.common().id);
         }
-        Ok(())
+        self.pools.insert(pool.common().id, pool);
     }
 
-    pub async fn replace_events_inner(
+    /// Indexes a new pool creation event.
+    pub async fn index_pool_creation(
```
nlordell, line 150: It was added here.

crates/shared/src/sources/balancer_v2/event_handler.rs
nlordell: See 👇.

I will add orders to the titles.

crates/shared/src/sources/balancer_v2/event_handler.rs
nlordell: Right - we now just implement the logic in 
```rust
fn replace_events() {
    // logic here...
}
```

Instead of 
```rust
fn replace_events_inner() {
    // logic here...
}
fn replace_events() {
    replace_events_inner();
}
```

Since I found it was easier to implement `PoolStorage` in terms of:
- Add events
- Remove events newer than block

And test those two methods interdependently instead of testing `replace_events_inner` assuming that it called the two above methods.

crates/shared/src/sources/balancer_v2/event_handler.rs
nlordell: Sorry, this was a bit confusing. When you reviewed the PR, I had accidentally also pushed #1463 on top of this one - so it included more changes than I wanted.

PR #1458, opened 2021-12-05
More descriptive error upon transfer simulation failure when placing order

### LINKED ISSUE #1085: Differentiate between insufficient funds and missing approval
When external teams integrate with GPv2 (e.g. the yearn team) they tend to run into issues with the various checks in our order creation. To ensure they can fix the issue without asking us directly for help, we need descriptive error messages (which we already have for the most part)

However, specifically our "InsufficientBalance" error covers multiple cases:

1. Not enough balance
2. Approval not yet given to the vault relayer

I think it would good to use a different error message in the latter case.

fleupold:
Fixes #1085

We have gotten mutliple complaints from people trying to integrate with CowSwap that our order creation errors are the same (InsufficientFunds) in these two cases
1. User doesn't have enough balance of the token
2. User hasn't given sufficient allowance to the vault relayer

This PR makes these order creation errors distinct. In particular we use the following errors in the following scenarios:
1. InsufficientFunds - user doesn't have enough Balance
2. InsufficientAllowance - Vault Relayer doesn't have enough allowance on behalf of the user
3. SellAmountOverflows - sell amount (U256) + fee amount (U256) overflow U256 (very pathological case)
4. TransferSimulationFailure - transfer failed for other reasons (e.g. the token being paused at the moment)

In order to not create more roundtrips in the happy path (e.g. for frontends that make sure users cannot place orders without having given prior approval) we keep the transfer simulation (which covers all 3 cases in one call) and only check the exact failure reason in case it fails.

### Test Plan

Added an integration test for the balance fetcher functions. It would be nice to use the order placement script with a custom endpoint (so I can easily create test orders locally)


crates/orderbook/src/api/order_validation.rs
```diff
@@ -111,7 +111,11 @@ pub enum ValidationError {
     Partial(PartialValidationError),
     InsufficientFee,
     InsufficientFunds,
+    InsufficientAllowance,
     InvalidSignature,
+    // If fee and sell amount overflow u256
+    SellAmountOverflow,
```
nlordell, line 117: I would be OK with a generic 500-error in these cases (unless this actually is happening to our users).

crates/orderbook/src/api/order_validation.rs
```diff
@@ -336,11 +362,35 @@ impl OrderValidating for OrderValidator {
             )
             .await
             .unwrap_or(false)
+        {
+            return Ok((order, unsubsidized_fee));
+        }
+
+        if !self
```
nlordell, line 369: Do you think it maybe makes sense to change `BalanceFetching` to have something like:
```rust
trait BalanceFetching {
    fn simulate_transfer(...) -> Result<(), TransferError>;
}
enum TransferError {
    InsufficientBalance,
    InsufficientAllowance,
    Other(...),
}
```

The advantage of doing it this way is that we keep the details of testing transfers and the possible places it could go wrong in the `BalanceFetching` component.

Some other advantages:
- It allows us to use a JSON RPC batch call for the balance and allowance query
- We won't accidentally forget to update `has_sufficient_balance` to also pass in the `SellTokenBalance` so that it works with Vault internal balances once we support it.

crates/orderbook/src/api/order_validation.rs
```diff
@@ -336,11 +362,35 @@ impl OrderValidating for OrderValidator {
...
+            .balance_fetcher
+            .has_sufficient_balance(order_creation.sell_token, owner, min_balance)
+            .await
+            .unwrap_or(false)
         {
             return Err(ValidationError::InsufficientFunds);
```
nlordell, line 375: uber-naming-nit: This should probably be renamed to `InsufficientBalance` now.

crates/orderbook/src/account_balances.rs
```diff
@@ -215,6 +227,32 @@ impl BalanceFetching for Web3BalanceFetcher {
         };
         Ok(success)
     }
+
+    async fn has_sufficient_balance(&self, token: H160, from: H160, amount: U256) -> Result<bool> {
```
nlordell, line 231: In theory, this also needs to pass in the `SellTokenSource` because we would need to query the `Vault::internal_balance` in the case of `SellTokenSource::Internal`.

crates/orderbook/src/account_balances.rs
```diff
@@ -215,6 +227,32 @@ impl BalanceFetching for Web3BalanceFetcher {
...
+        amount: U256,
+        source: SellTokenSource,
+    ) -> Result<bool> {
+        let success = match source {
+            SellTokenSource::Erc20 => {
+                let instance = ERC20::at(&self.web3, token);
+                instance.allowance(from, self.vault_relayer).call().await? >= amount
+            }
+            SellTokenSource::External => {
+                self.can_manage_user_balance_call(token, from, amount).await
```
nlordell, line 250: One slight worry is that this has an implicit requirement that `balance_fetcher.has_sufficient_balance()` was previously checked, otherwise `has_sufficient_allowance` would generate false-positives when the user has a sufficient ERC20 allowance on the Vault and set relayer approval but has not more ERC20 token balance.

nlordell (commented):
In general code looks good modulo false positives in `has_sufficient_allowance` when the user has approval, but not enough ERC20 token balance. I think it should be possible to:
- batch the `ERC20::balance_of`, `ERC20::allowance` and `Vault::get_relayer_approval` calls to determine the reason we can't transfer
- hide the "complexities" of this check in the `account_balances` module and not have to "leak" it in the order creation logic.

bh2smith (approved):
Glad to finally have some refined error messages.

sunce86 (approved):
Lgtm, although I would agree with @nlordell comment about converting return value of `can_transfer` from `Result<bool>` to `Result<(), TransferError>`, would make the code a bit clearer.

fleupold:
> I would agree with @nlordell comment about converting return value of can_transfer from Result<bool> to Result<(), TransferError>, would make the code a bit clearer.

I removed `has_sufficient_balance` and `has_sufficient_allowance` in favour of a more expressive `can_transfer` return type (should have already been there for the last two reviews).

Would you be able to take another look if this works now?

crates/orderbook/src/account_balances.rs
```diff
@@ -25,10 +25,18 @@ impl Query {
     }
 }
 
+#[derive(Debug, PartialEq, Eq)]
+pub enum TransferSimulationResult {
+    Ok,
+    InsufficientAllowance,
+    InsufficientBalance,
+    TransferFailed,
+}
```
nlordell, line 34: ubernit: Not a fan of these `CustomResult` types because they miss out on all the `Result` goodies (`?` support, all the standard library methods like `map`, `unwrap_or_else`, etc.) - I would much prefer:
```suggestion
pub enum TransferSimulationError {
    InsufficientAllowance,
    InsufficientBalance,
    TransferFailed,
    Other(anyhow::Error),
}
```

And then return `Result<(), TransferSimulationError>`.

In terms of how to go from `anyhow::Error` to `TransferSimulationError`, you can either
```rust
impl From<anyhow::Error> for TransferSimulationError {
    fn from(err: anyhow::Error) -> Self {
        Self::Other(err)
    }
}
```

Or, you can manually do it at the call site:
```rust
erc20_balance_query(&mut batch, token, from, self.vault_relayer)
    .await
    .map_err(TransferSimulationError::Other)?;
```

crates/orderbook/src/api/order_validation.rs
```diff
@@ -748,7 +784,7 @@ mod tests {
                     .await
                     .unwrap_err()
             ),
-            "InsufficientFunds"
+            "SellAmountOverflow"
```
nlordell, line 787: nit: completely unrelated to this PR (so feel free to not change that here), but this makes the test rely on how `#[derive(Debug)]` and seems a bit flaky to me... I think it would be much better to:
```rust
assert!(matches!(
    validator.validate_and_construct_order(...).await,
    Err(ValidationError::SellAmountOverflow),
));
```

nlordell (approved):
LGTM - just a small nit regarding the return type which feels a bit "un-Rust-y". Not blocking though.

crates/orderbook/src/api/order_validation.rs
```diff
@@ -470,93 +508,74 @@ mod tests {
...
+        assert!(matches!(
+            validator
+                .partial_validate(PreOrderData {
+                    valid_to: legit_valid_to,
+                    buy_token: BUY_ETH_ADDRESS,
+                    ..Default::default()
+                })
+                .await,
+            Err(PartialValidationError::TransferEthToContract)
+        ));
```
nlordell, line 571: 🎉

nlordell (approved):
Awesome!

crates/orderbook/src/api/order_validation.rs
anxolin: Thanks @nlordell for notifying, indeed is relevant. 

https://github.com/gnosis/cowswap/pull/2010

PR #1459, opened 2021-12-05
Run price/fee estimation queries concurrently for buy orders

fleupold:
We are struggling with latency for our `post_quote` route, because we issue two long running requests to a) to estimate the fee and b) to estimate the amount the user can expect to receive on their trade. This PR tackles a low hanging fruit, which is to run these two queries concurrently for buy orders (since the queries are fully independent).

For sell orders, it's slightly more tricky as the correct sell amount for the "price estimation query" depends on the fee amount computed (creating a dependency). One way to work around this could be to get the expected price based on the full sell amount (including fee) and then adjust for the fee afterwards (e.g. if we can sell 10 GNO for 1ETH, but there is 1 GNO fee, the expected buy amount would be 0.9 ETH). This should:tm be safe, since for smaller amounts the price we can achieve should be strictly better (less slippage). 

I also noticed this comment for `SellAmount::AfterFee` I read this comment

```
// TODO: Nice to have: true sell amount after the fee (more complicated).
```

which I'm confused, since as far as I understand this type of query could be implemented analogous to the way we do buy orders.

Looking forward to hearing your thoughts.

### Test Plan
Existing unit tests, see roundtrip time for buy orders decrease significantly:

```
main
________________________________________________________
Executed in    8.65 secs   fish           external
   usr time    6.89 millis    0.22 millis    6.68 millis
   sys time   13.01 millis    4.52 millis    8.49 millis
```

```
this branch
________________________________________________________
Executed in    3.05 secs   fish           external
   usr time    6.25 millis    0.32 millis    5.93 millis
   sys time   17.51 millis   10.35 millis    7.15 millis
```


nlordell:
> I also noticed this comment for SellAmount::AfterFee I read this comment
>
> ```
> // TODO: Nice to have: true sell amount after the fee (more complicated).
> ```

Right - I remember this as well. I think this was because the `POST /quote` was implemented by moving the `GET /fee_and_quote` logic which only had support for `sell_amount_before_fees` and was left as a TODO.

In terms of it being more complicated, I agree that the price quoting logic is quite easy. We would just have to adapt our internal price estimation API to allow for it (and verify that stored fee measurements for these kinds of quotes work when the order gets created).

nlordell (approved):
Nice! Great idea.

bh2smith:
Yes that todo was left after the rerouting of legacy quote logic. I wasn't sure how we were suppose to handle it at the time, so it was left that way. You suggestions sound reasonable.

crates/orderbook/src/api/post_quote.rs
```diff
@@ -281,29 +282,27 @@ impl OrderQuoter {
                     return Err(FeeError::PriceEstimate(PriceEstimationError::ZeroAmount));
                 }
 
-                let (fee, expiration) = self
-                    .fee_calculator
-                    .compute_subsidized_min_fee(
+                let price_estimation_query = price_estimation::Query {
```
sunce86, line 285: I see that there are common structures between `orderbook` and `shared` (`FeeData` and `price_estimation::Query`), is there a plan to extract these structures to common library? If not, should these have at least the same name, even if their current name reveals their usage intention?

sunce86 (approved):
Lg

crates/orderbook/src/api/post_quote.rs
fleupold: Nice observation! I personally think that while the fields are mostly overlapping on these two structs, they still have  different semantics that are kind of "specific" to their module. E.g. in the future the fee estimation query may depends on things like app data or user address (users with staked CoW token may get a fee discount) whereas the price estimation is likely never really depend on anything else but tokens, order type and amount.

I'd be all for aligning the naming and documenting the structs better though.

PR #1460, opened 2021-12-05
Implement SellAfterFee quote

fleupold:
As discussed in #1459, the implementation of sell orders with `AmountAfterFee` is actually straight forward. This PR implements it analogous to how we compute the quote for buy orders.

### Test Plan
Adjusted existing unit test

Also to verify fee estimation works check out https://github.com/gnosis/gp-v2-contracts/pull/931 and run

```
yarn hardhat --network mainnet place-order sellBeforeFee --from 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 --to 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48 --amount-atoms 1000000000000000000 --api-url http://localhost:8080
```

and look for the following fee log: 

>2021-12-07T11:47:59.839Z DEBUG orderbook::fee: found past fee FeeParameters { gas_amount: 205826.0, gas_price: 70000000000.0, sell_token_price: 1.0 }
2021-12-07T11:47:59.839Z DEBUG orderbook::fee: given fee matches past fee

### DIFF
--- a/crates/orderbook/src/api/post_quote.rs
+++ b/crates/orderbook/src/api/post_quote.rs
@@ -6,7 +6,7 @@ use crate::{
     },
     fee::{FeeData, MinFeeCalculating},
 };
-use anyhow::{anyhow, Result};
+use anyhow::Result;
 use chrono::{DateTime, Utc};
 use ethcontract::{H160, U256};
 use futures::try_join;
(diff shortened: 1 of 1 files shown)


bh2smith (approved):

nlordell:
One open question that we had that wasn't super clear to me (and I think we should check) is that the stored fee measurements for these quotes works when the order is created.

That is, if you get a fee quote this way, and then create an order with it, the API won't re-compute a new fee estimate and sometimes return an "insufficient fee error" in cases where the price moves against the user.

It would be nice to verify this somehow in the test plan.

nlordell (approved):
Code looks good.

sunce86 (approved):
lg

fleupold:
> One open question that we had that wasn't super clear to me (and I think we should check) is that the stored fee measurements for these quotes works when the order is created.

The key is that on order validation we check for `find_measurement_including_larger_amount`

https://github.com/gnosis/oba-services/blob/d881757133a9f52c30f0d991d793800c680e5271/crates/orderbook/src/fee.rs#L389

This is mainly for estimates where we "assume" the full amount (before deducting fees) in order to estimate a fee. In this case it will even be an exact match.

I also did a local test, and was able to place an order reliably (added to testplan)

PR #1461, opened 2021-12-06
Fix the fix for timeout logic in HTTP solver client

taminomara:
The MIP solver uses integer for timeout parameter, so we have to reserve one second for network integration, otherwise the timeout gets rounded down, causing unexpected errors.

### Test Plan
1: check that both solvers can produce valid solutions:
```
$ cargo run -p solver -- \
  --mip-solver-url "$MIP_SOLVER_URL" \
  --quasimodo-solver-url "$QUASIMODO_SOLVER_URL" \
  --settle-interval 30 \
  --solver-account 0xa6DDBD0dE6B310819b49f680F65871beE85f517e \
  --solvers Mip,Quasimodo \
  --transaction-strategy DryRun
```
2: run quasimodo price estimation test on mainnet.


taminomara:
Note: this is a continuation of #1457.

fleupold (approved):
Change makes sense given these constraints. It would be great if you could work with the solver team to overcome this limitation going forward, e.g. by adding a new field that allows specifying the timeout in milliseconds.

taminomara:
@fleupold I've updated the code and the test plan, could you give it another look before I merge, please?

fleupold:
> @fleupold I've updated the code and the test plan, could you give it another look before I merge, please?

Just took another look. Code looks good. The description is missing context on why the change from deadline to timeout is better, could you explain please?

fleupold:
Thanks, this makes sense. I wonder if we can somehow capture this detail somewhere in the code as I could easily see someone reversing this change in the future (or maybe already in the past - let's see if there was another "subtle" reason to do it the way it used to be)

PR #1462, opened 2021-12-06
Add remaining base pool methods

nlordell:
This PR adds all missing Balancer `BasePool` methods to the artifact.

Because the `BasePool` source has changed over time - this was created by checking out the Balancer V2 repo at the commit where the `WeightedPoolFactory` was deployed and building the `BasePool` source there. I then had to manually remove `{decrease,increase}Approval` methods since they are no longer part of the `BasePool` (and were renamed to `{decrease,increase}Allowance` for Stable pools).

This will be used as an ABI for a common `BasePool` contract type containing all common methods between all Balancer pool types for common query methods.

### Test Plan

Make sure that the common methods are indeed common:
```
$ cat crates/contracts/artifacts/BalancerV2BasePool.json  | jq -r '.abi[].name' > base.txt
$ cat crates/contracts/artifacts/BalancerV2WeightedPool.json  | jq -r '.abi[].name' > weight.txt
$ cat crates/contracts/artifacts/BalancerV2StablePool.json  | jq -r '.abi[].name' > stable.txt
$ diff base.txt weight.txt | grep '<'
$ diff base.txt stable.txt | grep '<'
```

### DIFF
--- a/crates/contracts/artifacts/BalancerV2BasePool.json
+++ b/crates/contracts/artifacts/BalancerV2BasePool.json
@@ -1,6 +1,220 @@
 {
   "abi": [
     {
+      "anonymous": false,
+      "inputs": [
+        {
+          "indexed": true,
+          "internalType": "address",
+          "name": "owner",
+          "type": "address"
+        },
+        {
+          "indexed": true,
+          "internalType": "address",
+          "name": "spender",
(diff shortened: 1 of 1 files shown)


fleupold (approved):
🙈

PR #1464, opened 2021-12-06
Disable mergify

taminomara:
We don't use it anymore.


bh2smith (approved):

bh2smith:
I think we need to also uninstall mergify from the repo project settings.

taminomara:
@bh2smith I thought so, but I can't find any place where I could disable it. I guess I'll merge and see if it breaks or not 🤷🏻‍♀️

bh2smith:
I think we need to get devops to disable it.

PR #1465, opened 2021-12-07
Move common pool state fetching logic

nlordell:
This PR moves the fetching logic for all common pool data (the swap fee, whether or not its paused and current reserves) into the `pools::common` module. Additionally, I simplified some of the pool fetching code to try and get rid of some of the code duplication.

Currently, stable and weighted pool fetching logic are still in the `pool_cache` module where they were before. They will be moved to their respective `pools::*` modules as a follow up, it was just too many changes for a single PR.

As a separate note, we now, unfortunately, have `pools::common::PoolState` and `pool_fetching::CommonPoolState` types for representing common pool state information. The plan is to eventually phase out `pool_fetching::CommonPoolState` in favour of the former in a follow up. Again, this is because otherwise, it would have touched a much larger portion of the project and made the PR too big to review. 

### Test Plan

New unit tests were added for the fetching of the common pool data that was moved from the `pool_cache` module.

<details><summary>Also you can check that pool fetching still works</summary>

Apply patch for some extra logging:
```diff
diff --git a/crates/shared/src/sources/balancer_v2/pool_fetching.rs b/crates/shared/src/sources/balancer_v2/pool_fetching.rs
index 37399ba..e1b72b7 100644
--- a/crates/shared/src/sources/balancer_v2/pool_fetching.rs
+++ b/crates/shared/src/sources/balancer_v2/pool_fetching.rs
@@ -240,6 +240,11 @@ impl BalancerPoolFetching for BalancerPoolFetcher {
             .weighted_pool_reserve_cache
             .fetch(pool_ids, at_block)
             .await?;
+        tracing::info!(
+            weighted = %fetched_weighted_pools.len(),
+            stable = %fetched_stable_pools.len(),
+            "FETCHED_POOLS",
+        );
         // Return only those pools which are not paused.
         Ok(FetchedBalancerPools {
             stable_pools: filter_paused(fetched_stable_pools),
```

And run the solver:
```
$ cargo run -p solver -- --baseline-sources BalancerV2 --orderbook-url https://protocol-mainnet.gnosis.io/ --solver-account 0xa6DDBD0dE6B310819b49f680F65871beE85f517e
...
2021-12-07T21:23:18.043Z  INFO shared::sources::balancer_v2::pool_fetching: FETCHED_POOLS weighted=12 stable=1
```

</details>

### DIFF
--- a/crates/shared/src/sources/balancer_v2/pool_cache.rs
+++ b/crates/shared/src/sources/balancer_v2/pool_cache.rs
@@ -156,8 +156,9 @@ fn accumulate_handled_results<T>(results: Vec<Result<T>>) -> Result<Vec<T>> {
     results
         .into_iter()
         .filter_map(|result| match result {
+            Ok(value) => Some(Ok(value)),
             Err(err) if is_contract_error(&err) => None,
-            _ => Some(result.map(Into::into)),
+            Err(err) => Some(Err(err)),
         })
         .collect()
 }
--- a/crates/shared/src/sources/balancer_v2/pool_cache.rs
+++ b/crates/shared/src/sources/balancer_v2/pool_cache.rs
@@ -1,25 +1,23 @@
+use super::{
+    event_handler::BalancerPoolRegistry,
+    pool_fetching::{AmplificationParameter, BalancerPoolEvaluating, StablePool, WeightedPool},
+    pools::common,
+};
 use crate::{
+    ethcontract_error::EthcontractErrorType,
     recent_block_cache::{Block, CacheFetching, CacheKey, CacheMetrics, RecentBlockCache},
-    sources::{
-        balancer_v2::{
-            event_handler::BalancerPoolRegistry,
-            pool_fetching::{BalancerPoolEvaluating, StablePool, WeightedPool},
-            pool_storage::{RegisteredStablePool, RegisteredWeightedPool},
-            swap::fixed_point::Bfp,
-        },
-        uniswap_v2::pool_fetching::handle_contract_error,
-    },
     transport::MAX_BATCH_SIZE,
     Web3,
 };
 use anyhow::Result;
-use contracts::{BalancerV2StablePool, BalancerV2Vault, BalancerV2WeightedPool};
-use ethcontract::{batch::CallBatch, errors::MethodError, BlockId, Bytes, H160, H256, U256};
+use contracts::BalancerV2StablePool;
+use ethcontract::{batch::CallBatch, errors::MethodError, BlockId, H256};
+use futures::future;
 use std::{collections::HashSet, sync::Arc};
 
 pub struct PoolReserveFetcher {
     pool_registry: Arc<BalancerPoolRegistry>,
-    vault: BalancerV2Vault,
+    common_pool_fetcher: Arc<dyn common::PoolInfoFetching>,
     web3: Web3,
 }
 
@@ -28,11 +26,14 @@ pub trait BalancerPoolCacheMetrics: Send + Sync {
 }
 
 impl PoolReserveFetcher {
-    pub async fn new(pool_registry: Arc<BalancerPoolRegistry>, web3: Web3) -> Result<Self> {
-        let vault = BalancerV2Vault::deployed(&web3).await?;
+    pub async fn new(
+        pool_registry: Arc<BalancerPoolRegistry>,
+        common_pool_fetcher: Arc<dyn common::PoolInfoFetching>,
+        web3: Web3,
+    ) -> Result<Self> {
         Ok(Self {
             pool_registry,
-            vault,
+            common_pool_fetcher,
             web3,
         })
     }
@@ -71,7 +72,7 @@ impl CacheFetching<H256, WeightedPool> for PoolReserveFetcher {
         pool_ids: HashSet<H256>,
         at_block: Block,
     ) -> Result<Vec<WeightedPool>> {
-        let mut batch = CallBatch::new(self.web3.transport());
+        let mut batch = CallBatch::new(self.web3.transport().clone());
         let block = BlockId::Number(at_block.into());
         let weighted_pool_futures = self
             .pool_registry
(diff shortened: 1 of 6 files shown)


crates/shared/src/sources/balancer_v2/pools/common.rs
```diff
@@ -77,6 +94,59 @@ impl PoolInfoFetching for PoolInfoFetcher {
...
+        // here to be `'static`. This requires us to clone and move `pool` into
+        // the async closure - otherwise it would only live for as long as
+        // `pool`, i.e. `'_`.
+        let pool = pool.clone();
+        async move {
+            let (paused, _, _) = paused.await?;
+            let swap_fee = Bfp::from_wei(swap_fee.await?);
+
+            let (token_addresses, balances, _) = balances.await?;
+            ensure!(pool.tokens == token_addresses, "pool token mismatch");
```
bh2smith, line 129: Was this constraint introduced here? Do you have reason to believe this would/could ever happen?

crates/shared/src/sources/balancer_v2/pools/common.rs
```diff
@@ -189,7 +274,132 @@ mod tests {
...
             }
-        )
+        );
+    }
+
+    #[tokio::test]
+    async fn fetch_pool_state() {
+        let pool_id = H256([0x90; 32]);
+        let tokens = [H160([1; 20]), H160([2; 20]), H160([3; 20])];
+        let balances = [bfp!("1000.0"), bfp!("10.0"), bfp!("15.0")];
```
bh2smith, line 284: cool macro... was this the only reason for introducing it? Is this easier than `bfp::from`?

crates/shared/src/sources/balancer_v2/pools/weighted.rs
```diff
@@ -64,10 +64,6 @@ mod tests {
     use ethcontract::{H160, H256};
     use ethcontract_mock::Mock;
 
-    fn bfp(amount: &str) -> Bfp {
-        amount.parse().unwrap()
-    }
```
bh2smith, line 69: I see, the macro is better than this... right?

bh2smith (approved):
Absolutely fantastic. my favourite file improvement is `pool_cache.rs`

crates/shared/src/sources/balancer_v2/pools/common.rs
nlordell: It was introduced here. It was just more of an "this is an in-code documentation of an assumption we make" kind of thing.

crates/shared/src/sources/balancer_v2/pools/common.rs
nlordell: Hmm good question! Didn't know we had a `From` conversion for strings to Balancer fixed point numbers.

I had seen `"...".parse::<Bfp>().unwrap()` before.

crates/shared/src/sources/balancer_v2/pools/weighted.rs
nlordell: I wouldn't say its better... Its kind of arbitrary here in my opinion. The reason I went with a macro here is that:
- Its primarily used for testing
- No `use ...::bfp` in the test modules

A function would have been fine as well.

crates/shared/src/sources/balancer_v2/pools/common.rs
nlordell: I checked and indeed there is no `From` conversion for strings implemented (which makes sense because it can fail).

fleupold (approved):
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2F3ohjUS0WqYBpczfTlm%2Fgiphy.gif&token=rebuild" data-gitmeme-token="rebuild"></a>

crates/shared/src/sources/balancer_v2/pool_cache.rs
```diff
@@ -71,50 +72,26 @@ impl CacheFetching<H256, WeightedPool> for PoolReserveFetcher {
         pool_ids: HashSet<H256>,
         at_block: Block,
     ) -> Result<Vec<WeightedPool>> {
-        let mut batch = CallBatch::new(self.web3.transport());
+        let mut batch = CallBatch::new(self.web3.transport().clone());
```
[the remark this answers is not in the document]
nlordell: Just because we made the type alias be `type Web3CallBatch = CallBatch<Web3Transport>` instead of `CallBatch<&Web3Transport>`. The choice was just to simplify lifetimes and `mockall` interoperability (it has the same issue as `BoxFuture<'a, ...>` not being able to specify the `'a`). Since transport are generally `Arc<...>` wrapped anyway, this should not have a significant impact.

crates/shared/src/sources/balancer_v2/pool_cache.rs
```diff
@@ -186,163 +152,48 @@ impl CacheMetrics for Arc<dyn BalancerPoolCacheMetrics> {
...
         .into_iter()
-        .try_fold(Vec::new(), |mut acc, fetched_pool| {
-            match fetched_pool.handle_results()? {
-                None => return Ok(acc),
-                Some(pool) => acc.push(pool),
-            }
-            Ok(acc)
+        .filter_map(|result| match result {
+            Err(err) if is_contract_error(&err) => None,
+            _ => Some(result.map(Into::into)),
```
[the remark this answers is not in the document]
nlordell: Here `result` can be `Ok(pool)` or `Err(err) if !is_contract_error(&err)`. This was a semantic change. I will split it out to be more obvious.

crates/shared/src/sources/balancer_v2/pools/common.rs
```diff
@@ -77,6 +94,59 @@ impl PoolInfoFetching for PoolInfoFetcher {
...
+            .block(block)
+            .batch_call(batch);
+
+        // Because of a `mockall` limitation, we **need** the future returned
+        // here to be `'static`. This requires us to clone and move `pool` into
+        // the async closure - otherwise it would only live for as long as
+        // `pool`, i.e. `'_`.
+        let pool = pool.clone();
+        async move {
+            let (paused, _, _) = paused.await?;
```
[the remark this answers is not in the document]
nlordell: One thing that `ether.js` does is to add requests to a buffer of RPC calls and execute them once every 100ms or so. We can do the something similar.

crates/shared/src/sources/balancer_v2/pool_cache.rs
```diff
@@ -186,163 +152,48 @@ impl CacheMetrics for Arc<dyn BalancerPoolCacheMetrics> {
...
-        };
-
-        Ok(Some(WeightedPool::new(
-            self.registered_pool,
-            balances,
-            Bfp::from_wei(swap_fee_percentage),
-            paused,
-        )))
-    }
+fn is_contract_error(err: &anyhow::Error) -> bool {
```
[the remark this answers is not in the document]
bh2smith: It was somewhere else, but has been simplified here.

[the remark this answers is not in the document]
nlordell: Exactly.

PR #1466, opened 2021-12-07
Move pool fetching logic to their own submodules

nlordell:
Follow up from #1465 

This PR moves pool fetching logic into their own submodules.

This PR has a few "oddities" that should be clarified:

#### Duplicate Pool State Types

One oddity is that we now have `common::PoolState`/`CommonPoolState`, `weighted::PoolState`/`WeightedPool` and `stable::PoolState`/`StablePool` types. Eventually the `CommonPoolState` and `{Weighted,Stable}Pool` types will go away in favour of the new ones, I just couldn't fit it into this PR. It will be done in a _follow-up PR_ coming soon to a Github near you!

#### Additional Complexity in `pool_cache` Module

The code in `pool_cache` module got a bit more complicated 😢. This is a "necessary evil" in order to make everything generic (note that we now can queue necessary `eth_call`s into a batch for fetching specialized pool information unique to stable and weighted pools, and that the `fetch_values` implementations for weighted and stable pools are almost identical with the exception of `s/weighted/stable/g`). Actually making this type generic will be the _next PR_.

#### `swap_fee` Duplicated For All `PoolState`s

`swap_fee` is common to both weighted and stable pools. It was duplicated to each `PoolState` so that we can implement `BaselineSolvable` for each individual `PoolState` without requiring any additional auxiliary data. This way, the `balancer_v2::pools::Pool` type can just contain common data that is important for interaction encoding, while each `*::PoolState` can have all the data required for computing input and output amounts.

#### Shared `common::PoolState` Future.

One final technical note of importance - `PoolInfoFetching::fetch_common_pool_state` returns a future that resolves to a `Result<common::PoolState>`, but we pass in a future that resolves to a `common::PoolState` (not a result) to the `FactoryIndexing::fetch_pool_state` method! The reason for this choice was that handling errors in fetching the common pool state should be handled by the caller and not by the `FactoryIndexing` implementation. In order to make this happen, we needed a `share_common_pool_state` method that would take a `Future<Output = Result<common::PoolState>` and produce a tuple of futures `(Future<Output = Result<common::PoolState>>, Future<Output = common::PoolState>)` where `poll`-ing the latter future would panic if the first future doesn't resolve to a success value. This means that using `share_common_pool_state` comes with some assumptions that if not upheld can cause panics. Since this is private and only used in the `pool_cache` module and these assumptions don't leak at all, I thought it was OK (especially since it allows for a cleaner `FactoryIndexing` interface which is arguably more important since future pool types will implement this and not require touching code in `pool_cache`). 

### Test Plan

Added a new tests for the specialized pool fetching logic that didn't exist before as well as new tests for the slightly "weird" `share_common_pool_state`

### DIFF
--- a/crates/shared/src/sources/balancer_v2/pool_cache.rs
+++ b/crates/shared/src/sources/balancer_v2/pool_cache.rs
@@ -1,7 +1,7 @@
 use super::{
     event_handler::BalancerPoolRegistry,
-    pool_fetching::{AmplificationParameter, BalancerPoolEvaluating, StablePool, WeightedPool},
-    pools::common,
+    pool_fetching::{BalancerPoolEvaluating, StablePool, WeightedPool},
+    pools::{common, FactoryIndexing},
 };
 use crate::{
     ethcontract_error::EthcontractErrorType,
@@ -10,14 +10,17 @@ use crate::{
     Web3,
 };
 use anyhow::Result;
-use contracts::BalancerV2StablePool;
+use contracts::{BalancerV2StablePoolFactory, BalancerV2WeightedPoolFactory};
 use ethcontract::{batch::CallBatch, errors::MethodError, BlockId, H256};
-use futures::future;
-use std::{collections::HashSet, sync::Arc};
+use futures::{future, FutureExt as _};
+use std::{collections::HashSet, future::Future, sync::Arc};
+use tokio::sync::oneshot;
 
 pub struct PoolReserveFetcher {
     pool_registry: Arc<BalancerPoolRegistry>,
     common_pool_fetcher: Arc<dyn common::PoolInfoFetching>,
+    weighted_pool_factory: BalancerV2WeightedPoolFactory,
+    stable_pool_factory: BalancerV2StablePoolFactory,
     web3: Web3,
 }
 
@@ -31,9 +34,14 @@ impl PoolReserveFetcher {
         common_pool_fetcher: Arc<dyn common::PoolInfoFetching>,
         web3: Web3,
     ) -> Result<Self> {
+        let weighted_pool_factory = BalancerV2WeightedPoolFactory::deployed(&web3).await?;
+        let stable_pool_factory = BalancerV2StablePoolFactory::deployed(&web3).await?;
+
         Ok(Self {
             pool_registry,
             common_pool_fetcher,
+            weighted_pool_factory,
+            stable_pool_factory,
             web3,
         })
     }
@@ -80,13 +88,28 @@ impl CacheFetching<H256, WeightedPool> for PoolReserveFetcher {
             .await
             .into_iter()
             .map(|registered_pool| {
-                let common_pool_state = self.common_pool_fetcher.fetch_common_pool_state(
-                    &registered_pool.common,
+                let (common_pool_state, common_pool_state_ok) =
+                    share_common_pool_state(self.common_pool_fetcher.fetch_common_pool_state(
+                        &registered_pool.common,
+                        &mut batch,
+                        block,
+                    ));
+                let weighted_pool_state = self.weighted_pool_factory.fetch_pool_state(
+                    &registered_pool,
+                    common_pool_state_ok.boxed(),
                     &mut batch,
                     block,
                 );
 
-                async move { Ok(WeightedPool::new(registered_pool, common_pool_state.await?)) }
+                async move {
+                    let common_pool_state = common_pool_state.await?;
+                    let weighted_pool_state = weighted_pool_state.await?;
+                    Ok(WeightedPool::new(
+                        registered_pool,
+                        common_pool_state,
+                        weighted_pool_state,
+                    ))
+                }
             })
             .collect::<Vec<_>>();
         batch.execute_all(MAX_BATCH_SIZE).await;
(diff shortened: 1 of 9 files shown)


crates/shared/src/sources/balancer_v2/pool_fetching.rs
```diff
@@ -86,33 +80,6 @@ impl BalancerPoolEvaluating for WeightedPool {
     }
 }
 
-#[derive(Clone, Debug, PartialEq)]
```
nlordell, line 89: This just moved to `shared::sources::balancer_v2::pools::stable` without any change.

crates/shared/src/sources/balancer_v2/pool_fetching.rs
```diff
@@ -298,27 +265,4 @@ mod tests {
         assert_eq!(filtered_pools.len(), 1);
         assert_eq!(filtered_pools[0].common.id, pools[1].common.id);
     }
-
```
nlordell, line 301: This just moved to `shared::sources::balancer_v2::pools::stable` without any change.

crates/shared/src/sources/balancer_v2/pools/stable.rs
```diff
@@ -22,20 +33,164 @@ impl PoolIndexing for PoolInfo {
     }
 }
 
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct PoolState {
+    pub tokens: BTreeMap<H160, common::TokenState>,
+    pub swap_fee: Bfp,
```
nlordell, line 39: `swap_fee` is common to both weighted and stable pools. It was duplicated to each `PoolState` so that we can implement `BaselineSolvable` for each individual `PoolState` without requiring any additional auxiliary data.

This way, the `balancer_v2::pools::Pool` type can just contain common data that is important for interaction encoding, while each `*::PoolState` can have all the data required for computing input and output amounts.

bh2smith (approved):
For such a complicated PR, it was actually not so difficult to read through. Things are looking up for our Balancer integration.

crates/shared/src/sources/balancer_v2/pools/stable.rs
```diff
@@ -22,20 +33,164 @@ impl PoolIndexing for PoolInfo {
...
 #[async_trait::async_trait]
 impl FactoryIndexing for BalancerV2StablePoolFactory {
     type PoolInfo = PoolInfo;
+    type PoolState = PoolState;
 
     async fn specialize_pool_info(&self, pool: common::PoolInfo) -> Result<Self::PoolInfo> {
         Ok(PoolInfo { common: pool })
     }
+
+    fn fetch_pool_state(
```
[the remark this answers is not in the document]
nlordell: `get_pool_tokens` returns the reserves. This was the same as the old code:

https://github.com/gnosis/gp-v2-services/blob/8189f04dba1c8f9a88a2191374799649b26fcf85/crates/shared/src/sources/balancer_v2/pool_cache.rs#L88-L92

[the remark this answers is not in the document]
bh2smith: Maybe we could rename this to get pool reserves which implies the tokens must code along with it.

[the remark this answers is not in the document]
nlordell: note that `getPoolTokens` is the name of the contract method we are calling on the Vault here. My worry about renaming it is that it will cause confusion where someone sees we are calling `get_reserves` and then go to the `IVault.sol` file and don't see that method implemented anywhere.

crates/shared/src/sources/balancer_v2/pool_cache.rs
```diff
@@ -31,9 +34,14 @@ impl PoolReserveFetcher {
         common_pool_fetcher: Arc<dyn common::PoolInfoFetching>,
         web3: Web3,
     ) -> Result<Self> {
+        let weighted_pool_factory = BalancerV2WeightedPoolFactory::deployed(&web3).await?;
```
fleupold, line 37: Is it the plan to keep this component manage one instance of each pool type factory or will this eventually also be united behind a single component per pool type?

crates/shared/src/sources/balancer_v2/pool_cache.rs
```diff
@@ -170,10 +190,49 @@ fn is_contract_error(err: &anyhow::Error) -> bool {
...
 
+/// An internal utility method for sharing the success value for an
+/// `anyhow::Result`.
+///
+/// Typically, this is pretty trivial using `FutureExt::shared`. However, since
+/// `anyhow::Error: !Clone` we need to use a different approach.
+///
+/// # Panics
+///
+/// Polling the future with the shared success value will panic if the result
```
fleupold, line 201: This is beyond my level of rust knowledge, but is there not something we can do in terms of lifetimes to indicate that the second future has to outlive the first one?

fleupold (approved):
Thanks for the context, while I find it quite hard to understand the "timing dependencies" between the futures I agree that given it's visibility it's OK (I was hoping the rust type system has a recipe for this but I guess not)

crates/shared/src/sources/balancer_v2/pool_cache.rs
nlordell: Correct. We will make an object-safe interface per factory (i.e. `Arc<dyn Thing>` - what I will call a `Registry`) that we can aggregate in a list `Vec<Arc<dyn Thing>>` (what I will call `AggregatePoolFetcher`). This way, adding new pools envolves:
- defining info and pool fetching details (i.e. implement `FactoryIndexing`)
- instantiating a new `Registry` component and adding it to our `AggregateFetcher`'s its list of registries.

crates/shared/src/sources/balancer_v2/pool_cache.rs
nlordell: Maybe... I couldn't think of anything cleaver though.

PR #1469, opened 2021-12-08
Make quote sell_amount_before_fee concurrent

fleupold:
**This may seem like a small PR but it may actually have severe impact, so review with caution.**

I'm trying to improve the roundtrip times for our post_quote endpoint as we are seeing some really bad worse case latencies (alerts of 95% taking longer than 5s are frequent).

This PR tries to parallelize the fee estimation and amount estimation query of a sell order where the amount is given before_fee. Normally, this requires two dependent queries: We first need to fetch the fee amount to deduct it from the sell amount (yielding sell_amount_after_fee) and can then use that amount to get the expected buy_amount_after_fee.

This PR changes the logic to use the original `sell_amount_before_fee` for the price estimation query. The returned buy_amount would be too much since the user is deducting the fee from the original sell amount. However, the price ratio can be used as a proxy (selling a larger amount should have a similar but strictly worse price than selling a smaller amount).

The risk I see is that the API will return too pessimistic prices for small amounts on illiquid tokens (where the fee is taking a large chunk of the trade and the price moves quickly when the amounts change slightly) as it doesn't account for the price impact that deducting the fee has. I personally feel like this is a risk worthwhile taking in favor of a snappier API but I think this needs to be discussed.

Evidence that cumulative requests are taking long (example of a 8s+ roundtrip): https://logs.gnosis.io/goto/bf8f9d2f1b2ecd9801e60816e3354c78

### Test Plan
Unit tests + performance tests (make sure no fee measurements are stored):

```
time curl 'http://localhost:8080/api/v1/quote' \                                                                                                                                  
                                      --data '{"sellToken": "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48", "buyToken": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",  "validTo": 4294967295,  "appData": "0x0000000000000000000000000000000000000000000000000000000000000000",  "partiallyFillable": false,  "sellTokenBalance": "erc20",  "buyTokenBalance": "erc20",  "from": "0x2CaEf7f0EE82fb0ABf1AB0dCD3A093803002e705",  "kind": "sell",  "sellAmountBeforeFee": "2000000000"}'
```

Cold performance before change: 7.66 s
Cold performance after change: 3.61s


crates/orderbook/src/api/post_quote.rs
```diff
@@ -231,36 +231,41 @@ impl OrderQuoter {
...
-                    })
-                    .await
-                    .map_err(FeeError::PriceEstimate)?;
+                let buy_amount_after_fee =
+                    match estimate.out_amount.checked_mul(sell_amount_after_fee) {
+                        // sell_amount_before_fee is at least 1 (cf. above)
+                        Some(product) => product / sell_amount_before_fee,
+                        None => (estimate.out_amount / sell_amount_before_fee)
+                            .checked_mul(sell_amount_after_fee)
+                            .unwrap_or(U256::MAX),
```
nlordell, line 265: Fancy way to avoid overflows on large amounts... My worry is that the error is quite high in these cases (`sell_amount_after_fee` - which is already very large if the overflow happens) and that we may run into some pathological cases. I would suggest just
- returning an error if this happens
- using `BigInt` for the math and doing a saturating cast to `U256`.

nlordell (approved):
Very pragmatic approach! Love it!

nlordell:
> The risk I see is that the API will return too pessimistic prices for small amounts on illiquid tokens (where the fee is taking a large chunk of the trade and the price moves quickly when the amounts change slightly) as it doesn't account for the price impact that deducting the fee has. I personally feel like this is a risk worthwhile taking in favor of a snappier API but I think this needs to be discussed.

I think this would mostly affect illiquid pools on xDai right? This may give us some pretty bad prices in those cases.

One thing to note, is that we should have pretty snappy price estimates on xDai already (because we only estimate with the Baseline solver) - so maybe this is something that can be enabled/disabled with a switch?

Another approach would be to "lower" the choice of parallelizing the fee computation and expected sell amount into the each `PriceEstimator`. This has a couple of advantages:
- We can keep more accurate price estimations with our snappy estimators (`Baseline` for example). This way we are not as affected on xDai (which I would argue 
- Once we get an `ETH-USD` price feed - it would be possible to avoid the additional query buying 1ETH with `sell_token` for 0x and ParaSwap. If we already "lowered" this logic into the price estimators, it may be easier to implement this.

bh2smith (approved):

josojo:
I have overseen this PR. Like the pragmatic appraoch as well!

PR #1471, opened 2021-12-08
Run fee estimation queries concurrently

fleupold:
Another PR to speed up our API queries. In particular, when we estimate the fee we issue two queries
1.  to estimate the fas used for a single order settlement selling buy token+amount for sell token+amount
2. to estimate the price of the sell token in native token (to price the gas amount in sell token instead of native token)

Those queries may use external APIs and can be long running, therefore it should be better to run them in parallel (they are independent) 

### Test Plan
Existing unit test + seeing performance increase when measuring roundtrip time for post_quote (although this benchmark is not very scientific and seemed to have high variance)

Before (current main): 7.71s
With this PR: 3.30s


nlordell (approved):
LGTM

crates/orderbook/src/fee.rs
```diff
@@ -281,27 +282,29 @@ impl MinFeeCalculator {
...
             buy_token: self.native_token,
             in_amount: self.native_token_price_estimation_amount,
             kind: OrderKind::Buy,
         };
-        let estimate = self.price_estimator.estimate(&query).await?;
-        let sell_token_price = estimate.price_in_sell_token_f64(&query);
+        let (gas_estimate, buy_token_estimate, native_token_estimate) = try_join!(
+            self.gas_estimator
+                .estimate()
+                .map_err(PriceEstimationError::from),
```
sunce86, line 300: Do we need to explicitly call `map_err` here? Previously we did not.

sunce86 (approved):
Lgtm

crates/orderbook/src/fee.rs
fleupold: `try_join` expects the same return type for all futures and `gas_estimator.estimate` returns and anyhow error whereas the other one returns a PriceEstimationError (and the compiler doesn't seem smart enough to auto convert the type.

PR #1473, opened 2021-12-09
Make pool info fetching component generic on its factory

nlordell:
Follow up to #1466 and #1465 

The `common::PoolInfoFetcher` is now generic on a factory type. This means that external consumers of this API can `fetch_pool_info` and `fetch_pool` directly instead of having to first `fetch_common_{pool_info, pool_state}` and then `fetch_specialized_{pool_info, pool_state}`.

This also has the advantage that setting up mocks for components like the `PoolStorage` are easier now, since we only need to setup mock for a `MockPoolInfoFetching` instead of setting up mocks for both a `MockPoolInfoFetcher` (for fetching common pool info) and `MockFactoryIndexing` (for specializing the pool info).

This also allowed us to make `PoolReserveFetcher` generic on a `Factory` type 🎉. Note that we still provide specialized implementations for `CacheFetching<H256, WeightedPool>` and `CacheFetching<H256, StablePool>`, but this will change once we start using the common `balancer_v2::pools::Pool` type - which is also coming in a follow up.

### Test Plan

Mostly just refactoring making things generic. Added some new unit tests to test fetching specialized state also when fetching paused pools (which behaves slightly differently now).

### DIFF
--- a/crates/shared/src/sources/balancer_v2/pool_cache.rs
+++ b/crates/shared/src/sources/balancer_v2/pool_cache.rs
@@ -94,14 +94,13 @@ impl CacheFetching<H256, WeightedPool> for PoolReserveFetcher<BalancerV2Weighted
 
                 async move {
                     let pool_status = pool_status.await?;
-                    let weighted_pool = match pool_status.active() {
+                    match pool_status.active() {
                         Some(Pool {
                             kind: PoolKind::Weighted(state),
                             ..
-                        }) => WeightedPool::new_unpaused(registered_pool, state),
-                        _ => return Ok(None),
-                    };
-                    Ok(Some(weighted_pool))
+                        }) => Ok(Some(WeightedPool::new_unpaused(registered_pool, state))),
+                        _ => Ok(None),
+                    }
                 }
             })
             .collect::<Vec<_>>();
@@ -133,14 +132,13 @@ impl CacheFetching<H256, StablePool> for PoolReserveFetcher<BalancerV2StablePool
 
                 async move {
                     let pool_status = pool_status.await?;
-                    let stable_pool = match pool_status.active() {
+                    match pool_status.active() {
                         Some(Pool {
                             kind: PoolKind::Stable(state),
                             ..
-                        }) => StablePool::new_unpaused(registered_pool, state),
-                        _ => return Ok(None),
-                    };
-                    Ok(Some(stable_pool))
+                        }) => Ok(Some(StablePool::new_unpaused(registered_pool, state))),
+                        _ => Ok(None),
+                    }
                 }
             })
             .collect::<Vec<_>>();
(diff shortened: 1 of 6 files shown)


crates/shared/src/sources/balancer_v2/pool_cache.rs
```diff
@@ -88,27 +88,20 @@ impl CacheFetching<H256, WeightedPool> for PoolReserveFetcher {
...
-                        registered_pool,
-                        common_pool_state,
-                        weighted_pool_state,
-                    ))
+                    let pool_status = pool_status.await?;
+                    let weighted_pool = match pool_status.active() {
+                        Some(Pool {
+                            kind: PoolKind::Weighted(state),
+                            ..
+                        }) => WeightedPool::new_unpaused(registered_pool, state),
```
fleupold, line 101: nanonit: I find it slightly easier to read when both return points are on the same branch level
```suggestion
                        }) => return Ok(Some(WeightedPool::new_unpaused(registered_pool, state))),
```

crates/shared/src/sources/balancer_v2/pools.rs
```diff
@@ -34,6 +34,36 @@ pub enum PoolKind {
...
+        impl From<$from> for PoolKind {
+            fn from(state: $from) -> Self {
+                Self::$variant(state)
+            }
+        }
+    };
+}
+
+impl_from_state!(weighted::PoolState, Weighted);
+impl_from_state!(stable::PoolState, Stable);
```
fleupold, line 48: clever!

crates/shared/src/sources/balancer_v2/pools/common.rs
```diff
@@ -224,20 +279,64 @@ fn scaling_exponent_from_decimals(decimals: u8) -> Result<u8> {
         .ok_or_else(|| anyhow!("unsupported token with more than 18 decimals"))
 }
 
+/// An internal utility method for sharing the success value for an
```
fleupold, line 282: Isn't this code already part of a previous PR?

fleupold (commented):
Looks good overall to me! I feel something is off with the rebase (some code I have seen before)

crates/shared/src/sources/balancer_v2/pools/common.rs
nlordell: It just moved here.

crates/shared/src/sources/balancer_v2/pool_cache.rs
```diff
@@ -190,101 +175,31 @@ fn is_contract_error(err: &anyhow::Error) -> bool {
     )
 }
 
-/// An internal utility method for sharing the success value for an
```
nlordell, line 193: It used to be here

bh2smith (approved):

bh2smith:
Sorry I forgot to review this earlier.

PR #1475, opened 2021-12-09
New system for reporting API metrics

taminomara:
Fixes over-reporting of API metrics for non-200 requests.

So far, we've tried two ways of reporting warp metrics. Both had issues.

The first version used the `map` filter. It didn't report requests that didn't end with an `Ok`. This is because `map` don't handle errors, and `map_err` is not available in the public warp interface. It also didn't report requests for URLs that don't match any endpoints.

The second version used one `warp::log::custom` for each endpoint. It ended up over-reporting 404s. This is becuase of how warp filters work. Consider the following setup:

```rust
    path!("endpoint_1").with(handle_metrics_for_endpoint_1)
.or(path!("endpoint_2").with(handle_metrics_for_endpoint_2))
.or(path!("endpoint_3").with(handle_metrics_for_endpoint_3))
```

In this case we have the following class structure:

```

or ---- handle_metrics_for_endpoint_1 ---- path!("endpoint_1")
|
or ---- handle_metrics_for_endpoint_2 ---- path!("endpoint_2")
|
+------ handle_metrics_for_endpoint_3 ---- path!("endpoint_3")

```

If we request `endpoint_3`, warp will run all filters top-to-bottom. It will first descend into `path!("endpoint_1")`. Then `path!("endpoint_1")` will reject url `endpoint_3` and bail an error into `handle_metrics_for_endpoint_1`. The `handle_metrics_for_endpoint_1` will report a 404 for endpoint 1 and bail an error into the first `or`. The first `or` will then call its right branch, eventually descending into `path!("endpoint_2")`... By the time we reach `path!("endpoint_3")`, a whole bunch of 404s would've been already reported.

So, the only correct way to report a request is by wrapping the top-level filter into a `warp::log::custom`.

For this to work correctly, we'll need to sanitize paths by removing variable parameters from them (i.e. replace `/api/v1/orders/<order_id>` with `/api/v1/orders/*`). I've tried my best to come up with a solution that don't require explicitly listing all allowed URLs, but failed. Unfortunately, we can't inspect warp filters and extract all endpoints and their parameters. Therefore, we have to create an explicit list of all valid URLs, and pass it to the sanitizer. I don't like it, but I don't know how to improve it.

### Test Plan
New and existing unit tests.
Manual testing:
- check that 200s are reported correctly for multiple API endpoints,
- check that 400s and 404s are reported correctly for multiple API endpoints,
- check that requests that don't match any endpoint are reported correctly.

### DIFF
--- a/crates/orderbook/src/api.rs
+++ b/crates/orderbook/src/api.rs
@@ -10,6 +10,7 @@ mod get_solvable_orders;
 mod get_solvable_orders_v2;
 mod get_trades;
 mod get_user_orders;
+mod metrics;
 pub mod order_validation;
 pub mod post_quote;
 
@@ -18,7 +19,7 @@ use crate::{
 };
 use anyhow::{Error as anyhowError, Result};
 use serde::{de::DeserializeOwned, Serialize};
-use shared::{metrics::get_metric_storage_registry, price_estimation::PriceEstimationError};
+use shared::price_estimation::PriceEstimationError;
 use std::fmt::Debug;
 use std::{convert::Infallible, sync::Arc};
 use warp::{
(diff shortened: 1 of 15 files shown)


crates/orderbook/src/api.rs
```diff
@@ -32,48 +33,110 @@ pub fn handle_all_routes(
...
-        get_markets::get_amount_estimate(quoter.price_estimator.clone()).boxed();
-    let get_fee_and_quote_sell = get_fee_and_quote::get_fee_and_quote_sell(quoter.clone()).boxed();
-    let get_fee_and_quote_buy = get_fee_and_quote::get_fee_and_quote_buy(quoter.clone()).boxed();
-    let get_user_orders = get_user_orders::get_user_orders(orderbook.clone()).boxed();
-    let get_orders_by_tx = get_orders_by_tx::get_orders_by_tx(orderbook).boxed();
-    let post_quote = post_quote::post_quote(quoter).boxed();
+    // api/v1
+    let create_order = (
+        "/api/v1/orders",
+        create_order::create_order(orderbook.clone()),
```
taminomara, line 39: I'm not sure if it's better to store the `"/api/v1/orders"` string here, or if `create_order::create_order` should return it as well. The concern here is that `create_order::create_order` doesn't know about the `"/api/v1"` prefix, hence it shouldn't return the `"/api/v1/orders"` string. On the other hand, we shouldn't rely on the contents of `create_order::create_order` function on the call site, hence we can't rely on the fact that it returns a filter for path `"/orders"`.

taminomara: One solution could be to move the `/api/v1/` part of the url into the `create_order::create_order` function.

nlordell (commented):
Very nice PR description!

One thing that may be worth exploring is to map the each of the "..._response" filters with metrics handling. Not sure if its possible.

fleupold:
Agree, that this is an excellent description of a very tricky issue. 

> The first version used the map filter. It didn't report requests that didn't end with an Ok.

Is this a problem though? It would only miss `Rejections` and the most common reason for rejection is that the URL pattern didn't match (which is equivalent to someone escaping the set of regexs that is now populated), isn't it? Even if an order isn't found we manually return an `Ok(reply)`, where the reply has status 400. Maybe I'm missing something.

taminomara:
# About API metrics reporting

CC: @nlordell @fleupold @vkgnosis

TL;DR: this is over-complicated, I'll go with a simpler version.


## Part 1: how warp filters work, exactly

Filters are small pieces of logic that encode how to process a request. Each filter can take a request, and return a transformed version of it, or an error. Filters can be nested to build complex request processing pipelines. When we build such pipeline, we ultimately end up with a single filter that takes a request and transforms it into a response or an error.

Consider:

```rust
let filter = (
        path!("api" / "endpoint_1").and_then(handle_endpoint_1)
    .or(path!("api" / "endpoint_2").and_then(handle_endpoint_2))
    .or(path!("api" / "endpoint_3").and_then(handle_endpoint_3))
    .or(path!("api" / "endpoint_4").and_then(handle_endpoint_4))
    ...
);
```

In this case our filter will have the following logical structure:


```
->- or ---- path!("api" / "endpoint_1") ---- handle_endpoint_1
    |
    or ---- path!("api" / "endpoint_2") ---- handle_endpoint_2
    |
    or----- path!("api" / "endpoint_3") ---- handle_endpoint_3
    |
    or----- path!("api" / "endpoint_4") ---- handle_endpoint_4
    |
    ...
```

Let's see what happens when we process a request to the url `"api/endpoint_3"`:

- the request will get to the first filter, `or`;
- `or` filter will pass it to its left sub-filter, `path!("api" / "endpoint_1")`;
- `path!("api" / "endpoint_1")` will match request's url against the pattern `"api/endpoint_1"`;
- since there is no match, `path!("api" / "endpoint_1")` will return `Err(404)` back to the `or` filter;
- `or` will see that the left sub-filter returned an error, and will call the right sub-filter, which happens to be another `or`;
- eventually, our request will reach the `path!("api" / "endpoint_3")` and match its pattern;
- `path!("api" / "endpoint_3")` will then call its inner filter, `handle_endpoint_3`;
- `handle_endpoint_3` will return `Or(Response)`, which will eventually bubble up to the top-most filter.

It is important to understand that the `or` filter is quite simple. Whenever the left sub-filter returns an error, `or` just calls the right sub-filter. Thus, if `handle_endpoint_3` in the above example were to return an error, the request would continue to `path!("api" / "endpoint_4")` and further down the filter tree.

To avoid this kind of behaviour, all our handlers return `Result<Response, Infallable>`, i.e. they never return errors. If a handler fails, we return `Ok(Response(500))`, not `Err(500)`.


## Part 2: how to report metrics, the correct way

Our current way of reporting the metrics is broken, the PR description above contains a detailed explanation of why it's happening. Thus, I will focus on two possible solutions.

**The first solution** is to add a filter that counts metrics for a specific path to each request handler. In code, it will look like this:

```rust
let filter = (
        path!("api" / "endpoint_1").and_then(handle_endpoint_1).and_then(report_metrics_for_endpoint_1)
    .or(path!("api" / "endpoint_2").and_then(handle_endpoint_2).and_then(report_metrics_for_endpoint_2))
    .or(path!("api" / "endpoint_3").and_then(handle_endpoint_3).and_then(report_metrics_for_endpoint_3))
    .or(path!("api" / "endpoint_4").and_then(handle_endpoint_4).and_then(report_metrics_for_endpoint_4))
    ...
);
```

Whenever `handle_endpoint_*` gets invoked, an appropriate `report_metrics_for_endpoint_*` also gets invoked.

We also need to report requests that don't match any endpoint. To do so, we add a filter at the very end of the `or` chain:

```rust
let filter = (
    ...
    .or(path!("api" / "endpoint_n").and_then(handle_endpoint_n).and_then(report_metrics_for_endpoint_n))
    .or(report_metrics_for_unknown_endpoint)
);
```

The `report_metrics_for_unknown_endpoint` filter will thus catch any unknown request and report it.

The advantage of this solution is that it's very easy to implement.

The disadvantage, however, is that the metrics are quite coarse. All requests that didn't match any endpoint will be counted towards a single metric, `unknown_endpoint`. We won't be able to distinguish between malformed urls, well-formed urls that weren't parsed due to bugs in our code, or completely bogus urls.

Arguably, this is not that big of an issue, though, because we can always look up failed requests in kibana.

**The second solution** is the one implemented in [PR 1475](https://github.com/gnosis/gp-v2-services/pull/1475). It is more complex, but also it is a general-purpose solution that allows for a more precise metrics.

The idea is to wrap the entire filter tree into a filter that reports all metrics:

```rust
let filter = (
    report_metrics(
        path!("api" / "endpoint_1").and_then(handle_endpoint_1)
        ...
    )
);
```

```
->- report_metrics
    |
    or ---- path!("api" / "endpoint_1") ---- handle_endpoint_1
    |
    ...
```

This filter doesn't depend on the tree structure, it is not affected by potential bugs in other filters.

However, precisely because of this independency, it also doesn't know which endpoint was matched. It only knows request's url and the returned status code. We therefore need to take the request url and remove all variable parameters from it (e.g. replace `/api/v1/orders/0x123...321` with `/api/v1/orders/*`), otherwise we'll end up with way too many metrics.

To do so, we build a list of all valid endpoint urls and pass it to the `report_metrics` function (see [the code][1]). The `report_metrics` then builds a prefix tree and uses it to sanitize urls.

The disadvantage of this method is obvious. It requires us to collect a list of all endpoint urls, and then implement a complex logic to sanitize a url.

The advantage is more precise metrics and independence from the filter tree structure.

## Conclusion

After writing this post, I've realised that the second solution is not as advantageous as I previously thought. Given that it requires writing and maintaining some complicated code, I've decided to close this PR and implement the first solution instead.

[1]: https://github.com/gnosis/gp-v2-services/pull/1475/files#diff-25ca3b362d29cb4f89851157ea343191211878a14628c4f92bcb0138c49ae726R115-R130

fleupold:
> The disadvantage, however, is that the metrics are quite coarse. All requests that didn't match any endpoint will be counted towards a single metric, unknown_endpoint

I do think that's fine as we are mainly interested in performance which should be the same for all unknown endpoints.

My understanding or `warp` is not as good as yours by now, but if we wanted to get a more fine-grained overview of our 404, we could maybe map each `Ok<Response>` into an `Ok<(label, Response)>` that we pass along the chain. In a last handler we could handle the generic 404 case returning `Ok<(label base on path, 404 Response)>`. At the very end of the chain (where we currently log requests) we could then [untuple_one](https://docs.rs/warp/latest/warp/trait.Filter.html#method.untuple_one) the label and report performance.

If the other approach simpler, this is fine imo too (the main goal is to get rid of the false 404 measurements)

PR #1476, opened 2021-12-09
CustomNodes strategy refactor

sunce86:
Task 2 from https://github.com/gnosis/gp-v2-services/issues/1407

Important changes:
1. CustomNodes strategy doesn't use crate `transaction_retry` nor `rpc::submit` file anymore. It relies completely on the new `Submitter` struct and its algorithm. (gas_price_stream.rs, retry.rs, rpc.rs are deleted)
2. Trait 'TransactionSubmitting' now returns custom error type (`SubmitApiError`) so we can better handle different situations in `Submitter`
3. `Submitter` doesn't get web3 as input but rather reuses web3 from contract object (not completely sure if this is safe in all use cases?)

Todos:
- [ ] ~~Implement cancellation for `CustomNodes` strategy~~ (skipping for now the cancellation for all strategies)
- [ ] ~~Add unit tests for `Submitter`~~ (no need for unit tests)
- [x] Cleanup of unused old code


fleupold (approved):
Looks good to me (I like that the overall LOC count was drastically reduced)!

I'm still not entirely sure how this abstraction allows us to mix and match between strategies. How would a strategy be configured that uses Eden and Flashbots in parallel?

sunce86:
> Looks good to me (I like that the overall LOC count was drastically reduced)!
> 
> I'm still not entirely sure how this abstraction allows us to mix and match between strategies. How would a strategy be configured that uses Eden and Flashbots in parallel?

We have two options, either to have `Vec<Submitter>` or `Vec<&'a dyn TransactionSubmitting>` inside `Submitter`. Will be considered in the next PR.

fleupold:
> We have two options, either to have Vec<Submitter> or Vec<&'a dyn TransactionSubmitting> inside Submitter

This will work for the parallel strategy, but how about a strategy where we only want to send certain txs to the public nodes (e.g. the ones not having any or only relatively fail safe interactions)?

sunce86:
I was thinking of having these structures dynamically filtered for each settlement and possibly for every iteration in run loop (for example if we want to remove eden in case the gas_price > X gwei)

PR #1480, opened 2021-12-13
Clippy updates, let's box it

josojo:
Fixing clippy failures. Needs to be merged, such that other PRs turn green again, as CI is failing currently

### Test Plan
non

### DIFF
--- a/crates/shared/src/price_estimation/competition.rs
+++ b/crates/shared/src/price_estimation/competition.rs
@@ -46,7 +46,12 @@ impl PriceEstimating for CompetitionPriceEstimator {
                         },
                     )
                     .map(|winning_estimate| {
-                        tracing::debug!(?query, ?winning_estimate, "winning price estimate",);
+                        tracing::debug!(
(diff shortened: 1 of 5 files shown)


crates/shared/src/price_estimation/competition.rs
```diff
@@ -55,18 +55,17 @@ impl PriceEstimating for CompetitionPriceEstimator {
 }
 
 #[derive(Debug)]
-struct EstimateData<'a> {
-    estimator_name: &'a str,
+struct EstimateData {
```
nlordell, line 58: Note that this change makes it so the `winning_estimate` log above no longer prints the estimator that was used.

nlordell (commented):
Just minor issue with logging the winning estimate.

nlordell (approved):
Weird that it suggests boxing here. `clippy` is always right though ;)

josojo:
>Weird that it suggests boxing here. clippy is always right though ;)

it's because the size for each enum  object is different, and it is a little bit more efficient, if we box it, so that we don't allocated the max object size.

PR #1482, opened 2021-12-14
Add Balancer SOR single order solver.

nlordell:
This PR adds a Balancer SOR single order solver to the mix!

The PR is unfortunately quite large - however more than half are tests. If this is too hard to review, let me know and I can split it up into smaller pieces.

There is an addition small change: single order solver now takes a `&Auction`. The reason for this is that the Balancer SOR API requires a `gas_price` which is uses for determining how many paths to "split" the route in. While we could have just passed a `GasPriceEstimating` to the `BalancerSorSolver`, I thought it was slightly cleaner to pass in a reference to the auction to all single order solvers. This allows implementation to (if needed) inspect things like the `external_prices` and `gas_price` to make some better decisions.

In a follow up PR, I will be adding required configuration parameters and enabling the solver!

### Test Plan

Added a bunch of new unit tests.
Also manual test to see what a solution looks like:
```
$ export BALANCER_SOR_URL="..."
$ export NODE_URL="..."
$ cargo test -p solver -- balancer_sor_solve --nocapture --ignored
...
test solver::balancer_sor_solver::tests::balancer_sor_solve ... ok
```


crates/shared/src/balancer_sor_api.rs
```diff
@@ -109,20 +131,22 @@ pub struct Swap {
     pub user_data: Bytes,
 }
 
-/// Module for deserializing the "wrapped" big integer types in the Balancer SOR
-/// API response (of the form `{ "type": "BigNumber", hex: "0x..." }`).
-mod serde_u256_wrapped {
+/// Module for deserializing the "wrapped" signed big integer types in the
+/// Balancer SOR API response (`{ "type": "BigNumber", hex: "0x..." }`).
+mod wrapped_big_integer {
```
bh2smith, line 136: Surely a great nice to have! I wonder if something like this could live more prominently in a web3 crate or `primitive_types` at some point.

bh2smith (approved):

crates/shared/src/balancer_sor_api.rs
nlordell: This is very specific to Balancer SOR. They return numbers like:
```json
{
  "foo": {
    "type": "BigNumber",
    "hex": "-0x1337"
  }
}
```

When the more expected way would be:
```json
{
  "foo": "-0x1337"
}
```

I wouldn't move this out for that reason.

sunce86 (approved):
lgtm

fleupold:
Very nice! This doesn't yet support the fast path yet right? Just wondering what is needed to support that as well?

PR #1484, opened 2021-12-15
Passes the parameter use_internal_buffers to http solvers.

marcovc:
I tried to keep the current design but I think the http solver parameter
passing should be refactored (in another PR). For example one could
have solver-specific SolverConfig structs with a virtual method that
simply inserts the config in the query.


nlordell (approved):
🎉 

I would prefer if it were enabled via a configuration parameter (like adding it to `shared::Arguments` for example) so that it is easier to turn it off without a code change (which is much more complicated in prod as well).

I'm OK with adding this right away to start collecting data, but I think we should parameterize this before releasing in prod.

marcovc:
> I would prefer if it were enabled via a configuration parameter (like adding it to shared::Arguments for example) 

Sure that makes sense. I can do it here.

marcovc:
@nlordell Added to Arguments (not shared Arguments) because the orderbook doesn't need it. I couldn't find what arguments the price estimator binaries use, but maybe that is why you were suggesting shared arguments instead? Anyway, happy to change it.

Also, I believe now this PR will also require a kubernetes change to pass the new parameter.

nlordell:
> Added to Arguments (not shared Arguments) because the orderbook doesn't need it. I couldn't find what arguments the price estimator binaries use, but maybe that is why you were suggesting shared arguments instead?

This is why I was suggesting it. The `orderbook` instantiates price estimators. The solver doesn't need price estimators since it doesn't need to quote orders for users nor need to determine min fees.

crates/solver/src/solver.rs
```diff
@@ -171,24 +172,29 @@ pub fn create(
...
-                token_info_fetcher.clone(),
-                buffer_retriever.clone(),
-                http_solver_cache.clone(),
-            )
-        };
+    let create_http_solver = |account: Account,
+                              url: Url,
+                              name: &'static str,
+                              config: SolverConfig,
+                              use_internal_buffers: bool|
```
nlordell, line 179: Since this gets passed to the `DefaultHttpSolverApi` anyway, can't this be included in the config `SolverConfig` above?

This might make it require way less plumbing.

marcovc: The SolverConfig is common to all solvers, so we would have two parameter there, one called `use_internal_buffers` and another `has_use_internal_buffers_parameter`. What about if we replace both by an Option<bool> ?

marcovc:
> The orderbook instantiates price estimators.

crates/shared/src/http_solver_api.rs
```diff
@@ -14,6 +14,7 @@ pub trait HttpSolverApi: Send + Sync {
         &self,
         model: &model::BatchAuctionModel,
         timeout: Duration,
+        use_internal_buffers: bool,
```
nlordell, line 17: We should remove this now that this value is passed in the `SolverConfig` right?

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -32,6 +32,7 @@ pub struct QuasimodoPriceEstimator {
     pub gas_info: Arc<dyn GasPriceEstimating>,
     pub native_token: H160,
     pub base_tokens: Arc<BaseTokens>,
+    pub use_internal_buffers: bool,
```
nlordell, line 35: This should also not be needed since the `SolverConfig` already has this value.

crates/solver/src/solver.rs
nlordell: Right now, we assume that our HTTP solvers have the same interface. We can make it an option so that:
```
None        => ?...
Some(false) => ?...&use_internal_buffers=false&...
Some(true)  => ?...&use_internal_buffers=true&...
```

marcovc: So the issue is indeed that http solvers do not have the same interface. Quasimodo is the only one that supports this parameter. The idea regarding the Option is to offer the 3 states we need: no parameter, parameter=false, parameter=true. Is this what you're saying?

bh2smith (approved):

crates/shared/src/http_solver_api.rs
```diff
@@ -96,10 +94,14 @@ impl HttpSolverApi for DefaultHttpSolverApi {
             url.query_pairs_mut()
                 .append_pair("ucp_policy", "EnforceForOrders");
         }
-        if self.config.has_use_internal_buffers_parameter {
+        if self.config.use_internal_buffers.is_some() {
```
nlordell, line 97: ubernit: This can be written slightly more "Rust"-y as:
```rust
if let Some(use_internal_buffers) = self.config.use_internal_buffers {
    url.query_pairs_mut().append_pair(
        "use_internal_buffers_parameter",
        &use_internal_buffers.to_string(),
    );
}
```

nlordell (approved):
Thanks for implementing the changes! Looks great, just a small Rust-related nit.

crates/shared/src/http_solver_api.rs
marcovc: Cool, thanks for the tip!

PR #1485, opened 2021-12-15
Refactor pool fetching to use N factories.

nlordell:
This PR refactors pool fetching to be on a `Vec<dyn InternalPoolFetching>`. This means that adding new pool factories _just_ requires you to add a new `InternalPoolFetching` implementation to that list. Additionally, a `Registry<Factory>: InternalPoolFetching` type was  introduced to wrap a `FactoryIndexing` implementation into a `InternalPoolFetching` making the process of adding new factories even easier.

This PR has +500 lines, but there is a good chunk of that is just moved code. For example the logic in `pool_cache` module was split into `pool_fetching::cache` and `pool_fetching::registry`. I tried to mark the sections of moved code without changes to make reviewing easier.

One additional note - the `BalancerPoolRegistry` type is actually no longer needed. The only place it gets used is in a unit test that verifies that registering pools via events matches pools in the subgraph. I want to remove this file, but will do it in a separate PR to not make this one even bigger.

There is an additional follow up "TODO" which is to get rid of the `FetchedBalancerPool`, `WeightedPool` and `StablePool` types in favour of `Vec<Pool>`. This would also be a larger change that touches a much larger part of the codebase, so it was left for later.

### Test Plan

There was _very_ little new logic in this PR, most of it was moving and combining existing logic in a different way. There was some new tests added for the new stuff, otherwise we moved and adapted existing tests.

You can also run the solver with _only_ Balancer pools as liquidity as a sanity check that we are indeed still fetching pools:
```
$ cargo run -p solver -- --baseline-sources BalancerV2 --solver-account 0xa6DDBD0dE6B310819b49f680F65871beE85f517e --transaction-strategy DryRun
...
2021-12-15T19:46:20.728Z DEBUG solver::liquidity_collector: got 35 AMMs
...
```


crates/shared/src/sources/balancer_v2/pool_fetching/cache.rs
```diff
@@ -0,0 +1,113 @@
...
+
+impl CacheKey<Pool> for H256 {
+    fn first_ord() -> Self {
+        H256::zero()
+    }
+
+    fn for_value(pool: &Pool) -> Self {
+        pool.id
+    }
+}
```
nlordell, line 90: Moved from `pool_cache`.

crates/shared/src/sources/balancer_v2/pool_fetching/cache.rs
```diff
@@ -0,0 +1,113 @@
...
+    async fn fetch_values(&self, pool_ids: HashSet<H256>, at_block: Block) -> Result<Vec<Pool>> {
+        self.0.pools_by_id(pool_ids, at_block).await
+    }
+}
+
+impl CacheMetrics for Arc<dyn BalancerPoolCacheMetrics> {
+    fn entries_fetched(&self, cache_hits: usize, cache_misses: usize) {
+        self.pools_fetched(cache_hits, cache_misses)
+    }
+}
```
nlordell, line 113: Moved from `pool_cache`.

crates/shared/src/sources/balancer_v2/pool_fetching/registry.rs
```diff
@@ -0,0 +1,174 @@
...
+        .collect()
+}
+
+fn is_contract_error(err: &anyhow::Error) -> bool {
+    matches!(
+        err.downcast_ref::<MethodError>()
+            .map(EthcontractErrorType::classify),
+        Some(EthcontractErrorType::Contract),
+    )
+}
```
nlordell, line 140: Moved from `pool_cache`. Tests also moved to this file.

crates/shared/src/sources/balancer_v2/pool_fetching/registry.rs
```diff
@@ -0,0 +1,174 @@
...
+        let pool_infos = self.updater.lock().await.store().pools_by_id(&pool_ids);
+        let pool_futures = pool_infos
+            .into_iter()
+            .map(|pool_info| self.fetcher.fetch_pool(&pool_info, &mut batch, block))
+            .collect::<Vec<_>>();
+
+        batch.execute_all(MAX_BATCH_SIZE).await;
+
+        let pools = future::join_all(pool_futures).await;
+        collect_pool_results(pools)
```
nlordell, line 101: Also moved from `pool_cache`. This is just a generic version of the `CacheFetching<H256, WeightedPool>` and `CacheFetching<H256, StablePool>` implementations.

So less code duplication 🎉.

crates/shared/src/price_estimation/competition.rs
```diff
@@ -46,7 +46,11 @@ impl PriceEstimating for CompetitionPriceEstimator {
                         },
                     )
                     .map(|winning_estimate| {
-                        tracing::debug!(?query, ?winning_estimate, "winning price estimate",);
+                        // The `EstimateData::estimator_name` field is getting
+                        // flagged as dead code despited being used in the log
+                        // below. Just convince the linter that we use it.
+                        let _ = winning_estimate.estimator_name;
+                        tracing::debug!(?query, ?winning_estimate, "winning price estimate");
```
nlordell, line 53: This is fixed on `main` - I just included a fix here to make CI past, but this will go away when rebasing + merging.

crates/shared/src/sources/balancer_v2/swap/error.rs
```diff
@@ -65,6 +65,6 @@ mod tests {
     #[test]
     #[should_panic]
     fn error_parse_fail() {
-        Error::from("Poop");
+        let _ = Error::from("Poop");
```
nlordell, line 68: Another lint error that was already fixed on `main` and will go away once rebased.

crates/solver/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::large_enum_variant)]
+
```
nlordell, line 2: More lint fixes.

crates/shared/src/event_handling.rs
```diff
@@ -79,6 +79,10 @@ where
         }
     }
 
+    pub fn store(&self) -> &S {
+        &self.store
+    }
```
bh2smith, line 84: This is so much nicer than pub crate, but syntactically less concise. Is this a personal preference or are there some other differences coming along with this choice?

If they are the same and the plan is to delete this file in a follow up PR, then I might argue that this change results in more touched lines on an already large PR. 

either way, it's fine.

crates/shared/src/sources/balancer_v2/swap/error.rs
bh2smith: Ha ha. Poop.

bh2smith (approved):
Easy to read, probably took a while to make.

crates/shared/src/event_handling.rs
nlordell: >  Is this a personal preference or are there some other differences coming along with this choice?

There is a subtle difference. Specifically with `pub(crate)` you could do something bad like:
```rust
let mut event_handler = ...;
event_handler.store.change_state_in_unexpected_way();
```

However, if we keep it private and add a public accessor, then you can only get an **immutable** reference to the store, meaning:
```rust
let mut event_handler = ...;
event_handler.store().change_state_in_unexpected_way(); // ERROR: not mutable
```

crates/shared/src/sources/balancer_v2/pool_fetching/registry.rs
nlordell: ESL 😛

crates/shared/src/sources/balancer_v2/pool_fetching.rs
```diff
@@ -199,74 +194,63 @@ impl BalancerPoolFetching for BalancerPoolFetcher {
...
 }
 
+/// Extract the pool address from an ID.
+///
+/// This takes advantage that the first 20 bytes of the ID is the address of
+/// the pool. For example the GNO-BAL pool with ID
+/// `0x36128d5436d2d70cab39c9af9cce146c38554ff0000200000000000000000009`:
+/// <https://etherscan.io/address/0x36128D5436d2d70cab39C9AF9CcE146C38554ff0>
+fn pool_address_from_id(pool_id: H256) -> H160 {
+    let mut address = H160::default();
```
sunce86, line 237: nit: Maybe a bit more succinct `H160::from_slice(&pool_id[..20])` would do the job?

sunce86: Would this function be useful maybe for other liquidity sources?

crates/shared/src/sources/balancer_v2/pool_fetching.rs
```diff
@@ -199,74 +194,63 @@ impl BalancerPoolFetching for BalancerPoolFetcher {
...
-            .fetch(pool_ids, at_block)
-            .await?;
-        // Return only those pools which are not paused.
-        Ok(FetchedBalancerPools {
-            stable_pools: filter_paused(fetched_stable_pools),
-            weighted_pools: filter_paused(fetched_weighted_pools),
-        })
-    }
-}
+        let pool_ids = self.fetcher.pool_ids_for_token_pairs(token_pairs).await;
```
sunce86, line 197: Do we need to have `pool_ids_for_token_pairs` and `pools_by_id` functions separated? I don't see any case when we call `pool_ids_for_token_pairs` without calling `pools_by_id` right after.

crates/shared/src/sources/balancer_v2/pool_fetching/cache.rs
```diff
@@ -0,0 +1,113 @@
...
+{
+    async fn pool_ids_for_token_pairs(
+        &self,
+        token_pairs: HashSet<model::TokenPair>,
+    ) -> HashSet<H256> {
+        self.inner.pool_ids_for_token_pairs(token_pairs).await
+    }
+
+    async fn pools_by_id(&self, pool_ids: HashSet<H256>, block: Block) -> Result<Vec<Pool>> {
+        self.cache.fetch(pool_ids, block).await
```
sunce86, line 68: Oh, is the caching the reason to separate `pool_ids_for_token_pairs` and `pools_by_id`? :)

sunce86 (approved):
Lgtm

crates/shared/src/sources/balancer_v2/pool_fetching.rs
nlordell: This is a Balancer specific detail.

crates/shared/src/sources/balancer_v2/pool_fetching.rs
nlordell: Its because of caching details. Specifically, the pool cache operates on `Map<H256, Pool>` (since a single pool can be over multiple token pairs, and multiple pools can trade the same token pair).

Additionally, we use an LRU cache strategy where we eagerly update commonly used pools every time we observe a new block. This means that we need to have a `pools_by_id` to be able to only update those pools.

PR #1486, opened 2021-12-15
Remove `event_handler` module

nlordell:
As a follow up to #1485, this PR removes the `event_handler` module and the `BalancerPoolRegistry` type as it is no longer needed (it was superseded by the `Aggregate: InternalPoolFetching` type from the aforementioned PR). We still index events, it was just mostly moved to `pool_storage` and `pool_fetching::registry` modules so that is generic on the factory type.

The logic in `pool_fetching` is unchanged (despite the PR changes). It would just moved into smaller helper methods to make the test setup and assertions a little easier.

### Test Plan

There was a very important unit test in the `event_handler` module that would index all pools by events and compare them to what the Balancer subgraph would return. This unit test has been re-written and now will actually fetch pool state for all those pools (on top of the requirement of indexing the events).

```
$ cargo test -p shared -- balancer_fetched_pools_match_subgraph --ignored --nocapture
...
2021-12-15T21:47:20.368Z  WARN shared::sources::balancer_v2::pool_fetching::tests: unknown_pools=[]
test sources::balancer_v2::pool_fetching::tests::balancer_fetched_pools_match_subgraph ... ok
...
```


crates/shared/src/sources/balancer_v2/pool_fetching.rs
```diff
@@ -227,6 +208,44 @@ impl Maintaining for BalancerPoolFetcher {
     }
 }
 
+/// Creates an aggregate fetcher for all supported pool factories.
+async fn create_all_pools_fetcher(
```
bh2smith, line 212: Nit. Might be better to call create all pool fetchers (I.e move the plural from pools to fetcher) since that's what it does.

crates/shared/src/sources/balancer_v2/pool_storage.rs
```diff
@@ -197,6 +202,44 @@ where
...
+        events: Vec<Event<BasePoolFactoryEvent>>,
+        range: RangeInclusive<BlockNumber>,
+    ) -> Result<()> {
+        tracing::debug!("replacing {} events for block {:?}", events.len(), range);
+
+        self.remove_pools_newer_than_block(range.start().to_u64());
+        self.append_events(events).await
+    }
+
+    async fn append_events(&mut self, events: Vec<Event<BasePoolFactoryEvent>>) -> Result<()> {
```
bh2smith, line 221: If were no longer have an event handle what are all these events handling method still used for?

bh2smith (approved):
Looks great. Just wondering why we keep all these event handling methods if we don't handle events anymore.

crates/shared/src/sources/balancer_v2/pool_storage.rs
nlordell: We do. Its in the `pool_storage` and `pool_fetching::registry` modules where we now have a registry per factory instead of a BalancerPoolRegistry for all three (which required manually adding a new field and code for each new factory you wanted to index).

PR #1493, opened 2021-12-17
implements timeout for price estimators 

josojo:
implements timeout for price estimators to quicker send results back to front-end

Currently, we have a lot of warnings that some price estimations take a long time on post quote. My suspicion is that the price estimator takes quite long for these instances. Hence, we eliminate all price estimations that take longer than 3 secs in order to provide a good result after 3 seconds from other estimators. Always waiting for the slowest makes no sense to me.

I am very unsure whether this is the right approach. Maybe just measuring the time for price-estimator and then working on the price estimators itself makes more sense. That is why this PR is in draft.
### Test Plan
Non

### DIFF
--- a/crates/shared/src/price_estimation/competition.rs
+++ b/crates/shared/src/price_estimation/competition.rs
@@ -3,7 +3,9 @@ use anyhow::{anyhow, Result};
 use futures::future;
 use num::BigRational;
 use std::cmp;
-
+use std::time::Duration;
+use tokio::time::error::Elapsed;
+use tokio::time::timeout;
 /// Price estimator that pulls estimates from various sources
 /// and competes on the best price.
 pub struct CompetitionPriceEstimator {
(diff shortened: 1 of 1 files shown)


nlordell:
One fear I have with timeouts is that we might have some price estimators that almost never get used as a result. Then it would be possible for a situation where we have a slow price estimator that produces good results and costs a lot of gas to not get considered here (resulting in a lower min fee) but then getting used for solutions (since the timeout when solving is much higher).

I do like the change in general, but think I think we would need to instrument the price estimators so that we count timeouts as a separate metric and does not get confounded with failures (so we can see if turning this off effectively disables a price estimator or if the timeouts happen only occasionally).

fleupold:
The frontend is allowing for a 10s timeout for the whole post quote. I'm not a fan of this change as it would make our price estimation much less complete (pretty likely some queries will timeout every once in a while and so we would be giving flaky quotes). 

Instead I'd suggest to push https://github.com/gnosis/gp-v2-services/pull/1469 which reduces the amount of successive roundtrips we are doing and maybe think of smarter ways of caching aspects of the relevant queries (e.g. a USD<>ETH price) in order to get gas estimation from the original price estimation instead of having to do 3 queries.

I also like the idea of measuring which price estimator is taking so long!

josojo:
thanks for the feedback. I think Felix approach is better, yes

PR #1494, opened 2021-12-17
Consider settlement with recent orders if they are part of "mature" settlements

### LINKED ISSUE #1310: Consider settlement with recent orders if they are part of "mature" settlements.
We currently only consider solutions if it has at least one order older than 30s. This causes problems, however, for orders recent orders (i.e. less than 30s old) where they may not get adequate protection by being under-represented in the objective value comparison.

To illustrate this with an example:
- `O1`: 1 minute old
- `O2`: 10 second old
- `O3`: 5 seconds old

And we have solutions:
- `S1`: trades orders `O1` and `O3`
- `S2`: trades orders `O2`
- `S3`: trades orders `O3`

Then both `S1` and `S3` need to be considered, otherwise we might not adequately protect `O3`. This can be particularly problematic for if `O3` is a "juicy" order that generates a lot of surplus (i.e. has a lot of slippage) and can get included at sub-optimal prices by a solver by using a "mature" order to get around the 30s rule.

MartinquaXD:
Fixes #1310 

See linked issue to read the motivation.  

Implementation:
We get a `Vec<Settlement>` and iterate over it. If a settlement includes a mature order (older than passed time) or a recent order which has been considered valid at one point, all orders within the current settlement (including recent orders) are considered valid and we look at the next settlement.
We keep looping over all settlements until we reach a point where no new orders have been added to the valid set. Now we know all of the valid orders.
Because this can get computationally expensive for many settlements with many orders, validated settlements get stored in a `HashSet` in order to quickly determine on each iteration whether we can skip looking at the current settlement.

### Test Plan
Added unit test for 3 scenarios:
1) No settlement includes any "mature" orders => No settlements should be considered
2) 2 disjoint settlements (one including mature order, one including no mature orders) => only the settlement with the mature order is considered
3) Multiple settlements (1 including mature order) arranged like a linked list => Because 1 settlement includes mature order the other settlements "linked" to it are also considered valid

### DIFF
--- a/crates/solver/src/driver/solver_settlements.rs
+++ b/crates/solver/src/driver/solver_settlements.rs
@@ -3,7 +3,10 @@ use ethcontract::U256;
 use num::BigRational;
 use primitive_types::H160;
 use shared::conversions::U256Ext;
-use std::{collections::HashMap, time::Duration};
+use std::{
+    collections::{HashMap, HashSet},
+    time::Duration,
+};
 
 pub fn has_user_order(settlement: &Settlement) -> bool {
     settlement
@@ -90,18 +93,64 @@ fn merge_at_most_settlements(
     }
 }
 
-pub fn filter_settlements_without_old_orders(
+/// Filters out all settlements without any orders which are mature by age or mature by association.
+/// Any order older than `min_order_age` is considered to be mature by age.
+/// Any younger order in a settlement containing an order mature by age or mature by association
+/// is considered to be mature by association.
+/// Because maturity by association is defined recursively it can "spread" across settlements,
+/// resulting in settlements being allowed where it's not immediately obvious by which association
+/// any order of a settlement has matured.
+pub fn retain_mature_settlements(
     min_order_age: Duration,
-    settlements: &mut Vec<Settlement>,
-) {
-    let settle_orders_older_than =
-        chrono::offset::Utc::now() - chrono::Duration::from_std(min_order_age).unwrap();
-    settlements.retain(|settlement| {
-        settlement
-            .trades()
-            .iter()
-            .any(|trade| trade.order.order_meta_data.creation_date <= settle_orders_older_than)
-    });
+    settlements: Vec<Settlement>,
+) -> Vec<Settlement> {
+    fn find_mature_settlements(
+        min_order_age: Duration,
+        settlements: &[Settlement],
+    ) -> HashSet<usize> {
+        let settle_orders_older_than =
+            chrono::offset::Utc::now() - chrono::Duration::from_std(min_order_age).unwrap();
+
+        let mut valid_trades = HashSet::<&model::order::OrderUid>::default();
+        let mut valid_settlement_indices = HashSet::<usize>::default();
+
+        loop {
+            let mut new_order_added = false;
+
+            for (index, settlement) in settlements.iter().enumerate() {
+                if valid_settlement_indices.contains(&index) {
+                    break;
+                }
+
+                let contains_valid_trade = settlement.trades().iter().any(|trade| {
+                    // mature by age
+                    trade.order.order_meta_data.creation_date <= settle_orders_older_than
+                    // mature by association
+                    || valid_trades.contains(&trade.order.order_meta_data.uid)
+                });
+
+                if contains_valid_trade {
+                    for trade in settlement.trades() {
+                        // make all orders within this settlement mature by association
+                        new_order_added |= valid_trades.insert(&trade.order.order_meta_data.uid);
+                    }
+                    valid_settlement_indices.insert(index);
+                }
+            }
+
+            if !new_order_added {
+                break valid_settlement_indices;
+            }
+        }
+    }
+
+    let valid_settlement_indices = find_mature_settlements(min_order_age, &settlements[..]);
+    settlements
+        .into_iter()
+        .enumerate()
+        .filter(|(index, _)| valid_settlement_indices.contains(index))
+        .map(|(_, item)| item)
+        .collect()
 }
 
 #[cfg(test)]
--- a/crates/solver/src/driver/solver_settlements.rs
+++ b/crates/solver/src/driver/solver_settlements.rs
@@ -157,12 +157,84 @@ pub fn retain_mature_settlements(
 mod tests {
     use super::*;
     use crate::settlement::Trade;
+    use chrono::{offset::Utc, DateTime, Duration, Local};
     use maplit::hashmap;
     use model::order::{Order, OrderCreation, OrderKind, OrderMetaData, OrderUid};
     use num::rational::BigRational;
     use num::traits::FromPrimitive;
     use primitive_types::U256;
     use std::collections::HashSet;
+    use std::ops::Sub;
+
+    fn trade(created_at: DateTime<Utc>, uid: u8) -> Trade {
+        Trade {
+            order: Order {
+                order_meta_data: OrderMetaData {
+                    creation_date: created_at,
+                    uid: OrderUid([uid; 56]),
+                    ..Default::default()
+                },
+                ..Default::default()
+            },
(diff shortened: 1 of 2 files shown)


MartinquaXD:
On second thought the operations we can skip because we have the settlements HashSet are very cheap. So having this set is maybe not worth the additional code complexity. Feedback is welcome.

crates/solver/src/driver/solver_settlements.rs
```diff
@@ -90,30 +93,138 @@ fn merge_at_most_settlements(
...
-    min_order_age: Duration,
-    settlements: &mut Vec<Settlement>,
-) {
+/// Filters out all settlements without any orders which are mature by age or mature by association.
+/// Any order older than `min_order_age` is considered to be mature by age.
+/// Any younger order in a settlement containing an order mature by age or mature by association
+/// is considered to be mature by association.
+/// Because maturity by association is defined recursively it can "spread" across settlements,
+/// resulting in settlements being allowed where it's not immediately obvious by which association
+/// any order of a settlement has matured.
```
bh2smith, line 102: Very nice description here.

crates/solver/src/driver/solver_settlements.rs
```diff
@@ -90,30 +93,138 @@ fn merge_at_most_settlements(
...
-        settlement
-            .trades()
-            .iter()
-            .any(|trade| trade.order.order_meta_data.creation_date <= settle_orders_older_than)
-    });
+
+    let mut valid_trades = HashSet::<&Order>::default();
+    // Use *const T to avoid expensive clones and hashing. This also doesn't require T: Hash + Eq.
+    // This approach is correct as long as 'settlements' doesn't reallocate.
+    let mut valid_settlements = HashSet::<*const Settlement>::default();
```
bh2smith, line 111: Cute trick!

crates/solver/src/driver/solver_settlements.rs
```diff
@@ -90,30 +93,138 @@ fn merge_at_most_settlements(
...
+
+        for settlement in settlements.iter() {
+            if valid_settlements.contains(&(settlement as _)) {
+                break;
+            }
+
+            let contains_valid_trade = settlement.trades().iter().any(|trade| {
+                //mature by age
+                trade.order.order_meta_data.creation_date <= settle_orders_older_than
+                    //mature by association
```
bh2smith, line 124: nit:
```suggestion
                // mature by age
                trade.order.order_meta_data.creation_date <= settle_orders_older_than
                    // mature by association
```

bh2smith (approved):
Looks very good! Nice first PR to the project.

crates/solver/src/driver/solver_settlements.rs
nlordell, line 111: I think one more condition on correctness here is that we don't mutate the `settlements: &mut Vec<Settlement>`. For example, something as innocent as `settlements.push(...)` could cause the vector to require growing and reallocating moving all the inner settlements to new memory locations and invalidating this `HashSet`.

How it is right now, its currently fine (since we only modify the `settlements: &mut Vec<Settlement>` after we are done modifying the `HashSet`) but maybe we should be more explicit and maybe capture `settlements` reference preventing mutation for as long as `HashSet` is mutable to the compiler helps you avoid this foot-gun?

nlordell (commented):
Nice clean code and tests. Looks good to me, just have a concern around the use of `HashSet<*const _>`.

crates/solver/src/driver/solver_settlements.rs
MartinquaXD: I pushed another commit to deal with that issue. I added an inner function which takes `&[Settlement]` and returns `HashSet<*const Settlement`. Invalidating the pointers within this function should now be caught by the compiler.
A programmer could still mutate the `&mut settlements` right after finding all settlements but I added a comment there to not do that.

crates/solver/src/driver/solver_settlements.rs
```diff
@@ -90,30 +93,144 @@ fn merge_at_most_settlements(
...
+            if !new_order_added {
+                break valid_settlements;
+            }
+        }
+    }
+
+    let valid_settlements = find_mature_settlements(min_order_age, &settlements[..]);
+    // After finding mature settlements DON'T mutate `settlements` other than filtering them with
+    // `valid_settlements`. Everything else might invalidate contents of `valid_settlements`!
+    settlements.retain(|settlement| valid_settlements.contains(&(settlement as _)))
```
nlordell, line 148: Hmm, I just thought about this now, but this is dependent on the specific implementation of `retain` which is not allow to move the inner `settlements` to new locations in memory temporarily. `Vec::retain` doesn't guarantee that the elements are pinned.

Although its less efficient (requires allocation, etc.) maybe it makes more sense to do this filtering on indices and construct a new `Vec<Settlement>` for the filtered result?

MartinquaXD: Yeah I checked that this isn't a problem with the current implementation of `retain` but I didn't think about that it might change in the future.
I now went with your idea of filtering based on indices and collecting into a new `Vec<Settlement>`.

crates/solver/src/driver/solver_settlements.rs
```diff
@@ -90,30 +93,148 @@ fn merge_at_most_settlements(
...
+    settlements: Vec<Settlement>,
+) -> Vec<Settlement> {
+    fn find_mature_settlements(
+        min_order_age: Duration,
+        settlements: &[Settlement],
+    ) -> HashSet<usize> {
+        let settle_orders_older_than =
+            chrono::offset::Utc::now() - chrono::Duration::from_std(min_order_age).unwrap();
+
+        let mut valid_trades = HashSet::<&model::order::Order>::default();
```
nlordell, line 114: Ubernit: I think it would be slightly more natural to use `HashSet::<OrderUid>` here since they uniquely identify orders (also, under the hood its just a `[u8; 56]` so it would be slightly more efficient - not that it really matters in the grand scheme of things).

nlordell (approved):
LGTM! One tiny ubernit - but feel free to ignore it.

crates/solver/src/driver/solver_settlements.rs
MartinquaXD: Done

crates/solver/src/driver/solver_settlements.rs
```diff
@@ -90,30 +93,148 @@ fn merge_at_most_settlements(
...
+        min_order_age: Duration,
+        settlements: &[Settlement],
+    ) -> HashSet<usize> {
+        let settle_orders_older_than =
+            chrono::offset::Utc::now() - chrono::Duration::from_std(min_order_age).unwrap();
+
+        let mut valid_trades = HashSet::<&model::order::OrderUid>::default();
+        let mut valid_settlement_indices = HashSet::<usize>::default();
+
+        loop {
```
fleupold, line 117: nanonit: might have been nice to upper bound this loop (intuitively I'd say by `settlements.length`) to make it more obvious that this can not loop endlessly (I'm always scared when I read unbounded loops).

fleupold (commented):
Just came here to say that this is a really nice PR (great description, test, comment, clean solution)

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FShjan97TyUJY4%2Fgiphy.gif&token=really_nice" data-gitmeme-token="really_nice"></a>

crates/solver/src/driver/solver_settlements.rs
MartinquaXD: That's a good point. :+1:
It didn't occur to me that giving an upper bound (while not changing the correctness of the solution) would help convey more information about the runtime and give peace of mind.
I will keep an eye out for `loop{}` in the future and see whether I could add more information that way.

PR #1499, opened 2021-12-21
Validate Price Estimation Queries Early and at a Single Point

### LINKED ISSUE #1495: BUG: Driver passing native ETH token with null decimals to Quasimodo solver
Reported by @marcovc who can provide instance details, but currently in production this is happening.

We need to find the PR that introduced passing raw orders (i.e. without mapping ETH to WETH) into the solvers.

MartinquaXD:
Fixes #1495

Some price estimators had problems estimating prices for ETH buy orders.
Because the price estimators don't really deal with ETH but with WETH we now detect ETH buy orders and pretend they are WETH buy orders.
To not have to implement that behavior for every price estimator a new type `SanitizedPriceEstimator` was added which is a wrapper for types implementing the `PriceEstimating` trait. I wrapped the `CompetitionPriceEstimator` which bundles all our estimators in a `SanitizedPriceEstimator` which means all price price estimator can assume that they will only ever get valid price estimation queries.
This lead to some duplicated code and tests being removed as well as the `bad_token_detector` being dropped from the individual estimators. The estimators only needed that member to find unsupported tokens which they no longer have to worry about.

Having this single point where all price estimations go through also enables us to easily implement a nice optimization Ben had in mind. If the buy_token was ETH we did not include the cost of unwrapping WETH->ETH in the price estimation. This is something the `SanitizedPriceEstimator` does as well.

### Test Plan
Unit tests for `SanitizedPriceEstimator`
1) normal order
2) unsupported buy_token
3) unsupported sell_token
4) ETH as buy_token where adding gas cost for unwrapping ETH would NOT overflow U256 for gas price
5) ETH as buy_token where adding gas cost for unwrapping ETH would overflow U256 for gas price
6) buy_token and sell_token are identical

Did manual integration tests:
1) Buying ETH
2) Buying WETH
3) Buying and selling same token
<details>
<summary>logs of orderbook</summary>
<pre>
2021-12-23T11:13:40.510Z DEBUG orderbook::api::post_quote: Received quote request OrderQuoteRequest { from: 0x1a015ad600829d8b288bf3c50b9f557a2c12e9ca, sell_token: 0x6b175474e89094c44da98b954eedeac495271d0f, buy_token: 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee, receiver: Some(0x0000000000000000000000000000000000000000), side: Buy { buy_amount_after_fee: 1000000000000000000 }, valid_to: 2147483648, app_data: 0x0000000000000000000000000000000000000000000000000000000000000000, partially_fillable: false, sell_token_balance: Erc20, buy_token_balance: Erc20 }
2021-12-23T11:13:40.630Z DEBUG orderbook::fee: computing subsidized fee fee_data=FeeData { sell_token: 0x6b175474e89094c44da98b954eedeac495271d0f, buy_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, amount: 1000000000000000000, kind: Buy } app_data=0x0000000000000000000000000000000000000000000000000000000000000000<br/>
<b>2021-12-23T11:13:40.631Z DEBUG shared::price_estimation::sanitized: estimate price for native token instead of ETH query=Query { sell_token: 0x6b175474e89094c44da98b954eedeac495271d0f, buy_token: 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee, in_amount: 1000000000000000000, kind: Buy } sanitized_query=Query { sell_token: 0x6b175474e89094c44da98b954eedeac495271d0f, buy_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, in_amount: 1000000000000000000, kind: Buy }</b>
<br/>2021-12-23T11:13:40.649Z DEBUG orderbook::fee: using existing fee measurement FeeParameters { gas_amount: 255402.0, gas_price: 56185118873.0, sell_token_price: 3947.210642883208 }
2021-12-23T11:13:40.650Z DEBUG orderbook::fee: computed subsidized fee of (56641650641400004608, 0x6b175474e89094c44da98b954eedeac495271d0f)
2021-12-23T11:13:40.752Z DEBUG shared::http_solver_api: http solver instance name is 2021-12-23_11:13:40.751862_UTC_Ethereum___Mainnet_1
2021-12-23T11:13:40.809Z DEBUG shared::price_estimation::competition: received price estimate estimator_name=Quasimodo query=Query { sell_token: 0x6b175474e89094c44da98b954eedeac495271d0f, buy_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, in_amount: 1000000000000000000, kind: Buy } estimate=Estimate { out_amount: 3947210642883208051882, gas: 255402 }
2021-12-23T11:13:40.810Z DEBUG shared::price_estimation::competition: winning price estimate query=Query { sell_token: 0x6b175474e89094c44da98b954eedeac495271d0f, buy_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, in_amount: 1000000000000000000, kind: Buy } winning_estimate=EstimateData { estimator_name: "Quasimodo", estimate: Estimate { out_amount: 3947210642883208051882, gas: 255402 }, sell_over_buy: Ratio { numer: 1973605321441604025941, denom: 500000000000000000 } }
<br/><b>2021-12-23T11:13:40.811Z DEBUG shared::price_estimation::sanitized: added cost of unwrapping WETH to price estimation query=Query { sell_token: 0x6b175474e89094c44da98b954eedeac495271d0f, buy_token: 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee, in_amount: 1000000000000000000, kind: Buy } final_estimation=Estimate { out_amount: 3947210642883208051882, gas: 269594 }</b><br/>
2021-12-23T11:13:40.859Z DEBUG orderbook::api::post_quote: Received quote request OrderQuoteRequest { from: 0x1a015ad600829d8b288bf3c50b9f557a2c12e9ca, sell_token: 0x6b175474e89094c44da98b954eedeac495271d0f, buy_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, receiver: Some(0x0000000000000000000000000000000000000000), side: Buy { buy_amount_after_fee: 1000000000000000000 }, valid_to: 2147483648, app_data: 0x0000000000000000000000000000000000000000000000000000000000000000, partially_fillable: false, sell_token_balance: Erc20, buy_token_balance: Erc20 }
2021-12-23T11:13:40.859Z DEBUG orderbook::fee: computing subsidized fee fee_data=FeeData { sell_token: 0x6b175474e89094c44da98b954eedeac495271d0f, buy_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, amount: 1000000000000000000, kind: Buy } app_data=0x0000000000000000000000000000000000000000000000000000000000000000
2021-12-23T11:13:40.867Z DEBUG orderbook::fee: using existing fee measurement FeeParameters { gas_amount: 255402.0, gas_price: 56185118873.0, sell_token_price: 3947.210642883208 }
2021-12-23T11:13:40.867Z DEBUG orderbook::fee: computed subsidized fee of (56641650641400004608, 0x6b175474e89094c44da98b954eedeac495271d0f)
2021-12-23T11:13:40.978Z DEBUG shared::http_solver_api: http solver instance name is 2021-12-23_11:13:40.978531_UTC_Ethereum___Mainnet_1
2021-12-23T11:13:41.032Z DEBUG shared::price_estimation::competition: received price estimate estimator_name=Quasimodo query=Query { sell_token: 0x6b175474e89094c44da98b954eedeac495271d0f, buy_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, in_amount: 1000000000000000000, kind: Buy } estimate=Estimate { out_amount: 3947210642883208051882, gas: 255402 }
2021-12-23T11:13:41.033Z DEBUG shared::price_estimation::competition: winning price estimate query=Query { sell_token: 0x6b175474e89094c44da98b954eedeac495271d0f, buy_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, in_amount: 1000000000000000000, kind: Buy } winning_estimate=EstimateData { estimator_name: "Quasimodo", estimate: Estimate { out_amount: 3947210642883208051882, gas: 255402 }, sell_over_buy: Ratio { numer: 1973605321441604025941, denom: 500000000000000000 } }
2021-12-23T11:13:41.081Z DEBUG orderbook::api::post_quote: Received quote request OrderQuoteRequest { from: 0x1a015ad600829d8b288bf3c50b9f557a2c12e9ca, sell_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, buy_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, receiver: Some(0x0000000000000000000000000000000000000000), side: Buy { buy_amount_after_fee: 1000000000000000000 }, valid_to: 2147483648, app_data: 0x0000000000000000000000000000000000000000000000000000000000000000, partially_fillable: false, sell_token_balance: Erc20, buy_token_balance: Erc20 }
2021-12-23T11:13:41.082Z  WARN orderbook::api::post_quote: post_quote error err=Order(Partial(SameBuyAndSellToken)) request=OrderQuoteRequest { from: 0x1a015ad600829d8b288bf3c50b9f557a2c12e9ca, sell_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, buy_token: 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, receiver: Some(0x0000000000000000000000000000000000000000), side: Buy { buy_amount_after_fee: 1000000000000000000 }, valid_to: 2147483648, app_data: 0x0000000000000000000000000000000000000000000000000000000000000000, partially_fillable: false, sell_token_balance: Erc20, buy_token_balance: Erc20 }
</pre>

### DIFF
--- a/crates/shared/src/price_estimation/sanitized.rs
+++ b/crates/shared/src/price_estimation/sanitized.rs
@@ -1,10 +1,10 @@
 use super::{Estimate, PriceEstimating, PriceEstimationError, Query};
-use crate::bad_token::BadTokenDetecting;
+use crate::bad_token::{BadTokenDetecting, TokenQuality};
 use crate::price_estimation::gas::GAS_PER_WETH_UNWRAP;
 use anyhow::Result;
 use model::order::BUY_ETH_ADDRESS;
 use primitive_types::H160;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
 /// Verifies that buy and sell tokens are supported and handles
@@ -36,9 +36,12 @@ impl<T: PriceEstimating> SanitizedPriceEstimator<T> {
         }
     }
 
-    async fn get_unsupported_tokens(&self, queries: &[Query]) -> Result<HashSet<H160>> {
-        let mut unsupported_tokens: HashSet<H160> = Default::default();
-        let mut supported_tokens: HashSet<H160> = Default::default();
+    async fn get_token_quality_errors(
+        &self,
+        queries: &[Query],
+    ) -> HashMap<H160, PriceEstimationError> {
+        let mut token_quality_errors: HashMap<H160, PriceEstimationError> = Default::default();
+        let mut checked_tokens = HashSet::<H160>::default();
 
         // TODO should this be parallelised?
         for token in queries
@@ -46,40 +49,43 @@ impl<T: PriceEstimating> SanitizedPriceEstimator<T> {
             .copied()
             .flat_map(|query| [query.buy_token, query.sell_token])
         {
-            if unsupported_tokens.contains(&token) || supported_tokens.contains(&token) {
+            if checked_tokens.contains(&token) {
                 continue;
             }
-            let quality = self.bad_token_detector.detect(token).await?;
-            if quality.is_good() {
-                supported_tokens.insert(token);
-            } else {
-                unsupported_tokens.insert(token);
-            }
+
+            match self.bad_token_detector.detect(token).await {
+                Err(err) => {
+                    token_quality_errors.insert(token, PriceEstimationError::Other(err));
+                }
+                Ok(TokenQuality::Bad { .. }) => {
+                    token_quality_errors
+                        .insert(token, PriceEstimationError::UnsupportedToken(token));
+                }
+                _ => (),
+            };
+            checked_tokens.insert(token);
         }
-        Ok(unsupported_tokens)
+        token_quality_errors
     }
 
     async fn bulk_estimate_prices(
         &self,
         queries: &[Query],
     ) -> (Vec<EstimationProgress>, Vec<EstimationResult>) {
-        let unsupported_tokens = self.get_unsupported_tokens(queries).await.unwrap();
+        let token_quality_errors = self.get_token_quality_errors(queries).await;
 
         let mut estimations_to_forward = Vec::new();
 
         let estimation_progress = queries
             .iter()
             .map(|query| {
-                if unsupported_tokens.contains(&query.buy_token) {
-                    return EstimationProgress::TrivialSolution(Err(
-                        PriceEstimationError::UnsupportedToken(query.buy_token),
-                    ));
+                if let Some(err) = token_quality_errors.get(&query.buy_token) {
+                    return EstimationProgress::TrivialSolution(Err(err.clone()));
                 }
-                if unsupported_tokens.contains(&query.sell_token) {
-                    return EstimationProgress::TrivialSolution(Err(
-                        PriceEstimationError::UnsupportedToken(query.sell_token),
-                    ));
+                if let Some(err) = token_quality_errors.get(&query.sell_token) {
+                    return EstimationProgress::TrivialSolution(Err(err.clone()));
                 }
+
                 if query.buy_token == query.sell_token {
                     return EstimationProgress::TrivialSolution(Ok(Estimate {
                         out_amount: query.in_amount,
@@ -105,19 +111,22 @@ impl<T: PriceEstimating> SanitizedPriceEstimator<T> {
     }
 
     fn merge_trivial_and_forwarded_estimations(
+        &self,
         partially_estimated: Vec<EstimationProgress>,
         forwarded_estimations: Vec<EstimationResult>,
     ) -> Vec<EstimationResult> {
+        use EstimationProgress::*;
+
         let mut forwarded_estimations = forwarded_estimations.into_iter();
 
         partially_estimated
             .into_iter()
             .map(|progress| match progress {
-                EstimationProgress::TrivialSolution(res) => res,
-                EstimationProgress::AwaitingErc20Estimation => forwarded_estimations
+                TrivialSolution(res) => res,
+                AwaitingErc20Estimation => forwarded_estimations
                     .next()
                     .expect("there is a result for every forwarded estimation"),
-                EstimationProgress::AwaitingEthEstimation => {
+                AwaitingEthEstimation => {
                     let mut res = forwarded_estimations
                         .next()
                         .expect("there is a result for every forwarded estimation")?;
@@ -138,10 +147,7 @@ impl<T: PriceEstimating> SanitizedPriceEstimator<T> {
 impl<T: PriceEstimating> PriceEstimating for SanitizedPriceEstimator<T> {
     async fn estimates(&self, queries: &[Query]) -> Vec<EstimationResult> {
         let (trivial_estimations, forwarded_estimations) = self.bulk_estimate_prices(queries).await;
-        SanitizedPriceEstimator::<T>::merge_trivial_and_forwarded_estimations(
-            trivial_estimations,
-            forwarded_estimations,
-        )
+        self.merge_trivial_and_forwarded_estimations(trivial_estimations, forwarded_estimations)
     }
 }
 
@@ -156,7 +162,7 @@ mod tests {
     const BAD_TOKEN: H160 = H160([0x12; 20]);
 
     #[tokio::test]
-    async fn works() {
+    async fn handles_trivial_estimates_on_its_own() {
         let mut bad_token_detector = MockBadTokenDetecting::new();
         bad_token_detector.expect_detect().returning(|token| {
             if token == BAD_TOKEN {
--- a/crates/shared/src/price_estimation/sanitized.rs
+++ b/crates/shared/src/price_estimation/sanitized.rs
@@ -1,0 +1,245 @@
+use super::{ensure_token_supported, Estimate, PriceEstimating, PriceEstimationError, Query};
+use crate::bad_token::BadTokenDetecting;
+use crate::price_estimation::gas::GAS_PER_WETH_UNWRAP;
+use anyhow::Result;
+use futures::future;
+use model::order::BUY_ETH_ADDRESS;
+use primitive_types::H160;
+use std::sync::Arc;
+
+/// Verifies that buy and sell tokens are supported and handles
+/// ETH as buy token appropriately.
+pub struct SanitizedPriceEstimator<T: PriceEstimating> {
+    inner: T,
+    bad_token_detector: Arc<dyn BadTokenDetecting>,
+    native_token: H160,
+}
+
+impl<T: PriceEstimating> SanitizedPriceEstimator<T> {
+    pub fn new(
+        inner: T,
+        native_token: H160,
+        bad_token_detector: Arc<dyn BadTokenDetecting>,
+    ) -> Self {
+        Self {
+            inner,
+            native_token,
+            bad_token_detector,
+        }
+    }
+    async fn estimate_sanitized_query(
+        &self,
+        query: &Query,
+    ) -> Result<Estimate, PriceEstimationError> {
+        if query.buy_token == query.sell_token {
+            return Ok(Estimate {
+                out_amount: query.in_amount,
+                gas: 0.into(),
+            });
+        }
+
+        ensure_token_supported(query.buy_token, self.bad_token_detector.as_ref()).await?;
+        ensure_token_supported(query.sell_token, self.bad_token_detector.as_ref()).await?;
+
+        let buy_eth = query.buy_token == BUY_ETH_ADDRESS;
+        let sanitized_query = if buy_eth {
+            Query {
+                buy_token: self.native_token,
+                ..*query
+            }
+        } else {
+            *query
+        };
+
+        let mut estimated_price = self.inner.estimate(&sanitized_query).await?;
+
+        if buy_eth {
+            estimated_price.gas = estimated_price
+                .gas
+                .checked_add(GAS_PER_WETH_UNWRAP.into())
+                .ok_or(anyhow::anyhow!(
+                    "cost of unwrapping ETH would overflow gas price"
+                ))?;
+        }
+
+        Ok(estimated_price)
+    }
+}
+
+#[async_trait::async_trait]
+impl<T: PriceEstimating> PriceEstimating for SanitizedPriceEstimator<T> {
+    async fn estimates(&self, queries: &[Query]) -> Vec<Result<Estimate, PriceEstimationError>> {
+        future::join_all(
+            queries
+                .iter()
+                .map(|query| self.estimate_sanitized_query(query)),
+        )
+        .await
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::bad_token::{MockBadTokenDetecting, TokenQuality};
+    use crate::price_estimation::MockPriceEstimating;
+    use mockall::predicate::eq;
+    use model::order::OrderKind;
+    use primitive_types::{H160, U256};
+
+    const BAD_TOKEN: H160 = H160([0x12; 20]);
+
+    #[tokio::test]
+    async fn works() {
+        let mut bad_token_detector = MockBadTokenDetecting::new();
+        bad_token_detector.expect_detect().returning(|token| {
+            if token == BAD_TOKEN {
+                Ok(TokenQuality::Bad {
+                    reason: "Token not supported".into(),
+                })
+            } else {
+                Ok(TokenQuality::Good)
+            }
+        });
+
+        let native_token = H160::from_low_u64_le(1);
+
+        let queries = [
+            // This is the common case (Tokens are supported, distinct and not ETH).
+            // Will be estimated by the wrapped_estimator.
+            Query {
+                sell_token: H160::from_low_u64_le(1),
+                buy_token: H160::from_low_u64_le(2),
+                in_amount: 1.into(),
+                kind: OrderKind::Buy,
+            },
+            // `sanitized_estimator` will replace `buy_token` with `native_token` before querying
+            // `wrapped_estimator`.
+            // `sanitized_estimator` will add cost of unwrapping ETH to Estimate.
+            Query {
+                sell_token: H160::from_low_u64_le(1),
+                buy_token: BUY_ETH_ADDRESS,
+                in_amount: 1.into(),
+                kind: OrderKind::Buy,
+            },
+            // Will cause buffer overflow of gas price in `sanitized_estimator`.
+            Query {
+                sell_token: H160::from_low_u64_le(1),
+                buy_token: BUY_ETH_ADDRESS,
+                in_amount: U256::MAX,
+                kind: OrderKind::Buy,
+            },
+            // Can be estimated by `sanitized_estimator` because `buy_token` and `sell_token` are identical.
+            Query {
+                sell_token: H160::from_low_u64_le(1),
+                buy_token: H160::from_low_u64_le(1),
+                in_amount: 1.into(),
+                kind: OrderKind::Sell,
+            },
+            // Will throw `UnsupportedToken` error in `sanitized_estimator`.
+            Query {
+                sell_token: BAD_TOKEN,
+                buy_token: H160::from_low_u64_le(1),
+                in_amount: 1.into(),
+                kind: OrderKind::Buy,
+            },
+            // Will throw `UnsupportedToken` error in `sanitized_estimator`.
+            Query {
+                sell_token: H160::from_low_u64_le(1),
+                buy_token: BAD_TOKEN,
+                in_amount: 1.into(),
+                kind: OrderKind::Buy,
+            },
+        ];
+
+        let mut wrapped_estimator = MockPriceEstimating::new();
+        // Tests for estimates[0]
+        wrapped_estimator
+            .expect_estimate()
+            .times(1)
+            .with(eq(queries[0]))
+            .returning(|_| {
+                Ok(Estimate {
+                    out_amount: 1.into(),
+                    gas: 100.into(),
+                })
+            });
+        // Tests for estimates[1]
+        wrapped_estimator
+            .expect_estimate()
+            .times(1)
+            .with(eq(Query {
+                // SanitizedPriceEstimator replaces ETH buy token with native token
+                buy_token: native_token,
+                ..queries[1]
+            }))
+            .returning(|_| {
+                Ok(Estimate {
+                    out_amount: 1.into(),
+                    gas: 100.into(),
+                })
+            });
+        // Tests for estimates[2]
+        wrapped_estimator
+            .expect_estimate()
+            .times(1)
+            .with(eq(Query {
+                // SanitizedPriceEstimator replaces ETH buy token with native token
+                buy_token: native_token,
+                ..queries[2]
+            }))
+            .returning(|_| {
+                Ok(Estimate {
+                    out_amount: 1.into(),
+                    gas: U256::MAX,
+                })
+            });
+
+        let sanitized_estimator = SanitizedPriceEstimator {
+            inner: wrapped_estimator,
+            bad_token_detector: Arc::new(bad_token_detector),
+            native_token,
+        };
+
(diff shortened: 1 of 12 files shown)


crates/shared/src/price_estimation/sanitized.rs
```diff
@@ -0,0 +1,245 @@
...
+use anyhow::Result;
+use futures::future;
+use model::order::BUY_ETH_ADDRESS;
+use primitive_types::{H160, U256};
+use std::sync::Arc;
+
+const ETH_UNWRAP_COST: U256 = U256([0, 0, 0, 123]);
+
+/// Verifies that buy and sell tokens are supported and handles
+/// ETH as buy token appropriately.
```
bh2smith, line 12: Nice!

crates/shared/src/price_estimation/sanitized.rs
```diff
@@ -0,0 +1,245 @@
+use super::{ensure_token_supported, Estimate, PriceEstimating, PriceEstimationError, Query};
+use crate::bad_token::BadTokenDetecting;
+use anyhow::Result;
+use futures::future;
+use model::order::BUY_ETH_ADDRESS;
+use primitive_types::{H160, U256};
+use std::sync::Arc;
+
+const ETH_UNWRAP_COST: U256 = U256([0, 0, 0, 123]);
```
bh2smith, line 9: The total gas cost (minus the tx overhead) is 
```
35192 - 21000 = 14192
```

This is based on the following transaction: https://etherscan.io/tx/0x50e67d83c14d73ec67a3abdb720350cfe7facd232d3277739be87304e0a82314

We could be more accurate and get average out the total gas used on every unwrap, but I think that these kinds of transactions are nearly the same.

Let me know if we want to use a value based on a larger sample size.

bh2smith (approved):
This looks very good and the description itself is also quite a nice story! Really nice work here.

I included the gas used for unwrap inline - please let me know if anyone thinks we should use a larger sample size.

crates/shared/src/price_estimation/sanitized.rs
nlordell, line 9: Just wanted to let you know that we already have a module that we keep gas cost constants for this kind of thing (I think its `shared/src/price_estimation/gas.rs`). I slightly prefer having all these things in the same module. That being said, this gas cost is very specific to this price estimator, so I see the argument for keeping it here.

I don't feel strongly either way, just thought I'd point out the existence of this module in case you missed it.

crates/shared/src/price_estimation/sanitized.rs
```diff
@@ -0,0 +1,245 @@
...
+
+#[async_trait::async_trait]
+impl<T: PriceEstimating> PriceEstimating for SanitizedPriceEstimator<T> {
+    async fn estimates(&self, queries: &[Query]) -> Vec<Result<Estimate, PriceEstimationError>> {
+        future::join_all(
+            queries
+                .iter()
+                .map(|query| self.estimate_sanitized_query(query)),
+        )
+        .await
```
nlordell, line 79: nit: The `BaselinePriceEstimator` takes advantage of bundling queries into a single `estimates` "bundle" to parallelize things like fetching pool data using batch `eth_calls` (since nodes handle requests one at a time and not in parallel). With that in mind, it means when processing multiple queries, it is better to send all the `Query`-ies in a single `estimates([...])` call instead of splitting and estimating concurrently many `estimate(...)` calls (because the limiting factor - the node - doesn't process call concurrently very well).

Maybe we should take a similar approach to the `Baseline` solver and first query all the unsupported tokens:

https://github.com/gnosis/gp-v2-services/blob/56f3fa6a16140037fc07e759ec2c6981e551e8b7/crates/shared/src/price_estimation/baseline.rs#L86-L102

And then do the `buy_eth` dance for the remaining valid queries.

nlordell: If this is way too complicated, I'm open to keeping it implemented as it is here, but it would be nice to at least get an idea if it affects our `POST /api/v1/quote` request times (I wouldn't be surprised if it didn't because I don't think that the `Baseline` price estimator is the slowest one, and the slower 0x and ParaSwap estimators don't benefit from this).

But we should at least measure 😄.

nlordell (commented):
Looks good! Just one nit that I think merits some discussion.

crates/shared/src/price_estimation/sanitized.rs
MartinquaXD: That sounds absolutely reasonable. I had a similar idea where I wanted to see whether it was possible to use `PriceEstimating::estimates` instead of `PriceEstimating::estimate`. I imagine the implementation of estimating a batch could easily be more efficient than estimating a single query.
I will play around with those ideas and hope they don't make the code too complex.

MartinquaXD: Also I would think that the set of supported tokens wouldn't change during the runtime of the service.
In that case we can keep a cache of unsupported tokens and only query tokens we have never seen before.
What do you think about that?

MartinquaXD: Nevermind. The signature `async fn estimates(&self, queries: &[Query]) -> Vec<Result<Estimate, PriceEstimationError>>` doesn't allow mutating the state of `self` during the estimation and introducing something like a global `RwLock<Cache>` just for that doesn't seem like a great idea.

nlordell: > Also I would think that the set of supported tokens wouldn't change during the runtime of the service.

Unfortunately, that isn't true. Some tokens can be paused and resumed.

MartinquaXD: I was able to implement both ideas but I think the code suffered quite a bit.
1) detecting bad tokens in bulk lead to very weird error handling where a single failure determining the quality of a token leads to the whole batch being dropped. (I took that behavior from the baseline estimator).
2) estimating prices in bulk lead to this weird pattern where I split trivial estimations from estimations required to be forwarded to the wrapped estimator. Merging those 2 back together while preserving the order of the results is not trivial to understand IMO.

If there is some Rust syntax sugar or clever iterator magic we can use to make those changes easier to understand, let me know.

crates/shared/src/price_estimation/sanitized.rs
```diff
@@ -0,0 +1,311 @@
...
+    bad_token_detector: Arc<dyn BadTokenDetecting>,
+    native_token: H160,
+}
+
+type EstimationResult = Result<Estimate, PriceEstimationError>;
+
+enum EstimationProgress {
+    TrivialSolution(EstimationResult),
+    AwaitingEthEstimation,
+    AwaitingErc20Estimation,
```
nlordell, line 23: ubernit: Instead of splitting into two vectors, could this be:
```suggestion
    AwaitingEthEstimation(EstimationResult),
    AwaitingErc20Estimation(EstimationResult),
```

That way the logic in `merge_trivial_and_forwarded_estimations` is slightly easier (it doesn't have to `.next().expect(...)`).

crates/shared/src/price_estimation/sanitized.rs
```diff
@@ -0,0 +1,311 @@
...
+    use super::*;
+    use crate::bad_token::{MockBadTokenDetecting, TokenQuality};
+    use crate::price_estimation::MockPriceEstimating;
+    use model::order::OrderKind;
+    use primitive_types::{H160, U256};
+
+    const BAD_TOKEN: H160 = H160([0x12; 20]);
+
+    #[tokio::test]
+    async fn works() {
```
nlordell, line 165: naming-nit: Maybe the test name could be more descriptive. (Don't have any good suggestions though 😛).

nlordell (approved):
LGTM. Just a small suggestion on how to simplify `merge_trivial_and_forwarded_estimations`. Not blocking merge though.

crates/shared/src/price_estimation/sanitized.rs
nlordell: > If there is some Rust syntax sugar or clever iterator magic

I suggest including the `EstimationResult` as part of `EstimationProgress` variants that use it.

> Merging those 2 back together while preserving the order of the results is not trivial to understand IMO.

I think its not too complex. Maybe ☝️ helps.

nlordell: As for:

> detecting bad tokens in bulk lead to very weird error handling where a single failure determining the quality of a token leads to the whole batch being dropped. (I took that behavior from the baseline estimator).

Hmm... I'm OK with this because it was the behaviour of `Baseline` solver. The only thing I can think of is storing `Result`s of the `self.bad_token_detector.detect(token).await` call in a `HashMap<H160, Result<BadTokenResult>>` instead of a `HashSet<H160>` of bad tokens, and then checking for each token in `bulk_estimate` if the result was successful (propagating errors ofc.) and if its "good" (returning an "unsupported token" error if not).

nlordell:
Would be cool to also do a manual test and run the orderbook API and request a quote for buying ETH to see it in action 😄. Let me know if you need help running the orderbook.

crates/shared/src/price_estimation/sanitized.rs
MartinquaXD: I don't think this is possible with my current approach.
The goal is to use the bulk estimation API of the wrapped estimator. Because the `SanitizedPriceEstimator` is supposed to do as much trivial work as possible before passing queries it can't solve to the wrapped estimator.
Because the bulk estimating API takes a `&[Query]` instead of an iterator I have to have 2 `Vec` if I'm not mistaken. (1 Vec holds the trivial results, 1 Vec gets passed into `PriceEstimating::estimates(queries: &[Query])`)
The Vec holding the trivial results is supposed to have these place holders so that I'm able to easily merge the other estimates back in order.
So if I put the `EstimationResult` into the enum variants I still have to have the second Vec (otherwise I'm not able to call the bulk estimate API) while also have to create those `EstimationResult` values out of thin air to put them in the place holders.

nlordell: Ah right, of course.

crates/shared/src/price_estimation/sanitized.rs
nlordell: My idea turned out to not work 😛. I also don't have a better idea then.

bh2smith:
Agree with Nicks last suggestion! Always fun to give it a spin!

crates/shared/src/price_estimation/sanitized.rs
MartinquaXD: I think the error handling improved a bit. Instead of `HashMap<H160, Result<BadTokenResult>>` I went for `HashMap<H160, PriceEstimationError>` and converted the different errors directly in `get_token_quality_errors()`. This lead to the early returns in `bulk_estimate_prices()` being less verbose in case of an error. I'm somewhat happy with that solution.

MartinquaXD:
Yes, learning to run services locally and doing some manual tests make sense. Might need a little bit of help, though. I already tried running the services with the help of the README.md some time ago but got nowhere. Maybe the README.md is a little outdated?

crates/shared/src/price_estimation/sanitized.rs
```diff
@@ -0,0 +1,313 @@
...
+        {
+            if checked_tokens.contains(&token) {
+                continue;
+            }
+
+            match self.bad_token_detector.detect(token).await {
+                Err(err) => {
+                    token_quality_errors.insert(
+                        token,
+                        PriceEstimationError::Other(crate::clone_anyhow_error(&err)),
```
nlordell, line 59: Can't we just move the `err` here instead of cloning?
```suggestion
                        PriceEstimationError::Other(err),
```

crates/shared/src/price_estimation/sanitized.rs
```diff
@@ -0,0 +1,313 @@
...
+                        res.gas
+                            .checked_add(GAS_PER_WETH_UNWRAP.into())
+                            .ok_or(anyhow::anyhow!(
+                                "cost of unwrapping ETH would overflow gas price"
+                            ))?;
+                    Ok(res)
+                }
+            })
+            .collect()
+    }
```
nlordell, line 145: Should we `debug_assert!(forwarded_estimatesion.is_empty())` here for good measure? Might help catch some logic errors in unit tests.

nlordell (approved):
Looks great! Like the improvement in the bad token handling.

crates/shared/src/price_estimation/sanitized.rs
nlordell: Nice solution!

crates/shared/src/price_estimation/sanitized.rs
MartinquaXD: Moving is not possible here.
Because we store the errors in a map, we only have at most 1 PriceEstimationError per token. If we have 5 Queries, all using the same unsupported token, we need to clone this single error into 5 different results.

crates/shared/src/price_estimation/sanitized.rs
MartinquaXD: Good idea!
`is_empty()` lead to following error:
```
[rustc E0658] [E] use of unstable library feature 'exact_size_is_empty'
see issue #35428 <https://github.com/rust-lang/rust/issues/35428> for more information
```
So I went with `debug_assert!(forwarded_estimations.next().is_none())`.

nlordell: My suggestions are usually half-pseudo-code 😛

> is_empty() lead to following error

Makes sense, its a `vec::IntoIter` and not a `Vec`. Cool that they have a proposal for `is_empty` for `ExactSizeIterator`s though!

crates/shared/src/price_estimation/sanitized.rs
nlordell: So, I agree we need to clone the `PriceEstimationError` when checking that the tokens are supported here (as you said, this error may be for more than one query):
```rust
                if let Some(err) = token_quality_errors.get(&query.buy_token) {
                    return EstimationProgress::TrivialSolution(Err(err.clone()));
                }
                if let Some(err) = token_quality_errors.get(&query.sell_token) {
                    return EstimationProgress::TrivialSolution(Err(err.clone()));
                }
```

But I'm not 100% sure I understand why we can't move the error into the `HashMap` (since we only insert a single error per token and the `err` is dropped at the end of the `match` branch anyway).

That being said, if the compiler says "no" then it knows better 😛.

MartinquaXD: Ah, sorry. Greped for `clone_anyhow_error` and found the wrong one. 🤦‍♂️ 
When the error gets put into the map it can be moved, indeed.
Good catch!

MartinquaXD:
I added some tracing debug logs because the `SanitizedPriceEstimator` changes some things under your nose.
See the PR description.

crates/shared/src/price_estimation/sanitized.rs
MartinquaXD: Done

crates/shared/src/price_estimation/sanitized.rs
MartinquaXD: Done

crates/shared/src/price_estimation/sanitized.rs
```diff
@@ -0,0 +1,331 @@
...
+                if query.buy_token == BUY_ETH_ADDRESS {
+                    let sanitized_query = Query {
+                        buy_token: self.native_token,
+                        ..*query
+                    };
+
+                    tracing::debug!(
+                        ?query,
+                        ?sanitized_query,
+                        "estimate price for native token instead of ETH"
```
bh2smith, line 110: This has become a bit confusing for me. I was under the impression we were referring to native token as the "native asset" of the network but apparently, throughout this repo it refers to the wrapped/erc20 version of this.

In any case, I suppose it makes sense then for us all to be aware that `native token = Wrapped(native asset)` for terminology. 

Here then I might suggest we change the comment to something like

```suggestion
                        "estimate price for wrapped native asset"
```

So not to hard-code ETH (since ETH isn't the native asset of all networks). For example
```
Gnosis Chain -> xDAI
Polygon -> MATIC
BSC -> BNB
```

bh2smith: This comment you highlighted in the code seems inconsistent with the terminology we're using. On some places we are referring to native token as native asset like here.

bh2smith (approved):

crates/shared/src/price_estimation/sanitized.rs
```diff
@@ -0,0 +1,331 @@
...
+        for token in queries
+            .iter()
+            .copied()
+            .flat_map(|query| [query.buy_token, query.sell_token])
+        {
+            if checked_tokens.contains(&token) {
+                continue;
+            }
+
+            match self.bad_token_detector.detect(token).await {
```
fleupold, line 55: nanonit: in theory these could be done in parallel (likely premature optimisation but maybe something to comment as a note?)

crates/shared/src/price_estimation/sanitized.rs
```diff
@@ -0,0 +1,331 @@
...
+            };
+            checked_tokens.insert(token);
+        }
+        token_quality_errors
+    }
+
+    async fn bulk_estimate_prices<'a, 'b>(
+        &self,
+        queries: &'a [Query],
+    ) -> (Vec<EstimationProgress<'b>>, Vec<EstimationResult>)
```
fleupold, line 73: nit: I think it's not easy to realize how the estimation results are supposed to line up with the estimation progress (it might be easier to grasp if each estimation progress had their estimation result directly attached to). Maybe already merging inside this method is easier (it's private so no big deal, alternatively maybe we can document this "alignment assumption" somwhere.

crates/shared/src/price_estimation/zeroex.rs
```diff
@@ -12,21 +9,10 @@ use std::sync::Arc;
 
 pub struct ZeroExPriceEstimator {
     pub api: Arc<dyn ZeroExApi>,
-    pub bad_token_detector: Arc<dyn BadTokenDetecting>,
 }
 
 impl ZeroExPriceEstimator {
     async fn estimate(&self, query: &Query) -> Result<Estimate, PriceEstimationError> {
-        if query.buy_token == query.sell_token {
```
fleupold, line 20: Maybe let's still assert those assumptions (in case someone configures the estimator chain incorrectly)

fleupold (approved):
> This is something the SanitizedPriceEstimator does as well.

Just wondering if you considered doing the WETH/ETH correction in a separate component of this chain (to keep the logic of the bad token detection/buy==sell token) separate?

crates/shared/src/price_estimation/zeroex.rs
MartinquaXD: Done.

crates/shared/src/price_estimation/sanitized.rs
MartinquaXD: Done.

crates/shared/src/price_estimation/sanitized.rs
MartinquaXD: I'm with you on that one. I already tried splitting the code in smaller functions and giving them descriptive names but that was not good enough.
The latest version has only 1 function again but adds extensive documentation about how it works.

fleupold: Thanks! I really like the documentation.

PR #1500, opened 2021-12-21
Reject Partially Fillable Orders

fleupold:
Despite our expectation, partially fillable orders seem to be working and causing a good amount of economically unviable solutions (cf. https://etherscan.io/tokentxns?a=0x042b32ac6b453485e357938bdc38e0340d4b9276&p=3)

To be safe, we should prevent people from placing such orders via the API

### Test Plan
Added unit test, also checking that the place order script from the contracts repo yields the new error when placing a partially fillable order.


josojo (approved):
LGTM

PR #1488, opened 2021-12-16
Initial support for Balancer LBPs

nlordell:
This PR adds Balancer Liquidity Bootstrapping Pools (or LBPs for short) 🎉.

Note that this PR is based initialization cleanup in #1487 and changes to allow enabling Balancer liquidity on a per-factory basis #1498.

One small detail is that LBPs have an additional `swapEnabled` property to signal whether or not you can swap with the pool (on top of the `paused` from the base pool contract). This required us to slightly modify the `FactoryIndexing` interface to allow factory implementations to be able to signal pools that aren't tradable (here, whether or not swapping is enabled).

One additional caveat - there is a new ["no protocol fee" LPB factory](https://etherscan.io/address/0x0f3e0c4218b7b0108a3643cfe9d3ec0d4f57c54e#code) which can be added in a follow up PR (in much the same way the weighted 2-token pools are added). 

### Test Plan

Added a unit test to verify the common pool fetching code deals with "disabled" pools as well as unit test for LBP fetching logic.

<details><summary>Additionally, you can run the manual indexing test to make sure we index all LBPs:</summary>

Run the test:
```
$ cargo test -p shared -- balancer_fetched_pools_match_subgraph --ignored --nocapture
...
2021-12-22T13:01:39.678Z  WARN shared::sources::balancer_v2::pool_fetching::tests: unknown_pools=[0x04953368a77af5b65512ee3536efe152b96aa453000200000000000000000100, 0x0a54996ce9ceaa449cde73da6aa0368bfe3df6dc000200000000000000000102, 0x32f0dc9e2d890bac95106a65eb82db70bc58badb0002000000000000000000f4, 0x6a9c95dd2cc3951f23a1ab9bfac7f5a9a8301117000200000000000000000110, 0x8175c3893a54238e0f1350075d17c177bf789a240002000000000000000000ff, 0xbccd9d4b953ada653b7c464665d7263e5e078a980002000000000000000000ef, 0xc065798f227b49c150bcdc6cdc43149a12c4d75700020000000000000000010b, 0xc35bdda2e93c401c6645e0d8a0b2c86906c51710000200000000000000000111, 0xc97dee022137a8c5f65b5138cc690fbe87806ed5000200000000000000000107, 0xf1b3256505159f6004da0b16e985ee46e6ffeea80002000000000000000000fa]
test sources::balancer_v2::pool_fetching::tests::balancer_fetched_pools_match_subgraph ... ok
...
```

[...]

PR #1502, opened 2021-12-22
Add support for "No Protocol Fee" Liquidity Bootstrapping Pools

nlordell:
This PR is a follow up to #1488

Here we add support for the `NoProtocolFeeLiquidityBootstrappingPoolFactory`, which is (as the name suggests) a version of `LiquidityBootstrappingPoolFactory` without protocol fees.

It shares all indexing logic with `LiquidityBootstrappingPoolFactory`, so this PR is mostly just the boiler plate to add the new factory deployment information, as well as factory type for configuring Balancer liquidity to index this pool on startup.

### Test Plan

No new logic. Expanded existing unit test to check deployment information is set in generated contract code.

Manual test results:
```
$ cargo test -p shared -- balancer_fetched_pools_match_subgraph --ignored --nocapture
...
2021-12-22T15:14:34.155Z  WARN shared::sources::balancer_v2::pool_fetching::tests: unknown_pools=[]
test sources::balancer_v2::pool_fetching::tests::balancer_fetched_pools_match_subgraph ... ok
...
```

### DIFF
--- a/crates/contracts/src/lib.rs
+++ b/crates/contracts/src/lib.rs
@@ -13,6 +13,10 @@ include!(concat!(
     env!("OUT_DIR"),
     "/BalancerV2LiquidityBootstrappingPoolFactory.rs"
 ));
+include!(concat!(
+    env!("OUT_DIR"),
+    "/BalancerV2NoProtocolFeeLiquidityBootstrappingPoolFactory.rs"
+));
 include!(concat!(env!("OUT_DIR"), "/BalancerV2StablePool.rs"));
 include!(concat!(env!("OUT_DIR"), "/BalancerV2StablePoolFactory.rs"));
 include!(concat!(env!("OUT_DIR"), "/BalancerV2Vault.rs"));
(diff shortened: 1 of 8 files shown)


crates/contracts/src/lib.rs
```diff
@@ -112,6 +116,8 @@ mod tests {
         }
         for network in &[1, 4] {
             assert_has_deployment_address!(BalancerV2Vault for *network);
+            assert_has_deployment_address!(BalancerV2LiquidityBootstrappingPoolFactory for *network);
```
nlordell, line 119: unrelated change that I snuck in.

crates/shared/src/sources/balancer_v2/pools/liquidity_bootstrapping.rs
```diff
@@ -1,4 +1,4 @@
-//! Module implementing two-token weighted pool specific indexing logic.
+//! Module implementing liquidity bootstrapping pool specific indexing logic.
```
nlordell, line 1: Trivial unrelated change that I snuck in.

bh2smith: This is part of the previous PR and I also a suggestion comment to fix it there. Just FYI

bh2smith (approved):
Looks good. I see that these no protocol fee pools are being treated as a completely different pool type. Is there any reason we didn't go with having the same pool type with an additional parameter or are they really all that different?

crates/shared/src/sources/balancer_v2/pools/liquidity_bootstrapping.rs
nlordell: Merged your suggestion :smil

nlordell:
> I see that these no protocol fee pools are being treated as a completely different pool type.

They are the same "pool type" in that they have the same `PoolInfo` and `PoolState` types and produce `WeightedPools`. They also reuse 100% of the logic from the existing LBP implementation. However, they need to be indexed separately because they have a separate factory address. Also, since they are a separate factory, we also want the ability to enable/disable them independently (hence the additional `BalancerFactoryKind` variant).

fleupold (approved):
```
unknown_pools=[]
```

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2FMdqE46HziuFJTlIwjw%2Fgiphy.gif&token=yay" data-gitmeme-token="yay"></a>

Issue #1479, opened 2021-12-13
Minimum Fee Amount


nlordell:
Add a configuration parameter for setting a minimum amount (in Ether) for fees. Last week when we had particularly network gas fees, our flat subsidy was configured high enough that it would cause the Orderbook API to accept orders with 0 fees (so free trading!).

While this did not cause us to burn a particularly high amount of gas money, it did have a side effect of enabling free trading which opens a whole can of worms (things like wash trading for example, or airdrop farming).

PR #1504, opened 2021-12-23
Configure minimum discounted fee

nlordell:
This PR adds a configuration option for setting a minimum discounted fee amount. We want this because we currently have a flat fee subsidy, and in times of low gas prices, it _can_ happen that we end up with a 0 or negative fee amount after applying this fee discount.

In order to avoid these situations (since free trading is nice for our users but causes a bunch of other problems - like wash trading for example), we add a new configuration option for a lower bound on how much the fee can be discounted.

One quirk about how this was implemented, the amount represents the minimum fee **before** any multiplicative subsidies are applied. The reason for doing this is that:
- It made the code much easier (we already had logic for this, just hard-coded at 0).
- It was unclear to me whether or not we wanted things like the global fee factor and specifically partner fee factors to allow the final subsidized fee to go lower than this (e.g. PMMs should be able to place orders below this minimum right?).

Since I'm not 100% sure about the semantics here, I started a discussion in Slack. Feel free to review the rest of the code in the meantime.

### Test Plan

Adapted unit test for fees being capped at 0 to use a minimum amount.

### DIFF
--- a/crates/orderbook/src/fee.rs
+++ b/crates/orderbook/src/fee.rs
@@ -37,11 +37,19 @@ pub struct FeeSubsidyConfiguration {
     ///
     /// Flat fee discounts are applied **before** any multiplicative discounts.
     pub fee_discount: f64,
+
+    /// Minimum fee amount after applying the flat subsidy. This prevents flat
+    /// fee discounts putting the fee amount below 0.
+    ///
+    /// Flat fee discounts are applied **before** any multiplicative discounts.
+    pub min_discounted_fee: f64,
+
     /// A factor to multiply the estimated trading fee with in order to compute
     /// subsidized minimum fee.
     ///
     /// Fee factors are applied **after** flat fee discounts.
     pub fee_factor: f64,
+
     /// Additional factors per order app ID for computing the subsidized minimum
     /// fee.
     ///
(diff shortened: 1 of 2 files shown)


fleupold (approved):
Great!

MartinquaXD (approved):
The changes look good overall.
I'm just a little concerned that the fee parameters are of type `f64`. Isn't that usually frowned upon if you are dealing with money because of rounding errors?
But since this is already an established pattern I don't see a reason to delay this PR.

bh2smith (commented):
This looks fine, but I was expecting to see a change in some of the formulas s discussed in the slack [here](https://gnosisinc.slack.com/archives/CC3C5SMB5/p1640271637140400) that relates the min discounted fee with the other calculated fees like `fee = max(minFee,calculatedFee)` to ensure we always use the lower bound. Perhaps I have misunderstood the conversation. Or is this planned for a follow up PR?

josojo (approved):

fleupold:
I'm going to merge this branch since it's accepted without any comments and this feature shouldn't wait another week.

MartinquaXD:
@nlordell this PR solves #1479, right?

nlordell:
It does! Sorry, forgot to link the issue with the PR.

PR #1507, opened 2021-12-29
Use Settlement Contract's Internal Buffer to Optimize Unwraps

### LINKED ISSUE #1496: Optimize Native Token Unwraps
Based on an idea to save gas costs, this change ultimately reduces the number of unwrap interactions executed in batch settlements.

### Background

This [Dune Query](https://dune.xyz/queries/304513) shows the total amount of ETH spent on unwraps. Compared with our overall costs this is approximately 3% of the all time.

Up until now, there has not been a single batch containing ETH as the buy token in which we did not unwrap WETH, which suggests that there is no logic in the driver code (as of yet) to use internal balances. Here is a [dune query](https://dune.xyz/queries/307139) showing all the trades buying ETH contained inside a batch which does not include an unwrap interaction (this can be used to monitor the effect of the change once it has been implemented).

To achieve this, we will break the task into three independent sub task, each of which having a small impact on the end result. Essentially take the total unwrap amount at settlement time, compare it with the contract's ETH balance, if it decides it needs to unwrap, then it unwraps ~60% of its WETH balance into so to ensure it doesn’t have to unwrap again for a while.

### Task Breakdown

1.  [ ] Add logic in the settlement submission that is aware of the settlement contract's native ETH balance and excludes the unwrap interaction from encoded settlement whenever sufficient to settle trades involved. This change would be effective on its own by simply "topping up" the settlement contract with ETH balance, so we would immediately see results in the query above and start saving gas. This is, however a manual process of sending ETH to the contract that we would like to avoid.
2. [ ] Rework the internal buffer logic to incorporate ETH as a supported token.
3. [ ] In the third, and final step we would like to eliminate the manual process of sending ETH into the contract by making the driver aware of both its WETH and ETH balances and unwrap more than necessary when necessary to, essentially, be responsible for topping up its own ETH balances. It was suggested by Martin K. to unwrap 60% of the total WETH balance whenever unwraps are warranted. In reality we will have to unwrap at least

```js
MAX(
    ETH needed to settle the batch, 
    0.6 * (settlement contract WETH balance + amount of WETH being purchased from DEXES in the current batch)
)
```

Note that in the third step, we must take special care and account more than just the contract's internal balances, but also the WETH that winds up being purchased within the settlement itself. 

### Example of weird scenario
For example, consider a batch containing a trade buying 2 ETH. The settlement contract has 1 WETH and 0 ETH. This implies we will have to make a dex interaction to settle the trade (buying 2 ETH), we would then have 3 WETH, but we need to unwrap at least 2 to settle the trade (i.e. more than 60%).

There is already an old open PR #1313 containing additional discussion on the topic that I will now close in favour of this well defined three task issue description. The only valuable code there at this point that could be recycled is the following block for WETH/ETH balance fetching

```rs
pub async fn get_eth_weth_balance(web3: &Web3, address: H160) -> Result<(U256, U256)> {
    let weth = WETH9::deployed(web3).await?;
    // TODO - fetch the next two in parallel
    let weth_balance = weth.methods().balance_of(address).call().await?;
    let native_asset_balance = &web3.eth().balance(address, None).await?;
    Ok((weth_balance, native_asset_balance.into()))
}
```

Note that the weth and eth balances can be fetched in parallel.

MartinquaXD:
Fixes #1496

This change will cause the internal buffers of the settlement contract to be used for unwraps if somebody buys ETH in a batch.
Before this change, every ETH buy order would result in an unwrap operation during the settlement. With this change we try to buffer unwraps such that we unwrap less frequently. But when we do, we unwrap way more than the current settlement needs so we don't have to unwrap again for a few settlements.
This basically uses the same idea as a `Vec<T>` trying to reduce the number of reallocations when you push elements into it.

If the settlement's unwraps can be paid with the settlement contract's ETH buffer, it does so and simply drops the unwrap operation of the settlement. If that is not possible, the driver computes a WETH amount to unwrap given a user defined factor. If that unwrap would exceed what the current settlement needs, the driver executes the unwrap. If that would NOT yield enough ETH for the settlement, the driver does nothing and leaves the settlement as is which results in the behavior we had before this PR.

The user can define how much of the WETH buffer may be unwrapped by passing `--weth-unwrap-factor` to the solver. This value is supposed to be in the range of [0.0, 1.0]. The unwrap amount is equal to `ETH_BALANCE * weth_unwrap_factor`. I followed the recommendation of 0.6 as the default value like it has been suggested in the linked issue.

For more information about the issue, have a look at the linked issue. It is quite detailed.

### Test Plan
Added unit tests for the algorithm optimizing unwraps.

Did a manual test trading GNO -> ETH on rinkeby without unwraps: https://rinkeby.etherscan.io/tx/0xf326ce6ce3d2da376d233f73d4d9646054eff0d91181825ba6eb8512f98003b5

### DIFF
--- a/crates/e2e/tests/e2e/onchain_settlement.rs
+++ b/crates/e2e/tests/e2e/onchain_settlement.rs
@@ -242,6 +242,7 @@ async fn onchain_settlement(web3: Web3) {
         10,
         create_orderbook_api(),
         create_order_converter(&web3, contracts.weth.address()),
+        0.0,
     );
     driver.single_run().await.unwrap();
 
--- a/crates/shared/src/arguments.rs
+++ b/crates/shared/src/arguments.rs
@@ -141,10 +141,10 @@ pub struct Arguments {
     pub mip_uses_internal_buffers: bool,
 }
 
-pub fn parse_unbounded_percentage_factor(s: &str) -> Result<f64> {
-    let percentage_factor = f64::from_str(s)?;
-    ensure!(percentage_factor.is_finite() && percentage_factor >= 0.);
-    Ok(percentage_factor)
+pub fn parse_unbounded_factor(s: &str) -> Result<f64> {
+    let factor = f64::from_str(s)?;
+    ensure!(factor.is_finite() && factor >= 0.);
+    Ok(factor)
 }
 
 pub fn parse_percentage_factor(s: &str) -> Result<f64> {
--- a/crates/shared/src/arguments.rs
+++ b/crates/shared/src/arguments.rs
@@ -141,10 +141,16 @@ pub struct Arguments {
     pub mip_uses_internal_buffers: bool,
 }
 
-pub fn parse_fee_factor(s: &str) -> Result<f64> {
-    let fee = f64::from_str(s)?;
-    ensure!(fee.is_finite() && fee >= 0.);
-    Ok(fee)
+pub fn parse_unbounded_percentage_factor(s: &str) -> Result<f64> {
+    let percentage_factor = f64::from_str(s)?;
+    ensure!(percentage_factor.is_finite() && percentage_factor >= 0.);
+    Ok(percentage_factor)
+}
+
+pub fn parse_percentage_factor(s: &str) -> Result<f64> {
+    let percentage_factor = f64::from_str(s)?;
+    ensure!(percentage_factor.is_finite() && percentage_factor >= 0. && percentage_factor <= 1.0);
+    Ok(percentage_factor)
 }
 
 pub fn duration_from_seconds(s: &str) -> Result<Duration, ParseFloatError> {
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -59,6 +59,7 @@ pub struct Driver {
     api: OrderBookApi,
     order_converter: OrderConverter,
     in_flight_orders: InFlightOrders,
+    weth_unwrap_factor: f64,
 }
 impl Driver {
     #[allow(clippy::too_many_arguments)]
@@ -83,6 +84,7 @@ impl Driver {
         max_settlements_per_solver: usize,
         api: OrderBookApi,
         order_converter: OrderConverter,
+        weth_unwrap_factor: f64,
     ) -> Self {
         Self {
             settlement_contract,
@@ -107,6 +109,7 @@ impl Driver {
             api,
             order_converter,
             in_flight_orders: InFlightOrders::default(),
+            weth_unwrap_factor,
         }
     }
 
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -500,9 +500,10 @@ impl Driver {
                 winning_settlement
             );
 
-            self.post_processing_pipeline
+            winning_settlement.settlement = self
+                .post_processing_pipeline
                 .optimize_settlement(
-                    &mut winning_settlement.settlement,
+                    winning_settlement.settlement,
                     winning_solver.account().clone(),
                     gas_price,
                 )
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -9,6 +9,7 @@ use crate::{
     metrics::{SolverMetrics, SolverRunOutcome},
     orderbook::OrderBookApi,
     settlement::Settlement,
+    settlement_post_processing::PostProcessingPipeline,
     settlement_simulation,
     settlement_submission::SolutionSubmitter,
     solver::{Auction, SettlementWithSolver, Solver, Solvers},
@@ -59,7 +60,7 @@ pub struct Driver {
     api: OrderBookApi,
     order_converter: OrderConverter,
     in_flight_orders: InFlightOrders,
-    weth_unwrap_factor: f64,
+    post_processing_pipeline: PostProcessingPipeline,
 }
 impl Driver {
     #[allow(clippy::too_many_arguments)]
@@ -86,6 +87,13 @@ impl Driver {
         order_converter: OrderConverter,
         weth_unwrap_factor: f64,
     ) -> Self {
+        let post_processing_pipeline = PostProcessingPipeline::new(
+            native_token,
+            web3.clone(),
+            weth_unwrap_factor,
+            settlement_contract.clone(),
+        );
+
         Self {
             settlement_contract,
             liquidity_collector,
@@ -109,7 +117,7 @@ impl Driver {
             api,
             order_converter,
             in_flight_orders: InFlightOrders::default(),
-            weth_unwrap_factor,
+            post_processing_pipeline,
         }
     }
 
@@ -491,6 +499,15 @@ impl Driver {
                 winning_solver.name(),
                 winning_settlement
             );
+
+            self.post_processing_pipeline
+                .optimize_settlement(
+                    &mut winning_settlement.settlement,
+                    winning_solver.account().clone(),
+                    gas_price,
+                )
+                .await;
+
             self.metrics
                 .complete_runloop_until_transaction(start.elapsed());
             let start = Instant::now();
--- a/crates/solver/src/settlement/settlement_encoder.rs
+++ b/crates/solver/src/settlement/settlement_encoder.rs
@@ -340,6 +340,24 @@ impl SettlementEncoder {
             None => U256::one().to_big_rational(),
         }
     }
+
+    /// Drops all UnwrapWethInteractions for the given token address.
+    /// This can be used in case the settlement contracts ETH buffer is big enough.
+    pub fn drop_unwrap(&mut self, token: H160) {
+        self.unwraps.retain(|unwrap| unwrap.weth.address() != token);
+    }
+
+    /// Calculates how much of a given token this settlement will unwrap during the execution.
+    pub fn amount_to_unwrap(&mut self, token: H160) -> U256 {
+        self.unwraps.iter().fold(U256::zero(), |sum, unwrap| {
+            if unwrap.weth.address() == token {
+                sum.checked_add(unwrap.amount)
+                    .expect("no settlement would pay out that much ETH at once")
+            } else {
+                sum
+            }
+        })
+    }
 }
 
 #[cfg(test)]
--- a/crates/solver/src/settlement/settlement_encoder.rs
+++ b/crates/solver/src/settlement/settlement_encoder.rs
@@ -348,7 +348,7 @@ impl SettlementEncoder {
     }
 
     /// Calculates how much of a given token this settlement will unwrap during the execution.
-    pub fn amount_to_unwrap(&mut self, token: H160) -> U256 {
+    pub fn amount_to_unwrap(&self, token: H160) -> U256 {
         self.unwraps.iter().fold(U256::zero(), |sum, unwrap| {
             if unwrap.weth.address() == token {
                 sum.checked_add(unwrap.amount)
--- a/crates/solver/src/settlement_post_processing/mod.rs
+++ b/crates/solver/src/settlement_post_processing/mod.rs
@@ -2,6 +2,7 @@ pub mod optimize_unwrapping;
 
 use crate::settlement::Settlement;
 use crate::settlement_simulation::simulate_and_estimate_gas_at_current_block;
+use crate::solver::http_solver::buffers::BufferRetriever;
 use contracts::{GPv2Settlement, WETH9};
 use ethcontract::Account;
 use gas_estimation::EstimatedGasPrice;
@@ -14,6 +15,7 @@ pub struct PostProcessingPipeline {
     unwrap_factor: f64,
     settlement_contract: GPv2Settlement,
     weth: WETH9,
+    buffer_retriever: BufferRetriever,
 }
 
 impl PostProcessingPipeline {
@@ -24,12 +26,14 @@ impl PostProcessingPipeline {
         settlement_contract: GPv2Settlement,
     ) -> Self {
         let weth = WETH9::at(&web3, native_token);
+        let buffer_retriever = BufferRetriever::new(web3.clone(), settlement_contract.address());
 
         Self {
             web3,
             unwrap_factor,
             settlement_contract,
             weth,
+            buffer_retriever,
         }
     }
 
@@ -50,21 +54,11 @@ impl PostProcessingPipeline {
             matches!(result, Ok(results) if results[0].is_ok())
         };
 
-        let get_weth_balance = || async {
-            // TODO cache the WETH balance if we add more optimizations in the future which need it
-            self.weth
-                .methods()
-                .balance_of(self.settlement_contract.address())
-                .call()
-                .await
-                .map_err(|e| anyhow::anyhow!(e))
-        };
-
         // an error will leave the settlement unmodified
         optimize_unwrapping(
             settlement,
             &settlement_would_succeed,
-            &get_weth_balance,
+            &self.buffer_retriever,
             &self.weth,
             self.unwrap_factor,
         )
--- a/crates/solver/src/settlement_post_processing/mod.rs
+++ b/crates/solver/src/settlement_post_processing/mod.rs
@@ -60,6 +60,7 @@ impl PostProcessingPipeline {
                 .map_err(|e| anyhow::anyhow!(e))
         };
 
+        // an error will leave the settlement unmodified
         let _ = optimize_unwrapping(
             settlement,
             &settlement_would_succeed,
--- a/crates/solver/src/settlement_post_processing/mod.rs
+++ b/crates/solver/src/settlement_post_processing/mod.rs
@@ -10,10 +10,38 @@ use optimize_unwrapping::optimize_unwrapping;
 use primitive_types::H160;
 use shared::Web3;
 
+/// Determines whether a settlement would be executed successfully.
+#[cfg_attr(test, mockall::automock)]
+#[async_trait::async_trait]
+pub trait SettlementSimulating: Send + Sync {
+    async fn settlement_would_succeed(&self, settlement: Settlement) -> bool;
+}
+
+pub struct SettlementSimulator {
+    web3: Web3,
+    settlement_contract: GPv2Settlement,
+    gas_price: EstimatedGasPrice,
+    solver_account: Account,
+}
+
+#[async_trait::async_trait]
+impl SettlementSimulating for SettlementSimulator {
+    async fn settlement_would_succeed(&self, settlement: Settlement) -> bool {
+        let result = simulate_and_estimate_gas_at_current_block(
+            std::iter::once((self.solver_account.clone(), settlement)),
+            &self.settlement_contract,
+            &self.web3,
+            self.gas_price,
+        )
+        .await;
+        matches!(result, Ok(results) if results[0].is_ok())
+    }
+}
+
 pub struct PostProcessingPipeline {
     web3: Web3,
-    unwrap_factor: f64,
     settlement_contract: GPv2Settlement,
+    unwrap_factor: f64,
     weth: WETH9,
     buffer_retriever: BufferRetriever,
 }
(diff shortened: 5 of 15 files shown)


MartinquaXD:
One thing I'm unhappy about is the introduction of the `shutdown_sender`. If anybody has a good idea how to ditch that, let me know.

crates/e2e/tests/e2e/onchain_settlement.rs
```diff
@@ -136,6 +136,7 @@ async fn onchain_settlement(web3: Web3) {
         block_stream,
         solvable_orders_cache,
         base_tokens,
+        ..
```
bh2smith, line 139: I guess this is the rust notation for a partial destructure.

crates/e2e/tests/e2e/unwrap_cost_optimization.rs
```diff
@@ -0,0 +1,388 @@
+use crate::deploy::Contracts;
```
bh2smith, line 1: Looks like the content of this file was moved from elsewhere. If there is any new code here it's difficult to tell....

bh2smith (commented):
Looks pretty good but that one new file is difficult to read... is that all new code?

crates/e2e/tests/e2e/unwrap_cost_optimization.rs
MartinquaXD: I started with the eth_integration tests as the basis for this new test but had to rewrite quite a bit.

crates/e2e/tests/e2e/unwrap_cost_optimization.rs
```diff
@@ -0,0 +1,388 @@
...
+            0_u64.into(),
+            solver.address(),
+            U256::max_value(),
+        )
+    );
+
+    (solver, trader_a, trader_b, token)
+}
+
+async fn set_up_driver_and_services(
```
sunce86, line 113: Could we avoid calling `OrderbookServices::new` twice for one test by separating this function to two functions? Is it necessary to have two `OrderbookServices` objects in one test?

sunce86 (approved):
Looks good. Well explained PR.

crates/shared/src/arguments.rs
```diff
@@ -147,6 +147,12 @@ pub fn parse_fee_factor(s: &str) -> Result<f64> {
     Ok(fee)
 }
 
+pub fn parse_unwrap_factor(s: &str) -> Result<f64> {
```
fleupold, line 150: nit: My initial thought was "why don't we reuse the same parsing as for the fee factor?" (answer, because this one here is upper bound by 1).

Maybe we can rename these methods to be more explanatory on the kind of parsing they do (e.g. `parse_percentage_factor` vs `parse_unbounded_percentage_factor`) rather than which field they are used for (I would like them to be reused if we add another field)?

crates/solver/src/driver.rs
```diff
@@ -145,13 +148,70 @@ impl Driver {
         .await
     }
 
+    async fn optimize_cost_of_unwrapping(&self, settlement: &mut Settlement) -> Result<()> {
```
fleupold, line 151: I would like to avoid adding this logic to `driver.rs` as it is already little bit our "messy cellar" where we have a lot of different logic together.
I was originally thinking that this kind of "settlement post processing" could be it's own component/pipeline where we can add multiple optimisations in a row. With this being our first one, it might be a bit too early to build an abstraction (we don't really know yet what other processing steps we will have, if any), but maybe we can move this logic into it's own submodule (settlement_post_processing or the like).

Also, I feel this method doesn't necessarily require a e2e test, but could well be covered by unit tests (given that the settlement is supposed to be pretty much a data struct and we are "just" doing some logically involved modifications to it). Currently the difficult part for a unit test is that we need to know the buffer which requires an async call on an unmockable web3 instance.

We already have a `BufferRetrieving` trait which is used by the `http_solver` to pass internal buffers into our MiP and Quasimodo solver. This trait could be extended to also support the native token (right now it's limited to ERC20) and then used here. We could then easily mock it and have unit tests covering all of the scenarios.

This is also brings the small performance advantage that we could create a cache of buffers "per settlement run" to avoid double fetching these balances (not needed for this PR).

crates/solver/src/driver.rs
```diff
@@ -145,13 +148,70 @@ impl Driver {
...
+
+        if buffered_eth >= eth_for_settlement {
+            // No need to unwrap yet, we have enough ETH buffered
+            tracing::debug!(?eth_for_settlement, "use internal buffer for unwraps");
+            settlement.encoder.drop_unwrap(self.native_token);
+            return Ok(());
+        }
+
+        let weth = contracts::WETH9::at(&self.web3, self.native_token);
+        let weth_balance = weth
```
fleupold, line 174: One complexity that I didn't think of before is, what if the solution is already planning to use part of the WETH buffer (we now allow MiP and Quasimodo to use internal buffers during their solution finding)? Then converting 60% of it to ETH may cause the settlement to fail. I'm not sure if we can somehow know the state of the buffers after the settlement was applied. 

It may require a simulation + balance query which is not trivial to implement without using [`trace_callMany`](https://openethereum.github.io/JSONRPC-trace-module#trace_callmany) (a feature of OpenEthereum, the node type we want to stop depending on) or building a [`StorageReaderExtension`](https://www.youtube.com/watch?v=4BfAEIp2_No) for our smart contract.

Alternatively, we could simulate the "optimised settlement" and discard if it fails or only perform this optimisation for our non-http-solvers (basically requiring solvers to either do all or none of the "internal buffer" optimisation)

crates/solver/src/driver.rs
```diff
@@ -145,13 +148,70 @@ impl Driver {
...
+            );
+            settlement.encoder.drop_unwrap(self.native_token);
+            settlement
+                .encoder
+                .add_unwrap(crate::interactions::UnwrapWethInteraction {
+                    weth,
+                    amount: amount_to_unwrap,
+                });
+        } else {
+            tracing::warn!("consider topping up settlement contract's internal buffer");
```
fleupold, line 197: Not needed for now, but maybe a metric by how much we would have had to top up the contract could be useful to quantify the correct amount to use on average (can maybe be done using Dune as well)

fleupold (commented):
I would really like to explore if we can get around without an e2e test. They are quite heavy weight, so we try to only test major features (new order types, settlement flows). Post processing could be considered such a major feature, but if we can test the logic in a unit test, I'd prefer that.

Also one comment regarding our HTTP solvers using buffers already partially now.

MartinquaXD:
Thanks for the great feedback!
It seems that the major concerns can be addressed by replacing the e2e test with a unit test. This means I don't have to manage the lifetime of the orderbook service which caused me to introduce the shutdown mechanism in the first place.
I initially wanted to have an e2e test to see whether I overlooked any relevant nuances in the interaction between the contract and the driver. Since that doesn't seem to be the case, a unit test should suffice.

crates/e2e/tests/e2e/services.rs
```diff
@@ -187,12 +188,20 @@ impl OrderbookServices {
...
             db.clone(),
             orderbook,
             quoter,
             API_HOST[7..].parse().expect("Couldn't parse API address"),
-            pending(),
+            async {
+                if shutdown_receiver.await.is_err() {
+                    // Tests dropping the shutdown_sender would terminate the service by accident.
+                    // Let's wait for an explicit shutdown message before terminating the service.
+                    std::future::pending::<()>().await;
```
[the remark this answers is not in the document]
MartinquaXD: The pending future we had before never got dropped. The problem started happening after I started using the oneshot channel for the shutdown mechanism.
Just replacing the `pending()` future with a future waiting for a message from a oneshot channel caused the future to terminate once the channel sender has been dropped. This meant all the existing e2e tests would have to keep the `shutdown_sender` alive. To not make every test manage the `shutdown_sender` when they never had the intention to terminate the service, I required an actual shutdown message to be sent (which can be differentiated from termination caused by dropping the sender). If `shutdown_receiver.await` terminated because the sender got dropped, it was one of the tests which didn't want to terminate the service anyway so I used `pending()` as the expected fallback behavior.
But that doesn't matter anymore anyway because most of the problems can be solved by replacing the e2e with an unit test altogether.

crates/shared/src/arguments.rs
MartinquaXD: Good suggestions, will do.

crates/solver/src/driver.rs
MartinquaXD: I assumed `SettlementEncoder::unwraps` contained all unwraps which will happen during the execution of the settlement but judging by this comment this probably only accounts for the ETH which gets paid out to the traders in the end.
Could we maybe get the needed information by analyzing `SettlementEndoder::execution_plan`? Like sum up all `Interactions` which take from our internal buffer.

crates/e2e/tests/e2e/services.rs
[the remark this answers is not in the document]
fleupold: I see, my bad was that I didn't realise that if shutdown_receiver.await.is_ok() we also resolve the future.

crates/solver/src/driver.rs
fleupold: Afaik, `SettlementEncoder::unwraps` does contains all unrwaps. The problem is that we don't necessarily execute all swaps the are needed for the settlement to "perfectly clear" (sample settlement: https://etherscan.io/tx/0xec7fcc9c59e3b56d29ea7feb3f79afc2fb1c0c73e7c0049075be1666a64f2ced, see how we do a bunch of trades without any external swap - the imbalance is paid from "internal buffers")

Here is an example where this new logic would fail: Let's say we have 10 WETH (0 ETH) in the contract and there is a trade selling 20k USDC for 5 WETH together with a trade selling some shitcoin (which is not "whitelisted" for internal buffer trading) for 1 ETH. 
Since recently, our http solvers receive information about the buffers as part of the auction instance and may - in order to be more gas efficient - decide to use less interactions (e.g. skip the $20k -> 5 WETH swap) and pay those from internal buffers.

In this case the logic here would decide to unwrap 6 WETH to ETH, however this wouldn't leave enough WETH to pay for the $20k -> 5 WETH trade.

Judging the expected in/out amounts from interactions would require adding the in/out token + amounts to the `Interaction` trait - something we have spoken about doing in the past (it might also be relevant for a smarter solutions submission strategy that @sunce86 is working on). I'm not sure how big of a refactoring this would be and if it would be simpler for now to simulate the optimised solution (maybe only necessary if we add an unwrap) and discard it if it's no longer valid.

MartinquaXD:
* created a new module for post processing `Settlements` (instead of putting it in `driver.rs` directly)
* replaced expensive e2e tests with cheap unit tests

Making the API better testable made it a little weird. Maybe I should pass `impl Trait` instead of closures in the optimization algorithm?

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -340,6 +340,24 @@ impl SettlementEncoder {
             None => U256::one().to_big_rational(),
         }
     }
+
+    /// Drops all UnwrapWethInteractions for the given token address.
+    /// This can be used in case the settlement contracts ETH buffer is big enough.
+    pub fn drop_unwrap(&mut self, token: H160) {
```
vkgnosis, line 346: I think there can only be one "weth" address per network. So it doesn't make sense to have the token address as a parameter here. Maybe someone else could confirm that.

crates/solver/src/settlement_post_processing/optimize_unwrapping.rs
```diff
@@ -0,0 +1,201 @@
+use crate::settlement::Settlement;
+use anyhow::Result;
+use contracts::WETH9;
+use primitive_types::U256;
+
+pub async fn optimize_unwrapping<V, VFut, B, BFut>(
+    settlement: &mut Settlement,
```
vkgnosis, line 7: Since we have to clone the settlement anyway maybe it would be nicer to take and return it by value. That way the caller can't ignore the result so easily and it is clearer under what conditions the settlement changes. With the current signature it might not be clear whether the settlement will be modified if the function returns Err.

crates/solver/src/settlement_post_processing/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+        let _ = optimize_unwrapping(
+            settlement,
+            &settlement_would_succeed,
+            &get_weth_balance,
+            &self.weth,
+            self.unwrap_factor,
+        )
+        .await;
+
+        debug_assert!(settlement_would_succeed(settlement.clone()).await);
```
vkgnosis, line 72: I don't think this is useful because optimize_unwrapping already does it and it feels weird  to do IO in a debug assert but not sure.

vkgnosis:
> Making the API better testable made it a little weird. Maybe I should pass `impl Trait` instead of closures in the optimization algorithm?

It might be easier to extract the IO you need into a trait that can be mocked. You can look at other examples in the code where use mockall::automock.

crates/solver/src/settlement_post_processing/optimize_unwrapping.rs
MartinquaXD: I used an out parameter to leave the original `Settlement` untouched while also reporting an error. With `-> Result<Settlement>` we would loose the original settlement in case of an error and with `-> Settlement` we would not be able to report any errors at all.
Because there is currently no good way to handle an error anyway, I could simply drop the error handling and always return a settlement.

crates/solver/src/settlement_post_processing/mod.rs
MartinquaXD: I thought it nicely conveys the post conditions of the `PostProcessingPipeline` in case some more stages get added to it in the future. But I would also be fine with dropping it.

crates/solver/src/settlement_post_processing/optimize_unwrapping.rs
vkgnosis: We would have to clone it to not lose it, yes (or return it in the Error). I do think this way is fine too so no need to change it. Maybe add a comment saying that settlement is not modified if the Result is Err.

crates/solver/src/settlement/settlement_encoder.rs
MartinquaXD: In that case wouldn't it also make sense to make `SettlementEncoder::unwraps` of type `U256` instead of `Vec<UnwrapWethInteraction`? Or maybe `UnwrapWethInteraction` if we still need the included `WETH9` struct?

vkgnosis: I think atm we need the weth9 instance so that we can implement the Interaction trait. With some refactoring we might be able to change that too and it probably would be slightly cleaner.

crates/solver/src/settlement_post_processing/mod.rs
nlordell: My worry is that network changes between the initial simulation and the final simulation for the `debug_assert` may cause settlements to start failing.

Since this is a `debug_assert` and doesn't actually run in our deployments (we use release builds AFAIK) its not that big of an issue, but it may cause unexpected panics when running locally in debug mode.

crates/solver/src/settlement_post_processing/optimize_unwrapping.rs
```diff
@@ -0,0 +1,206 @@
...
+
+/// Tries to do one of 2 optimizations.
+/// 1) Drop WETH unwraps and instead pay ETH with the settlment contract's buffer.
+/// 2) Top up settlement contract's ETH buffer by unwrapping way more WETH than this settlement
+///    needs. This will cause the next few settlements to use optimization 1.
+/// If this function returns Err(), the original settlement will not be modified in any way.
+pub async fn optimize_unwrapping<V, VFut, B, BFut>(
+    settlement: &mut Settlement,
+    settlement_would_succeed: V,
+    get_weth_balance: B,
```
nlordell, line 14: Cool use of closures to make the method testable instead of using a trait and mocks.

crates/solver/src/settlement_post_processing/optimize_unwrapping.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    // simulate settlement without unwrap
+    let mut optimized_settlement = settlement.clone();
+    optimized_settlement.encoder.drop_unwrap(weth.address());
+
+    if settlement_would_succeed(optimized_settlement.clone()).await {
+        tracing::debug!("use internal buffer to unwraps");
+        *settlement = optimized_settlement;
+        return Ok(());
+    }
```
nlordell, line 37: At first, I thought that we could just use the condition `if required_eth_payout <= weth_balance { ... }` here (in order to avoid a node round-trip), but it turns out, you can't because it doesn't capture solvers using `ETH` buffers for some other purpose which could leave not enough `ETH` for the payout at the end. Maybe add a comment here as to why we need a simulation instead of just checking amounts?

crates/solver/src/settlement/settlement_encoder.rs
nlordell: > I think there can only be one "weth" address per network. So it doesn't make sense to have the token address as a parameter here.

The reason for this is that there are multiple "unwrap" tokens - cDAI <-> DAI, stETH <-> ETH to name a few that we potentially wanted to support here.

That being said, I would be OK with ripping this out of the the settlement encoder in favour of a single `Unwrap` interaction per settlement since we don't use this feature and it is causing confusion ATM. That being said, what I just mentioned is unrelated to this change though.

nlordell (commented):
Code looks good! Just not a fan of the `debug_assert`.

crates/solver/src/settlement_post_processing/optimize_unwrapping.rs
nlordell: > It might be easier to extract the IO you need into a trait that can be mocked. You can look at other examples in the code where use mockall::automock.

Personally, I'm OK either way. I would only slightly prefer mocking just to be more consistent with the rest of the codebase.

crates/solver/src/settlement_post_processing/optimize_unwrapping.rs
MartinquaXD: I decided to drop the error propagation after all.

crates/solver/src/settlement_post_processing/mod.rs
MartinquaXD: Dropped the assert.

crates/solver/src/settlement_post_processing/optimize_unwrapping.rs
MartinquaXD: I refactored the code to use traits and automock. Although it was possible to test the code with the closures, it made the signature of the optimization function very complex. Also having an automocked trait return different values across multiple invocations is way nicer than make a closure do that.

crates/solver/src/settlement_post_processing/optimize_unwrapping.rs
MartinquaXD: Done.

crates/solver/src/settlement/settlement_encoder.rs
nlordell: > maybe open another PR refactoring the unwraps properly

Of course! This is what I meant with:
> what I just mentioned is unrelated to this change though

crates/solver/src/settlement_post_processing/optimize_unwrapping.rs
```diff
@@ -0,0 +1,201 @@
+use super::SettlementSimulating;
+use crate::settlement::Settlement;
+use crate::solver::http_solver::buffers::BufferRetrieving;
```
nlordell, line 3: Nice!

nlordell (approved):
LGTM

Might be cool to do a manual test on Rinkeby and actually submit a settlement to see it use buffers (might require sending some ETH to the settlement contract though).

PR #1509, opened 2021-12-30
submit settlement to multiple targets at the same time

sunce86:
This PR implements the task 3 from: https://github.com/gnosis/gp-v2-services/issues/1407

Up until now, we had the option to submit settlements either to public mempool, Flashbots or Eden.
Now, we can submit settlement to multiple targets at the same time.

`transaction_strategy` is now a vector. If it contains `DryRun` option or if it contains several options where `DryRun` is one of them, then `transaction_strategy` is resolved to `DryRun`. In other cases it is resolved to other strategies or combination of other strategies.

Note that, since we support different `additional_tip` for eden and flashbots, submitted transactions to respective networks will have different hash (because `additional_tip` affects the `gas_price` of the submitted transaction). This is not an issue, they cant be both mined since they have the same nonce.

### Test Plan
Manual test with Flashbots and Eden options passed. The plan is to extensively test and monitor the changes on staging.

### DIFF
--- a/crates/solver/src/main.rs
+++ b/crates/solver/src/main.rs
@@ -191,8 +191,17 @@ struct Arguments {
     zeroex_slippage_bps: u32,
 
     /// How to to submit settlement transactions.
-    #[structopt(long, env, default_value = "PublicMempool")]
-    transaction_strategy: TransactionStrategyArg,
+    /// Expected to contain either:
+    /// 1. One value equal to TransactionStrategyArg::DryRun or
+    /// 2. One or more values equal to any combination of enum variants except TransactionStrategyArg::DryRun
+    #[structopt(
+        long,
+        env,
+        default_value = "Flashbots,Eden",
+        possible_values = &TransactionStrategyArg::variants(),
+        case_insensitive = true,
+        use_delimiter = true)]
+    transaction_strategy: Vec<TransactionStrategyArg>,
 
     /// Additional tip in gwei that we are willing to give to eden above regular gas price estimation
     #[structopt(
@@ -517,15 +526,29 @@ async fn main() {
             .await
             .map_err(|err| tracing::error!("Couldn't fetch market makable token list: {}", err))
             .ok();
-    let solution_submitter = SolutionSubmitter {
-        web3: web3.clone(),
-        contract: settlement_contract.clone(),
-        gas_price_estimator: gas_price_estimator.clone(),
-        target_confirm_time: args.target_confirm_time,
-        max_confirm_time: args.max_submission_seconds,
-        retry_interval: args.submission_retry_interval_seconds,
-        gas_price_cap: args.gas_price_cap,
-        transaction_strategy: match args.transaction_strategy {
+    let submission_nodes = args
+        .transaction_submission_nodes
+        .into_iter()
+        .enumerate()
+        .map(|(index, url)| {
+            let transport = create_instrumented_transport(
+                HttpTransport::new(client.clone(), url, index.to_string()),
+                metrics.clone(),
+            );
+            web3::Web3::new(transport)
+        })
+        .collect::<Vec<_>>();
+    for node in &submission_nodes {
+        let node_network_id = node.net().version().await.unwrap();
+        assert_eq!(
+            node_network_id, network_id,
+            "network id of custom node doesn't match main node"
+        );
+    }
+    let transaction_strategy = args
+        .transaction_strategy
+        .iter()
+        .map(|strategy| match strategy {
             TransactionStrategyArg::PublicMempool => {
                 TransactionStrategy::CustomNodes(StrategyArgs {
                     submit_api: Box::new(CustomNodesApi::new(vec![web3.clone()])),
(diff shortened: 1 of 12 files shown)


crates/solver/src/settlement_submission/submitter/eden_api.rs
```diff
@@ -24,6 +24,7 @@ impl TransactionSubmitting for EdenApi {
         &self,
         tx: TransactionBuilder<DynTransport>,
     ) -> Result<TransactionHandle, SubmitApiError> {
+        tracing::info!("sending transaction to Eden network...");
```
fleupold, line 27: Will we be able to match this log statement with the one that is posting the tx hash (to know the Eden vs. flashbots tx hash)? Otherwise maybe just pass a "Tag" into `common::submit_raw_transaction` to make sure it gets printed in the same statement.

crates/solver/src/settlement_submission.rs
```diff
@@ -55,68 +67,66 @@ impl SolutionSubmitter {
...
+                            strategy_args.submit_api.as_ref(),
+                            &gas_price_estimator,
+                        )?;
+                        submitter.submit(settlement.clone(), params).await
+                    }
+                    .boxed()
+                })
+                .collect::<Vec<_>>();
+
+            loop {
```
fleupold, line 119: This logic is quite involved (what happens if there is a dry run in the list, what happens if flashbots fails but eden succeeds, etc) and would really benefit from some unit tests imo.

fleupold (approved):
Looks good for testing in staging. 

However, I think this logic is quite important (and will become even more complex as we add truly smart submission strategies) and could therefore benefit from being architected in a way that allows for easy unit tests of the actual strategy logic.

crates/solver/src/settlement_submission/submitter/common.rs
```diff
@@ -43,7 +43,11 @@ pub async fn submit_raw_transaction(
         .map_err(|err| SubmitApiError::Other(err.into()))?;
 
     let handle = parse_json_rpc_response::<H256>(&body)?;
-    tracing::debug!("transaction handle: {}", handle);
+    tracing::info!(
```
MartinquaXD, line 46: Is there a reason to change `tracing::debug!` to `tracing:info!`?
Could this have unexpected side effects, like flooding logs are such?

crates/solver/src/main.rs
```diff
@@ -517,15 +526,29 @@ async fn main() {
...
+        .map(|(index, url)| {
+            let transport = create_instrumented_transport(
+                HttpTransport::new(client.clone(), url, index.to_string()),
+                metrics.clone(),
+            );
+            web3::Web3::new(transport)
+        })
+        .collect::<Vec<_>>();
+    for node in &submission_nodes {
+        let node_network_id = node.net().version().await.unwrap();
```
MartinquaXD, line 542: nit: How likely is this `unwrap` to fail? Seems odd to have an `unwrap` right before an assertion. Maybe they could be combined somehow. Ultimately it probably wouldn't make huge difference in case of a failed `unwrap`.

MartinquaXD (approved):
Looks good to me.

crates/solver/src/main.rs
```diff
@@ -517,15 +526,29 @@ async fn main() {
...
+        assert_eq!(
+            node_network_id, network_id,
+            "network id of custom node doesn't match main node"
+        );
+    }
+    let transaction_strategy = args
```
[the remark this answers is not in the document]
sunce86: Fill included in the following PR

crates/solver/src/settlement_submission/submitter/eden_api.rs
sunce86: Fix included in the following PR

crates/solver/src/settlement_submission.rs
sunce86: DryRun does actually messes everything up, as a completely different strategy fails to fit into the flow for other strategies.
For now I went with this, but I'm open for suggestions how to improve the arguments:
`transaction_strategy is now a vector. If it contains DryRun option or if it contains several options where DryRun is one of them, then transaction_strategy is resolved to DryRun. In other cases it is resolved to other strategies or combination of other strategies.`
Regarding unit tests, this is one of those components that is hard to cover with UTs, because we have infinite loop with deadline, also I believe it's not easy to simulate the nonce change. Maybe some e2e test would be more appropriate here?

crates/solver/src/settlement_submission/submitter/common.rs
sunce86: Seems like a replacement change but it is more a moved log from here: 
https://github.com/gnosis/gp-v2-services/blob/f07fc12f1d08bb05b40c3123f0ae4915fbef94a6/crates/solver/src/settlement_submission/submitter.rs#L316
This log contained hash print a few PRs behind, so just wanted to stay consistent with that.

crates/solver/src/main.rs
sunce86: Good point. This code is not new (I just moved it) but I guess the reasoning behind the unwrap is that it happens in main at the app startup so panicking is not that big of a deal and also enforcing the user to enter valid arguments.

crates/solver/src/settlement_submission/submitter/common.rs
MartinquaXD: Ahh, I see. 👌

crates/solver/src/settlement_submission.rs
fleupold: > I'm open for suggestions how to improve

I probably don't see all the difficulties, but I personally would maybe try to make the `SolutionSubmitter` component abstract (having a trait and multiple implementations). Then you can have "simple strategy" submitters (e.g. DryRun, Flashbots only, Eden only, etc) and more complex, composed strategies (e.g. using multiple sub strategies in a vector, using sub strategy A if condition C is met, strategy B otherwise).

Such an abstraction may allow us to implement arbitrary logic (maybe one day we can get a MEV searcher to actually write us a really smart strategy).

Maybe this is a big enough "design" exercise to brainstorm with the rest of the backend team.

> Regarding unit tests, this is one of those components that is hard to cover with UTs

This is usually just as sign that the code is currently not architected in a way that allows for unit tests. Specifically because there is an endless loop, I'm concerned about not covering this with a test. If we cannot simulate a nonce change, we should make the future that gets created for submission mockable so that in the unit test we can simulate it coming back with `Ok` and `Error` (without actually requiring an actual RPC conneciton).

I don't know if an e2e test can work here, since Flashbots/Eden are hard to simulate e2e unless we try to run it on mainnet (which would be super costly on CI).

PR #1525, opened 2022-01-02
add possibility to disable strategies

sunce86:
This PR implements task 4 from: https://github.com/gnosis/gp-v2-services/issues/1407

The idea is following:
1. All transaction strategies that are considered are received on program start up (`transaction_strategy` argument)
2. For each submission loop (one specific settlement) disable some of the strategies before entering the submission loop (for example, we know if settlement is MEV extractable before starting the submission loop)
2. For each loop inside submission loop, additionally disable/reenable some of the strategies based on some dynamically changed parameters (for example, if effective gas price becomes > 500gwei, disable Eden network temporarily)

### DIFF
--- a/crates/solver/src/metrics.rs
+++ b/crates/solver/src/metrics.rs
@@ -45,6 +45,8 @@ pub enum SettlementSubmissionOutcome {
     Revert,
     /// Submission timed-out while waiting for the transaction to get mined.
     Timeout,
+    /// Submission disabled
+    Disabled,
     /// A transaction failed to be submitted or, in the case of private network
     /// submission, the blockchain state changed and the transaction is no
     /// longer valid.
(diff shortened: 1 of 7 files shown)


nlordell:
Interesting approach. So, the way I understand it is the end goal to just configure a list of transaction strategies to use, and then have the shared component decide which one to use in the end?

We would have to be careful to not accidentally disable all strategies.

Maybe another approach is to make the strategies composable and configurable like:
```
--transaction-strategy 'MevExtractable(Parallel(Eden,Flashbots),CustomNodes)'
```

Would send "mev-extractable" settlements to `Parallel(Eden,Flashbots)` which would submit the settlement in parallel to `Eden` and `Flashbots`, and for non-mev-extractable settlements would just use `CustomNode` strategy? Maybe my suggestion is overkill though 🙈.

sunce86:
Closed in favor of https://github.com/gnosis/gp-v2-services/pull/1612

PR #1527, opened 2022-01-04
Add buffered transport with automatic batching

nlordell:
This PR introduces a new `Buffered` transport that wraps a `BatchTransport` and tries to automatically group RPC calls into batches.

The motivation behind this change is to simplify how we build batched calls. The idea is that we will now be able to more intuitively:
```rust
let tokens: Vec<ERC20>;
let user: H160;

let futures = tokens.iter().map(|token| token.balance_of(user).call());
let balances = future::try_join_all(futures).await?;
```

Instead of having to manually build batches with:
```rust
let tokens: Vec<ERC20>;
let user: H160;

let mut batch = CallBatch::new(self.web3.transport());
let futures = tokens
    .iter()
    .map(|token| token.balance_of(user).batch_call(&mut batch))
    .collect::<Vec<_>>();
batch.execute_all(MAX_BATCH_SIZE).await;
let balances = future::try_join_all(futures).await?;
```

Furthermore, it will help [clean up some of our APIs where we pass `CallBatch`-es around](https://github.com/gnosis/gp-v2-services/blob/397043bb1d4bf2cdf3a59dc94c7ecae2309097c8/crates/shared/src/sources/balancer_v2/pools.rs#L111-L120) in order allow batching across multiple calls.
 
The way this implementation works is to setup a multi-producer single-consumer channel that the `Buffered` transport implementation uses to push new JSON RPC requests to a background worker task. The background worker task then splits the incoming JSON RPC requests into batches and uses the inner `BatchTransport` implementation to execute them in one round trip. The batching logic is very similar to [`StreamExt::ready_chunks`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.ready_chunks) but allows for an additional configured "delay" to wait for additional requests to make single request batches less likely.

### Test Plan

Added some unit tests to check automatic batching logic.

### DIFF
--- a/crates/shared/src/transport/buffered.rs
+++ b/crates/shared/src/transport/buffered.rs
@@ -160,13 +160,10 @@ where
 
 /// Batches a stream into chunks.
 ///
-/// This is very similar to `futures::future::FutureExt::ready_chunks` with the
-/// difference that it allows configuring:
-/// - Maximum concurrency for work done on the chunks (allowing things like only
-///   processing one chunk at a time, and have the subsequent chunk already
-///   start accumulating incoming items)
-/// - Minimum delay for a batch, so waiting for a small amount of time to allow
-///   the stream to produce additional items, thus increasing batch sizes.
+/// This is very similar to `futures::stream::StreamExt::ready_chunks` with the
+/// difference that it allows configuring a minimum delay for a batch, so
+/// waiting for a small amount of time to allow the stream to produce additional
+/// items, thus decreasing the chance of batches of size 1.
 fn batched_for_each<T, St, F, Fut>(
     config: Configuration,
     items: St,
--- a/crates/shared/src/transport/buffered.rs
+++ b/crates/shared/src/transport/buffered.rs
@@ -76,21 +76,22 @@ where
         config: Configuration,
         calls: mpsc::UnboundedReceiver<CallContext>,
     ) -> JoinHandle<()> {
-        tokio::task::spawn(batched_for_each(config, calls, move |mut batch| {
+        tokio::task::spawn(batched_for_each(config, calls, move |batch| {
             let inner = inner.clone();
             async move {
-                match batch.len() {
+                let (mut requests, mut senders): (Vec<_>, Vec<_>) = batch
+                    .into_iter()
+                    .filter(|(_, _, sender)| !sender.is_canceled())
+                    .map(|(id, request, sender)| ((id, request), sender))
+                    .unzip();
+                match requests.len() {
                     0 => (),
                     1 => {
-                        let (id, request, sender) = batch.remove(0);
+                        let ((id, request), sender) = (requests.remove(0), senders.remove(0));
                         let result = inner.send(id, request).await;
                         let _ = sender.send(result);
                     }
                     n => {
-                        let (requests, senders): (Vec<_>, Vec<_>) = batch
-                            .into_iter()
-                            .map(|(id, request, sender)| ((id, request), sender))
-                            .unzip();
                         let results = inner
                             .send_batch(requests)
                             .await
@@ -128,8 +129,13 @@ where
     }
 
     fn send(&self, id: RequestId, request: Call) -> Self::Out {
-        let response = self.queue_call(id, request);
-        async move { response.await.expect("worker task unexpectedly dropped") }.boxed()
+        let this = self.clone();
+
+        async move {
+            let response = this.queue_call(id, request);
+            response.await.expect("worker task unexpectedly dropped")
+        }
+        .boxed()
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/shared/src/transport/buffered.rs
```diff
@@ -0,0 +1,304 @@
...
+                    1 => {
+                        let (id, request, sender) = batch.remove(0);
+                        let result = inner.send(id, request).await;
+                        let _ = sender.send(result);
+                    }
+                    n => {
+                        let (requests, senders): (Vec<_>, Vec<_>) = batch
+                            .into_iter()
+                            .map(|(id, request, sender)| ((id, request), sender))
+                            .unzip();
```
vkgnosis, line 93: Not sure if worth it but we could detect the receiver being dropped through `Sender::is_canceled` and in that case not send a request.

vkgnosis (approved):
Neat. When I read the description I was slightly worried this might be hiding too many implementation details and it would be better to manually create batches. After understanding it more I think it's fine.

There is a non obvious change in behavior: Before if you created a bunch of futures on the transport and did not poll them they wouldn't be executed. Now the background task will execute them anyway. I think that's unproblematic.

crates/shared/src/transport/buffered.rs
nlordell: Good idea.

> Before if you created a bunch of futures on the transport and did not poll them they wouldn't be executed.

That's a good point. It still behaves slightly differently in the case where you don't drop a request future but just don't poll it (before, it would not do anything, now it still sends the request, even if we check if the `oneshot::Receiver` has been dropped).

Another approach would be to only queue the call once it gets polled. We should still see benefits of batching (we can set a default very small delay - like 1ms or something, to make batching slightly more likely when `join!`-ing futures). WDYT?

vkgnosis: Depends on how much more complex it gets. Should be easy to either await a small timeout or even await next poll. If it stays easy then it's probably worth it.

crates/shared/src/transport/buffered.rs
```diff
@@ -0,0 +1,305 @@
...
+                    0 => (),
+                    1 => {
+                        let (id, request, sender) = batch.remove(0);
+                        let result = inner.send(id, request).await;
+                        let _ = sender.send(result);
+                    }
+                    n => {
+                        let (requests, senders): (Vec<_>, Vec<_>) = batch
+                            .into_iter()
+                            .filter(|(_, _, sender)| sender.is_canceled())
```
vkgnosis, line 92: Missing NOT ?

nlordell: :see_no_evil: I was wondering why my tests were failing.

crates/shared/src/transport/buffered.rs
nlordell: It was an easy change: https://github.com/gnosis/gp-v2-services/pull/1527/commits/449c3bddca58d9ab12f8237a94b6b12c81d2737a

The behaviour is now:
- Don't send requests for cancelled futures (i.e. dropped `oneshot::Receiver`)
- Only queue the request if it gets polled

crates/shared/src/transport/buffered.rs
```diff
@@ -0,0 +1,311 @@
...
+    stream::{self, FusedStream, Stream, StreamExt as _},
+};
+use serde_json::Value;
+use std::{future::Future, sync::Arc, time::Duration};
+use tokio::task::JoinHandle;
+
+/// Buffered transport configuration.
+pub struct Configuration {
+    /// The maximum amount of concurrent batches to send to the node.
+    pub max_concurrent_requests: usize,
```
MartinquaXD, line 21: nit: I think `for_each_concurrent` taking an `Into<Option<usize>>` is an unfortunate choice because `Some(0)` (which would mean unlimited parallelism) feels like an invalid state. Making `max_concurrent_requests` a `Into<NonZeroUsize>` to "fix" the API of `for_each_concurrent` seems like too much of a hassle but documenting that passing 0 would result in unlimited parallelism seems reasonable.

MartinquaXD (approved):
I think the code is quite elegant. I had to look up a couple of the "advanced" future helper functions, but it looks like they have been used correctly and the code make sense.
Do you think this could be upstreamed to `web3` or is this use case too niche?

nlordell:
> Do you think this could be upstreamed to web3 or is this use case too niche?

I don't see why not.

crates/shared/src/transport/buffered.rs
nlordell: Good point. I don't like "magic values". I like the idea of an `Option<NonZeroUsize>`.

fleupold (commented):
This is very cool! Is the plan to have a single batch being passed around all our components that do RPC requests so that we can bundle all requests together (not just across the ones where we already know we need to batch for efficiency reasons)?

nlordell:
> Is the plan to have a single batch being passed around all our components that do RPC requests so that we can bundle all requests together (not just across the ones where we already know we need to batch for efficiency reasons)?

Not sure, this is something I was still thinking about. There are two options:
1. Create a batched transport for our global web3 instance (this makes all RPC requests over all components automatically batch).
2. Create a batched transport for components that need it.

The advantage of 1. over 2. is that we always automatically batch requests without thinking about it. The disadvantage however is that we may unintentionally include more critical requests in a large batch, causing the time to complete the request to increase (for example, doing a `eth_sendTransaction` with 99 other `eth_call`s).

nlordell:
Going to merge this for now, even if we don't really make use of it. I think factoring out all the `CallBatch` calls into using this is something that needs to be done over smaller PRs.

PR #1528, opened 2022-01-05
Move balancer contract creation to main

vkgnosis:
Follow up to https://github.com/gnosis/gp-v2-services/pull/1435 . Another PR changed the way the balancer contracts were handled which I didn't update the previous PR with.

The point of the change is still to not hide calls to deployed, and have them be clear in main instead.

### Test Plan
CI


nlordell (approved):
Nice!

PR #1532, opened 2022-01-06
Drop Unnecessary Tokens and Clearing Prices

### LINKED ISSUE #779: Optimize price vector in settlement

MartinquaXD:
Fixes #779 

### Idea
The settlement account is only interested in a token and its price if there is a trade in the settlement trading against it. Note that tokens which are ONLY used for intermediate hops in a multi hop trading route, are irrelevant for the settlement contract.
To reduce gas costs for the call data when calling the settle function on the contract, we can drop the unnecessary token addresses and token prices from the `EncodedSettlement`.

### Implementation
1) Collect all `buy_token` and `sell_token` of trades included in the settlement in a set.
2) Drop all tokens from the `Settlement` which are not in the set.
3) Sort `Settlements'` tokens and update indices within trades which refer to the tokens.
4) Calculate price vector like before but now with the reduced list of tokens.

### Test Plan
Unit test: Created a `Settlement` with some "intermediate" tokens assert that they and their prices are not included in the `EncodedSettlement`.

### DIFF
--- a/crates/solver/src/settlement/settlement_encoder.rs
+++ b/crates/solver/src/settlement/settlement_encoder.rs
@@ -239,7 +239,27 @@ impl SettlementEncoder {
             })
     }
 
-    pub fn finish(self) -> EncodedSettlement {
+    fn drop_unnecessary_tokens_and_prices(&mut self) {
+        let traded_tokens: HashSet<_> = self
+            .trades()
+            .iter()
+            .flat_map(|trade| {
+                [
+                    trade.order.order_creation.buy_token,
+                    trade.order.order_creation.sell_token,
+                ]
+            })
+            .collect();
+
+        self.tokens.retain(|token| traded_tokens.contains(token));
+        self.sort_tokens_and_update_indices();
+        self.clearing_prices
+            .retain(|token, _price| traded_tokens.contains(token));
+    }
+
+    pub fn finish(mut self) -> EncodedSettlement {
+        self.drop_unnecessary_tokens_and_prices();
+
         let clearing_prices = self
             .tokens
             .iter()
@@ -653,5 +673,49 @@ pub mod tests {
             .unwrap();
 
         assert!(encoder0.merge(encoder1).is_err());
+    }
+
+    #[test]
+    fn encoding_strips_unnecessary_tokens_and_prices() {
+        let prices = hashmap! {token(1) => 7.into(), token(2) => 2.into(),
+        token(3) => 9.into(), token(4) => 44.into()};
+
+        let mut encoder = SettlementEncoder::new(prices);
+
+        let order_1_3 = OrderBuilder::default()
+            .with_sell_token(token(1))
+            .with_sell_amount(33.into())
+            .with_buy_token(token(3))
+            .with_buy_amount(11.into())
+            .build();
+        encoder
+            .add_trade(order_1_3, 4.into(), 0.into(), false)
+            .unwrap();
+
+        let weth = dummy_contract!(WETH9, token(2));
+        encoder.add_unwrap(UnwrapWethInteraction {
+            weth,
+            amount: 12.into(),
+        });
+
+        let encoded = encoder.finish();
+
+        // only token 1 and 2 have been included in orders by traders
+        let expected_tokens: Vec<_> = [1, 3].into_iter().map(token).collect();
+        assert_eq!(expected_tokens, encoded.tokens);
+
+        // only the prices for token 1 and 2 remain and they are in the correct order
+        let expected_prices: Vec<_> = [7, 9].into_iter().map(U256::from).collect();
+        assert_eq!(expected_prices, encoded.clearing_prices);
+
+        let encoded_trade = &encoded.trades[0];
+
+        // dropping unnecessary tokens did not change the sell_token_index
+        let updated_sell_token_index = encoded_trade.0;
+        assert_eq!(updated_sell_token_index, 0.into());
+
+        // dropping unnecessary tokens decreased the buy_token_index by one
+        let updated_buy_token_index = encoded_trade.1;
+        assert_eq!(updated_buy_token_index, 1.into());
     }
 }


MartinquaXD:
@fedgiac in your original issue #779 you mentioned that only the prices can be dropped. Since the price vector and the token vector have to be synchronized to each other, I figured token addresses could be dropped as well. I had a quick look through the settlement contract and didn't find anything speaking against that.
Could you confirm that this is indeed the case?
No e2e tests are failing because of it so I think this assumption should be true.

vkgnosis:
> Could you confirm that this is indeed the case?

I think that's true. Like you said tokens and prices have to synchronized so it doesn't make sense to remove only a price. Federico probably meant both.

vkgnosis (approved):
Looks right to me. The only thing I was wondering about is whether we need special logic for BUY_ETH_ADDRESS . Looking at the contract I don't think so but would be nice to have confirmed by someone who knows the contract better.

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -239,7 +239,25 @@ impl SettlementEncoder {
...
+            .iter()
+            .flat_map(|trade| {
+                [
+                    trade.order.order_creation.buy_token,
+                    trade.order.order_creation.sell_token,
+                ]
+            })
+            .collect();
+
+        self.tokens.retain(|token| traded_tokens.contains(token));
```
vkgnosis, line 254: I know this happens in finish anyway, but it would be clearer to also retain on self.clearing_prices here? Considering the function is called "and prices" .

nlordell:
> The only thing I was wondering about is whether we need special logic for BUY_ETH_ADDRESS .

No. This is only used in the context of transfers and is basically:
```
if buyToken == BUY_ETH_ADDRESS {
    transfer(receiver, executedBuyAmount);
} else {
    buyToken.transfer(receiver, executedBuyAmount);
}
```

Like all other tokens, if there is an order with a buy token of `ETH_BUY_TOKEN`, then it needs to be included in the token list.

nlordell:
> Could you confirm that this is indeed the case?

Definitely. `price[i]` is the price of `token[i]`. In theory, you can add extra prices or tokens and still have a valid settlement, but then you are just paying for additional call data that doesn't get used.

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -654,4 +672,38 @@ pub mod tests {
...
+    fn encoding_strips_unnecessary_tokens_and_prices() {
+        let prices = hashmap! {token(1) => 7.into(), token(2) => 2.into(),
+        token(3) => 9.into(), token(4) => 44.into()};
+
+        let mut encoder = SettlementEncoder::new(prices);
+
+        let order_1_2 = OrderBuilder::default()
+            .with_sell_token(token(1))
+            .with_sell_amount(33.into())
+            .with_buy_token(token(2))
```
nlordell, line 686: nit: Note that trades also use indices for the token addresses. Could you maybe make this a `token(3)` so that in theory the token indices for the trade would change by optimizing the token array/price vector?

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -654,4 +672,38 @@ pub mod tests {
...
+        let encoded = encoder.finish();
+
+        // only token 1 and 2 have been included in orders by traders
+        let expected_tokens: Vec<_> = [1, 2].into_iter().map(token).collect();
+        assert_eq!(expected_tokens, encoded.tokens);
+
+        // only the prices for token 1 and 2 remain and they are in the correct order
+        let expected_prices: Vec<_> = [7, 2].into_iter().map(U256::from).collect();
+        assert_eq!(expected_prices, encoded.clearing_prices);
+    }
```
nlordell, line 708: nit: Can we also assert that the trade's token indices are correct here?

nlordell (commented):
LGTM! Just some small change requests around the test for some extra peace of mind 😛

crates/solver/src/settlement/settlement_encoder.rs
MartinquaXD: Done 👍

crates/solver/src/settlement/settlement_encoder.rs
MartinquaXD: Done 👍

crates/solver/src/settlement/settlement_encoder.rs
MartinquaXD: Done 👍

MartinquaXD:
Added assertions about correctness of token indices changed by dropping unnecessary token data.
Also filtered out unnecessary clearing prices to make the name of `drop_unnecessary_tokens_and_prices()` actually correct.

nlordell (approved):
LGTM

bh2smith (approved):

bh2smith:
This is great! Do we save on gas by reducing the call data size? I love saving gas.

MartinquaXD:
@bh2smith I assume token addresses are usually non-zero bytes (86 gas per byte) whereas token prices probably contain mostly zeros (4 gas per byte). With 68 gas per non-zero byte and 4 gas per zero byte that leaves us with:
token address: 32 * 68 = 2176
token price (pessimistic) = 31 * 4 + 68 = 192 (at least 1 non-zero byte otherwise the token would be free)
Totaling ~2368 gas per dropped token information in savings if I'm not mistaken.

bh2smith:
That's great. Maybe a bit far fetched and better asked to the data guy, but do we happen to know how much superfluous data we had been passing into the contract before up until now? Was it one solver in particular or...

MartinquaXD:
Sorry, don't know. I yet have to make a plan how to measure the impact of the change. Maybe comparing `avg(bytes of calldata / number of trades)` before and after the change would make sense?

fleupold:
> I yet have to make a plan how to measure the impact of the change

This might not be the best idea for every small change, but one way is to create a metric that counts whenever this code path does an improvement (e.g. in the other task you did you could simply increment a counter for every unwrap saved). Piping metrics into a new component used to be a bit of boilerplate overhead, but I think @taminomara refactoring made it quite easy to create a new metric also just for "impact" measure.

nlordell:
> This might not be the best idea for every small change, but one way is to create a metric that counts whenever this code path does an improvement

I think for this change a metric might be overkill, in particular I think it would be easier to  just use Dune to:
- Find all settlements with trades that bought ETH **without** a `WETH9::withdraw` interaction since the change is released
- We know how much gas an unwrap is (we have historical data that can approximate this)
- Compute the sum `weth_unwrap_gas * settlement_gas_price * eth_price` to get a USD measurement of the impact.

Maybe this is something @bh2smith can help with.

MartinquaXD:
@fleupold good suggestion. I will have a look at the metrics API for future PRs.

PR #1536, opened 2022-01-08
Adjust Swagger URLs to new `cow.fi` Domain

nlordell:
This PR adjusts the Swagger URLs to point to our new `cow.fi` domain! 🎉 🐮

I also decided to order the YAML pairs as `descritpion` -> `url` as I found it slightly more intuitive.

### Test Plan

Check it out! https://barn.api.cow.fi/docs/

image

### DIFF
--- a/crates/orderbook/openapi.yml
+++ b/crates/orderbook/openapi.yml
@@ -3,20 +3,20 @@ info:
   version: 0.0.1
   title: Order Book API
 servers:
-  - url: https://protocol-mainnet.dev.gnosisdev.com
-    description: Mainnet (Staging)
-  - url: https://protocol-mainnet.gnosis.io
-    description: Mainnet (Prod)
-  - url: https://protocol-rinkeby.dev.gnosisdev.com
-    description: Rinkeby (Staging)
-  - url: https://protocol-rinkeby.gnosis.io
-    description: Rinkeby (Prod)
-  - url: https://protocol-xdai.dev.gnosisdev.com
-    description: xDai (Staging)
-  - url: https://protocol-xdai.gnosis.io
-    description: xDai (Prod)
-  - url: http://localhost:8080
-    description: Local
+  - description: Mainnet (Staging)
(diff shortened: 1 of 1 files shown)


crates/orderbook/openapi.yml
```diff
@@ -3,20 +3,20 @@ info:
...
+  - description: Mainnet (Prod)
+    url: https://api.cow.fi/mainnet
+  - description: Rinkeby (Staging)
+    url: https://barn.api.cow.fi/rinkeby
+  - description: Rinkeby (Prod)
+    url: https://api.cow.fi/rinkeby
+  - description: xDai (Staging)
+    url: https://barn.api.cow.fi/xdai
+  - description: xDai (Prod)
+    url: https://api.cow.fi/xdai
```
MartinquaXD, line 17: Shouldn't those refer to Gnosis chain?

nlordell: Changed the description to mention `Gnosis Chain`. The URLs still use `xdai` as we haven't moved them over yet.

MartinquaXD: Might be worth also updating the screen shot in the PR description.

nlordell: 🦅 👀

MartinquaXD (approved):
Looks good to me.
One thing which surprises me is that the gnosis chain URLs still refer to xDai. Is there a discussion where one can read up about that decision? Could be confusing to only change the name but not add any new URLs with the new name. But this is besides the point of this PR.

nlordell:
> Is there a discussion where one can read up about that decision?

Most of the discussion is happening in Slack. We've marked the `xdai` API endpoints as:
> Maybe later: Should we do at some point the rename to Gnosis Chain??

The issue is that the FE and our ingress configuration still use the `/xdai` path, so we need to change those before we can update the OpenAPI spec to use `/gnosischain` or `/gno` or `/gc` or whatever we settle on.

Issue #1540, opened 2022-01-10
Remove code to filter out pools with 0 amounts for http solver


vkgnosis:
This should be done once the solver version is released that internally filters them instead of erroring. Applies to quasimodo price estimator in orderbook crate and http solver in solver crate.

nlordell:
Specifically here:

https://github.com/gnosis/gp-v2-services/blob/a7915d1529cbebf0a85a2396fc8f6e098c69f65f/crates/solver/src/solver/http_solver.rs#L342

And here:

https://github.com/gnosis/gp-v2-services/blob/3eef426abf33b4ead9c3a9c88275e2d5d061af37/crates/shared/src/http_solver_api/model.rs#L48

PR #1543, opened 2022-01-10
Use balancer liquidity in quasimodo price estimator

### LINKED ISSUE #1412: Support non-uniswap pools for quasimodo price estimator
Right now we only consider uniswap v2 pools for quasimodo price estimator. This happened because we use different representations for liquidity in the `orderbook` and `solver` crates, so supporting other pools is somewhat difficult. Once pool representations are unified, we can support all of them. Alternatively, we can try to add support for paraswap before that happens.

vkgnosis:
Fixes https://github.com/gnosis/gp-v2-services/issues/1412

Moved some code around which happens in individual commits which might be easier to review.

As a follow up to this we can get rid of the 0 pool amount filter as part of https://github.com/gnosis/gp-v2-services/issues/1540 .
And I would like to refactor parts of the balancer and liquidity types so more code can be shared between the http solver invocations in solver crate and shared crate.

### Test Plan
adapted the test in the quasimodo estimator module and ran it against staging http solver


crates/shared/src/sources/balancer_v2/pools/common.rs
```diff
@@ -273,6 +273,26 @@ pub struct TokenState {
     pub scaling_exponent: u8,
 }
 
+/// Compute the scaling rate from a Balancer pool's scaling exponent.
```
vkgnosis, line 276: Moved here from solver crate file

crates/shared/src/http_solver/gas_model.rs
```diff
@@ -0,0 +1,29 @@
+use super::model::CostModel;
```
vkgnosis, line 1: Moved here from solver crate file

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -86,11 +75,17 @@ impl QuasimodoPriceEstimator {
...
+        let gas_model = GasModel {
+            native_token: self.native_token,
+            gas_price: gas_price.to_f64_lossy(),
+        };
 
-        let amms = self
+        let balancer_pools = self.balancer_pools(pairs.clone(), &gas_model).await?;
+        let mut amms: BTreeMap<usize, AmmModel> = self
             .pools
             .fetch(pairs, Block::Recent)
```
nlordell, line 86: Maybe for symmetry we can have a `uniswaplike_pools()` with this original fetching code, then all the pools becomes `self.uniswaplike_pools().into_iter().chain(self.balancer_pools()).enumerate().collect()`.

crates/shared/src/price_estimation/quasimodo.rs
```diff
@@ -103,24 +98,60 @@ impl QuasimodoPriceEstimator {
...
             })
+            .chain(balancer_pools)
             .enumerate()
             .collect();
 
+        // Solver cannot handle 0 reserves so filter these pools out until that is fixed.
+        amms.retain(|_, v| match &v.parameters {
+            AmmParameters::ConstantProduct(p) => p.reserves.values().all(|r| !r.is_zero()),
+            AmmParameters::WeightedProduct(p) => p.reserves.values().all(|r| !r.balance.is_zero()),
+            AmmParameters::Stable(p) => p.reserves.values().all(|r| !r.is_zero()),
```
nlordell, line 112: For some reason I thought there was a `has_sufficient_reserves()` method for this already.

nlordell (commented):
LGTM, just one place where it looks like we could use `has_sufficient_reserves` instead of re-implementing it.

crates/shared/src/price_estimation/quasimodo.rs
vkgnosis: I did it this way to make the diff smaller. But agree for the future separate methods is better. Changed.

crates/shared/src/price_estimation/quasimodo.rs
vkgnosis: Changed to use it.

nlordell (approved):
LGTM!

PR #1544, opened 2022-01-10
Trade liquidity orders at limit price

josojo:
This PR introduces the logic for trading liquidity orders at order limit prices. In order to do so, I am modifying the settlement struct to also store non-uniform clearing price information. These prices are stored by using the sellPrice from the uniform clearing prices, and a separately stored buyPrice. This pair of (sellPrice, buyPrice) will then be similar to the limit order price of the liquidity order modulo rounding.
This new information gets encoded in the following way for the smart contract: I am replacing the `[uniform_clearing_price_vec]` with `[[uniform_clearing_price_vec], [buy_tokens_of_liquidity_orders]]`
 and then all I need to do is to give the liquidity orders an index referring to the second part of the price vector.

### Test Plan
[This](https://dashboard.tenderly.co/gp-v2/staging/tx/mainnet/0x6756c294eb84c899247f2ec64d6eee73e7aaf50d6cb49ba9bab636f450240f51) is the settlement where Martin(0xe63a) only needs to send 11688 stable tokens for his buy order to our settlement contract, as he gets the uniform clearing price.
With logic from this PR, we generate the [following settlement](https://dashboard.tenderly.co/gp-v2/staging/simulator/428383b2-1e53-4cd5-b40b-24efe4698ee0), Martin has to pay 11725. Observe the correctness of the solution.

### Other comments
The PR looks very long, but most of the code is associated with testing. I made an extensive e2e test - taking the quasimodo solution and building the tender simulation from it - in order to see how the new code behaves in the smart contract. If I should remove that test, as it is quite unconventional and requires the settlement module to be public, let me know.


crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -102,27 +115,75 @@ impl SettlementEncoder {
         let sell_token_index = self
             .token_index(order.order_creation.sell_token)
             .expect("missing sell token with price");
+        let (trade, execution) = match is_liquidity_order {
```
marcovc, line 118: Rust question: is `match {true, false}` more idiomatic than an `if .. else` block?

vkgnosis: I wouldn't say so. There is no big difference so it can be personal preference.

nlordell:
For me it depends on the context:
```
let x = if some_thing { 0 } else { 1 };
```

Then with `if/else` you get it in 1 line (kind of like the ternary
operator).

Otherwise with:
```
let x = match some_thing {
    true => some_thing_that_is_longer(),
    false => some_other_thing_that_is_longer(),
};
```

It's in 4 lines. instead of 5 lines for `if/else`.

And lastly:
```
if some_thing {
    stuff();
    over();
    multiple();
    lines();
} else {
    stuff();
    over();
    multiple();
    lines();
}
```

I prefer that over `match` because the blocks are 1 indentation level less
so reduces right-ward drift.

nlordell:
@marcovc, regarding your question, for me it depends on the context:

```rust
let x = if some_thing { 0 } else { 1 };
```

Then with `if/else` you get it in 1 line (kind of like the ternary operator).

Otherwise with:
```
let x = match some_thing {
    true => some_thing_that_is_longer(),
    false => some_other_thing_that_is_longer(),
};
```

It's in 4 lines. instead of 5 lines for `if/else` and more closely ties the `true/false` value to the outcome.

And lastly:
```
if some_thing {
    stuff();
    over();
    multiple();
    lines();
} else {
    stuff();
    over();
    multiple();
    lines();
}
```

I prefer that over `match` because the blocks are 1 indentation level less so reduces right-ward drift.

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -260,7 +334,15 @@ impl SettlementEncoder {
...
-        let clearing_prices = self
+        let (mut additional_tokens, mut additional_prices): (Vec<H160>, Vec<U256>) = self
+            .non_uniform_clearing_prices
+            .iter()
+            .map(|x| (x.token, x.price))
+            .unzip();
+        let uniform_clearing_price_vec_length = self.tokens.len();
+        let mut tokens = self.tokens.clone();
+        tokens.append(&mut additional_tokens);
+        let mut clearing_prices: Vec<U256> = self
```
marcovc, line 345: It is not obvious to me how the call in line 351 won't fail for tokens that are not in `self.clearing_prices` but are in `tokens` due to line 344. Could you please explain?

crates/solver/src/settlement.rs
```diff
@@ -12,11 +13,23 @@ use primitive_types::{H160, U256};
 use shared::conversions::U256Ext as _;
 use std::collections::HashMap;
 
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum TokenIndex {
+    UniformClearingPriceIndex(usize),
+    CustomBuyPriceIndex(usize),
+}
```
vkgnosis, line 20: Probably too much for this PR but I feel like settlement.rs and settlement_encoder.rs are in need of a refactor. Having to special case it like this feels weird. Maybe we could keep track of liquidity orders separately so they can use their own struct, vector, prices, etc.

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -102,27 +115,75 @@ impl SettlementEncoder {
...
+            true => {
+                // Liquidity orders are settled at their limit price. We set:
+                // buy_price =  sell_price * order.sellAmount / order.buyAmount, where sell_price is given from uniform clearing prices
+
+                // Rounding error checks:
+                // Following limit price constraint is checked in the smart contract:
+                // order.sellAmount.mul(sellPrice) >= order.buyAmount.mul(buyPrice),
+                // For each order, we get
+                // order.sellAmount.mul(sellPrice)  >= order.buyAmount.mul(buyPrice)
+                // <=> order.sellAmount.mul(sellPrice)  >= order.buyAmount.mul(sell_price * order.sellAmount / order.buyAmount)
```
vkgnosis, line 152: So on the right side buyAmount is cancelled out and we're left with sellAmount * sellPrice >= sellAmount * sellPrice ?
And since division buy buyAmount happens first before it is multiplied out any rounding error goes down?
If that is right we should spell it out the comment here to make it clear.

marcovc (approved):
Thanks for this PR. I am not 100% sure about everything but it looks promising!

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -157,7 +218,9 @@ impl SettlementEncoder {
                 // have the same price (i.e. are equivalent).
                 return Ok(());
             }
-            (None, None) => bail!("tokens not part of solution for equivalency"),
+            // If token_b is only the buy token of a liquidity order,
+            // the price does not need to be in the uniform clearing prices
+            (None, None) => return Ok(()),
```
vkgnosis, line 223: If what the comment says is true then why are we adding a token equivalency? Shouldn't it still error? Maybe the correct way to handle this is to not call add_token_equivalency for liquidity orders because they shouldn't be buying the native token?

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -317,8 +399,21 @@ impl SettlementEncoder {
...
+                    price: big_rational_to_u256(
+                        &(token_price_pair.price.to_big_rational() * &scaling_factor),
+                    )
+                    .context("Invalid price scaling factor")?,
+                })
+            })
+            .collect::<Result<Vec<CustomBuyPrice>>>()?;
 
-        for other_trade in other.trades.iter() {
+        for other_trade in other.trades.iter().clone() {
```
vkgnosis, line 416: This clone does nothing

crates/solver/src/settlement/settlement_encoder.rs
marcovc: Could it be that you're counting that every token traded by a liquidity order has an entry in the price vector? (I don't believe that must be the case)

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -102,27 +115,75 @@ impl SettlementEncoder {
...
+                let buy_price = self
+                    .clearing_prices
+                    .get(&order.order_creation.sell_token)
+                    .context("settlement missing buy token")?
+                    .checked_mul(order.order_creation.sell_amount)
+                    .context("buy_price calculation failed")?
+                    .checked_div(order.order_creation.buy_amount)
+                    .context("buy_price calculation failed")?;
+
+                let non_uniform_clearing_prices = CustomBuyPrice {
```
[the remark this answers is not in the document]
josojo: ? It is named: `non_uniform_clearing_price`? What is your suggestion? 

I see that the non_uniform_clearing_price is not optimal, as `CustomBuyPrice` only represents halve of the prices to build the prices for liquidity orders. But I wanted to indicate that there are two different variables: normal clearing prices and then the additional non_uniform clearing prices.

Maybe we can name it: liquidity_order_prices

crates/solver/src/settlement/settlement_encoder.rs
josojo: Good catch!
I moved the       `  tokens.append(&mut additional_tokens);` further down, then only the tokens for which we really have clearing prices are in the tokens vect.

crates/solver/src/settlement.rs
josojo: Yeah..., agree.

What do you think about the following approach:
```

enum Trade{
Liquidity_Trade(TradeInner),
NormalTrade(TradeInner),
}

#[derive(Debug, Clone)]
pub struct SettlementEncoder {
    tokens: Vec<H160>,
    clearing_prices: HashMap<H160, U256>,
    trades: Vec<Trade>,
    execution_plan: Vec<Arc<dyn Interaction>>,
    unwraps: Vec<UnwrapWethInteraction>,
}

```

and then build the final clearing prices for liquidity orders only at the very end within the `finish` function.

That would be simplifying it, as then we don't need to store the non_uniform_clearing prices  in the settlementEncoder

crates/solver/src/settlement/settlement_encoder.rs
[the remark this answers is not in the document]
MartinquaXD: Had to double check as well but the new `CustomBuyPrice` is currently named `non_uniform_clearing_prices` (plural) not `non_uniform_clearing_price` (singular).

crates/solver/src/settlement.rs
vkgnosis: I could see this working. Although we have some code that currently iterates over all orders and does something with them. That would be harder. One alternative I meant was to not have enum. So there would be
Vec<NormalTrade>
Vec<LiquidityTrade>
and separate functions to get the from the encoder. This is hard to say without trying it out though.

nlordell:
Regarding the test plan, Is the settlement contract keeping the difference?

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -102,27 +115,77 @@ impl SettlementEncoder {
         let sell_token_index = self
             .token_index(order.order_creation.sell_token)
             .expect("missing sell token with price");
+        let (trade, execution) = match is_liquidity_order {
```
MartinquaXD, line 118: Since most of the function is different based on whether you add a liquidity order, could it be nicer to have 2 separate functions instead?

crates/solver/src/settlement.rs
nlordell: I also think we can refactor the settlement and settlement encoder to make this use-case fit in as I think the uniform clearing prices were somewhat baked into the assumptions of this module.

Maybe separate `Vec`s makes it easier here? I'm also not sure.

crates/solver/src/settlement/settlement_encoder.rs
marcovc: Isn't this deeper than that? I understand that you're counting on one of the prices to be in the clearing price vector, but that doesn't necessary needs to be the case, right?

MartinquaXD (commented):
I find it quite hard to understand how `SettlementEncoder::non_uniform_clearing_prices` are supposed to be used and when and how token indices referencing them have to be updated.
Would it also be possible to make `SettlementEncoder::clearing_prices` store a type which knows the values of both price types?
```rust
struct TokenPrice {
    uniform_clearing_price: U256,
    non_uniform_clearing_price: Option<U256>,
}

struct SettlementEncoder {
    clearing_prices: HashMap<H160, TokenPrice>,
    //...
}
```
I believe in that case managing the indices can stay as it was before the PR and because trades now know if they are a liquidity order, deciding which price to use should be easy.

crates/solver/src/settlement.rs
josojo:
```
Vec<NormalTrade>
Vec<LiquidityTrade>
```
I went for this suggestion, just with a little bit longer names:
```
Vec<OrderTrade>
Vec<LiquidityOrderTrade>
```
It seems to work nice.

I was also trying out another approach:
Stripping away the `buy_token_index` and `sell_token_index` in the `struct Trade` and only finding the indexes during the finish() function of the `SettlementEncoder`. This worked nicely, as then the `Trade` struct of  a liquidity order and a normal order do not differ. Then also, the work of maintaining the indices would not be needed during merges etc. 
But it had a major drawback: If the `SettlementEncoder` would not be valid - as some token indices do not exist - it would only fail during the finish() functionality. In the end, I found it cleaner to not have these hidden assumptions and go with your proposal.

crates/solver/src/settlement/settlement_encoder.rs
josojo: > I understand that you're counting on one of the prices to be in the clearing price vector, but that doesn't necessary needs to be the case, right?

I made an explicit comment about this. I think it's reasonable to have this assumption.
I would propose  to force the solvers giving the nominator price - the price in the sell token via the uniform cl earing price - as then the driver has one less degree of freedom: 
If the solvers would not give a price for the sell token, then the driver could choose [a*sell_price]/ [a*buy_price] with an arbitrary `a` as valid prices for the liquidity order. If we assume that the sell_price is always given, then the scaling factor is in the control of the solver.

josojo:
> Regarding the test plan, Is the settlement contract keeping the difference?

 In the end, the solvers can steer where the surplus is going to. 
 In this particular instance, it looks like the settlement contract gets 21129USDT, but does send only 7916+1479+11727 out
 ```
 21129-7916-1479-11727=7
 ```
 and hence pockets the difference.

josojo:
>I find it quite hard to understand how SettlementEncoder::non_uniform_clearing_prices are supposed to be used and when and how token indices referencing them have to be updated.

In the new code, I managed to avoid a new vector non_uniform_clearing_prices, but stored the execution price within the liquidity order trade. I think it's cleaner, but I am happy to hear your proposal on how to improve the readability even more.

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -243,15 +342,25 @@ impl SettlementEncoder {
...
             .iter()
-            .flat_map(|trade| {
-                [
-                    trade.order.order_creation.buy_token,
-                    trade.order.order_creation.sell_token,
+            .flat_map(|order_trade| {
+                vec![
+                    order_trade.trade.order.order_creation.buy_token,
+                    order_trade.trade.order.order_creation.sell_token,
                 ]
```
vkgnosis, line 349: Here and 10 lines below `vec!` isn't needed. Plain array `[...]` is preferred because it doesn't allocate.

vkgnosis (commented):
The implementation makes sense to me now.

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -111,19 +128,84 @@ impl SettlementEncoder {
...
+        // needs to be the sell_token_index present. Hence, we ask
+        // solvers to provide it in their solution. But actually, we could
+        // also set the sell token index in this function and not require the solvers
+        // to provide it.
+        let sell_token_index = self
+            .token_index(order.order_creation.sell_token)
+            .expect("missing sell token with price");
+
+        // Liquidity orders are settled at their limit price. We set:
+        // buy_price =  sell_price * order.sellAmount / order.buyAmount, where sell_price is given from uniform clearing prices
```
MartinquaXD, line 171: nit: extra space after `buy_price = `

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -243,15 +342,25 @@ impl SettlementEncoder {
...
             })
             .collect();
 
-        self.tokens.retain(|token| traded_tokens.contains(token));
+        let liquidity_traded_tokens: HashSet<_> = self
+            .liquidity_order_trades
+            .iter()
+            .flat_map(|liquidity_order_trade| {
+                vec![liquidity_order_trade.trade.order.order_creation.sell_token]
+            })
```
MartinquaXD, line 358: nit: Using `flat_map` for a single result seems surprising. A normal map would achieve the same thing:
```rust
.map(|liquidity_order_trade| {
    liquidity_order_trade.trade.order.order_creation.sell_token
})
```

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -260,7 +369,19 @@ impl SettlementEncoder {
...
+                .iter()
+                .map(|liquidity_order_trade| {
+                    (
+                        liquidity_order_trade.trade.order.order_creation.buy_token,
+                        liquidity_order_trade.buy_token_price,
+                    )
+                })
+                .unzip();
+        let uniform_clearing_price_vec_length = self.tokens.len();
+        let mut tokens = self.tokens.clone();
```
MartinquaXD, line 383: Not too long ago `drop_unnecessary_tokens_and_prices()` was introduced to make sure the settlement data has no unnecessary token or price data to save gas costs.
Judging by the test `settlement_reflects_different_price_for_normal_and_liquidity_order` duplicated tokens can appear in the final settlement data now.
I don't think this PR has to be updated to include this optimization but we should be aware of it and maybe add a comment or open an issue to track that.

MartinquaXD (commented):
Overall it's now easier to understand what's going on. 👍

crates/solver/src/settlement/settlement_encoder.rs
josojo: Thanks for your comments. They are all great.

Regarding this specific one: Yes, duplicated tokens can appear in the final settlement now. They are required to provide the different prices for the same token. 
Previously, we had only 1 price per token. But with this new PR, we want to settle liquidity orders at different prices. Hence, we no longer have 1 price per token, but can several prices. To provide these several prices, we need to provide the token as reference several times.

But for sure, I agree with you that `drop_unnecessary_tokens_and_prices` could double check that 
there are no double entries for the same token - the double entries will be added later. But let's do this in another PR, as even your original implementation does not do that - it only kick's out unnecessary ones.

MartinquaXD: Ahh, I forgot about the 2 different prices for the same token.
Thanks for the explanation. 👍

vkgnosis (approved):

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -111,19 +127,84 @@ impl SettlementEncoder {
...
+        order: Order,
+        executed_amount: U256,
+        scaled_fee_amount: U256,
+    ) -> Result<TradeExecution> {
+        let sell_price = self
+            .clearing_prices
+            .get(&order.order_creation.sell_token)
+            .context("settlement missing sell token")?;
+
+        // Note, for the encoding strategy of liquidity orders, there
```
josojo, line 160: @marcovc Is this okay for you?

The reason why I prefer to have it provided by the solvers is the following:
- If solvers do not provide the sell price for the liquidity orders, then the driver can choose it price freely up to a scalar: E.g. [sell_price, buy_price] or [a*sell_price, a*buy_price] would both be okay. In order to have it more deterministic and in order to give the solvers more control over the submitted solution, I would prefer to require the solvers to provide the sell_token_price for liquidity orders.

marcovc: Currently the json includes the order executed buy and sell amounts, I guess everything can be derived from that, right?

josojo: Yes, we could just use the executed_amounts as sell and buy prices, thought this might be confusing: 
- If the sellTokenPrice is provided then we want to reuse it for liquidity order trades
- If the sellTokenPrice is not provided, then we would use the executed_buy_amount.

Hence, my proposal is to always provide the sellTokenPrice. But yeah, we can also go with  your logic. I don't have a strong opinion

marcovc: I think I might need to read your PR again, but I'm currently very busy. Please go ahead with your suggestion, I'm sure it makes sense and that we can adapt how we express that exchange rate later.

josojo: :) I will wait for you @marcovc. If we merge it as it is and the solver does not provide the sell_token_price, then the solution will not be accepted. Feel free to huddle me, to speed it up. No need to read the whole PR.

josojo: Marco convinced me to use the uniform clearing price vector only for price information of usual user orders. I will refactor this part.

josojo:
Could I get one more review of this commit? 
https://github.com/gnosis/gp-v2-services/pull/1544/commits/fc6c9ff5a9e88f88ee4eefb980afbc7c42260342

Probably, not, the whole PR needs to be checked again.

vkgnosis (approved):

crates/solver/src/liquidity/order_converter.rs
```diff
@@ -76,16 +76,22 @@ impl SettlementHandling<LimitOrder> for OrderSettlementHandler {
     fn encode(&self, executed_amount: U256, encoder: &mut SettlementEncoder) -> Result<()> {
         let is_native_token_buy_order = self.order.order_creation.buy_token == BUY_ETH_ADDRESS;
 
-        if is_native_token_buy_order {
+        if !self.is_liquidity_order && is_native_token_buy_order {
```
nlordell, line 79: Why do we not add token equivalency for liquidity orders? They still need a `0xeee..eee` price if they are buying ETH right?

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -26,8 +29,15 @@ pub struct SettlementEncoder {
...
-    trades: Vec<Trade>,
+    // Order trades are trades of usual user orders. They will be settled using the
+    // uniform clearing prices. Hence, every trade's buy and sell token has an entry
+    // in clearing_prices
+    order_trades: Vec<OrderTrade>,
+    // Liquidity orders will be settled at their limit prices.
+    // In order to represent the limit price, the sell_price is taken from the uniform
+    // clearing price vector and the buy_price is a custom buy_price defined in the
+    // struct LiquidityOrderTrade
+    liquidity_order_trades: Vec<LiquidityOrderTrade>,
```
nlordell, line 40: I'm a fan of this separation!

crates/solver/src/liquidity/order_converter.rs
nlordell: **edit**: I understand why its not now from the code below (when the additional prices are added to the price vector for liquidity orders).

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -121,19 +137,92 @@ impl SettlementEncoder {
...
+                self.tokens.push(sell_token);
+                self.clearing_prices.insert(sell_token, sell_price);
+                (self.tokens.len() - 1, sell_price)
+            }
+        };
+
+        // Liquidity orders are settled at their limit price. We set:
+        // buy_price = sell_price * order.sellAmount / order.buyAmount, where sell_price is given from uniform clearing prices
+
+        // Rounding error checks:
```
nlordell, line 190: One other solution is to always add two new entries and have a duplicated:
```
tokens = [...tokens, sell_token, buy_token];
prices = [...prices, buy_amount, sell_amount];
```

This way you avoid rounding errors at the cost of potentially slightly more calldata (64 bytes of mostly 0s - so around 500 - 700 additional gas).

nlordell (approved):
Changes look good to me. Just one suggestion for simplifying the logic for computing for the additional liquidity order prices.

crates/solver/src/liquidity/order_converter.rs
josojo: Good point. Not obvious, hence I added a comment: https://github.com/gnosis/gp-v2-services/pull/1544/commits/59c699363dbdc808a7c1136a96a23a5d282d82ed

crates/solver/src/settlement/settlement_encoder.rs
josojo: Yes, one would avoid rounding errors and have less complexity in the driver. But since we have the rounding errors for nearly all solutions, I decided for the gas efficiency. Hope this is okay.

nlordell: Yes of course, the suggestion wasn't blocking the approval.

MartinquaXD:
One more suggestion:
Since the sorting issue wasn't caught by any test could you modify `settlement_inserts_sell_price_for_new_liquidity_order_if_price_did_not_exist` such that a second order gets added which would affect the token indices of the first order?

nlordell:
@josojo Note that we don't have `merge when green` support anymore (we had issues with Mergify) and instead use the native Github automerge feature:

image

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -687,6 +687,8 @@ pub mod tests {
         assert!(settlement
             .add_liquidity_order_trade(order01.clone(), 20.into(), 0.into())
             .is_ok());
+        // ensures that the output of add_liquidity_order is sorted
+        assert_eq!(settlement.tokens, vec![token(0), token(1)]);
```
MartinquaXD, line 691: This still leaves room for forgetting to update the indices after adding a new token. The following assertions would have caught that.
```rust
assert_eq!(
    settlement.liquidity_order_trades[0].trade.sell_token_index, // <-- is the sell token index of liquidity order
    0
);
```

josojo: Added in here: 
[50ecc3d](https://github.com/gnosis/gp-v2-services/pull/1544/commits/50ecc3d7230bcd04df1c2120f64f29e5c1cc6e6f)

But I don't think that it is important, as updating and sorting is always done "atomic" in the function sort_and_update_indicies. But yeah, now we are safe. thanks

MartinquaXD (approved):

crates/solver/src/settlement/settlement_encoder.rs
MartinquaXD: I think it's nice to have that extra peace of mind. Thanks for adding that. 👍

nlordell: I like the added assert as it protects us from refactoring errors that may happen in the future (if we decide to change how `sort_and_update_indicies` works for example).

PR #1546, opened 2022-01-11
Making the gas limit for simulations configurable

josojo:
Currently, we are seeing a ton of simulations errors due to a high gas limit and not sufficient funds: 
https://logs.gnosisdev.com/goto/f87ec3e0-72bf-11ec-8530-c3498405903c

I want to provide a high gas limit for xdai, as at peaks we are seeing batches with over 60 orders.

But for mainnet, we never had a solution that consumed more than 6million gas. Hence, I think we can provide smaller values, especially for the shadow-solver.


crates/solver/src/settlement_simulation.rs
```diff
@@ -103,7 +104,7 @@ pub async fn simulate_and_error_with_tenderly_link(
                 // set a gas limit so we don't get failed simulations because of insufficient
                 // solver balance for the default ~15M gas limit. Limit to around the
                 // block gas limit / 2 (as settling bigger tx is unrealistic anyways).
-                .gas(15_000_000.into());
+                .gas(simulation_gas_limit.into());
```
vkgnosis, line 107: While reading your PR I noticed that the comment right above here says that the intended mainnet limit is 7.5m instead of 15m. Does anyone know what is right?
Either way I'd update this comment to remove the last sentence because the value is now passed in.

sunce86: 30M is block gas limit while there is a gas target of 15M which is chased by increasing/decreasing block base fee.
Anyway, I would leave 15M as a default value for this parameter (as in this PR).

sunce86 (approved):
looks good

nlordell (approved):
Nice idea!

vkgnosis (approved):

PR #1547, opened 2022-01-11
Stop filtering out pools with 0 amounts for http solver

### LINKED ISSUE #1540: Remove code to filter out pools with 0 amounts for http solver

vkgnosis:
With a recent change to the http solver already running on staging this
is no longer necessary.

TODO: get confirmation that this is ok for the older non quasimodo solver too
Update: Should be fine on staging. Maybe wait with merging this until until the solvers had a release to be sure we don't break prod.

### Test Plan
ran the ignored quasimodo test with staging solver that used to fail without filtering out pools


nlordell (approved):

MartinquaXD (approved):
LGTM

MartinquaXD:
I guess this should have "Fixes #1540" in the description?

vkgnosis:
Yes, thanks. Also can be merged now.

Issue #1533, opened 2022-01-06
Cache Price Estimation API Requests


nlordell:
We are currently being rate-limited by ParaSwap because we are making around 1000 requests per minute. It looks like we are sending a lot of duplicate requests (especially when computing the native token price for some sell token - since the amounts are always the same and it is likely that a user makes multiple price estimate requests for the same sell token while tweaking amounts).

vkgnosis:
There are three potential places where caching could happen:

- Paraswap
- Backend
- Frontend

We can't influence Paraswap but still it is one perspective to say that they should be caching (and they probably are) so we don't have to do it for them. Still it is work to send network requests so it does make sense to have caching in the backend. It depends on how often the amounts are the same. We probably cannot answer this without actually implementing a cache and metrics.
If these requests you mention as being duplicate originate from the frontend then there could be some caching there too. But not sure from top of my head if that is the case or these are prices we request in backend logic.

nlordell:
Specifically, I was thinking of caching ParaSwap API requests. For example, for two different FE requests:
- quote SELL 100 BAL for GNO
- quote SELL 1000 BAL for GNO

We have a the duplicate ParaSwap API to compute the sell token's ETH sell price:
- quote BUY 1ETH for BAL

In fact, the API request for getting ETH sell prices is much more likely to be a cache hit because only the sell token varies (instead of sell token, buy token and amount for usual price requests).

fleupold:
>  It depends on how often the amounts are the same

No exact science, but I checked https://logs.gnosis.io/goto/c30951d0-6efa-11ec-9c50-a508084c3610 to see that some requests are indeed sent 2-6 times within the matter of a few seconds.

sunce86:
> * quote SELL 100 BAL for GNO
> * quote SELL 1000 BAL for GNO

Are these REALLY the same requests or we just approximate them as being the same?
Asking because I am not sure how Paraswap works internally. These two queries should theoretically make different slippages right?

vkgnosis:
We are not going to cache when the amounts are different. What Nick means is that for both of those requests we will make the same "quote BUY 1ETH for BAL" estimate so that one can be cached.

nlordell:
Closing as this should be solved by #1548 and #1555

PR #1548, opened 2022-01-11
Create price estimator cache

vkgnosis:
Part of https://github.com/gnosis/gp-v2-services/issues/1533 .

This is a generic price estimation cache that we can use to wrap the external price estimation apis.

I initially implemented the cache myself but decided on using a dependency that already implements what we want and feels well made.

In a follow up PR I would change the Paraswap and 0x estimators to use the cache.

### Test Plan
unit test

### DIFF
--- a/crates/shared/src/price_estimation/cached.rs
+++ b/crates/shared/src/price_estimation/cached.rs
@@ -1,15 +1,36 @@
 use super::{Estimate, PriceEstimating, PriceEstimationError, Query};
 use cached::{Cached, TimedSizedCache};
-use std::{sync::Mutex, time::Duration};
+use std::{
+    sync::{Arc, Mutex},
+    time::Duration,
+};
 
 /// Price estimator wrapper that caches Ok results for some time.
 pub struct CachingPriceEstimator {
     inner: Box<dyn PriceEstimating>,
     cache: Mutex<TimedSizedCache<Query, Estimate>>,
+    metrics: Arc<dyn Metrics>,
+    name: String,
+}
+
+#[cfg_attr(test, mockall::automock)]
+pub trait Metrics: Send + Sync + 'static {
+    fn price_estimator_cache(&self, name: &str, misses: usize, hits: usize);
+}
+
+struct NoopMetrics;
+impl Metrics for NoopMetrics {
+    fn price_estimator_cache(&self, _: &str, _: usize, _: usize) {}
 }
 
 impl CachingPriceEstimator {
-    pub fn new(inner: Box<dyn PriceEstimating>, max_age: Duration, max_size: usize) -> Self {
+    pub fn new(
+        inner: Box<dyn PriceEstimating>,
+        max_age: Duration,
+        max_size: usize,
+        metrics: Arc<dyn Metrics>,
+        name: String,
+    ) -> Self {
         Self {
             inner,
             cache: Mutex::new(TimedSizedCache::with_size_and_lifespan_and_refresh(
@@ -17,6 +38,8 @@ impl CachingPriceEstimator {
                 max_age.as_secs(),
                 false,
             )),
+            metrics,
+            name,
         }
     }
 }
@@ -41,6 +64,12 @@ impl PriceEstimating for CachingPriceEstimator {
                 }
             }
         }
+
+        self.metrics.price_estimator_cache(
+            &self.name,
+            missing.len(),
+            queries.len() - missing.len(),
+        );
 
         let inner_results = self.inner.estimates(&missing).await;
         {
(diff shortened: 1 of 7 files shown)


nlordell (approved):
Nice! "And there is a crate for that" 😛.

On a more ominous note, we should probably consider vendoring and auditing our dependencies at some point.

crates/shared/src/price_estimation/cached.rs
```diff
@@ -0,0 +1,107 @@
+use super::{Estimate, PriceEstimating, PriceEstimationError, Query};
+use cached::{Cached, TimedSizedCache};
+use std::{sync::Mutex, time::Duration};
```
MartinquaXD, line 3: Don't we have to use `tokio::sync::Mutex` to avoid stopping every future in the current thread?
I believe if we make the `CachingPriceEstimator` the outer most layer of the estimator hierarchy this is probably fine but if we were to only wrap the "leaf" estimators in this cache this would lead to problems, no?

nlordell: I think `tokio::sync::Mutex` is only needed if the lock is held across `await`s. Otherwise, I imagine `std::sync::Mutex` to actually be more performant for this CPU bound code.

From the [`tokio-rs` docs](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use) itself:
> Contrary to popular belief, it is ok and often preferred to use the ordinary Mutex from the standard library in asynchronous code.

nlordell: Also, the compiler helps you out a bit here if you try to make a `BoxFuture` where you hold `MutexGuard`s across `await`s:
```rust
use futures::future::{BoxFuture, FutureExt as _};
use std::{sync::Mutex, future::Future};

fn foo(thing: impl Future<Output = ()> + Send + 'static) -> BoxFuture<'static, ()> {
    let mutex = Mutex::new(1u64);
    async move {
        // thing.await; // Here is OK.
        {
            let mut guard = mutex.lock().unwrap();
            thing.await; // Here is NOT OK.
            *guard += 1;
        }
        // thing.await; // Here is OK.
    }.boxed()
}
```

Will error:
```
error: future cannot be sent between threads safely
  --> src/main.rs:16:7
   |
16 |     }.boxed()
   |       ^^^^^ future created by async block is not `Send`
   |
   = help: within `impl futures::Future`, the trait `std::marker::Send` is not implemented for `std::sync::MutexGuard<'_, u64>`
note: future is not `Send` as this value is used across an await
  --> src/main.rs:12:13
   |
11 |             let mut guard = mutex.lock().unwrap();
   |                 --------- has type `std::sync::MutexGuard<'_, u64>` which is not `Send`
12 |             thing.await; // Here is NOT OK.
   |             ^^^^^^^^^^^ await occurs here, with `mut guard` maybe used later
13 |             *guard += 1;
14 |         }
   |         - `mut guard` is later dropped here
```

Playground link: <https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7af6ba330b2b4cb2211c748411a749fa>

vkgnosis:
I did somewhat audit this crate btw. Not 100% but I read some of the source, checked what it depends on, etc.

crates/shared/src/price_estimation/cached.rs
vkgnosis: You have to use an async aware mutex when you hold the lock across awaits or get deadlocks. When it isn't held across awaits like in this case you can decide. The mutex is only locked for short duration (read or write a couple of elements into the cache) so it will not block async executions for a long time. Non async mutexes have overall less overhead so I use them in situations like these.

MartinquaXD: Thanks for the explanation. It seems like I have to do more research.
Then the PR seems fine. Will approve.

MartinquaXD (approved):
LGTM

nlordell:
> I did somewhat audit this crate btw.

Not worried for this particular case, just meant something we should probably do in general and was unrelated to this PR. I was just thinking out loud. Created https://github.com/gnosis/gp-v2-services/issues/1552 so we don't forget to discuss this.

crates/shared/src/price_estimation/cached.rs
MartinquaXD: The [shared state](https://tokio.rs/tokio/tutorial/shared-state) section of the tokio tutorial also has some great insights on this topic. It specifically addresses some edge cases I was still unsure about. Namely high contention and it seems `tokio::sync::Mutex` wouldn't help in that scenario either.

crates/shared/src/price_estimation/cached.rs
```diff
@@ -0,0 +1,107 @@
...
+pub struct CachingPriceEstimator {
+    inner: Box<dyn PriceEstimating>,
+    cache: Mutex<TimedSizedCache<Query, Estimate>>,
+}
+
+impl CachingPriceEstimator {
+    pub fn new(inner: Box<dyn PriceEstimating>, max_age: Duration, max_size: usize) -> Self {
+        Self {
+            inner,
+            cache: Mutex::new(TimedSizedCache::with_size_and_lifespan_and_refresh(
```
sunce86, line 15: If query is not updated after max_age, then cache.cache_get(query) returns None right?

sunce86 (approved):
Lgtm

crates/shared/src/price_estimation/cached.rs
vkgnosis: Yes, it will return None for too old entries.

PR #1549, opened 2022-01-11
Hotfix: Change default parameter to be parsable.

nlordell:
This is a hotfix to fix the:
> Invalid value for '--simulation-gas-limit <simulation-gas-limit>': invalid digit found in string

Alerts we've been seeing in the past hour or so.


josojo (approved):
sorry. thanks for correcting

PR #1550, opened 2022-01-11
Implemented cancel_submission for Eden and CustomNodes strategies

sunce86:
Yesterday, we tried out the Flashbots+Eden transaction submission strategy on prod and had too many failing transactions.
Based on my analysis, quite a few of them could be prevented by:
1. Lowering the cli argument submission_retry_interval_seconds from 5s to 1-2s in order to detect simulation failure as soon as possible.
2. Cancelling the tx on Eden network by sending a noop transaction with higher gas price to replace the old tx (I was aiming to avoid this during the refactor of submission logic but it seems its inevitable).
3. Bidding for slot on Eden network.

This PR implements point 2.

This PR is not quite clean. Additional effort could be invested to clean it up, but for now I want to test the effects of this change prior to investing more time in further implementation.

### DIFF
--- a/crates/solver/src/settlement_submission/submitter.rs
+++ b/crates/solver/src/settlement_submission/submitter.rs
@@ -341,7 +341,7 @@ impl<'a> Submitter<'a> {
 
             match self.submit_api.submit_transaction(method.tx).await {
                 Ok(handle) => {
-                    previous_tx = Some((gas_price, handle.clone()));
+                    previous_tx = Some((gas_price, handle));
                     transactions.push(handle.tx_hash)
                 }
                 Err(err) => match err {
--- a/crates/solver/src/settlement_submission/submitter.rs
+++ b/crates/solver/src/settlement_submission/submitter.rs
@@ -65,6 +65,14 @@ pub struct TransactionHandle {
     pub tx_hash: H256,
 }
 
+#[derive(Debug, Clone)]
+pub struct CancelHandle {
+    /// transaction previosly submitted using TransactionSubmitting::submit_transaction()
+    pub submitted_transaction: TransactionHandle,
+    /// empty transaction with the same nonce used for cancelling the previously submitted transaction
+    pub noop_transaction: TransactionBuilder<DynTransport>,
+}
+
 #[cfg_attr(test, mockall::automock)]
 #[async_trait::async_trait]
 pub trait TransactionSubmitting: Send + Sync {
@@ -74,8 +82,10 @@ pub trait TransactionSubmitting: Send + Sync {
         &self,
         tx: TransactionBuilder<DynTransport>,
     ) -> Result<TransactionHandle, SubmitApiError>;
-    /// Cancels already submitted transaction using the transaction handle
-    async fn cancel_transaction(&self, id: &TransactionHandle) -> Result<()>;
+    /// Cancels already submitted transaction using the cancel handle
+    async fn cancel_transaction(&self, id: &CancelHandle) -> Result<()>;
+    /// Marks previously submitted transaction as outdated and prevents from being mined
+    async fn mark_transaction_outdated(&self, id: &TransactionHandle) -> Result<()>;
 }
 
 /// Gas price estimator specialized for sending transactions to the network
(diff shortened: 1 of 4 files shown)


crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -65,6 +65,14 @@ pub struct TransactionHandle {
     pub tx_hash: H256,
 }
 
+#[derive(Debug, Clone)]
+pub struct CancelHandle {
+    /// transaction previosly submitted using TransactionSubmitting::submit_transaction()
+    pub submitted_transaction: TransactionHandle,
```
MartinquaXD, line 71: This doesn't seem to be used anywhere. Is this for debugging purposes only?

crates/solver/src/settlement_submission/submitter/flashbots_api.rs
```diff
@@ -25,7 +28,11 @@ impl TransactionSubmitting for FlashbotsApi {
         super::common::submit_raw_transaction(self.client.clone(), URL, tx).await
     }
 
-    async fn cancel_transaction(&self, _id: &TransactionHandle) -> Result<()> {
+    async fn cancel_transaction(&self, _id: &CancelHandle) -> Result<()> {
+        Ok(())
+    }
```
MartinquaXD, line 33: Canceling on flashbots does nothing because our transaction is guaranteed to be included and therefore we never have to cancel, right?
Should this be documented or asserted?

crates/solver/src/settlement_submission/submitter.rs
sunce86: Nice catch. A bit of context: we have two different ways of cancelling the submitted transaction:
1. Using the previously submitted transaction hash/bundle_id (Flashbots are using this way)
2. Using the noop transaction (CustomNodes and Eden are using this way)
Seems like an Enum would be a perfect solution for this, however `CancelHandle` is populated inside `Submitter` which abstracts type of strategy, therefore we cant decide on the `CancelHandle` variant inside it, therefore we need to populate both 1 and 2 in order to provide valid `CancelHandle`.

And the response to the question, `submitted_transaction` is not used because currently Flashbots cancellation implementation is missing since its not mandatory for this PR experiment.

MartinquaXD: I see. Thanks for the detailed explanation. Will resolve the other conversation then. 👍

crates/solver/src/settlement_submission/submitter/flashbots_api.rs
MartinquaXD: See thread above for the explanation.

sunce86: Flashbots do provide cancelling logic, but two things could happen:
1. Our tx is sitting in flashbots mempool and its not forwarded to miners (still) -> our cancel removes tx from flashbots mempool
2. Our tx is forwarded to miners -> our cancel only removes tx from flashbots mempool but does not guarantee that our tx wont be mined (because there is a chance that miners mined a block containing our tx in the time between our cancel request and miners receiving a new block proposal not containing our tx)

On our side, cancelling on Flashbots is not that critical, because once we detect the tx will fail, our submission logic ends, and it starts running a new loop with a new settlement. In that case, our tx wont be mined as failed, because Flashbots gives that guarantee.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -65,6 +65,14 @@ pub struct TransactionHandle {
...
+    /// empty transaction with the same nonce used for cancelling the previously submitted transaction
+    pub noop_transaction: TransactionBuilder<DynTransport>,
```
nlordell, line 73: Not a huge fan of having this "leak" to the `submitter` (to me the fact that a strategy uses a "no-op transaction" is an implementation detail of that strategy).

That being said, this would require the submitter to keep additional information (nonce, last gas price used, etc.) so I don't see a very good way around this unfortunately.

nlordell (approved):
LGTM. Not a huge fan of the `noop_transaction` leaking into the "submitter" logic as it seems strategy specific. That being said, I can't think of a good solution to that given the data that is needed to build it, so I think its fine to compromise on that.

crates/solver/src/settlement_submission/submitter.rs
sunce86: True. But at least I don't see a third way of cancelling the tx than these two (in the future), so this structure can be treated as sealed.

Issue #1529, opened 2022-01-05
Add 1Inch Price Estimator


nlordell:
1Inch usually has good prices but very high gas costs. This made us only consider 1inch for bigger orders where the good prices were able to outweigh the bad gas costs.
The `solver` currently has a command line argument `min-order-size-one-inch` which controls whether an order is big enough that using the 1inch solver makes sense.
To get rid of this heuristic and use actual data to decide whether a settlement with 1inch makes sense, we can implement a 1inch price estimator.

Remaining tasks are:
- [x] move `OneInchClient` into `shared` crate
- [x] simplify `OneInchClient` error handling to reuse it for the quoting API
- [x] extend `OneInchClient` with quoting API
- [x] write unit tests for the quoting API
- [x] implement `OneInchPriceEstimator` using `OneInchClient`
- [x] write unit tests for `OneInchPriceEstimator`
- [x] make `orderbook` use new price estimator
- [ ] drop CLI argument `min-order-size-one-inch` from `solver`
  
For reference, something similar was already done for 0x here: https://github.com/gnosis/gp-v2-services/pull/1277

MartinquaXD:
@nlordell it seems like the [1Inch quoting API](https://docs.1inch.io/docs/aggregation-protocol/api/quote-params) only enables us to estimate sell orders.

Does it make sense to use the sell order estimation to calculate the price of the sell_token and buy_token relative to each other in order to estimate a buy order ourselves? I'm not sure if there are scenarios where "correct" estimations for buy orders would produce vastly different results than this workaround.

Alternatively we could disable estimating buy orders completely. I'm guessing since we have multiple estimators competing against each other we wouldn't have to use a possibly inaccurate result.

What are your thoughts?

PR #1553, opened 2022-01-12
Create a generic type for 1Inch REST API responses

MartinquaXD:
All 1Inch REST API endpoints return the same error type. To make some code within the `OneInchClient` reusable for the upcoming API to get quotes for an order, I refactored the error handling in `OneInchClient` a bit.
Instead of having a `QuoteResponse` and a `SwapResponse` I made the `enum RestResponse<T>` which can take different types for the `Ok` value but always uses the `RestError` for the `Err` value.
I also dropped the indirection of storing `Box`ed values within `Err` and `Ok`.
This resolves task 2 of #1529.

### Test Plan
Just refactored some things so no new tests.


vkgnosis (approved):

nlordell (approved):
Nice cleanup.

PR #1555, opened 2022-01-12
Use price estimator cache in orderbook

vkgnosis:
Next PR will add metrics to see cache hits and misses.

I intentionally used cache on the inner Paraswap and 0x estimators instead of later in the chain because not all estimators need to be cached.

Part of https://github.com/gnosis/gp-v2-services/issues/1533 

### Test Plan
CI

### DIFF
--- a/crates/orderbook/src/main.rs
+++ b/crates/orderbook/src/main.rs
@@ -20,7 +20,6 @@ use orderbook::{
     verify_deployed_contract_constants,
 };
 use primitive_types::H160;
-use shared::http_solver_api::{DefaultHttpSolverApi, SolverConfig};
 use shared::network::network_name;
 use shared::price_estimation::quasimodo::QuasimodoPriceEstimator;
 use shared::price_estimation::sanitized::SanitizedPriceEstimator;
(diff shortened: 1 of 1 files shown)


nlordell (approved):

crates/orderbook/src/main.rs
```diff
@@ -175,6 +178,19 @@ struct Arguments {
     /// The API endpoint to call the mip v2 solver for price estimation
     #[structopt(long, env)]
     quasimodo_solver_url: Option<Url>,
+
+    /// The maximum time price estimates will be cache for.
+    #[structopt(
+        long,
+        default_value = "10",
```
sunce86, line 185: Should we maybe try with lower value than 10s? 
Not sure if applicable in this situation but sometimes we have 3-4 blocks mined in 10s, so ethereum state could change drastically in very short time.

sunce86 (approved):
lg

vkgnosis:
FYI I changed this so that we instrument first and then cache so that the instrumentation isn't influenced by the cache.

crates/orderbook/src/main.rs
vkgnosis: I think the reason we see many blocks mined in short duration is the time it takes them to propagate through the network. So in one update of our node it suddenly sees a couple of new blocks. In that situation even caching for 1 second wouldn't help.

MartinquaXD: typo: "cache" -> "cached"

MartinquaXD (approved):
LGTM, just one typo.

PR #1556, opened 2022-01-12
Metrics for price estimator cache

vkgnosis:
Made InstrumentedPriceEstimator take Box<dyn> instead of generic because that made construction easier in main. If it is generic then I can't use a closure there. And overall I find less generics less complicated and performance doesn't matter here.

### Test Plan
CI


nlordell (approved):
Nice!

As for boxing, reasoning sounds good to me!

PR #1557, opened 2022-01-12
Add 1Inch quoting API

MartinquaXD:
In order to implement a `OneInchPriceEstimator` our `OneInchClient` needs to support the 1Inch quote API.
Unfortunately the [quote API](https://docs.1inch.io/docs/aggregation-protocol/api/quote-params) only supports quoting sell orders so I named the involved types and functions accordingly.
This PR resolves task 3 & 4 of #1529.

### Test Plan
For the new unit tests I tried to cover the same scenarios as the tests for the `swap` API.
- serializing `SellOrderQuoteQuery` only with required parameters set
- serializing `SellOrderQuoteQuery` with all parameters set
- deserializing authentic 1Inch quote response
- deserializing authentic 1Inch error response
- executing an actual quote request only with required parameters set
- executing an actual quote request only with all parameters set

### DIFF
--- a/crates/shared/src/oneinch_api.rs
+++ b/crates/shared/src/oneinch_api.rs
@@ -673,4 +673,236 @@ mod tests {
             .unwrap();
         println!("{:#?}", spender);
     }
+
+    #[test]
+    fn sell_order_quote_query_serialization() {
+        let base_url = Url::parse("https://api.1inch.exchange/").unwrap();
+        let url = SellOrderQuoteQuery {
+            from_token_address: addr!("EeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE"),
+            to_token_address: addr!("111111111117dc0aa78b770fa6a738034120c302"),
+            amount: 1_000_000_000_000_000_000u128.into(),
+            protocols: None,
+            fee: None,
+            gas_limit: None,
+            connector_tokens: None,
+            complexity_level: None,
+            main_route_parts: None,
+            virtual_parts: None,
+            parts: None,
+            gas_price: None,
+        }
+        .into_url(&base_url);
+
+        assert_eq!(
+            url.as_str(),
+            "https://api.1inch.exchange/v4.0/1/quote\
+                ?fromTokenAddress=0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee\
+                &toTokenAddress=0x111111111117dc0aa78b770fa6a738034120c302\
+                &amount=1000000000000000000"
+        );
+    }
+
+    #[test]
+    fn sell_order_quote_query_serialization_optional_parameters() {
+        let base_url = Url::parse("https://api.1inch.exchange/").unwrap();
+        let url = SellOrderQuoteQuery {
+            from_token_address: addr!("EeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE"),
+            to_token_address: addr!("111111111117dc0aa78b770fa6a738034120c302"),
+            amount: 1_000_000_000_000_000_000u128.into(),
+            protocols: Some(vec!["WETH".to_string(), "UNISWAP_V3".to_string()]),
+            fee: Some(0.5),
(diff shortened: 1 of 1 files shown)


vkgnosis:
Looks like there are some duplicate commits that are already in main. Probably something went wrong with a rebase. That is one of the disadvantage with the multi PR strategy when you merge into main first. It changes the commit hashes and git will probably complain about conflicts now.

MartinquaXD:
Now everything should be back in order.

nlordell (commented):
Code looks good.

> needs to support the 1Inch quote API.

Is there a reason we can't use the `/swap` API for getting prices as well?

vkgnosis:
It might just be a little less work on their side if they don't have to generate the full transaction for quote. But not sure.

vkgnosis (approved):

nlordell:
I think the `/swap` builds call data. Also,
> contract's allowance for the sell token is guaranteed to be big enough to not cause errors

I think there is a `disable_estimate` flag that makes it so it doesn't simulate the calldata in order so you don't need balance or approvals.

It sounds like the `/quote` API is more appropriate for price estimates so lets use that.

nlordell (approved):
LGTM!

nlordell:
> It might just be a little less work on their side if they don't have to generate the full transaction for quote. But not sure.

It could be like for 0x, with `/swap` they need to get binding quotes from PMMs. Given the number of requests we will be sending to their service, we probably don't want to do that.

crates/shared/src/oneinch_api.rs
```diff
@@ -37,6 +37,114 @@ impl<const MIN: usize, const MAX: usize> Display for Amount<MIN, MAX> {
...
+    /// Maximum number of token-connectors to be used in a transaction. Default: 2
+    pub complexity_level: Option<Amount<0, 3>>,
+    /// Limit maximum number of main route parts. Default: 20
+    pub main_route_parts: Option<Amount<1, 50>>,
+    /// Limit the number of virtual split parts.
+    pub virtual_parts: Option<Amount<1, 500>>,
+    /// Limit maximum number of parts each main route part can be split into. Default: 20
+    pub parts: Option<Amount<1, 100>>,
+    /// Gas price in atoms. Default: "fast" from network
+    pub gas_price: Option<U256>,
```
sunce86, line 74: nit: What are atoms? 
Do you know when or how 1inch uses gas_price data?

crates/shared/src/oneinch_api.rs
```diff
@@ -37,6 +37,114 @@ impl<const MIN: usize, const MAX: usize> Display for Amount<MIN, MAX> {
...
 }
 
+// The `Display` implementation for `H160` unfortunately does not print
+// the full address and instead uses ellipsis (e.g. "0xeeee…eeee"). This
+// helper just works around that.
+fn addr2str(addr: H160) -> String {
+    format!("{:?}", addr)
+}
+
+/// A query to get a quote forl a sell order with 1Inch.
```
sunce86, line 47: nit: typo

sunce86 (approved):

sunce86:
https://github.com/gnosis/gp-v2-services/blob/413f7bce7bdaae52c01a020dc32a374065ba9329/crates/shared/src/oneinch_api.rs#L4
Broken link. Maybe a good opportunity to fix it.

crates/shared/src/oneinch_api.rs
nlordell: Atoms are smallest units of tokens. For example, `USDC` has 6 decimals. So a "base" amount of `1.50` corresponds to `1500000` atoms. Similarly, for `DAI` with 18 decimals, an amount of `42.1337` corresponds to `42133700000000000000` atoms.

Its equivalent to `wei` for `(W)ETH`.

MartinquaXD: Atoms are the smallest divisible unit. I copied the term "atoms" from the `SwapQuery` comments but using the language of the 1Inch API makes more sense. 👍 
Now it's smallest divisible unit.
Not sure how 1Inch uses this data.

PR #1559, opened 2022-01-13
Add price estimator for 1Inch

MartinquaXD:
This PR adds a price estimator for 1Inch. There are is a caveat, though.
The 1Inch REST API only supports estimating sell orders. Because of that I introduced  `PriceEstimationError::UnsupportedOrderType`.
This PR resolves task 5, 6, 7 of #1529.

### Test Plan
New unit tests:
-estimating sell order succeeds
-estimating buy orders yields new error type
-REST API errors get propagated as `PriceEstimationError::Other`
-other errors from the `OneInchClientImpl` also get propagated as `PriceEstimationError::Other`
-test that disabling 1Inch protocols works and that the required API call is cached
-one ignored test which issues a request to the real 1Inch REST API

### DIFF
--- a/crates/orderbook/src/api.rs
+++ b/crates/orderbook/src/api.rs
@@ -295,11 +295,8 @@ impl IntoWarpReply for PriceEstimationError {
                 StatusCode::BAD_REQUEST,
             ),
             Self::UnsupportedOrderType => with_status(
-                error(
-                    "UnsupportedOrderType",
-                    "Can not estimate price for this order type",
-                ),
-                StatusCode::BAD_REQUEST,
+                internal_error(anyhow::anyhow!("UnsupportedOrderType").context("price_estimation")),
+                StatusCode::INTERNAL_SERVER_ERROR,
             ),
             Self::Other(err) => with_status(
                 internal_error(err.context("price_estimation")),
--- a/crates/orderbook/src/api.rs
+++ b/crates/orderbook/src/api.rs
@@ -294,6 +294,13 @@ impl IntoWarpReply for PriceEstimationError {
                 error("ZeroAmount", "Please use non-zero amount field"),
                 StatusCode::BAD_REQUEST,
             ),
+            Self::UnsupportedOrderType => with_status(
+                error(
+                    "UnsupportedOrderType",
+                    "Can not estimate price for this order type",
+                ),
+                StatusCode::BAD_REQUEST,
+            ),
             Self::Other(err) => with_status(
                 internal_error(err.context("price_estimation")),
                 StatusCode::INTERNAL_SERVER_ERROR,
--- a/crates/orderbook/src/main.rs
+++ b/crates/orderbook/src/main.rs
@@ -518,11 +518,11 @@ async fn main() {
                         &estimator.name(),
                     )),
                     PriceEstimatorType::OneInch => Box::new(instrumented_and_cached(
-                            Box::new(OneInchPriceEstimator::new(
-                                Arc::new(OneInchClientImpl::new(OneInchClientImpl::DEFAULT_URL, client.clone()).unwrap()),
-                                args.disabled_one_inch_protocols.clone()
-                            )),
-                            &estimator.name(),
+                        Box::new(OneInchPriceEstimator::new(
+                            Arc::new(OneInchClientImpl::new(OneInchClientImpl::DEFAULT_URL, client.clone()).unwrap()),
+                            args.disabled_one_inch_protocols.clone()
+                        )),
+                        &estimator.name(),
                     ))
                 },
             )
--- a/crates/orderbook/src/main.rs
+++ b/crates/orderbook/src/main.rs
@@ -197,6 +197,12 @@ struct Arguments {
     /// The maximum number of price estimates that will be cached.
     #[structopt(long, default_value = "1000")]
     price_estimator_cache_size: usize,
+
+    /// The list of disabled 1Inch protocols. By default, the `PMM1` protocol
+    /// (representing a private market maker) is disabled as it seems to
+    /// produce invalid swaps.
+    #[structopt(long, env, default_value = "PMM1", use_delimiter = true)]
+    disabled_one_inch_protocols: Vec<String>,
 }
 
 pub async fn database_metrics(metrics: Arc<Metrics>, database: Postgres) -> ! {
@@ -511,10 +517,11 @@ async fn main() {
                         }),
                         &estimator.name(),
                     )),
-                    PriceEstimatorType::OneInch => Box::new(instrumented(
-                            Box::new(OneInchPriceEstimator {
-                                api: Arc::new(OneInchClientImpl::new(OneInchClientImpl::DEFAULT_URL, client.clone()).unwrap())
-                            }),
+                    PriceEstimatorType::OneInch => Box::new(instrumented_and_cached(
+                            Box::new(OneInchPriceEstimator::new(
+                                Arc::new(OneInchClientImpl::new(OneInchClientImpl::DEFAULT_URL, client.clone()).unwrap()),
+                                args.disabled_one_inch_protocols.clone()
+                            )),
                             &estimator.name(),
                     ))
                 },
--- a/crates/orderbook/src/main.rs
+++ b/crates/orderbook/src/main.rs
@@ -20,6 +20,8 @@ use orderbook::{
     verify_deployed_contract_constants,
 };
 use primitive_types::H160;
+use shared::oneinch_api::OneInchClientImpl;
+use shared::price_estimation::oneinch::OneInchPriceEstimator;
 use shared::price_estimation::quasimodo::QuasimodoPriceEstimator;
 use shared::price_estimation::sanitized::SanitizedPriceEstimator;
 use shared::price_estimation::zeroex::ZeroExPriceEstimator;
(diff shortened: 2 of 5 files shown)


crates/orderbook/src/api.rs
```diff
@@ -198,6 +198,13 @@ impl IntoWarpReply for PriceEstimationError {
                 error("ZeroAmount", "Please use non-zero amount field"),
                 StatusCode::BAD_REQUEST,
             ),
+            Self::UnsupportedOrderType => with_status(
+                error(
+                    "UnsupportedOrderType",
+                    "Can not estimate price for this order type",
+                ),
+                StatusCode::BAD_REQUEST,
+            ),
```
vkgnosis, line 207: Imo: In our top level api we always want to support all order types. That some price estimators cannot support buy orders is an implementation detail. So I would treat this like `Other`.
Likewise in competition we should give Other priority over unsupported.

MartinquaXD: That's a good point. An internal server error seems more fitting than saying the order type is not supported at all.

MartinquaXD:
One thing which is still missing is that the `OneInchSolver` disables some 1Inch protocols because they lead to wrong settlements.
In order to get comparable results the `OneInchPriceEstimator` should disable the same protocols `OneInchSolver` does.
Before every swap request the `OneInchSolver` requests all protocols 1Inch supports (`/v3.0/1/protocols`) and disallows some of those protocols for the swap. Fetching the protocols on every price estimation is probably too much because 1Inch price estimations would happen way more often than 1Inch settlements. This means we probably have to cache the known protocols.

crates/orderbook/src/main.rs
```diff
@@ -509,6 +511,12 @@ async fn main() {
                         }),
                         &estimator.name(),
                     )),
+                    PriceEstimatorType::OneInch => Box::new(instrumented(
```
vkgnosis, line 514:
```suggestion
                    PriceEstimatorType::OneInch => Box::new(instrumented_and_cached(
```

Yesterday I introduced this cache for the other external price estimators. It probably makes sense for 1Inch too.

vkgnosis (approved):

vkgnosis:
> Fetching the protocols on every price estimation is probably too much because 1Inch price estimations would happen way more often than 1Inch settlements. This means we probably have to cache the known protocols.

Agree. Can have a long cache for this.

crates/orderbook/src/api.rs
```diff
@@ -198,6 +198,10 @@ impl IntoWarpReply for PriceEstimationError {
                 error("ZeroAmount", "Please use non-zero amount field"),
                 StatusCode::BAD_REQUEST,
             ),
+            Self::UnsupportedOrderType => with_status(
+                internal_error(anyhow::anyhow!("UnsupportedOrderType").context("price_estimation")),
+                StatusCode::INTERNAL_SERVER_ERROR,
```
nlordell, line 203: Maybe this should be a `BAD_REQUEST`?

crates/shared/src/price_estimation/oneinch.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+}
+
+#[async_trait::async_trait]
+impl PriceEstimating for OneInchPriceEstimator {
+    async fn estimates(
+        &self,
+        queries: &[Query],
+    ) -> Vec<anyhow::Result<Estimate, PriceEstimationError>> {
+        debug_assert!(queries.iter().all(|query| {
```
nlordell, line 56: Maybe add a message to the `debug_assert` to add some context as to why this should never happen? (The way I see it this is because it should always be wrapped in `SanitizedPriceEstimator` which guarantees these invariants, but it may not be obvious to someone less familiar with this code).

nlordell (approved):
Nice and easy PR to review.

One open question about choice of error priority and a small nit on `debug_assert` context. Nothing blocking merging though.

crates/shared/src/price_estimation/competition.rs
```diff
@@ -119,14 +119,18 @@ fn join_error(a: PriceEstimationError, b: PriceEstimationError) -> PriceEstimati
...
         | (_, err @ PriceEstimationError::ZeroAmount) => err,
         (err @ PriceEstimationError::UnsupportedToken(_), _)
         | (_, err @ PriceEstimationError::UnsupportedToken(_)) => err,
         (err @ PriceEstimationError::NoLiquidity, _)
         | (_, err @ PriceEstimationError::NoLiquidity) => err,
-        (err @ PriceEstimationError::Other(_), _) => err,
+        (err @ PriceEstimationError::Other(_), _) | (_, err @ PriceEstimationError::Other(_)) => {
+            err
+        }
+        (err @ PriceEstimationError::UnsupportedOrderType, _) => err,
```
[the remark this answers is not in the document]
vkgnosis: Related to my comment above.

crates/orderbook/src/api.rs
nlordell: 🙈 I see that @vkgnosis suggested the opposite here.

crates/orderbook/src/api.rs
nlordell: Good point. I also naively assumed `BAD_REQUEST` here. Maybe we can add a comment about this in code as it wasn't immediately obvious to me.

MartinquaXD:
Added the option to disable 1Inch protocols while also limiting the number of API requests by caching the API resonses. I'm not particularly happy about the caching implementation but that was the least messy thing I could come up with using the `cached` crate.

crates/orderbook/src/main.rs
```diff
@@ -509,6 +517,13 @@ async fn main() {
                         }),
                         &estimator.name(),
                     )),
+                    PriceEstimatorType::OneInch => Box::new(instrumented_and_cached(
+                            Box::new(OneInchPriceEstimator::new(
+                                Arc::new(OneInchClientImpl::new(OneInchClientImpl::DEFAULT_URL, client.clone()).unwrap()),
+                                args.disabled_one_inch_protocols.clone()
+                            )),
+                            &estimator.name(),
```
nlordell, line 525: Indentation feels off. Surprised `cargo fmt` didn't catch this.

crates/shared/src/price_estimation/oneinch.rs
```diff
@@ -0,0 +1,318 @@
...
+use cached::{Cached, TimedSizedCache};
+use futures::future;
+use model::order::OrderKind;
+use primitive_types::U256;
+use std::sync::{Arc, Mutex};
+
+pub struct OneInchPriceEstimator {
+    api: Arc<dyn OneInchClient>,
+    disabled_protocols: Vec<String>,
+    allowed_protocols: Arc<Mutex<TimedSizedCache<Vec<String>, Vec<String>>>>,
```
nlordell, line 14: Since this logic is also used by the 1Inch solver, could we maybe move it to a shared component instead of re-implementing it here (I'm happy that `cached` is used instead of custom code!).

I'm fine with doing it in a separate PR to not have too much "scope creep" in this one.

crates/shared/src/price_estimation/oneinch.rs
```diff
@@ -0,0 +1,318 @@
...
+            .unwrap()
+            .cache_get(&self.disabled_protocols)
+        {
+            return Ok(Some(allowed_protocols.clone()));
+        }
+
+        let current_protocols = self.api.get_protocols().await?.protocols;
+        let allowed_protocols: Vec<String> = current_protocols
+            .into_iter()
+            // linear search through the Vec is okay because it's very small
```
nlordell, line 80: I'm willing to bet this is a tiny fraction of the total time taken to complete this method (most of it being waiting on I/O).

nlordell (approved):
Changes look great!

I would have been fine to have the `protocols` filtering to be done in a separate follow up PR to keep this one leaner and given that we had already reviewed the original code. That being said, I don't see a point in splitting it up now that the code is here.

crates/orderbook/src/main.rs
MartinquaXD: When I run cargo fmt locally it doesn't work properly on the big main files. I guess CI has the same problem.

MartinquaXD: `rustfmt` doesn't give me any error message for that particular file. I just fixed the indentation manually now.

crates/shared/src/price_estimation/oneinch.rs
MartinquaXD: Sure, but I still like to be aware of the runtime of the CPU bound code.

crates/shared/src/price_estimation/oneinch.rs
```diff
@@ -0,0 +1,318 @@
...
+impl PriceEstimating for OneInchPriceEstimator {
+    async fn estimates(
+        &self,
+        queries: &[Query],
+    ) -> Vec<anyhow::Result<Estimate, PriceEstimationError>> {
+        debug_assert!(queries.iter().all(|query| {
+            query.buy_token != model::order::BUY_ETH_ADDRESS
+                && query.sell_token != model::order::BUY_ETH_ADDRESS
+                && query.sell_token != query.buy_token
+        }), "the hierarchy of price estimators should be set up such that OneInchPriceEstimator is a descendant of SanitizedPriceEstimator");
```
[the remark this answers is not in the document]
MartinquaXD: Did that as the last finishing touch of the PR. Will merge now.

crates/shared/src/price_estimation/oneinch.rs
nlordell: I was mostly agreeing with your statement 😛. That the runtime of the code was fine in this instance.

Issue #1534, opened 2022-01-06
Refactor Price Estimation to Use Rich Price API Data


nlordell:
Both 0x and ParaSwap APIs have "rich" price responses that would allow us to cut the price estimation time in half, as well as half the number of requests we make.

Specifically for 0x:
```
% curl -s 'https://api.0x.org/swap/v1/quote?sellToken=0xba100000625a3754423978a60c9317c58a424e3d&buyToken=0x6810e776880c02933d47db1b9fc05908e5386b96&sellAmount=1000000000000000000000&skipValidation=true' | jq '{sellTokenToEthRate,buyTokenToEthRate}'
{
  "sellTokenToEthRate": "180.75973275686745708",
  "buyTokenToEthRate": "7.01021738251805285"
}
```

This means that the API request we make for estimating the traded amount already contains enough information to give us the amount of sell token to cover a specific ETH gas fee.

And for ParaSwap:
```
% curl -s 'https://apiv4.paraswap.io/v2/prices/?from=0xba100000625a3754423978a60c9317c58a424e3d&to=0x6810e776880C02933D47DB1b9fc05908e5386b96&amount=1000000000000000000000&side=SELL&network=1' | jq '.priceRoute | {fromUSD,toUSD}'
{
  "fromUSD": "16760.0000000000",
  "toUSD": "16349.1885254736"
}
```

This API request provide a USD value of the amounts being traded, so with a simple ETH/USD price feed (which we can have by relying on our current liquidity sources which has more than enough ETH/USD stable coin liquidity to give a very indicative price) we can also compute the required sell token amount to cover a specific ETH gas fee.

**Note**: Implementing this approach might make #1533 less relevant.

nlordell:
One issue with using the price estimates from ParaSwap and 0x rich price data, is that we don't quite know where the prices come from. Furthermore, it isn't clear for illiquid tokens whether or not the price will be accurate for our given `native_token_price_estimation_amount`.

nlordell:
Quasimodo can send the order for the price estimate along with a synthetic "buy 1 ETH for `sell_token`" order. If both are marked as `mandatory` then we can get the estimated token amounts along with the sell token price in a single round-trip.

nlordell:
Closing this as we have decided against using the rich pricing API for now and instead will use a new component for computing these prices.

PR #1561, opened 2022-01-13
Cache Quasimodo price estimator

vkgnosis:
Just like the other external price estimator we can cache Quasimodo to reduce some stress on it.
Based on conversation on Slack that Quasimodo is under high load.

### Test Plan
CI

### DIFF
--- a/crates/orderbook/src/main.rs
+++ b/crates/orderbook/src/main.rs
@@ -482,7 +482,7 @@ async fn main() {
                         }),
                         &estimator.name(),
                     )),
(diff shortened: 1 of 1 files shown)


MartinquaXD (approved):

nlordell (approved):

nlordell:
Also, in the long term, we can consider making a single `solve` request with a virtual buy ETH order. This is something that @marcovc suggested to reduce one of the round-trips.

Related to #1534

PR #1564, opened 2022-01-13
added cancellation after deadline is hit

sunce86:
When deadline is reached in transaction submission loop, we want to cancel the last submitted transaction in order to avoid entering next submission loop with potentially different solver (therefore different sender address). In this situation, two different transactions with similar settlement orders could be mined, one of them will be successful, other one will be mined as failed with error from settlement contract: `GPv2: order filled`

This cancellation significantly reduces the chances of this error happening.

The rest of the code is refactor/cleanup so we can call cancelling on specific line in the code.

### DIFF
--- a/crates/solver/src/settlement_submission/submitter.rs
+++ b/crates/solver/src/settlement_submission/submitter.rs
@@ -186,7 +186,15 @@ impl<'a> Submitter<'a> {
                 Ok(None)
             },
             _ = deadline_future.fuse() => {
-                tracing::info!("stopping submission because deadline has been reached");
+                tracing::info!("stopping submission because deadline has been reached. cancelling last submitted transaction...");
+
+                if let Some(transaction) = transactions.last() {
+                    if let Ok(gas_price) = self.gas_price_estimator.estimate_with_limits(21000., Duration::from_secs(0)).await {
+                        if let Err(err) = self.cancel_transaction(transaction, &gas_price, nonce).await {
+                            tracing::warn!("cancellation failed: {:?}", err);
+                        }
+                    }
+                }
                 Ok(None)
             },
         };
@@ -211,6 +219,11 @@ impl<'a> Submitter<'a> {
                 "waiting up to {} seconds to see if a transaction was mined",
                 MINED_TX_PROPAGATE_TIME.as_secs()
             );
+
+            let transactions = transactions
+                .into_iter()
+                .map(|handle| handle.tx_hash)
+                .collect::<Vec<_>>();
 
             loop {
                 if let Some(receipt) =
@@ -268,11 +281,10 @@ impl<'a> Submitter<'a> {
         settlement: Settlement,
         nonce: U256,
         params: &SubmitterParams,
-        transactions: &mut Vec<H256>,
+        transactions: &mut Vec<TransactionHandle>,
     ) -> SubmissionError {
         let target_confirm_time = Instant::now() + params.target_confirm_time;
 
-        // gas price and raw signed transaction
         let mut previous_tx: Option<(EstimatedGasPrice, TransactionHandle)> = None;
 
         loop {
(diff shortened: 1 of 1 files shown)


vkgnosis:
> therefore different sender address). In this situation, two different transactions with similar settlement orders could be mined, one of them will be successful, other one will be mined as failed

Unfortunate side effect of having per solver accounts. If we had only one account for all settlements then we could run different submission strategies at the same time without worrying because they'd have the same nonce. Relevant to @nlordell as we talked about that today.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -186,7 +186,15 @@ impl<'a> Submitter<'a> {
                 Ok(None)
             },
             _ = deadline_future.fuse() => {
-                tracing::info!("stopping submission because deadline has been reached");
+                tracing::info!("stopping submission because deadline has been reached. cancelling last submitted transaction...");
+
+                if let Some(transaction) = transactions.last() {
+                    if let Ok(gas_price) = self.gas_price_estimator.estimate_with_limits(21000., Duration::from_secs(0)).await {
+                        if let Err(err) = self.cancel_transaction(transaction, &gas_price, nonce).await {
```
bh2smith, line 193: I wish python had if let. So useful!

bh2smith (approved):
Looks pretty good! Nice PR description. Helped a lot.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -186,7 +186,15 @@ impl<'a> Submitter<'a> {
                 Ok(None)
             },
             _ = deadline_future.fuse() => {
-                tracing::info!("stopping submission because deadline has been reached");
+                tracing::info!("stopping submission because deadline has been reached. cancelling last submitted transaction...");
+
+                if let Some(transaction) = transactions.last() {
+                    if let Ok(gas_price) = self.gas_price_estimator.estimate_with_limits(21000., Duration::from_secs(0)).await {
```
fleupold, line 192: Shouldn't this just be the gas price of the last attempt * some increment factor? We don't care that the cancellation is executed quickly, but just that it has higher priority than the last settlement attemp?

How do we ensure the value that is returned here is actually higher than the transaction that is currently in the mem-pool.

vkgnosis: @sunce86 Did you see this comment? Wondering because today there was an error

```
ERROR (pod: dev-dfusion-v2-solver-mainnet-845665679-8ccc2):
2022-01-18T10:14:52.994Z ERROR solver::settlement_submission::submitter: submission failed: invalid rpc response: {"jsonrpc":"2.0","id":1,"error":{"code":"-32000","message":"replacement transaction underpriced","data":"0x...
```

sunce86: @vkgnosis @fleupold Sorry, I somehow missed this comment. 

Anyway, I wanted to use the latest gas price estimation as a more secure option here, because network could have sudden surges in gas prices, then deadline could be hit, and our replacement tx could be sent with gas price < block base fee. I am not confident that all ethereum clients are respecting the same replacement strategy "rules" (these rules are not even consensus rules I believe, so each client could have its own rules and specifics dependent on their tx pool implementation - that could even change in the future). 
If you have a strong opinion that we should use the latest submitted gas price + bump, I can make the PR and experiment with that?

Regarding the "replacement transaction underpriced" message, this is a message from geth client: https://github.com/ethereum/go-ethereum/blob/51eb5f8ca816b6253233177ae952b2941d4520dd/core/tx_pool.go#L717
happening when configured price bump is not respected.
My guess is there are some miners using geth client with non-default configuration (higher) for price bump, and there is nothing we can do about that. I've also reached out to Eden crew to give me more info about this issue, however they did not respond.
For now we just ignore that error message, and wait for next loop iteration to submit with even higher gas price until our price bump is high enough.

sunce86: > How do we ensure the value that is returned here is actually higher than the transaction that is currently in the mem-pool.

We don't. If gas price estimator returns the lower gas price than latest submitted tx, then we assume that last submitted tx is mined in the meantime -> our tx cancel fails which is ok

vkgnosis: Thanks for the explanation.

> I am not confident that all ethereum clients are respecting the same replacement strategy "rules"

Agree with that. 

> I wanted to use the latest gas price estimation as a more secure option here, because network could have sudden surges in gas prices

Could maybe do `max(current_gas_price, previous_tx_gas_price * 1.125`).

PR #1566, opened 2022-01-14
Merge price estimator calls into one

vkgnosis:
The price estimator trait already supports multiple queries in one call
which can be more efficient than multiple calls with one query each.

### Test Plan
CI


MartinquaXD (approved):

nlordell (approved):
Nice catch!

PR #1572, opened 2022-01-17
Add ProtocolCache for 1Inch

MartinquaXD:
The `OneInchPriceEstimator` and `OneInchSolver` need to disable certain protocols (`PMM1`) because they can lead to wrong settlements.
To not duplicate the code responsible for limiting the protocols, I added `ProtocolCache` to `oneinch_api.rs` and made both structs us it.

This PR introduces caching to the `OneInchSolver` but it shouldn't be a big deal. 1Inch doesn't change the supported protocols on a regular basis, so having a small delay (at most 1 minute) until `OneInchSolver` is able to utilize new protocols can not lead to many suboptimal solutions.

### Test Plan
Unit tests for `ProtocolCache`

### DIFF
--- a/crates/shared/src/oneinch_api.rs
+++ b/crates/shared/src/oneinch_api.rs
@@ -4,13 +4,14 @@
 //! <https://docs.1inch.io/docs/aggregation-protocol/api/swagger>
 use crate::solver_utils::{deserialize_prefixed_hex, Slippage};
 use anyhow::{ensure, Context, Result};
-use cached::{Cached, TimedSizedCache};
+use cached::{Cached, TimedCache};
 use ethcontract::{H160, U256};
 use model::u256_decimal;
 use reqwest::{Client, IntoUrl, Url};
 use serde::Deserialize;
 use std::fmt::{self, Display, Formatter};
 use std::sync::{Arc, Mutex};
+use std::time::Duration;
 
 /// Parts to split a swap.
 ///
@@ -387,17 +388,14 @@ where
 }
 
 #[derive(Debug, Clone)]
-pub struct ProtocolCache(Arc<Mutex<TimedSizedCache<(), Vec<String>>>>);
+pub struct ProtocolCache(Arc<Mutex<TimedCache<(), Vec<String>>>>);
 
 impl ProtocolCache {
-    pub fn new(cache_validity_in_seconds: u64) -> Self {
-        Self(Arc::new(Mutex::new(
-            TimedSizedCache::with_size_and_lifespan_and_refresh(
-                1,
-                cache_validity_in_seconds,
-                false,
-            ),
-        )))
+    pub fn new(cache_validity_in_seconds: Duration) -> Self {
+        Self(Arc::new(Mutex::new(TimedCache::with_lifespan_and_refresh(
+            cache_validity_in_seconds.as_secs(),
+            false,
+        ))))
     }
 
     pub async fn get_all_protocols(&self, api: &dyn OneInchClient) -> Result<Vec<String>> {
@@ -436,7 +434,7 @@ impl ProtocolCache {
 
 impl Default for ProtocolCache {
     fn default() -> Self {
-        Self::new(60)
+        Self::new(Duration::from_secs(60))
     }
 }
 
--- a/crates/shared/src/oneinch_api.rs
+++ b/crates/shared/src/oneinch_api.rs
@@ -956,4 +956,38 @@ mod tests {
             .unwrap();
         println!("{:#?}", swap);
     }
+
+    #[tokio::test]
+    async fn allowing_all_protocols_will_not_use_api() {
+        let mut api = MockOneInchClient::new();
+        api.expect_get_protocols().times(0);
+        let allowed_protocols = ProtocolCache::default()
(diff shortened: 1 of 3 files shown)


crates/shared/src/oneinch_api.rs
```diff
@@ -384,6 +386,55 @@ where
     serde_json::from_str(&response?).context("1inch result parsing failed")
 }
 
+#[derive(Debug, Clone)]
+pub struct ProtocolCache(Arc<Mutex<TimedSizedCache<Vec<String>, Vec<String>>>>);
```
vkgnosis, line 390: I think we should cache the raw list of protocols that 1nch returns instead of this construct where the entries in the cache depend on the disabled protocols. The underlying data is always the same so I imagine there would be a "get current protocols" function that either returns the cached value or updates the cache and returns the value.
That you might or might not implement with this cached trait. I'm saying might not because when there is only one value it might be so simple that the crate doesn't add anything.

nlordell: One advantage of caching the protocol list instead of the enabled/disabled protocols is that it, in theory, allows multiple instances of `1Inch{PriceEstimator,Solver}` to share the same cache.

That being said, we don't really do that now so maybe its a bit of "overarchitecting" on my part.

> there is only one value it might be so simple that the crate doesn't add anything.

> But even if the general use case is only storing a single value, I would prefer using the crate over implementing a cache myself.

I don't have a strong opinion here. If we added `cached` as a dependency _just_ for this cache, I would tend to agree that we shouldn't use, but since its already included in our dependency tree - why not?

crates/shared/src/oneinch_api.rs
```diff
@@ -384,6 +386,55 @@ where
...
+    pub async fn get_allowed_protocols(
+        &self,
+        #[allow(clippy::ptr_arg)] disabled_protocols: &Vec<String>,
+        api: &dyn OneInchClient,
+    ) -> Result<Option<Vec<String>>> {
+        if disabled_protocols.is_empty() {
+            return Ok(None);
+        }
+
+        if let Some(allowed_protocols) = self.0.lock().unwrap().cache_get(disabled_protocols) {
```
nlordell, line 412: Interesting... This doesn't use `Borrow::Target` so you have to use a `&Vec<String>` instead of the more Rust-y `&[String]`.

crates/shared/src/oneinch_api.rs
nlordell: You could use `TimedCache<(), Vec<String>>` if you wanted to keep a single list right?

vkgnosis: > I wanted to skip filtering out disabled protocols every time we need the allowed list

I like doing it that way a lot more. Filtering takes no time compared to making the network request and it has the advantage of simpler code because only one thing is cached.

crates/shared/src/oneinch_api.rs
```diff
@@ -384,6 +386,60 @@ where
     serde_json::from_str(&response?).context("1inch result parsing failed")
 }
 
+#[derive(Debug, Clone)]
+pub struct ProtocolCache(Arc<Mutex<TimedSizedCache<(), Vec<String>>>>);
```
nlordell, line 390: Can this just be `TimedCache` since the size can only ever be 1 so we don't care about it?

nlordell (approved):
Personally, I prefer it this way as well (caching a single value).

Also, generally speaking, very clean PR with some nice unit tests 🎉.

crates/shared/src/oneinch_api.rs
```diff
@@ -384,6 +386,60 @@ where
...
+
+impl ProtocolCache {
+    pub fn new(cache_validity_in_seconds: u64) -> Self {
```
vkgnosis, line 393: Nit: Could be `Duration` and then use .as_secs() to pass into the cache constructor which sadly accepts only integer seconds instead of Duration.

crates/shared/src/oneinch_api.rs
vkgnosis: I think that is correct.

crates/shared/src/oneinch_api.rs
```diff
@@ -384,6 +386,60 @@ where
...
+            .filter(|protocol| !disabled_protocols.contains(protocol))
+            .collect();
+
+        Ok(Some(allowed_protocols))
+    }
+}
+
+impl Default for ProtocolCache {
+    fn default() -> Self {
+        Self::new(60)
```
vkgnosis, line 439: Nit: We can make this a lot longer than 60 seconds. I don't expect them to add protocols often and even if they do it wouldn't be a problem to not use a new protocol right that moment. Could be an hour. On the other hand fetching the protocols isn't expensive either so not sure.

vkgnosis (approved):

crates/shared/src/oneinch_api.rs
MartinquaXD: They should behave the same in our case but their implementation is different. `SizedCache` and `TimedSizedCache` use quite a bit of `unsafe` in the authors homegrown implementation of the inner map storing the cached values.
`TimedCache` uses a simple `std::collections::HashMap` so let's go with that instead.

PR #1569, opened 2022-01-16
Box warp routes

MartinquaXD:
Boxing the warp routes improves build times dramatically.

### Test Plan
functionality didn't change -> no new tests

PR #1574, opened 2022-01-17
Add 60 minute timeout to all CI jobs

taminomara:
This commit adds 60-minute timeouts to all CI jobs triggered by PRs and commits to main. Release jobs are unaffected.

This will help to spot issues when builds take too much time due to, for example, compilation issues (see #1569).

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -7,6 +7,7 @@ on:
     branches: [main]
 jobs:
   rust-misc:
+    timeout-minutes: 60
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


taminomara:
Note: taking a look at the current build history, all builds seem to take approximately 10 minutes, with maximum up to 30 minutes.

nlordell:
> I think 60 minutes timeout is a reasonable one.

Sounds good to me.

nlordell (approved):
LGTM. Lets wait a bit to see what others have to say about the timeout amount.

MartinquaXD:
> Lets wait a bit to see what others have to say about the timeout amount.
I'm guessing there aren't too many ways to add 30 minutes of compile time with a single PR. 🤔 
I think we shouldn't make the timeout too long to actually get some use out of it besides REALLY exceptional cases.
Maybe 40 minutes is long enough to cover our regular bad cases and would still be valuable for catching smaller regressions.

vkgnosis (approved):
Looks fine to me. There is always correct timeout so we have to arbitrarily pick somehow. Could go as low as 30 minutes.

taminomara:
I'd say we should account for other possible delays, like something else happening on the machine where the build is running. Wouldn't want false-positive failures. As for regression testing build times, I'd say if we want to implement this, we should go with analysing general build time trend, not setting up a timeout.

taminomara:
BTW I've realised that I haven't seen the issue locally because I was always building in debug mode, i.e. with minimal optimisations.

PR #1575, opened 2022-01-17
Eden and Flashbots Api URL made configurable

sunce86:
Last 48 hours we have an issue with the submission of the transaction to the Eden network. Their Api endpoint is failing.
Eden crew suggested to try out the different URL. That said, this seems like a convenient time to set these URLs configurable.


vkgnosis (approved):

bh2smith (approved):
Looks good to me.

nlordell (approved):

PR #1576, opened 2022-01-18
Prepare 1Inch API for gnosis chain support

MartinquaXD:
So far `OneInchClientImpl` was hardcoded to only support mainnet. Because the 1Inch team plans to support Gnosis Chain in the future, we should be prepared to enable it with minimal effort.

Instead of hardcoding the chain_id `1` everywhere the chain_id has to be passed to functions which need it.
Instead of asserting a supported chain_id in `OneInchSolver` it is now done in the `OneInchClientImpl`. This way we don't have have that code in the `OneInchSolver` AND `-PriceEstimator` because they can't even be constructed if `OneInchClientImpl` fails to construct.

### Test Plan
Added unit test which tries to instantiate `OneInchClientImpl` with an unsupported chain_id.

### DIFF
--- a/crates/orderbook/src/main.rs
+++ b/crates/orderbook/src/main.rs
@@ -519,7 +519,7 @@ async fn main() {
                     )),
                     PriceEstimatorType::OneInch => Box::new(instrumented_and_cached(
                         Box::new(OneInchPriceEstimator::new(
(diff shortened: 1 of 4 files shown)


nlordell (approved):
LGTM

vkgnosis (approved):

PR #1577, opened 2022-01-18
Use same parameters for 1Inch swap and quote

MartinquaXD:
In order for `OneInchPriceEstimator` quotes to be meaningful they need to use the same arguments as the `OneInchSolver` swaps wherever possible. Otherwise the price estimation and the final price could differ quite a bit.
I created a new struct which holds parameters which are common to `/swap` and `/quote` requests. Creating those requests is now done with a function which sets the correct defaults. There is only a single source of truth so no need to manually keep the defaults for `/quote` and `/swap` queries synchronized.

### Test Plan
No new tests

### DIFF
--- a/crates/shared/src/oneinch_api.rs
+++ b/crates/shared/src/oneinch_api.rs
@@ -46,6 +46,7 @@ pub struct QuoteAndSwapCommonOptions {
     /// Contract address of a token to buy.
     pub to_token_address: H160,
     /// Amount of a token to sell, set in atoms.
+    pub amount: U256,
     /// Maximum number of token-connectors to be used in a transaction.
     pub complexity_level: Option<Amount<0, 3>>,
     /// List of protocols to use for the swap.
@@ -63,10 +64,12 @@ impl QuoteAndSwapCommonOptions {
         sell_token: H160,
         buy_token: H160,
         protocols: Option<Vec<String>>,
+        amount: U256,
     ) -> Self {
         Self {
             from_token_address: sell_token,
             to_token_address: buy_token,
+            amount,
             protocols,
             complexity_level: Some(Amount::new(2).unwrap()),
             gas_limit: Some(Amount::new(750_000).unwrap()),
@@ -86,8 +89,6 @@ fn addr2str(addr: H160) -> String {
 /// A query to get a quote for a sell order with 1Inch.
 #[derive(Clone, Debug)]
 pub struct SellOrderQuoteQuery {
-    /// Amount of a token to sell, set in smallest divisible unit.
-    pub amount: U256,
     /// Percentage how much of the from_token_address amount should be sent to the referrer
     /// address. Values: [0, 3], Default 0.0
     pub fee: Option<f64>,
@@ -113,7 +114,7 @@ impl SellOrderQuoteQuery {
                 &addr2str(self.common.from_token_address),
             )
             .append_pair("toTokenAddress", &addr2str(self.common.to_token_address))
-            .append_pair("amount", &self.amount.to_string());
+            .append_pair("amount", &self.common.amount.to_string());
 
         if let Some(protocols) = self.common.protocols {
             url.query_pairs_mut()
@@ -167,13 +168,12 @@ impl SellOrderQuoteQuery {
         protocols: Option<Vec<String>>,
     ) -> Self {
         Self {
-            amount: in_amount,
             fee: None,
             connector_tokens: None,
             virtual_parts: None,
             gas_price: None,
             common: QuoteAndSwapCommonOptions::with_default_options(
-                sell_token, buy_token, protocols,
+                sell_token, buy_token, protocols, in_amount,
             ),
         }
     }
@@ -199,7 +199,6 @@ pub struct SellOrderQuote {
 /// added incrementally as needed.
 #[derive(Clone, Debug)]
 pub struct SwapQuery {
-    pub amount: U256,
     /// Address of a seller.
     ///
     /// Make sure that this address has approved to spend `from_token_address`
@@ -225,7 +224,7 @@ impl SwapQuery {
                 &addr2str(self.common.from_token_address),
             )
             .append_pair("toTokenAddress", &addr2str(self.common.to_token_address))
-            .append_pair("amount", &self.amount.to_string())
+            .append_pair("amount", &self.common.amount.to_string())
             .append_pair("fromAddress", &addr2str(self.from_address))
             .append_pair("slippage", &self.slippage.to_string());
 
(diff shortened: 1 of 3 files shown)


crates/shared/src/price_estimation/oneinch.rs
```diff
@@ -19,6 +19,7 @@ impl OneInchPriceEstimator {
             return Err(PriceEstimationError::UnsupportedOrderType);
         }
 
+        // If applicable, keep these parameters synchronized with OneInchSolver swap parameters!
```
nlordell, line 22: Instead of manually synchronizing can we maybe add something like:
```
impl SellOrderQuoteQuery {
    pub fn with_default_options(sell_token, buy_token, in_amount, protocols) -> Self {
        // ...
    }
}
```

To construct a `SellOrderQuoteQuery` with the default options? This way it there is no manual requirement to keep the two files in sync.

nlordell (changes requested):
Nice catch! Like the purpose of the PR in general. I suggested a slightly different approach to avoid doing manual synchronization of parameters across files.

crates/shared/src/price_estimation/oneinch.rs
MartinquaXD: To have a "single source of truth" we would have to have a struct storing the defaults for settings which are supported in both `SwapQuery` and `SellOrderQuoteQuery` and construct the respective types from it.
If it's good enough to manually keep code synchronized within a single file I would add `with_default_options` function to both structs.

nlordell: Ah! They are different structs.

One other possibility would be to:
```rust
struct SellOrderQuoteQuery {
  things: Things,
  common: CommonQueryParameters,
}

struct SwapQuery {
  other_things: OtherThings,
  common: CommonQueryParameters,
}

struct CommonQueryParameters {
  common_things: CommonThings,
}

impl Default for CommonQueryParameters {
  fn default() -> Self {
    // ...
  }
}
```

This way we can keep track of the common parameter defaults that we used. I'm always wary of duplicating "magic values" as, in my experience, they very quickly get out of sync despite comments urging them to stay in sync.

crates/shared/src/oneinch_api.rs
```diff
@@ -39,6 +39,43 @@ impl<const MIN: usize, const MAX: usize> Display for Amount<MIN, MAX> {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct QuoteAndSwapCommonOptions {
+    /// Contract address of a token to sell.
+    pub from_token_address: H160,
+    /// Contract address of a token to buy.
+    pub to_token_address: H160,
+    /// Amount of a token to sell, set in atoms.
```
nlordell, line 48: I think this may be a typo:
```suggestion
    /// Amount of a token to sell, set in atoms.
    pub amount: U256,
```

crates/shared/src/oneinch_api.rs
```diff
@@ -49,31 +86,18 @@ fn addr2str(addr: H160) -> String {
 /// A query to get a quote for a sell order with 1Inch.
 #[derive(Clone, Debug)]
 pub struct SellOrderQuoteQuery {
-    /// Contract address of a token to sell.
-    pub from_token_address: H160,
-    /// Contract address of a token to buy.
-    pub to_token_address: H160,
     /// Amount of a token to sell, set in smallest divisible unit.
     pub amount: U256,
```
nlordell, line 90: This can be moved to the `Common` struct right?

nlordell (commented):
Looks great! Just a couple questions then I think its good to merge.

MartinquaXD:
Moved `amount` into common parameters and added comments for when our defaults overwrite 1Inch defaults.

crates/shared/src/oneinch_api.rs
```diff
@@ -483,17 +539,19 @@ mod tests {
...
             slippage: Slippage::percentage_from_basis_points(50).unwrap(),
-            protocols: None,
             disable_estimate: None,
-            complexity_level: None,
-            gas_limit: None,
-            main_route_parts: None,
-            parts: None,
+            common: QuoteAndSwapCommonOptions {
+                from_token_address: addr!("EeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE"),
+                to_token_address: addr!("111111111117dc0aa78b770fa6a738034120c302"),
```
bh2smith, line 547: I like this new addr! macro. Way better looking than `H160::from(n.into())`!

crates/shared/src/oneinch_api.rs
```diff
@@ -512,17 +570,15 @@ mod tests {
...
-            protocols: Some(vec!["WETH".to_string(), "UNISWAP_V3".to_string()]),
             disable_estimate: Some(true),
-            complexity_level: Some(Amount::new(1).unwrap()),
-            gas_limit: Some(133700),
-            main_route_parts: Some(Amount::new(28).unwrap()),
-            parts: Some(Amount::new(42).unwrap()),
+            common: QuoteAndSwapCommonOptions::with_default_options(
+                addr!("EeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE"),
+                addr!("111111111117dc0aa78b770fa6a738034120c302"),
+                Some(vec!["WETH".to_string(), "UNISWAP_V3".to_string()]),
```
bh2smith, line 579: Having these protocol options as raw strings seems a bit prone to error. Shouldn't we have an enum somewhere with explicit supported protocols that we use instead?

bh2smith (approved):
Looks pretty good.

crates/shared/src/oneinch_api.rs
MartinquaXD: Besides testing (where everything is mocked anyway) we would only ever "instantiate" a protocol ourselves when a user passes a protocol to `solver` or `orderbook` with `--disabled_one_inch_protocols`.
I think introducing an enum for that might be a bit overkill. I fear that it would add too much complexity for the limited benefit it gives us. But asserting that the values passed to `--disabled_one_inch_protocols` are actual protocols and then keep on using the raw strings would already help while being a minimal change.

MartinquaXD: Should be something for another PR, though. To not blow up this one.

nlordell: > Shouldn't we have an enum

My issue with using an `enum` is that these are arbitrarily defined by 1Inch. So if they add a new protocol or change a protocol name, then we need to do a code change (instead of a configuration change) to adjust for it.

nlordell (approved):
Looks great!

crates/shared/src/oneinch_api.rs
MartinquaXD: Sending 1Inch protocols which don't exist, leads to internal server errors.
So I think we could fetch all supported protocols on startup and then either:
1) terminate if we find an unsupported protocol
2) filter out unsupported tokens and print a waning

MartinquaXD: I think this PR will cause major conflicts with #1578. I will merge this PR now to make progress on the other one, but will keep this issue in mind for a future PR.

PR #1578, opened 2022-01-18
Update 1Inch REST API to v4.1

### LINKED ISSUE #1560: Update API endpoints of OneInchClientImpl to v4.1
`OneInchClientImpl` mostly relies on the older v3.0 API. This is a bit worrying because the endpoints seem to be no longer documented anywhere.
Most of the endpoints we currently also exist under the `v4.1` API. Some endpoints now support different parameters, though.
It seems like the functionality of `/v3.0/protocols` can now be found under `/v4.1/1/liquidity-sources`.
Since the endpoints didn't change too much updating them shouldn't be too much work.

MartinquaXD:
Fixes #1560

We have been using 1Inch API v3.0 for quite a while. There isn't even any documentation online available anymore.
Upgrading to v4.1 makes it easier to reason about changes in the future and also offers some new parameters to configure quotes and swaps.
Note that  I'm not able to find any documentation about v4.1 specifically but I got confirmation from the 1Inch team that v4.0 and v4.1 are equivalent except that v4.1 is using EIP-1559 gas prices instead of the old gas prices.

The `/swap` endpoint got some new parameters and `/protocols` became `/liquidity-sources`.

### Test Plan
Update existing tests
Added new test parsing the response from `/liquidity-sources`

### Release notes
Use 1Inch API v4.1 instead of v3.0

### DIFF
--- a/crates/shared/src/oneinch_api.rs
+++ b/crates/shared/src/oneinch_api.rs
@@ -42,7 +42,7 @@ impl<const MIN: usize, const MAX: usize> Display for Amount<MIN, MAX> {
 }
 
 #[derive(Debug, Clone)]
-pub struct QuoteAndSwapCommonOptions {
+pub struct SellOrderQuoteQuery {
     /// Contract address of a token to sell.
     pub from_token_address: H160,
     /// Contract address of a token to buy.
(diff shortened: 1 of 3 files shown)


crates/solver/src/solver/oneinch_solver.rs
```diff
@@ -114,6 +114,14 @@ impl OneInchSolver {
             // Use only 3 main route for cheaper trades.
             main_route_parts: Some(Amount::new(3).unwrap()),
             parts: Some(Amount::new(3).unwrap()),
+            dest_receiver: None,
+            referrer_address: None,
```
josojo, line 118: For the other apis, we used our settlement contract address. But this can/should be done in another PR.

crates/shared/src/oneinch_api.rs
```diff
@@ -179,12 +181,33 @@ pub struct SwapQuery {
     pub main_route_parts: Option<Amount<1, 50>>,
     /// Limit maximum number of parts each main route part can be split into.
     pub parts: Option<Amount<1, 100>>,
+    /// Receiver of destination currency. default: fromAddress
+    pub dest_receiver: Option<H160>,
```
josojo, line 185: Nit: Some variables will never be used, e.g. dest_receiver will always be the fromAddress. 
Hence, in order to keep code shorter, it might make sense to exclude these struct fields.
(I really don't have a strong opinion here, and feel free to keep it as it is.)

crates/shared/src/oneinch_api.rs
```diff
@@ -412,7 +483,7 @@ impl ProtocolCache {
         ))))
     }
 
-    pub async fn get_all_protocols(&self, api: &dyn OneInchClient) -> Result<Vec<String>> {
+    pub async fn get_all_protocols(&self, api: &dyn OneInchClient) -> Result<Vec<LiquidiySource>> {
```
josojo, line 486: Not sure whether it's worth the effort, but you could rename the functions /structs now : `get_all_liquidity_sources` or `LiquiditySourceCache` are two examples. I think then it would be easier to maintain in the future.

josojo (commented):
Nice. Looks already quite promising.

crates/shared/src/oneinch_api.rs
MartinquaXD: Great feedback! I'm a bit conflicted about this, though.
1Inch's naming scheme seems inconsistent so I believe whatever we end up naming the functions and types will never by 100% consistent as well.
For example the `/liquidity-sources` endpoint returns a list of `protocols`. But the items of this list are not identical to items in the `protocols` field of the `/swap` and `/quote` endpoints.
Right now the term `protocol` seems to be more common in the 1Inch language than `liquidity source`.
Maybe renaming `LiquiditySource -> Protocol` and `Protocol -> TradeRouteSegment` (or something similar) could be an option. This way the 1Inch naming inconsistencies don't leak into the user API.
Then the user will not have to wonder how they can convert a `LiquiditySource` to a thing they can pass as a `protocol` argument, when they are basically the same thing.

MartinquaXD:
Note that `SellOrderQuoteQuery` no longer has any unique fields. Technically we no longer need `SellOrderQuoteQuery` but changing the APIs to expect `QuoteAndSwapCommonOptions` instead of it feels weird. Not sure about that.

nlordell:
> Note that SellOrderQuoteQuery no longer has any unique fields. Technically we no longer need SellOrderQuoteQuery but changing the APIs to expect QuoteAndSwapCommonOptions instead of it feels weird. Not sure about that.

Maybe just keep two types:
```rust
pub struct QuoteQuery {
  // What we have in QuoteAndSwapCommonOptions now
}
pub struct SwapQuery {
  thing: Things,
  // Common options with the quote query.
  quote: QuoteQuery,
}
```

crates/shared/src/oneinch_api.rs
nlordell: Since this is implemented based on the 1Inch API documentation, I think its fine to include it here. No strong preference either way.

crates/shared/src/oneinch_api.rs
nlordell: I'd be slightly in favour of `get_liquidity_sources` just because that is what the URL is called (it makes it easier for me to match the 1Inch API client function to the endpoint it corresponds to).

As for the protocol names, I would call the struct `ProtocolInfo` just because its in a the `protocols` field of the returned JSON, but isn't a protocol name (as passed into the route). Also, we can rename the current `Protocol` struct to `ProtocolRouteSegment` or something (again, just because of my _weak_ argument that the field in the JSON block is `protocols`). For the protocol names, I really don't have a strong opinion. I'm also happy with your suggestions as well!

nlordell (approved):
LGTM,

There is some open discussion around struct and function naming, but I don't feel particularly strongly about any suggestion and am OK with merging regardless of the outcome.

vkgnosis (approved):
Code LGTM. Also no strong feeling about naming so don't want to bikeshed it too much.

PR #1580, opened 2022-01-18
Allow coverage CI to fail

vkgnosis:
Currently we get annoying notifications because the coverage part of CI is failing. The cause is a regression in nightly rust that we cannot fix.
This PR allows the coverage job to fail because it is not important for us, just nice to have.

### Test Plan
see that error notification goes away after this PR


vkgnosis:
Sadly this doesn't work. Still shows the red x.

nlordell:
We can just pin an older nightly version.

PR #1585, opened 2022-01-19
dont fail if one of the gas price estimators fail to initialize

sunce86:
We want to skip sending fail back to the caller in case one of the gas price estimators failed to initialize.

### DIFF
--- a/crates/shared/src/gas_price_estimation.rs
+++ b/crates/shared/src/gas_price_estimation.rs
@@ -73,7 +73,9 @@ pub async fn create_priority_estimator(
                 } else {
                     http::header::HeaderMap::new()
                 };
-                estimators.push(Box::new(BlockNative::new(client.clone(), headers).await?))
+                if let Ok(estimator) = BlockNative::new(client.clone(), headers).await {
+                    estimators.push(Box::new(estimator))
+                }
             }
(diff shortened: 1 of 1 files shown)


vkgnosis (commented):
The problem this way is that we silently drop the estimators and never use them if they fail once at start up.

It should either error or continue to try to initialize them later. This is related to the actual estimator implementations because for some of them it is a design decision whether `new` should fail if it cannot get the first datapoint or if it should fail during estimation later.

Overall I like hard failure because it shows us immediately that something went wrong.

sunce86:
I do agree with the "silently dropping" part, however I am not sure if being noticed that something is wrong is worth our service going down. I would prefer to skip it like this PR suggests, but build an alert that would notify us in the channel so we could manually react. This would be a middleground between your idea of trying to initialize later, which is a bit more sophisticated and technically challenging. 
Not sure...

vkgnosis:
Let's at least at `tracing::error` with the error then if they fail to initialize.

nlordell:
Agree that we should log errors in these cases - since it means we are not running with the configured parameters.

Since the error would cause an alert at deployment time (so when we are applying new prod configurations) I think its reasonable to expect it not to be easily missed.

vkgnosis (approved):

PR #1587, opened 2022-01-20
Disable balance check for presign

### LINKED ISSUE #1545: Allow Pre-Sign Orders to be Created Without Allowance and/or Balance
Balance and allowance checks were used as a form of DoS attack prevention. Since we aren't protected for pre-sign orders anyway, we can remove this restriction for them enabling Yearn to use CowSwap for harvesting swaps.

nlordell:
Fixes #1545 
 
This PR disables balance and allowance check for pre-sign orders.

The rational is that this allows SCs to place orders bundled with other transactions that either produce the required balance or set the allowance. This would, for example, allow a Gnosis Safe to bundle the pre-signature transaction with a WETH wrap and WETH approval to the vault relayer contract. Additionally, this enables the Yearn-team integration mentioned in the issue ☝️ 

### Test Plan

New unit test.

### DIFF
--- a/crates/orderbook/src/api/order_validation.rs
+++ b/crates/orderbook/src/api/order_validation.rs
@@ -9,6 +9,7 @@ use model::{
     order::{
         BuyTokenDestination, Order, OrderCreation, OrderKind, SellTokenSource, BUY_ETH_ADDRESS,
     },
+    signature::SigningScheme,
     DomainSeparator,
 };
 use shared::{bad_token::BadTokenDetecting, web3_traits::CodeFetching};
@@ -376,6 +377,11 @@ impl OrderValidating for OrderValidator {
             .await
         {
             Ok(_) => Ok((order, unsubsidized_fee)),
+            Err(TransferSimulationError::InsufficientAllowance)
+                if order.order_creation.signature.scheme() == SigningScheme::PreSign =>
+            {
+                Ok((order, unsubsidized_fee))
+            }
             Err(err) => match err {
                 TransferSimulationError::InsufficientAllowance => {
                     Err(ValidationError::InsufficientAllowance)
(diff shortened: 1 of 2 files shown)


vkgnosis (approved):

crates/orderbook/src/api/order_validation.rs
```diff
@@ -815,4 +829,72 @@ mod tests {
...
+                    validator
+                        .validate_and_construct_order(
+                            order.clone(),
+                            None,
+                            &Default::default(),
+                            Default::default()
+                        )
+                        .await,
+                    Err(ValidationError::$err)
+                ));
```
MartinquaXD, line 878: nit: Because SigningScheme is an enum with very few variants, it could be nice to assert an error for every SigningScheme except PreSign.

crates/orderbook/src/api/order_validation.rs
```diff
@@ -376,6 +377,19 @@ impl OrderValidating for OrderValidator {
...
+                TransferSimulationError::InsufficientAllowance
+                | TransferSimulationError::InsufficientBalance,
+            ) if order.order_creation.signature.scheme() == SigningScheme::PreSign => {
+                // We have an exception for pre-sign orders where they do not
+                // require sufficient balance or allowance. The idea, is that
+                // this allows smart contracts to place orders bundled with
+                // other transactions that either produce the required balance
+                // or set the allowance. This would, for example, allow a Gnosis
+                // Safe to bundle the pre-signature transaction with a WETH wrap
+                // and WETH approval to the vault relayer contract.
```
MartinquaXD, line 390: Very helpful comment! Without it this exception to the rule could seem arbitrary.

MartinquaXD (approved):
Looks good!
The code change is easy to understand and the test is very well structured. Just one little nit but I think you could merge the PR regardless.

PR #1589, opened 2022-01-20
Pin nightly to older version to re-enable coverage

nlordell:
This PR pins nightly to `2022-01-14` to fix the coverage report.

It appears to be an issue with the nightly compiler (at least that is what is suggested in the `cargo-llvm-cov` project issues page). I created an issue so we don't forget to switch( back eventually #1588 (this also contains a link to the issue in the `cargo-llvm-cov` project for reference).

This PR also introduces a small unrelated change: the [recommended way to install `cargo-llvm-cov` is to use a dedicated Github action](https://github.com/taiki-e/cargo-llvm-cov#continuous-integration), so this PR also changes the workflow to use it instead of the manual install it was doing previously.

### Test Plan

CI

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -108,11 +108,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install Rust
-        run: rustup toolchain install nightly-2022-01-13 --component llvm-tools-preview
+        run: rustup toolchain install nightly-2022-01-14 --component llvm-tools-preview
       - name: Install cargo-llvm-cov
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):
Looks like this specific nightly hasn't helped? Another possibility is that a code change caused it to start failing on all nightlies so we can't fix it this way.
Anyway, I'm approving assuming we will repeat until we find a working nightly.

nlordell:
> Looks like this specific nightly hasn't helped?

Issue with the workflow I think. Its complaining that `nightly` isn't installed.

MartinquaXD (approved):

Issue #1583, opened 2022-01-18
Error on transaction submission: "replacement transaction underpriced"


sunce86:
@vkgnosis reported error with message "replacement transaction underpriced".
Seems like this message also appears in one special case when we:

1. Submit tx
2. Try to cancel tx for whatever reason
3. Submit new tx with the same nonce in the next submission loop, while (2) is not mined yet.

In order to fix this, we need to revert and use the `recover_gas_price_from_pending_transaction` function that was deleted here:
https://github.com/gnosis/gp-v2-services/pull/1476/files#diff-9537ac92f7f737a1932e307020d27e0b620aa671417eebc67c78eb7560564af0L88

PR #1590, opened 2022-01-20
cancel transaction using the latest submitted tx + bump

sunce86:
This PR is inspired by the @fleupold comment: https://github.com/gnosis/gp-v2-services/pull/1564#discussion_r784772357

In case of deadline, we want to cancel the last submitted tx by using its gas price + 12.5% bump.
After extensive analysis of geth and erigon clients, I made sure this is safe to do (base fee is ignored on tx pool import). This change should reduce the number of "replacement transaction underpriced" error messages, context here: https://github.com/gnosis/gp-v2-services/issues/1583

Additionally, this PR removes `let mut previous_tx: Option<(EstimatedGasPrice, TransactionHandle)> = None;` since its not needed anymore, all data is already available in variable `transactions: &mut Vec<(TransactionHandle, EstimatedGasPrice)>`

### DIFF
--- a/crates/solver/src/settlement_submission/submitter.rs
+++ b/crates/solver/src/settlement_submission/submitter.rs
@@ -188,11 +188,10 @@ impl<'a> Submitter<'a> {
             _ = deadline_future.fuse() => {
                 tracing::info!("stopping submission because deadline has been reached. cancelling last submitted transaction...");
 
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):

nlordell (approved):
LGTM

PR #1591, opened 2022-01-20
Add Metrics to Competition Price Estimator

### LINKED ISSUE #1586: Add metrics to the `CompetitionPriceEstimation`
We should record metrics in the competition price estimator implementation to keep track of how many times each price estimator wins.

nlordell:
Fixes #1586.

This PR adds metrics or the competition price estimation to show the winner.

I decided to include a label for the price estimator name **and** for the query kind (so `buy` vs `sell`). The decision was driven by the fact that the additional label only has two possible values (so we have twice as many counters now - nothing crazy). Furthermore, it could show how different price estimators fare on buy vs sell orders (which are fundamentally different). This was also driven by the fact that some price estimators (1Inch for example) don't 

### Test Plan

Run the orderbook and checkout the metrics!
```
% curl -s http://localhost:9586/metrics | grep gp_v2_api_competition_price_estimator_queries_won
# HELP gp_v2_api_competition_price_estimator_queries_won Number of wins for a particular price estimator and order kind.
# TYPE gp_v2_api_competition_price_estimator_queries_won counter
gp_v2_api_competition_price_estimator_queries_won{estimator_type="Baseline",order_kind="buy"} 2
gp_v2_api_competition_price_estimator_queries_won{estimator_type="OneInch",order_kind="sell"} 4
```

### Release notes

Additional metrics have been added.

### DIFF
--- a/crates/shared/src/price_estimation/competition.rs
+++ b/crates/shared/src/price_estimation/competition.rs
@@ -1,6 +1,8 @@
 use super::{Estimate, PriceEstimating, PriceEstimationError, Query};
+use crate::metrics;
 use anyhow::{anyhow, Result};
 use futures::future;
+use model::order::OrderKind;
 use num::BigRational;
 use std::cmp;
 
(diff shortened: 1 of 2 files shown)


vkgnosis (approved):

crates/shared/src/price_estimation/competition.rs
```diff
@@ -66,11 +74,22 @@ impl PriceEstimating for CompetitionPriceEstimator {
...
+impl EstimateData<'_> {
+    /// Returns a label for the estimate's query kind.
+    pub fn kind_label(&self) -> &'static str {
+        match self.kind {
+            OrderKind::Buy => "buy",
+            OrderKind::Sell => "sell",
+        }
+    }
+}
+
```
MartinquaXD, line 92: nit: Could make sense to implement something like `as_str` on `OrderKind` instead.

MartinquaXD (approved):

crates/shared/src/price_estimation/competition.rs
nlordell: I ended up adding a `OrderKind::label` method. I didn't go with `as_str` just because I wanted the method to reflect that it was meant to be used for metrics labels.

PR #1593, opened 2022-01-20
recover pending transaction from mempool

### LINKED ISSUE #1583: Error on transaction submission: "replacement transaction underpriced"

sunce86:
This PR fixes https://github.com/gnosis/gp-v2-services/issues/1583

Notice that this is enabled only for `CustomNodes` strategy at the moment. Eden and Flashbots can't benefit from this change since transactions sent through them are not seen in the public mempool. Flashbots does say they send transactions to public mempool if tx.gas < 41000, however I wasn't able to confirm this.

Additionally, with previous PR https://github.com/gnosis/gp-v2-services/pull/1590 we should receive less error messages of this type. Anyway, in this PR we suppress this type of message by handling it in the `Submitter`. Will continue to monitor these messages in logs.

### DIFF
--- a/crates/solver/src/settlement_submission.rs
+++ b/crates/solver/src/settlement_submission.rs
@@ -80,15 +80,16 @@ impl SolutionSubmitter {
                 .iter()
                 .map(|strategy| {
                     async {
-                        match &*strategy {
+                        let try_to_recover_gas_price = match &*strategy {
                             TransactionStrategy::Eden(_) | TransactionStrategy::Flashbots(_) => {
                                 if !matches!(account, Account::Offline(..)) {
                                     return Err(SubmissionError::from(anyhow!(
                                         "Submission to private network requires offline account for signing"
                                     )));
                                 }
+                                false
                             }
-                            TransactionStrategy::CustomNodes(_) => {}
+                            TransactionStrategy::CustomNodes(_) => true,
                             TransactionStrategy::DryRun => unreachable!(),
                         };
 
@@ -98,6 +99,7 @@ impl SolutionSubmitter {
                             gas_estimate,
                             deadline: Some(Instant::now() + self.max_confirm_time),
                             retry_interval: self.retry_interval,
+                            try_to_recover_gas_price,
                         };
                         let gas_price_estimator = SubmitterGasPriceEstimator {
                             inner: self.gas_price_estimator.as_ref(),
--- a/crates/solver/src/settlement_submission.rs
+++ b/crates/solver/src/settlement_submission.rs
@@ -80,16 +80,15 @@ impl SolutionSubmitter {
                 .iter()
                 .map(|strategy| {
                     async {
-                        let try_to_recover_gas_price = match &*strategy {
+                        match &*strategy {
                             TransactionStrategy::Eden(_) | TransactionStrategy::Flashbots(_) => {
                                 if !matches!(account, Account::Offline(..)) {
                                     return Err(SubmissionError::from(anyhow!(
                                         "Submission to private network requires offline account for signing"
                                     )));
                                 }
-                                false
                             }
-                            TransactionStrategy::CustomNodes(_) => true,
+                            TransactionStrategy::CustomNodes(_) => {},
                             TransactionStrategy::DryRun => unreachable!(),
                         };
 
@@ -99,7 +98,6 @@ impl SolutionSubmitter {
                             gas_estimate,
                             deadline: Some(Instant::now() + self.max_confirm_time),
                             retry_interval: self.retry_interval,
-                            try_to_recover_gas_price,
                         };
                         let gas_price_estimator = SubmitterGasPriceEstimator {
                             inner: self.gas_price_estimator.as_ref(),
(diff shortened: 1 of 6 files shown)


crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -319,17 +334,16 @@ impl<'a> Submitter<'a> {
...
+            if let Some(previous_gas_price) = transactions
+                .last()
+                .map(|(_, previous_gas_price)| previous_gas_price)
+                .or_else(|| pending_gas_price.as_ref())
+            {
                 let previous_gas_price = previous_gas_price.bump(1.125).ceil();
-                if gas_price.cap() > previous_gas_price.cap()
-                    && gas_price.tip() > previous_gas_price.tip()
+                if gas_price.tip() < previous_gas_price.tip()
+                    || gas_price.cap() < previous_gas_price.cap()
```
MartinquaXD, line 345: I think this is a subtle change in behavior. Shouldn't it be `gas_price.tip() <= previous_gas_price.tip()` and same for `.cap()` to invert the if?

Also I'm not sure about how the gas price has to change in order to warrant resubmitting the order with a higher gas price. Do the `.tip()` *AND* `.cap()` of the current gas price need to be higher than what we submitted previously? Intuitively I would think it's enough if either value exceeds what we previously submitted.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -319,17 +334,16 @@ impl<'a> Submitter<'a> {
...
             }
 
-            // if gas price has increased cancel old transaction so new transaction could be submitted.
-
-            if let Some((previous_tx, previous_gas_price)) = transactions.last() {
+            // if gas price has not increased enough, skip submitting the transaction.
+            if let Some(previous_gas_price) = transactions
+                .last()
+                .map(|(_, previous_gas_price)| previous_gas_price)
+                .or_else(|| pending_gas_price.as_ref())
```
MartinquaXD, line 341: Is it correct that we only ever use `pending_gas_price` on the first iteration of the loop? On subsequent iterations `transactions.last()` would probably yield a value from which we can take the gas price, right?

Also since it's possible to return from this function before we reach this line, could we do something like `.or_else(calculate_pending_gas_price)` instead of `.or_else(|| pre_calculated_pending_gas_price)`. This way we could delay a possibly (I don't know) expensive operation until we really need it.
Not sure if it's worth the effort though, given the fact that the reasons to return before we reach this line seem quite exotic.

MartinquaXD (commented):
I think this code is somewhat hard to reason about because correctly cancelling or resubmitting transactions has to deal with some subtleties.
Not sure that I know this topic well enough to give an educated feedback but here are my 2 cents regardless.

crates/solver/src/settlement_submission.rs
```diff
@@ -98,6 +99,7 @@ impl SolutionSubmitter {
                             gas_estimate,
                             deadline: Some(Instant::now() + self.max_confirm_time),
                             retry_interval: self.retry_interval,
+                            try_to_recover_gas_price,
```
nlordell, line 102: Another alternative instead of specifying whether or not we should try to recover gas prices would be to always try, and silently fail if its not there.

Alternatively, it could be possible for a future private network API to have a request for recovering a users transaction requiring some PK signature (otherwise, what is the point to private transactions 🤣). So the other options would be to add a new method `TransactionSubmitting::recover_pending_transaction()` where for now the `CustomNodes` implementation uses the mem-pool, but `Flashbots` and `Eden` just return `None`.

nlordell (commented):
Nice change!

Added a suggestion to make the recovery of pending transactions a `TransactionSubmitting` method instead of being behind a flag. Let me know what you think about it.

crates/solver/src/settlement_submission/submitter.rs
sunce86: It is possible to use `pending_gas_price` more then just in first iteration, because we have a `continue` between usage of `pending_gas_price` and submitting the transaction (after which `transactions.last()` yields something).

MartinquaXD: Ahh, I see. In that case trying to delay computing the pending gas cost and caching that result doesn't seem to be worth the added complexity.

crates/solver/src/settlement_submission.rs
sunce86: Refactored according to the above explanation.

nlordell (approved):
LGTM!

PR #1595, opened 2022-01-21
Update gp-gas-estimation

vkgnosis:
One more update to fix compatibility with our fallback nodes.

### Test Plan

CI


sunce86 (approved):
I don't see this version here: https://github.com/gnosis/gp-gas-estimation/releases maybe we should add them

PR #1596, opened 2022-01-24
Drop min_order_size_one_inch CLI argument from solver

### LINKED ISSUE #1529: Add 1Inch Price Estimator

MartinquaXD:
Fixes #1529

Because 1Inch is now represented in the price estimation and we also have enough solution competition, costly 1Inch settlements should simply loose based on their objective value. We no longer need to manually limit `OneInchSolver` to settlements with a total trading volume bigger than a user specified value.

This PR resolves the final task of the linked issue.

### Test Plan
No tests

### DIFF
--- a/crates/solver/src/main.rs
+++ b/crates/solver/src/main.rs
@@ -1,7 +1,7 @@
 use anyhow::anyhow;
 use clap::{ArgEnum, Parser};
 use contracts::{BalancerV2Vault, IUniswapLikeRouter, WETH9};
-use ethcontract::{Account, PrivateKey, H160, U256};
+use ethcontract::{Account, PrivateKey, H160};
 use reqwest::Url;
 use shared::{
     baseline_solver::BaseTokens,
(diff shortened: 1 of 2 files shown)


MartinquaXD:
As was discussed in the planning meeting, it makes sense to first set `min_order_size_one_inch` to 0 in staging. This way we can evaluate the effect of this PR without changing the code base. If the test run causes no problems, this PR can be considered for a merge.

vkgnosis (approved):

nlordell (approved):

PR #1597, opened 2022-01-24
Revert 1Inch API update

MartinquaXD:
Updating the 1Inch API lead to some parsing errors. Quickly fixing those parsing errors lead to some settlement simulation failures. It's not yet clear to me what's causing those.
Let's revert the API update until all problems have been ironed out because I'm not sure how long it would take to find the cause.


sunce86 (approved):
I just did a sanity check of changed files, I assume you just reverted your commits.

MartinquaXD:
> I just did a sanity check of changed files, I assume you just reverted your commits.

Yes.

PR #1598, opened 2022-01-24
Detecting failing transactions with receipt.status field.

### LINKED ISSUE #1570: Add alert for failing transactions
Transaction submission could end in mined transaction, however, that does not mean it is successful.
Mined transaction could be one of the three:

1. Successful transaction
2. Cancelled transaction - noop transaction
3. Failing transaction

We need to handle failing transactions through `SettlementSubmissionOutcome` error type, so we can build an alert around that.

sunce86:
This PR fixes https://github.com/gnosis/gp-v2-services/issues/1570

Two important things are added:
1. Up until now, we did not detect failing transaction - in case `Submitter` found a mined transaction, `Submitter` returned `Ok(receipt)` and that was interpreted as `SettlementSubmissionOutcome::Success` in solver metrics. Now, `Submitter` detects failing transactions by inspecting `TransactionReceipt::status` field of the mined transaction.
2. Detection of cancelled transactions. In order to implement this, `noop` cancel transactions also need to be pushed into the `mut Transactions` variable, so their hash could be examined at the end of the `Submitter` logic.

### DIFF
--- a/crates/solver/src/metrics.rs
+++ b/crates/solver/src/metrics.rs
@@ -41,18 +41,14 @@ pub enum SolverRunOutcome {
 pub enum SettlementSubmissionOutcome {
     /// A settlement transaction was mined and included on the blockchain.
     Success,
-    /// A settlement transaction was mined and included on the blockchain with a failed status.
-    MinedFailed,
-    /// A transaction reverted.
+    /// A settlement transaction was mined and included on the blockchain but reverted.
     Revert,
+    /// A transaction reverted in the simulation stage.
+    SimulationRevert,
     /// Submission timed-out while waiting for the transaction to get mined.
     Timeout,
-    /// Transaction sucessfully cancelled after revert or timeout
+    /// Transaction sucessfully cancelled after simulation revert or timeout
     Cancel,
-    /// A transaction failed to be submitted or, in the case of private network
-    /// submission, the blockchain state changed and the transaction is no
-    /// longer valid.
-    Failure,
 }
 
 pub trait SolverMetrics: Send + Sync {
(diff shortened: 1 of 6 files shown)


crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -421,6 +429,20 @@ impl<'a> Submitter<'a> {
     }
 }
 
+fn status(receipt: TransactionReceipt) -> Result<TransactionReceipt, SubmissionError> {
+    if let Some(status) = receipt.status {
+        if status == U64::zero() {
+            // failing transaction
+            return Err(SubmissionError::Revert(None));
```
nlordell, line 436: As a future improvement we could try and detect the revert message by re-simulating the transaction at the same block and transaction index. Not needed for this PR though.

nlordell (approved):
LGTM.

vkgnosis (approved):

sunce86:
Added new value `MinedFailed` since `Revert` is already reserved for simulation failure, therefore we cant build an alert to detect only failing transactions. Re-request review incomming...

crates/solver/src/metrics.rs
```diff
@@ -41,10 +41,14 @@ pub enum SolverRunOutcome {
 pub enum SettlementSubmissionOutcome {
     /// A settlement transaction was mined and included on the blockchain.
     Success,
+    /// A settlement transaction was mined and included on the blockchain with a failed status.
+    MinedFailed,
```
nlordell, line 45: very nit-picky naming-nit: To me, its more natural for `Revert` to indicated that a transaction got mined and reverted. Originally, I had intended for `Failure` to be used for this:

> A transaction failed to be submitted or, in the case of private network submission, the blockchain state changed and the transaction is no longer valid.

That being said maybe we should differentiate between "failed to submit" and "transaction started simulating failure" and have a `SimulationRevert` just for simulations?

nlordell: With that in mind, we should probably update the documentation for the `Failure` variant as it doesn't seem accurate to me.

nlordell (approved):
Small naming nit (very nit-picky, feel free to ignore), otherwise looks good.

PR #1600, opened 2022-01-24
Take time to solvability into account for alerter

### LINKED ISSUE #1434: Fix alerter false positive if single order just became settlable
The alerter is supposed to alert when an order is settlable but no settlement happens for X minutes. Currently this can trigger when a sole order is not settlable, then the on chain price moves and it becomes settlable. If there were no other orders then it is not surprising that no settlement happened for X minutes so we should wait for this order to have been settlable for those X minutes.
We could fix that by keeping track of the time orders became solvable per order.

vkgnosis:
This eliminates some false positives where an order is created while
being unsolvable and becomes solvable later. In that case the current
alerter will alert even though the driver didn't have time to settle
that order yet.
To fix this we move from checking the order creation time to checking
the time the order became solvable at.

Fixes https://github.com/gnosis/gp-v2-services/issues/1434 .

### Test Plan

Ran manually and CI.

### DIFF
--- a/crates/alerter/src/main.rs
+++ b/crates/alerter/src/main.rs
@@ -10,7 +10,7 @@ use model::{
 };
 use primitive_types::{H160, U256};
 use reqwest::Client;
-use std::time::{Duration, Instant, SystemTime};
+use std::time::{Duration, Instant};
 use structopt::StructOpt;
 use url::Url;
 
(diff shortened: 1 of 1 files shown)


crates/alerter/src/main.rs
```diff
@@ -214,20 +218,22 @@ impl Alerter {
         ) {
             return Ok(());
         }
-        for order in &self.open_orders {
-            if !self.order_has_minimum_age(order) {
-                continue;
-            }
-
-            if self
+        for i in 0..self.open_orders.len() {
```
nlordell, line 221: Does Rust 2021 borrow checker let you do:
```
for (order, time) in self.open_orders.iter_mut() {
    // ...
}
```

And figure out that only `open_orders` is borrowed mutably (so you can do stuff with other fields in `self`)? Might make the code slightly more succinct (less array indexing).

At least I have some vague memory that this changed with the new 2021 edition.

Fine as is though.

nlordell (approved):
Nice! Thanks for following up on this change. Has been in the backlog for a while.

crates/alerter/src/main.rs
vkgnosis: I had that first but didn't work.

PR #1601, opened 2022-01-25
return early from transaction submission if reverted or canceled

sunce86:
In case of `Revert` or `Cancel` error type, we definitely know what happened -> transaction is mined. There is no need to wait for other futures.

This prevents racing when, for example, we have two submissions running at the same time, first one finishes with `Revert` error while other detects change in nonce, but does not detect mined transaction because it has a different set of submitted transaction hashes, therefore it returns `SimulationRevert`. In this case, we want `Revert` as a final error.


nlordell (approved):
Gains a little at a time.

PR #1603, opened 2022-01-26
Refactor solvable orders

sunce86:
We had multiple `SolvableOrders` structs defined in our code, with small differences.
Now, we have `SolvableOrders` and `CachedSolvableOrders` and both defined in `model` crate, and used by `orderbook` and `solver`.

_`CachedSolvableOrders` is used only by `orderbook`, not sure if I should leave it in `model` or move to `orderbook`._


vkgnosis (commented):
I don't agree with this change. The main intention of the model crate is contain types that are used in the api. The new struct CachedSolvableOrders does not appear in the api so it does not fit there. It is not in general a problem to have multiple slightly different definitions of SolvableOrders for different purposes.

nlordell:
I found having two identical `SolvableOrders` types could be slightly confusing (hard to tell which one you should use where) and I like the rename to `CachedSolvableOrders`.

I agree that the `models` crate should stick with API model types. Maybe we should move `CachedSolvableOrders` to the `orderbook` crate since it is an implementation detail of the `orderbook` API.

sunce86:
`CachedSolvableOrders` moved to `orderbook`. Still not completely sure if we should accept this PR, no strong opinion.

nlordell:
@sunce86 I'm closing this PR here. Feel free to re-open in cowprotocol/services

PR #1604, opened 2022-01-28
Replace structopt with clap

vkgnosis:
Clap has absorbed and succeeded structopt in the newly released v3.

Based on my testing this should be backward compatible but it is easy to accidentally make a mistake that will only show up once this hits staging so I will merge this earliest after Kaffeekraenzchen and stand by to unbreak if needed.

### Test Plan

CI and manual test

### DIFF
--- a/crates/shared/src/gas_price_estimation.rs
+++ b/crates/shared/src/gas_price_estimation.rs
@@ -7,18 +7,16 @@ use gas_estimation::{
 };
 use serde::de::DeserializeOwned;
 use std::sync::{Arc, Mutex};
-use structopt::clap::arg_enum;
 
-arg_enum! {
-    #[derive(Debug)]
-    pub enum GasEstimatorType {
-        EthGasStation,
-        GasNow,
-        GnosisSafe,
-        Web3,
-        BlockNative,
-        Native,
-    }
+#[derive(Copy, Clone, Debug, clap::ArgEnum)]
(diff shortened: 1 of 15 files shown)


crates/shared/src/gas_price_estimation.rs
```diff
@@ -7,18 +7,16 @@ use gas_estimation::{
...
-    pub enum GasEstimatorType {
-        EthGasStation,
-        GasNow,
-        GnosisSafe,
-        Web3,
-        BlockNative,
-        Native,
-    }
+#[derive(Copy, Clone, Debug, clap::ArgEnum)]
+#[clap(rename_all = "verbatim")]
```
vkgnosis, line 12: This is needed because by default clap will convert the enum values like eth-gas-station which would break our existing configuration.

nlordell:
> Based on my testing this should be backward compatible

I think they might do `mutliple_values/multiple_occurences` differently than `structopt`. At least that was my experience with it.

vkgnosis:
Based on my testing and documentation it should automatically enable multiple values and this works for Vec<T> and Option<Vec<T>>. There might be some edge cases like if you do --flag value1 --flag value2 to specify multiple values but we don't rely on that anywhere.

nlordell (approved):

MartinquaXD (approved):
LGTM.
I didn't find any breaking changes of clap 3.0 which apply to us.

Issue #1538, opened 2022-01-09
Specialized Settlement Submission Strategies


nlordell:
With the current work going into refactoring the settlement submission strategies, it would be great to add some specialized logic to our strategies to:
- For settlements without any trade interactions (WETH unwraps are OK to include) don't send the settlement to a private network
- For settlements only including small orders (so not enough to be attractive to MEV) over very deep pools (USDT-WETH for example - where the chance of price movement beyond the slippage is very very low) do not submit to a private network

sunce86:
Related to task 4 from https://github.com/gnosis/gp-v2-services/issues/1407

I would argue (for both) that we should substitute `do not submit to private network` with `submit to both private networks and public mempool`, because Flashbots give us also the protection against failing transaction that both of these situations would benefit from.

PR #1612, opened 2022-02-01
add possibility to disable strategies (copied)

sunce86:
Copy of https://github.com/gnosis/gp-v2-services/pull/1525 (tried to merge with main and resolve conflicts but got strange build errors)

This PR implements task 4 from: https://github.com/gnosis/gp-v2-services/issues/1407 and also https://github.com/gnosis/gp-v2-services/issues/1538

The point of this PR is to be able to select one or multiple transaction strategies dynamically, based on the different attributes of the Settlement. Right now, the only attribute of interest is if the Settlement is MEV Safe. Additional atributes could be taken into consideration in the future:
1. Disable Eden and CustomNodes if gas price > 500
2. Disable Eden if tx.gas > 1.5M (Eden rpc constraint)
3. ... etc

The idea is following:
1. All transaction strategies that are considered are received on program start up (`transaction_strategy` argument). I expect the default to be Flashbots + Eden + CustomNodes
2. In each submission loop, Settlement is checked if it is MEV Safe (theoretically needs to be checked in every loop, since Ethereum State changes block by block, therefore our Settlement could change from MEV Safe to MEV Unsafe, if for example, Settlement touches the liquidity pool and in the next block large portion of the liquidity is removed from the pool). Based on the result, CustomNodes strategy is enabled(if MEV Safe)/disabled(if not MEV Safe), while Flashbots and Eden are always enabled.

MEV Safe check is the hard part and it will be handled in the following PR.

### DIFF
--- a/crates/solver/src/settlement_submission.rs
+++ b/crates/solver/src/settlement_submission.rs
@@ -90,11 +90,7 @@ impl SolutionSubmitter {
                                     )));
                                 }
                             }
-                            TransactionStrategy::CustomNodes(_) => {
-                                if settlement.mev_extractable() {
-                                    return Err(SubmissionError::Disabled(DisabledReason::CustomNodesDisabledMevExtractable));
-                                }
-                            }
+                            TransactionStrategy::CustomNodes(_) => {}
                             TransactionStrategy::DryRun => unreachable!(),
                         };
 
--- a/crates/solver/src/settlement_submission.rs
+++ b/crates/solver/src/settlement_submission.rs
@@ -74,6 +74,8 @@ impl SolutionSubmitter {
             .iter()
             .any(|strategy| matches!(strategy, TransactionStrategy::DryRun));
 
+        let network_id = self.web3.net().version().await?;
+
         if is_dry_run {
             Ok(dry_run::log_settlement(account, &self.contract, settlement).await?)
         } else {
@@ -100,6 +102,7 @@ impl SolutionSubmitter {
                             gas_estimate,
                             deadline: Some(Instant::now() + self.max_confirm_time),
                             retry_interval: self.retry_interval,
+                            network_id: network_id.clone(),
                         };
                         let gas_price_estimator = SubmitterGasPriceEstimator {
                             inner: self.gas_price_estimator.as_ref(),
@@ -196,6 +199,12 @@ impl SubmissionError {
             SubmissionError::Other(_) => false,
             SubmissionError::Disabled(_) => false,
         }
+    }
+}
+
+impl From<web3::Error> for SubmissionError {
+    fn from(err: web3::Error) -> Self {
+        Self::Other(err.into())
     }
 }
 
--- a/crates/solver/src/settlement_submission.rs
+++ b/crates/solver/src/settlement_submission.rs
@@ -163,7 +163,7 @@ impl SubmissionError {
             Self::Revert => SettlementSubmissionOutcome::Revert,
             Self::Canceled => SettlementSubmissionOutcome::Cancel,
             Self::Disabled(_) => SettlementSubmissionOutcome::Disabled,
-            Self::Other(_) => SettlementSubmissionOutcome::SimulationRevert,
+            Self::Other(_) => SettlementSubmissionOutcome::Failed,
         }
     }
 
--- a/crates/solver/src/settlement_submission.rs
+++ b/crates/solver/src/settlement_submission.rs
@@ -16,7 +16,9 @@ use std::{
     sync::Arc,
     time::{Duration, Instant},
 };
-use submitter::{Submitter, SubmitterGasPriceEstimator, SubmitterParams, TransactionSubmitting};
+use submitter::{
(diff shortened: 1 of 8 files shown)


crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
```diff
@@ -118,4 +118,14 @@ impl TransactionSubmitting for CustomNodesApi {
...
+
+    fn submission_status(&self, settlement: &Settlement) -> SubmissionLoopStatus {
+        if !settlement.mev_safe() {
+            return SubmissionLoopStatus::Disabled(
+                DisabledReason::CustomNodesDisabledMevExtractable,
+            );
+        }
+
+        SubmissionLoopStatus::Enabled
+    }
```
MartinquaXD, line 130: This would effectively disable `TransactionStrategyArg::CustomNodes` until the follow-up PR is merged.
Depending on how long that is expected to take, I think it would be better to simply return `SubmissionLoopStatus::Enabled` here and leave a `TODO` comment to replace the implementation later on.
In that case it would also make sense to drop `mev_safe()` from this PR and only introduce it in the follow-up PR.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -60,6 +60,17 @@ impl From<anyhow::Error> for SubmitApiError {
...
 
+#[derive(Debug)]
+pub enum SubmissionLoopStatus {
+    Enabled,
+    Disabled(DisabledReason),
+}
+
+#[derive(Debug)]
+pub enum DisabledReason {
+    CustomNodesDisabledMevExtractable,
```
MartinquaXD, line 71: I think `MevExtractable` would be a better name here. It's an implementation detail that only `CustomNodes` suffers from this problem and `Disabled` is already in the enum name. Also all submission strategies have access to this value so it would theoretically be possible that an eden submitter has `CustomNodesDisabledMevExtractable` as its status which would be confusing.

MartinquaXD (commented):
Seems like a useful idea. 👍

crates/solver/src/settlement_submission/submitter.rs
sunce86: Yes, `MevExtractable` seems clearer.

MartinquaXD (approved):

crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
sunce86: Indeed, `mev_safe` is not relevant, but let's keep it for the idea to be more clear, I don't plan to merge this PR before the following PR for `mev_safe` is done, anyways.
`mev_safe` changed to `true` to avoid disabling `CustomNodes`.

crates/solver/src/settlement_submission.rs
```diff
@@ -155,6 +159,7 @@ impl SubmissionError {
             Self::Timeout => SettlementSubmissionOutcome::Timeout,
             Self::Revert => SettlementSubmissionOutcome::Revert,
             Self::Canceled => SettlementSubmissionOutcome::Cancel,
+            Self::Disabled(_) => SettlementSubmissionOutcome::Disabled,
             Self::Other(_) => SettlementSubmissionOutcome::SimulationRevert,
```
nlordell, line 163: Unrelated to this PR, but does it makes sense for this to be the `Failure` outcome that we used to have? From what I understand, you can get here by errors like failing to establish a TCP connection to the node when submitting a transaction which seems misclassified as a "simulation revert".

Not suggesting fixing it in this PR, just something I noticed while reviewing.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -325,6 +338,16 @@ impl<'a> Submitter<'a> {
...
             };
 
+            // before submitting, check if the currently executing strategy is temporarily disabled
+
+            if let SubmissionLoopStatus::Disabled(reason) =
+                self.submit_api.submission_status(&settlement)
+            {
+                tracing::debug!("strategy temporarily disabled, reason: {:?}", reason);
+                tokio::time::sleep(params.retry_interval).await;
+                continue;
```
nlordell, line 348: Is the reasoning here that a strategy may be temporarily disabled and then re-enabled in the future? Is this something that we should account for now or can we just `break` when a strategy becomes disabled?

nlordell (commented):
LGTM! Just one small question.

crates/solver/src/settlement_submission/submitter.rs
sunce86: Exactly, enable/reenable is the idea behind the `continue`. At this moment I don't see any downside of it, and also I don't see any benefit of ending the submission loop for currently executing strategy.

nlordell (approved):

crates/solver/src/settlement_submission.rs
sunce86: I think you are right. Fixed.

Issue #1615, opened 2022-02-02
Implement optimized native price estimation


MartinquaXD:
#1582 collected some ideas and requirements how to optimize the native token price calculation. This issue reiterates once more the requirements and presents concrete implementation ideas to achieve them.

1. We need a component optimized to calculate a tokens' price in terms of the native token (ETH, xDai)
2. Make use of existing `PriceEstimators` by being compatible with the `PriceEstimating` trait
3. It should utilize a cache to keep latency low
4. This cache needs to stay up-to-date by doing maintenance periodically (similar to `solvable_orders`)
5. The code needs to be shareable with the planned `/auction` endpoint

### Idea
Use existing PriceEstimators to compose a second estimator hierarchy which is optimized for caching native token price queries. To be able to maintain the native price cache in the background, a new type will be implemented which holds a `Weak<CachingPriceEstimator>` that will refetch token prices periodically triggered by `ServiceMaintenance`.
This way the requirements can be met without introducing too much new code while still having the option to easily switch out the implementation later on.

### Roadmap
- [ ] Implement `NativePriceEstimating` trait for all types implementing `PriceEstimating`
- [ ] Refactor existing price estimator code a bit
  - [ ] Make `SantitizedPriceEstimator` store a `Arc<dyn PriceEstimating>` instead of `T: PriceEstimating` (this is needed to create a second price estimator hierarchy specialized for native token price estimation)
  - [ ] Add helper fn in `/orderbook/src/main.rs` which instantiates PriceEstimators based on `&[PriceEstimatorType]` (this makes it possible to create an entirely separate estimator hierarchy with minimal code duplication)
- [ ] Update `MinFeeCalculator::compute_unsubsidized_min_fee` to use optimized price estimator
  - [ ] Create new estimator hierarchy `SanitizedPE` -> `CachingPE` -> `CompetitionPE`
  - [ ] Store new estimator in `MinFeeCalculator`
  - [ ] Use new estimator in `compute_unsubsidized_min_fee`
- [ ] Maintain cache for native token prices
  - [ ] Implement `PriceCacheMaintainer` which implements `Maintaining` and manages a `CachingPE`
  - [ ] Instantiate `PriceCacheMaintainer` for the `CachingPE` inside the native price estimator hierarchy and pass that to `ServiceMaintenance`

### The new `NativePriceEstimating` trait
```rust
#[mockall::automock]
#[async_trait::async_trait]
pub trait NativePriceEstimating: Send + Sync {
[...]

PR #1617, opened 2022-02-03
Add NativePriceEstimator

MartinquaXD:
Related: #1615

To make it more convenient to estimate prices, I added a new trait `NativePriceEstimating` and a wrapper around `Arc<dyn PriceEstimating` which implements the new trait for the wrapped price estimator.
This component is not yet in use but a follow up PR shall change that and make the `MinFeeCalculator` use it.

### Test Plan
Unit test for new component

### DIFF
--- a/crates/shared/src/price_estimation.rs
+++ b/crates/shared/src/price_estimation.rs
@@ -133,6 +133,24 @@ pub async fn ensure_token_supported(
     }
 }
 
+#[mockall::automock]
+#[async_trait::async_trait]
+pub trait NativePriceEstimating: Send + Sync {
+    async fn estimate_native_price(&self, token: &H160) -> Result<f64, PriceEstimationError> {
+        self.estimate_native_prices(std::slice::from_ref(token))
+            .await
+            .into_iter()
+            .next()
+            .unwrap()
+    }
+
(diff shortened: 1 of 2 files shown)


crates/shared/src/price_estimation.rs
```diff
@@ -133,6 +134,24 @@ pub async fn ensure_token_supported(
     }
 }
 
+#[mockall::automock]
+#[async_trait::async_trait]
+pub trait NativePriceEstimating: Send + Sync {
```
vkgnosis, line 139: The return type is a plain `f64`. Could we make it clear in what unit this is? Like the comment  for price estimate

> The resulting price is how many units of sell_token needs to be sold for one unit of buy_token (sell_amount / buy_amount).

vkgnosis, line 139: Nit: I feel this fits better in the `native` module / file too next to the implementation but not sure.

vkgnosis (approved):

crates/shared/src/price_estimation.rs
MartinquaXD: Will do.

sunce86 (approved):

PR #1618, opened 2022-02-03
Use NativePriceEstimator in fee.rs

MartinquaXD:
Related to #1615

Now that we have the `NativePriceEstimating` trait we can use it to clean up `MinFeeCalculator` a little bit.
Note that effectively nothing changed because the `NativePriceEstimator` wraps around the same price estimator hierarchy we've been using before.

### Test Plan
No new behavior => only updated existing tests


sunce86 (approved):
Cleaner

vkgnosis (approved):

PR #1619, opened 2022-02-03
Add task maintaining estimator cache in the background

MartinquaXD:
Related: #1615 

This PR adds a helper function which can be used to update the cache of a `CachingPriceEstimator` in the background.
This function will be used in a follow up commit which will create a new price estimator hierarchy specifically for native price estimations.

### Test Plan
added new unit test

### DIFF
--- a/crates/shared/src/price_estimation/cached.rs
+++ b/crates/shared/src/price_estimation/cached.rs
@@ -1,8 +1,8 @@
 use super::{Estimate, PriceEstimating, PriceEstimationError, Query};
 use cached::{Cached, TimedSizedCache};
 use std::{
-    sync::{Arc, Mutex},
-    time::Duration,
+    sync::{Arc, Mutex, Weak},
+    time::{Duration, Instant},
 };
 
 /// Price estimator wrapper that caches Ok results for some time.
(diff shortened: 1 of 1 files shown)


vkgnosis (commented):
It might be better to have an explicit native token price estimator cache instead of reusing the general price estimator cache.

I'm worried that the components have different goals that align right now but would drift more in the future and then it would be come awkward to maintain.

For example the general cache needs to store a full Query as the key. But the native cache only needs a single H160 representing the other token. The general estimator is meant to be a short term cache while the native cache has more complicated logic tying the cache and the updating together that can likely be optimized if it is implemented on the same struct.

One idea I had related to that is that we could continuously update the native estimates. So instead of sleeping 30 second then updating all of them at the same time we could in a loop keep updating the N oldest entries.
This is not something we need to start with right now but it shows that we might want to use have a closer relation between the cache and the updating code.

Let me know what you think about think about this (specific cache for native tokens). I'm somewhat in favor of it based on gut feeling but not very sure.

vkgnosis:
Another point is that the general cache was added mostly for the native estimates. Once we have those in their own component we can get rid of the general cache again.

MartinquaXD:
I initially started out working on a specialized cache like you are suggesting but came to the conclusion that we could get a pretty good solution by composing existing code but I agree that optimizing this solution further would incur significant maintenance overhead.
I don't really like the idea of having 2 similar (but not identical) caching implementations. But ditching the general cache because it would no longer be needed seems worthwhile.

vkgnosis:
It is very likely that i will be ditched so I'm in favor of pretending it doesn't exist for the purposes for the native token cache.

MartinquaXD:
> It is very likely that i will be ditched ...

Oh no! 😅 
Then I will work on a PR with a proposal for a new cache.

MartinquaXD:
Closed in favor of #1626

PR #1620, opened 2022-02-04
Do not use join_all in 1Inch price estimator

vkgnosis:
The `estimates` functions is meant to allow more efficient estimating
when the estimator can evalute several queries together as is the case
for the baseline estimator.
For the external estimators whose api only allows single estimates at a
time there is no such efficiency gain. The 1inch estimator was
nonetheless doing all price estimation calls in parallel. This is
problematic when there are a large amount of queries as it would likely
look like a DOS to their backend.
For now I decided to ensure that all of the external estimators work the same way.
In the future we could decide to allow some fixed level of
parallelness like max 3 requests at the same time.

Related to https://github.com/gnosis/gp-v2-services/pull/1619

### DIFF
--- a/crates/shared/src/price_estimation/oneinch.rs
+++ b/crates/shared/src/price_estimation/oneinch.rs
@@ -2,7 +2,6 @@ use super::gas;
 use crate::oneinch_api::{OneInchClient, ProtocolCache, RestResponse, SellOrderQuoteQuery};
 use crate::price_estimation::{Estimate, PriceEstimating, PriceEstimationError, Query};
 use anyhow::Result;
-use futures::future;
 use model::order::OrderKind;
 use primitive_types::U256;
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


sunce86 (approved):
lg

crates/shared/src/price_estimation/oneinch.rs
```diff
@@ -72,12 +71,11 @@ impl PriceEstimating for OneInchPriceEstimator {
             a SanitizedPriceEstimator"
         );
 
-        future::join_all(
-            queries
-                .iter()
-                .map(|query| async move { self.estimate(query).await }),
-        )
-        .await
+        let mut results = Vec::new();
```
MartinquaXD, line 74: nit: This could allocate the correct size immediately: `Vec::with_capacity(queries.len())`

MartinquaXD (approved):

PR #1622, opened 2022-02-04
Composed estimator hierarchy for cached native price estimations

MartinquaXD:
Since #1619 triggered a discussion about the viability of the approach of composing a new native price estimator entirely from existing components, I wanted to share what I got so far.

The current code already highlights one weird thing with reusing the `CachingPriceEstimator`:
The `CachingPriceEstimator` as well as the task maintaining it in the background need to be parameterized a little bit different but not that different that it would make much sense introducing more CLI arguments for that.

Note creating the new estimator hierarchy is quite verbose but relatively painless in this approach but optimizing the task to maintain the cache in the background further would probably end up quite hacky.


MartinquaXD:
Closed in favor of #1626

PR #1624, opened 2022-02-04
Fix panic in share_common_pool_state

### LINKED ISSUE #1605: Panics in share_common_pool_state
We had two panics in production:

https://gnosisinc.slack.com/archives/CUD4MUCUF/p1643404096116769
https://gnosisinc.slack.com/archives/CUD4MUCUF/p1643442554550739?thread_ts=1643404096.116769&cid=CUD4MUCUF

Maybe the author of the share_common_pool_state could look in this? 

From the comments in code, we can read:
>This method is only ever
/// meant to be used internally, so we don't have to worry that these
/// assumptions leak out of this module.

Does this assumption still holds true? What is meant with Internally?

vkgnosis:
As the comment on the method states this function assumes that the first
future is resolved first. Only when it resolves as Ok the second future
is allowed to be awaited. Based on this assumptino the method body uses
now_or_never.
However, in tokio some futures can spuriously return pending because of
cooperative multi tasking: https://docs.rs/tokio/latest/tokio/task/index.html#cooperative-scheduling
This violates the assumption that now_or_never always succeeds.
To work around this we use try_recv which isn't an async function and so
doesn't take part in cooperative yielding.

Fixes https://github.com/gnosis/gp-v2-services/issues/1605

### Test Plan

CI

### DIFF
--- a/crates/shared/src/sources/balancer_v2/pools/common.rs
+++ b/crates/shared/src/sources/balancer_v2/pools/common.rs
@@ -320,7 +320,7 @@ fn share_common_pool_state(
     impl Future<Output = Result<PoolState>>,
     impl Future<Output = PoolState>,
 ) {
-    let (pool_sender, pool_receiver) = oneshot::channel();
+    let (pool_sender, mut pool_receiver) = oneshot::channel();
 
     let result = fut.inspect(|pool_result| {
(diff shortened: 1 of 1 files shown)


MartinquaXD (approved):
LGTM.
Really interesting problem.

nlordell (approved):
Nice catch!

PR #1626, opened 2022-02-07
Implement optimized CachingNativePriceEstimator

MartinquaXD:
Related: #1615
Opened to replace #1619 & #1622 

Added a new component which has been optimized to cache native price estimates and allows us to improve cache maintenance more easily than a solution composed from existing components.
The plan is to use this component in a second price estimator hierarchy which only ever estimates native prices.

The cache is unbounded in size and has a max age for the cached prices.
Price estimates resulting in a `PriceEstimationError` don't get cached.
The underlying cache is protected by a `RwLock` to keep latency low when multiple user need price estimates at the same time.
The provided maintenance strategy can be configured to run at any interval and to update every or at most `n` outdated prices at once per update cycle.

### Test Plan
New unit tests

### DIFF
--- a/crates/shared/src/price_estimation/native_price_cache.rs
+++ b/crates/shared/src/price_estimation/native_price_cache.rs
@@ -1,0 +1,317 @@
+use crate::price_estimation::native::NativePriceEstimating;
+use crate::price_estimation::PriceEstimationError;
+use primitive_types::H160;
+use std::collections::HashMap;
+use std::sync::{Arc, RwLock, Weak};
+use std::time::{Duration, Instant};
+
+#[derive(Debug, Clone)]
+struct CachedPrice {
+    price: f64,
+    updated_at: Instant,
+}
+
+struct Inner {
+    cache: RwLock<HashMap<H160, CachedPrice>>,
+    estimator: Box<dyn NativePriceEstimating>,
+    max_age: Duration,
+}
+
+impl Inner {
+    async fn estimate_prices_and_update_cache(
+        &self,
+        tokens: &[H160],
+    ) -> Vec<Result<f64, PriceEstimationError>> {
+        if tokens.is_empty() {
+            return Vec::default();
+        }
+
+        let now = Instant::now();
+        let prices = self.estimator.estimate_native_prices(tokens).await;
+        {
+            let mut cache = self.cache.write().unwrap();
+            for (token, result) in tokens.iter().zip(prices.iter()) {
+                if let Ok(price) = result {
+                    let mut entry = cache.entry(*token).or_insert_with(|| CachedPrice {
+                        price: *price,
+                        updated_at: now,
+                    });
+                    entry.updated_at = now;
+                    entry.price = *price;
+                }
+            }
+        }
+        prices
+    }
+
+    fn get_cached_prices(&self, tokens: &[H160]) -> Vec<Option<f64>> {
+        if tokens.is_empty() {
+            return Vec::default();
+        }
+
+        let now = Instant::now();
+        let cache = self.cache.read().unwrap();
+        tokens
+            .iter()
+            .map(|token| match cache.get(token) {
+                Some(entry) if now.saturating_duration_since(entry.updated_at) < self.max_age => {
+                    Some(entry.price)
+                }
+                _ => None,
+            })
+            .collect()
+    }
+}
+
+/// Wrapper around `Box<dyn PriceEstimating>` which caches successful price estimates for some time
+/// and supports updating the cache in the background.
+/// The size of the underlying cache is unbounded.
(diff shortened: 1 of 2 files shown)


vkgnosis (approved):

crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -0,0 +1,317 @@
...
+            cache: RwLock::new(Default::default()),
+            max_age,
+        }))
+    }
+
+    /// Spawns a background task maintaining the cache once per `update_interval`.
+    /// Only outdated prices get updated and older prices have a higher priority.
+    /// If `update_size` is `Some(n)` at most `n` prices get updated per interval.
+    /// If `update_size` is `None` no limit gets applied.
+    pub fn with_maintenance(self, update_interval: Duration, update_size: Option<usize>) -> Self {
```
vkgnosis, line 86: I think this doesn't have to consume `self`. Could take by normal reference.

sunce86 (approved):

crates/shared/src/price_estimation/native_price_cache.rs
MartinquaXD: No, it doesn't **have** to consume `self`. I just thought it would be nicer to read if you are able to chain those calls when you create a new estimator.
```rust
let estimator = CachingNativePriceEstimator::new(...).with_maintenenace(...);
```
vs
```rust
let estimator = CachingNativePriceEstimator::new(...);
estimator.with_maintenance(...);
```
If that is not desired I would prefer the function to be renamed as well:
```rust
pub fn spawn_maintenance_task(&self, update_interval: Duration, update_size: Option<usize>)
```

vkgnosis: Unless you explicitly want to use the builder pattern I prefer the third version. Usually you would use it when there are lots of things to set.
The difference is small so either way is fine I just slightly prefer it that way but we can continue with current version too.

MartinquaXD: I don't really care that much but I wanted to explain my reasoning.
Changed the signature and name now.

crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -0,0 +1,316 @@
...
+    /// Spawns a background task maintaining the cache once per `update_interval`.
+    /// Only outdated prices get updated and older prices have a higher priority.
+    /// If `update_size` is `Some(n)` at most `n` prices get updated per interval.
+    /// If `update_size` is `None` no limit gets applied.
+    pub fn spwan_maintenance_task(&self, update_interval: Duration, update_size: Option<usize>) {
+        tokio::spawn(update_most_outdated_prices(
+            Arc::downgrade(&self.0),
+            update_interval,
+            update_size,
+        ));
```
nlordell, line 91: No strong opinion here, but I slightly prefer making `impl ServiceMaintaining for CachingNativePriceEstimator`.

While a timed interval also makes sense, I think for things that are dependent on block chain state (such as token prices - which tend to depend on AMM states), updating at discrete block intervals makes slightly more sense to me.

Additionally, since we already only update at most `update_size` entries that are older than `max_age` I don't see an issue with updating prices too often.

The other possible strategy would be an LRU cache.

nlordell (approved):

crates/shared/src/price_estimation/native_price_cache.rs
MartinquaXD: One concern with implementing `ServiceMaintaining` was that some REST APIs might not appreciate it if we updated too many price estimates at once for every block.
I think the current approach gives us fairly up-to-date prices while doing not too much work at once (of course depending on the parameters we settle on).

nlordell: > was that some REST APIs might not appreciate it if we updated too many price estimates at once for every block.

Right, but don't we only update prices that are older than `max_age` and at most `update_size`? I thought that would be enough to prevent rate limiting.

To be clear, I'm fine with how it is now.

MartinquaXD: It all depends on what we want the caches properties to be like.
I aimed for:
- the prices of all used tokens are reasonably up-to-date at all times
- a user request should rarely ever miss the cache
- not DOSing REST APIs while updating

To do the same thing when implementing `Maintaining` could get a little weird because we would need to batch our work based on some timer while also dealing with the fact that block times are not 100% predictable.
If we only do one small batch of updates per block price estimations of rarely traded tokens could encounter higher latency.
Overall I think the timer based approach is easier to grasp and evaluate but as I mentioned it all depends on the goals we want to achieve.

PR #1627, opened 2022-02-07
change the additional-tip calculation

sunce86:
Alex noticed that with the current network conditions (gas price ~ 50Gwei) we might be overpaying the Flashbots, giving them fixed 7Gwei tip increase.

This PR aims at lowering the additional tip in low gas price network conditions.

### DIFF
--- a/crates/solver/src/settlement_submission/submitter.rs
+++ b/crates/solver/src/settlement_submission/submitter.rs
@@ -32,6 +32,9 @@ use shared::Web3;
 use std::time::{Duration, Instant};
 use web3::types::{TransactionReceipt, U64};
 
+/// Used for calculating the maximum additional tip based on the current value of max_fee_per_gas
+const MAX_FEE_PER_GAS_PERCENT: f64 = 0.05; //5%
+
 /// Parameters for transaction submitting
 #[derive(Clone, Default)]
 pub struct SubmitterParams {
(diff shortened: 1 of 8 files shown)


crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -32,6 +32,9 @@ use shared::Web3;
 use std::time::{Duration, Instant};
 use web3::types::{TransactionReceipt, U64};
 
+/// Used for calculating the maximum additional tip based on the current value of max_fee_per_gas
+const MAX_FEE_PER_GAS_PERCENT: f64 = 0.05; //5%
```
josojo, line 36: Could we get that as a parameter for the driver?

Personally, I think that 0.03 should be sufficient, because the max_fee_per_gas is usually much higher than then spent base_price. E.g. over here its over 50%: https://etherscan.io/tx/0x0dcbbf13a3577b769f9245095901de04172ce1b5a2b20c4425b33a8c33ab4b90

nlordell: Can we make this configurable so that we can change it without requiring code changes when the network conditions change?

josojo (commented):
Generally, I like the PR a lot. I think it is an effective way to reduce our gas costs.

Given that flashbots is building quite large bundles, e.g. [this one](https://flashbots-explorer.marto.lol/?block=14158811), I think it makes sense to reassess experiments with lower additional tips in general.

vkgnosis (approved):

josojo (approved):
Nice

nlordell (approved):

Issue #1573, opened 2022-01-17
Record Settlement Gas Cost vs Order Fees Collected


nlordell:
We currently record in the `orderbook` database rich `fee_amount` data with each order (so AFAIU, enough to recover the estimated amount of gas for an order in ETH).

This issue captures the work to recover additional settlement fee information (so gas used and effective gas price) as well so we can gather better data on how far off our fee estimates are vs. what they actually cost.

nlordell:
Note that this data can be read from the block-chain "after the fact". This means that, while we haven't been recording this, we can still retroactively compute cost coverage for our orders.

Maybe this data recording is something that can be delegated to a background worker of some sort.

vkgnosis:
We have logging in the solver now for easy manual analysis. I am going to write a program that uses the database to automatically find settlements an their orders and gets the transaction gas data from an ethereum node. If that works well I like that approach because it is decoupled from the rest of the system.

nlordell:
Is this something that we should add to our database or should it be completely separate process?

vkgnosis:
I'm not sure if we need it in the database or if it would be fine to run the tool from time to time to generative a csv file or something like that. If we do want it in the database I'm not sure. Separate process feels nice because it *can* be a separate process as it doesn't need to be synchronized to the main process in any way (?). Currently I feel I should prototype it as a standalone application.

vkgnosis:
Created https://github.com/gnosis/gp-v2-fee-analysis . Currently prints information to stdout for settlements in the last 100 blocks.

vkgnosis:
Next step: More comprehensive analysis using this tool. (Not necessarily by me.) Should be easy to adapt it for specific use case.

PR #1630, opened 2022-02-08
Print earned fee in driver

vkgnosis:
Part of https://github.com/gnosis/gp-v2-services/issues/1573

- Store the originally signed fee amount in several place
- Make the distinction clear between that fee (unscaled subsidized) and the scaled unsubsidized fee that is already used
- Log both fees in the driver

This is the minimum data we need to compare settlement transaction cost to expected cost. After we see that this data makes sense we might extend the way it is logged as part of the above issue.

### Test Plan

CI

### DIFF
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -62,6 +62,7 @@ pub struct Driver {
     in_flight_orders: InFlightOrders,
     post_processing_pipeline: PostProcessingPipeline,
     simulation_gas_limit: u128,
+    fee_objective_scaling_factor: BigRational,
 }
 impl Driver {
     #[allow(clippy::too_many_arguments)]
@@ -88,6 +89,7 @@ impl Driver {
         order_converter: OrderConverter,
         weth_unwrap_factor: f64,
         simulation_gas_limit: u128,
+        fee_objective_scaling_factor: f64,
     ) -> Self {
         let post_processing_pipeline = PostProcessingPipeline::new(
             native_token,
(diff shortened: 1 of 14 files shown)


crates/solver/src/driver.rs
```diff
@@ -329,25 +333,30 @@ impl Driver {
                 .get(&self.native_token)
                 .expect("Price of native token must be known.");
 
-        let rate_settlement = |settlement: Settlement, gas_estimate| {
+        let rate_settlement = |id, settlement: Settlement, gas_estimate| {
```
sunce86, line 336: Should settlement id be part of the `Settlement` struct to be more clear?

sunce86 (approved):

crates/solver/src/settlement.rs
```diff
@@ -251,6 +256,23 @@ impl Settlement {
             .sum()
     }
 
+    // Computes the total scaled unsubsidized fee of all protocol trades (in wei ETH).
+    pub fn total_unscaled_subsidized_fees(
```
sunce86, line 260: nit: maybe add unit test for this function (noticed the decrease in unit test coverage)

nlordell (commented):
LGTM

nlordell (approved):

crates/solver/src/settlement.rs
vkgnosis: Decided against this because the functions are so similar didn't feel test was worth it.

PR #1631, opened 2022-02-08
Make 1Inch pricing parameters more competive

MartinquaXD:
We no longer need to make sure 1Inch generates low gas cost trading routes.
Especially for big volume trades there often exists a trading route which yields an overall better price even if you consider higher gas costs.
To get the best prices I removed our default arguments which are below the 1Inch default and added values for `complexity_level` and `main_parts` which are bigger than 1Inch's defaults. Those parameters where the only ones which yielded higher prices compared to queries with their default values instead.
Note that growing the solution space by giving more generous query arguments also increases the runtime of the query which will result in higher latency for price estimations.
I tried to measure the runtime regression bu the variance between the runs were quite. Overall the latency regression seems worth the up to ~2% better prices for big volume trades.
The routes generated for smaller volume trades most of the time didn't change with the new parameters.

### Test Plan
Manual test

### DIFF
--- a/crates/shared/src/oneinch_api.rs
+++ b/crates/shared/src/oneinch_api.rs
@@ -146,13 +146,12 @@ impl SellOrderQuoteQuery {
             to_token_address: buy_token,
             amount,
             protocols,
-            // Use at most 2 connector tokens
-            complexity_level: Some(Amount::new(2).unwrap()),
-            // Cap swap gas to 750K.
-            gas_limit: Some(Amount::new(750_000).unwrap()),
-            // Use only 3 main route for cheaper trades.
-            main_route_parts: Some(Amount::new(3).unwrap()),
-            parts: Some(Amount::new(3).unwrap()),
+            // Use max value instead of default
+            complexity_level: Some(Amount::new(3).unwrap()),
+            gas_limit: None,
+            // use max value instead of default
+            main_route_parts: Some(Amount::new(50).unwrap()),
+            parts: None,
             fee: None,
             gas_price: None,
             virtual_parts: None,


crates/shared/src/oneinch_api.rs
```diff
@@ -146,13 +146,12 @@ impl SellOrderQuoteQuery {
...
-            // Cap swap gas to 750K.
-            gas_limit: Some(Amount::new(750_000).unwrap()),
-            // Use only 3 main route for cheaper trades.
-            main_route_parts: Some(Amount::new(3).unwrap()),
-            parts: Some(Amount::new(3).unwrap()),
+            // Use max value instead of default
+            complexity_level: Some(Amount::new(3).unwrap()),
+            gas_limit: None,
+            // use max value instead of default
+            main_route_parts: Some(Amount::new(50).unwrap()),
```
nlordell, line 153: Are these the same values 1Inch uses in their front-end? I would be inclined to just use those so the "quickly compare prices on 1Inch" gives us the same results.

nlordell:
Very nice analysis of how the 1Inch API responds to tweaking parameters!

crates/shared/src/oneinch_api.rs
MartinquaXD: **TLDR:** no

This is how the 1Inch front-end gets their quotes:
```
https://pathfinder.1inch.io/v1.2/chain/1/router/v4/quotes-by-presets?chainId=1&fromTokenAddress=0x0d438f3b5175bebc262bf23753c1e53d03432bde&toTokenAddress=0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48&amount=24062474987761938271780&gasPrice=64974614477&maxReturnProtocols=WETH,UNISWAP_V1,UNISWAP_V2,SUSHI,MOONISWAP,BALANCER,COMPOUND,CURVE,CURVE_V2_SPELL_2_ASSET,CURVE_V2_THRESHOLDNETWORK_2_ASSET,CHAI,OASIS,KYBER,AAVE,IEARN,BANCOR,PMM1,CREAMSWAP,SWERVE,BLACKHOLESWAP,DODO,DODO_V2,VALUELIQUID,SHELL,DEFISWAP,COFIX,SAKESWAP,LUASWAP,MINISWAP,MSTABLE,PMM2,AAVE_LIQUIDATOR,SYNTHETIX,AAVE_V2,ST_ETH,ONE_INCH_LP,ONE_INCH_LP_1_1,ONE_INCH_LP_MIGRATOR,ONE_INCH_LP_MIGRATOR_V1_1,UNISWAP_V2_MIGRATOR,SUSHISWAP_MIGRATOR,LINKSWAP,S_FINANCE,PSM,POWERINDEX,INDEXED_FINANCE,PMM3,XSIGMA,CREAM_LENDING,SMOOTHY_FINANCE,SADDLE,PMM4,KYBER_DMM,BALANCER_V2,UNISWAP_V3,SETH_WRAPPER,CURVE_V2,CURVE_V2_EURS_2_ASSET,CURVE_V2_EURT_2_ASSET,CURVE_V2_XAUT_2_ASSET,CURVE_V2_ETH_CRV,CURVE_V2_ETH_CVX,CONVERGENCE_X,ONE_INCH_LIMIT_ORDER,ONE_INCH_LIMIT_ORDER_V2,DFX_FINANCE,FIXED_FEE_SWAP,DXSWAP,CLIPPER,SHIBASWAP,UNIFI,PMMX,PMM5,PSM_PAX,PMM2MM,PMM2MM1,WSTETH,DEFI_PLAZA,FIXED_FEE_SWAP_V3,SYNTHETIX_WRAPPER,SYNAPSE&time=1644327191081
```

This includes 2 interesting pieces of information:
1. They specify `&maxReturnProtocols=WETH,UNISWAP_V1,UNISWAP_V2,SUSHI,MOONISWAP,...` which does have a 1-to-1 representation (by name) in the `/quote` or `/swap` endpoints but it sounds and looks like that would be equivalent to the `protocols` argument which limits which protocols may be included in a trade route.
2. They query `https://pathfinder.1inch.io/v1.2/chain/1/router/v4/**quotes-by-presets**` this looks like they have an endpoint which sets some default arguments to each request.
The default `complexity_level` is 2 but I only ever got the comparable results as the 1Inch front-end when I set it to 3 and also increased the `main_route_parts` which makes me think that those parameters are added by the `quotes-by-presets` endpoint.

vkgnosis (approved):

josojo (approved):

nlordell (approved):

PR #1634, opened 2022-02-09
Driver logging improvements

vkgnosis:
Use the new rated settlement id in more places and log rated settlements
over multiple lines.

### Test Plan

New test. It is ignored because it makes sense for a human to look at the output, not ci.

### DIFF
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -182,8 +182,8 @@ impl Driver {
             Ok(receipt) => {
                 let name = solver.name();
                 tracing::info!(
-                    "Successfully submitted {} settlement: {:?}",
-                    name,
+                    "Successfully submitted settlement id {} as tx hash {:?}",
+                    rated_settlement.id,
                     receipt.transaction_hash
                 );
                 trades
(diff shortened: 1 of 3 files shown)


crates/solver/src/driver.rs
```diff
@@ -182,8 +182,8 @@ impl Driver {
             Ok(receipt) => {
                 let name = solver.name();
                 tracing::info!(
-                    "Successfully submitted {} settlement: {:?}",
-                    name,
+                    "Successfully submitted settlement id {} as tx hash {}",
```
sunce86, line 185: I found it is generally more useful to use `{:?}` for transaction hash because of the hash length, since with `{}` only part of the hash is logged making it useless...

sunce86 (approved):

crates/solver/src/driver.rs
vkgnosis: Fixed. Didn't intentionally change this.

nlordell (approved):

crates/solver/src/driver.rs
```diff
@@ -656,4 +669,52 @@ mod tests {
         );
         assert!(!is_only_selling_trusted_tokens(&settlement, &token_list));
     }
+
+    #[test]
+    #[ignore]
+    fn print_settlements_() {
```
MartinquaXD, line 675: nit: I guess it would be nice name the test `print_settlements` (without _).
This could be done by calling the actual `print_settlements` function with `super::print_settlements`.

MartinquaXD (approved):

PR #1637, opened 2022-02-09
disable strategies by revertable risk level

sunce86:
This PR implements https://github.com/gnosis/gp-v2-services/issues/1538

`Settlement` can have different `Revertable` risk level:
1. `Revertable::NoRisk` - `Settlement` has no interactions, trades against internal buffers + cows
2. `Revertable::LowRisk` - `Settlement` has interactions, but all of them have <X eth sell/buy amount
3. `Revertable::HighRisk` - `Settlement` has interactions, but at least one of them has >X eth sell/buy amount

Depending on the `Revertable` attribute, transaction strategies are enabled/disabled:
1. `Revertable::NoRisk` - All strategies are enabled (`CustomNodes+Eden+Flashbots`), but `Eden+Flashbots` premium tip is set to zero (we don't want to pay extra since there is no need)
2. `Revertable::LowRisk` - `CustomNodes` disabled, `Eden+Flashbots` enabled with premium tip. `Flashbots` have protection against reverted transactions, while `Eden` should work decently with staked Eden on solver accounts.
3. `Revertable::HighRisk` - Only `Flashbots` enabled.

This way, we always have at least 1 strategy enabled, avoiding to accidentally disable all strategies.

### EDIT (keeping the original idea for future):
Initial idea greatly simplified by removing the `executed_amounts` from `Interaction` trait, until we decide how to inspect interactions the best way. Also, `Revertable::LowRisk` is removed. This simplified version is easier to review, deploy, test and monitor.


crates/solver/src/settlement.rs
```diff
@@ -278,9 +289,25 @@ impl Settlement {
...
+
+        match self
+            .encoder
+            .interactions()
+            .iter()
+            .filter_map(|interaction| {
+                let (token, amount) = interaction.executed_amount()?;
+                amount.checked_mul(self.clearing_price(token)?)
+            })
+            .any(|executed_amount| executed_amount > U256::from_f64_lossy(5e36))
```
sunce86, line 306: 5e36 is bad way I guess, since not all tokens have 18digits base right?

MartinquaXD: Aren't clearing prices already computed in WETH so it wouldn't matter what precision each token has?

MartinquaXD:
I guess as a cost saving measure we could also only enable `CustomNodes` for `Revertable::NoRisk`?

crates/solver/src/settlement.rs
sunce86: Hm, clearing price for ETH/WETH is "0x0de0b6b3a7640000" = 1000000000000000000, so they are denominated in ETH but expressed in wei.

sunce86: The same is true for amount, so basically need to bump x18 twice

crates/solver/src/settlement/settlement_encoder.rs
```diff
@@ -87,6 +87,10 @@ impl SettlementEncoder {
         &self.trades
     }
 
+    pub fn interactions(&self) -> &Vec<Arc<dyn Interaction>> {
```
nlordell, line 90: nit: Lets call this `execution_plan` since it is only interactions from the execution plan and don't include things like WETH unwraps.

nlordell (commented):
I like this change.

I find just having the simple has/doesn't have interactions decide how submission works be low hanging optimization for paying less priority fees and faster execution (by using the public mempool).

Just to nitpick a bit, I think the PR description is out of date.

nlordell:
> To do it non intrusively we can look at the orders.

I think ultimately we need to look at interaction execution amounts. Orders volume in it of itself is a very poor indication on revert risk/MEV amount since two perfectly overlapping orders already have a very small revert risk (user cancels approval or transfers funds) and no MEV possibility.

crates/solver/src/settlement_submission/submitter/eden_api.rs
```diff
@@ -57,11 +57,14 @@ impl TransactionSubmitting for EdenApi {
...
     }
 
-    fn submission_status(
-        &self,
-        _settlement: &Settlement,
-        _network_id: &str,
-    ) -> SubmissionLoopStatus {
-        SubmissionLoopStatus::Enabled
+    fn submission_status(&self, settlement: &Settlement, network_id: &str) -> SubmissionLoopStatus {
+        // disable strategy if there is a high posibility to be reverted (check done only for mainnet)
```
MartinquaXD, line 61: nit: `posibility` -> `possibility`

crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
```diff
@@ -120,11 +123,13 @@ impl TransactionSubmitting for CustomNodesApi {
     }
 
     fn submission_status(&self, settlement: &Settlement, network_id: &str) -> SubmissionLoopStatus {
-        // disable strategy if not mev safe (check done only for mainnet)
-        if !settlement.mev_safe() && shared::gas_price_estimation::is_mainnet(network_id) {
-            return SubmissionLoopStatus::Disabled(DisabledReason::MevExtractable);
+        // disable strategy if there is a slighest posibility to be reverted (check done only for mainnet)
```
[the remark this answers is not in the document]
sunce86: Fixed.

crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
```diff
@@ -120,11 +123,13 @@ impl TransactionSubmitting for CustomNodesApi {
     }
 
     fn submission_status(&self, settlement: &Settlement, network_id: &str) -> SubmissionLoopStatus {
-        // disable strategy if not mev safe (check done only for mainnet)
-        if !settlement.mev_safe() && shared::gas_price_estimation::is_mainnet(network_id) {
-            return SubmissionLoopStatus::Disabled(DisabledReason::MevExtractable);
+        // disable strategy if there is a slightest possibility for a transaction to be reverted (check done only for mainnet)
+        if shared::gas_price_estimation::is_mainnet(network_id) {
```
nlordell, line 127: Why do we only do this for mainnet? In theory this applies to all networks right?

nlordell (commented):
LGTM,

The only thing that isn't clear to me is why we only disable non-`revertable` settlements on Mainnet (to me this would also, in theory, apply to Gnosis Chain for example).

crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
sunce86: Currently on gnosis chain our configuration is to use only `CustomNodes`, so If we were to allow disabling it we would be left without any strategy. Eden and Flashbots rpcs are not enabled on gnosis chain afaik (maybe wrong)

nlordell: Ah! Of course!

nlordell (approved):

crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
sunce86: Although I do agree it seems a bit hardcoded and hard to maintain (it will be easy to forget about this `if` when we add new network).

MartinquaXD (approved):

vkgnosis (approved):

PR #1638, opened 2022-02-10
Add metrics for native price cache hits/misses

MartinquaXD:
To see if the new native price cache is worth it we need to see how many cache hits vs misses there are.
Note that cache misses and hits are not updated during maintenance cycles.
This way we only track how much our users get out of the new cache.

### Test Plan
Manual test
```
curl -s http://localhost:9586/metrics | grep native 
# HELP gp_v2_api_native_price_cache Native price cache hit/miss counter.
# TYPE gp_v2_api_native_price_cache counter
gp_v2_api_native_price_cache{result="hits"} 2
gp_v2_api_native_price_cache{result="misses"} 1
```

### DIFF
--- a/crates/shared/src/price_estimation/native_price_cache.rs
+++ b/crates/shared/src/price_estimation/native_price_cache.rs
@@ -5,6 +5,16 @@ use std::collections::HashMap;
 use std::sync::{Arc, RwLock, Weak};
 use std::time::{Duration, Instant};
 
+#[cfg_attr(test, mockall::automock)]
+pub trait Metrics: Send + Sync + 'static {
+    fn native_price_cache(&self, misses: usize, hits: usize);
+}
+
+struct NoopMetrics;
+impl Metrics for NoopMetrics {
+    fn native_price_cache(&self, _: usize, _: usize) {}
+}
+
 #[derive(Debug, Clone)]
 struct CachedPrice {
     price: f64,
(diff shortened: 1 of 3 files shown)


crates/shared/src/price_estimation/native_price_cache.rs
```diff
@@ -5,6 +5,16 @@ use std::collections::HashMap;
 use std::sync::{Arc, RwLock, Weak};
 use std::time::{Duration, Instant};
 
+#[cfg_attr(test, mockall::automock)]
+pub trait Metrics: Send + Sync + 'static {
+    fn native_price_cache(&self, misses: usize, hits: usize);
+}
+
+struct NoopMetrics;
```
nlordell, line 13: For some reason, I think we may already have some `NoopMetrics` type somewhere... Might be useful to reuse that one (if I'm not remembering incorrectly).

nlordell (approved):
Nice!

sunce86 (approved):
Lg

crates/shared/src/price_estimation/native_price_cache.rs
MartinquaXD: There is currently only a `NoopMetrics` struct in the `solver` crate.
We could move that into `shared/src/metrics` and use that across all crates.

nlordell: Ah OK. Here is fine then.

vkgnosis (approved):

PR #1640, opened 2022-02-10
Delay OneInchClientImpl error until estimator gets initialized

MartinquaXD:
#1635 caused the 1Inch API to always get instantiated even if no 1Inch estimator is configured. This causes the  orderbook to always crash on rinkeby.

Now we still try to always create the 1Inch API but only propagate that error if the `OneInchEstimator` needs to get instantiated.

### Test Plan
Manual test on rinkeby

### DIFF
--- a/crates/orderbook/src/main.rs
+++ b/crates/orderbook/src/main.rs
@@ -445,9 +445,9 @@ async fn main() {
         )
         .unwrap(),
     );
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):

fedgiac (approved):
Looks reasonable!

PR #1644, opened 2022-02-11
Propose new `/auction` endpoint

nlordell:
This PR adds an initial version of the `/auction` endpoint to the open API specification for the orderbook service.

This proposes what could be the new `/auction` endpoint that the driver uses for populating getting batches. Currently, it is just `v2/solvable_orders` with the reference prices, but we may decide to add more to it in the future (there was discussion of including baseline liquidity here as well to reduce the amount of work required for new solvers - allowing them to integrate with our public `orderbook` API directly without needing to also run a `driver`).

One thing that we should probably discuss (but not necessarily important for this PR) is whether or not auctions should be sealed (i.e. multiple queries of `/auction` within a small time frame return the exact same data) or continuous (a new auction on each block). For canonical batches - it makes more sense to have them "sealed", but in the general case, I'm slightly leaning towards having them continuous, although this makes solution competition a bit more difficult if we decide to not have it happen in the driver.

### Test Plan

CI - we validate OpenAPI file 🎉.

Added an empty route  always returning 500 - you can try it out with:
```
% curl -s http://localhost:8080/api/v1/auction
{"errorType":"InternalServerError","description":""}
```

And check that the error metrics for it increased!:
```
% curl -s http://localhost:9586/metrics | rg 'gp_v2_api_api_requests_complete' | rg 'v1/auction'
gp_v2_api_api_requests_complete{method="v1/auction",status_code="500"} 1
```

### DIFF
--- a/crates/orderbook/openapi.yml
+++ b/crates/orderbook/openapi.yml
@@ -667,7 +667,7 @@ components:
         - $ref: "#/components/schemas/OrderMetaData"
     Auction:
       description: |
-        A batched auction for solving.
+        A batch auction for solving.
       type: object
       properties:
         block:
--- a/crates/orderbook/openapi.yml
+++ b/crates/orderbook/openapi.yml
@@ -240,6 +240,20 @@ paths:
                   latestSettlementBlock:
                     type: integer
                     description: The block number in which the most recent settlement was observed.
+  /api/v1/auction:
+    get:
+      summary: Gets the current batch auction.
+      description: |
+        The current batch auction that solvers should be solving right now. Includes the list of
+        solvable orders, the block on which the batch was created, as well as prices for all tokens
+        being traded (used for objective value computation).
+      responses:
+        200:
+          description: the auction
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Auction"
   /api/v1/fee:
     get:
       description: |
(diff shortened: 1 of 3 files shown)


crates/orderbook/openapi.yml
```diff
@@ -651,6 +665,31 @@ components:
...
+          items:
+            $ref: "#/components/schemas/Order"
+          description: |
+            The solvable orders included in the auction.
+        prices:
+          type: object
+          additionalProperties:
+            $ref: "#/components/schemas/BigUint"
+          description: |
+            The reference prices for all traded tokens in the auction. These prices are used for
```
vkgnosis, line 689: Not sure how to specify this but it would be nice if it said that the keys are token addresses.

vkgnosis (approved):

crates/orderbook/openapi.yml
```diff
@@ -651,6 +665,31 @@ components:
       allOf:
         - $ref: "#/components/schemas/OrderCreation"
         - $ref: "#/components/schemas/OrderMetaData"
+    Auction:
+      description: |
+        A batched auction for solving.
```
MartinquaXD, line 670: nit: I was under the impression we call them "batch auction". Could be mishearing, though.
Also wouldn't it be "A batch(ed) auction to solve".

bh2smith: To cover all case we could say "An auction batch"

crates/orderbook/openapi.yml
```diff
@@ -651,6 +665,31 @@ components:
...
+      type: object
+      properties:
+        block:
+          type: integer
+          description: |
+            The block number for the auction. Orders and prices are guaranteed to be valid on this
+            block. Proposed settlements should be valid for this block as well.
```
bh2smith, line 677: I like this. The order set would then be those valid orders in the given block window. Seems coherent, straightforward, clear and easy to query for!

bh2smith (approved):

bh2smith:
While, I approve the spec... wouldn't it make sense to include the implementation here or at least some response saying "not implemented" so nobody attempts to use the described end point and gets error response.

josojo (approved):

crates/orderbook/openapi.yml
nlordell: I meant it to be "batch auction" - nice catch!

nlordell:
> wouldn't it make sense to include the implementation

I don't think so. Specifically, this PR was more about the discussion around "how the auction end point should look like" and not the implementation for it.

I  can add an empty route that always returns a 500 error.

PR #1648, opened 2022-02-14
Add new Auction model type

nlordell:
This PR just adds a new `Auction` model type (doesn't use it yet). This model will be returned by the new `v1/auction` endpoint that was introduced in #1644 

### Test Plan

New unit test to check serialization/deserialization.

### DIFF
--- a/crates/model/src/auction.rs
+++ b/crates/model/src/auction.rs
@@ -1,0 +1,70 @@
+//! Module defining a batch auction.
+
+use crate::{order::Order, u256_decimal::DecimalU256};
+use primitive_types::{H160, U256};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+use std::collections::BTreeMap;
+
+/// A batch auction.
+#[serde_as]
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+pub struct Auction {
+    /// The block that this auction is valid for.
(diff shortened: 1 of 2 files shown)


vkgnosis (approved):

MartinquaXD (approved):

crates/model/src/auction.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// The block number for the auction. Orders and prices are guaranteed to be
+    /// valid on this block.
+    pub block: u64,
+
+    /// The solvable orders included in the auction.
+    pub orders: Vec<Order>,
+
+    /// The reference prices for all traded tokens in the auction.
+    #[serde_as(as = "BTreeMap<_, DecimalU256>")]
+    pub prices: BTreeMap<H160, U256>,
```
sunce86, line 23: We used `HashMap` for clearing prices in the `Settlement`, any particular reason to use `BTreeMap` here?

sunce86 (approved):
lg

crates/model/src/auction.rs
nlordell: I chose `BTreeMap` just to make sure that the JSON output is always ordered. So multiple calls to `v1/auction` return the same JSON blob (and don't depend on the randomness of `HashMap` and its use of `Hash`).

PR #1649, opened 2022-02-14
Implement auction endpoint to return prices with solvable orders

nlordell:
This PR adds an initial implementation for the new `v1/auction` endpoint.

Currently, its mostly a wrapper around the `solvable_orders` API endpoint with a couple of differences:
- The block is reported as the last processed block by the event updater instead of the latest settlement block. This makes the block number closer to "latest" and be a better representation of the block the batch auction is valid for. In case the event handler has not processed any blocks - it returns the block number of the last settlement (as the `v2/solvable_orders` endpoint). This should only happen right on startup so I'm not particularly worried about optimizing the case where no events have been processed.
- We filter out orders for which there are no native prices. Since we use the same native price estimator from the `/quote` endpoint, this shouldn't happen often. Currently, we only log a warning when this happens. I want to change this to bump some metrics to keep track of this a bit better (will do it in a follow up PR).

Things still left to do:
- We no longer need to augment prices in the driver (since, not only do we use the native token estimator which supports more tokens, additionally orders for which there are no native prices shouldn't get added to the orderbook anymore, and finally we also filter the orders before sending them to the driver so augmenting would never really do anything anyway).
- Change the driver to use the `Auction` data from this new endpoint.
- Metrics for filtered orders as mentioned ☝️.

### Test Plan

Added new unit test to verify order filtering logic.

Additionally, did some manual testing. Ran the orderbook locally and [added an order](). Then queried the orderbook's new `auction` endpoint:

<details><summary><code>curl -s http://localhost:8080/api/v1/auction | jq</code></summary>

```
{
  "block": 14210281,
  "orders": [
    {
      "creationDate": "2022-02-15T10:46:21.893840Z",
      "owner": "0xb54872859733a3dfbb2c5401ac68cd9ca84b3cd1",
      "uid": "0x6c85ed0264a62ba594b7edc7b9209b05c15ce6814c4b11fc1ae855be48092471b54872859733a3dfbb2c5401ac68cd9ca84b3cd180000000",
      "availableBalance": "600000000000000000000",
      "executedBuyAmount": "0",
      "executedSellAmount": "0",
      "executedSellAmountBeforeFees": "0",
      "executedFeeAmount": "0",
      "invalidated": false,
      "status": "open",
      "settlementContract": "0x9008d19f58aabd9ed0d60971565aa8510560ab41",
      "fullFeeAmount": "9726151661805737984",
      "sellToken": "0xde30da39c46104798bb5aa3fe8b9e0e1f348163f",
      "sellAmount": "590273848338194262016",
      "buyAmount": "11899022238076966313",
      "validTo": 2147483648,
      "feeAmount": "9726151661805737984",
      "kind": "sell",
      "partiallyFillable": false,
      "signingScheme": "eip712",
      "signature": "0x8535eb8070a7327406bcf69574bda0c494304354c6a36e00886a8461bf2f57e738830870bff9abd75d243dba3c4999e118da11243c093044fa58812d1fc51ea91b",
      "sellTokenBalance": "erc20",
      "buyTokenBalance": "erc20"
    }
  ],
  "prices": {
    "0x6810e776880c02933d47db1b9fc05908e5386b96": "117755220144748432",
    "0xde30da39c46104798bb5aa3fe8b9e0e1f348163f": "2415672455701688"
  }
}
```

</details>

This trades GNO priced at ~$372.74 for GTC priced at ~$7.66. Given an ETH priced at ~$3,109.71. The prices in the vector make sense (GNO is ~Ξ0.1177 and GTC is ~Ξ0.0024).

### DIFF
--- a/crates/orderbook/src/orderbook.rs
+++ b/crates/orderbook/src/orderbook.rs
@@ -1,19 +1,28 @@
 use crate::{
     api::order_validation::{OrderValidating, OrderValidator, ValidationError},
     database::orders::{InsertionError, OrderFilter, OrderStoring},
+    event_updater::EventUpdating,
     solvable_orders::{SolvableOrders, SolvableOrdersCache},
 };
 use anyhow::{ensure, Context, Result};
 use chrono::Utc;
-use ethcontract::H256;
+use ethcontract::{H256, U256};
 use model::{
+    auction::Auction,
     order::{Order, OrderCancellation, OrderCreationPayload, OrderStatus, OrderUid},
     signature::SigningScheme,
     DomainSeparator,
 };
 use primitive_types::H160;
-use shared::{bad_token::BadTokenDetecting, metrics::LivenessChecking};
-use std::{collections::HashSet, sync::Arc, time::Duration};
+use shared::{
+    bad_token::BadTokenDetecting, metrics::LivenessChecking,
+    price_estimation::native::NativePriceEstimating,
+};
+use std::{
+    collections::{BTreeMap, HashSet},
+    sync::Arc,
+    time::Duration,
+};
 
 #[derive(Debug)]
 pub enum AddOrderResult {
@@ -40,10 +49,12 @@ pub struct Orderbook {
     settlement_contract: H160,
     database: Arc<dyn OrderStoring>,
     bad_token_detector: Arc<dyn BadTokenDetecting>,
+    native_price_estimator: Arc<dyn NativePriceEstimating>,
     enable_presign_orders: bool,
     solvable_orders: Arc<SolvableOrdersCache>,
     solvable_orders_max_update_age: Duration,
     order_validator: Arc<OrderValidator>,
+    event_updater: Arc<dyn EventUpdating>,
 }
 
 impl Orderbook {
@@ -53,20 +64,24 @@ impl Orderbook {
         settlement_contract: H160,
         database: Arc<dyn OrderStoring>,
         bad_token_detector: Arc<dyn BadTokenDetecting>,
+        native_price_estimator: Arc<dyn NativePriceEstimating>,
         enable_presign_orders: bool,
         solvable_orders: Arc<SolvableOrdersCache>,
         solvable_orders_max_update_age: Duration,
         order_validator: Arc<OrderValidator>,
+        event_updater: Arc<dyn EventUpdating>,
     ) -> Self {
         Self {
             domain_separator,
             settlement_contract,
             database,
             bad_token_detector,
+            native_price_estimator,
             enable_presign_orders,
             solvable_orders,
             solvable_orders_max_update_age,
             order_validator,
+            event_updater,
         }
     }
 
@@ -194,6 +209,24 @@ impl Orderbook {
         Ok(solvable_orders)
     }
 
+    pub async fn get_auction(&self) -> Result<Auction> {
+        let last_handled_block = self.event_updater.last_handled_block().await;
+        let solvable_orders = self.get_solvable_orders().await?;
+
+        let block = last_handled_block.unwrap_or(solvable_orders.latest_settlement_block);
+        let (orders, prices) = filter_tokens_without_native_prices(
+            solvable_orders.orders,
+            &*self.native_price_estimator,
+        )
+        .await;
+
+        Ok(Auction {
+            block,
+            orders,
+            prices,
+        })
+    }
+
     pub async fn get_user_orders(
         &self,
         owner: &H160,
@@ -243,13 +276,58 @@ fn set_available_balances(orders: &mut [Order], cache: &SolvableOrdersCache) {
     }
 }
 
+async fn filter_tokens_without_native_prices(
+    mut orders: Vec<Order>,
+    native_price_estimator: &dyn NativePriceEstimating,
+) -> (Vec<Order>, BTreeMap<H160, U256>) {
+    let traded_tokens = orders
+        .iter()
+        .flat_map(|order| {
+            [
+                order.order_creation.sell_token,
+                order.order_creation.buy_token,
+            ]
+        })
+        .collect::<HashSet<_>>()
+        .into_iter()
+        .collect::<Vec<_>>();
+    let prices = native_price_estimator
+        .estimate_native_prices(&traded_tokens)
+        .await
+        .into_iter()
+        .zip(traded_tokens)
+        .filter_map(|(price, token)| match price {
+            Ok(price) => Some((token, to_normalized_price(price))),
+            Err(err) => {
+                tracing::warn!(?token, ?err, "error estimating native token price");
+                None
+            }
+        })
+        .collect::<BTreeMap<_, _>>();
+
+    orders.retain(|order| {
+        prices.contains_key(&order.order_creation.sell_token)
+            && prices.contains_key(&order.order_creation.buy_token)
+    });
+
+    (orders, prices)
+}
+
+fn to_normalized_price(price: f64) -> U256 {
+    U256::from_f64_lossy(price * 1e18)
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
     use ethcontract::H160;
     use futures::FutureExt;
+    use maplit::{btreemap, hashset};
     use model::order::OrderBuilder;
-    use shared::bad_token::list_based::ListBasedDetector;
+    use shared::{
+        bad_token::list_based::ListBasedDetector,
+        price_estimation::{native::MockNativePriceEstimating, PriceEstimationError},
+    };
 
     #[test]
     fn filter_unsupported_tokens_() {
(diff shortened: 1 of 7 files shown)


crates/orderbook/src/orderbook.rs
```diff
@@ -243,13 +276,58 @@ fn set_available_balances(orders: &mut [Order], cache: &SolvableOrdersCache) {
     }
 }
 
+async fn filter_tokens_without_native_prices(
```
MartinquaXD, line 279: Maybe `retain_orders_with_native_prices()` would be better?

crates/orderbook/src/orderbook.rs
```diff
@@ -277,4 +355,72 @@ mod tests {
...
+            token3 => 2.0,
+        };
+
+        let mut native_price_estimator = MockNativePriceEstimating::new();
+        native_price_estimator
+            .expect_estimate_native_prices()
+            // deal with undeterministic ordering of `HashSet`.
+            .withf(move |tokens| {
+                tokens.iter().cloned().collect::<HashSet<_>>() == hashset!(token1, token2, token3)
+            })
```
MartinquaXD, line 398: This is a little unfortunate but changing the implementation of `filter_tokens_without_native_prices` so you can compare against `&[H160]` doesn't seem worth it.

crates/orderbook/src/orderbook.rs
```diff
@@ -277,4 +355,72 @@ mod tests {
...
+        );
+    }
+
+    #[tokio::test]
+    async fn filters_tokens_without_native_prices() {
+        let token1 = H160([1; 20]);
+        let token2 = H160([2; 20]);
+        let token3 = H160([3; 20]);
+
+        let orders = vec![
```
MartinquaXD, line 373: nit: I think covering all possibilities would be nice. That would need a test where neither `sell_token` nor `buy_token` of an order has a native price.

MartinquaXD (commented):
Looks good overall just 2 minor suggestions.

crates/orderbook/src/orderbook.rs
nlordell: Its just a unit test so I'm not too worried about optimizing it.

The other option is to include a:
```rust
#[cfg(test)]
let traded_tokens = {
    let mut tokens = traded_tokens;
    tokens.sort_unstable();
    tokens
};
```

In the implementation. I slightly prefer making the mock code a little dirtier than sorting the token list when building in `test`, but I have no strong feelings either way.

crates/orderbook/src/orderbook.rs
nlordell: Here I prefer `filter_` just because the signature is `(Vec<Order>, NativePriceEstimating) -> (Vec<Order>, BTreeSet<H160, U256>)`. A name of `retain_` would suggest to me a signature of `(&mut Vec<Order>, NativePriceEstimating) -> BTreeSet<H160, U256>`.

I'm fine with renaming it, but then I would also suggest changing the signature as well to better match the name.

crates/orderbook/src/orderbook.rs
nlordell: Good suggestion! Added.

crates/orderbook/src/orderbook.rs
MartinquaXD: I agree with you and I don't see a need to change anything here.
This was just something that stood out to me.

crates/orderbook/src/orderbook.rs
MartinquaXD: I see your point.
Naming this function is a little hard because it does multiple things:
It filters orders (1) based on their tokens (2) and also returns prices (3).

Maybe `get_orders_with_native_prices` could also be an option?
`get` => to avoid the signature argument (because the current signature is fine IMO)
`orders` => because it operates on orders instead of `tokens`
`with_native_prices` => can be interpreted as "only orders which have native prices" or "return orders but also prices" (both interpretations seem fine to me)

nlordell: I like it! Man, naming things really is the hardest part of software development...

nlordell: Applied your suggestion.

MartinquaXD (approved):

crates/orderbook/src/orderbook.rs
```diff
@@ -194,6 +209,22 @@ impl Orderbook {
         Ok(solvable_orders)
     }
 
+    pub async fn get_auction(&self) -> Result<Auction> {
+        let last_handled_block = self.event_updater.last_handled_block().await;
+        let solvable_orders = self.get_solvable_orders().await?;
```
vkgnosis, line 214: It would be nice to synchronize last_handled_block with solvable orders. If you look in SolvableOrders.rs, it gets from the database the orders and last_settlement_block in one transaction. We can extend that to also include the last event block from EventStoring::last_event_block.
The way it is currently implemented the orders and event block can get out of sync and getting the last handled block from the new EventUpdating trait has to go through a mutex which feels weird.
Implementing this might be somewhat annoying because we have to share the last event query with both places so has to be a pub str somewhere.

nlordell: Note that, AFAIU, `last_handled_block != last_event_block`. Specifically, the last handled block is the last block that was used to request events from the node:
```rust
self.last_handled_block = Some(range.end().to_u64());
```

While `last_event_block` is:
```sql
SELECT GREATEST(
  (SELECT COALESCE(MAX(block_number), 0) FROM trades),
  (SELECT COALESCE(MAX(block_number), 0) FROM settlements),
  (SELECT COALESCE(MAX(block_number), 0) FROM invalidations),
  (SELECT COALESCE(MAX(block_number) , 0) FROM presignature_events)
);
```

This means that even if there are no events which we process for a long time, we would still have `last_handled_block ~= "latest"` (which is what I intended from the auction end point, specifically because prices will be around this block number and if we are too far in the past then the prices are not necessarily valid).

We could also fetch the last event block with the orders (instead of using just the last settlement block as it does now) for the case where the event updater hasn't updated at all yet, which would be slightly more correct. I briefly touched on this in the PR description:

> In case the event handler has not processed any blocks - it returns the block number of the last settlement (as the `v2/solvable_orders` endpoint). This should only happen right on startup so I'm not particularly worried about optimizing the case where no events have been processed.

nlordell:
I got the prices backwards 🙈. I noticed when I did the manual test.

crates/orderbook/src/orderbook.rs
vkgnosis: Your explanation makes sense. I missed the distinction.

crates/orderbook/src/orderbook.rs
```diff
@@ -243,13 +274,61 @@ fn set_available_balances(orders: &mut [Order], cache: &SolvableOrdersCache) {
...
+
+    (orders, prices)
+}
+
+fn to_normalized_price(price: f64) -> U256 {
+    // NOTE: The `NativePriceEstimating` component returns prices denominated
+    // in the token that it is estimating and not in ETH. This means that we
+    // need to invert the price in order for it to be correct.
+    U256::from_f64_lossy(1e18 / price)
+}
```
vkgnosis, line 319: This can panic if the price is 0 which is allowed by the price estimator interface. I think we should filter 0 prices out.
As a second point, maybe it makes sense to also guarantee that prices in the auction endpoint itself are never zero so we would have to check here that the U256 didn't round to 0.

nlordell: Changed to check that the resulting `U256` prices are valid.

crates/orderbook/src/orderbook.rs
```diff
@@ -243,13 +274,68 @@ fn set_available_balances(orders: &mut [Order], cache: &SolvableOrdersCache) {
...
+fn to_normalized_price(price: f64) -> Option<U256> {
+    let uint_max = 2.0_f64.powi(256);
+
+    // NOTE: The `NativePriceEstimating` component returns prices denominated
+    // in the token that it is estimating and not in ETH. This means that we
+    // need to invert the price in order for it to be correct.
+    let price_in_eth = 1e18 / price;
+    if price_in_eth < 1. || price_in_eth >= uint_max {
+        return None;
+    }
```
vkgnosis, line 323: One more nit on this. Testing on my local machine 1e18 / 0.0 is INF. But I don't know if that is guaranteed. If the result was NAN then the if condition would be false and we would continue with U256::from_f64_lossy(NAN). So maybe should check for 0 price explicitly.

vkgnosis (approved):

crates/orderbook/src/orderbook.rs
nlordell: According to the IEEE-754 standard page on Wikipedia:

> Division by zero: an operation on finite operands gives an exact infinite result, e.g., 1/0 or log(0). By default, returns ±infinity.

Note that `0. / 0. is NaN` though!

nlordell: Made this stricter. I additionally added a `price_in_eth.is_normal()` condition so that the price can't be `NaN`, zero, subnormal, or infinite.

Issue #1359, opened 2021-11-07
Price estimations take a long time


josojo:
Marco [highlighted](https://gnosisinc.slack.com/archives/C020UHS2UEL/p1636228074017700) that many instances have little time to left.

I think it is due to the fact that price estimations quite often take a long time. If one looks through the logs:
https://logs.gnosisdev.com/app/kibana#/discover?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15m,to:now))&_a=(columns:!(_source),filters:!(),index:de1f5360-a9ae-11ea-acbb-9d2e3cb124ca,interval:auto,query:(language:kuery,query:'log:%22estimate_prices%20took%22'),sort:!())
one sees many instances with >=5 secs.

Here is one example, where there was little time to solve the instance: 
https://logs.gnosisdev.com/app/kibana#/discover/doc/de1f5360-a9ae-11ea-acbb-9d2e3cb124ca/logstash-2021.10.18?id=apyrlHwBGsivd2V8bgsx

and just before the price estimation took 11 secs:
https://logs.gnosisdev.com/app/kibana#/discover/doc/de1f5360-a9ae-11ea-acbb-9d2e3cb124ca/logstash-2021.10.18?id=dQarlHwBIBf7G29ggfyG

nlordell:
Looking into this a bit, it looks like some of the batched `eth_call` requests are taking a long time. This one took 3 seconds for example: https://logs.gnosisdev.com/app/kibana#/discover?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2021-10-18T18:00:00.000Z',to:'2021-10-18T19:00:00.000Z'))&_a=(columns:!(_source),filters:!(),index:de1f5360-a9ae-11ea-acbb-9d2e3cb124ca,interval:auto,query:(language:kuery,query:'log:%22102854%22'),sort:!())

nlordell:
I would also expect `eth_calls` to be cached (they are retrieving Uniswap pool data here for example) but they appear not to be.

vkgnosis:
I would expect the same since #1137 .

vkgnosis:
Since #1137 we should fetch all pools in one go. I'm not sure about why they're not cached. The baseline price estimator is indicating it is OK with "recent' results for the pools.

vkgnosis:
It is expected that there are some uncached pools btw for the http solver. Because it needs more price estimates than the driver.

[...]

PR #1650, opened 2022-02-15
Implement RacingCompetitionPriceEstimator

MartinquaXD:
Related to: #1359 

Right now it takes quite a bit of time until a user gets a first price estimation for their trade.
Unfortunately it looks as if there is no easy way to speed up price estimation by simply tweaking the query parameters we use. One alternative solution could be to offer a way to get faster but less optimal price estimations. The frontend could request a fast and an optimal price at the same time and show the fast one until the optimal one arrives.

My initial idea was to just put a time limit on the price estimation and use the best result we got so far. This turned out to be problematic because there are no good ways to configure that time limit. Sometimes even our faster estimators take quite a bit of time to find a solution and if we configure the time limit to be high enough to get a result from the fast ones most  of the time, we waste quite a bit on time on average.

Another option which seems to perform better is to let the estimators race each other and return a result early when there is at least a (configurable) number of successful estimates for every query.
This is nicer because it's way simpler to configure a "good" threshold (compared to the time limit solution). I found 2 results to return early quite nice (if all estimators are enabled). Smaller orders usually get estimated by the faster (on average) `Baseline` and `Quasimodo` estimators whereas bigger orders cause `Quasimodo` to return `NoLiquidity` so one of the other "bigger" estimators has to give an estimation instead. But it looks like `Paraswap`, `ZeroEx` and `OneInch` are faster on bigger orders so that cancels our nicely.
Based on some manual tests this would bring "fast" sell order price estimations consistently below 1s (where smaller orders are around 200-500ms and bigger ones around 600-900ms). Big buy orders perform worse because more estimators returns errors.

Notes
- to properly test returning fast results early I had to make the mocked `PriceEstimating` trait return a future instead of a normal value. (This is also closer to the reality)
- a follow up PR is supposed to make the `/quote` endpoint differentiate between "fast" and "optimal" and  actually use the new estimator

### Test Plan
Unit test for correctness of `RacingCompetitionPriceEstimator`
Manual test to see viability of this approach

### DIFF
--- a/crates/shared/src/price_estimation.rs
+++ b/crates/shared/src/price_estimation.rs
@@ -105,8 +105,8 @@ impl Estimate {
     }
 }
 
-#[mockall::automock]
 #[async_trait::async_trait]
+#[mockall::automock]
 pub trait PriceEstimating: Send + Sync {
     async fn estimate(&self, query: &Query) -> Result<Estimate, PriceEstimationError> {
         self.estimates(std::slice::from_ref(query))
--- a/crates/shared/src/price_estimation/competition.rs
+++ b/crates/shared/src/price_estimation/competition.rs
@@ -247,6 +247,8 @@ mod tests {
     use anyhow::anyhow;
     use model::order::OrderKind;
     use primitive_types::H160;
+    use std::time::Duration;
+    use tokio::time::sleep;
 
     #[tokio::test]
     async fn works() {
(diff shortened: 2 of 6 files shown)


vkgnosis (approved):
LGTM and I like your explanation of this design.

Optional idea I had is to implement the non racing competition estimator as a wrapper around racing with successful_count == inner.len(). Not sure if that is better.

crates/shared/src/price_estimation.rs
```diff
@@ -105,8 +105,8 @@ impl Estimate {
     }
 }
 
-#[mockall::automock]
 #[async_trait::async_trait]
+#[mockall::automock]
```
nlordell, line 109: Funny how significant this small change actually is!

nlordell (approved):
Looks great.

Also, very nice PR description and research 👏.

MartinquaXD:
> Optional idea I had is to implement the non racing competition estimator as a wrapper around racing with successful_count == inner.len(). Not sure if that is better.

I also thought about that but I was a little hesitant to add unnecessary work to the `CompetitionPriceEstimator` (e.g. `join_all` vs checking if we have enough estimates for every estimator) but that really doesn't matter in the grand scheme of things while allowing us to drop a little bit of duplicated code.
The last commit now implements that idea. Let me know if anyone has concerns with that and I will revert it.
I don't have strong feelings about that.

crates/shared/src/price_estimation.rs
MartinquaXD: To be honest I'm a little surprised that the old version even worked considering that async traits aren't stable yet.
BTW they specifically mention `async_trait` in their [docs](https://docs.rs/mockall/latest/mockall/#async-traits).

crates/shared/src/price_estimation/competition.rs
```diff
@@ -2,76 +2,146 @@ use super::{Estimate, PriceEstimating, PriceEstimationError, Query};
...
+                            name,
+                            previous_result,
+                            estimates[i].clone(),
+                        )
+                    },
+                )
+                .map(|winning_estimate| {
+                    // The `EstimateData::estimator_name` field is getting
+                    // flagged as dead code despited being used in the log
+                    // below. Just convince the linter that we use it.
```
nlordell, line 130: I believe this comment is outdated now that we use `estimator_name` with the metrics. Can probably be removed now.

nlordell (approved):
Looks good to me!

Also a fan of fewer code paths 😄.

PR #1651, opened 2022-02-15
Verify that external single order solvers do not violate limit price

vkgnosis:
As on call I was investigating

> 2022-02-15T06:31:09.418Z ERROR solver::settlement: Overflow computing objective value

and could not understand how this could happen.

Eventually I found the following exchange with 0x:
<details>

> 2022-02-15T06:30:55.692Z DEBUG shared::zeroex_api: Querying 0x API: https://gated.api.0x.org/swap/v1/quote?sellToken=0x7ae1d57b58fa6411f32948314badd83583ee0e8c&buyToken=0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2&slippagePercentage=0.003&buyAmount=2000000000000000000&affiliateAddress=0x9008D19f58AAbD9eD0D60971565AA8510560ab41&skipValidation=true&intentOnFilling=false

> 2022-02-15T06:30:56.354Z DEBUG shared::zeroex_api: Response from 0x API: {"chainId":1,"price":"147052.264902334297684418","guaranteedPrice":"147493.421697041300577472","estimatedPriceImpact":"2.5337","to":"0xdef1c0ded9bec7f1a1670819833240f027b25eff","data":"0x6af479b2000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000300cf9dc4d42af9360410000000000000000000000000000000000000000000000001559b69e3c99d88f0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002b7ae1d57b58fa6411f32948314badd83583ee0e8c002710c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2000000000000000000000000000000000000000000869584cd0000000000000000000000009008d19f58aabd9ed0d60971565aa8510560ab410000000000000000000000000000000000000000000000ec58c55eff620b48a0","value":"0","gas":"126000","estimatedGas":"126000","gasPrice":"45000000000","protocolFee":"0","minimumProtocolFee":"0","buyTokenAddress":"0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2","sellTokenAddress":"0x7ae1d57b58fa6411f32948314badd83583ee0e8c","buyAmount":"1538461538461538447","sellAmount":"226234253695898917376575","sources":[{"name":"0x","proportion":"0"},{"name":"Uniswap","proportion":"0"},{"name":"Uniswap_V2","proportion":"0"},{"name":"Eth2Dai","proportion":"0"},{"name":"Kyber","proportion":"0"},{"name":"Curve","proportion":"0"},{"name":"Balancer","proportion":"0"},{"name":"Balancer_V2","proportion":"0"},{"name":"Bancor","proportion":"0"},{"name":"mStable","proportion":"0"},{"name":"Mooniswap","proportion":"0"},{"name":"Swerve","proportion":"0"},{"name":"SnowSwap","proportion":"0"},{"name":"SushiSwap","proportion":"0"},{"name":"Shell","proportion":"0"},{"name":"MultiHop","proportion":"0"},{"name":"DODO","proportion":"0"},{"name":"DODO_V2","proportion":"0"},{"name":"CREAM","proportion":"0"},{"name":"LiquidityProvider","proportion":"0"},{"name":"CryptoCom","proportion":"0"},{"name":"Linkswap","proportion":"0"},{"name":"Lido","proportion":"0"},{"name":"MakerPsm","proportion":"0"},{"name":"KyberDMM","proportion":"0"},{"name":"Smoothy","proportion":"0"},{"name":"Component","proportion":"0"},{"name":"Saddle","proportion":"0"},{"name":"xSigma","proportion":"0"},{"name":"Uniswap_V3","proportion":"1"},{"name":"Curve_V2","proportion":"0"},{"name":"ShibaSwap","proportion":"0"},{"name":"Synapse","proportion":"0"}],"orders":[{"makerToken":"0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2","takerToken":"0x7ae1d57b58fa6411f32948314badd83583ee0e8c","makerAmount":"1538461538461538447","takerAmount":"226234253695898917376575","fillData":{"router":"0xe592427a0aece92de3edee1f18e0157c05861564","tokenAddressPath":["0x7ae1d57b58fa6411f32948314badd83583ee0e8c","0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"],"uniswapPath":"0x7ae1d57b58fa6411f32948314badd83583ee0e8c002710c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"},"source":"Uniswap_V3","sourcePathId":"0x1b6352b938195de9d2d72d38bfc0f6fe14e99d9ddebed329265c11ae49dd6cf9","type":0}],"allowanceTarget":"0xdef1c0ded9bec7f1a1670819833240f027b25eff","sellTokenToEthRate":"143326.5157040987654247","buyTokenToEthRate":"1"}

</details>

The original user order here is a buy order with sell amount 2.27e23 and buy amount 2.00e18. We forward this to 0x with a query setting the buy amount to 2.00e18 . The response we get back has sell amount 2.26e23 and buy amount 1.54e18.
The returned buy amount is significantly less than what we asked for in the buy order. This is a problem.
First, we asked for a fill-or-kill order buy order. It does not make sense that the resulting buy amount is less than our input buy amount.
Second, because we assume 0x respects our wish for a fill-or-kill order, the limit price check is simplified to `executed_sell_amount <= order_sell_amount` (for a buy order). This is incorrect if the executed buy amount is not at least as large as the order buy amount.
The first case is benign. Either the settlement will fail to simulate because there is not enough buy token or we will unintentionally use the buffers in the settlement contract.
The second case is malignant. Like in the first case it is possible that we would be unintentionally using the buffers. But here the order limit price has not been respected. We are paying the user their limit price out of our buffers even though the on chain price does not match.

I do not know if this ever really happened. In the case I was investigating the 0x solution did not win the competition or failed to simulate (don't remember which). If the situation occurs we should always see the "Overflow computing objective value" error which is caused by `settlement.rs` `fn (buy|sell)_order_surplus` returning None when the surplus is negative which always happens when the price is not respected.
Note that the error does not cause the settlement to be disregarded, the surplus is treated as 0.

To fix this I changed all single order solvers to use the same validation function which always compares both sell and buy amounts.

### Test Plan

Code still compiles. The new function is simple enough that a test is not useful (but please look at it carefully).

### DIFF
--- a/crates/solver/src/solver/balancer_sor_solver.rs
+++ b/crates/solver/src/solver/balancer_sor_solver.rs
@@ -1,7 +1,7 @@
 //! Solver using the Balancer SOR.
 
 use super::{
-    single_order_solver::{SettlementError, SingleOrderSolving},
+    single_order_solver::{execution_respects_order, SettlementError, SingleOrderSolving},
     Auction,
 };
 use crate::{
@@ -76,15 +76,16 @@ impl SingleOrderSolving for BalancerSorSolver {
             }
         };
 
-        if !quote_respects_limit_price(&quote, &order) {
-            tracing::debug!("Order limit price not respected");
-            return Ok(None);
-        }
-
         let (quoted_sell_amount, quoted_buy_amount) = match order.kind {
             OrderKind::Sell => (quote.swap_amount, quote.return_amount),
             OrderKind::Buy => (quote.return_amount, quote.swap_amount),
         };
+
+        if !execution_respects_order(&order, quoted_sell_amount, quoted_buy_amount) {
+            tracing::debug!("execution does not respect order");
+            return Ok(None);
+        }
+
         let (quoted_sell_amount_with_slippage, quoted_buy_amount_with_slippage) = match order.kind {
             OrderKind::Sell => (
                 quoted_sell_amount,
@@ -135,17 +136,6 @@ impl SingleOrderSolving for BalancerSorSolver {
 
     fn name(&self) -> &'static str {
         "BalancerSOR"
-    }
-}
-
-fn quote_respects_limit_price(quote: &Quote, order: &LimitOrder) -> bool {
-    match order.kind {
-        OrderKind::Sell => {
-            quote.swap_amount == order.sell_amount && quote.return_amount >= order.buy_amount
-        }
-        OrderKind::Buy => {
-            quote.return_amount <= order.sell_amount && quote.swap_amount == order.buy_amount
-        }
     }
 }
 
@@ -245,76 +235,6 @@ mod tests {
         Web3,
     };
     use std::env;
-
-    #[test]
-    fn check_limit_prices_for_sell_order() {
-        let sell_order = LimitOrder {
-            kind: OrderKind::Sell,
-            sell_amount: 10.into(),
-            buy_amount: 10.into(),
-            ..Default::default()
-        };
-
-        assert!(quote_respects_limit_price(
-            &Quote {
-                swap_amount: 10.into(),
-                return_amount: 11.into(),
-                ..Default::default()
-            },
-            &sell_order,
-        ));
-        assert!(!quote_respects_limit_price(
-            &Quote {
-                swap_amount: 10.into(),
-                return_amount: 9.into(),
-                ..Default::default()
-            },
-            &sell_order,
-        ));
-        assert!(!quote_respects_limit_price(
-            &Quote {
-                swap_amount: 9.into(), // not fill or kill!
-                return_amount: 10.into(),
-                ..Default::default()
-            },
-            &sell_order,
-        ));
-    }
-
-    #[test]
-    fn check_limit_prices_for_buy_order() {
-        let buy_order = LimitOrder {
-            kind: OrderKind::Buy,
-            buy_amount: 10.into(),
-            sell_amount: 10.into(),
-            ..Default::default()
-        };
-
-        assert!(quote_respects_limit_price(
-            &Quote {
-                swap_amount: 10.into(),
-                return_amount: 9.into(),
-                ..Default::default()
-            },
-            &buy_order,
-        ));
-        assert!(!quote_respects_limit_price(
-            &Quote {
-                swap_amount: 10.into(),
-                return_amount: 11.into(),
-                ..Default::default()
-            },
-            &buy_order,
-        ));
-        assert!(!quote_respects_limit_price(
-            &Quote {
-                swap_amount: 11.into(), // not fill or kill!
-                return_amount: 10.into(),
-                ..Default::default()
-            },
-            &buy_order,
-        ));
-    }
 
     #[test]
     fn computed_swap_sets_sign() {
--- a/crates/solver/src/solver/paraswap_solver.rs
+++ b/crates/solver/src/solver/paraswap_solver.rs
@@ -1,5 +1,5 @@
 use super::{
-    single_order_solver::{SettlementError, SingleOrderSolving},
+    single_order_solver::{execution_respects_order, SettlementError, SingleOrderSolving},
     Auction,
 };
 use crate::{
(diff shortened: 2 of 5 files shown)


crates/solver/src/solver/paraswap_solver.rs
```diff
@@ -93,8 +93,12 @@ impl SingleOrderSolving for ParaswapSolver {
             .get_token_infos(&[order.sell_token, order.buy_token])
             .await;
         let (price_response, amount) = self.get_price_for_order(&order, &token_info).await?;
-        if !satisfies_limit_price(&order, &price_response) {
-            tracing::debug!("Order limit price not respected");
+        if !execution_respects_order(
+            &order,
+            price_response.src_amount,
+            price_response.dest_amount,
```
vkgnosis, line 99: TODO: Have to double check this is right. I think we might have to select based on order_kind here but in that case wasn't the original code wrong too?

nlordell: I think this is correct. `src_token` always refers to the token being sold and `dest_token` to the token being bought. Reference: https://developers.paraswap.network/api/get-rate-for-a-token-pair

Specifically the documentatino for the `amount` parameter makes it quite clear:
> srcToken amount (in case of SELL) or destToken amount (in case of BUY).

nlordell:
Nice catch!

I think this is missing the `BalancerSor` solver though:
https://github.com/gnosis/gp-v2-services/blob/01e1983853179b47aedd783f46eb95357893d405/crates/solver/src/solver/balancer_sor_solver.rs#L79-L82

vkgnosis:
added

crates/solver/src/solver/balancer_sor_solver.rs
```diff
@@ -76,8 +76,12 @@ impl SingleOrderSolving for BalancerSorSolver {
             }
         };
 
-        if !quote_respects_limit_price(&quote, &order) {
-            tracing::debug!("Order limit price not respected");
+        let (executed_sell, executed_buy) = match order.kind {
```
nlordell, line 79: nit: We already compute:
```rust
        let (quoted_sell_amount, quoted_buy_amount) = match order.kind {
            OrderKind::Sell => (quote.swap_amount, quote.return_amount),
            OrderKind::Buy => (quote.return_amount, quote.swap_amount),
        };
```

A couple lines below, maybe we don't need to re-match on `order.kind` and only do this once?

crates/solver/src/solver/single_order_solver.rs
```diff
@@ -105,6 +105,20 @@ impl From<anyhow::Error> for SettlementError {
...
+// amount for sell orders) is not enough.
+pub fn execution_respects_order(
+    order: &LimitOrder,
+    executed_sell_amount: U256,
+    executed_buy_amount: U256,
+) -> bool {
+    // note: This would be different for partially fillable orders but LimitOrder does currently not
+    // contain the remaining fill amount.
+    executed_sell_amount <= order.sell_amount && executed_buy_amount >= order.buy_amount
+}
```
nlordell, line 120: We can copy the tests from `balancer_sor.rs` here:
```rust
    #[test]
    fn test_satisfies_limit_price() {
        // ...
    }
```

vkgnosis: I'm just wondering if that is useful. This function is very simple so I don't feel a test will catch an error. Either the logic is right or I will make the same mistake in the test too. Don't mind adding it if other people think it is useful.

crates/solver/src/solver/balancer_sor_solver.rs
vkgnosis: Somehow missed that. Fixed.

crates/solver/src/solver/single_order_solver.rs
nlordell: Agree in general that this function is quite simple, but since the was already there, it is useful for verifying that there are no regressions with the refactor.

vkgnosis: Added tests. One (expected) difference is that the new function allows the input amount to unexpectedly be less as long as the out amount is still respected. The sor solver was more strict than necessary here even though it would be weird for the api to return that.

nlordell (approved):

MartinquaXD (approved):
LGTM

PR #1652, opened 2022-02-15
`NativePriceEstimating` return prices denominated in Ether

nlordell:
This PR changes the `NativePriceEstimating` implementation to return query token prices denominated in native token instead of prices of the native token denominated in query tokens.

I noticed this when implementing #1649 and thought it made slightly more sense to have native token prices be denominated in native token. That is:
```rust
let prices = native_price_estimator.estimate_native_prices([foo, bar]);
```

This will give us a price vector that are in the same unit instead of having a different unit per estimate (meaning that comparisons of prices returned from the request are meaningful):
```rust
prices[0] = price_of_foo_in_eth;
prices[1] = price_of_bar_in_eth;
```

I don't feel strongly about this, it just confused me a bit when implementing `v1/auction` endpoint and wanted to propose this change to see what you all think. I'm also OK with just closing this PR if you don't think its necessary.

### Test Plan

CI and existing unit tests.

### DIFF
--- a/crates/orderbook/src/fee.rs
+++ b/crates/orderbook/src/fee.rs
@@ -102,7 +102,7 @@ pub struct FeeParameters {
 // 2. When validating fees this consistently picks the same amount.
 impl FeeParameters {
     pub fn amount_in_sell_token(&self) -> U256 {
-        U256::from_f64_lossy((self.gas_amount * self.gas_price * self.sell_token_price).ceil())
+        U256::from_f64_lossy((self.gas_amount * self.gas_price / self.sell_token_price).ceil())
     }
 
     fn apply_fee_factor(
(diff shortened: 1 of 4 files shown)


crates/orderbook/src/fee.rs
```diff
@@ -837,7 +839,7 @@ mod tests {
             ..Default::default()
         };
         let native_token_price_estimation_amount = 100.;
-        let sell_token_price = 1.25;
+        let sell_token_price = 0.1;
```
nlordell, line 842: Not an actually needed change - but if you make the sell token less valuable than the native token then it avoid rounding errors in the asserts 👇.

vkgnosis (approved):

MartinquaXD (approved):

PR #1655, opened 2022-02-15
Use new auction endpoint for solving

nlordell:
This PR makes the solver use the new `v1/auction` endpoint as well as prices from the orderbook for settlement competition. These prices are used instead of the prices that used to be computed internally in the solver and result in quite a bit of removed code (🎉).

Of note, the `augument_prices` preprocessing step has been removed (this was briefly discussed in #1649). The rationale for this is that augmenting prices should no longer be needed since we used to be able to accept orders in the orderbook without a baseline price (because the orderbook uses a different set of estimators). Because the `estimated_prices` computed in the driver came from baseline prices, this caused some issues where orders could be added to the orderbook but not be considered by the driver. To remedy this, the `augment_prices` preprocessing step was added to augment the `external_prices` vector under certain conditions - allowing these orders to be traded. Since the external price vector comes from the orderbook now (and uses the same set of native prices used for computing fees when adding orders - i.e. they should always exist), it is no longer needed.

### Test Plan

CI. I will additionally run manual tests and update the PR once they are done. One concern that I thought about that I want to make sure it works are ETH/WETH orders (I don't remember seeing any special paths in the native price estimation to handle them and want to make sure that they don't get filtered out).

### Release Notes

Since this relies on new APIs that aren't yet in production, this PR will break the shadow solver until next release.


crates/solver/src/auction_preprocessing.rs
```diff
@@ -1,345 +1,9 @@
...
-            removed_orders.len(),
-            removed_orders,
-        );
-    }
-    orders
-}
-
-// In case we cannot estimate the price for a token but there is only one pair traded in the batch the accuracy of the price estimate for ranking a solution doesn't matter that much (surplus is not competing with other tokens).
-// Therefore we can deduct a price estimate by looking at the average limit price of the orders on the single token pair
-fn augment_prices(orders: &[LimitOrder], prices: &mut HashMap<H160, BigRational>) {
```
MartinquaXD, line 97: I think this currently doesn't happen in the `/auction` endpoint.
This optimization doesn't seem very important to me so I would be fine with dropping it completely. I just thought that this should be highlighted. Maybe somebody else has a different opinion. 🤷‍♂️

crates/solver/src/auction_preprocessing.rs
```diff
@@ -1,345 +1,9 @@
...
-            );
-        }
-    }
-}
-
 // vk: I would like to extend this to also check that the order has minimum age but for this we need
 // access to the creation date which is a more involved change.
 pub fn has_at_least_one_user_order(orders: &[LimitOrder]) -> bool {
     orders.iter().any(|order| !order.is_liquidity_order)
 }
```
MartinquaXD, line 9: This is the only function left in the file. It's only used at one place in the driver. I think it's simple enough that we can inline it there or maybe move it there to get rid of this file entirely.

MartinquaXD (commented):
I'm really happy about how much code can be dropped. 🧹

crates/solver/src/auction_preprocessing.rs
nlordell: Sorry! There is some missing context that I had included in #1649:

> We no longer need to augment prices in the driver (since, not only do we use the native token estimator which supports more tokens, additionally orders for which there are no native prices shouldn't get added to the orderbook anymore, and finally we also filter the orders before sending them to the driver so augmenting would never really do anything anyway).

crates/solver/src/auction_preprocessing.rs
nlordell: I also wanted to do that, just didn't want to add noise to the review.

As you said since it is only a small 1-line function I think its fine to just move to the `driver` module as you suggested.

crates/solver/src/auction_preprocessing.rs
MartinquaXD: Forgot about that. Thanks for reiterating. 👍

crates/solver/src/driver.rs
```diff
@@ -374,35 +368,41 @@ impl Driver {
...
             .collect::<Vec<_>>();
         tracing::info!("got {} orders: {:?}", orders.len(), orders);
+
+        let estimated_prices = auction
+            .prices
+            .into_iter()
+            .map(|(token, price)| {
+                (
+                    token,
+                    price.to_big_rational() / BigInt::from(1_000_000_000_000_000_000_u128),
```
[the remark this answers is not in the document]
nlordell: My understanding of the `estimated_prices` were that they were exchange rates from Ether to the token. Since the prices are denominated in Ether with 18 decimals (i.e. the amount of Ether required to buy `1e18` of the token) we need to divide by `1e18` to get an actual exchange rate (such that `X ETH * estimated_prices[TOK] = Y TOK`).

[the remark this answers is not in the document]
nlordell: Looks correct:
```rust
        .map(|token| price_estimation::Query {
            // For ranking purposes it doesn't matter how the external price vector is scaled,
            // but native_token is used here anyway for better logging/debugging.
            sell_token: native_token,
            buy_token: token,
            in_amount: native_token_amount_to_estimate_prices_with,
            kind: model::order::OrderKind::Sell,
        })

    // ...

            let price = match estimate.price_in_sell_token_rational(&query) {

    // ...

    // Always include the native token.
    prices.insert(native_token, num::one());
    // And the placeholder for its native counterpart.
    prices.insert(BUY_ETH_ADDRESS, num::one());
```

So it looks like the prices in the `estimated_prices` maps are indeed exchange rates from token to ETH. Since the service returns prices in ETH with 18 decimals, this is computed by: `price / 1e18`.

I will add a small helper function to `auction_preprocessing` with comments to make this more obvious.

crates/solver/src/auction_preprocessing.rs
nlordell: Given our discussion ☝️, I will keep this module around and add a helper method for converting prices to exchange rates with tests.

crates/solver/src/driver.rs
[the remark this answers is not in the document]
nlordell: Added a separate method with tests that hopefully clarify the format used in the `estimated_prices` mapping.

crates/solver/src/auction_preprocessing.rs
```diff
@@ -1,9 +1,59 @@
...
+    use super::*;
+    use num::One as _;
+
+    #[test]
+    fn converts_prices_to_exchange_rates() {
+        // By definition, the price of ETH is 1e18 and its xrate is 1.
+        let eth_price = U256::from(1_000_000_000_000_000_000_u128);
+        assert_eq!(to_native_xrate(eth_price), BigRational::one());
+
+        // GNO is typically traded at around Ξ0.1. With the price
```
MartinquaXD, line 42: Didn't know `Ξ` was the symbol for 1 Ether. 👀

MartinquaXD (commented):
Once the CI passes this PR would be good to go from my point of view. Will delay approval until then.

crates/solver/src/auction_preprocessing.rs
nlordell: Its the symbol for Ether as a currency (not a specific quantity). Just like $ is for dollar, this means that Ξ13.37 == `13_370_000_000_000_000_000 wei`.

nlordell:
> Once the CI passes this PR would be good to go from my point of view.

This is related to native price estimates for `0xeee..eee` not working. I have a few small PRs coming up that should fix it. Once they are merged I can rebase this one and hopefully everything will just work™

vkgnosis (approved):

crates/solver/src/auction_preprocessing.rs
```diff
@@ -1,150 +1,57 @@
...
+) -> HashMap<H160, BigRational> {
+    let mut exchange_rates = prices
+        .into_iter()
+        .map(|(token, price)| (token, to_native_xrate(price)))
+        .collect::<HashMap<_, _>>();
+
+    // Ensure there is a price for the native asset and wrapped native token.
+    // While this is not strictly needed since we are returning a mapping of
+    // exchange rates to the native token, certain components in the driver
+    // expect their existence. The value is just 1.0.
```
nlordell, line 41: To me this is a bit of a code smell...

That being said, I didn't feel super comfortable trying to fix at all the locations where it happens. I found one place for sure but its hard to know with confidence if I missed anywhere:

https://github.com/gnosis/gp-v2-services/blob/14c739ad2a4772be348f0f293e9a8740abecf966/crates/solver/src/driver.rs#L338-L342

I created an issue to address this #1666.

nlordell:
@MartinquaXD 

CI should be passing now. This PR introduced a bit of a code smell (documented inline). I created an issue to not forget to come back to it.

MartinquaXD (approved):

crates/solver/src/auction_preprocessing.rs
nlordell: Changed this to return errors on malformed `/auction` prices and proposed #1669 (although, TBH, I'm not 100% happy with the PR).

PR #1656, opened 2022-02-16
Add case in sanitizing price estimator for WETH -> ETH

nlordell:
When implementing #1649 I noticed that native ETH orders aren't working as expected because estimating native token price for `0xeee...eee` are failing. I started looking at how to fix it and noticed that there is an insidious way to get queries where `sell_token == buy_token` by trying to do a price query from `WETH -> ETH`.

Currently, this will never happen because:
- The FE will automatically make unwrap transactions if you try to trade WETH -> ETH
- The `EthAwareMinFeeCalculator` is already swapping normalizing `buy_token == BUY_ETH_ADDRESS` to `buy_token = native_token` for us, so it would always reach the price estimators as a `sell_token: WETH, buy_token: WETH` query anyway. (**NOTE**: This means that the price estimate doesn't seem to be adding extra gas for the WETH unwrap in the price estimates. I will fix this in a follow up).

Note that this still doesn't fix my original problem, but that is a larger fix that happen in pieces (including this one 😄).

### Test Plan

Added this particular case to the existing unit test.

### DIFF
--- a/crates/shared/src/price_estimation/sanitized.rs
+++ b/crates/shared/src/price_estimation/sanitized.rs
@@ -122,7 +122,17 @@ impl PriceEstimating for SanitizedPriceEstimator {
                     return TrivialSolution(Ok(estimation));
                 }
 
-                if query.buy_token == BUY_ETH_ADDRESS {
+                if query.sell_token == self.native_token && query.buy_token == BUY_ETH_ADDRESS {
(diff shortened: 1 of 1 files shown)


MartinquaXD (approved):
Nice catch!

vkgnosis (approved):

PR #1657, opened 2022-02-16
Deprecate `EthAwareMinFeeCalculator`

nlordell:
This PR removes the `orderbook::fee::{EthAdapter, EthAwareMinFeeCalculator}` types. Follow up to #1656.

These types existed for computing minimum order fees for orders taking the `BUY_ETH_ADDRESS` marker value for the `buy_token` into account. However, this is no longer needed since the `SanitizingPriceEstimator` already does this for us **and** adds a gas amount to compensate for the WETH unwrap cost.

With this change, we should still continue supporting native token buy orders as before (i.e. orders where `buy_token == BUY_ETH_ADDRESS`) while charging a small premium on the fee to account for the WETH unwrap interaction that the solvers need to perform.

One downside to this is that we can't take advantage of the cached fee measurement entry in the database between ETH and WETH orders with the same amounts. One alternative would be to also include logic in the `impl MinFeeCalculating for EthAdapter` implementation to add the WETH unwrap gas premium. That being said I slightly prefer the solution here as it removes the necessity of having multiple components aware of `native_token` for the purpose of normalizing `buy_token` for estimates.

### Test Plan

Existing CI - just removed some code. The E2E test that buys ETH should verify that this change does not cause any regressions.

You can also see that it still works with the `/quote` API:
```
% curl -s \
  -X POST \
  --data '@-' \
  <<JSON | jq '.quote | { sellAmount, buyAmount, feeAmount }'
{
  "sellToken": "0xba100000625a3754423978a60c9317c58a424e3d",
  "buyToken": "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee",
  "validTo": 2147483648,
  "kind": "buy",
  "partiallyFillable": false,
  "buyTokenBalance": "erc20"
}
JSON

{
  "sellAmount": "200274328056010037013",
  "buyAmount": "1000000000000000000",
  "feeAmount": "3736440792904630784"
}
```

### DIFF
--- a/crates/e2e/tests/e2e/services.rs
+++ b/crates/e2e/tests/e2e/services.rs
@@ -7,7 +7,7 @@ use orderbook::{
     api::post_quote::OrderQuoter,
     database::Postgres,
     event_updater::EventUpdater,
-    fee::{EthAwareMinFeeCalculator, FeeSubsidyConfiguration},
+    fee::{FeeSubsidyConfiguration, MinFeeCalculator},
     orderbook::Orderbook,
     solvable_orders::SolvableOrdersCache,
 };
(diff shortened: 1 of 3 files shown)


MartinquaXD (approved):
Love it. 🧹 
I thought about doing the same thing when I implemented the native price estimator because it looked like all functionality of `EthAwareMinFeeCalculator` would have been moved elsewhere but a tiny bit of code remained.
I think the one downside you mentioned is worth the reduced code complexity.

vkgnosis (approved):

PR #1658, opened 2022-02-16
Use error types instead of result enums for order operations

nlordell:
This PR refactors the `add_order` and `cancel_order` orderbook operations to return `Result<_, Error>` instead of `Result<ResultEnum>`.

This is mostly a cleanup and doesn't change any behaviour. The rationale behind the change is that:
- The `Result::Ok(ResultEnum::SomeErrorCondition)` actually represent errors (since they return `4xx` and `5xx` error codes) and it made more sense to me to use error types instead of nested result types.
- Allows for `?` operator to be used in the implementation, simplifying things a bit.

I also noticed when refactoring that the `cancel_order` API endpoint returns `200 OK` with a static string. I think it would be slightly more correct to return `204 No Content` but that is a separate change that I don't want to include here (and may require coordinating with the FE team in case they are relying on the current behaviour).

### Test Plan

Existing unit tests continue to pass.

### DIFF
--- a/crates/orderbook/src/api/cancel_order.rs
+++ b/crates/orderbook/src/api/cancel_order.rs
@@ -1,6 +1,6 @@
-use crate::api::extract_payload;
-use crate::orderbook::{OrderCancellationResult, Orderbook};
-use anyhow::{Context, Result};
+use crate::api::{convert_json_response, extract_payload, IntoWarpReply};
+use crate::orderbook::{OrderCancellationError, Orderbook};
+use anyhow::Result;
 use model::signature::EcdsaSignature;
 use model::{
     order::{OrderCancellation, OrderUid},
@@ -8,6 +8,7 @@ use model::{
 };
 use serde::{Deserialize, Serialize};
 use std::{convert::Infallible, sync::Arc};
+use warp::reply::with_status;
 use warp::{hyper::StatusCode, Filter, Rejection};
 
 #[derive(Debug, Deserialize, PartialEq, Serialize)]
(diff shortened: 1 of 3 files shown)


vkgnosis:
The reason for the current behavior was to distinguish internal errors (like not being able to connect to database) from user errors like invalid signature. For internal errors we cannot do anything except bubble them up with `?` but user errors might be handled differently. For Orderbook::AddOrder you could argue that for user errors from the orderbook's pov everything went fine so this shouldn't be Result::Error. It just so happens that the operating the user requested is invalid. But figuring out that it is invalid worked correctly.
That said, I'm fine with this change. Just wanted to explain why it is this way.

nlordell:
Makes sense.

I just prefer `Result<_, OperationError>` because it is slightly more ergonomic. Also, distinguishing user errors from internal errors seems more like an API thing than something that `Orderbook::add_order` cares about. For example, you could argue for some hypothetical  programmatically added order by some component of the orderbook, that an error - because the signature would be invalid for example - would be an internal error and not a user error. To me this indicates that the responsibility for handling different errors differently should be the caller's and not the callee's.

That being said, there is no real benefit to this (besides `add_order` and `cancel_order` being a tiny bit simpler with `?` instead of `match`-es) and am OK to drop this PR if you/others think it obscures intent of the original code.

vkgnosis (approved):
The original intent doesn't matter compared to what is better now so no need to keep it for its own sake. I don't have a strong preference here but I think your new version is slightly more natural.

MartinquaXD (approved):

PR #1660, opened 2022-02-16
Remove extra create_order error

vkgnosis:
This started printing benign order creation errors with a recent change.
The log itself is not needed as internal errors are already logged and
the extra payload logging is not important.

### Test Plan

CI


nlordell (approved):
Thanks for fixing this 🙈.

PR #1664, opened 2022-02-17
Reject orders selling ETH

nlordell:
This PR changes the behaviour of the order validator to respond with a new error for orders that use `BUY_ETH_ADDRESS` as the sell token.

Previously, we had decided to not have any special handling for this case since, theoretically, `sellToken = BUY_ETH_ADDRESS` has no special meaning at the contract level and it will attempt to interact with it as if it were an ERC20. Practically speaking, it is impossible to have an ERC20 contract at that address (unless you have found a hash collision, but then the whole 🌍 would 🔥). With that in mind, I propose having a dedicated error for this case so that:
- It is clearer from an API level that `BUY_ETH_ADDRESS` isn't supported for sell tokens
- More descriptive error (currently `NoLiquidity` since no ERC20 exists at that address). 

Furthermore, this would allow us to remove the following check from the `SanitizedPriceEstimator`:

https://github.com/gnosis/gp-v2-services/blob/dd160fe01116452151ca7b0db60f7c139b164769/crates/shared/src/price_estimation/sanitized.rs#L105-L107

And modify it to treat `sell_token == BUY_ETH_ADDRESS` as it does with `buy_token == BUY_ETH_ADDRESS`. I plan on opening a follow up PR for this particular change, so I won't go into details on the rationale here.

Even if we decide not to change the `SanitizedPriceEstimator`, I still think this is a valid change for the "clearer API" reason above.

### Test Plan

Added case to cover this new path in the existing unit test. You can also compare:

<details><summary>How the API responds with the code on <code>main</code></summary>

```
% curl -s \
  -X POST \
  --data '@-' \
  'https://api.cow.fi/mainnet/api/v1/quote' \
  <<JSON | jq                                                
{
  "sellToken": "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee",
  "buyToken": "0x6810e776880C02933D47DB1b9fc05908e5386b96",
  "validTo": 2147483648,
  "kind": "buy",
  "partiallyFillable": false,
  "buyTokenBalance": "erc20"
}
JSON

{
  "errorType": "NoLiquidity",
  "description": "not enough liquidity"
}
```

</details>

<details><summary>How the API responds with this change</summary>

```
% curl -s \
  -X POST \
  --data '@-' \
  <<JSON | jq
{
  "sellToken": "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee",
  "buyToken": "0x6810e776880C02933D47DB1b9fc05908e5386b96",
  "validTo": 2147483648,
  "kind": "buy",
  "partiallyFillable": false,
  "buyTokenBalance": "erc20"
}
JSON

{
  "errorType": "InvalidNativeSellToken",
  "description": "The chain's native token (Ether/xDai) cannot be used as the sell token"
}
```

</details>

### DIFF
--- a/crates/orderbook/src/api/order_validation.rs
+++ b/crates/orderbook/src/api/order_validation.rs
@@ -27,6 +27,7 @@ pub trait OrderValidating: Send + Sync {
     ///
     /// Specifically, but *does* verify:
     ///     - if buy token is native asset, receiver is not a smart contract,
+    ///     - the sell token is not the native asset,
     ///     - the sender is not a banned user,
     ///     - the order validity is appropriate,
     ///     - buy_token is not the same as sell_token,
@@ -57,6 +58,7 @@ pub enum PartialValidationError {
     Forbidden,
     InsufficientValidTo,
     TransferEthToContract,
+    InvalidNativeSellToken,
     SameBuyAndSellToken,
     UnsupportedBuyTokenDestination(BuyTokenDestination),
     UnsupportedSellTokenSource(SellTokenSource),
@@ -97,6 +99,13 @@ impl IntoWarpReply for PartialValidationError {
                 super::error(
                     "TransferEthToContract",
                     "Sending Ether to smart contract wallets is currently not supported",
+                ),
+                StatusCode::BAD_REQUEST,
+            ),
+            Self::InvalidNativeSellToken => with_status(
+                super::error(
+                    "InvalidNativeSellToken",
+                    "The chain's native token (Ether/xDai) cannot be used as the sell token",
                 ),
                 StatusCode::BAD_REQUEST,
             ),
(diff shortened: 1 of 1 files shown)


bh2smith (approved):

bh2smith (approved):
I'm a bit surprised this is only happening now. What was the reason for not doing this a long time ago?

nlordell:
> What was the reason for not doing this a long time ago?

It is in the PR description here:
> Previously, we had decided to not have any special handling for this case since, theoretically, `sellToken = BUY_ETH_ADDRESS` has no special meaning at the contract level and it will attempt to interact with it as if it were an ERC20.

vkgnosis (approved):

MartinquaXD (approved):

PR #1665, opened 2022-02-17
Support Wrapping Native Asset In Sanitized Estimator

nlordell:
This PR add support for wrapping native asset (i.e. `sell_token = BUY_ETH_ADDRESS` queries) to the `SanitizedPriceEsimator`.

The main motivation for this PR is to unblock #1649, which currently is failing the native token E2E test. The reason that PR effectively changes the services to use `estimated_prices` computed by `NativePriceEstimating` component - which doesn't support `0xeee..eee` (i.e. `BUY_ETH_ADDRESS`). This is because, under the hood, we use a `SanitizedPriceEsimator` to compute token prices, and it automatically rejects queries where `sell_token == BUY_ETH_ADDRESS`.

There are two straight-forward ways of fixing this issue that I could think of:
1. Lift the restriction on `sell_token == BUY_ETH_ADDRESS` queries in `SanitizedPriceEsimator` (Note that this is not an issue at an API level since we now reject orders and quotes where `sell_token == BUY_ETH_ADDRESS` with a specialized error since #1664 ).
2. Automatically add an entry for `prices[BUY_ETH_ADDRESS] = 1e18` before filtering orders that don't have token prices.

The reason I went for 1. is that, conceptually, it makes sense to be able to estimate conversions from the native asset to another token using our price estimator (even if the contracts don't support such orders). One example of where this can, conceptually, make sense is if we have an interaction that swaps an ERC20 for ETH and **not** WETH (for example Uniswap V1 - although in practice there is absolutely no liquidity there) and then we need some way to convert the ETH to some other token. This way, the price estimator would be able to give us a price estimate for such a conversion (regardless of whether or not CowSwap supports such orders).

That being said, I am OK with going for option 2. as well, WDYT?

Regarding the implementation - it is a little verbose and am not 100% happy with how the code turned out (it seemed nicer before 😛). I couldn't figure out anything cleaver to reduce some of the duplicated code across wrap and unwrap paths.

### Test Plan

Added more cases for new logic in existing unit test.

### DIFF
--- a/crates/orderbook/src/api/get_fee_info.rs
+++ b/crates/orderbook/src/api/get_fee_info.rs
@@ -1,12 +1,17 @@
-use crate::fee::MinFeeCalculating;
-use crate::{api::convert_json_response, fee::FeeData};
+use crate::{
+    api::{convert_json_response, order_validation::PartialValidationError, IntoWarpReply},
+    fee::FeeData,
+    fee::MinFeeCalculating,
+};
 use anyhow::Result;
 use chrono::{DateTime, Utc};
-use model::{order::OrderKind, u256_decimal};
+use model::{
+    order::{OrderKind, BUY_ETH_ADDRESS},
+    u256_decimal,
+};
 use primitive_types::{H160, U256};
 use serde::{Deserialize, Serialize};
-use std::convert::Infallible;
-use std::sync::Arc;
+use std::{convert::Infallible, sync::Arc};
 use warp::{Filter, Rejection};
 
 #[derive(Debug, Serialize, Deserialize)]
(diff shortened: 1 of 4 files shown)


vkgnosis (approved):
Code is fine and your explanation makes. No strong opinion on 1 vs 2.

crates/orderbook/src/api/get_fee_info.rs
```diff
@@ -39,6 +44,12 @@ pub fn get_fee_info(
     get_fee_info_request().and_then(move |query: Query| {
         let fee_calculator = fee_calculator.clone();
         async move {
+            if query.sell_token == BUY_ETH_ADDRESS {
+                return Result::<_, Infallible>::Ok(
+                    PartialValidationError::InvalidNativeSellToken.into_warp_reply(),
+                );
+            }
```
nlordell, line 51: This change is unrelated to this PR, but the legacy (and deprecated) fee endpoint uses a different path for estimating fees than the `quote` and `fee_and_quote` end points.

I snuck it in here, but can move it to a separate PR if you think I should.

MartinquaXD (approved):
LGTM

PR #1669, opened 2022-02-17
Introduce `ExternalPrices` Type

### LINKED ISSUE #1666: Refactor Solver External Prices Into a Rust "New Type"
Small clean up task.

Now that the external prices are, in fact, a mapping of exchange rates of tokens to Ether, we should refactor it to be something like:
```rust
struct ExternalPrices(HashMap<H160, BigRational>);

impl ExternalPrices {
    fn get_native_value(&self, token: H160, amount: U256|BigInt) -> U256|BigInt;
}
```

The reason for this suggestion is so that the invariant that the `BigRational` stored in the `HashMap` is an exchange rate doesn't "leak" out of this type, and so that amount conversions can be done without any knowledge of the representation of the prices. It also helps get rid of any implicit assumptions on the format of the price map being passed around (is it prices of tokens in Ether? Is it price of Ether in tokens? etc.).

If we ever decide to change the price format in the `/auction` endpoint, then it would just involve changes to how the `ExternalPrices` get constructed and its method implementations and not have to look around in the code for where additional changes are required.

nlordell:
Fixes #1666 

This PR introduces a new `ExternalPrices` type for external prices from the `/auction` endpoint. Internally, it tracks this as exchange rates from tokens to Ether (although this detail shouldn't "leak" out of the type). It also internalizes the logic of how to convert a token amount into a native asset amount (so you don't have to remember if you should divide or multiply by the price 😵‍💫).

Additionally it verifies that the prices returned by the `/auction` endpoint satisfy our external price invariant (i.e. Ether prices are 1).

One notable small change in behaviour (which should have no practical bearing) is that the `SellVolumeFilteringSolver` now panics for orders without an external price. Since all orders should have external prices anyway, and we don't actually use this solver anymore, the behaviour of the driver should not have changed.

Overall, I don't think this is a huge improvement (when I opened the issue #1666 it seemed a lot more of a cut and dry improvement, however, as I implemented it, I'm not so sure any more...). Eager to hear your thoughts and feedback on the change!

### Test Plan

Adjusted unit tests to use new type. They continue to pass. It is worth noting, however, that now, external prices need to have a concept of a "native token", this means that unit tests that used external prices were modified to have one of the tokens be the native token.


vkgnosis (approved):
I think it's an improvement but no strong opinion again.

crates/solver/src/settlement/external_prices.rs
```diff
@@ -0,0 +1,193 @@
...
+
+use anyhow::{bail, Result};
+use ethcontract::{H160, U256};
+use lazy_static::lazy_static;
+use model::order::BUY_ETH_ADDRESS;
+use num::{BigInt, BigRational, One as _, ToPrimitive as _};
+use shared::conversions::U256Ext as _;
+use std::collections::{BTreeMap, HashMap};
+
+/// A collection of external prices use for converting token amounts to native
```
vkgnosis, line 16:
```suggestion
/// A collection of external prices used for converting token amounts to native
```

crates/solver/src/settlement/external_prices.rs
```diff
@@ -0,0 +1,193 @@
...
+/// assets.
+#[derive(Clone, Debug)]
+pub struct ExternalPrices(HashMap<H160, BigRational>);
+
+impl ExternalPrices {
+    /// Creates a new set of external prices for the spcified
```
vkgnosis, line 22:
```suggestion
    /// Creates a new set of external prices for the specified exchange rates.
```

crates/solver/src/settlement/external_prices.rs
```diff
@@ -0,0 +1,193 @@
...
+    pub fn new(native_token: H160, mut xrates: HashMap<H160, BigRational>) -> Result<Self> {
+        // Make sure to verify our inviant that native asset price and native
```
vkgnosis, line 24:
```suggestion
        // Make sure to verify our invariant that native asset price and native
```

crates/solver/src/settlement/external_prices.rs
```diff
@@ -0,0 +1,193 @@
...
+                    xrates.insert(token, BigRational::one());
+                }
+            }
+        }
+
+        Ok(Self(xrates))
+    }
+
+    /// Returns a set of external prices for the specified auction model prices.
+    pub fn from_auction_prices(native_token: H160, prices: BTreeMap<H160, U256>) -> Result<Self> {
```
MartinquaXD, line 45: nit: I think `try_from_action_prices` would be more fitting considering this returns a `Result`.

MartinquaXD (approved):
I like the idea of having a component abstract away those error prone/confusing conversions.
Also the documentation looks nice. 👍

PR #1671, opened 2022-02-18
Improve native price cache hit rate

MartinquaXD:
The native price cache performs very well on staging but quite poorly on production.
When I decided on the update strategy I didn't expect the price cache to get saturated (many more traded tokens than can be kept up to date under optimal conditions). Because the set of traded tokens on production is way bigger this assumption doesn't hold and results in too many cache misses.

To find a better solution I set up a prometheus and grafana instance locally and "replayed" a ~10 minutes recording of requests on  production and compared the resulting graphs. Note that the first section of the graph is not identical to the original graph from grafana during that time. I assume this is because of some other variables which I didn't control but I think the improvements compared against the first "vanilla" replay still warrant a PR.

Screenshot 2022-02-18 at 12 40 11

|section|update scheme|max prices to update per cycle|notes|
|-------|----------------|---------------------------------|------|
|1|most outdated first|3||
|2|most recently used first|3||
|3|most often used first|3||
|4|most recently used first|6||
|5|most recently used first|1|aborted due to poor performance|
|6|most recently used first|3|prefetching prices before they are outdated|

Ultimately I settled on updating most recently used (but outdated) prices first as well as prefetching them a little.
Also note since `CachedPrice::requested_at` needs to be updated on every cache hit, using `RwLock` over a `Mutex` doesn't give us anything anymore so I replaced it.

### Test Plan
Manual test
Update existing unit tests


vkgnosis (approved):
Good idea.
I think we should also increase the max update age from 30 to at least 60 which allows twice as many tokens to be in the cache assuming same max updates per second. 60 should still be completely fine with regards to volatility.
I will make a staging PR for this.

nlordell (approved):
Nice. So this is an LRU-ish cache (prioritizes updates of recently used prices) with some logic to keep the cache warm for the very recently used tokens.

Neat idea!

PR #1674, opened 2022-02-19
Stop inspecting Quasimodo metadata field

vkgnosis:
Whether a returned settlement is usable depends on the actual
settlements and not what the metadata fields say. The previous check
caused us to interpret soft errors as hard "other error"s when no
liquidity makes more sense.

### Test Plan

CI, no need for extra unit test


MartinquaXD (approved):

nlordell (approved):
LGTM.

PR #1675, opened 2022-02-22
Handle EDEN transaction fee too high errors

nlordell:
This PR adds logic to [handle errors](https://gnosisinc.slack.com/archives/CUD4MUCUF/p1645464295682109) that we were seeing from EDEN network RPC endpoint when the total gas cost (i.e. `gas_limit  max_fee_per_gas`) would exceed their configured cap of 1.0 Ether:
```
Server error: tx fee (1.14 ether) exceeds the configured cap (1.00 ether)
```

Additionally, this PR refactors the `submitter::common` module to implement the `submit_raw_transaction` method as a `web3::api::Namespace` and use the internal Rust `web3` code for doing HTTP requests and parsing JSON RPC responses. The motivation for this was that it was hard for me to debug where the error was originating because we didn't log the HTTP requests and responses like we do for regular transports. Instead of just adding logs, I thought it was better to do the refactor above and just use our current logging `HttpTransport` implementation.

Another small implementation note - I decided against `impl From<web3::Error> for SubmitApiError` since:
- It is only used in one place, so implementing the trait wasn't really necessary
- It is not generally true - the conversion only really makes sense for `web3.eth().send_raw_transaction()` errors so having a function that we need to manually `.map_err(...)` with made a little more sense to me.

### Test Plan

CI. There were no tests for this code to adapt. I can add some around the error conversion if you think it is worth it.

### DIFF
--- a/crates/solver/src/settlement_submission/submitter.rs
+++ b/crates/solver/src/settlement_submission/submitter.rs
@@ -55,7 +55,7 @@ pub enum SubmitApiError {
     OpenEthereumTooCheapToReplace, // todo ds safe to remove after dropping OE support
     /// EDEN network will reject transactions where the maximum gas cost in ETH
     /// is over 1.0.
-    TransactionTooExpensive,
+    EdenTransactionTooExpensive,
     Other(anyhow::Error),
 }
 
@@ -437,7 +437,13 @@ impl<'a> Submitter<'a> {
             // execute transaction
 
             match self.submit_api.submit_transaction(method.tx).await {
-                Ok(handle) => transactions.push((handle, gas_price)),
+                Ok(handle) => {
+                    tracing::info!(
+                        submitter = %self.submit_api.name(), ?handle,
+                        "submitted transaction",
+                    );
+                    transactions.push((handle, gas_price));
+                }
                 Err(err) => match err {
                     SubmitApiError::InvalidNonce => {
                         tracing::warn!("submission failed: invalid nonce")
@@ -448,8 +454,8 @@ impl<'a> Submitter<'a> {
                     SubmitApiError::OpenEthereumTooCheapToReplace => {
                         tracing::debug!("submission failed: OE has different replacement rules than our algorithm")
                     }
-                    SubmitApiError::TransactionTooExpensive => {
-                        tracing::warn!("submission failed: transaction too expensive")
+                    SubmitApiError::EdenTransactionTooExpensive => {
+                        tracing::warn!("submission failed: eden transaction too expensive")
                     }
                     SubmitApiError::Other(err) => tracing::error!("submission failed: {}", err),
                 },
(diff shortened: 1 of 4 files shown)


vkgnosis (approved):

sunce86:
Is the 1ETH cap documented somewhere? I know about a 1.5M gas cap.

crates/solver/src/settlement_submission/submitter/common.rs
```diff
@@ -1,88 +1,62 @@
...
+            Transaction::Raw { bytes, hash } => (bytes.0, hash),
+        };
+
+        let handle = self
+            .0
+            .eth()
+            .send_raw_transaction(Bytes(raw_signed_transaction))
+            .await
+            .map_err(convert_web3_to_submission_error)?;
+
```
sunce86, line 41: Here I would revert the deleted log:
`tracing::info!(
        "created transaction with hash: {:?} and handle: {:?}, url: {}",
        tx_hash,
        handle,
        url.as_str()
    );
`

sunce86: Not sure if we have a type of private network now... (previously it was recognized by `url`)

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -53,6 +53,9 @@ pub enum SubmitApiError {
     InvalidNonce,
     ReplacementTransactionUnderpriced,
     OpenEthereumTooCheapToReplace, // todo ds safe to remove after dropping OE support
+    /// EDEN network will reject transactions where the maximum gas cost in ETH
+    /// is over 1.0.
+    TransactionTooExpensive,
```
sunce86, line 58: Not super important, but here I was aiming to give a prefix to errors that are tied to a specific submitter (for example `OpenEthereumTooCheapToReplace`) so we could easily recognize common errors vs submitter specific errors. If we wanted to remove Eden in the future, we wouldn't be wondering if any of the other submitters return the same error.

sunce86: So the suggestion would be renaming to `EdenTransactionTooExpensive`

sunce86 (approved):
Lg, just a few comments...

crates/solver/src/settlement_submission/submitter/common.rs
nlordell: Right, didn't mean to remove it.

> Not sure if we have a type of private network now...

I added names to the HTTP transports, so you should see `[eden]` or `[flashbots]` in the JSON RPC logs around the message. We can also log it further up the call chain to have more context for the message.

crates/solver/src/settlement_submission/submitter.rs
nlordell: Renamed.

crates/solver/src/settlement_submission/submitter/common.rs
nlordell: I added the log back in. Good catch!

https://github.com/gnosis/gp-v2-services/blob/65ab1cc090290023318ab7bb9520305dc26005b2/crates/solver/src/settlement_submission/submitter.rs#L441-L444

Note that I moved it higher up in the call-chain so that it can have more context (specifically, it now includes the submitter name instead of the URL).

Issue #1647, opened 2022-02-14
Extra subsidy for vcow holders


vkgnosis:
When an account has more than some threshold Vcow balance we want to give them an extra fee subsidy.

The implementation of this would happen in FeeEstimator which can already create arbitrary fees per order. To get the balance we could re-use the balance fetcher that we already use to check that the user really has the required sell tokens for their order.

I'll self assign for now or @fedgiac can take over.

vkgnosis:
We have an initial implementation in the above PR. This can be adapted once we know exactly how the subsidy should work.

PR #1676, opened 2022-02-22
Support extra subsidy for cow token owners

vkgnosis:
This is the main work for https://github.com/gnosis/gp-v2-services/issues/1647 . We support configuring the service with an erc20 token and extra subsidy for users that have a minimum balance in this token.

The model is a simple binary yes if the threshold is exceed for multiplying an extra factor. We could extend it with other functions in the future for example to scale the subsidy with the amount held.

I have not yet looked at the actual cow token contract so might need some adjustments with that regard in another PR.

In the default configuration no fee subsidy is given.

### Test Plan

new unit test and new ignored test

### DIFF
--- a/crates/model/src/order.rs
+++ b/crates/model/src/order.rs
@@ -45,13 +45,7 @@ impl Default for Order {
             .signature
             .validate(domain, &order.hash_struct())
             .unwrap();
-        Self::from_order_creation(
-            OrderCreation::default(),
-            domain,
-            H160::default(),
-            Default::default(),
-            owner,
-        )
+        Self::from_order_creation(order, domain, H160::default(), Default::default(), owner)
     }
 }
 
--- a/crates/model/src/order.rs
+++ b/crates/model/src/order.rs
@@ -39,13 +39,19 @@ pub struct Order {
 
 impl Default for Order {
     fn default() -> Self {
+        let domain = &DomainSeparator::default();
+        let order = OrderCreation::default();
+        let owner = order
+            .signature
+            .validate(domain, &order.hash_struct())
+            .unwrap();
         Self::from_order_creation(
             OrderCreation::default(),
-            &DomainSeparator::default(),
+            domain,
             H160::default(),
             Default::default(),
+            owner,
         )
-        .unwrap()
     }
 }
 
@@ -65,11 +71,9 @@ impl Order {
         domain: &DomainSeparator,
         settlement_contract: H160,
         full_fee_amount: U256,
-    ) -> Option<Self> {
-        let owner = order_creation
-            .signature
-            .validate(domain, &order_creation.hash_struct())?;
-        Some(Self {
+        owner: H160,
+    ) -> Self {
+        Self {
             order_meta_data: OrderMetaData {
                 creation_date: chrono::offset::Utc::now(),
                 owner,
@@ -79,8 +83,9 @@ impl Order {
                 ..Default::default()
             },
             order_creation,
-        })
+        }
     }
+
     pub fn contains_token_from(&self, token_list: &HashSet<H160>) -> bool {
         token_list.contains(&self.order_creation.buy_token)
             || token_list.contains(&self.order_creation.sell_token)
--- a/crates/orderbook/src/cow_subsidy.rs
+++ b/crates/orderbook/src/cow_subsidy.rs
@@ -11,39 +11,51 @@ const CACHE_LIFESPAN: Duration = Duration::from_secs(60 * 60);
 /// Check whether a user qualifies for an extra fee subsidy because they own enough cow token.
 #[async_trait::async_trait]
 pub trait CowSubsidy: Send + Sync + 'static {
-    async fn qualifies_for_subsidy(&self, user: H160) -> Result<bool>;
+    async fn cow_subsidy_factor(&self, user: H160) -> Result<f64>;
 }
 
-#[derive(Default)]
-pub struct NoopCowSubsidy(pub bool);
+pub struct NoopCowSubsidy(pub f64);
+
+impl Default for NoopCowSubsidy {
+    fn default() -> Self {
+        Self(1.0)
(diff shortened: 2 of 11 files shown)


crates/orderbook/src/fee.rs
```diff
@@ -322,7 +351,21 @@ impl MinFeeCalculating for MinFeeCalculator {
...
+        let cow_owner = async {
+            match user {
+                Some(user) => self.cow_subsidy.qualifies_for_subsidy(user).await,
+                None => Ok(false),
+            }
+        };
+        let past_fee = self
+            .measurements
+            .find_measurement_including_larger_amount(fee_data, (self.now)());
+        // TODO: make this function return a proper error distinguishing errors from fee too low;
```
vkgnosis, line 365: Want to do this in another PR

crates/orderbook/src/fee.rs
```diff
@@ -150,6 +159,9 @@ impl From<u32> for FeeParameters {
 #[cfg_attr(test, mockall::automock)]
 #[async_trait::async_trait]
 pub trait MinFeeCalculating: Send + Sync {
+    // TODO: User should be mandatory. It is not because some old routes like /api/v1/fee do not
+    // know the user. When we remove those apis we can remove the Optional.
+
```
nlordell, line 164: Instead, can't the `api/v1/fee` just use `H160::default()`? Note that `address(0)` is guaranteed to never have a balance:

https://github.com/gnosis/cow-token/blob/6cc9c8ba02116869b319a705f933bcad99633529/src/contracts/vendored/mixins/ERC20.sol#L272-L273

nlordell (commented):
LGTM.

I had some concerns about this making the fee computation time longer, but since we `join` with the price estimates, it will very likely not be the bottleneck.

I also think that `user` should not be optional. We should be able to safely use `H160::default()` in the legacy route because the contract prevents transferring tokens to that address anyway (same for the COW token).

crates/orderbook/src/fee.rs
vkgnosis: I was testing with weth where the 0 address actually has a lot of balance so I thought optional was better for now. Since the cow token does not allow this agree we can use it instead of None.

crates/model/src/order.rs
```diff
@@ -39,13 +39,19 @@ pub struct Order {
...
 impl Default for Order {
     fn default() -> Self {
+        let domain = &DomainSeparator::default();
+        let order = OrderCreation::default();
+        let owner = order
+            .signature
+            .validate(domain, &order.hash_struct())
+            .unwrap();
         Self::from_order_creation(
             OrderCreation::default(),
```
sunce86, line 49: nit: Can we reuse the already created `order` here?

crates/orderbook/src/fee.rs
vkgnosis: Done

sunce86 (approved):
The current implementation LGTM, however I was under impression that the idea for cow token subsidy is not finalized yet (still under discussion in the forum) and could make this impl obsolete.

crates/orderbook/src/cow_subsidy.rs
```diff
@@ -0,0 +1,85 @@
...
+use contracts::ERC20;
+use primitive_types::{H160, U256};
+
+const CACHE_SIZE: usize = 10000;
+const CACHE_LIFESPAN: Duration = Duration::from_secs(60 * 60);
+
+/// Check whether a user qualifies for an extra fee subsidy because they own enough cow token.
+#[async_trait::async_trait]
+pub trait CowSubsidy: Send + Sync + 'static {
+    async fn qualifies_for_subsidy(&self, user: H160) -> Result<bool>;
```
nlordell, line 14: Does it make sense for this component to be a simple `CowBalance` component to keep all subsidy logic in `fee.rs`?

The main reason I was suggesting this is that I'm 90% sure that we will want this to be tiered so we will need the balance amount and not a binary "has balance or not". Additionally, it feels a bit weird to me that the subsidy threshold is configured here and the subsidy amount is configured by the `MinFeeCalculating` implementation (i.e. the code could be a bit more colocated).

nlordell (commented):
In general looks good. Just some comments about how the logic is split.

crates/orderbook/src/cow_subsidy.rs
vkgnosis: Alternatively we can store the threshold and factor both in this component and have `fn cow_subsidy(&self, user: H160) -> Result<f64>`. One difference is that having all the cow subsidy related code is in one place then so that the balance fetching is not separate from the fee calculation. It could be good that this keeps the fee component simpler. This would be reflected in the cache too as at the moment the cache can store `bool` but in your example it would store `U256`.
I'm not sure what is better.

nlordell: Hmm, good point. I guess it does make sense to keep the fee component simpler. Then I would suggest `fn cow_subsidy(&self, user: H160) -> Result<f64>` so that COW related subsidy code is in one place.

crates/orderbook/src/cow_subsidy.rs
```diff
@@ -0,0 +1,110 @@
...
+const CACHE_SIZE: usize = 10000;
+const CACHE_LIFESPAN: Duration = Duration::from_secs(60 * 60);
+
+/// Check whether a user qualifies for an extra fee subsidy because they own enough cow token.
+#[async_trait::async_trait]
+pub trait CowSubsidy: Send + Sync + 'static {
+    async fn cow_subsidy_factor(&self, user: H160) -> Result<f64>;
+}
+
+pub struct NoopCowSubsidy(pub f64);
```
nlordell, line 17: naming-nit: 
```suggestion
pub struct FixedCowSubsidy(pub f64);
```

nlordell (approved):
LGTM

PR #1677, opened 2022-02-23
Add `BufferingPriceEstimator`

MartinquaXD:
In order to enable our frontend to get faster price estimates #1659 got merged. The idea is to send 2 price estimation requests at the same time (one for fast prices, one for optimal prices) and let the frontend display the prices as they come in.
This will lead to many duplicated identical price estimation requests which can result in us getting rate limited by external services more often.

The new `BufferingPriceEstimator` wraps around any `Box<dyn PriceEstimating>` and keeps track of all the requests which are currently in flight. This does not happen on a batch basis but on a `Query` basis. When it gets asked to estimate a batch of queries it only actually calls the wrapped `dyn PriceEstimating` for the queries which are not already in flight.

This is accomplished by creating a `Shared` future for each `Query` which simply awaits the original batch future and returns the single result which corresponds to the `Query`. We store `WeakShared` version of each one and await the `Shared` future instead of the underlying batch request to avoid invalidating the `WeakShared` immediately.
This has also the effect that even if a price estimator decides that it no longer wants to make progress on a price estimation (which happens in the `RacingCompetitionPriceEstimator`) price estimations which depend on that batch can still poll that batch to completion on their own.

To not waste any memory the `BufferingPriceEstimator` collects garbage (i.e. completed futures or one which nobody holds on to anymore) after each batch call.

Note that this PR on its own does nothing. To actually be effective the creation of the price estimators in the orderbook main, needs to be refactored such that the fast and optimal fee calculator use the same instance of each estimator wrapped in a `BufferingPriceEstimator`. This will happen in a follow up PR.

### Test Plan
Added a unit test which for garbage collection and finishing a request which got dropped by the original initiator.

### DIFF
--- a/crates/shared/src/price_estimation/buffered.rs
+++ b/crates/shared/src/price_estimation/buffered.rs
@@ -59,8 +59,11 @@ impl BufferingPriceEstimator {
             }),
         }
     }
+}
 
-    async fn estimate_buffered(&self, queries: &[Query]) -> Vec<EstimationResult> {
+#[async_trait::async_trait]
+impl PriceEstimating for BufferingPriceEstimator {
+    async fn estimates(&self, queries: &[Query]) -> Vec<EstimationResult> {
         let (active_requests, new_requests) = {
             let mut in_flight_requests = self.inner.in_flight_requests.lock().unwrap();
 
@@ -138,13 +141,6 @@ impl BufferingPriceEstimator {
                 None => new_results.next().unwrap(),
             })
             .collect()
-    }
-}
-
-#[async_trait::async_trait]
-impl PriceEstimating for BufferingPriceEstimator {
-    async fn estimates(&self, queries: &[Query]) -> Vec<EstimationResult> {
-        self.estimate_buffered(queries).await
     }
 }
 
--- a/crates/shared/src/price_estimation/buffered.rs
+++ b/crates/shared/src/price_estimation/buffered.rs
@@ -54,61 +54,47 @@ impl BufferingPriceEstimator {
     }
 
     async fn estimate_buffered(&self, queries: &[Query]) -> Vec<EstimationResult> {
-        // For each `Query` either get an in-flight request or keep the `Query` to forward it to the
-        // inner price estimator.
-        let (active_requests, remaining_queries): (Vec<_>, Vec<_>) = {
-            let requests = self.inner.in_flight_requests.lock().unwrap();
-            queries
+        let (active_requests, fetch_remaining_estimates) = {
+            let mut in_flight_requests = self.inner.in_flight_requests.lock().unwrap();
+
+            // For each `Query` either get an in-flight request or keep the `Query` to forward it to the
+            // inner price estimator.
+            let (active_requests, remaining_queries): (Vec<_>, Vec<_>) = queries
                 .iter()
-                .map(|query| match requests.get(query).cloned() {
+                .map(|query| match in_flight_requests.get(query).cloned() {
                     Some(active_request) => (Some(active_request), None),
                     None => (None, Some(*query)),
                 })
-                .unzip()
-        };
+                .unzip();
 
-        // Create future which estimates all `remaining_queries` in a single batch.
-        let fetch_remaining_estimates = {
-            let remaining_queries: Vec<_> = remaining_queries.iter().flatten().cloned().collect();
-            let inner = self.inner.clone();
-            async move { inner.estimator.estimates(&remaining_queries).await }
-                .boxed()
-                .shared()
-        };
+            // Create future which estimates all `remaining_queries` in a single batch.
+            let fetch_remaining_estimates = {
+                let remaining_queries: Vec<_> =
+                    remaining_queries.iter().flatten().cloned().collect();
+                let inner = self.inner.clone();
+                async move { inner.estimator.estimates(&remaining_queries).await }
+                    .boxed()
+                    .shared()
+            };
 
-        // Create a `SharedEstimationRequest` for each individual `Query` of the batch. This
-        // makes it possible for a `batch_2` to await the queries which it is interested in of the
-        // in-flight `batch_1`. Even if the estimator which requested `batch_1` stops polling it, the
-        // estimator of `batch_2` can still poll `batch_1` to completion by polling the
-        // `SharedEstimationRequest` it is actually interested in.
-        //
-        // Build those shared futures up front to keep the critical section short when inserting
-        // them into the `active_requests`.
-        #[allow(clippy::needless_collect)]
-        let individual_requests_for_batch: Vec<_> = remaining_queries
-            .iter()
-            .flatten()
-            .enumerate()
-            .map(|(index, query)| {
-                let fetch_remaining_estimates = fetch_remaining_estimates.clone();
-                (
-                    *query,
-                    async move { fetch_remaining_estimates.await[index].clone() }
-                        .boxed()
-                        .shared(),
-                )
-            })
-            .collect();
-        self.inner
-            .in_flight_requests
-            .lock()
-            .unwrap()
-            // It is possible that someone stored a `SharedEstimationRequest` in the meantime which
-            // could be overwritten now. In those rare cases 2 identical estimation requests would
-            // be in-flight at the same time but it wouldn't produce errors. No memory would leak
-            // and a calling estimator could still decide to stop polling a future while other
-            // estimators interested in the result are able poll the future to completion.
-            .extend(individual_requests_for_batch.into_iter());
+            // Create a `SharedEstimationRequest` for each individual `Query` of the batch. This
+            // makes it possible for a `batch_2` to await the queries which it is interested in of the
+            // in-flight `batch_1`. Even if the estimator which requested `batch_1` stops polling it, the
+            // estimator of `batch_2` can still poll `batch_1` to completion by polling the
+            // `SharedEstimationRequest` it is actually interested in.
+            in_flight_requests.extend(remaining_queries.iter().flatten().enumerate().map(
+                |(index, query)| {
+                    let fetch_remaining_estimates = fetch_remaining_estimates.clone();
+                    (
+                        *query,
+                        async move { fetch_remaining_estimates.await[index].clone() }
+                            .boxed()
+                            .shared(),
+                    )
+                },
+            ));
+            (active_requests, fetch_remaining_estimates)
+        };
 
         // Await all the estimates we need (in-flight and the new ones) in parallel.
         let results = futures::join!(
--- a/crates/shared/src/price_estimation/buffered.rs
+++ b/crates/shared/src/price_estimation/buffered.rs
@@ -39,6 +39,7 @@ impl Inner {
 
 /// A price estimator which doesn't issue another estimation request while an identical one is
 /// already in-flight.
+#[derive(Clone)]
 pub struct BufferingPriceEstimator {
     inner: Arc<Inner>,
 }
--- a/crates/shared/src/price_estimation/buffered.rs
+++ b/crates/shared/src/price_estimation/buffered.rs
@@ -1,0 +1,145 @@
+use crate::price_estimation::{Estimate, PriceEstimating, PriceEstimationError, Query};
+use anyhow::Result;
+use futures::future::Shared;
+use futures::FutureExt;
+use std::collections::HashMap;
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::{Arc, Mutex};
+
+type EstimationResult = Result<Estimate, PriceEstimationError>;
+type SharedEstimationRequest = Shared<Pin<Box<dyn Future<Output = EstimationResult> + Send>>>;
+
+struct Inner {
+    estimator: Box<dyn PriceEstimating>,
+    in_flight_requests: Mutex<HashMap<Query, SharedEstimationRequest>>,
+}
+
+impl Inner {
+    fn collect_garbage(&self) {
+        // TODO: Refactor this to use `HashMap::drain_filter` when it's stable:
+        // https://github.com/rust-lang/rust/issues/59618
+        let mut active_requests = self.in_flight_requests.lock().unwrap();
+        let completed_and_ignored_requests: Vec<_> = active_requests
+            .iter()
+            .filter_map(|(query, handle)| {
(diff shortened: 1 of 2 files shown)


crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,248 @@
...
+            .iter()
+            .flatten()
+            .enumerate()
+            .map(|(index, query)| {
+                let fetch_remaining_estimates = fetch_remaining_estimates.clone();
+                (
+                    *query,
+                    async move { fetch_remaining_estimates.await[index].clone() }
+                        .boxed()
+                        .shared(),
```
nlordell, line 98: I did a double take on the double boxing/sharing of the future, but I think I understand why it is needed: We are creating a single batched query future and so it needs to be shared for each query of the batch (which also needs to be shared for the `inflight_requests_cache`.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,248 @@
...
+        );
+        let (mut in_flight_results, mut new_results) =
+            (results.0.into_iter(), results.1.into_iter());
+
+        // Return the results of new and in-flight requests merged into one.
+        active_requests
+            .iter()
+            .map(|request| match request {
+                Some(_) => in_flight_results.next().unwrap(),
+                None => new_results.next().unwrap(),
```
nlordell, line 132: Cleaver, don't need to keep track of indices this if they are ordered this way.

nlordell (approved):
LGTM and composes nicely with our other price estimation components.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,248 @@
...
+
+impl Inner {
+    fn collect_garbage(&self) {
+        // TODO: Refactor this to use `HashMap::drain_filter` when it's stable:
+        // https://github.com/rust-lang/rust/issues/59618
+        let mut active_requests = self.in_flight_requests.lock().unwrap();
+        let completed_and_ignored_requests: Vec<_> = active_requests
+            .iter()
+            .filter_map(|(query, handle)| {
+                if matches!(handle.strong_count(), Some(1)) {
```
nlordell, line 26: Should this maybe also include:
```suggestion
                if handle.strong_count().unwrap_or_default() <= 1 {
```

I don't think it can happen in practice though... Maybe I'm just paranoid.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,248 @@
...
+        // For each `Query` either get an in-flight request or keep the `Query` to forward it to the
+        // inner price estimator.
+        let (active_requests, remaining_queries): (Vec<_>, Vec<_>) = {
+            let requests = self.inner.in_flight_requests.lock().unwrap();
+            queries
+                .iter()
+                .map(|query| match requests.get(query).cloned() {
+                    Some(active_request) => (Some(active_request), None),
+                    None => (None, Some(*query)),
+                })
```
vkgnosis, line 66: Avoid clone when not needed
```suggestion
                .map(|query| match requests.get(query) {
                    Some(active_request) => (Some(active_request.clone()), None),
                    None => (None, Some(*query)),
                })
```

crates/shared/src/price_estimation/buffered.rs
MartinquaXD: This would be a very slight change in behavior if I'm reading the [docs](https://docs.rs/futures/latest/futures/future/struct.Shared.html#method.strong_count) right.
If future changes cause the `BufferingPriceEstimator` to poll the stored futures to completion, `strong_count()` would return `None` which would cause the future to get dropped regardless of other futures depending on it.
I'm not sure if that would cause problems, but probably not. 🤔

vkgnosis: I think this cannot happen because we never poll this future directly, we just give out copies. If that is true then I would say 

```rust
 if handle.strong_count().unwrap() == 1 {
```

is the most appropriate because it asserts this invariant.

vkgnosis: Based on your comment maybe I misunderstood how this works.

MartinquaXD: I think you understood it right. In my comment I was worried about a possible change in the future.
I agree with you that right now `unwrap()` seems like the best option.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,248 @@
...
+                in_flight_requests: Mutex::new(Default::default()),
+            }),
+        }
+    }
+
+    async fn estimate_buffered(&self, queries: &[Query]) -> Vec<EstimationResult> {
+        // For each `Query` either get an in-flight request or keep the `Query` to forward it to the
+        // inner price estimator.
+        let (active_requests, remaining_queries): (Vec<_>, Vec<_>) = {
+            let requests = self.inner.in_flight_requests.lock().unwrap();
```
vkgnosis, line 60: Should this lock be held until line 111 where in_flight_requests has been extended? During this time we do not perform any async operations. Not holding the lock allows the following to happen:
```
thread 1 reaches this line, locks the mutex, determines it needs to create a new future, unlocks the mutex
thread 2 reaches this line, locks the mutex, determines it needs to create a new future, unlocks the mutex
thread 1 reaches line 102, locks the mutex, inserts its future as in flight, unlocks the mutex
thread 2 reaches line 102, locks the mutex, inserts its future as in flight, unlocks the mutex
```
In this situation we have duplicate in flight requests which could have been prevented.

crates/shared/src/price_estimation/buffered.rs
MartinquaXD: If I'm not mistaken switching to `WeakShared` would not change the behavior of the `BufferingPriceEstimator` while better communicating the intent.

crates/shared/src/price_estimation/buffered.rs
MartinquaXD: I initially held the lock while creating the futures but preferred to have a shorted critical section with the downside of possibly allowing duplicated requests.
I would be fine with either version.

crates/shared/src/price_estimation/buffered.rs
nlordell: I read the docs as `None` means that your particular instance of `Shared` has already been polled (not that the actual shared future has resolved, i.e. `shared.peek().is_some())`).

I tried it out and what I found seems to confirm this:
```rust
use futures::prelude::*;
use std::{
    pin::Pin,
    task::{Context, Poll},
};

fn main() {
    let a = async { 42_i64 }.shared();
    let mut b = a.clone();

    dbg!(a.strong_count()); // Some(2)
    dbg!(b.strong_count()); // Some(2)

    let waker = futures::task::noop_waker();
    let mut context = Context::from_waker(&waker);
    assert_eq!(Pin::new(&mut b).poll(&mut context), Poll::Ready(42));

    dbg!(a.peek()); // Some(42)
    dbg!(a.strong_count()); // Some(1)
    dbg!(b.strong_count()); // None
}
```

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,248 @@
...
+            })
+            .collect()
+    }
+}
+
+#[async_trait::async_trait]
+impl PriceEstimating for BufferingPriceEstimator {
+    async fn estimates(&self, queries: &[Query]) -> Vec<EstimationResult> {
+        let result = self.estimate_buffered(queries).await;
+        self.inner.collect_garbage();
```
vkgnosis, line 142: Cleaning up only after completed estimates a problem. Someone could keep creating new estimates which internally creates in flight futures and then dropping those estimate futures before this call to estimate_buffered completes. This causes collect_garbage to never run.
There are several solutions:

1. Ignore the problem because it is a rare edge case and we complete these futures most of the time
2. Create background clean up task
3. Create a drop guard that performs the cleanup.

3 Would look like
```rust
        struct DropGuard<'a>(&'a BufferingPriceEstimator);
        impl<'a> Drop for DropGuard<'a> {
            fn drop(&mut self) {
                self.0.inner.collect_garbage();
            }
        }
        let _guard = DropGuard(self);
        let result = self.estimate_buffered(queries).await;
        result
```

I'm not sure what is best.

crates/shared/src/price_estimation/buffered.rs
vkgnosis: I agree with this interpretation and under it I think my comment is correct that we can do unwrap. The WeakShared would communicate the intent better because it doesn't implement Future. Using it we would need to call upgrade when finding existing futures and react to None.

crates/shared/src/price_estimation/buffered.rs
vkgnosis: To me it makes sense because we're just creating a couple of vectors in between and this completely guarantees that there won't be duplicate requests in flight at the same time.
It might even be faster overall because there is only one lock instead of two.

crates/shared/src/price_estimation/buffered.rs
vkgnosis: With a drop guard we could even use the still existing `&[Queries]` to only garbage collect the exact queries we might have created.

crates/shared/src/price_estimation/buffered.rs
nlordell: One nicety of using `WeakShared` is that if we get `None` when we try to upgrade, this would mean that the future stopped being polled/resolved for all `Shared` references. In this case it would probably be good to re-query.

crates/shared/src/price_estimation/buffered.rs
MartinquaXD: Alright. Then I will extend the lock accordingly. 👍

nlordell: This also makes sense to me.

crates/shared/src/price_estimation/buffered.rs
MartinquaXD: One downside of `WeakShared` is that it might look a little awkward.
We would have to keep the `Shared` futures around to the end of `buffered_estimate()` only to prevent the ref counted future from being dropped immediately. Besides that we wouldn't do anything with them.
I guess I can make a commit switching to `WeakShared` and we can decide whether to keep it or revert it.

nlordell: Ah I see, because we await `fetch_remaining_estimates` and not the shared futures that get added to the `in_flight_requests.`

crates/shared/src/price_estimation/buffered.rs
nlordell: Instead of a drop guard, shouldn't we just garbage collect before calling `estimate_buffered`? This way we remove any shared futures that have already resolved or are no longer being polled (so we don't use stale requests).

This way we don't have to change the code to use `WeakShared` either.

MartinquaXD: I think the drop guard should be the way to go.
Coming to think of it the current implementation has a problem:
Imagine `batch_1` estimates many queries. This will cause us to store many `Shared` (or `WeakShared`) futures.
Immediately afterwards a totally unrelated batch finishes and collects garbage. This would cause many futures to be dropped which are currently unused but could have gotten used until `batch_a` terminates. This would cause us to issue many more requests than needed.

vkgnosis: That does make me think of a new problem. If the Shared has internally finished but someone keeps a reference to a future using alive because for whatever reason they don't poll it, then we continue to give out more copies of this Shared even though the internal future might be very old.
I don't immediately see how to fix that with this design.

vkgnosis: Another alternative design idea I had but that I didn't think the implications through is to `tokio::spawn` the inner futures. That might simplify some things or give more control in some  cases.

MartinquaXD:
I'm now using a single critical section.
Also I now store `WeakShared` instead of `Shared`. In order to not make the code too awkward because of it, I decided to not await the single batch estimation request but await each individual request (which is just an adapter around the batch request). Otherwise I would have to keep the individual queries alive artificially in order to not make the stored `WeakShared` immediately unusable.
Every `WeakShared` which can not be upgraded or already finished is now considered garbage.

One issue I thought about is the timing of garbage collection. Garbage collection can basically happen at any time if you consider that multiple requests can happen in parallel. With that in mind there is really no "right" time to collect garbage (also we now have the lifetimes of the `WeakShared` futures in order). That means we might as well collect garbage during the big critical section to avoid wasting time acquiring the lock one more time.
What do you think about that idea?

vkgnosis:
> That means we might as well collect garbage during the big critical section to avoid wasting time acquiring the lock one more time.

Yes I thought about this too. If the code stays the way it is now where we go through all futures in collect_garbage then I agree. On the other hand if we do something like the drop guard checking only those queries that were used then there would be a small difference.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,242 @@
...
+
+            // For each `Query` either get an in-flight request or keep the `Query` to forward it to the
+            // inner price estimator.
+            let (active_requests, remaining_queries): (Vec<_>, Vec<_>) = queries
+                .iter()
+                .map(
+                    |query| match in_flight_requests.get(query).map(WeakShared::upgrade) {
+                        Some(Some(active_request)) => (Some(active_request), None),
+                        _ => (None, Some(*query)),
+                    },
```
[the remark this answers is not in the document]
MartinquaXD: Yes, I see the issue now. 🤔 
To be honest I would just ignore it until it actually becomes a problem in the future. Right nothing which would end up using the `BufferingPriceEstimator` follows the pattern of poll-store-abandon.

If we were to tackle the issue I suspect that solution 3 would be cleaner to implement than solution 2 (no handling of special cases). And because the tasks we are dealing with basically just wait until an external service responds, polling every future to completion will not incur much overhead compared to canceling.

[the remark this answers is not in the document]
vkgnosis: I agree that it is ok to ignore it. I would put a description of this in the file somewhere for future reference.

vkgnosis (approved):

PR #1679, opened 2022-02-24
Use BufferingPriceEstimator

MartinquaXD:
Follow up to #1677

In order to let the `optimal` and `fast` price estimators profit from each others work the different estimator hierarchies need to use the same base estimator instances wrapped in a `BufferedPriceEstimator`.
I added `get_or_create_base_estimator()` which stores newly created `Arc<dyn PriceEstimating>` in a map. If someone requests an estimator which already has an instance they will receive a cloned shared pointer to the original.

This change also required that the `RacingCompetitionPriceEstimator` stores `Arc<dyn PriceEstimating>` instead of `Box<dyn PriceEstimating>`.

### Test Plan
Manual test: Requesting a `fast` and `optimal` price quote right after another resulted in a single 1Inch `/quote` request.

### DIFF
--- a/crates/orderbook/src/main.rs
+++ b/crates/orderbook/src/main.rs
@@ -41,6 +41,7 @@ use shared::{
     paraswap_api::DefaultParaswapApi,
     price_estimation::{
         baseline::BaselinePriceEstimator,
+        buffered::BufferingPriceEstimator,
         competition::{CompetitionPriceEstimator, RacingCompetitionPriceEstimator},
         instrumented::InstrumentedPriceEstimator,
         native::NativePriceEstimator,
@@ -500,7 +501,7 @@ async fn main() {
     let instrumented = |inner: Box<dyn PriceEstimating>, name: String| {
         InstrumentedPriceEstimator::new(inner, name, metrics.clone())
     };
-    let create_base_estimator = |&estimator| -> (String, Box<dyn PriceEstimating>) {
+    let create_base_estimator = |estimator| -> (String, Arc<dyn PriceEstimating>) {
         let instance: Box<dyn PriceEstimating> = match estimator {
             PriceEstimatorType::Baseline => Box::new(BaselinePriceEstimator::new(
                 pool_fetcher.clone(),
@@ -551,8 +552,19 @@ async fn main() {
 
         (
             estimator.name(),
-            Box::new(instrumented(instance, estimator.name())),
+            Arc::new(BufferingPriceEstimator::new(Box::new(instrumented(
+                instance,
+                estimator.name(),
+            )))),
         )
+    };
+
+    let mut base_estimators_instances: HashMap<_, _> = Default::default();
+    let mut get_or_create_base_estimator = move |estimator| {
+        base_estimators_instances
+            .entry(estimator)
+            .or_insert_with(|| create_base_estimator(estimator))
+            .clone()
     };
 
     let sanitized = |estimator| {
@@ -566,14 +578,14 @@ async fn main() {
     let price_estimator = Arc::new(sanitized(Box::new(CompetitionPriceEstimator::new(
         args.price_estimators
             .iter()
-            .map(create_base_estimator)
+            .map(|estimator| get_or_create_base_estimator(*estimator))
             .collect(),
     ))));
 
     let fast_price_estimator = Arc::new(sanitized(Box::new(RacingCompetitionPriceEstimator::new(
         args.price_estimators
             .iter()
-            .map(create_base_estimator)
+            .map(|estimator| get_or_create_base_estimator(*estimator))
             .collect(),
         args.fast_price_estimation_results_required,
     ))));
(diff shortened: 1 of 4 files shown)


crates/orderbook/src/main.rs
```diff
@@ -538,10 +542,23 @@ async fn main() {
...
 
+    let base_estimators_instances: Rc<RefCell<HashMap<_, _>>> = Default::default();
+    let get_or_create_base_estimator = |estimator: PriceEstimatorType| {
+        base_estimators_instances
+            .clone()
+            .borrow_mut()
+            .entry(estimator)
+            .or_insert_with(|| create_base_estimator(estimator))
+            .clone()
+    };
```
vkgnosis, line 560: Should be able to do
```rust
    let mut base_estimators_instances: HashMap<_, _> = Default::default();
    let mut get_or_create_base_estimator = move |estimator: PriceEstimatorType| {
        base_estimators_instances
            .entry(estimator)
            .or_insert_with(|| create_base_estimator(estimator))
            .clone()
    };
```
to get rid of Rc RefCell. Although for some reason I also had to replace 

.map(get_or_create_base_estimator)

with

 .map(|a| get_or_create_base_estimator(a))

below

MartinquaXD: That's really cool. I was sure there has to be a way to get it working without `Rc` and `RefCell` but I didn't find it.
I think I didn't try `.map(|a| get_or_create_base_estimator(a))`.
I'm not really sure why this makes such a difference. Will do some research on the topic.

vkgnosis: I was surprised by having to manually use the closure there too. Haven't seen that before. There is a chance it would with 

.map(&mut get_or_create_base_estimator)

but that's still weird.

vkgnosis (approved):

crates/orderbook/src/main.rs
MartinquaXD: I found out what's going on.
The docs hold the [answer](https://doc.rust-lang.org/std/marker/trait.Copy.html#additional-implementors):
> Closure types, if they capture no value from the environment or if all such captured values implement `Copy` themselves. Note that variables captured by shared reference always implement `Copy` (even if the referent doesn’t), while variables captured by mutable reference never implement `Copy`.

The second sentence is important here. `create_base_estimator()` only captures shared references therefore the closure also implements `Copy`.
`get_or_create_base_estimator()` modifies `base_estimator_instances` which therefore has to be captured by an exclusive reference preventing the closure from implementing `Copy`.
If we take a look at 
```rust
    fn map<B, F>(self, f: F) -> Map<Self, F>
    where
        Self: Sized,
        F: FnMut(Self::Item) -> B,
    {
    // ...
```
we see that it takes an `FnMut` by value and consumes it. If we pass `get_base_estimator` into `.map()` it automatically gets copied and `.map()` only consumes the copy of the closure whereas `get_or_create_base_estimator` would actually be consumed which is why we can only pass it to `.map()` once.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -0,0 +1,251 @@
...
+                        async move { fetch_remaining_estimates.await[index].clone() }
+                            .boxed()
+                            .shared(),
+                    )
+                })
+                .collect();
+            in_flight_requests.extend(
+                new_requests
+                    .iter()
+                    .map(|(query, fut)| (*query, fut.downgrade().unwrap())),
```
nlordell, line 122: uber-nano-can-be-ignored-nit: When I read this line, I found myself needing to consult the docs to read up on why the `downgrade` could fail (future has already been polled to completion). Maybe this could be changed to an `expect` since understanding details of how `Shared` work is not common knowledge?

That being said, it took me 20 seconds at most to understand what was going on, so no strong feelings about this suggestion.

nlordell (approved):
Looks good to me!

Sorry for taking a while to review.

crates/shared/src/price_estimation/buffered.rs
MartinquaXD: At some point in time I must have forgotten to rebase this PR properly. The code you highlighted was already merged into main.
But I can sneak this suggestion into this PR, no problem.

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -119,7 +119,7 @@ impl PriceEstimating for BufferingPriceEstimator {
             in_flight_requests.extend(
                 new_requests
                     .iter()
-                    .map(|(query, fut)| (*query, fut.downgrade().unwrap())),
+                    .map(|(query, fut)| (*query, fut.downgrade().expect("future completed"))),
```
MartinquaXD, line 122: Unrelated to this PR. Just something which slipped through the cracks in the previous PR.

vkgnosis: Another way of doing it is keep unwrap with a comment on top of it. Personally I don't find the expect messages useful because in a panic you already get the line number so you know where it happened and can check it in the code. Comments are a lot easier to write and format than the expect string so I like them more.

nlordell: I'm OK with either a comment or `expect`.

PR #1681, opened 2022-02-24
Refactor order validation and bubble up internal error

vkgnosis:
Building on the previous PR this merges into we bubble up the internal error.
In order to do this I moved some code in the validation logic around to change the order of validation to do fast operations early. This caused the existing test to break as it relied on the exact ordering. To fix this I split the test up into several independent tests. This is where most of the added lines are.

### Test Plan

CI, adapted unit tests

### DIFF
--- a/crates/model/src/order.rs
+++ b/crates/model/src/order.rs
@@ -45,7 +45,7 @@ impl Default for Order {
             .signature
             .validate(domain, &order.hash_struct())
             .unwrap();
-        Self::from_order_creation(order, domain, H160::default(), Default::default(), owner)
+        Self::from_order_creation(&order, domain, H160::default(), Default::default(), owner)
     }
 }
 
@@ -61,7 +61,7 @@ pub enum OrderStatus {
 
 impl Order {
     pub fn from_order_creation(
-        order_creation: OrderCreation,
+        order_creation: &OrderCreation,
         domain: &DomainSeparator,
         settlement_contract: H160,
         full_fee_amount: U256,
@@ -76,22 +76,13 @@ impl Order {
                 full_fee_amount,
                 ..Default::default()
             },
-            order_creation,
+            order_creation: *order_creation,
         }
     }
 
     pub fn contains_token_from(&self, token_list: &HashSet<H160>) -> bool {
         token_list.contains(&self.order_creation.buy_token)
             || token_list.contains(&self.order_creation.sell_token)
-    }
-
-    pub fn actual_receiver(&self) -> H160 {
-        let receiver = self.order_creation.receiver.unwrap_or_default();
-        if receiver == H160::zero() {
-            self.order_meta_data.owner
-        } else {
-            receiver
-        }
     }
 }
 
@@ -196,7 +187,7 @@ impl OrderBuilder {
 
 /// An order as provided to the orderbook by the frontend.
 #[serde_as]
-#[derive(Eq, PartialEq, Clone, Deserialize, Debug, Serialize, Hash)]
+#[derive(Eq, PartialEq, Copy, Clone, Deserialize, Debug, Serialize, Hash)]
 #[serde(rename_all = "camelCase")]
 pub struct OrderCreation {
     pub sell_token: H160,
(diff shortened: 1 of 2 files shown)


crates/model/src/order.rs
```diff
@@ -76,23 +76,14 @@ impl Order {
...
         }
     }
 
     pub fn contains_token_from(&self, token_list: &HashSet<H160>) -> bool {
         token_list.contains(&self.order_creation.buy_token)
             || token_list.contains(&self.order_creation.sell_token)
     }
-
-    pub fn actual_receiver(&self) -> H160 {
-        let receiver = self.order_creation.receiver.unwrap_or_default();
```
vkgnosis, line 89: Moved into order_validation.rs as it was the only user.

crates/orderbook/src/api/order_validation.rs
```diff
@@ -373,7 +354,44 @@ impl OrderValidating for OrderValidator {
...
+                    err,
+                )) => ValidationError::Other(err),
+                GetUnsubsidizedMinFeeError::InsufficientFee => ValidationError::InsufficientFee,
+                // Some of the possible errors here have been already checked in this function or
+                // should have been checked when the order was pre-validated. There is no good way
+                // and not much need to bubble them up and we don't want to error log for them so
+                // treat them as insufficient fee.
+                GetUnsubsidizedMinFeeError::PriceEstimationError(_) => {
+                    ValidationError::InsufficientFee
+                }
```
vkgnosis, line 391: This is the part that has changed.

sunce86 (approved):
lg

crates/model/src/order.rs
```diff
@@ -196,7 +187,7 @@ impl OrderBuilder {
 
 /// An order as provided to the orderbook by the frontend.
 #[serde_as]
-#[derive(Eq, PartialEq, Clone, Deserialize, Debug, Serialize, Hash)]
+#[derive(Eq, PartialEq, Copy, Clone, Deserialize, Debug, Serialize, Hash)]
```
MartinquaXD, line 190: Should this really be `Copy`? The struct is pretty big.

crates/model/src/order.rs
```diff
@@ -61,7 +61,7 @@ pub enum OrderStatus {
 
 impl Order {
     pub fn from_order_creation(
-        order_creation: OrderCreation,
+        order_creation: &OrderCreation,
```
MartinquaXD, line 64: What is the reason for this change?
They way it was before makes more sense to me.

crates/model/src/order.rs
vkgnosis: Copy is about whether the type can be copied by just copying bits without extra logic. This is the case for order so it makes sense to be copy. At one point in this PR I noticed I had to call clone a bunch so I made this change.
This is also why at your other comment I turned that into a reference, because there is nothing to be gained by passing it by value. For Copy types moving them is the same as copying them.

MartinquaXD: Alright. I was under the impression that a type should only be `Copy` if it's also small-ish.

nlordell: [TIL](https://doc.rust-lang.org/std/marker/trait.Copy.html#when-should-my-type-be-copy):

> Generally speaking, if your type _can_ implement `Copy`, it should.

MartinquaXD (approved):

crates/model/src/order.rs
vkgnosis: Keep in mind is that for Copy types moving them is the same as copying them. So if you have a big type that is Copy (regardless of whether you actually derive Copy) and you take it by value as a function argument, then it will always be copied completely. So whether your type is Copy and whether it should be passed by value is not related which I think is where the confusion stems from. It just so happens that many types you would pass by value because they are small are also Copy.

Unfortunately (?) Rust uses the same trait Copy to signal several things:
1. Type can be copied by just copying bits
2. Type does not have a custom Drop implementation
3. Type automatically gets copied in several instances like when dereferencing.

To me the most important part is 1 and then 2 and 3 is a slight downside for big Copy types because you might accidentally copy them when not intended. Arguably there should have been several traits for this.

PR #1682, opened 2022-02-24
Add tenderly api simulation

sunce86:
This PR adds possibility to call tenderly simulation api in order to generate access list for the given transaction.

Note that this code is not used anywhere. In next PR I will use it in settlement submission and refactor to be more generic if needed. (will add `eth_createAccessList` call). For now it's in the simplest form possible.

### Test Plan

UT passed.


crates/solver/src/settlement_access_list.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub generate_access_list: bool,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct TenderlyResponse {
+    pub generated_access_list: Vec<AccessListItem>,
+}
+
+/// Had to introduce copy of the web3 AccessList because tenderly responds with snake_case fields
+/// and tenderly storage_keys field does not exist if empty (it should be empty Vec instead)
```
MartinquaXD, line 24: nit: The wording sounds more like an internal comment than a documentation comment.
I think either making this a normal comment `//` or rewording the comment for users of the type would make sense.

crates/solver/src/settlement_access_list.rs
```diff
@@ -0,0 +1,104 @@
...
+                header.insert("x-access-key", HeaderValue::from_str(api_key).unwrap());
+                header
+            },
+        }
+    }
+}
+
+impl TenderlyApi {
+    #[allow(dead_code)]
+    pub async fn access_list(&self, body: TenderlyRequest) -> reqwest::Result<TenderlyResponse> {
```
MartinquaXD, line 66: This can be moved into the existing `impl` block.

crates/solver/src/settlement_access_list.rs
```diff
@@ -0,0 +1,104 @@
...
+
+#[derive(Debug, Serialize)]
+pub struct TenderlyRequest {
+    pub network_id: String,
+    pub block_number: u64,
+    pub from: Address,
+    pub input: String,
+    pub to: Address,
+    pub generate_access_list: bool,
+}
```
MartinquaXD, line 16: To reduce the number of ways you can create a nonsensical `TenderlyRequest` one could change the `network_id` to be a `u64` and `input` to be a `Vec<u8>`.
To make writing tests with `Vec<u8>` somewhat ergonomic one would probably have to use `hex::decode()`.

crates/solver/src/settlement_access_list.rs
```diff
@@ -0,0 +1,104 @@
+use ethcontract::{Address, H256};
+use reqwest::{
+    header::{HeaderMap, HeaderValue},
+    Client, Url,
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize)]
+pub struct TenderlyRequest {
```
MartinquaXD, line 9: nit: I think there is no harm in deriving `Clone` for all the new structs.

crates/solver/src/settlement_access_list.rs
sunce86: True, changed to `//`

crates/solver/src/settlement_access_list.rs
sunce86: Fixed.

crates/solver/src/settlement_access_list.rs
sunce86: Changed `input` to be `Bytes`, since it will be needed for usage in settlement submission as bytes anyway.
I will keep the `network_id` as `String`, since node clients response for network id is a `String` and also tenderly api expects `String`, so I don't see any advantage of having to convert `String -> u64 -> String`, although It's obvious it should be u64.

crates/solver/src/settlement_access_list.rs
sunce86: Sure, added.

crates/solver/src/settlement_access_list.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// Had to introduce copy of the web3 AccessList because tenderly responds with snake_case fields
+// and tenderly storage_keys field does not exist if empty (it should be empty Vec instead)
+#[derive(Debug, Clone, Deserialize)]
+pub struct AccessListItem {
+    /// Accessed address
+    pub address: Address,
+    /// Accessed storage keys
+    pub storage_keys: Option<Vec<H256>>,
```
vkgnosis, line 31:
```suggestion
    #[serde(default)]
    pub storage_keys: Vec<H256>,
```

MartinquaXD (approved):

crates/solver/src/settlement_access_list.rs
```diff
@@ -0,0 +1,103 @@
...
+            .send()
+            .await?
+            .error_for_status()?
+            .json()
+            .await
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
vkgnosis, line 79: Would be nice to have one unignored deserialize test for the response imo. As in copy a real response into a string and deserialize it in the test.

vkgnosis (approved):

crates/solver/src/settlement_access_list.rs
sunce86: Cool

crates/solver/src/settlement_access_list.rs
sunce86: Indeed. Added.

PR #1684, opened 2022-02-25
Sunce86/use access list tenderly

sunce86:
This PR extracts common behavior of estimating access list into a separate trait. This trait is now fully aligned with the future needs of the `eth_createAccessList` call, and also with the usage in the settlement simulation component (for which access list for multiple transactions is needed).

Tenderly API related struct are now internal to this file, so the `pub` is removed.

I am not pushing the rest of the code (usage of this trait in other components) since this code is dependent on the `ethcontract` changes that are not buildable at the moment, so I find this incremental way of pushing the code easier to review.

### Test Plan

Created another real example of usage of `estimate_access_lists` and it passes.


crates/solver/src/settlement_access_list.rs
```diff
@@ -1,35 +1,59 @@
...
     Client, Url,
 };
 use serde::{Deserialize, Serialize};
-use web3::types::Bytes;
+use shared::Web3;
+use web3::types::{AccessList, Bytes};
+
+#[async_trait::async_trait]
+pub trait AccessListEstimating: Send + Sync {
+    async fn estimate_access_list(
```
MartinquaXD, line 13: The naming seems off.
Since there is actually no estimating involved `trait AccessListGenerating` or `trait AccessListing` would fit better?
Function names could then be `generate_access_list()` or `list_accesses()` (this one sounds a little weird, though).

crates/solver/src/settlement_access_list.rs
```diff
@@ -76,16 +100,63 @@ impl TenderlyApi {
...
+                .data
+                .clone()
+                .ok_or(anyhow!("transaction data does not exist"))?;
+            let from = tx
+                .from
+                .clone()
+                .ok_or(anyhow!("transaction from does not exist"))?
+                .address();
+            let to = tx.to.ok_or(anyhow!("transaction to does not exist"))?;
+            let block_number = web3.eth().block_number().await?.as_u64();
```
MartinquaXD, line 122: Would it be problematic if the current block changed during the execution of `estimate_access_lists()`?
Does it make sense to fetch the `block_number` once and reuse it for all the futures?

crates/solver/src/settlement_access_list.rs
sunce86: Hm, I was comparing this to the `eth_estimateGas` functionality which is also used for settlement simulation/submission. Also, with `estimation` in the name I was aiming to say that this trait does not necessarily return 100% precise and always the same result. The result could be different depending on the type of access list estimator and also the exact timing of the call.

crates/solver/src/settlement_access_list.rs
sunce86: Nice catch. Here, I was going for a safe option (delay the `block_number` call to the last minute possible). Theoretically, with each new block access list could change (and also our submission logic last 120s which means a lot of blocks could be mined), but practically, and given our app specific needs, it is possible that this is overkill. Notified this, so I can check later when we build the metrics.

crates/solver/src/settlement_access_list.rs
```diff
@@ -76,16 +100,63 @@ impl TenderlyApi {
     }
 }
 
+#[async_trait::async_trait]
+impl AccessListEstimating for TenderlyApi {
+    async fn estimate_access_lists(
+        &self,
+        web3: &Web3,
+        network_id: String,
```
vkgnosis, line 108: Would it make sense to remove these two parametes from the function and store them into the struct implemeneting the trait instead. To me it makes sense because they should be constant so there is no need to dynamically adjust what is passed.

crates/solver/src/settlement_access_list.rs
nlordell: On suggestion would be `AccessListCreating` so that it is more 1:1 with the RPC method (`eth_createAccessList`).

That being said, I do think its fine as is.

crates/solver/src/settlement_access_list.rs
```diff
@@ -1,35 +1,59 @@
...
+            .into_iter()
+            .next()
+            .unwrap()
+    }
+    async fn estimate_access_lists(
+        &self,
+        web3: &Web3,
+        network_id: String,
+        txs: &[TransactionBuilder<DynTransport>],
+    ) -> Vec<Result<AccessList>>;
```
nlordell, line 30: Do we need to have a batched operation? I can't think of any access list generation method that benefits from that.

I would propose just keeping `estimate_access_list` to keep it simpler.

crates/solver/src/settlement_access_list.rs
nlordell: Also we can use Tenderly's view of the what the latest block is:
```
```

Additionally, at least for the `api/.../simulate` (which you can observe in the developer tools when simulating a transaction in their UI) they use `block_number: null` to indicate simulation on the pending block (which is probably what we want right?) and it includes the actual block that was used in the response. It might also work for the access list API?

image

crates/solver/src/settlement_access_list.rs
sunce86: I thought it would be beneficial to batch several `eth_createAccessList` calls in a single RPC call? This is one place where we could benefit from it: https://github.com/gnosis/gp-v2-services/blob/7dd32c1290b4e9ff1a610e9feadf513ef0b2db70/crates/solver/src/settlement_simulation.rs#L36

sunce86: Do you think it's overkill?

nlordell: Ah good point. We can do a single batched Eth RPC call for the `eth_crateAccessList` implementation.

In that case I don't see why we shouldn't.

crates/solver/src/settlement_access_list.rs
```diff
@@ -1,35 +1,59 @@
...
-use web3::types::Bytes;
+use shared::Web3;
+use web3::types::{AccessList, Bytes};
+
+#[async_trait::async_trait]
+pub trait AccessListEstimating: Send + Sync {
+    async fn estimate_access_list(
+        &self,
+        web3: &Web3,
+        network_id: String,
```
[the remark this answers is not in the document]
sunce86: Ah yeees, missed this.

crates/solver/src/settlement_access_list.rs
sunce86: Tried with null, not working. Added the `block_number` query.

crates/solver/src/settlement_access_list.rs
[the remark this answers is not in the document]
sunce86: Resolved with the new `block_number` query.

crates/solver/src/settlement_access_list.rs
sunce86: Fixed.

vkgnosis (approved):

crates/solver/src/settlement_access_list.rs
```diff
@@ -74,22 +100,81 @@ impl TenderlyApi {
...
+impl AccessListEstimating for TenderlyApi {
+    async fn estimate_access_lists(
+        &self,
+        txs: &[TransactionBuilder<DynTransport>],
+    ) -> Vec<Result<AccessList>> {
+        futures::future::join_all(txs.iter().map(|tx| async {
+            let input = tx
+                .data
+                .clone()
+                .ok_or(anyhow!("transaction data does not exist"))?;
```
nlordell, line 128: ubernit: It would be slightly better to use `ok_or_else` here so that we don't build an error if we don't need it:
```suggestion
                .ok_or_else(|| anyhow!("transaction data does not exist"))?;
```

also, I _think_ `anyhow` has a `Context` implementation for `Option<_>` so that you can turn it into a `Result<_, anyhow::Error>` a little more succinctly:

```rust
use anyhow::Context as _;
```

```suggestion
                .context("transaction data does not exist")?;
```

crates/solver/src/settlement_access_list.rs
```diff
@@ -74,22 +100,81 @@ impl TenderlyApi {
...
+            let from = tx
+                .from
+                .clone()
+                .ok_or(anyhow!("transaction from does not exist"))?
```
nlordell, line 132: Same uber-nit here.

crates/solver/src/settlement_access_list.rs
```diff
@@ -74,22 +100,81 @@ impl TenderlyApi {
...
+                .address();
+            let to = tx.to.ok_or(anyhow!("transaction to does not exist"))?;
```
nlordell, line 134: And here.

crates/solver/src/settlement_access_list.rs
```diff
@@ -74,22 +100,81 @@ impl TenderlyApi {
...
+                block_number: block_number.block_number,
+                from,
+                input,
+                to,
+                generate_access_list: true,
+            };
+
+            match self.access_list(tenderly_request).await {
+                Ok(response) if response.generated_access_list.is_empty() => {
+                    Err(anyhow!("empty access list"))
```
nlordell, line 148:
```suggestion
                    bail!("empty access list")
```

crates/solver/src/settlement_access_list.rs
```diff
@@ -74,22 +100,81 @@ impl TenderlyApi {
...
+                }
+                Ok(response) => Ok(response
+                    .generated_access_list
+                    .into_iter()
+                    .map(Into::into)
+                    .collect()),
+                Err(err) => Err(err.into()),
+            }
```
nlordell, line 156: ubernit: A slightly more succinct way to express this with less "right-ward drift" would be to use `?` and `ensure`:
```suggestion
            let response = self.access_list(tenderly_request).await?;
            ensure!(!response.generated_access_list.is_empty(), "empty access list");
            Ok(response
                .generated_access_list
                .into_iter()
                .map(Into::into)
                .collect())
```

nlordell (approved):
Looks good! Just some small nits, but noything blocking.

crates/solver/src/settlement_access_list.rs
sunce86: Cool, switched to `context`

crates/solver/src/settlement_access_list.rs
sunce86: fixed

crates/solver/src/settlement_access_list.rs
sunce86: also nice :)

crates/solver/src/settlement_access_list.rs
sunce86: fixed

Issue #1663, opened 2022-02-16
Add 0x Limit Order Support


nlordell:
This issue captures work for adding 0x limit order support.

I think this can be done fairly straightforwardly with our current architecture. We would need to define a new liquidity source in the driver that would query the 0x API for limit orders. The following paginated query, for example, returns a page (of 7 at the time of writing) of 0x limit orders that we would be able to use for any of our solvers:
```sh
curl -s 'https://api.0x.org/orderbook/v1/orders?taker=0x0000000000000000000000000000000000000000&sender=0x0000000000000000000000000000000000000000&verifyingContract=0xDef1C0ded9bec7F1a1670819833240f027b25EfF&page=2&perPage=100'
```
 

Some notes:
- `taker` needs to be `address(0)` (or, technically, the address of the EOA that submits the transction filling this order)
- `sender` needs to be `address(0)` or `settlement.address`. Since I don't think anyone is creating orders for GPv2Settlement contract specifically, we can probably just use `address(0)` to start
- `verifyingContract` is set to the latest 0x v4 contract. While we could support older versions (v2 was popular for example) I don't think its worth the effort of the additional logic for encoding interactions (at the time of writing there were 3/651 orders that were not for v4).

The API is speedy enough that we should be able to realistically query it every batch:
```
% cat get_all_orders.sh 
#!/usr/bin/env bash

for page in $(seq 1 7); do
  curl -s "https://api.0x.org/orderbook/v1/orders?taker=0x0000000000000000000000000000000000000000&sender=0x0000000000000000000000000000000000000000&verifyingContract=0xDef1C0ded9bec7F1a1670819833240f027b25EfF&page=${page}&perPage=100" > /dev/null
done
% /usr/bin/time bash get_all_orders.sh 
        5.71 real         0.16 user         0.09 sys
```

From what I can tell there isn't any ordering that we can take advantage of to speed this up.

The solver already support liquidity orders (i.e. limit orders orders that should not receive any surplus) so most of the work will be towards implementing the calldata encoding logic for filling the 0x orders. I think they may be some differences with costs/fees but I can't think of anything substantial.

nlordell:
Since all the orders fit within a single page at the max page size (`1000`) lets just use the HTTP API since it is super easy to use.

[...]

PR #1687, opened 2022-02-28
Add method to get open orders from 0x API

MartinquaXD:
Related to #1663 

In order to allow solvers to utilize 0x limit orders as liquidity our `ZeroExApi` needs to get extended.
Because the 0x endpoint to query those orders is paginated I added 2 functions. `get_orders_with_pagination()` retrieves a specific page of the results whereas `get_orders()` retrieves all orders (modulo some which don't pass a sanity check).
`get_orders()` uses `get_orders_with_pagination()` under the hood.

### Test Plan
New (ignored) unit test which makes sure that deserializing the API responses works.


crates/shared/src/zeroex_api.rs
```diff
@@ -183,24 +369,66 @@ pub enum ZeroExResponseError {
...
+        &self,
+        query: &OrdersQuery,
+    ) -> Result<Vec<OrderRecord>, ZeroExResponseError> {
+        let mut results = Vec::default();
+        let mut page = 1;
+        loop {
+            let mut response = self.get_orders_with_pagination(query, 100, page).await?;
+            let results_in_response = response.records.len();
+            results.append(&mut response.records);
+            if response.per_page != results_in_response as u64 {
```
sunce86, line 404: Wondering if it is possible (and meaningful) to add a UT to confirm this condition? (to make sure that if we send a request for which 0 records are returned that empty Vec is returned instead of some kind of error).

crates/shared/src/zeroex_api.rs
```diff
@@ -75,6 +79,168 @@ impl SwapQuery {
...
+        Self {
+            taker: Some(H160::zero()),
+            sender: Some(H160::zero()),
+            verifying_contract: Some(DefaultZeroExApi::DEFAULT_VERIFICATION_CONTRACT),
+        }
+    }
+}
+
+#[derive(Clone, Derivative, Deserialize, PartialEq)]
+#[derivative(Debug)]
```
vkgnosis, line 135: Is this needed instead of derive Debug? Here and in following structs.

sunce86 (approved):
lg

crates/shared/src/zeroex_api.rs
MartinquaXD: I'm not sure. This is the result of copy-and-paste. 🙈  
I will check if it can be dropped.

crates/shared/src/zeroex_api.rs
MartinquaXD: I think testing `get_orders()` would require a mockable [`reqwest::Client`](https://docs.rs/reqwest_mock/latest/reqwest_mock/). This would require to replace `reqwest::Client` with `reqwest_mock::Client`. Either for this piece of code (but it would be weird to use 2 different `Client` types) or for the entire crate, which should then be done in a separate crate (if desired).

Without replacing the `Client` we could at least have a test to see how `get_orders_with_pagination()` behaves on empty results. This could be done easily by requesting something reasonable like `&page=10000000`.

sunce86: Yes, I was thinking about the latter one.

vkgnosis: Another way of doing it could be to factor out "handle_single_page" or something like that into a standalone function. As input it would the data the client downloaded and it would append to a mutable reference results vector and return whether there are more pages.
I often like this style of testing because there is less machinery to set up than with mocking.

crates/shared/src/zeroex_api.rs
```diff
@@ -75,6 +79,163 @@ impl SwapQuery {
...
+    fn default() -> Self {
+        Self {
+            taker: Some(H160::zero()),
+            sender: Some(H160::zero()),
+            verifying_contract: Some(DefaultZeroExApi::DEFAULT_VERIFICATION_CONTRACT),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Derivative, Deserialize, PartialEq)]
```
vkgnosis, line 134:
```suggestion
#[derive(Debug, Clone, Deserialize, PartialEq)]
```

crates/shared/src/zeroex_api.rs
```diff
@@ -75,6 +79,163 @@ impl SwapQuery {
...
+fn deserialize_epoch_timestamp<'de, D>(deserializer: D) -> Result<DateTime<Utc>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let epoch: &str = Deserialize::deserialize(deserializer)?;
+    let naive = NaiveDateTime::from_timestamp(epoch.parse().map_err(serde::de::Error::custom)?, 0);
+    Ok(DateTime::from_utc(naive, Utc))
+}
+
+#[derive(Debug, Clone, Derivative, Deserialize, PartialEq)]
```
vkgnosis, line 151:
```suggestion
#[derive(Debug, Clone, Deserialize, PartialEq)]
```

crates/shared/src/zeroex_api.rs
```diff
@@ -75,6 +79,163 @@ impl SwapQuery {
...
+
+        match signature.signature_type {
+            2 => Ok(Signature::Eip712(ecdsa)),
+            3 => Ok(Signature::EthSign(ecdsa)),
+            _ => Err(anyhow::anyhow!("unsupported signature type")),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Derivative, Deserialize, PartialEq)]
```
vkgnosis, line 178:
```suggestion
#[derive(Debug, Clone, Deserialize, PartialEq)]
```

crates/shared/src/zeroex_api.rs
```diff
@@ -75,6 +79,163 @@ impl SwapQuery {
...
+    /// The address of the ERC20 token the taker is selling to the maker.
+    pub taker_token: H160,
+    /// Amount of takerToken paid by the taker to the feeRecipient.
+    pub taker_token_fee_amount: U256,
+    /// Address of the contract where the transaction should be sent, usually this is
+    /// the 0x exchange proxy contract.
+    pub verifying_contract: H160,
+}
+
+#[derive(Debug, Clone, Derivative, Deserialize, PartialEq)]
```
vkgnosis, line 222:
```suggestion
#[derive(Debug, Clone, Deserialize, PartialEq)]
```

crates/shared/src/zeroex_api.rs
```diff
@@ -75,6 +79,163 @@ impl SwapQuery {
...
+#[serde(rename_all = "camelCase")]
+pub struct OrderRecord {
+    pub meta_data: OrderMetaData,
+    pub order: Order,
+}
+
+/// A Ox API `orders` response.
+#[derive(Debug, Clone, Default, Derivative, Deserialize, PartialEq)]
```
vkgnosis, line 230:
```suggestion
#[derive(Debug, Clone, Default, Deserialize, PartialEq)]
```

vkgnosis (approved):

MartinquaXD:
I used @vkgnosis idea for testing the pagination loop and used it as well to test retaining the valid orders.
This caused me to actually require 3 instances of `#[derive(Derivative)]` but I removed the other ones.

crates/shared/src/zeroex_api.rs
```diff
@@ -75,6 +81,170 @@ impl SwapQuery {
...
+            v: signature.v,
+        };
+
+        match signature.signature_type {
+            2 => Ok(Signature::Eip712(ecdsa)),
+            3 => Ok(Signature::EthSign(ecdsa)),
+            _ => Err(anyhow::anyhow!("unsupported signature type")),
+        }
+    }
+}
```
nlordell, line 183: Not sure we need this conversion, since 0x signatures are different from ours (it might also exclude orders signed in different ways that we could support). In fact these signatures get encoded in a different way (they will be added to calldata for an interaction and not part of the settlement trades), so I think its fine just keeping `ZeroExSignature`s for these orders and not converting them into `Signature`.

crates/shared/src/zeroex_api.rs
```diff
@@ -118,6 +288,20 @@ pub trait ZeroExApi: Send + Sync {
...
     /// - https://0x.org/docs/api#get-swapv1price
     async fn get_price(&self, query: SwapQuery) -> Result<PriceResponse, ZeroExResponseError>;
+
+    /// Retrieves specific page of current limit orders.
+    async fn get_orders_with_pagination(
+        &self,
+        query: &OrdersQuery,
+        results_per_page: usize,
+        page: usize,
+    ) -> Result<OrdersResponse, ZeroExResponseError>;
```
nlordell, line 298: Don't think this needs to be part of the `trait` if we only ever plan on calling the `get_orders` to get the full book.

Agree it makes 100% sense to have this method as an internal implementation detail though.

crates/shared/src/zeroex_api.rs
```diff
@@ -132,6 +316,12 @@ impl DefaultZeroExApi {
     /// Default 0x API URL.
     pub const DEFAULT_URL: &'static str = "https://api.0x.org/";
 
+    /// Default 0x verifying contract.
+    /// The currently latest 0x v4 contract.
+    pub const DEFAULT_VERIFICATION_CONTRACT: H160 = H160(hex_literal::hex!(
```
nlordell, line 321: ubernit: We should have an `addr` macro:
```suggestion
    pub const DEFAULT_VERIFICATION_CONTRACT: H160 = addr!(
```

crates/shared/src/zeroex_api.rs
```diff
@@ -183,24 +373,80 @@ pub enum ZeroExResponseError {
...
+}
+
+fn retain_valid_orders(orders: &mut Vec<OrderRecord>) {
+    let mut included_orders = HashSet::new();
+    let now = chrono::offset::Utc::now();
+    orders.retain(|order| {
+        // only keep orders which are still valid
+        order.order.expiry > now
+            // properly signed
+            && Signature::try_from(order.order.signature.clone()).is_ok()
```
nlordell, line 436: I don't think this limitation is needed. How 0x orders are signed and verified doesn't need to map to how CowSwap signatures work.

crates/shared/src/zeroex_api.rs
```diff
@@ -183,24 +373,80 @@ pub enum ZeroExResponseError {
...
+            // not a duplicate
+            && included_orders.insert(order.order.salt.clone())
```
nlordell, line 438:
```suggestion
            && included_orders.insert(order.metadata.hash)
```

I don't think the salt is necessarily unique (it looks like a Unix timestamp to me). The order hash is though.

nlordell (changes requested):
Looks good!

Some small nits (although nothing blocking).

Requesting changes for two things:
- From my understanding, the `salt` is not a unique identifier for orders, it just salts the orders (so you can have two distinct orders with identical paramaters). I suggested using the order hash instead from the order metadata in the response.
- We shouldn't require conversions from 0x orders to our signature format. We have inherently different signature formats and we encode them in different places (as part of trades for CowSwap orders and as part of interaction calldata for 0x orders) and the encoding is different anyway (packed signature bytes for CowSwap, ABI encoded signature struct for 0x).

crates/shared/src/zeroex_api.rs
MartinquaXD: Yeah that makes way more sense. I misread and thought the `salt` itself would be unique and not that the `salt` was added to make 2 otherwise identical orders unique.

nlordell (approved):

PR #1688, opened 2022-03-01
Precompute auction like solvable orders

vkgnosis:
instead of creating it on demand.

This PR mostly moves code around from orderbook into solvable_orders. We want to update the solvable orders and the auctions in sync together so it makes sense to me that this would happen in the same component. This did make the file pretty big so we might want to refactor some of the into separate modules while continuing to tie them together in the solvable orders struct.

Precomputing the auction makes sense because the auction is mostly static until the solvable orders change. The direct motivation for this change is that when there are a lot of missing native prices because we just got a large amount of new orders then the auction endpoint still returns instantly. The potentially long running operation of querying these prices has moved into the solvable orders update loop.

As a follow up change we should bound the time that loop takes too. We should do this by restricting the amount of balances and native prices that are fetched in some way if the update would take too long otherwise.

### Test Plan

CI


sunce86 (commented):
Code itself seems ok, however if I understand correctly, auction will be calculated every N seconds now (since it is placed inside solvable orders cache) . What happens if the duration of the auction update > N seconds?

vkgnosis:
The longer an individual update takes the lower the update frequency is. Nothing special happens at that moment where an auction update takes as long as the interval.

MartinquaXD (approved):

crates/orderbook/src/orderbook.rs
```diff
@@ -239,22 +222,13 @@ impl Orderbook {
...
-        Ok(Auction {
-            block: solvable_orders.block,
-            latest_settlement_block: solvable_orders.latest_settlement_block,
-            orders,
-            prices,
-        })
+    pub fn get_auction(&self) -> Result<Auction> {
+        let (auction, update_time) = self.solvable_orders.cached_auction();
+        ensure!(
+            update_time.elapsed() <= self.solvable_orders_max_update_age,
```
nlordell, line 228: Unrelated to this PR, but would it make sense to instead of checking whether or not it is out of date based on time, if it is out of date based on block numbers? I.e. you return the "last updated block" with the auction and compare it to the current block from the `CurrentBlockStream`? There are in rare cases 1min between blocks - and if we are at the end of this window but 30s late, it doesn't really matter because the state of the blockchain hasn't changed.

Also, slightly unrelated to this PR, could we add a historgram metric to get a better sense of how much and how often the auction is out of date? It would help to evaluate the impact of these changes

nlordell (approved):
Looks good.

I would propose changing the check of whether or not the auction is out of date to be more block-centric rather than time-centric and to keep track of just how out of date our auctions tend to be with metrics as some potential follow-up improvements.

> As a follow up change we should bound the time that loop takes too. We should do this by restricting the amount of balances and native prices that are fetched in some way if the update would take too long otherwise.

Agree this is important, otherwise we might just not be able to serve `/auction`s very well because we are constantly updating and it can't stay up to date.

sunce86 (approved):

vkgnosis:
Will merge now. Can continue discussion about remaining point here or in the next PR.

crates/orderbook/src/orderbook.rs
nlordell: Makes sense. In that case measuring orders that get left out makes more sense than auction delay.

Also, sorry if it wasn't clear before - this discussion wasn't blocking the merge for me (hence the approve), just thinking about how we can measure the impact of this change to get a better understanding at what the tradeoffs would be, and be able to pick the right constants for the time limits.

PR #1695, opened 2022-03-02
Eden implementation for eth_sendSlotTx

sunce86:
Specifically for Eden, when sending a transaction to their network, first try to send using method "eth_sendSlotTx" in order to leverage slot priority, if not possible, fallback to "eth_sendRawTransaction".

### DIFF
--- a/crates/solver/src/settlement_submission/submitter/eden_api.rs
+++ b/crates/solver/src/settlement_submission/submitter/eden_api.rs
@@ -7,21 +7,26 @@ use super::{
     common::PrivateNetwork,
     AdditionalTip, CancelHandle, SubmissionLoopStatus,
 };
-use anyhow::{anyhow, Context, Result};
+use anyhow::{Context, Result};
 use ethcontract::{
     transaction::{Transaction, TransactionBuilder},
-    Bytes, H160, U256,
+    Bytes, H160, H256, U256,
 };
 use futures::{FutureExt, TryFutureExt};
 use gas_estimation::EstimatedGasPrice;
-use jsonrpc_core::Output;
 use reqwest::{Client, IntoUrl};
+use serde::Deserialize;
 use shared::{transport::http::HttpTransport, Web3, Web3Transport};
 use web3::Transport;
 
 #[derive(Clone)]
 pub struct EdenApi {
     rpc: Web3,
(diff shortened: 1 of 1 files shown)


crates/solver/src/settlement_submission/submitter/eden_api.rs
```diff
@@ -37,9 +69,16 @@ impl TransactionSubmitting for EdenApi {
         &self,
         tx: TransactionBuilder<Web3Transport>,
     ) -> Result<TransactionHandle, SubmitApiError> {
-        self.rpc
-            .api::<PrivateNetwork>()
-            .submit_raw_transaction(tx)
+        // try to submit with slot method
+        self.submit_slot_transaction(tx.clone())
```
nlordell, line 73: Does this fail if you don't have a valid slot?

nlordell (commented):
Code looks good.

I would add a comment explaining why we do the fallback strategy. Specifically, some context explaining that we first try to submit a slot tx, and that will fail if we don't have a slot claimed delegating to the settlement contract address, at which point we fall back to a "regular" Eden transaction.

crates/solver/src/settlement_submission/submitter/eden_api.rs
sunce86: I _think_ it shouldn't fail because when I send a dummy signed tx it returns the hash. But I wasn't really confident in this new function (these RPCs often tend to disrespect the json standard and similar) so wanted to keep the old gold `eth_sendRawTransaction` in case of any kind of error with `submit_slot_transaction`.
How about I add a comment about this in the code so we can remove the fallback call after some time if not needed, wdyt?

MartinquaXD (approved):

crates/solver/src/settlement_submission/submitter/eden_api.rs
nlordell: Comment would be great!

nlordell (approved):

crates/solver/src/settlement_submission/submitter/eden_api.rs
sunce86: Resolved with comment.

PR #1697, opened 2022-03-02
Rename Order Types and Fields

nlordell:
This PR just does some very light cleanup on something that was bothering me (admittedly, more than it should have 🙈).

- Rename `MetaData` to `Metadata`. AFAIK, "metadata" the correct in English so we should not camelCase it as it is a single word. I _think_ "meta-data" is an alternative form, but I don't really ever see it used and when I checked both Oxford and Webster English dictionaries it was spelt as "metadata".
- Rename `meta_data` to `metadata`, same reasoning as ☝️.
- Rename `Order::order_{creation,meta_data}` to `Order::{creation,metadata}`. The reasoning here is that typing `order.order_THING` seems redundant and `order.THING` is just as informative and more succinct. Note that instances of the kind `let order_creation = OrderCreation { .. }` were not changed, as they make sense as-is to me.

### Test Plan

The compiler. There are no code changes beyond renaming type names and fields.

### DIFF
--- a/crates/e2e/tests/e2e/eth_integration.rs
+++ b/crates/e2e/tests/e2e/eth_integration.rs
@@ -153,7 +153,7 @@ async fn eth_integration(web3: Web3) {
             SecretKeyRef::from(&SecretKey::from_slice(&TRADER_BUY_ETH_A_PK).unwrap()),
         )
         .build()
-        .order_creation;
+        .creation;
     let placement = client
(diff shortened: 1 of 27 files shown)


vkgnosis (approved):

sunce86 (approved):

MartinquaXD (approved):
Glad you got it out of your system. 😅

PR #1700, opened 2022-03-03
Refactor price estimator to return Stream

### LINKED ISSUE #1689: Restrict maximum solvable orders and auction update time in solvable_orders
Follow up from https://github.com/gnosis/gp-v2-services/pull/1688 .

We should do this by restricting the amount of balances and native prices that are fetched in some way if the update would take too long otherwise.

vkgnosis:
This allows estimators to make results available as they come in which gives them and their user more flexibility in how they want to handle multiple requests.

This is motivated by #1689 where we want to try to get as many estimates as possible in some time interval. After a bunch of testing and back and forth I decided the long term cleanest way to implement this would require this change.

To keep the commit reasonably short I introduce the new trait function in `price_estimation.rs` and several helper functions that allow existing code to keep working mostly as before. Then most price estimator implementations and uses are naively replaced with these helper functions.

In follow ups I will
- Change NativePriceEstimator to be Stream based.
- See if there are smarter PriceEstimating implementations possible for the more complicated buffered, competition, instrumented estimators.
- Make use of the streaming in `solvable_orders.rs` to fix #1689 .

### Test Plan

Existing tests in CI

### DIFF
--- a/crates/shared/src/price_estimation.rs
+++ b/crates/shared/src/price_estimation.rs
@@ -15,10 +15,12 @@ pub mod zeroex;
 use crate::{bad_token::BadTokenDetecting, conversions::U256Ext};
 use anyhow::Result;
 use ethcontract::{H160, U256};
+use futures::{stream::BoxStream, StreamExt};
 use model::order::OrderKind;
 use num::BigRational;
 use std::{
     cmp::{Eq, PartialEq},
+    future::Future,
     hash::Hash,
 };
 use thiserror::Error;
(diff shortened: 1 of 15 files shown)


MartinquaXD (approved):
Cool PR!
Great idea to add the helper functions. This made it quite easy to review.
TBH I'm not really sure if #1689 is the way to go, but the change to the API makes sense to me regardless.

crates/shared/src/price_estimation.rs
```diff
@@ -121,19 +123,58 @@ impl Estimate {
...
-            .into_iter()
-            .next()
-            .unwrap()
+pub trait PriceEstimating: Send + Sync + 'static {
+    // The '_ lifetime in the return value is the same as 'a but we need to write it as underscore
+    // because of a mockall limitation.
+
+    /// Returns one result for each query in arbitrary order. The usize is the index into the queries slice.
+    fn estimates<'a>(&'a self, queries: &'a [Query])
+        -> BoxStream<'_, (usize, PriceEstimateResult)>;
```
nlordell, line 135: Because we are eliding the returned lifetime, can you just:
```suggestion
    fn estimates(&self, queries: &[Query])
        -> BoxStream<'_, (usize, PriceEstimateResult)>;
```

crates/shared/src/price_estimation/buffered.rs
```diff
@@ -92,7 +90,7 @@ impl PriceEstimating for BufferingPriceEstimator {
                 let remaining_queries: Vec<_> =
                     remaining_queries.iter().flatten().cloned().collect();
                 let inner = self.inner.clone();
-                async move { inner.estimator.estimates(&remaining_queries).await }
+                async move { vec_estimates(inner.estimator.as_ref(), &remaining_queries).await }
```
nlordell, line 93: Is this what we want here? Doesn't this mean that you need to wait for all queries to resolve and only then will you start returning `Poll::Ready` in the stream? In that case, waiting for the first one or all of them has the same delay right?

nlordell: > See if there are smarter PriceEstimating implementations possible for the more complicated buffered, competition, instrumented estimators.

I see, this is already covered in the PR description.

crates/shared/src/price_estimation/competition.rs
```diff
@@ -95,49 +109,42 @@ impl CompetitionPriceEstimator {
...
-                    winning_estimate.estimate
-                })
-        })
-        .collect()
+fn merge_estimates_from_multiple_estimators<'a>(
+    queries: &'a [Query],
+    all_estimates: Vec<(&'a String, Vec<PriceEstimateResult>)>,
+) -> impl Iterator<Item = PriceEstimateResult> + 'a {
+    queries.iter().enumerate().map(move |(i, query)| {
+        all_estimates
```
nlordell, line 126: `rustfmt` really wasn't our friend here 🤣.

nlordell (approved):
Change makes a lot of sense!

The only concern I have is that it looks including a `BufferedPriceEstimator` or `CompetingPriceEstimator` anywhere in the hierarchy loses your "stream-goodness", but I see that this is already discussed in the PR description and plans on being tackled.

crates/shared/src/price_estimation.rs
vkgnosis: Looking at https://doc.rust-lang.org/reference/lifetime-elision.html it doesn't have the same meaning because the lifetime wouldn't be tied to the argument, only self.

```rust
fn args1<T: ToCStr>(&mut self, args: &[T]) -> &mut Command;                  // elided
fn args2<'a, 'b, T: ToCStr>(&'a mut self, args: &'b [T]) -> &'a mut Command; // expanded
```

PR #1701, opened 2022-03-04
using client instead of transport for sending the slot request

sunce86:
This PR fixes the issue with deserializing the response from Eden. When `eth_sendSlotTx` method is used in request, non-standard json response is received. Since this response is not parse-able by `Transport` (because `Transport` automatically tries to deserialize response to `json_core::Output`), need to use `Client` directly and deserialize response manually.

Note that request IS standard json request, that's why I used `helpers::build_request` to build it, if anyone knows a cleaner way to handle this situation I'm all ears.

https://gnosisinc.slack.com/archives/CQZBU9T5J/p1646370943931949

There is another issue with this method in staging, the error is related to wrong Eden url needed for `eth_sendSlotTx`, but this one will be resolved with configuration change.

### DIFF
--- a/crates/solver/src/settlement_submission/submitter/eden_api.rs
+++ b/crates/solver/src/settlement_submission/submitter/eden_api.rs
@@ -14,13 +14,15 @@ use ethcontract::{
 };
 use futures::{FutureExt, TryFutureExt};
 use gas_estimation::EstimatedGasPrice;
-use reqwest::{Client, IntoUrl};
+use reqwest::{Client, IntoUrl, Url};
 use serde::Deserialize;
 use shared::{transport::http::HttpTransport, Web3, Web3Transport};
-use web3::Transport;
+use web3::helpers;
 
 #[derive(Clone)]
 pub struct EdenApi {
+    client: Client,
+    url: Url,
     rpc: Web3,
 }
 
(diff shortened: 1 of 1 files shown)


vkgnosis (approved):
I think we should add a comment in the file explaining what you also explained in the PR description.

crates/solver/src/settlement_submission/submitter/eden_api.rs
```diff
@@ -51,20 +54,24 @@ impl EdenApi {
         };
         let params =
             serde_json::to_value(Bytes(raw_signed_transaction)).context("failed to serialize")?;
+        let request = helpers::build_request(1, "eth_sendSlotTx", vec![params]);
```
nlordell, line 57: Can we log the request and response?

```rust
tracing::debug!(%request, "sending Eden API request");
// ...
tracing::debug!(%response, "received Eden API response");
```

crates/solver/src/settlement_submission/submitter/eden_api.rs
```diff
@@ -51,20 +54,24 @@ impl EdenApi {
...
+            .json(&request)
+            .send()
             .await
-            .context("transport failed")?;
-        tracing::debug!("response from eden: {:?}", response);
+            .context("failed sending request")?
+            .text()
+            .await
+            .context("failed converting to text")?;
+        tracing::debug!("response from eden: {}", response);
```
nlordell, line 69: ubernit: `tracing` has a cool macro feature for you to add values to the log message that makes it slightly more concise (also I think it allows tighter integration with subscribers, but I'm not 100% sure about that):
```suggestion
        tracing::debug!(%response, "response from eden");
```

nlordell (approved):
LGTM.

sunce86:
> I think we should add a comment in the file explaining what you also explained in the PR description.

Added comment for the function `submit_slot_transaction`

crates/solver/src/settlement_submission/submitter/eden_api.rs
sunce86: Yup

crates/solver/src/settlement_submission/submitter/eden_api.rs
sunce86: Cool, updated.

PR #1703, opened 2022-03-05
eth_createAccessList implementation

sunce86:
This PR implements two things: 

- Calling the `eth_createAccessList` function in order to estimate access list (implemented in struct `NodeApi`).

- Also, implements `PriorityAccessListEstimator` that contains a vector of estimators (the idea is to have `[NodeApi, TenderlyApi]` in production). The priority estimator returns either all results from one of the estimators or nothing. I decided not to mix results from different estimators (for example, if we send 5 txs to priority estimator and first inner estimator has success for only 2, I don't want to fallback to secondary estimator for the rest of the 3 txs). The reasoning for not mixing is:
1. I don't ever expect that primary estimator (`NodeApi`) return partial results.
2. I don't lose much without mixing, since not having access list is not critical.
3. I do lose much if mixing, because I need to send each tx as a separate RPC call - lose batch option.

~~Our Erigon and Nethermind nodes require auth, therefore custom headers are used. If auth is removed in the next days/weeks as a consequence of a OE -> Erigon switch, then I will use web3 transport to refactor and simplify the code a bit.~~ -> Fixed

~~Still not sure if `estimate_access_lists` for `NodeApi` should guarantee the same number of elements in result, compared to a number of transactions for which we are estimating the access lists.~~  -> Decided to have the guarantee.


crates/solver/src/settlement_access_list.rs
```diff
@@ -14,15 +18,95 @@ pub trait AccessListEstimating: Send + Sync {
...
+
+#[derive(Debug)]
+pub struct NodeApi {
+    url: Url,
+    client: Client,
+    header: HeaderMap, //custom header requires direct usage of client instead of transport
+}
+
+impl NodeApi {
+    pub fn new(url: impl IntoUrl, client: Client, api_key: &str) -> Result<Self> {
```
sunce86, line 47: @nlordell the reasoning behind this api_key is to simulate user:pass needed to access the node, but this was only needed for local testing on my machine. After you comment on slack https://gnosisinc.slack.com/archives/C6Z2XNL5Q/p1646642796786589?thread_ts=1645611988.320009&cid=C6Z2XNL5Q it seems like this is not needed and I can use `BatchTransport` here directly ?

crates/solver/src/settlement_access_list.rs
```diff
@@ -14,15 +18,95 @@ pub trait AccessListEstimating: Send + Sync {
...
+                    data: Some(data),
+                    ..Default::default()
+                };
+                let params = helpers::serialize(&request);
+                helpers::build_request(1, "eth_createAccessList", vec![params])
+            })
+            .collect::<Vec<_>>();
+
+        let batch_response = self.send_batch(batch_request).await?;
+        // if we receive less results than txs, we cant figure out how to map them
```
sunce86, line 94: Actually seems we can use json rpc `id` field, fixing it...

crates/solver/src/settlement_access_list.rs
sunce86: Fixed.

crates/solver/src/settlement_access_list.rs
sunce86: Resolved by using the transport.

crates/solver/src/settlement_access_list.rs
```diff
@@ -14,15 +19,73 @@ pub trait AccessListEstimating: Send + Sync {
...
+#[async_trait::async_trait]
+impl AccessListEstimating for NodeApi {
+    async fn estimate_access_lists(
+        &self,
+        txs: &[TransactionBuilder<DynTransport>],
+    ) -> Result<Vec<Result<AccessList>>> {
+        let batch_request = txs
+            .iter()
+            .map(|tx| {
+                let (from, to, data) = resolve_call_request(tx).unwrap_or_default();
```
MartinquaXD, line 60: This seems somewhat problematic depending on how often `resolve_call_request()` returns an error.
First we send a request which will fail (right?) and the final `Vec<Result<_>>` hides the fact that we weren't able to resolve this request in the first place and instead reports this as a web3 or parsing error.

Ideally we would only send the requests which we expect to return sensible results and propagate the error of `resolve_call_request()`. This is how I handled the pattern of `(1) filter initial errors (2) send only sensible requests (3) merge results` in the price estimator code. This is the most elegant I could think of but maybe there is still room for improvement:
```rust
async fn estimate_access_lists(
    &self,
    txs: &[TransactionBuilder<DynTransport>],
) -> Result<Vec<Result<AccessList>>> {
    let batch_request = txs
        .iter()
        .map(|tx| -> Result<_, anyhow::Error> {
            let (from, to, data) = resolve_call_request(tx)?;
            let request = CallRequest {
                from: Some(from),
                to: Some(to),
                data: Some(data),
                ..Default::default()
            };
            let params = helpers::serialize(&request);
            let (id, request) = self
                .web3
                .transport()
                .prepare("eth_createAccessList", vec![params]);
            Ok((id, request))
        })
        .collect::<Vec<_>>();

    // send_batch guarantees the size and order of the responses to match the requests
    let mut batch_response = self
        .web3
        .transport()
        .send_batch(batch_request.iter().flatten().cloned())
        .await?
        .into_iter();

    Ok(batch_request
        .into_iter()
        // merge results of unresolved call requests with responses of resolved requests
        .map(|req| match req {
            // error during `resolve_call_request()`
            Err(e) => Err(e),
            Ok(_req) => match batch_response.next().unwrap() {
                Ok(response) => serde_json::from_value::<NodeAccessList>(response)
                    // error parsing the response
                    .context("unexpected response format")
                    .map(|response| response.access_list),
                // error during transport
                Err(err) => Err(anyhow!("web3 error: {}", err)),
            },
        })
        .collect())
}
```

crates/solver/src/settlement_access_list.rs
```diff
@@ -151,7 +213,63 @@ impl AccessListEstimating for TenderlyApi {
...
+    let to = tx.to.context("transaction to does not exist")?;
+    let data = tx.data.clone().context("transaction data does not exist")?;
+    Ok((from, to, data))
+}
+
+// this function should remove duplicates and elements that are not useful
+// currently only eliminating addresses and storages with value '1'
+// should be updated continually as we learn more about the imperfections of 3rd party access_list calculators
+#[allow(dead_code)]
+fn filter_access_list(access_list: AccessList) -> AccessList {
```
MartinquaXD, line 235: This functionality seems specific to each `AccessListEstimating` implementation using a 3rd party service. This is probably fine for now but in case we find more imperfections unique to each implementation it would make sense to make this a member function of `NodeApi` and `TenderlyApi`.

crates/solver/src/settlement_access_list.rs
```diff
@@ -75,21 +138,26 @@ pub struct TenderlyApi {
...
+        client: Client,
+        api_key: &str,
+        network_id: String,
+    ) -> Result<Self> {
+        Ok(Self {
+            url: url.into_url()?,
+            client,
             header: {
                 let mut header = HeaderMap::new();
                 header.insert("x-access-key", HeaderValue::from_str(api_key).unwrap());
```
MartinquaXD, line 152: Since the function now returns a `Result` I think it would be nice to also have `HeaderValue::from_str(api_key)?`.

crates/solver/src/settlement_access_list.rs
```diff
@@ -151,7 +213,63 @@ impl AccessListEstimating for TenderlyApi {
...
+}
+
+#[async_trait::async_trait]
+impl AccessListEstimating for PriorityAccessListEstimating {
+    async fn estimate_access_lists(
+        &self,
+        txs: &[TransactionBuilder<DynTransport>],
+    ) -> Result<Vec<Result<AccessList>>> {
+        for (i, estimator) in self.estimators.iter().enumerate() {
+            match estimator.estimate_access_lists(txs).await {
```
MartinquaXD, line 265: If latency is a big concern the access lists could be estimated in parallel here with `futures::future::select_all()`. This will only give us anything if the high priority estimator fails. Since this would make the implementation non-trivial it might be worth it to measure first how often this scenario happens.

crates/solver/src/settlement_access_list.rs
```diff
@@ -120,18 +188,12 @@ impl AccessListEstimating for TenderlyApi {
...
-            let input = tx.data.clone().context("transaction data does not exist")?;
-            let from = tx
-                .from
-                .clone()
-                .context("transaction from does not exist")?
-                .address();
-            let to = tx.to.context("transaction to does not exist")?;
+    ) -> Result<Vec<Result<AccessList>>> {
+        Ok(futures::future::join_all(txs.iter().map(|tx| async {
+            let (from, to, input) = resolve_call_request(tx)?;
```
MartinquaXD, line 193: In the context of price estimators @vkgnosis had some [reservations](https://github.com/gnosis/gp-v2-services/commit/d0fe1beda68c694065f0237ab2802f816832f9fc) about using `join_all()` to implement a similar function for 3rd party services which don't offer a batch request endpoint. I don't have a strong opinion here I just thought it's something to consider if you haven't already.

crates/solver/src/settlement_access_list.rs
```diff
@@ -14,15 +19,73 @@ pub trait AccessListEstimating: Send + Sync {
         tx: &TransactionBuilder<DynTransport>,
     ) -> Result<AccessList> {
         self.estimate_access_lists(std::slice::from_ref(tx))
-            .await
+            .await?
             .into_iter()
             .next()
             .unwrap()
     }
+    /// The function guarantee the same size and order of input and output containers
```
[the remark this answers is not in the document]
sunce86: fixed

crates/solver/src/settlement_access_list.rs
sunce86: You are right. :)

crates/solver/src/settlement_access_list.rs
sunce86: fixed, thanks

crates/solver/src/settlement_access_list.rs
sunce86: I expect the high priority estimator to be quite reliable (and much faster than the second one). But I will remember this comment if improvements are needed.

crates/solver/src/settlement_access_list.rs
sunce86: For now I will assume it won't be an issue because we have a pro account. But in general, also not sure about this one.

MartinquaXD (approved):

crates/solver/src/settlement_access_list.rs
nlordell: Can we add a unit test for this filtering function (might help me understand how it was supposed to work)? I'm not sure I understand exactly what its why `address(1)` and `bytes32(1)` slots are special.

Also, if this is specific to a node/tenderly, can we prefix the method name with `erigon_...` or `tenderly_...` for example?

sunce86: We have two tests. I renamed them now to be more clear, `filter_access_list_node` and `filter_access_list_tenderly`.
The `filter_access_list` is used for both apis.

sunce86: And also I am not sure why `1` is returned, would probably need to dig into evm

crates/solver/src/settlement_access_list.rs
vkgnosis: It is probably fine here because this is a more controlled environment. For the price estimator we were calling it different places with different number of queries so I didn't want an unconditional join_all.  The access list code is only used in one or two places.
What you might still want to do is introduce a parallelization limit by converting the input iterator to a stream and then doing https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.for_each_concurrent . Not required for the PR just an idea.

crates/solver/src/settlement_access_list.rs
vkgnosis: I would like to have an explanatory comment for the 1 address and 1 hash too.  It should explain why you included this filter. If the answer is that this some quirk of the nodes then that's fine but should still be a comment.

nlordell (commented):
Code looks good to me,

The only thing that is missing is something explaining why we filter out `address(1)+slot(1)` from the access list because it was not immediately clear to me.

Note that `address(1)` is the ECDSA signature recovery precompile which we use for order signatures, so it makes sense why it is always included in access lists for our settlement transactions. Do access lists affect gas usage of precompiles?

crates/solver/src/settlement_access_list.rs
nlordell: > And also I am not sure why 1 is returned

Note that `address(1)` is the precompile that we use for signature recovery in our settlement contract. Likely related.

sunce86: Added the comment.

nlordell (approved):

crates/solver/src/settlement_access_list.rs
```diff
@@ -14,15 +19,85 @@ pub trait AccessListEstimating: Send + Sync {
...
+            .into_iter();
+
+        Ok(batch_request
+            .into_iter()
+            // merge results of unresolved call requests with responses of resolved requests
+            .map(|req| match req {
+                // error during `resolve_call_request()`
+                Err(e) => Err(e),
+                Ok(_req) => match batch_response.next().unwrap() {
+                    Ok(response) => serde_json::from_value::<NodeAccessList>(response)
```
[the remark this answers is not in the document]
sunce86: I believe this should be already guaranteed here:
https://github.com/gnosis/gp-v2-services/blob/188199640c0509d1c42d0bbbb98327a2c9302477/crates/shared/src/transport/http.rs#L175

In the same function the `id` is stripped, so it's not really possible to match ids at this place in the code.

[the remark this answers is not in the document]
nlordell: Ah good point!

PR #1706, opened 2022-03-07
Fixup zeroex `OrderRecord`

MartinquaXD:
In order to implement a settlement encoding strategy for an upcoming `ZeroExInteraction` we basically have to pass an `OrderRecord` to a smart contract function.
This showed that some fields didn't have the correct type. I also added 2 other things which makes working with the type more ergonomic.

Changes:
- `remaining_fillable_taker_amount`, `maker_amount`, `taker_amount`, `taker_token_fee_amount`: `U256` => `u128`
- `pool`: `Bytes` => `H256`
- `salt`: `String` => `U256`
- `expiry`: `DateTime<Utc>` => `u64` (`retain_valid_order()` now does the conversion to check validity)
- `ZeroExSignature` fields are now public
- added helper function `remaining_maker_amount` which gives the scaled amount for partially filled orders

### Test Plan
added an unit test for deserializing an `OrdersResponse`


crates/shared/src/zeroex_api.rs
```diff
@@ -213,6 +209,21 @@ pub struct OrderRecord {
...
+    /// amount was already used.
+    pub fn remaining_maker_amount(&self) -> u128 {
+        let to_big_int = |num: u128| BigInt::from_u128(num).unwrap();
+        let percent_remaining = BigRational::new(
+            to_big_int(self.metadata.remaining_fillable_taker_amount),
+            to_big_int(self.order.taker_amount),
+        );
+        let scaled_amount = percent_remaining * to_big_int(self.order.maker_amount);
+        // This value can be at most as big as `taker_amount` which fits into an `u128`.
+        scaled_amount.to_integer().to_u128().unwrap()
```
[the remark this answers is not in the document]
MartinquaXD: That's true. I will change it.

sunce86 (approved):

crates/shared/src/zeroex_api.rs
```diff
@@ -196,23 +190,46 @@ pub struct Order {
...
+    #[serde(rename = "metaData")]
     pub metadata: OrderMetadata,
     pub order: Order,
 }
 
+impl OrderRecord {
+    /// Scales the `maker_amount` according to how much of the partially fillable
+    /// amount was already used.
+    pub fn remaining_maker_amount(&self) -> Result<u128> {
+        let to_big_int = |num: u128| BigInt::from_u128(num).unwrap();
```
nlordell, line 216: Might actually make sense to use `U256` checked math here instead of `BigInt` since that is what the SC on-chain does. (Also saves some heap allocations 🎉).

That being said, I don't think anything here can overflow a `U256` so I think its OK as is as well.

nlordell (approved):
LGTM!

One small suggestion, but IMO not blocking approval/merge.

crates/shared/src/zeroex_api.rs
MartinquaXD: Good idea. This also made the good nicer to look at because `BigInt` is somewhat awkward to convert to and from.

PR #1707, opened 2022-03-07
Fix Liquidity Orders Not Getting Considered In Certain Places

nlordell:
This PR fixes a couple small issues where liquidity orders weren't getting considered everywhere they should.

Specifically, we weren't:
- Recording trades for liquidity orders in metrics.
- Considering order UIDs as in-flight (which may cuase `order filled` reverts)
- Naive solver not checking limit prices of liquidity orders are satisfied.

The latter was particularly problematic, because the Naive solver would match direct Cows against liquidity orders at Uniswap V2 spot prices, which could be WAY off for tokens for which there are only shallow pools. This leads to the solvers paying out more than it should to the users under certain conditions (example: https://etherscan.io/tx/0x02e858f10c5b3ab41031421f6634dc0c7799c9aec65160f516af53673dafa92c).

### Test Plan

Added a new unit test to verify case from https://gnosisinc.slack.com/archives/C020UHS2UEL/p1646628310450819 gets fixed.


vkgnosis (approved):

josojo (approved):
Thanks a lot for fixing it so quickly.

I should have noticed it in my liquidity order PR.

I will make a hotfix release with this fix once it was running for 1 day

josojo:
I will merge this test it a little bit and then bring it to production. Hope this is okay for everyone.

The one comment of mine can then also be resolved in a follow up PR, if really needed

PR #1708, opened 2022-03-07
Draft: Add helper to turn async fn into stream

vkgnosis:
I noticed that having this would simplify some of the streaming price
estimation code a lot.
Generators are an unstable nightly only feature that async/await builds
on. By using channels we can emulate most of the generator power with
stable rust.

Could also consider releasing this as a separate crate in case it is useful for others but then I'd feel compelled to polish it more which would take too long at the moment. Maybe something for later.

### Test Plan

CI with new tests

### DIFF
--- a/crates/shared/src/generator.rs
+++ b/crates/shared/src/generator.rs
@@ -1,0 +1,164 @@
+use futures::{channel::mpsc, stream::FusedStream, SinkExt, Stream};
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+/// Creates a stream from an async generator function.
+///
+/// This is similar to [`unfold`](https://docs.rs/futures/0.3.21/futures/stream/fn.unfold.html) in
+/// the sense that it bridges Future to Stream. In addition, it can be more ergonomic because it
+/// allows the compiler to automatically create the state machine that you would otherwise have to
+/// write by hand in the same way that async/await syntax makes it easier to create a Future.
+///
+/// The generator takes an argument `Sender` who's `send` function is used to yield values to the
+/// Stream.
+///
+/// The generator must ensure that the sender does not outlive the generator. This is usually the
+/// case but could be violated by leaking the sender or moving it into a global task. If you want to
+/// use a task then you do not need this function because you can communicate over a channel
+/// directly.
+pub fn async_generator_to_stream<T, Fut>(
+    generator: impl FnOnce(Sender<T>) -> Fut,
+) -> impl Stream<Item = T> + FusedStream
+where
+    Fut: Future<Output = ()>,
+{
+    let (sender, receiver) = mpsc::channel(0);
+    let generator = generator(Sender(sender));
+    StreamAndFuture::new(receiver, Some(generator))
+}
+
+pub struct Sender<T>(mpsc::Sender<T>);
+
+impl<T> Sender<T> {
+    /// Yields a value from the generator into the stream.
+    pub async fn send(&mut self, t: T) {
+        // We assume that the receiver has not been dropped so sending cannot fail.
+        // We use SinkExt::send instead of SinkExt::feed. `send` blocks until the value
+        // has been received (read from the stream). This works because the channel has 0 capacity
+        // and counts as flushed only when it has space for at least 1 message.
+        // `feed` would send the value without waiting for it to be read. This would make the the
+        // generator future calling this function progress too early.
+        self.0.send(t).await.unwrap();
+    }
+}
+
+/// Forward items from the stream while simultaneously running the future.
+///
+/// This can be useful when the future is in some way tied to the stream.
+///
+/// The stream ends when the inner stream ends. This means the future is dropped at this point.
+#[pin_project::pin_project]
+pub struct StreamAndFuture<St, Fut> {
+    // future should come first in the struct so that it gets dropped first which protects against
+    // weird Future implementations that try to use Sender on drop.
+    #[pin]
+    future: Option<Fut>,
+    #[pin]
+    stream: St,
+}
+
+impl<St, Fut> StreamAndFuture<St, Fut> {
+    pub fn new(stream: St, future: Option<Fut>) -> Self {
+        Self { stream, future }
+    }
+}
+
+impl<T, St, Fut> Stream for StreamAndFuture<St, Fut>
+where
+    St: Stream<Item = T>,
+    Fut: Future<Output = ()>,
+{
+    type Item = T;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        if let Some(fut) = self.as_mut().project().future.as_pin_mut() {
+            match fut.poll(cx) {
+                Poll::Ready(()) => self.as_mut().project().future.set(None),
+                Poll::Pending => (),
+            }
+        }
+        self.project().stream.poll_next(cx)
+    }
+}
+
+impl<St, Fut> FusedStream for StreamAndFuture<St, Fut>
+where
+    St: FusedStream,
+    Fut: Future<Output = ()>,
+{
+    fn is_terminated(&self) -> bool {
+        self.stream.is_terminated()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use futures::{FutureExt, StreamExt};
+
+    #[test]
+    fn works() {
+        async fn generator(mut sender: Sender<usize>) {
+            for i in 0..3 {
+                sender.send(i).await;
+            }
+        }
+        let stream = async_generator_to_stream(|sender| async { generator(sender).await });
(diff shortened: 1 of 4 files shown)


crates/orderbook/src/generator.rs
```diff
@@ -0,0 +1,164 @@
...
+    let generator = generator(Sender(sender));
+    StreamAndFuture::new(receiver, Some(generator))
+}
+
+pub struct Sender<T>(mpsc::Sender<T>);
+
+impl<T> Sender<T> {
+    /// Yields a value from the generator into the stream.
+    pub async fn send(&mut self, t: T) {
+        // We assume that the receiver has not been dropped so sending cannot fail.
```
MartinquaXD, line 38: > We **assume** that the receiver has not been dropped...

Don't we know that for sure?
The `StreamAndFuture` owns the receiver and the sender (at least by owning the generator) and doesn't expose access to them.
Also the `Sender` doesn't implement any trait which would allow it to be copied and moved out of the generator to outlive the `StreamAndFuture`, no?

vkgnosis: This is explained a bit in the documentation for `async_generator_to_stream`. One way of this happening is if you spawn a background task and move the sender into that task. Then StreamAndFuture can be dropped which drops the receiver but you still access to the sender.
`Sender` can still be moved even when it isn't Clone or Copy and I didn't find a reasonable way of preventing moves of it in the generator closure.

MartinquaXD: Ahh, yes. Forgot about that.
Could that be prevented by requiring the generator to return its `Sender` on completion?
This is just a thought experiment because the worse ergonomics would probably not be worth it to prevent this edge case.
```rust
pub fn async_generator_to_stream<T, Fut>(
    generator: impl FnOnce(Sender<T>) -> Fut,
) -> impl Stream<Item = T> + FusedStream
where
    Fut: Future<Output = Sender<T>>,
```

vkgnosis: That's one of the ideas I had. It doesn't work because you can do something like
```rust
async fn generator(sender: Sender) -> Sender {
    move_sender_to_task(sender);
    futures::pending::<Sender>().await
}
```

Then StreamAndFuture can get dropped and you're back to having a sender with no receiver. This works because you can write functions like
```rust
fn foo() -> Sender<usize> {
    loop {}
}
```
that "return" any output type even those that have no way to be constructed because the compiler understands that the function never actually returns anything.

vkgnosis: Another idea is using `&mut Sender` but then lifetime become awkward and nothing can be Unpin because the pointed to Sender isn't allowed to move.

MartinquaXD: Oh boy! Very interesting problem.

Another idea: AFAIK the type of every closure is unique. Maybe this fact plus the requirement that the generator returns what we put in could avoid the issue of people "faking" the return type (i.e. `futures::pending::<Sender>().await`).

One could imagine to pass a closure which does the sending instead of a type `Sender`:
```rust
pub fn async_generator_to_stream<T, Fut>(
    generator: impl FnOnce(SENDER) -> Fut,
) -> impl Stream<Item = SENDER> + FusedStream
where
    Fut: Future<Output = SENDER>,
    SENDER: Fn(T) -> ()
```

vkgnosis: While you cannot write out the type you still have access to it. If I make my generator a real function instead of a closure  it would look like
```rust
fn generator<T>(sender: T) where T: Fn(T) -> () {}
```
Inside of it I can still call the type T and call pending. Even if I was forced to write `sender: impl ...` I believe `pending` would still be able to infer the correct type.

MartinquaXD: Of course. 🤦‍♂️ 
I had the order of "usability" of the function types backwards. I thought `FnOnce` would always require a closure (which can't use generic argument or return types AFAIK), but since all other function types are more generic they could be used as well.
In that case I have nothing to add here and might as well approve the PR. 😅

MartinquaXD (approved):
Would have been nice to get rid of the `unwrap` in `send` without compromising ergonomics (error handling for each send) but that fact has been brought up in the documentation comments so I would be fine with merging it as is.
Very cool helper function!

crates/shared/src/generator.rs
```diff
@@ -0,0 +1,164 @@
...
+    // weird Future implementations that try to use Sender on drop.
+    #[pin]
+    future: Option<Fut>,
+    #[pin]
+    stream: St,
+}
+
+impl<St, Fut> StreamAndFuture<St, Fut> {
+    pub fn new(stream: St, future: Option<Fut>) -> Self {
+        Self { stream, future }
```
nlordell, line 65: ubernit: Maybe this doesn't need to be public (assuming that it will only ever be created with `async_generator_to_stream`)?

Additionally, if that is the case, does it make sense to make this type:

```
pub struct StreamAndFuture<T, Fut> {
    // future should come first in the struct so that it gets dropped first which protects against
    // weird Future implementations that try to use Sender on drop.
    #[pin]
    future: Option<Fut>,
    #[pin]
    stream: mpsc::Receiver<T>,
}
```

crates/shared/src/generator.rs
```diff
@@ -0,0 +1,164 @@
...
+/// case but could be violated by leaking the sender or moving it into a global task. If you want to
+/// use a task then you do not need this function because you can communicate over a channel
+/// directly.
+pub fn async_generator_to_stream<T, Fut>(
+    generator: impl FnOnce(Sender<T>) -> Fut,
+) -> impl Stream<Item = T> + FusedStream
+where
+    Fut: Future<Output = ()>,
+{
+    let (sender, receiver) = mpsc::channel(0);
```
nlordell, line 28: If we were to have this buffered, that just means that it is possible for the `Future` to produce more values than actually get consumed by the `Stream` right?

crates/shared/src/generator.rs
```diff
@@ -0,0 +1,164 @@
...
+impl<T> Sender<T> {
+    /// Yields a value from the generator into the stream.
+    pub async fn send(&mut self, t: T) {
+        // We assume that the receiver has not been dropped so sending cannot fail.
+        // We use SinkExt::send instead of SinkExt::feed. `send` blocks until the value
+        // has been received (read from the stream). This works because the channel has 0 capacity
+        // and counts as flushed only when it has space for at least 1 message.
+        // `feed` would send the value without waiting for it to be read. This would make the the
+        // generator future calling this function progress too early.
+        self.0.send(t).await.unwrap();
```
nlordell, line 44: Thanks for the explanation. In general, the way I reasoned about the `unwrap` being OK here is that, internally, we guarantee that the generator `Future` will never get polled after the receiver gets dropped.

I think even if the channel was buffered, or if we used `feed` there would be no issue beyond potentially producing values that might not get consumed by the `Stream` (i.e. "progress too fast"). I'm not sure I understand the issue with that.

crates/shared/src/generator.rs
```diff
@@ -0,0 +1,164 @@
...
+    let (sender, receiver) = mpsc::channel(0);
+    let generator = generator(Sender(sender));
+    StreamAndFuture::new(receiver, Some(generator))
+}
+
+pub struct Sender<T>(mpsc::Sender<T>);
+
+impl<T> Sender<T> {
+    /// Yields a value from the generator into the stream.
+    pub async fn send(&mut self, t: T) {
```
nlordell, line 37: uber-naming-nit: Maybe call it `yield` to be more "generator-y"?
```suggestion
    pub async fn yield(&mut self, t: T) {
```

nlordell (approved):
LGTM.

Also worth checking out is [`async-stream`](https://docs.rs/async-stream) which uses a procedural macro to do something similiar.

crates/orderbook/src/generator.rs
nlordell: You can also expose an additional `try_send` on `Sender` that bubbles up the error and have the generator handle it? This would allow moving the future to a background task and still handle the stream being dropped, while still keeping the `send` for ergonomics with simple generators?

> Another idea is using `&mut Sender` but then lifetime become awkward and nothing can be Unpin because the pointed to Sender isn't allowed to move.

Can you pass in a `Pin<&mut Sender>` and handle pinning in the caller?

vkgnosis:
Thank you for mentioning this crate. I looked at several others which I disliked because they were either nightly only or used too much unsafe. Didn't know about async-stream. It seems to do exactly I want and is probably a higher quality implementation. I will take a closer look at it. I prefer to use that crate rather than my reimplementation, if it holds up.

vkgnosis:
I prefer async_stream. Will put this PR on hold and probably delete in a couple of weeks when I'm sure we really won't use this code.

nlordell:
> Didn't know about async-stream

They include it as a utility crate with `rocket-rs` which I played around with for my `hdnode` project - that's how I know about it.

PR #1710, opened 2022-03-08
Use zeroex liquidity in http solver

### LINKED ISSUE #1663: Add 0x Limit Order Support

MartinquaXD:
Fixes #1663

Added basic machinery to utilize 0x limit orders as liquidity.
For that a new type of liquidity was introduced `Liquidity::LimitOrder(LimitOrder)` which is just a wrapper around our usual `LimitOrder` type.
Also `ZeroExInteraction` was added which just generates call data for `IZeroEx::fillOrKillLimitOrder()` as its encoding strategy.
The http solver takes advantage of the new liquidity limit orders by adding them to the list of solvable orders of the auction.

Note that the Baseline solver does not take advantage of the new liquidity yet.

Whether zeroex limit order should be considered as liquidity can be controlled with the `--baseline-sources` argument. By default the baseline source `ZeroEx` is enabled on mainnet.

### Test Plan
Manual test where I ran the driver locally configured to use quasimodo until a [winning settlement](https://dashboard.tenderly.co/gp-v2/staging/simulator/cac03704-f932-4677-9d28-29c0bddd5c32) was found which included a `ZeroExInteraction`. (call stack includes `fillOrKillLimitOrder()`)

Since multi-0x-hop solutions required another fix I did another manual test to see that way more complex [settlements](https://dashboard.tenderly.co/gp-v2/staging/simulator/df4aa94c-c953-4cff-8c22-38ac78fe85e0) also work.

I didn't add a unit test for generating the call data (as opposed to other interactions) because this is done with a crate which is hopefully properly tested.

### Release notes
Added liquidity of 0x limit orders to quasimodo

### DIFF
--- a/crates/solver/src/interactions/zeroex.rs
+++ b/crates/solver/src/interactions/zeroex.rs
@@ -12,7 +12,7 @@ pub struct ZeroExInteraction {
 
 impl Interaction for ZeroExInteraction {
     fn encode(&self) -> Vec<EncodedInteraction> {
-        let method = self.zeroex.fill_limit_order(
+        let method = self.zeroex.fill_or_kill_limit_order(
             (
                 self.order.maker_token,
                 self.order.taker_token,
--- a/crates/solver/src/interactions/zeroex.rs
+++ b/crates/solver/src/interactions/zeroex.rs
@@ -6,7 +6,7 @@ use shared::zeroex_api::Order;
 #[derive(Clone, Debug)]
 pub struct ZeroExInteraction {
     pub order: Order,
-    pub executed_amount: u128,
+    pub taker_token_fill_amount: u128,
     pub zeroex: IZeroEx,
 }
 
@@ -33,7 +33,7 @@ impl Interaction for ZeroExInteraction {
                 Bytes(self.order.signature.r.0),
                 Bytes(self.order.signature.s.0),
             ),
-            self.executed_amount,
+            self.taker_token_fill_amount,
         );
         let calldata = method.tx.data.expect("no calldata").0;
         vec![(self.zeroex.address(), 0.into(), Bytes(calldata))]
--- a/crates/solver/src/interactions/zeroex.rs
+++ b/crates/solver/src/interactions/zeroex.rs
@@ -12,7 +12,7 @@ pub struct ZeroExInteraction {
 
 impl Interaction for ZeroExInteraction {
     fn encode(&self) -> Vec<EncodedInteraction> {
-        let method = self.zeroex.fill_or_kill_limit_order(
+        let method = self.zeroex.fill_limit_order(
             (
                 self.order.maker_token,
                 self.order.taker_token,
--- a/crates/solver/src/interactions/zeroex.rs
+++ b/crates/solver/src/interactions/zeroex.rs
@@ -1,0 +1,41 @@
+use crate::{encoding::EncodedInteraction, settlement::Interaction};
+use contracts::IZeroEx;
+use ethcontract::Bytes;
+use shared::zeroex_api::Order;
+
+#[derive(Clone, Debug)]
+pub struct ZeroExInteraction {
+    pub order: Order,
+    pub executed_amount: u128,
+    pub zeroex: IZeroEx,
+}
+
+impl Interaction for ZeroExInteraction {
+    fn encode(&self) -> Vec<EncodedInteraction> {
+        let method = self.zeroex.fill_or_kill_limit_order(
+            (
+                self.order.maker_token,
+                self.order.taker_token,
+                self.order.maker_amount,
+                self.order.taker_amount,
+                self.order.taker_token_fee_amount,
+                self.order.maker,
+                self.order.taker,
+                self.order.sender,
+                self.order.fee_recipient,
+                Bytes(self.order.pool.0),
+                self.order.expiry,
+                self.order.salt,
+            ),
+            (
+                self.order.signature.signature_type,
+                self.order.signature.v,
+                Bytes(self.order.signature.r.0),
+                Bytes(self.order.signature.s.0),
+            ),
+            self.executed_amount,
+        );
+        let calldata = method.tx.data.expect("no calldata").0;
+        vec![(self.zeroex.address(), 0.into(), Bytes(calldata))]
+    }
+}
--- a/crates/solver/src/liquidity/zeroex.rs
+++ b/crates/solver/src/liquidity/zeroex.rs
@@ -1,0 +1,73 @@
+use super::SettlementHandling;
+use crate::interactions::ZeroExInteraction;
+use crate::liquidity::{LimitOrder, Liquidity};
+use crate::settlement::SettlementEncoder;
+use anyhow::Result;
+use contracts::IZeroEx;
+use model::order::OrderKind;
+use primitive_types::U256;
+use shared::zeroex_api::{Order, OrderRecord, OrdersQuery, ZeroExApi};
+use std::sync::Arc;
+
+pub struct ZeroExLiquidity {
+    pub api: Arc<dyn ZeroExApi>,
+    pub zeroex: IZeroEx,
+}
+
+impl ZeroExLiquidity {
+    pub async fn get_liquidity(&self) -> Result<Vec<Liquidity>> {
+        Ok(self
+            .api
+            .get_orders(&OrdersQuery::default())
+            .await?
+            .into_iter()
+            .filter_map(|record| self.record_into_liquidity(record))
+            .collect())
+    }
+
+    /// Turns 0x OrderRecord into liquidity which solvers can use.
+    fn record_into_liquidity(&self, record: OrderRecord) -> Option<Liquidity> {
+        let sell_amount: U256 = record.remaining_maker_amount().ok()?.into();
+        if sell_amount.is_zero() || record.metadata.remaining_fillable_taker_amount == 0 {
+            // filter out orders with 0 amounts to prevent errors in the solver
+            return None;
+        }
+
+        let limit_order = LimitOrder {
+            id: hex::encode(&record.metadata.order_hash.0),
+            sell_token: record.order.maker_token,
+            buy_token: record.order.taker_token,
+            sell_amount,
+            buy_amount: record.metadata.remaining_fillable_taker_amount.into(),
+            kind: OrderKind::Buy,
+            partially_fillable: true,
+            unscaled_subsidized_fee: U256::zero(),
+            scaled_unsubsidized_fee: U256::zero(),
+            is_liquidity_order: true,
+            settlement_handling: Arc::new(OrderSettlementHandler {
+                order: record.order,
+                zeroex: self.zeroex.clone(),
+            }),
+        };
+        Some(Liquidity::LimitOrder(limit_order))
+    }
+}
+
+struct OrderSettlementHandler {
+    order: Order,
+    zeroex: IZeroEx,
+}
+
+impl SettlementHandling<LimitOrder> for OrderSettlementHandler {
+    fn encode(&self, executed_amount: U256, encoder: &mut SettlementEncoder) -> Result<()> {
+        if executed_amount > u128::MAX.into() {
+            anyhow::bail!("0x only supports executed amounts of size u128");
+        }
+        encoder.append_to_execution_plan(ZeroExInteraction {
+            executed_amount: executed_amount.as_u128(),
+            order: self.order.clone(),
+            zeroex: self.zeroex.clone(),
+        });
+        Ok(())
+    }
+}
--- a/crates/solver/src/liquidity/zeroex.rs
+++ b/crates/solver/src/liquidity/zeroex.rs
@@ -64,7 +64,7 @@ impl SettlementHandling<LimitOrder> for OrderSettlementHandler {
             anyhow::bail!("0x only supports executed amounts of size u128");
         }
         encoder.append_to_execution_plan(ZeroExInteraction {
-            executed_amount: executed_amount.as_u128(),
+            taker_token_fill_amount: executed_amount.as_u128(),
             order: self.order.clone(),
             zeroex: self.zeroex.clone(),
         });
--- a/crates/solver/src/liquidity/zeroex.rs
+++ b/crates/solver/src/liquidity/zeroex.rs
@@ -48,6 +48,7 @@ impl ZeroExLiquidity {
                 order: record.order,
                 zeroex: self.zeroex.clone(),
             }),
+            has_atomic_execution: true,
         };
         Some(Liquidity::LimitOrder(limit_order))
     }
--- a/crates/solver/src/liquidity/zeroex.rs
+++ b/crates/solver/src/liquidity/zeroex.rs
@@ -1,6 +1,6 @@
 use super::SettlementHandling;
 use crate::interactions::ZeroExInteraction;
-use crate::liquidity::{LimitOrder, Liquidity};
+use crate::liquidity::{Exchange, LimitOrder, Liquidity};
 use crate::settlement::SettlementEncoder;
 use anyhow::Result;
 use contracts::IZeroEx;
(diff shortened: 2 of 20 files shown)


crates/solver/src/solver/http_solver.rs
```diff
@@ -367,6 +370,20 @@ impl Solver for HttpSolver {
...
+        orders.extend(
+            liquidity
+                .iter()
+                .filter_map(|liquidity| match liquidity {
+                    Liquidity::LimitOrder(order) => Some(order.clone()),
+                    _ => None,
+                })
+                .filter(|order| {
+                    prices.get(&order.sell_token).is_some()
+                        && prices.get(&order.buy_token).is_some()
```
MartinquaXD, line 383: I'm not entirely sure this is needed because of:
```rust
        // We are guaranteed to have price estimates for all tokens that are relevant to the
        // objective value by the driver. It is possible that we have AMM pools that contain tokens
        // that are not any order's tokens. We used to fetch these extra prices but it would often
        // slow down the solver and the solver can estimate them on its own.
        let price_estimates = external_prices.into_http_solver_prices();
```
But when I passed all orders I always got [settlement simulation failures](https://dashboard.tenderly.co/gp-v2/staging/simulator/72e65804-2bf8-4b5b-ad52-dc6b637a5b6f). I think the error is always caused by some transfer exceeding the balance of our settlement contract. I'm not sure how filtering out the orders without prices affects that but it seems to be related.
Any ideas on that would be greatly appreciated.

MartinquaXD, line 383: I'm not entirely sure this is needed because of:
```rust
        // We are guaranteed to have price estimates for all tokens that are relevant to the
        // objective value by the driver. It is possible that we have AMM pools that contain tokens
        // that are not any order's tokens. We used to fetch these extra prices but it would often
        // slow down the solver and the solver can estimate them on its own.
        let price_estimates = external_prices.into_http_solver_prices();
```
But when I passed all orders I always got [settlement simulation failures](https://dashboard.tenderly.co/gp-v2/staging/simulator/72e65804-2bf8-4b5b-ad52-dc6b637a5b6f). I think the error is always caused by some transfer exceeding the balance of our settlement contract. I'm not sure how filtering out the orders without prices affects that but it seems to be related.
Any ideas on that would be greatly appreciated.

crates/solver/src/interactions/zeroex.rs
```diff
@@ -0,0 +1,41 @@
...
+#[derive(Clone, Debug)]
+pub struct ZeroExInteraction {
+    pub order: Order,
+    pub executed_amount: u128,
+    pub zeroex: IZeroEx,
+}
+
+impl Interaction for ZeroExInteraction {
+    fn encode(&self) -> Vec<EncodedInteraction> {
+        let method = self.zeroex.fill_or_kill_limit_order(
```
nlordell, line 15: Why do you use `fill_or_kill` here if the orders are partially fillable?

**edit**: The 0x function name threw me off: https://github.com/0xProject/protocol/blob/development/docs/basics/functions.rst#fillorkilllimitorder

the "fill-or-kill` means that it fills the entire _specified executed amount_ and **not** the _entire remaining amount_ of order. The alternate `fillLimitOrder` method allows the order to be partially filled **up to** the specified amount.

MartinquaXD: I must have read the docs not carefully enough. 🤦‍♂️ 
Will change it.

crates/solver/src/interactions/zeroex.rs
```diff
@@ -0,0 +1,41 @@
+use crate::{encoding::EncodedInteraction, settlement::Interaction};
+use contracts::IZeroEx;
+use ethcontract::Bytes;
+use shared::zeroex_api::Order;
+
+#[derive(Clone, Debug)]
+pub struct ZeroExInteraction {
+    pub order: Order,
+    pub executed_amount: u128,
```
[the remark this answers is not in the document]
MartinquaXD: Sure, no problem.
I will go for the second suggestion.

crates/solver/src/solver/http_solver.rs
nlordell: Hmm, looking at the settlement, something looks really suspicious. For a single trade, the solver decided to use 7 0x orders and a Balancer interaction.
1. `0x2b59...` -> `0xa0b8...`
2. `0xa0b8...` -> `0x2b59...`
3. `0x3819...` -> `0xa0b8...`
4. `0x2b59...` -> `0xa0b8...`
5. `0x2b59...` -> `0x3819...`
6. `0x2b59...` -> `0xa0b8...`
7. `0xa0b8...` -> `0x3819...`

And finally, the Balancer V2 swap was:
7. `0xa0b8...` -> `0xdac1...`


Which... looks wrong for matching an order trading `0xa0b8...` to `0xdac1...`

> I'm not sure how filtering out the orders without prices affects that but it seems to be related.

I'm guessing that fewer 0x orders were considered to the problem didn't manifest itself.

nlordell: Also worth noting is that **execution order matters** for 0x orders (and not for GP orders). cc @marcovc

crates/solver/src/interactions/zeroex.rs
nlordell: Nono! It is correct as is!

MartinquaXD: Arrgh. Just pushed the change. 😅 
Will revert it again.

crates/solver/src/solver/http_solver.rs
MartinquaXD: It's not obvious to me why execution order would matter for 0x orders. If their price is constant it shouldn't matter in which order they get executed because they can't influence each others prices, no?

nlordell: Its because they are executed immediately, so the settlement contract needs sufficient balance at execution time, and not at the end of the settlement.

marcovc: Thanks for letting me know about this. I think we then need to add some flag to these orders to differentiate them from other liquidity orders where order of execution does not matter, and then do something different with them in the solver if this flag is set.

vkgnosis:
> I make the http solver take advantage of the new liquidity by

This sentence cuts off.

crates/solver/src/solver/http_solver.rs
nlordell: This is actually quite interesting. I ran the numbers and computed how much of each token was leaving the settlement contract and how much was being added to see why all these 0x orders were being traded, and it looks like the optimization solver found some arbitrage opportunity with the orders:
```
┌───────┬──────────────────────────────────────────────┬─────────────┐
│ (idx) │ tokenAddress                                 │ value       │
├───────┼──────────────────────────────────────────────┼─────────────┤
│     0 │ "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48" │ "32130726"  │
│     1 │ "0x2b591e99afE9f32eAA6214f7B7629768c40Eeb39" │ "-51"       │
│     2 │ "0x3819f64f282bf135d62168C1e513280dAF905e06" │ "110326224" │
└───────┴──────────────────────────────────────────────┴─────────────┘
```

Specifically, around `32 USDC` - which isn't a lot. I'd be interested in understanding why it chose to do this (as I imagine the gas we spend is WAY more than 32$). I imagine this could indicate that there is an issue with our fee/cost model to the orders that are being passed in.

@MartinquaXD, maybe you can collect the instance JSON for this kind of settlement to give to @marcovc in order to investigate?

All in all though - it looks like the solver is doing what it is supposed to (modulo including a bunch of orders for marginal amount of arbitrage, although that may be an issue with how the limit orders are being passed in).

MartinquaXD:
This PR showed that some work has to be done on the solver side in order to not sometimes cause settlement simulation failures. The progress on the solver side is captured [here](https://github.com/gnosis/gp-v2-solver-lib/issues/89).

nlordell:
> This PR showed that some work has to be done on the solver side in order to not sometimes cause settlement simulation failures.

Maybe we can highjack this PR to already add the `has_atomic_execution` flag to limit orders that need to get passed to the solver? Can also be separate - not sure how much extra code it would add.

crates/solver/src/solver/http_solver.rs
```diff
@@ -367,6 +370,20 @@ impl Solver for HttpSolver {
         if orders.is_empty() {
             return Ok(Vec::new());
         };
+        let prices = external_prices.clone().into_http_solver_prices();
+        orders.extend(
```
nlordell, line 374: I think we should remove this filtering because I also don't think its needed.

nlordell (commented):
In general the code looks good.

Another thing to consider is having a custom "cost-model" that more closely captures the actual gas usage of a 0x order.

MartinquaXD:
>  Maybe we can highjack this PR to already add the has_atomic_execution flag to limit orders...

I'm already in the process of doing that. It's not much work I just got sidetracked a little.

> Another thing to consider is having a custom "cost-model" that more closely captures the actual gas usage of a 0x order.

Yeah, I think so too. Since they don't provide gas cost with their `OrderRecord`s I think we will have to simulate each order, no? Not sure if that is too much work to do for every 0x order.

Also @marcovc brought up that we should probably also send those orders during price estimation to get a representative estimate. This seems to involve either reimplementing a bunch of code or moving liquidity related code into from the `solver` crate into `shared`. I would do that in another PR because this would blow up this one too much.

nlordell:
> I think we will have to simulate each order, no?

IMO its fine to just use the mean gas cost for calling `fillOrKillLimitOrder` on the `0x` contract.

MartinquaXD:
With the help of @bh2smith we were able to establish a lower bound for a trade on 0x (min over last 2 months).
[This](https://dune.xyz/queries/492900/933779) is the query we used.

vkgnosis (approved):
Not sure if this is ready yet based on the discussion of whether the solver can handle these but code looks good.

nlordell:
> Also @marcovc brought up that we should probably also send those orders during price estimation to get a representative estimate. This seems to involve either reimplementing a bunch of code or moving liquidity related code into from the solver crate into shared.

YES! This has been something in the backlog for ages! The liquidity we use in the API is different than in the solver. That would be awesome to tackle (but quite an extensive task with a long thread of small changes to pull on).

MartinquaXD:
> Why did you chose a lower bound?

It looks like `gas_model.rs` uses lower bounds for all gas calculations so I thought it wouldn't be fair to use a different measurement for zeroex.

 > Also, this seems to be using the event which includes fillLimitOrder calls which I don't think we want...
 
 We were aware of this issue but is was our best idea to get **some** estimate. If you have a good alternative query I'm happy to use that instead.

nlordell:
> If you have a good alternative query I'm happy to use that instead.

Likely need to search the transaction table by sighash (first 4 bytes of calldata).

MartinquaXD:
I did a manual test where I checked simulations of a driver on this branch using a solver on the [corresponding branch](https://github.com/gnosis/gp-v2-solver-lib/pull/93).
In order to verify that complex multi-hop settlements don't produce simulation errors I made the solver assume that 0x trades don't incur any gas costs. This results in solutions which are way more complex than we would normally generate but that gives me more confidence in the test.
[simulation](https://dashboard.tenderly.co/gp-v2/staging/simulator/df4aa94c-c953-4cff-8c22-38ac78fe85e0) with 3 user orders, 2 uni v2 hops and 9 0x hops.
Before carefully ordering the execution plan simulations with as few as 2 0x hops already produced errors. I think it's safe to merge this PR now.

nlordell:
Very thorough testing.

Lets merge it!

MartinquaXD:
Updated the query to calculate the min gas cost per 0x transaction.
Previously we used the minimum of all transactions which created a `LimitOrderFilled` event which can also be emitted by the slightly cheaper `fillLimitOrder()` function (as opposed to `fillOrKillLimitOrder()`).

crates/solver/src/main.rs
```diff
@@ -280,6 +280,10 @@ struct Arguments {
     /// but at the same time we don't restrict solutions sizes too much
     #[clap(long, env, default_value = "15000000")]
     simulation_gas_limit: u128,
+
+    /// Controls whether 0x limit orders will be used as additional liquidity.
+    #[clap(long, env)]
+    with_zeroex_liquidity: bool,
```
nlordell, line 286: nit: Can this be part of the existing `--baseline-sources` flag?

nlordell (commented):
LGTM - just one nit about how 0x orders are enabled and disabled.

crates/solver/src/liquidity/zeroex.rs
```diff
@@ -0,0 +1,74 @@
...
+use shared::zeroex_api::{Order, OrderRecord, OrdersQuery, ZeroExApi};
+use std::sync::Arc;
+
+pub struct ZeroExLiquidity {
+    pub api: Arc<dyn ZeroExApi>,
+    pub zeroex: IZeroEx,
+}
+
+impl ZeroExLiquidity {
+    pub async fn get_liquidity(&self) -> Result<Vec<Liquidity>> {
```
nlordell, line 18: I think we should filter orders that aren't in the set of `user_order_tokens ∪ base_tokens`. This would be consistent with what we do for other liquidity sources like Uniswap and Balancer.

I think this can go in a follow up.

nlordell (approved):
LGTM

crates/solver/src/liquidity/zeroex.rs
MartinquaXD: I mean I can do that but I'm not sure what the purpose would be.
If quasimodo is able to handle all the orders wouldn't it be better to give it all the orders we have?
Same for Uniswap and Balancer.

nlordell: Maybe this is something that @marcovc can weigh-in on. Agree that if the solver can handle it than we should give it to the solver.

> Same for Uniswap

<details><summary>I think that would be 10s of thousands of pools.</summary>

```js
let count = 0;
let current = ethers.constants.AddressZero;
do {
  const { data: { pairs } } = await graphQuery(
    "uniswap/uniswap-v2",
    `{
      pairs(
        first: 1000,
        where: {
          id_gt: "${current}"
        }
      ) {
        id
      }
    }`,
  );
  count += pairs.length;
  current = pairs.pop()?.id;
} while (current);

console.log(`Final count: ${count}`);
```

I ran this and got `67538` pools.

</details>

PR #1711, opened 2022-03-09
Rewrite competition price estimator

vkgnosis:
To use the stream interface. In the process of this I also simplified most other parts of the file which is why I'm calling this rewrite rather than refactor.

- When all estimators error use the highest priority existing error rather than creating a new error "no successful price estimates".
- Logging the winner and observing it in metrics is separated from calculating the winner.
- Get rid of EstimateData intermediate struct.
- Get rid of computation of the sell_over_buy BigRational. Instead we compare the estimate out amount directly.

I thought I needed my generator helper for this but it turns out I don't.

### Test Plan

CI with existing tests and new one


crates/shared/src/price_estimation/competition.rs
```diff
@@ -34,62 +28,53 @@ impl RacingCompetitionPriceEstimator {
...
-        let mut computed_estimates = Vec::with_capacity(queries.len());
-
-        while !remaining_estimates.is_empty() {
-            let (estimate, _, rest) = future::select_all(remaining_estimates).await;
-            computed_estimates.push(estimate);
-            if self.enough_successful_estimates_for_each_query(queries, &computed_estimates) {
-                break;
+        let combined_stream = futures::stream::select_all(self.inner.iter().enumerate().map(
+            |(i, (_, estimator))| estimator.estimates(queries).map(move |result| (i, result)),
+        ));
```
MartinquaXD, line 46: nit: I find it a little confusing that the code to handle the `combined_streams` is split in two parts with `handle_single_result` in between. AFAICS this part could as well move to the bottom where `combined_stream` gets used to not have related logic together.

crates/shared/src/price_estimation/competition.rs
```diff
@@ -34,62 +28,53 @@ impl RacingCompetitionPriceEstimator {
...
+        // Stores the estimates for each query and estimator. When we have collected enough results
+        // to produce a result of our own the corresponding element is set to None.
+        let mut estimates: Vec<Option<Vec<(usize, PriceEstimateResult)>>> =
+            vec![Some(Vec::with_capacity(self.inner.len())); queries.len()];
+        let mut handle_single_result = move |estimator_index: usize, query_index: usize, result| {
```
MartinquaXD, line 51: nit: I find `handle_single_result` very hard to parse because everything can happen out of order so we have to juggle 2 indices around.
I think adding an empty line between the different blocks of the function and adding a small comment for each one instead of a single comment for the whole function could be more helpful for people unfamiliar with the code.
Maybe:
1) // Store result and determine if we have enough for this query.
2) // Determine the best result for this query.
3) // Log result.

MartinquaXD (commented):
I think the code itself is fine.
Just have some suggestions to rearrange/format the code in `estimates()` to make it easier to follow.

crates/shared/src/price_estimation/competition.rs
vkgnosis: The reason this code isn't directly in filter_map below is that I wanted to avoid more rightward drift as it would be indented an extra level and that I wanted to be able to use `?` inside of it which I can't do if it's inline. Then there was some extra annoyances where I couldn't just put an async block down there because it would run into lifetime problems.

The reason I ordered it this way is it made sense to me that you first see how the streams are combined and then how each result in the combined stream is handled. The filter_map really just glues them together but doesn't do anything interesting.

crates/shared/src/price_estimation/competition.rs
vkgnosis: Good suggestion. Will add some comments to make it easier to understand.

crates/shared/src/price_estimation/competition.rs
```diff
@@ -118,105 +103,54 @@ impl PriceEstimating for CompetitionPriceEstimator {
...
-                        winning_estimate.kind.label(),
-                    ])
-                    .inc();
-                winning_estimate.estimate
-            })
+fn best_result<'a>(
+    query: &Query,
+    results: impl Iterator<Item = &'a PriceEstimateResult>,
+) -> Option<(usize, &'a PriceEstimateResult)> {
+    results.enumerate().max_by(|a, b| {
```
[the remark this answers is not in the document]
vkgnosis: Not sure I understand you right. The thing that doesn't get used by the caller of this function is the &PriceEstimateResult reference so we could move that in here. In that case I think we should. I originally wrote this function before implementing the body above and thought I wanted the reference too but now it's clear we only need the index.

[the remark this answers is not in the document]
MartinquaXD: I think I misunderstood a little bit how `best_result()` worked so my comment didn't make too much sense.
The latest changes to `best_result()` helped resolve my confusion.
Looks good to me. 👍

crates/shared/src/price_estimation/competition.rs
MartinquaXD: Thanks. The new comments/white space help me very much. 👍

crates/shared/src/price_estimation/competition.rs
MartinquaXD: I totally agree that it's worth having a named closure for `handle_single_result`.

I can follow your reasoning for "interleaving" the code I just would have it done differently which added a bit of cognitive load to `estimates()` for me. I thought I would have to keep all the code of the `combined_stream` logic and `handle_single_result()` in memory at the same time for me to understand what's going on.

With the comments added to `handle_single_result()` it's easier to understand separate chunks of the logic so I don't feel strongly about reordering the logic anymore.

MartinquaXD (approved):

nlordell (approved):
Looks very good.

PR #1712, opened 2022-03-09
Using access lists

### LINKED ISSUE #1602: Populate access list for eip1559 transactions
Make use of the EIP-2930: Optional access lists:  https://eips.ethereum.org/EIPS/eip-2930 for transaction submission.

Gas reduction should be 100gas per accessed address and 100gas per accessed storage. 

For this simple transaction cost reduction should be 1800gas:
https://etherscan.io/tx/0x831c5057528e122540959b5d78167e46d28071129a7609c4d126b70775bbd600

~~Todo: investigate how to extract access list from `Settlement`~~
Access list should be fetched by calling the `eth_createAccessList` (geth should have it implemented)
Access list should be rebuilt on every new block, since theoretically, different EVM execution (therefore different access list) happens on a different blockchain state. This means that we need to rebuild the access list in every iteration of the submission loop.

sunce86:
This PR implements using the access lists for two components, settlement simulation and settlement submission.


crates/solver/src/settlement_simulation.rs
```diff
@@ -48,41 +54,47 @@ pub async fn simulate_and_estimate_gas_at_current_block(
...
+                    gas_price,
+                    &contract_with_buffered_transport,
+                    settlement.clone(),
+                    account.clone(),
+                )
+                .tx
+            })
+            .collect::<Vec<_>>();
+        let mut access_lists = access_list_estimator
+            .estimate_access_lists(txs.as_slice())
```
vkgnosis, line 76: I'm wondering if it would be cleaner to create the access lists when we also create the rest of the transaction itself and then pass it around to the simulation here and to the submitter later. That way the access list only has to be created once. Also it feels more proper to me that the access list would be part of the transaction and not be created on demand in several places.

crates/solver/src/settlement_simulation.rs
```diff
@@ -48,41 +54,47 @@ pub async fn simulate_and_estimate_gas_at_current_block(
...
+            .await
+            .unwrap_or_default()
+            .into_iter();
+        let calls = txs
+            .into_iter()
+            .map(|tx| {
+                let tx = match access_lists.next().unwrap() {
```
MartinquaXD, line 83: I believe `unwrap()` is able to panic here if `estimate_access_lists()` doesn't return an access list for every transaction because some bogous acess lists got filtered out.

Also if `estimate_access_lists()` throws an error `unwrap_or_default()` would return an empty `Vec` resulting in the same problem.

crates/solver/src/settlement_simulation.rs
```diff
@@ -91,12 +103,36 @@ pub async fn simulate_and_error_with_tenderly_link(
     network_id: &str,
     block: u64,
     simulation_gas_limit: u128,
+    access_list_estimator: Arc<dyn AccessListEstimating>,
 ) -> Vec<Result<()>> {
+    let settlements = settlements.collect::<Vec<_>>();
```
MartinquaXD, line 108: I think this `collect()` can be dropped by using the `settlements` iterator directly a few lines down.

crates/solver/src/settlement_simulation.rs
```diff
@@ -91,12 +103,36 @@ pub async fn simulate_and_error_with_tenderly_link(
...
+    let mut access_lists = access_list_estimator
+        .estimate_access_lists(&txs)
+        .await
+        .unwrap_or_default()
+        .into_iter();
+
+    // append access lists to existing methods if access lists exist, otherwise ignore access lists
+    let methods = methods
+        .into_iter()
+        .map(|method| match access_lists.next().unwrap() {
```
MartinquaXD, line 125: Same problem with `unwrap()` as above.

crates/solver/src/settlement_simulation.rs
```diff
@@ -91,12 +103,36 @@ pub async fn simulate_and_error_with_tenderly_link(
...
+            Ok(access_list) => method.access_list(access_list),
+            Err(_) => method,
+        })
+        .collect::<Vec<_>>();
```
MartinquaXD, line 129: No `collect()` needed here.

crates/solver/src/settlement_simulation.rs
sunce86: Access list is something that is theoretically dependent on the current network state, so it makes sense to me to be created at the moment of using. That is the reason why I estimate access list in every iteration of the submission loop, instead of doing it only once before entering the loop. Not sure about this though, if others think we should calculate it only once, I can change it

crates/solver/src/settlement_simulation.rs
sunce86: Changed the logic a bit.

crates/solver/src/settlement_simulation.rs
sunce86: Yup. Copy paste.

crates/solver/src/settlement_simulation.rs
sunce86: Fixed.

crates/solver/src/settlement_simulation.rs
sunce86: True :)

crates/solver/src/main.rs
```diff
@@ -184,6 +185,32 @@ struct Arguments {
...
+    /// `Web3`: supports every network.
+    #[clap(
+        long,
+        env,
+        default_values = &["Web3", "Tenderly"],
+        arg_enum,
+        ignore_case = true,
+        use_value_delimiter = true
+    )]
+    pub access_list_estimators: Vec<AccessListEstimatorType>,
```
nlordell, line 200: ubernit:
```suggestion
    access_list_estimators: Vec<AccessListEstimatorType>,
```

crates/solver/src/settlement_simulation.rs
nlordell: Makes sense to me. It reduces the change of paying for storage access you don't end up using.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -436,51 +444,88 @@ impl<'a> Submitter<'a> {
...
+            }
+        }
+    }
+
+    /// Estimate access list and validate
+    async fn estimate_access_list(
+        &self,
+        tx: &TransactionBuilder<DynTransport>,
+    ) -> Result<AccessList> {
+        let access_list = self.access_list_estimator.estimate_access_list(tx).await?;
```
nlordell, line 514: Maybe the sanitization of the access list should also move here? This way each implementation doesn't need to call it themselves.

nlordell (commented):
In general looks good.

How long does it take to generate access lists? I'm a bit concerned that we are adding more time to preparing and simulating settlement transactions which may have issues if we get "too busy".

Agree that we should re-compute the access list on each submission loop - but maybe we can try and reuse the first one (also reuse it for tenderly link).

crates/solver/src/settlement_submission/submitter.rs
sunce86: Hm, then I would have to add the sanitization for settlement simulation also, so it's not that beneficial, What we can do is to add the sanitization to `PriorityAccessListEstimating` but also make `NodeApi` and `TenderlyApi` private so we are forced to use `PriorityAccessListEstimating`

crates/solver/src/main.rs
sunce86: Fixed.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -436,51 +444,88 @@ impl<'a> Submitter<'a> {
...
+        let gas_before_access_list = tx.clone().estimate_gas().await?;
+        let tx = tx.clone().access_list(access_list.clone());
+        let gas_after_access_list = tx.clone().estimate_gas().await?;
```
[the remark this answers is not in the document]
sunce86: Indeed.

crates/solver/src/settlement_submission/submitter.rs
sunce86: Fixed as explained.
Also feels like filtering is the response of the estimator, not the client.

crates/solver/src/settlement_simulation.rs
sunce86: Refactored these two functions to receive already prepared access list. Access lists are now created once in driver.

sunce86:
Based on the logs it takes around 130ms for tenderly. Anyway, it shouldn't be an issue, will be checking the logs on staging.

nlordell:
> Based on the logs it takes around 130ms for tenderly.

My concern is more about this during "busy times". For example, if we need to simulate 30 settlements (which can happen because of settlement merging for example) then we need to compute 30x access lists.

To be clear, I'm not too worried about this to the point that I think it should block this change, just hoping we could find a way to quantify it.

nlordell:
Thanks very much for looking into this for me @sunce86.

crates/solver/src/main.rs
```diff
@@ -184,6 +185,32 @@ struct Arguments {
...
     transaction_strategy: Vec<TransactionStrategyArg>,
 
+    /// Which access list estimators to use. Multiple estimators are used in sequence if a previous one
+    /// fails. Individual estimators might support different networks.
+    /// `Tenderly`: supports every network.
+    /// `Web3`: supports every network.
+    #[clap(
+        long,
+        env,
+        default_values = &["Web3", "Tenderly"],
```
nlordell, line 195: nit: Can we make the default an empty list. My rationale is:
- I _think_ if you have default values for `use_value_delimiter` parameters, there is actually no way to configure it to use an empty list. At least this was the case a few months ago.
- Including `Tenderly` in the default requires an access key - this is annoying to configure for local testing
- Including `Web3` requires a node with `createAccessList` support which is not always the case for local testing or for all networks (does this work on GChain and Rinkeby for example?)

That being said, I do think that the value for our Mainnet solver pod should be what you have here, just maybe it makes more sense to set that in our Kubernetes deployment configuration instead of hardcoded defaults in the binary.

nlordell (approved):
Changes look good to me!

One small follow-up: Now that we have access-list support, we should be able to allow Smart Contracts to place orders buying Ether 🚀. So, we can now allow a `--allow-smart-contracts-buying-native-token` in the `orderbook` so that we can turn on this feature (my rationale for not having it on by default is that we can, in theory, configure no access list estimators, which means we need to be able to disable it in the orderbook API as well).

vkgnosis (approved):

crates/solver/src/main.rs
```diff
@@ -184,6 +185,32 @@ struct Arguments {
...
+        ignore_case = true,
+        use_value_delimiter = true
+    )]
+    access_list_estimators: Vec<AccessListEstimatorType>,
+
+    /// The URL for tenderly transaction simulation.
+    #[clap(
+        long,
+        env,
+        default_value = "http://api.tenderly.co/api/v1/account/sunce86/project/project/simulate"
```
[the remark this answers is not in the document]
sunce86: Indeed. Should probably define a new endpoint for deployment configuration. Fixed.

crates/solver/src/main.rs
sunce86: Yes, makes sense. Removed default.

> Including Web3 requires a node with createAccessList support which is not always the case for local testing or for all networks (does this work on GChain and Rinkeby for example?)

It does require a node with `createAccessList` support, however the code should not panic in case we define `Web3` and `createAccessList` is not supported on that node, because we agreed to have the access list functionality as an additional bonus, and to fallback to no access list if one does not exist for any reason.
Both client node (if it implements `createAccessList`) and tenderly support GChain and Rinkeby.

PR #1714, opened 2022-03-10
Store gas amount as u64 in price estimator

vkgnosis:
While some gas amounts are technically U256 on chain they are limited in practice to 30e6 which easily fits in u64 or even u32. Working with a native integer type is easier for programmers and faster so this PR converts the price estimator interface to use u64.
This might be followed up with similar conversions in other places.

### Test Plan

CI, existing tests ran some external price estimator tests manually as I changed the serialization type

### DIFF
--- a/crates/orderbook/src/api/post_quote.rs
+++ b/crates/orderbook/src/api/post_quote.rs
@@ -622,7 +622,7 @@ mod tests {
         let fee_calculator = Arc::new(fee_calculator);
         let price_estimator = FakePriceEstimator(price_estimation::Estimate {
             out_amount: 14.into(),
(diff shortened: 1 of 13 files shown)


nlordell:
> they are limited in practice to 30e6

This is not necessarily true for all usage of "gas". I think Geth, for example, accepts `eth_call`s with a gas limit much higher than the block gas limit.

Agree that it is more ergonomic and I can't think of any reason why a `u64` wouldn't be large enough for all our gas needs.

nlordell (approved):

MartinquaXD (approved):

PR #1715, opened 2022-03-10
Only do buffer trades, if sell and buy token are trusted

josojo:
We wanted to keep buffer tokens concentrated in a smaller set of tokens to enable buffer trading.
In order to archive this, we should not trade buffers of trusted tokens into the long-tail of arbitrary tokens. Hence, this PR ensures that both tokens, the buy and sell token are in the trusted token list, if an internal buffer trade is made.

### Test Plan
unit tests

### DIFF
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -212,7 +212,7 @@ impl Driver {
         settlement: &RatedSettlement,
         gas_price: EstimatedGasPrice,
     ) -> Result<bool> {
-        // We want to buy and sell only tokens that we trust. If no list is set, we settle with external liquidity.
+        // We want to only buy and sell tokens that we trust. If no list is set, we settle with external liquidity.
         if !self
             .market_makable_token_list
             .as_ref()
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -212,11 +212,11 @@ impl Driver {
         settlement: &RatedSettlement,
         gas_price: EstimatedGasPrice,
     ) -> Result<bool> {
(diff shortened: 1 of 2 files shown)


crates/solver/src/driver.rs
```diff
@@ -563,10 +563,14 @@ impl Driver {
...
 
-fn is_only_selling_trusted_tokens(settlement: &Settlement, token_list: &TokenList) -> bool {
-    !settlement
-        .traded_orders()
-        .any(|order| token_list.get(&order.creation.sell_token).is_none())
+fn is_only_selling_and_buying_trusted_tokens(
+    settlement: &Settlement,
+    token_list: &TokenList,
+) -> bool {
+    !settlement.traded_orders().any(|order| {
```
MartinquaXD, line 570: nit: Seems more intuitive to use to use [`Iterator::all()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all) to avoid a double negative here.

MartinquaXD (approved):

fleupold (changes requested):
> we should not trade buffers of trusted tokens into the long-tail of arbitrary tokens.

I don't think we are currently doing this. We are only trading with internal buffers if the user is **selling us** tokens that are in the buffer list. This means our balance in the tokens we like will increase. We are happy to have the user buy any shitcoin from us (and reduce the balance of this coin).

josojo:
Sorry, I got confused. Not sure what was going on in my head.

fleupold:
> Not sure what was going on in my head.

No problem, I have seen this confusion in the past. It may be because the solver team usually views this interaction from the other side? There, the smart contracts buys token (the sell token of the user) and sells token (the user's buy token).

PR #1717, opened 2022-03-11
Convert InstrumentedPriceEstimator to Stream

vkgnosis:
CI, existing tests

### DIFF
--- a/crates/shared/src/price_estimation/instrumented.rs
+++ b/crates/shared/src/price_estimation/instrumented.rs
@@ -1,9 +1,9 @@
-use crate::price_estimation::{
-    old_estimator_to_stream, vec_estimates, PriceEstimateResult, PriceEstimating,
-    PriceEstimationError, Query,
+use crate::price_estimation::{PriceEstimating, PriceEstimationError, Query};
+use futures::stream::StreamExt;
+use std::{
+    sync::Arc,
+    time::{Duration, Instant},
 };
-use std::sync::Arc;
-use std::time::{Duration, Instant};
 
 /// An instrumented price estimator.
 pub struct InstrumentedPriceEstimator {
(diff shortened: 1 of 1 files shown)


crates/shared/src/price_estimation/instrumented.rs
```diff
@@ -42,7 +30,20 @@ impl PriceEstimating for InstrumentedPriceEstimator {
...
+            self.metrics
+                .price_estimation_timed(&self.name, start.elapsed());
+        };
+        self.inner
+            .estimates(queries)
+            .map(move |result| {
+                let success = !matches!(&result.1, Err(PriceEstimationError::Other(_)));
+                self.metrics.price_estimated(&self.name, success);
+                result
+            })
```
MartinquaXD, line 44: nit: Seems like this would be a good use for [`inspect`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.inspect)

crates/shared/src/price_estimation/instrumented.rs
```diff
@@ -42,7 +30,20 @@ impl PriceEstimating for InstrumentedPriceEstimator {
...
+            .chain(futures::stream::once(measure_time).filter_map(|_| async { None }))
```
MartinquaXD, line 45: I think this does the right thing but working with streams certainly takes some time getting used to. 😅

MartinquaXD (approved):

nlordell (commented):
LGTM

I also slightly prefer `inspect` as it feels more appropriate than `map`.

nlordell (approved):

crates/shared/src/price_estimation/instrumented.rs
vkgnosis: Changed

PR #1718, opened 2022-03-11
Tiered cow subsidy

### LINKED ISSUE #1686: Add tiered COW subsidy support
This issue captures the work to add "tiered" support for COW/vCOW subsidies.

Since we don't really have an idea of how many tiers or amounts for each tier, we should make it configurable like:

```
--cow-fee-factors "10.2:0.9;100.0:0.75;1000.0:0.5"
```

This would meant that:
- If your balance is greater than `10.2 COW` (or `10200000000000000000` atoms) and less than `100.0 COW` then you get an additional 0.9 factor on your trade
- If your balance is greater than `100.0 COW` but less than `1000.0 COW` then you get a 0.75 additional factor
- If your balance is greater than `1000.0 COW` then you get a 0.5 additional factor

MartinquaXD:
Fixes #1686 

To be more flexible when offering subsidies based a user's COW balance 2 things got added:
1) Support different subsidy tiers based on the COW balance
2) Update CLI arguments to configure subsidy tiers at startup

The help for the new CLI argument looks like this:
```
--cow-fee-factors <COW_FEE_FACTORS>
    Used to configure how much of the regular fee a user should pay based on their COW
    balance.

    The expected format is "10.2:0.75,150:0.5" for 2 subsidy tiers. A balance of [10.2,150)
    COW will cause you to pay 75% of the regular fee and a balance of [150, inf) COW will
    cause you to pay 50% of the regular fee.

    If multiple tiers get configured for the same threshold the first will be used.

    [env: COW_FEE_FACTORS=]
```

Note that a `SubsidyTier` can only be created with a fee factor in the range of [0.0, 1.0].

### Test Plan
Added unit test which verifies that the correct fee factor gets looked up based on a balance amount.


crates/orderbook/src/cow_subsidy.rs
```diff
@@ -24,67 +24,114 @@ impl Default for FixedCowSubsidy {
...
+            .map_err(|_| anyhow::anyhow!("can not parse threshold {} as f64", threshold))?;
+
+        let fee_factor = parts.next().ok_or(anyhow::anyhow!("missing fee factor"))?;
+        let fee_factor: f64 = fee_factor
+            .parse()
+            .map_err(|_| anyhow::anyhow!("can not parse fee factor {} as f64", fee_factor))?;
+
+        if parts.next().is_some() {
+            anyhow::bail!("too many arguments for subsidy tier");
+        }
```
nlordell, line 58: I recently discovered [`split_once`](https://doc.rust-lang.org/std/primitive.str.html#method.split_once) when doing this year's Advent of Code which I think would be quite helpful here.

crates/orderbook/src/cow_subsidy.rs
```diff
@@ -24,67 +24,114 @@ impl Default for FixedCowSubsidy {
...
+    }
 }
 
 pub struct CowSubsidyImpl {
     cow_token: ERC20,
-    threshold: U256,
-    factor: f64,
-    cache: Mutex<TimedSizedCache<H160, SubsidyTier>>,
+    // sorted by threshold in increasing order, no duplicated thresholds
+    subsidy_tiers: Vec<SubsidyTier>,
```
nlordell, line 81: Maybe a `BTreeMap<Threshold, Factor>` might be appropriate here?

MartinquaXD: I thought since we will never lookup a key but instead always loop through the keys a map would be weird.
Don't have a strong opinion, though.

nlordell (commented):
Code looks good to me!

Just one small suggestion around using `split_once` and discussion around using a btree for the subsidy factors.

crates/orderbook/src/cow_subsidy.rs
nlordell: > never lookup a key but instead always loop through the keys a map would be weird

It allows you to binary search for the key. [For example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=67bc32d9559e6905dafd1bbe270062ba):
```rust
let tier = map.range(..balance).rev().next();
```

MartinquaXD: `BTreeMap::range` seems to work. In that case the reasons for the map outweigh the one against. 👍

nlordell (approved):
Approving because I don't feel strongly about using a b-tree here.

vkgnosis (approved):

crates/orderbook/src/cow_subsidy.rs
```diff
@@ -24,66 +25,89 @@ impl Default for FixedCowSubsidy {
...
+                .parse()
+                .map_err(|_| anyhow::anyhow!("can not parse threshold \"{}\" as f64", threshold))?;
+            let threshold = U256::from_f64_lossy(threshold * 1e18);
+
+            let fee_factor: f64 = fee_factor.parse().map_err(|_| {
+                anyhow::anyhow!("can not parse fee factor \"{}\" as f64", fee_factor)
+            })?;
+            if !(0.0..=1.0).contains(&fee_factor) {
+                anyhow::bail!("fee factor must be in the range of [0.0, 1.0]");
+            }
```
nlordell, line 60: Last `anyhow`-suggestion, I promise 😛. There is an `ensure!` macro which is like an `assert!` but returns an error instead of panicking.
```suggestion
            anyhow::ensure!(
                (0.0..=1.0).contains(&fee_factor),
                "fee factor must be in the range of [0.0, 1.0]"
            );
```

nlordell (approved):
Made some tiny suggestions to use some additional `anyhow` ergonomic features. Again, not blocking merge from my point of view.

sunce86 (approved):
Lg.

PR #1719, opened 2022-03-12
Abstract Uniswap V2 Pool State Reading

nlordell:
This PR modifies the Uniswap pool fetching component to be abstract over some pool reading logic.

Specifically, the existing `PoolFetcher` is now generic over a `PoolReading` which implements the logic of actually reading the pool reserves on-chain.

Why? I'm glad you asked! We previously had initial Swapr pool support (a Uniswap V2 clone on Gnosis Chain) but had to disable it because it wasn't working as expected. Specifically, the Uniswap state reading logic **assumes a 0.3% fee**. This, however, is not true for Swapr pools (they default to 0.25% and can be modified by a special "owner" account). This means that we need to additionally read the current fee for Swapr pools on top of reading pool reserves and token balances.

In a follow up PR, I plan on introducing a `SwaprPoolReader` component that wraps a `DefaultPoolReader` and additionally fetches the current pool fee on-chain in the same batch call.

### Test Plan

This PR doesn't add or change functionality, just moves the Uniswap V2 pool state reading logic into its own component, so existing CI should be sufficient testing.


bh2smith (approved):
Great description and not a very invasive change. I notice that this merely adds the uniswap pool fetching and doesn't do much to reinstate swapr. Is it safe to assume this is being saved for later?

vkgnosis (approved):

crates/shared/src/sources/uniswap_v2/pool_fetching.rs
```diff
@@ -25,6 +29,16 @@ pub trait PoolFetching: Send + Sync {
...
 
+/// Trait for abstracting the on-chain reading logic for pool state.
+pub trait PoolReading: Send + Sync {
+    fn read_state(
+        &self,
+        pair: TokenPair,
+        batch: &mut Web3CallBatch,
+        block: BlockId,
+    ) -> BoxFuture<'_, Result<Option<Pool>>>;
+}
```
MartinquaXD, line 40: Is there a reason why this doesn't use `async_trait`? I think having that consistent across the code base would be nice.

MartinquaXD (approved):

crates/shared/src/sources/uniswap_v2/pool_fetching.rs
nlordell: Yes there is. Specifically, when making a "batched" fetching interface, you need to add Ehereum RPC request to the batch synchronously and then return a future that combines the result of those batched RPC requests into a `Result<Pool>`. If you use a `async_trait`, then the requests will only get added to the batch after the first poll.

I will add a comment explaining this.

PR #1720, opened 2022-03-12
Add Swapr-specific `PoolReading` Implementation

nlordell:
Follow up to #1719

This PR adds a Swapr-specific `PoolReading` implementation that wraps the exising `DefaultPoolReader` to additionally query AMM fees. This is required because:
- The fee defaults to something different than Uniswap (0.25% instead of 0.3%)
- Can be modified at any point in time by some DEX admin (so can't be read once and cached forever).

Note that this reader is still not being used. In order to hook it up to the current `PoolAggregator` that fetches all Uniswap-v2-like pools requires some additional refactoring that will come in a follow up PR.

### Test Plan

Added some unit tests around the logic for building a pool for its asynchronously fetched parts. Also added a manual ignore test:
```
% cargo test -p shared -- --ignored --nocapture fetch_swapr_pool
WETH <> wxDAI pool: Pool {
    tokens: TokenPair(
        0x6a023ccd1ff6f2045c3309768ead9e68f978f6e1,
        0xe91d153e0b41518a2ce8dd3d7944fa863463a97d,
    ),
    reserves: (
        379003626006978207230,
        979677523238372752737977,
    ),
    fee: Ratio {
        numer: 1,
        denom: 400,
    },
}
```

Notice that the fee is `1/400 = 0.0025 = 0.25% != 0.3%` 🎉.


crates/contracts/src/bin/vendor.rs
```diff
@@ -48,7 +48,7 @@ fn run() -> Result<()> {
         )?
         .github(
             "BalancerV2StablePoolFactory",
-            "balancer-labs/balancer-subgraph-v2/2b97edd5e65aed06718ce64a69111ccdabccf048/\
+            "balancer-labs/balancer-v2-monorepo/ad1442113b26ec22081c2047e2ec95355a7f12ba/\
```
nlordell, line 51: Unrelated change, but this just fixes an issue where the `vendor` script was failing because the URI to this artifact was wrong (accidentally introduced in https://github.com/gnosis/gp-v2-services/pull/1403).

crates/contracts/src/bin/vendor.rs
```diff
@@ -96,7 +96,7 @@ fn run() -> Result<()> {
         .github(
             "BalancerV2StablePool",
             "balancer-labs/balancer-subgraph-v2/2b97edd5e65aed06718ce64a69111ccdabccf048/\
-             pkg/deployments/tasks/20210624-stable-pool/abi/StablePool.json",
+             abis/StablePool.json",
```
nlordell, line 99: Same here.

crates/contracts/src/bin/vendor.rs
```diff
@@ -129,7 +129,12 @@ fn run() -> Result<()> {
         .github(
             "IZeroEx",
             "0xProject/protocol/c1177416f50c2465ee030dacc14ff996eebd4e74/\
-            packages/contract-artifacts/artifacts/IZeroEx.json",
+             packages/contract-artifacts/artifacts/IZeroEx.json",
+        )?
+        .github(
+            "ISwaprPair",
+            "levelkdev/dxswap-core/3511bab996096f9c9c9bc3af0d94222650fd1e40/\
+             build/IDXswapPair.json",
```
nlordell, line 137: Swapr is formally known as DXswap - so the naming is confusing 😵‍💫

crates/shared/src/sources/swapr/reader.rs
```diff
@@ -0,0 +1,134 @@
...
+use ethcontract::{errors::MethodError, BlockId};
+use futures::{future::BoxFuture, FutureExt as _};
+use model::TokenPair;
+use num::rational::Ratio;
+
+/// A specialized Uniswap-like pool reader for DXdao Swapr pools.
+///
+/// Specifically, Swapr pools have dynamic fees that need to be fetched with the
+/// pool state.
+pub struct SwaprPoolReader(DefaultPoolReader);
```
bh2smith, line 18: So the default pool reader is the one which also fetches fees. Makes sense because uniswap fees are permanently fixed to be 0.3%. Is that true? Would it be possible for the uni holders to vote to change the fees for a specific pool? I haven't actually checked, just wondering.

bh2smith (approved):

crates/shared/src/sources/swapr/reader.rs
nlordell: > So the default pool reader is the one which also fetches fees

No, only the `SwaprPoolReader` does. The default one doesn't have to because Uniswap pools have a fixed 0.3% fee.

> Would it be possible for the uni holders to vote to change the fees for a specific pool?

No, its a hard-coded value: 

https://github.com/Uniswap/v2-core/blob/4dd59067c76dea4a0e8e4bfdda41877a6b16dedc/contracts/UniswapV2Pair.sol#L180-L181

vkgnosis (approved):

PR #1721, opened 2022-03-12
Use Swapr-specific Pool Reader

nlordell:
Follow-up to #1719 and #1720 

This PR moves `dyn PoolFetching` initialization into each UniswapV2-like liquidity source module. Previously, each module only created their own `PairProvider` instances since they were liquidity-configuration-dependent (specifically, it depended on the factory address and init code hash which was different for each UniswapV2-like liquidity source). However, it turns out Swapr also has specific pool state fetching logic (see aforementioned PRs for more details), which means that each UniswapV2-like liquidity module **also** needs to be responsible for creating their own `dyn PoolFetching` instance (since the `PoolReading` implementation varies from liquidity source to liquidity source).

The end-result: Swapr liquidity works correctly now! It should no longer have a fixed fee of 0.3% and now read it from the block chain when fetching liquidity.

### Test Plan

CI and Rust type checker. The only "new logic" is that the Swapr `dyn PoolFetching` is created with `SwaprPoolReader` instead of a `DefaultPoolReader`. The existing manual test that fetched an on-chain Swapr pool was modified to use the module-level created `dyn PoolFetching`:

```
% cargo test -p shared -- --ignored --nocapture swapr
    Finished test [unoptimized + debuginfo] target(s) in 0.21s
     Running unittests (target/debug/deps/shared-25ec9fa0fee90af3)

running 1 test
WETH <> wxDAI pool: Pool {
    tokens: TokenPair(
        0x6a023ccd1ff6f2045c3309768ead9e68f978f6e1,
        0xe91d153e0b41518a2ce8dd3d7944fa863463a97d,
    ),
    reserves: (
        379042671808702239633,
        979576897975408422373175,
    ),
    fee: Ratio {
        numer: 1,
        denom: 400,
    },
}
test sources::swapr::reader::tests::fetch_swapr_pool ... ok
```

Note that this test still returns a on-chain fetched fee of 0.25% instead of defaulting to 0.3%

### Release Notes

Swapr liquidity can now be re-enabled on Gnosis Chain.

### DIFF
--- a/crates/e2e/tests/e2e/services.rs
+++ b/crates/e2e/tests/e2e/services.rs
@@ -116,7 +116,7 @@ impl OrderbookServices {
                 maximum_recent_block_age: 4,
                 ..Default::default()
             },
-            Box::new(PoolFetcher::uniswap(pair_provider, web3.clone())),
+            Arc::new(PoolFetcher::uniswap(pair_provider, web3.clone())),
             current_block_stream.clone(),
             Arc::new(NoopPoolCacheMetrics),
         )
--- a/crates/orderbook/src/main.rs
+++ b/crates/orderbook/src/main.rs
@@ -58,10 +58,7 @@ use shared::{
     sources::{
         self,
         balancer_v2::{pool_fetching::BalancerContracts, BalancerPoolFetcher},
-        uniswap_v2::{
-            pool_cache::PoolCache,
-            pool_fetching::{PoolFetcher, PoolFetching},
-        },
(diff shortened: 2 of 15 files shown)


vkgnosis (approved):

MartinquaXD (approved):
LGTM.
> This change makes Swapr liquidity work correctly now! It should no

This sentence in the PR description cuts off.

nlordell:
> This sentence in the PR description cuts off.

Oops 🙈! Updated the description to be more grammatically correct.

PR #1724, opened 2022-03-13
Change In-Flight Order Check To Be Pessimistic

nlordell:
This PR changes the in-flight order check to be pessimistic when it comes to in-flight partially fillable orders.

Specifically, it now filters out orders that are partially fillable and in flight **even if the order has a remaining amount to be filled**. This is in contrast to the optimistic approach it used to take where partially fillable orders would not be filtered out, meaning that a potentially invalid order was being passed to the solvers (where its maximum fillable amount was misrepresented). This will prevent partially fillable orders where only a portion was executed to be considered in the following batch even if it has some remaining amount (see issue https://github.com/gnosis/gp-v2-services/issues/1723).

### Test Plan

Adjusted unit test.


bh2smith (approved):
Nice and easy!

vkgnosis (approved):

PR #1725, opened 2022-03-13
Scale Partially Filled Limit Orders

nlordell:
This PR modifies the limit order converter to scale partially filled orders into the `LimitOrder` liquidity type. This uses the shared code for computing remaining amounts that was introduced in #1722.

### Test Plan

Added a unit test verifying partially filled orders get scaled.


vkgnosis (approved):

MartinquaXD (approved):
Code looks good.

PR #1726, opened 2022-03-14
Allow Liquidity Order Owners to Place Partially Fillable Orders

nlordell:
This PR moves the `--liquidity-order-owners` (a.k.a. PMMs) flag into the shared arguments and allows partially fillable orders to be placed when the owner is one of those specified addresses.

### Test Plan

Added unit test case to check that partially fillable orders from liquidity order owners.

### DIFF
--- a/crates/orderbook/src/api/order_validation.rs
+++ b/crates/orderbook/src/api/order_validation.rs
@@ -681,7 +681,7 @@ mod tests {
 
     #[tokio::test]
     async fn pre_validate_ok() {
-        let liquidity_order_owner = H160::from_low_u64_be(1);
+        let liquidity_order_owner = H160::from_low_u64_be(0x42);
         let min_order_validity_period = Duration::from_secs(1);
         let validator = OrderValidator::new(
             Box::new(MockCodeFetching::new()),
--- a/crates/orderbook/src/api/order_validation.rs
+++ b/crates/orderbook/src/api/order_validation.rs
@@ -15,7 +15,7 @@ use model::{
 use shared::{
     bad_token::BadTokenDetecting, price_estimation::PriceEstimationError, web3_traits::CodeFetching,
 };
-use std::{sync::Arc, time::Duration};
+use std::{collections::HashSet, sync::Arc, time::Duration};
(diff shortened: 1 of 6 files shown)


crates/orderbook/src/api/order_validation.rs
```diff
@@ -679,24 +681,33 @@ mod tests {
...
+            min_order_validity_period,
             Arc::new(MockMinFeeCalculating::new()),
             Arc::new(MockBadTokenDetecting::new()),
             Arc::new(MockBalanceFetching::new()),
         );
+        let order = || PreOrderData {
+            valid_to: shared::time::now_in_epoch_seconds()
+                + min_order_validity_period.as_secs() as u32
+                + 2,
+            sell_token: H160::from_low_u64_be(1),
```
MartinquaXD, line 700: nit: Little surprising that the `sell_token` uses the same address as `liquidity_order_owner`.

crates/shared/src/arguments.rs
```diff
@@ -130,6 +130,15 @@ pub struct Arguments {
     /// The 1Inch REST API URL to use.
     #[structopt(long, env, default_value = "https://api.1inch.exchange/")]
     pub one_inch_url: Url,
+
+    /// The configured addresses whose orders should be considered liquidity and
+    /// not regular user orders.
+    ///
+    /// These orders have special semantics such as not being considered in the
+    /// settlements objective funtion, not receiving any surplus, and being
+    /// allowed to place special kinds of orders.
```
MartinquaXD, line 139: > special kinds of orders.

Does it make sense to explicitly name partially fillable orders here?

MartinquaXD (approved):

vkgnosis (approved):

crates/orderbook/src/api/order_validation.rs
nlordell: Good point, it doesn't matter that they are the same, but it makes the intent of the test less clear. Will change.

nlordell: Fixed.

PR #1727, opened 2022-03-14
No Minimum Fee Required for Liquidity Owners

nlordell:
This PR modifies the `MinFeeCalculator` to be aware of of liquidity order owners and require 0 minimum fees.

The rationale is that they already pay very small fees, so there is no reason not to make them 0. This is especially true considering we want to allow PMMs to place partially fillable orders, where they can pretty much achieve 0-fees by placing larger orders such that the executed fee is very tiny (and effectively pretty much 0).

One concern we had about this was that PMMs could use 0-fee orders for wash trading. However, I don't think this is a legitimate concern since solvers already don't settle PMM orders unless it facilitates a user trade and improves over a similar settlement with an AMM. Thus, 0-fee PMM orders would not be very effective for wash trading.

I had implemented this by extending the `MinFeeCalculator` type. It could have been implemented by wrapping it instead with something like:
```rust
struct LiquidityOrderOwnerAwareMinFeeCalculator {
    inner: Arc<dyn MinFeeCalculating>,
    liquidity_order_owners: HashSet<H160>,
}
```

But it felt a bit overkill for the small additional condition this change added. WDYT?

### Test Plan

Added a unit test verifying that PMMs indeed pay no fees.

Additionally, I ran a local orderbook and solver on Rinkeby and placed a user order with a counter PMM order: <https://rinkeby.etherscan.io/tx/0xd37d653712d05d33506095331cf2d4cab9aa3dcd159bcc0ee26f7bc2b75dfb37>. Unfortunately, [Tenderly isn't printing a very helpful call trace](https://dashboard.tenderly.co/gp-v2/staging/tx/rinkeby/0xd37d653712d05d33506095331cf2d4cab9aa3dcd159bcc0ee26f7bc2b75dfb37/debugger?trace=0.1), but it does at least indicate that its hitting the `partiallyFillable` order path:

image

<details><summary>Additionally, I was able to query my local orderbook for the PMM order and see it was only partially filled:</summary>

```
% curl -s 'http://localhost:8080/api/v1/orders/0x8D0F6FD101C12B7032F4D7F00C95D147F5C79F1C83B24FC4DA54C9AD4C1EFB743A2C5D8F209F12BDCD84A46D6EEB136CBB0CCB9C80000000' | jq
{
  "creationDate": "2022-03-14T22:09:35.167076Z",
  "owner": "0x3a2c5d8f209f12bdcd84a46d6eeb136cbb0ccb9c",
  "uid": "0x8d0f6fd101c12b7032f4d7f00c95d147f5c79f1c83b24fc4da54c9ad4c1efb743a2c5d8f209f12bdcd84a46d6eeb136cbb0ccb9c80000000",
  "availableBalance": "7666960556003232805048",
  "executedBuyAmount": "999779582997575397",
  "executedSellAmount": "1333039443996767194952",
  "executedSellAmountBeforeFees": "1333039443996767194952",
  "executedFeeAmount": "0",
  "invalidated": false,
  "status": "open",
  "settlementContract": "0x9008d19f58aabd9ed0d60971565aa8510560ab41",
  "fullFeeAmount": "0",
  "sellToken": "0xa7d1c04faf998f9161fc9f800a99a809b84cfc9d",
  "buyToken": "0xc778417e063141139fce010982780140aa0cd5ab",
  "sellAmount": "2400000000000000000000",
  "buyAmount": "1800000000000000000",
  "validTo": 2147483648,
  "feeAmount": "0",
  "kind": "sell",
  "partiallyFillable": true,
  "signingScheme": "eip712",
  "signature": "0xbbfe981aa66900a36d7f13d2407c0b0062ae0bf5913c89b7803f4d4bb488ed2f236abd920c023608670ded935890ceadad307a43d12315396d4a2f9ee43de90f1c",
  "sellTokenBalance": "erc20",
  "buyTokenBalance": "erc20"
}
```

</details>

### Release notes

PMM addresses should now be configured as liquidity order owners instead of using custom app IDs.

### DIFF
--- a/crates/orderbook/src/fee.rs
+++ b/crates/orderbook/src/fee.rs
@@ -109,6 +109,12 @@ impl Default for FeeParameters {
         Self {
             gas_amount: 0.,
             gas_price: 0.,
+            // We can't use `derive(Default)` because then this field would have
+            // a value of `0.` and it is used in division. The actual value we
+            // use here doesn't really matter as long as its non-zero (since the
+            // resulting amount in native token or sell token will be 0
+            // regardless), but the multiplicative identity seemed like a
+            // natural default value to use.
             sell_token_price: 1.,
         }
     }
--- a/crates/orderbook/src/fee.rs
+++ b/crates/orderbook/src/fee.rs
@@ -11,7 +11,7 @@ use shared::{
     price_estimation::{native::NativePriceEstimating, single_estimate},
 };
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     sync::{Arc, Mutex},
 };
 
@@ -84,6 +84,7 @@ pub struct MinFeeCalculator {
     fee_subsidy: FeeSubsidyConfiguration,
     native_price_estimator: Arc<dyn NativePriceEstimating>,
     cow_subsidy: Arc<dyn CowSubsidy>,
+    liquidity_order_owners: HashSet<H160>,
 }
 
 #[derive(Debug, Default, Clone, Copy, Eq, PartialEq)]
@@ -96,11 +97,21 @@ pub struct FeeData {
 }
 
 /// Everything required to compute the fee amount in sell token
-#[derive(Debug, Default, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, Copy, PartialEq)]
 pub struct FeeParameters {
     pub gas_amount: f64,
     pub gas_price: f64,
     pub sell_token_price: f64,
+}
+
+impl Default for FeeParameters {
+    fn default() -> Self {
+        Self {
+            gas_amount: 0.,
+            gas_price: 0.,
+            sell_token_price: 1.,
+        }
+    }
 }
 
 // We want the conversion from f64 to U256 to use ceil because:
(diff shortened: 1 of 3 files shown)


vkgnosis (approved):
Argument for allowing 0 fees makes sense to me.

crates/orderbook/src/fee.rs
```diff
@@ -96,13 +97,23 @@ pub struct FeeData {
...
     pub gas_price: f64,
     pub sell_token_price: f64,
 }
 
+impl Default for FeeParameters {
+    fn default() -> Self {
+        Self {
+            gas_amount: 0.,
+            gas_price: 0.,
+            sell_token_price: 1.,
```
MartinquaXD, line 112: Just to make sure I understand the change around `Default` right:
When we have to return a `FeeParameter` for a partially fillable order it doesn't really matter which exact `sell_token_price` gets used because we also specify a fee amount of `0`. As long as `sell_token_price` is not `0` (to avoid division by 0), we are fine?
If that's not the case I'm not sure how you arrived at `1.0` for the price.

nlordell: This is correct. I just thought that the multiplicative identity made sense for the price. I will add a comment.

MartinquaXD (approved):

PR #1732, opened 2022-03-16
flashbots cancellation added

### LINKED ISSUE #1467: Implement cancellation for flashbots settlement submitter
Flashbots announced that they now support cancellation of transactions.

Need to implement the function: https://github.com/gnosis/gp-v2-services/blob/7dab6e20e5c8f5e9e640bbb17abf45a8fa897b73/crates/solver/src/settlement_submission/flashbots_api.rs#L64

sunce86:
Fixes https://github.com/gnosis/gp-v2-services/issues/1467

Manual test passed.


vkgnosis (approved):
Not sure if others feel it is needed but it might be nice to add the same flashbots docs link that you put into the issue here in the code to explain readers at a later point why this is the correct way to cancel a tx.

MartinquaXD (approved):

PR #1733, opened 2022-03-17
Move price estimator request sharing into more generic component

vkgnosis:
While working on refactoring the buffered price estimator to use the streaming interface I realized that I was dealing with a lot of complexity due to the batch interface when the actual estimators that we want this for are all mostly based around making independent single requests (the only price estimator really making use of batches is naive).

While playing with this idea I realized I could also build a more generic request sharing component that isn't tied to the price estimator interface which could be used in any component that needs similar functionality.

This PR is the result of this idea. There is a new single request -> single response shared future component that can be used directly by the price estimators for which this makes sense.

The shared request component is based mostly on Martin's BufferedEstimator but simplified because it doesn't have to handle batches. I played around with a few other approaches (like garbage collect on Drop) but decided extra complexity wasn't worth it. It possible that in the future we end up needing batched request sharing after all, in which case some of that code can be revived.

This type of sharing also has the advantage that it is more fine grained. We can share only the actual request to the external service without having to share all the async work performed. For example the naive price estimator uses an internal pool cache. There is no need to request-share the updating of that cache but with the existing BufferingPriceEstimator we cannot control this.

TODO:
I only made one price estimator use the new component as an example for the PR. Before merging I need to do this to the other external estimators.

### Test Plan

CI, new component has new tests


crates/shared/src/price_estimation/oneinch.rs
```diff
@@ -1,15 +1,22 @@
...
-use futures::StreamExt;
+use futures::{future::BoxFuture, FutureExt, StreamExt};
 use model::order::OrderKind;
 use std::sync::Arc;
 
 pub struct OneInchPriceEstimator {
     api: Arc<dyn OneInchClient>,
+    sharing: RequestSharing<
+        Query,
+        BoxFuture<'static, Result<RestResponse<SellOrderQuote>, PriceEstimationError>>,
```
vkgnosis, line 18: Slightly awkward here is that the future must be 'static even though a hypothetical 'self would fit better.

vkgnosis, line 18: The return type here is what OneInchClient returns for the request we want to share. There is no need to share the get_allowed_protocols call below because that is already cached and independent of the specific query.
The error type had to be adjusted to PriceEstimationError because the future's output must be clonable.

MartinquaXD (approved):
Awesome that this can be simplified so much when you not have to take batches into account.

crates/shared/src/price_estimation/oneinch.rs
```diff
@@ -1,15 +1,22 @@
...
+    request_sharing::RequestSharing,
 };
-use futures::StreamExt;
+use futures::{future::BoxFuture, FutureExt, StreamExt};
 use model::order::OrderKind;
 use std::sync::Arc;
 
 pub struct OneInchPriceEstimator {
     api: Arc<dyn OneInchClient>,
+    sharing: RequestSharing<
```
nlordell, line 16: One thing that I'm not super fond of is that each estimator needs its own `sharing` instance right?

Would't you also be able to implement a `BufferedPriceEstimator` around another `SinglePriceEstimating` using the exact same helper you created? Might make composition easier (instead of relying on adding a new field on all existing estimators).

Kind of how we have a specialized `SingleOrderSolver` in the driver.

nlordell, line 16: One thing that I'm not super fond of is that each estimator needs its own `sharing` instance right?

Would't you also be able to implement a `BufferedPriceEstimator` around another `SinglePriceEstimating` using the exact same helper you created? Might make composition easier (instead of relying on adding a new field on all existing estimators).

Kind of how we have a specialized `SingleOrderSolver` in the driver.

crates/shared/src/price_estimation/oneinch.rs
nlordell: Self-referential struct 🙅

nlordell: Self-referential struct 🙅

nlordell (commented):
This component is super cool! I really like how it became much simpler because it deals only with single requests.

That being said, I think it might be good to create an abstraction around a `SingleQueryPriceEstimating` that can be used to create a `BufferedPriceEstimator` using the new request sharing type you created.

crates/shared/src/price_estimation/oneinch.rs
vkgnosis: I thought about this while implementing but didn't felt the abstraction was that useful because there might be some work that you want to make part of the shared request in some estimators but not others. Also adding this directly is easy so not sure of extra trait etc was worth it. I can try to doing this in a follow up PR to be more sure.

nlordell (approved):

crates/shared/src/price_estimation/oneinch.rs
nlordell: OK, might be worth experimenting. You are right that you have more control if you can chose which part of the estimate to share.

PR #1734, opened 2022-03-18
Limit time spent in auction creation

vkgnosis:
This makes use of the streaming price estimator interface changes by streaming as many as prices as possible within a fixed time limit that we reserve for this. Orders for which we did not get prices in time are excluded. Since price estimation is cached it is likely that in the next round of auction creation more orders can be included.

### Test Plan

CI, new test

### DIFF
--- a/crates/orderbook/Cargo.toml
+++ b/crates/orderbook/Cargo.toml
@@ -52,3 +52,4 @@ web3 = { version = "0.18", default-features = false }
 secp256k1 = "0.21"
 mockall = "0.11"
 testlib = { path = "../testlib" }
+tokio = { version = "1.15", features = ["test-util"] }
--- a/crates/orderbook/src/solvable_orders.rs
+++ b/crates/orderbook/src/solvable_orders.rs
@@ -4,22 +4,25 @@ use crate::{
     orderbook::filter_unsupported_tokens,
 };
 use anyhow::{Context as _, Result};
+use futures::StreamExt;
 use model::{auction::Auction, order::Order};
 use primitive_types::{H160, U256};
 use shared::{
-    bad_token::BadTokenDetecting,
-    current_block::CurrentBlockStream,
-    maintenance::Maintaining,
-    price_estimation::native::{native_vec_estimates, NativePriceEstimating},
-    time::now_in_epoch_seconds,
+    bad_token::BadTokenDetecting, current_block::CurrentBlockStream, maintenance::Maintaining,
+    price_estimation::native::NativePriceEstimating, time::now_in_epoch_seconds,
 };
 use std::{
     collections::{BTreeMap, HashMap, HashSet},
     iter::FromIterator,
     sync::{Arc, Mutex, Weak},
-    time::{Duration, Instant},
+    time::Duration,
 };
-use tokio::sync::Notify;
+use tokio::{sync::Notify, time::Instant};
+
+// When creating the auction after solvable orders change we need to fetch native prices for a
+// potentially large amount of tokens. This is the maximum amount of time we allot for this
+// operation.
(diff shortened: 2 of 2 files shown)


crates/orderbook/Cargo.toml
```diff
@@ -52,3 +52,4 @@ web3 = { version = "0.18", default-features = false }
 secp256k1 = "0.21"
 mockall = "0.11"
 testlib = { path = "../testlib" }
+tokio = { version = "1.15", features = ["test-util"] }
```
nlordell, line 55: Aren't we on `tokio v1.17` now?

crates/orderbook/src/solvable_orders.rs
```diff
@@ -327,41 +334,64 @@ impl Maintaining for SolvableOrdersCache {
...
+            let price = match to_normalized_price(price) {
+                Some(price) => price,
+                None => continue,
+            };
+            prices.insert(*token, price);
+        }
+    };
+    match tokio::time::timeout_at(deadline, collect_prices).await {
+        Ok(()) => (),
+        Err(_) => tracing::warn!(
```
nlordell, line 366: Can we maybe add some metrics to this to?
- How many orders are missing?
- How often the timeout happens

Can be in a separate PR as well.

crates/orderbook/src/solvable_orders.rs
```diff
@@ -327,41 +334,64 @@ impl Maintaining for SolvableOrdersCache {
...
+            "auction native price collection took too long, got {} out of {}",
+            prices.len(),
+            traded_tokens.len()
+        ),
+    }
 
+    // Filter both orders and prices so that we only return orders that have prices and prices that
+    // have orders.
```
nlordell, line 374: Ah! This is important because we may have gotten the price for one of the tokens of an order, but not the other, and the token we have the price for only appears in that order. I.e. we are missing a token for an order, so we exclude the order; in excluding the order, we have a price with no equivalent order.

crates/orderbook/src/solvable_orders.rs
```diff
@@ -327,41 +334,64 @@ impl Maintaining for SolvableOrdersCache {
...
-        .zip(traded_tokens)
-        .filter_map(|(price, token)| match price {
-            Ok(price) => Some((token, to_normalized_price(price)?)),
-            Err(err) => {
-                tracing::warn!(?token, ?err, "error estimating native token price");
-                None
-            }
-        })
-        .collect::<BTreeMap<_, _>>();
+    let mut prices = BTreeMap::new();
```
nlordell, line 345: Might make sense to have this be a `HashMap` for faster access. We return the `used_prices` which needs to be a `BTreeMap`.
```suggestion
    let mut prices = HashMap::new();
```

nlordell (approved):
Looks great!

Would like to see metrics for:
- How many times we timeout updating auctions
- How many orders we exclude because of missing prices.

This will help us keep an eye and understand just how often our batches are "incomplete".

crates/orderbook/Cargo.toml
vkgnosis: I copied the version from another line. If tokio 1.17 is the newest version then we're probably already using it in the lock file since it is semver compatible.

MartinquaXD (approved):
I think the PR is fine as it is.

Just one idea: Did you consider updating the auction in a streaming manner after the timeout was reached. I think computing a big chunk of the auction like as it's done right now makes sense (to not return tiny auctions when somebody requests one right after we started updating) but theoretically the auction in the cache could get more orders as they come in.

This is probably overkill and too complicated so I don't really think we should go for it right now but maybe in the future this idea might be helpful.

PR #1738, opened 2022-03-21
Base submission error alerts off of metrics

MartinquaXD:
Submitting a transaction can error for all sorts of reasons. Because so many different things can go wrong on-call can receive many alerts which ultimately don't point to a systemic problem but to a bunch of one-off errors.
Instead of always alerting for certain errors I would like to try keeping metrics for every type of submission strategy and alert when one consistently reports errors.

Individual submission success/failure metrics will be maintained for eden, flashbots and each individual custom node the driver is using.
This way it should be mostly obvious from the alert which submission strategy is having an actual problem.
It's mostly obvious because there is no nice way of labeling the individual custom nodes so they will be named based on the index they have in the `--transaction-submission-nodes` argument.

Another notable change is that we are currently try to be smart about errors we receive from the rpc endpoint. This means we try to read the error message and turn the generic error into a custom error. Every error which we didn't anticipate will cause an alert (instead of being logged as a warning). This can lead to too many alerts for benign errors and needs to be kept in sync with the specific nodes rpc client implementation (which error messages does it produce?).
Also we don't really make decisions based on that custom error type. We just log more generic error messages for each known error. Combined with the transition to metrics based errors I don't really see the need to keep this logic and instead use `tracing::warn()` for every (uninterpreted) `web3::Error`. If the metrics cause an alert it's still easy enough to use kibana to search for all the errors related to that submission strategy.

### Test Plan
Used a local grafana and prometheus instance to see that the metrics can be used to generate meaningful graphs.
Also this is how the new metrics show when you query the `/metrics` endpoint:
```
# HELP gp_v2_solver_submission_strategies_submissions Tracks how many transactions get successfully submitted with the different submission strategies.
# TYPE gp_v2_solver_submission_strategies_submissions counter
gp_v2_solver_submission_strategies_submissions{result="error",submitter="custom_nodes_0"} 62
gp_v2_solver_submission_strategies_submissions{result="success",submitter="custom_nodes_0"} 62
```

### DIFF
--- a/crates/solver/src/settlement_submission/submitter/common.rs
+++ b/crates/solver/src/settlement_submission/submitter/common.rs
@@ -1,4 +1,5 @@
 use super::super::submitter::TransactionHandle;
+use anyhow::Result;
 use ethcontract::transaction::{Transaction, TransactionBuilder};
 use futures::FutureExt;
 use shared::{Web3, Web3Transport};
@@ -23,7 +24,7 @@ impl PrivateNetwork {
     pub async fn submit_raw_transaction(
         &self,
         tx: TransactionBuilder<Web3Transport>,
-    ) -> anyhow::Result<TransactionHandle> {
+    ) -> Result<TransactionHandle> {
         let (raw_signed_transaction, tx_hash) = match tx.build().now_or_never().unwrap().unwrap() {
             Transaction::Request(_) => unreachable!("verified offline account was used"),
             Transaction::Raw { bytes, hash } => (bytes.0, hash),
--- a/crates/solver/src/settlement_submission/submitter/common.rs
+++ b/crates/solver/src/settlement_submission/submitter/common.rs
@@ -1,8 +1,5 @@
-use super::super::submitter::{SubmitApiError, TransactionHandle};
-use ethcontract::{
-    jsonrpc::types::error::Error as RpcError,
-    transaction::{Transaction, TransactionBuilder},
-};
+use super::super::submitter::TransactionHandle;
+use ethcontract::transaction::{Transaction, TransactionBuilder};
 use futures::FutureExt;
 use shared::{Web3, Web3Transport};
 use web3::{api::Namespace, types::Bytes};
@@ -26,7 +23,7 @@ impl PrivateNetwork {
     pub async fn submit_raw_transaction(
         &self,
         tx: TransactionBuilder<Web3Transport>,
-    ) -> Result<TransactionHandle, SubmitApiError> {
+    ) -> anyhow::Result<TransactionHandle> {
         let (raw_signed_transaction, tx_hash) = match tx.build().now_or_never().unwrap().unwrap() {
             Transaction::Request(_) => unreachable!("verified offline account was used"),
             Transaction::Raw { bytes, hash } => (bytes.0, hash),
@@ -37,26 +34,8 @@ impl PrivateNetwork {
             .eth()
             .send_raw_transaction(Bytes(raw_signed_transaction))
             .await
-            .map_err(convert_web3_to_submission_error)?;
+            .map_err(anyhow::Error::new)?;
 
         Ok(TransactionHandle { tx_hash, handle })
     }
-}
-
-fn convert_web3_to_submission_error(err: web3::Error) -> SubmitApiError {
-    if let web3::Error::Rpc(RpcError { message, .. }) = &err {
-        if message.starts_with("invalid nonce") || message.starts_with("nonce too low") {
-            return SubmitApiError::InvalidNonce;
-        } else if message.starts_with("Transaction gas price supplied is too low") {
-            return SubmitApiError::OpenEthereumTooCheapToReplace;
-        } else if message.starts_with("replacement transaction underpriced") {
-            return SubmitApiError::ReplacementTransactionUnderpriced;
-        } else if message.contains("tx fee") && message.contains("exceeds the configured cap") {
-            return SubmitApiError::EdenTransactionTooExpensive;
-        }
-    }
-
-    anyhow::Error::new(err)
-        .context("transaction submission error")
-        .into()
 }
--- a/crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
+++ b/crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
@@ -33,7 +33,7 @@ impl TransactionSubmitting for CustomNodesApi {
     async fn submit_transaction(
         &self,
         tx: TransactionBuilder<DynTransport>,
-    ) -> anyhow::Result<TransactionHandle> {
+    ) -> Result<TransactionHandle> {
         let transaction_request = tx.build().now_or_never().unwrap().unwrap();
         let mut futures = self
             .nodes
@@ -75,7 +75,7 @@ impl TransactionSubmitting for CustomNodesApi {
         }
     }
 
-    async fn cancel_transaction(&self, id: &CancelHandle) -> anyhow::Result<TransactionHandle> {
+    async fn cancel_transaction(&self, id: &CancelHandle) -> Result<TransactionHandle> {
         self.submit_transaction(id.noop_transaction.clone()).await
     }
 
--- a/crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
+++ b/crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
@@ -52,9 +52,11 @@ impl TransactionSubmitting for CustomNodesApi {
             .collect::<Vec<_>>();
 
         loop {
-            let (result, _index, rest) = futures::future::select_all(futures).await;
+            let (result, index, rest) = futures::future::select_all(futures).await;
+            let lable = "custom_nodes_".to_string() + &index.to_string();
             match result {
                 Ok(tx_hash) => {
+                    super::track_submission_success(lable.as_str(), true);
                     tracing::info!("created transaction with hash: {:?}", tx_hash);
                     return Ok(TransactionHandle {
                         tx_hash,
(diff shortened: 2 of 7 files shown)


crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
```diff
@@ -52,33 +52,30 @@ impl TransactionSubmitting for CustomNodesApi {
             .collect::<Vec<_>>();
 
         loop {
-            let (result, _index, rest) = futures::future::select_all(futures).await;
+            let (result, index, rest) = futures::future::select_all(futures).await;
+            let lable = "custom_nodes_".to_string() + &index.to_string();
```
vkgnosis, line 56: Feels more natural to me
```suggestion
            format!("custom_nodes_{}", index)
```

vkgnosis (approved):
Metric code looks good.
I like the categorization of errors because it shows us what errors can show up in case we want to react to them in the future. It feels like a better api. On the other hand if we're not using them now and it is work to keep track of them maybe it's not worth it.

crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
nlordell: I would even suggest edition 2021 magic:
```rust
            format!("custom_nodes_{index}")
```

crates/solver/src/settlement_submission/submitter/common.rs
```diff
@@ -26,7 +23,7 @@ impl PrivateNetwork {
     pub async fn submit_raw_transaction(
         &self,
         tx: TransactionBuilder<Web3Transport>,
-    ) -> Result<TransactionHandle, SubmitApiError> {
+    ) -> anyhow::Result<TransactionHandle> {
```
nlordell, line 26: ubernit: This file already `use anyhow::Result` so you should be able to:
```suggestion
    ) -> Result<TransactionHandle> {
```

crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
```diff
@@ -52,33 +52,30 @@ impl TransactionSubmitting for CustomNodesApi {
             .collect::<Vec<_>>();
 
         loop {
-            let (result, _index, rest) = futures::future::select_all(futures).await;
+            let (result, index, rest) = futures::future::select_all(futures).await;
+            let lable = format!("custom_nodes_{}", index);
```
nlordell, line 56: ubernit: Edition 2021 niceness!
```suggestion
            let lable = format!("custom_nodes_{index}");
```

crates/solver/src/settlement_submission/submitter/eden_api.rs
```diff
@@ -97,13 +98,21 @@ impl TransactionSubmitting for EdenApi {
...
+            .await;
+
+        let successful = match &result {
+            Ok(_) => true,
+            // Sometimes `submit_slot_transaction()` times out and the fallback submission
+            // strategy reveals that the network is already aware of the transaction by returning
+            // the error message "already known".
+            Err(err) => err.to_string().contains("already known"),
+        };
+        super::track_submission_success("eden", successful);
```
nlordell, line 110: Instead of tracking submission in each submitter implementation, should we instead introduce an `InstrumentedSubmitter` instead? My concern is that if we add a new submitter we may forget to instrument it.

I'm on the fence about this suggestion - just wanted to bring it up.

crates/solver/src/settlement_submission/submitter/common.rs
MartinquaXD: Tried it; didn't work.
I changed the API of the  trait from `-> Result` to `-> anyhow::Result` so at no place implementing the trait should the existing code be okay since the Results are different types.
Or am I missing something?

crates/solver/src/settlement_submission/submitter/eden_api.rs
nlordell: > Eden is the best example here: AFAIK it's the only submitter where we can say a submission was successful even if we are not able to produce a TransactionHandle.

Right, so that would require even more work to change the interface (to bubble up the fact of whether or not an error is a "hard" error or a "soft" error).

This also feels icky because you are modifying the interface in order to introduce a instrumenting component further up the call-chain.

With that in mind, I think it makes sense in this case to instrument each submitter individually then.

nlordell (approved):
Approving assuming a test plan will be added/run.

I also liked the error classification that we had before. One possible use for it would be to implement some `InstrumentedSubmitter` wrapper component that would use the differently classified errors to determine how to update metrics. On the other hand, this does cause submitter-specific implementation details to leak into the `Submitting` trait and in the shared `InstrumentedSubmitter` which also feels a bit icky to me. With that in mind, I am OK with getting rid of the error classification.

crates/solver/src/settlement_submission/submitter/common.rs
nlordell: My point is that `Result` should already have been `anyhow::Result` because we `use anyhow::Result` in the file (or at least I thought we did - but maybe I saw wrong).

Anyway, just an ubernit - so not blocking merge from my perspective.

MartinquaXD: But if `Result` was already `anyhow::Result` the trait wouldn't have been correctly implemented before because the old trait expected `std::result::Result` instead of `anyhow::Result`. So either the signature was wrong before it's wrong now, no?

vkgnosis: This is confusing because anyhow::Result is defined like `type anyhow::Result<T> = std::result::Result<T, anyhow::Error>` so it is not a normal type. Writing `anyhow::Result<TransactionHandle>` and `std::result::Result<TransactionHandle, anyhow::Error>`(or leaving out the std::result part) is equivalent.

MartinquaXD: Ahh, I see. Thanks for clearing that up. 👍

crates/solver/src/settlement_submission/submitter/eden_api.rs
sunce86: I'm more in favor of removing the `submit_slot_transaction` call since we are not going to bid for Eden slot anymore. It's unreliable and forcing us to have these kind of weird cases to handle.

crates/solver/src/settlement_submission/submitter/common.rs
sunce86: There are also a few more places to replace. Personally, I don't favor any of the versions `anyhow::Result<TransactionHandle>` vs `Result<TransactionHandle>`, but I do favor using only one of them through out the code for consistency.

crates/solver/src/settlement_submission/submitter/eden_api.rs
nlordell: Good point. That probably makes sense in a separate PR though (since its unrelated to instrumentation right?).

sunce86 (approved):

crates/solver/src/settlement_submission/submitter/flashbots_api.rs
```diff
@@ -34,18 +34,20 @@ impl TransactionSubmitting for FlashbotsApi {
...
-        self.rpc
+    ) -> anyhow::Result<TransactionHandle> {
+        let result = self
+            .rpc
             .api::<PrivateNetwork>()
             .submit_raw_transaction(tx)
-            .await
+            .await;
+
+        super::track_submission_success("flashbots", result.is_ok());
```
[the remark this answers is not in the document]
sunce86: Seems like a copy of the @nlordell comment above.

crates/solver/src/settlement_submission/submitter/eden_api.rs
MartinquaXD: So if we consider dropping `submit_slot_transaction()` the `InstrumentedSubmitter` idea seems way nicer but I encountered one problem while implementing it:

At the moment we have a single instance of `CustomNodesApi` which handles any number of nodes. To have metrics for each individual node we would have to have a `CustomNodesApi` instance per node. This is because the `CustomNodesApi` API only gives us the information that either all nodes failed or some node succeeded.

This can be solved by either leaving the code mostly as is and have n `CustomNodesApi` instances each having handling a `Vec` of nodes with size 1. Or we could rewrite `CustomNodesApi` to always handle a single node.

IMO the cleaner approach would be the latter but in the interest of time and small PRs I can do the former and change the `CustomNodesApi` afterwards.

sunce86: Note that if `CustomNodesApi` contains only one node, then we will have 5 transaction strategies (Eden + Flashbots + 3nodes) instead of 3 (Eden + Flashbots + CustomNodes), meaning our submission component will now run this function additional two times for two additional nodes without actual need (all 3 nodes have exact code execution while Eden and Flashbots have some specifics and that was the reason to parallelize the execution at the level of `Submitter` and not at the level of `Submitter::TransactionSubmitting`: https://github.com/gnosis/gp-v2-services/blob/c2abb1955aa7909258264b92dbeb4ceb7d623621/crates/solver/src/settlement_submission/submitter.rs#L211

MartinquaXD: That's unfortunate. :(
The only alternative I see would be to create another type implementing `TransactionSubmitting` (let's say `CustomNode`) which takes only a single node (`Web3`) and make it that `CustomNodesApi` operates on a `Vec<Box<dyn TransactionSubmitting>>` instead of `Vec<Web3>`. That way we could wrap `CustomNode` in an `InstrumentedNode` and pass a vector of those to `CustomNodesApi`.
`CustomNodesApi` would then be similar to the `RacingPriceEstimator` but a little bit simpler.

vkgnosis: To me the instrumenting part is secondary and it feels like we're having to make the primary logic more complicated to make the instrumenting nicer which I dislike. The concern about forgetting to add instrumenting to a new submitter doesn't outweight these disadvantages.
(But I haven't looked at this code in detail.)

PR #1743, opened 2022-03-22
Load uniswapv3 fee values at runtime

### LINKED ISSUE #1568: Fetch uniswapv3 fee values on start up
Can we fetch and cache this instead of hardcoding it?

_Originally posted by @nlordell in https://github.com/gnosis/gp-v2-services/pull/1567#discussion_r784886607_

vkgnosis:
Fixes #1568

### Test Plan

CI and ran ignored test manually to see fee values printed

### DIFF
--- a/crates/shared/src/bad_token/trace_call.rs
+++ b/crates/shared/src/bad_token/trace_call.rs
@@ -1,13 +1,14 @@
 use super::{BadTokenDetecting, TokenQuality};
 use crate::{
     ethcontract_error::EthcontractErrorType,
+    event_handling::MAX_REORG_BLOCK_COUNT,
     sources::{uniswap_v2::pair_provider::PairProvider, uniswap_v3_pair_provider},
     trace_many, Web3,
 };
 use anyhow::{anyhow, bail, ensure, Context, Result};
 use contracts::{IUniswapV3Factory, ERC20};
 use ethcontract::{
-    batch::CallBatch, dyns::DynTransport, transaction::TransactionBuilder, PrivateKey,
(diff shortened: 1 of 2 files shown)


crates/shared/src/bad_token/trace_call.rs
```diff
@@ -55,13 +56,43 @@ impl TokenOwnerFinding for BalancerVaultFinder {
...
-    const FEE: [u32; 3] = [500, 3000, 10000];
+    async fn fee_values(factory: &IUniswapV3Factory, current_block: u64) -> Result<Vec<u32>> {
+        // We expect there to be few of these kind of events (currently there are 4) so fetching all
+        // of them is fine. Alternatively we could index these events in the database.
+        let events = factory
+            .events()
+            .fee_amount_enabled()
+            .from_block(BlockNumber::Earliest)
+            .to_block(BlockNumber::Number(
+                current_block.saturating_sub(MAX_REORG_BLOCK_COUNT).into(),
```
sunce86, line 89: What bad could happen if we don't substract MAX_REORG_BLOCK_COUNT?

sunce86 (approved):

crates/shared/src/bad_token/trace_call.rs
vkgnosis: Theoretically they could have just added a new fee constant which we load and then the block gets re-orged so that fee constant disappears. This is unlikely and the only impact is that we would waste some calls to the node. Still feels slightly nicer to subtract the constant.

MartinquaXD (approved):
Appears to be compatible with the uniswap v3 contract.

nlordell (approved):
LGTM.

PR #1744, opened 2022-03-23
Refactor Allowance Manager Interface

nlordell:
This PR modifies the `AllowanceManaging` interface to allow for querying batched approval interactions.

The `AllowanceManaging` interface is responsible for checking the current approvals of the settlement contract to various token transfer proxy contracts for on-chain DEXs (for example the Uniswap router, the ZeroEx "approval target" contract) and creating `Approval` interactions that get encoded into the settlement. If the settlement contract has a sufficiently high allowance set for a spender, then it returns `Approval::SufficientAllowance`, which encodes to no interactions in the final settlement, or a `Approval::Approve { token, spender }` which encodes to a single `token.approve(spender, type(uint256).max)` interaction in the final settlement. This is a gas optimization to avoid setting allowances for every AMM interaction instead of only once per spender per token.

Previously, the allowance manager was used in two contexts:
- Pre-fetching allowances for liquidity. For example, the Uniswap liquidity fetcher would fetch all settlement allowances to the Uniswap router for all pool tokens.
- Fetching a single `Approval` in single order solvers.

However, now that our HTTP solvers are going to be fetching their own liquidity - it needs a way to set multiple approvals for different tokens and different spenders. One approach would have been to make the HTTP solver responsible for fetching the current settlement contract allowances and implementing its own logic for deciding when to add an approval or not. However, in order to reduce the barrier to entry for solvers, it makes sense for the driver to accept a list of "approvals" to encode for the solver, allowing us to reuse the existing logic for fetching settlement contract allowances and encoding approvals.

Support for HTTP solver solutions specifying required approvals will be added in a follow up.

### Test Plan

Existing CI.


vkgnosis (approved):

sunce86 (approved):

MartinquaXD (approved):

crates/solver/src/interactions/allowances.rs
```diff
@@ -135,45 +156,66 @@ impl AllowanceManager {
...
+            .iter()
+            .map(|request| {
+                allowances
+                    .get(&request.spender)
+                    .with_context(|| {
+                        format!("no allowances found for spender {}", request.spender)
+                    })?
+                    .approve_token(request.token, request.amount)
+            })
+            .collect()
```
MartinquaXD, line 189: I was very confused here. I didn't know [this](https://doc.rust-lang.org/std/result/enum.Result.html#impl-FromIterator%3CResult%3CA%2C%20E%3E%3E) was a thing.

PR #1747, opened 2022-03-23
Use cow token and virtual token for subsidy calculation

### LINKED ISSUE #1647: Extra subsidy for vcow holders

### LINKED ISSUE #1685: Consider Multiple Token Balances for Solver Subsidies
We should probably consider multiple token balances for additional trading subsidies:
- The _swappable_ vCOW balance (rather than the total balance)
- The total COW balance

vkgnosis:
Fixes #1685 Fixes #1647 

This PR bases the subsidy on the sum of the Cow balance and vCow swappable balance. I would appreciate someone independently thinking about whether this makes sense.

### Test Plan

CI, ran the updated ignored test

### DIFF
--- a/crates/orderbook/src/cow_subsidy.rs
+++ b/crates/orderbook/src/cow_subsidy.rs
@@ -2,7 +2,7 @@ use std::{sync::Mutex, time::Duration};
 
 use anyhow::{Context, Result};
 use cached::{Cached, TimedSizedCache};
-use contracts::ERC20;
+use contracts::{CowProtocolToken, CowProtocolVirtualToken};
 use primitive_types::{H160, U256};
 use std::collections::BTreeMap;
 
@@ -71,7 +71,8 @@ impl std::str::FromStr for SubsidyTiers {
 }
 
 pub struct CowSubsidyImpl {
-    cow_token: ERC20,
+    token: CowProtocolToken,
+    vtoken: CowProtocolVirtualToken,
     subsidy_tiers: SubsidyTiers,
     cache: Mutex<TimedSizedCache<H160, f64>>,
 }
@@ -89,7 +90,11 @@ impl CowSubsidy for CowSubsidyImpl {
 }
 
 impl CowSubsidyImpl {
-    pub fn new(cow_token: ERC20, subsidy_tiers: SubsidyTiers) -> Self {
+    pub fn new(
+        token: CowProtocolToken,
+        vtoken: CowProtocolVirtualToken,
+        subsidy_tiers: SubsidyTiers,
+    ) -> Self {
         // NOTE: A long caching time might bite us should we ever start advertising that people can
         // buy COW to reduce their fees. `CACHE_LIFESPAN` would have to pass after buying COW to
         // qualify for the subsidy.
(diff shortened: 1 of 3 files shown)


crates/orderbook/src/cow_subsidy.rs
```diff
@@ -100,16 +105,21 @@ impl CowSubsidyImpl {
...
             cache: Mutex::new(cache),
         }
     }
 
     async fn subsidy_factor_uncached(&self, user: H160) -> Result<f64> {
-        let balance = self.cow_token.balance_of(user).call().await?;
-        let tier = self.subsidy_tiers.0.range(..=balance).rev().next();
-        Ok(tier.map(|tier| *tier.1).unwrap_or(1.0))
+        let balance = self.token.balance_of(user).call().await?;
+        let vbalance = self.vtoken.swappable_balance_of(user).call().await?;
```
MartinquaXD, line 117: I think those balances could get fetched in parallel.

MartinquaXD (approved):
Code looks good. As for
> ... independently thinking about whether this makes sense.

I don't see anything wrong with your approach to calculate the total balance. Using cow balance + the swappable part of vcow should prevent counting balances which are still vesting.
The only noteworthy thing I can think of is that using a timed cache with a long timeout can be bad for people buying COW specifically for the subsidy. But that is nothing which needs to be addressed in this PR IMO.
Do you have some specific concerns in mind?

crates/orderbook/src/cow_subsidy.rs
nlordell: Maybe try out the `BufferedTransport` so that you can `join` the futures and get batched calls for free? 😄.

crates/orderbook/src/main.rs
```diff
@@ -591,12 +590,32 @@ async fn main() {
...
+    let cow_vtoken = match CowProtocolVirtualToken::deployed(&web3).await {
+        Err(DeployError::NotFound(_)) => None,
+        other => Some(other.unwrap()),
+    };
+    let cow_tokens = match (cow_token, cow_vtoken) {
+        (None, None) => None,
+        (Some(token), Some(vtoken)) => Some((token, vtoken)),
+        _ => panic!("should either have both cow token contracts or none"),
+    };
+    let cow_subsidy = match cow_tokens {
```
nlordell, line 606: Personally, I would have just used
```suggestion
    let cow_subsidy = match cow_token.zip(cow_vtoken) {
```

And not handled the case where only one of the tokens exist (its not like we can ever hit this codepath because of how we do contract deployments).

On the other hand - its already in there and this feels more "correct".

crates/orderbook/src/cow_subsidy.rs
nlordell: Hmm, I think this is wrong. I think that the total subsidy should be:
```suggestion
        let balance = self.token.balance_of(user).call().await?;
        let vbalance = self.vtoken.balance_of(user).call().await?;
```

Let me double check.

nlordell: Right, it is the combined balance and not just the swappable vCOW balance. From the CIP:

> Combined balance of vCOW and COW is accounted for per network 

https://forum.cow.fi/t/cip-3-trading-fee-discount-for-cow-holders/325

vkgnosis: So swappable balance of is correct or balance of? The docs https://github.com/gnosis/cow-token/blob/main/src/contracts/CowProtocolVirtualToken.sol say that balance_of would include future vested amounts which seemed wrong.

MartinquaXD: I personally read the quoted section such that it's supposed to highlight the fact that only the balance of the current chain is considered while being a little vague on the exact calculation of the "correct" balance.
IMO opinion not considering the unvested amounts makes more sense, but depending on how you read the CIP it would be wrong.

nlordell: > IMO opinion not considering the unvested amounts makes more sense

Interesting. I would argue the opposite. Unvested vCOW holders are locked in for 4 years and are way more long on COW than regular holders - so to me it makes sense for them to also get benefits.

I asked on Slack to confirm.

bh2smith: I was also expecting it to be the cow balance plus the still vested balance. Not just the cow plus swappable vCOW balance.

nlordell: It was confirmed by Anna on Slack that we want the full vCOW token balance and not just the vested (a.k.a. swappable) balance for the purpose of computing subsidies.

nlordell: @vkgnosis - sorry but I high jacked this comment and I think the original suggestion by @MartinquaXD was missed:

> I think those balances could get fetched in parallel.

PR #1748, opened 2022-03-23
Flag to Enable Buying Ether with Smart Contracts

nlordell:
⚠️ Currently this PR is **on hold** and should not be merged because of issues with access list generation.

This PR adds a new flag to the orderbook API to enable orders buying Ether where the receiver is a smart contract.

Because of recent gas fee schedule changes, transferring Ether to smart contracts from the Settlement contract as part of trade doesn't work without access lists. Specifically, the transfer is encoded as a `CALL` opcode with a very small gas limit. Gnosis Safes, for example, are implemented as proxy contracts that first read the "singleton" address at storage slot 0 and then do a `DELEGATECALL` to that address. The Ether transfer triggers these two things, however, the small gas stipend is not enough to actually execute them. With access lists, which would "warm up" the storage and addresses, would make it so the gas stipend would be enough.

Now that we have access list generation support, I decided to try and it out and see if we can finally buy Ether with smart contract receivers! 

### Test Plan

Added unit tests to verify new logic. Also ran a local test on Rinkeby. Unfortunately it doesn't work because the access list being generated doesn't warm up enough addresses and slots for the transfer to succeed. Specifically the Safe singleton copy address is not getting warmed up and leads to reverts like:

https://dashboard.tenderly.co/gp-v2/staging/simulator/4590f4e6-71fa-47eb-bff4-0a4f51251e77/public

I will try with Erigon access list estimation, and have contacted the Tenderly team to see if there is anything that can be done about it.

### DIFF
--- a/crates/e2e/tests/e2e/services.rs
+++ b/crates/e2e/tests/e2e/services.rs
@@ -179,6 +179,7 @@ impl OrderbookServices {
             HashSet::default(),
             HashSet::default(),
             Duration::from_secs(120),
+            false,
             fee_calculator.clone(),
             bad_token_detector.clone(),
             balance_fetcher,
--- a/crates/orderbook/src/api/order_validation.rs
+++ b/crates/orderbook/src/api/order_validation.rs
@@ -216,6 +216,7 @@ pub struct OrderValidator {
     banned_users: HashSet<H160>,
     liquidity_order_owners: HashSet<H160>,
     min_order_validity_period: Duration,
+    enable_smart_contract_buy_native_token: bool,
     /// For Full-Validation: performed time of order placement
     fee_validator: Arc<dyn MinFeeCalculating>,
     bad_token_detector: Arc<dyn BadTokenDetecting>,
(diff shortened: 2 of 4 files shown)


vkgnosis (approved):
(I see it is on hold.)
As an alternative to just support safes we could have a flag like "this is a safe so please add address and storage slot 0 to access list".

nlordell:
> "this is a safe so please add address and storage slot 0 to access list".

Yeah, it just gets complicated because it will be SC dependent (Safe may be different than an Argent wallet for example). This is why having access list generation was important - so something can tell us what address/slots we need instead of having some hard-coded values.

nlordell:
Another more reliable solution is to transfer add 0-calldata and 0-eth interaction for smart-contract receivers.

The downside is that these interactions don't have gas limits and can be used to, for example, mint gas tokens.

sunce86:
For this to work, we need to make access list mandatory for sending transactions, right? Otherwise, if we allow smart contracts to receive ETH and don't have access list enabled / access list generation fails, then our simulation will be constantly failing for these settlements.

nlordell:
> Otherwise, if we allow smart contracts to receive ETH and don't have access list enabled / access list generation fails, then our simulation will be constantly failing for these settlements.

Right - this is why I decided to put it behind a flag for now. Its a tricky problem because if we solve a settlement with order X under the assumption we can generate an access list and then fail, then the solution with order X is invalid, and we can't generate access lists without already having a solution with order X. 🐔 🥚

PR #1749, opened 2022-03-23
Use u64 to enter subsidy tier threshold to avoid rounding errors

MartinquaXD:
Since people are voting on very specific tier thresholds we should avoid rounding errors if possible.
Luckily this can be done easily in this case by parsing the cow based tier subsidy threshold as an `u64` which allows a lossless conversion to `U256`.

Note:
Also updated the help text to reflect the contents of what is currently being voted on: [CIP-3](https://forum.cow.fi/t/cip-3-trading-fee-discount-for-cow-holders/325)

### Test Plan
CI

### DIFF
--- a/crates/orderbook/src/main.rs
+++ b/crates/orderbook/src/main.rs
@@ -177,10 +177,11 @@ struct Arguments {
     )]
     partner_additional_fee_factors: HashMap<AppId, f64>,
 
-    /// Used to configure how much of the regular fee a user should pay based on their COW balance.
+    /// Used to configure how much of the regular fee a user should pay based on their
+    /// COW + VCOW balance of the current network.
     ///
-    /// The expected format is "10.2:0.75,150:0.5" for 2 subsidy tiers.
-    /// A balance of [10.2,150) COW will cause you to pay 75% of the regular fee and a balance of
+    /// The expected format is "10:0.75,150:0.5" for 2 subsidy tiers.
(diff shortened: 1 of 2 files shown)


crates/orderbook/src/main.rs
```diff
@@ -177,10 +177,11 @@ struct Arguments {
     )]
     partner_additional_fee_factors: HashMap<AppId, f64>,
 
-    /// Used to configure how much of the regular fee a user should pay based on their COW balance.
+    /// Used to configure how much of the regular fee a user should pay based on their
+    /// COW + VCOW balance of the current network.
```
nlordell, line 181: nit: Maybe specify that this amount is not in atoms but in base units.

nlordell (approved):
LGTM. 

One downside is that if we setup a tier for 0.5 COW that we can't represent it with `u64`.

Another a approach would be to use atoms instead (so 1 COW would be specified as `1000000000000000000`), but that is less convenient in terms of configuration than what you propose.

MartinquaXD:
I think `u64` offers the best balance of input range and verbosity. Since a single base unit of COW will probably never be very expensive needing 0.5 COW more or less to qualify for a fee doesn't make a difference for the end user and simplifies things for us.

vkgnosis (approved):

nlordell:
> I think `u64` offers the best balance of input range and verbosity.

I also agree. Additionally, changing this to take `U256` is fairly small amount of work that we can do if the need arises. I think the current implementation is a good compromise between precision and ease of configuration.

PR #1752, opened 2022-03-24
No mature liquidity order

josojo:
Addresses the discussion point from here:
https://github.com/gnosis/gp-v2-services/pull/1707#discussion_r820944092

For users, it should be preferable if they are waiting for a real cow, and not just for a cow with market makers.
Hence, mature liquidity orders should not contribute to the "maturity level" of a settlement.

### Test Plan

unit tests.


vkgnosis (approved):

MartinquaXD (approved):
LGTM

sunce86 (approved):
Lg. Might be useful to update the function `retain_mature_settlements` description with this new info.

nlordell (approved):
LGTM

PR #1753, opened 2022-03-24
Change auction metrics

vkgnosis:
I was looking at the auction metrics and felt that we should have a solvable order count metric too and that this and the filtered orders metric make more sense as gauges. The solvable orders we technically get in the driver too but it is nicer to create it at the source which is the orderbook.

### Test Plan

CI

### DIFF
--- a/crates/orderbook/src/metrics.rs
+++ b/crates/orderbook/src/metrics.rs
@@ -1,7 +1,8 @@
 use anyhow::Result;
 use gas_estimation::EstimatedGasPrice;
 use prometheus::{
-    Gauge, Histogram, HistogramOpts, HistogramVec, IntCounter, IntCounterVec, IntGaugeVec, Opts,
+    Gauge, Histogram, HistogramOpts, HistogramVec, IntCounter, IntCounterVec, IntGauge,
+    IntGaugeVec, Opts,
 };
 use shared::{
     metrics::get_metrics_registry,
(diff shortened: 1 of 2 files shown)


sunce86 (approved):

MartinquaXD (commented):
Code looks good and I like collecting more data but I'm not sure using a Gauge is the way to go here.
Isn't an `IntCounter` in this scenario strictly more powerful than a `Gauge`?
I think graphing how many solvable orders are in a batch over time would be useful to quickly see a timespan where suspicious activity happened and debug it after the fact.
I believe a `Gauge` wouldn't allow for that, right?

vkgnosis:
The gauge measures the current number of solvable orders while the counter measures the total number of solvable orders over all auctions. If the prometheus sampling interval is smaller than the auction creation interval then either metric can be used to get the current number (for the gauge the gauge itself, for the counter `current - prev`).
- This stops working when the prometheus sampling interval is larger than the auction creation interval.
- A counter allows you to calculate the rate of the measured events happening but in our case this rate has no meaning. We already measure the rate of auction creation and the rate of solvable orders depends on that.

For these reasons a gauge makes more sense to me.

MartinquaXD (approved):
Thanks for the explanation.
I didn't think about the prometheus sampling interval being a problem.

nlordell (approved):

PR #1754, opened 2022-03-24
Access list logging and metrics

sunce86:
This PR adds metrics for gas saved by using access lists for settlement submission.

Additional refactoring is done to enable this:
1. Extracted basic tenderly simulation and moved to `settlement_simulation.rs`
2. Basic tenderly simulation made generic over response type, since the native response is huge, therefore different components using basic tenderly simulation should define their own struct for deserialization of response.
3. Refactored `TenderlyAccessList` to use basic tenderly simulation.

### DIFF
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -14,7 +14,7 @@ use crate::{
     settlement_submission::SolutionSubmitter,
     solver::{Auction, SettlementWithError, SettlementWithSolver, Solver, Solvers},
 };
-use anyhow::{ensure, Context, Result};
+use anyhow::{Context, Result};
 use contracts::GPv2Settlement;
 use futures::future::join_all;
 use gas_estimation::{EstimatedGasPrice, GasPriceEstimating};
@@ -220,12 +220,13 @@ impl Driver {
     }
 
     async fn metric_access_list_gas_saved(&self, transaction_hash: H256) -> Result<()> {
-        ensure!(self.tenderly.is_some(), "tenderly disabled");
-        let web3 = &self.web3;
-        let tenderly = self.tenderly.as_ref().unwrap();
-        let network_id = self.network_id.clone();
-        let gas_saved =
-            simulate_before_after_access_list(web3, tenderly, network_id, transaction_hash).await?;
+        let gas_saved = simulate_before_after_access_list(
+            &self.web3,
+            self.tenderly.as_ref().context("tenderly disabled")?,
+            self.network_id.clone(),
+            transaction_hash,
+        )
+        .await?;
         self.metrics.settlement_access_list_saved_gas(gas_saved);
         Ok(())
     }
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -10,17 +10,17 @@ use crate::{
     orderbook::OrderBookApi,
     settlement::{external_prices::ExternalPrices, Settlement},
     settlement_post_processing::PostProcessingPipeline,
-    settlement_simulation::{self, settle_method},
+    settlement_simulation::{self, settle_method, simulate_before_after_access_list, TenderlyApi},
     settlement_submission::SolutionSubmitter,
     solver::{Auction, SettlementWithError, SettlementWithSolver, Solver, Solvers},
 };
-use anyhow::{Context, Result};
+use anyhow::{ensure, Context, Result};
 use contracts::GPv2Settlement;
 use futures::future::join_all;
 use gas_estimation::{EstimatedGasPrice, GasPriceEstimating};
 use itertools::{Either, Itertools};
 use num::{BigRational, ToPrimitive};
-use primitive_types::H160;
+use primitive_types::{H160, H256};
 use rand::prelude::SliceRandom;
 use shared::{
     current_block::{self, CurrentBlockStream},
@@ -58,6 +58,7 @@ pub struct Driver {
     post_processing_pipeline: PostProcessingPipeline,
     simulation_gas_limit: u128,
     fee_objective_scaling_factor: BigRational,
+    tenderly: Option<TenderlyApi>,
 }
 impl Driver {
     #[allow(clippy::too_many_arguments)]
@@ -83,6 +84,7 @@ impl Driver {
         weth_unwrap_factor: f64,
         simulation_gas_limit: u128,
         fee_objective_scaling_factor: f64,
+        tenderly: Option<TenderlyApi>,
     ) -> Self {
         let post_processing_pipeline = PostProcessingPipeline::new(
             native_token,
@@ -116,6 +118,7 @@ impl Driver {
             simulation_gas_limit,
             fee_objective_scaling_factor: BigRational::from_float(fee_objective_scaling_factor)
                 .unwrap(),
+            tenderly,
         }
     }
 
@@ -188,6 +191,12 @@ impl Driver {
                     crate::metrics::SettlementSubmissionOutcome::Success,
                     name,
                 );
+                if let Err(err) = self
+                    .metric_access_list_gas_saved(receipt.transaction_hash)
+                    .await
+                {
+                    tracing::debug!("access list metric not saved: {}", err);
+                }
                 Ok(receipt)
             }
             Err(err) => {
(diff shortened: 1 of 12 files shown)


crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -474,16 +485,23 @@ impl<'a> Submitter<'a> {
             tx.clone().access_list(access_list.clone()).estimate_gas()
         )?;
 
+        ensure!(
+            gas_before_access_list > gas_after_access_list,
+            "access list exist but does not lower the gas usage"
+        );
+        ensure!(gas_before_access_list > U256::zero(), "bad gas estimation");
```
MartinquaXD, line 492: This error will never happen because the previous line already ensures that `gas_before_access_list > 0`.

sunce86: Indeed. Nice.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -474,16 +485,22 @@ impl<'a> Submitter<'a> {
...
             gas_before_access_list,
             gas_after_access_list,
             access_list,
+            gas_percent_saved
         );
-        ensure!(
-            gas_before_access_list > gas_after_access_list,
-            "access list exist but does not lower the gas usage"
-        );
+        track_access_list_gas_saved(gas_percent_saved);
```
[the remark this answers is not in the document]
sunce86: In general yes, but I was aiming to measure/compare only the success of access list estimators (now we are on tenderly but will switch to node).

In case we want to know the real onchain effect of access lists, I guess I would need to do that via tenderly simulation (because nodes do not support simulation with `transaction_index`).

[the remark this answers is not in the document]
nlordell: OK - makes sense that we will always approximate. But still shouldn't we only count this if we see a mined transaction? My reasoning is that, for example, we may count the gas saved from a Flashbots transaction timeout, even though its 0 - and that will get us misrepresented (higher) savings metric than we actually get.

I agree that the complexity of simulating at transaction index to get the exact amount is not necessarily worth it though.

crates/solver/src/settlement_submission/submitter.rs
```diff
@@ -474,16 +485,22 @@ impl<'a> Submitter<'a> {
...
         Ok(access_list)
```
nlordell, line 504: nit: I think we should be measuring the gas saved in absolute units and not in percentage.

vkgnosis: Currently the metric is a histogram. We should probably adjust the buckets either way for the value range we expect. Or this might be a metric where direct logging is more appropriate than prometheus.

crates/solver/src/settlement_submission/submitter.rs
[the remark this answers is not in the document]
sunce86: Added metric for the mined transactions only. One concern I have now is that our metric (now based on tenderly) will be skipped if tenderly sync is behind our node sync. Let's see.

crates/solver/src/settlement_submission/submitter.rs
nlordell: Ah ok... In that case it makes sense to use a histogram for the % savings, but it might also be interesting in recording the total gas savings as a counter to see how much gas we have saved over time.

crates/solver/src/settlement_simulation.rs
```diff
@@ -171,6 +245,72 @@ pub fn tenderly_link(
     )
 }
 
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
```
sunce86, line 248: This is old code just moved.

crates/solver/src/settlement_submission/submitter.rs
vkgnosis: Instead of a histogram we could have "transaction count" and "gas saved count". Don't think we need to have individual buckets that much. Or you can implement this as a histogram with only one bucket (+inf).

crates/solver/src/main.rs
```diff
@@ -616,6 +618,15 @@ async fn main() {
         liquidity_order_owners: args.shared.liquidity_order_owners.into_iter().collect(),
         fee_objective_scaling_factor: args.fee_objective_scaling_factor,
     };
+    let tenderly = {
+        if let Some(tenderly_url) = args.tenderly_url {
+            if let Some(tenderly_api_key) = args.tenderly_api_key.as_ref() {
+                TenderlyApi::new(tenderly_url, client.clone(), tenderly_api_key).ok();
+            }
+        }
```
MartinquaXD, line 626: nit: This rightward drift is a little unfortunate. Looks like `tenderly_url` and `tenderly_api_key` only make sense together. Maybe they can be passed around as a single struct?
I missed that on the previous PR. 🙈

crates/solver/src/driver.rs
```diff
@@ -200,11 +209,27 @@ impl Driver {
...
+                }
                 Err(err.into_anyhow())
             }
         }
     }
 
+    async fn metric_access_list_gas_saved(&self, transaction_hash: H256) -> Result<()> {
+        ensure!(self.tenderly.is_some(), "tenderly disabled");
+        let web3 = &self.web3;
+        let tenderly = self.tenderly.as_ref().unwrap();
```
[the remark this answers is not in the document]
sunce86: Indeed. Fixed. Thanks.

crates/solver/src/driver.rs
```diff
@@ -200,11 +209,27 @@ impl Driver {
...
+                    }
+                }
                 Err(err.into_anyhow())
             }
         }
     }
 
+    async fn metric_access_list_gas_saved(&self, transaction_hash: H256) -> Result<()> {
+        ensure!(self.tenderly.is_some(), "tenderly disabled");
+        let web3 = &self.web3;
```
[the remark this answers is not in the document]
sunce86: Sure, with `ensure` removed I also think this is fine.

crates/solver/src/main.rs
nlordell: Can also be refactored as:
```rust
let tenderly = args.tenderly_url.zip(args.tenderly_api_key).map(|(url, api_key)| {
    TenderlyApi::new(url, client.clone(), api_key).ok();
});
```

sunce86: Took @nlordell suggestion, tnx both!

crates/solver/src/driver.rs
```diff
@@ -200,11 +209,28 @@ impl Driver {
                 );
                 self.metrics
                     .settlement_submitted(err.as_outcome(), solver.name());
+                if let Some(transaction_hash) = err.transaction_hash() {
+                    if let Err(err) = self.metric_access_list_gas_saved(transaction_hash).await {
```
nlordell, line 213: Ah nice - so this will show us the "downside" of access lists if we revert early in a transaction and the access list causes us to pay for warm ups we don't actually use.

crates/solver/src/settlement_simulation.rs
```diff
@@ -119,6 +124,75 @@ pub async fn simulate_and_error_with_tenderly_link(
...
+    web3: &Web3,
+    tenderly: &TenderlyApi,
+    network_id: String,
+    transaction_hash: H256,
+) -> Result<f64> {
+    let transaction = web3
+        .eth()
+        .transaction(transaction_hash.into())
+        .await?
+        .ok_or_else(|| anyhow!("no transaction found"))?;
```
nlordell, line 147: ubernit: Some `anyhow` niceties are slightly more ergonomic.
```suggestion
        .context("no transaction found")?;
```

Not blocking PR though.

crates/solver/src/settlement_simulation.rs
```diff
@@ -119,6 +124,75 @@ pub async fn simulate_and_error_with_tenderly_link(
...
+            .block_number
+            .ok_or_else(|| anyhow!("no block number field exist"))?
+            .as_u64(),
+        transaction
+            .from
+            .ok_or_else(|| anyhow!("no from field exist"))?,
+        transaction.to.ok_or_else(|| anyhow!("no to field exist"))?,
+        transaction
+            .transaction_index
+            .ok_or_else(|| anyhow!("no transaction_index field exist"))?
```
nlordell, line 166: Same as before:
```suggestion
            .context("no block number field exist")?
            .as_u64(),
        transaction
            .from
            .context("no from field exist")?,
        transaction.to.context("no to field exist")?,
        transaction
            .transaction_index
            .context("no transaction_index field exist")?
```

nlordell (approved):
This is very cool!

crates/solver/src/driver.rs
sunce86: Exactly. It should appear as a negative value of saved gas. Note that we also enter the function for cancelled transactions, but for those we don't use access list so calculation will be skipped.

crates/solver/src/settlement_simulation.rs
sunce86: Added,

crates/solver/src/settlement_simulation.rs
sunce86: Added.

MartinquaXD (approved):

crates/solver/src/settlement_submission/submitter.rs
sunce86: Currently we have a histogram but will adjust if we conclude it's needed.

PR #1755, opened 2022-03-24
Settlement maturity across all solvers

josojo:
Martin found some problematic batches today: [1](https://cowservices.slack.com/archives/C036JAGRQ04/p1648110887695829) and [2](https://cowservices.slack.com/archives/C036JAGRQ04/p1648113437434039)

This PR implements settlement maturity association across all solvers, and not just within one solver. That means that if quasimodo or MIP finds valid settlements, because their settlements include 1 valid mature order, then all other orders included in this valid batch should also be mature for all other solvers, like 1inch or 0x.

With this new logic, the issues would not have occurred.

### Test Plan
Only reviewing the code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,7 +53,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "chrono",
- "clap 3.1.3",
+ "clap 3.1.6",
  "model",
  "primitive-types",
  "reqwest",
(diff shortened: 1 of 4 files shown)


sunce86 (approved):

nlordell (approved):

MartinquaXD (approved):

Cargo.lock
```diff
@@ -53,7 +53,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "chrono",
- "clap 3.1.3",
```
vkgnosis, line 56: There are many changes in the lock file which wasn't mentioned in the PR description. I think you might have accidentally run `cargo update`. It would be cleaner to not update the lock file in this PR. You could reset it with `git restore --source origin/main -- Cargo.lock`.

josojo: Sorry, this was not my intention. Thanks. I removed it.

PR #1761, opened 2022-03-29
Liquidity orders fee should not be accounted during solving

josojo:
Currently, the http solvers receive the info that liquidity orders can contribute to the fees within the objective. 
But later on, liquidity orders are not considered for calculating total fees collected. 
See for example the function: `total_scaled_unsubsidized_fees`, it operates only on 'trade_orders' and not on 'liquidity_orders'.

Hence, solvers are "miss-informed" if we specify positive fee amounts for liquidity orders. This PR changes that.
For more information, see here: 
https://cowservices.slack.com/archives/C036JAGRQ04/p1648539991062179?thread_ts=1648474423.631859&cid=C036JAGRQ04

### DIFF
--- a/crates/solver/src/liquidity.rs
+++ b/crates/solver/src/liquidity.rs
@@ -91,6 +91,7 @@ pub struct LimitOrder {
     /// factor to make matching orders more valuable from an objective value
     /// perspective.
     pub scaled_unsubsidized_fee: U256,
+    pub actual_fee_amount: U256,
     pub is_liquidity_order: bool,
     #[cfg_attr(test, derivative(PartialEq = "ignore"))]
     pub settlement_handling: Arc<dyn SettlementHandling<Self>>,
@@ -142,6 +143,7 @@ impl Default for LimitOrder {
             partially_fillable: Default::default(),
             unscaled_subsidized_fee: Default::default(),
             scaled_unsubsidized_fee: Default::default(),
+            actual_fee_amount: Default::default(),
             settlement_handling: tests::CapturingSettlementHandler::arc(),
             is_liquidity_order: false,
             id: Default::default(),
--- a/crates/solver/src/liquidity.rs
+++ b/crates/solver/src/liquidity.rs
@@ -91,7 +91,6 @@ pub struct LimitOrder {
     /// factor to make matching orders more valuable from an objective value
     /// perspective.
     pub scaled_unsubsidized_fee: U256,
-    pub actual_fee_amount: U256,
     pub is_liquidity_order: bool,
     #[cfg_attr(test, derivative(PartialEq = "ignore"))]
     pub settlement_handling: Arc<dyn SettlementHandling<Self>>,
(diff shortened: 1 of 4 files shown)


crates/solver/src/solver/http_solver.rs
```diff
@@ -175,8 +175,12 @@ fn map_tokens_for_solver(orders: &[LimitOrder], liquidity: &[Liquidity]) -> Vec<
 }
 
 fn order_fee(order: &LimitOrder) -> FeeModel {
+    let amount = match order.is_liquidity_order {
+        true => U256::zero(),
```
nlordell, line 179: Instead of 0, should this be the actual fee?
```suggestion
        true => order.actual_fee, // field doesn't currently exist, but maybe we should add it?
```

nlordell (approved):

crates/solver/src/solver/http_solver.rs
josojo: Thanks Nick!
Yes, I was under the impression that all liquidity orders would be charged a fee of zero, hence the shortcut. But you are right that the actual amount is the right spot.


Here, a quick explainer why liquidity orders were included before, but now shouldn't be anymore:

Normally, liquidity orders would not be included as the overall objective, as long as the following would hold:
```
 fees of liqudity order - cost + surplus 
 ```
 But before this PR, we had:
 ```
 fees of liqudity order  + fee subsidy - cost + surplus 
 ```
 Now, it happened that fee subsidy was sometimes higher than the cost. Then these orders would be able to steal surplus from other orders as long as their `fees of liqudity order ` was higher than the surplus stolen from another order

Comment (josojo) [reply crates/solver/src/solver/http_solver.rs]: I will add the additions to the overall objective function from the liquiidty orders (their new fee) into the function `total_scaled_unsubsidized_fees` in another PR

File: crates/solver/src/liquidity.rs [hunk 2]
```diff
@@ -91,6 +91,7 @@ pub struct LimitOrder {
     /// factor to make matching orders more valuable from an objective value
     /// perspective.
     pub scaled_unsubsidized_fee: U256,
+    pub actual_fee_amount: U256,
```
vkgnosis, line 94: Is this different from unscaled_subsidized_fee above?

josojo: I am very sorry. Either, I should not do things on the fly or I should take holiday.

vkgnosis (approved):

crates/solver/src/liquidity.rs
vkgnosis: No worries. That's why we have reviews.

crates/solver/src/solver/http_solver.rs
nlordell: Created #1763 to not forget.

marcovc: Don't we also need to change the solvers themselves?

marcovc: Nevermind, I think the purpose of this PR is to make the driver as it is already in the solver, right?

PR #1762, opened 2022-03-29
Printing auction_id for easier solution finding

josojo:
The solver team spends quite some time to find their logs for a corresponding tx-settlement. With this PR, it should get much easier. One can just search for the tx_hash in kibana and will find an auction id. This auction id will be also passed to the http_solvers and then the http solver will be able to make all their data searchable with this auction_id in kibana.

Hopefully, this saves us some time in the future.

### Test Plan
none

### 

in a future PR, we can make the auction_id unique, in order to make it even more easy

### DIFF
--- a/crates/solver/src/driver.rs
+++ b/crates/solver/src/driver.rs
@@ -156,6 +156,7 @@ impl Driver {
 
     async fn submit_settlement(
         &self,
+        auction_id: u64,
         solver: Arc<dyn Solver>,
         rated_settlement: RatedSettlement,
     ) -> Result<TransactionReceipt> {
@@ -177,8 +178,9 @@ impl Driver {
             Ok(receipt) => {
                 let name = solver.name();
                 tracing::info!(
-                    "Successfully submitted settlement id {} as tx hash {:?}",
+                    "Successfully submitted settlement id {} for the auction {} with tx hash {:?}",
                     rated_settlement.id,
+                    auction_id,
                     receipt.transaction_hash
                 );
                 traded_orders
@@ -437,7 +439,7 @@ impl Driver {
             external_prices: external_prices.clone(),
         };
         tracing::debug!("solving auction ID {}", auction.id);
-        let run_solver_results = self.run_solvers(auction).await;
+        let run_solver_results = self.run_solvers(auction.clone()).await;
         for (solver, settlements) in run_solver_results {
             let name = solver.name();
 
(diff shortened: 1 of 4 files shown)


crates/solver/src/driver.rs
```diff
@@ -437,7 +439,7 @@ impl Driver {
             external_prices: external_prices.clone(),
         };
         tracing::debug!("solving auction ID {}", auction.id);
-        let run_solver_results = self.run_solvers(auction).await;
+        let run_solver_results = self.run_solvers(auction.clone()).await;
```
MartinquaXD, line 442: `auction` can get pretty huge so I think only copying `auction.id` would be nice.
Also if the `auction` doesn't get copied its ownership moves into `run_solvers()` which nicely signals that the `auction` object has no use after that.

crates/solver/src/driver.rs
```diff
@@ -529,9 +536,10 @@ impl Driver {
             .rate_settlements(solver_settlements, &external_prices, gas_price)
             .await?;
         tracing::info!(
-            "{} settlements passed simulation and {} failed",
+            "{} settlements passed simulation and {} failed for auctionId: {}",
```
MartinquaXD, line 539: I think consistently using either `auction`, `auction id` or `auctionId` would also be nice for grep-ability.
The 3 new/updated log messages all use slightly different wording.

MartinquaXD (commented):
I really like the idea!
Just some small nits.

marcovc (approved):
Love it. Coming next: print the auction id somewhere in the user screen that Martin usually posts about :)

marcovc:
I suppose when quasimodo is called by the price estimator then the auction_id will not be set, right?

crates/solver/src/driver.rs
```diff
@@ -468,7 +471,12 @@ impl Driver {
             };
 
             for settlement in &settlements {
-                tracing::debug!("solver {} found solution:\n{:?}", name, settlement);
+                tracing::debug!(
+                    "solver {} found solution:\n{:?} for auction id: {:}",
+                    name,
+                    settlement,
+                    auction_id
```
nlordell, line 478: ubernit: I would reorder the line so that the the ID is near the front. The `{:?}` debug print of the settlement data is very long so it makes it hard to see the auction ID when scanning the logs.

```suggestion
                    "solver {name} found solution for auction {auction_id}:\n{settlement:?}",
```

nlordell (approved):
LGTM. Just one tiny suggestion to make the logs easier to scan.

MartinquaXD (approved):

nlordell:
> I suppose when quasimodo is called by the price estimator then the auction_id will not be set, right?

We have a "request ID" that we can use for this instead of an auction ID.

PR #1765, opened 2022-03-30
Reduce max memory usage when enabling zeroex liquidity

MartinquaXD:
Edit: Filtering results after they got parsed is actually good enough so the solution this PR eventually implemented is way simpler than what is described 👇 . But I will leave the original description in place because it could be an interesting reference for similar problems in the future.

<br>  

### Original Desription
ZeroEx usually offers >1000 limit orders as liquidity. The current implementation simply queries all orders and forwards them to the solvers without filtering out any orders. This results in a significantly higher max memory usage than usual.
Unfortunately the solution is not as easy as deserializing every order and filtering the results afterwards because the max memory usage happens while parsing.
A simple solution would be to decrease the page size when querying the zeroex orders. This unfortunately adds some latency because we would have to do 10 times as many requests to get reasonable memory usage.

Instead I extended the `ZeroExApi` such that you can pass a `Filter<OrderRecord>` which allows you to discard any orders you are not interested in while they are being parsed.
This approach has a few consequences:
1. Because this use case is very niche we can not simply derive the implementation of `Deserialize` but have to write some specialized deserialization logic instead.
2. The code to determine when we reached the final page of limit orders needed to be adapted slightly.
3. `get_orders()` can't use the generic `DefaultZeroExApi::request()` and has to use a specialised implementation which knows how to apply the passed `Filter<OrderRecord>`.

With that functionality in place we can filter out many zeroex limit orders by only collecting the ones where the order trades one of our 6 base tokens against one of the user order tokens. This behavior is already used for other liquidity sources so it should be fine.

The manual benchmarks showed a decrease of the max used memory from ~213MB to ~40MB which is only slightly higher than when the zeroex liquidity is disabled (~37MB).
1. Current implementation (0x disabled)
2. Filtering orders while parsing (0x disabled)
3. Current implementation (0x enabled)
4. Filtering orders while parsing (0x enabled)
Screenshot 2022-03-30 at 11 11 16

Notes:
1. The benchmarks were done in release mode.
2. The benchmark also shows a significant reduction in CPU utilization while parsing.
3. The reduced number of collected limit orders should also decrease the computation time of the quasimodo solver.
4. During my benchmarks the current implementation always got a lower max memory usage than in the staging environment so the memory usage improvements of this change might be a little different when it gets deployed.
[here](https://cowservices.slack.com/archives/C0361CDD1FZ/p1648239893644589?thread_ts=1648218890.934519&cid=C0361CDD1FZ) is the memory usage on staging.

### Test Plan
Benchmarks of max memory usage
added unit test for new filter while parsing logic
updated some existing tests

### DIFF
--- a/crates/solver/src/liquidity/zeroex.rs
+++ b/crates/solver/src/liquidity/zeroex.rs
@@ -5,24 +5,39 @@ use crate::settlement::SettlementEncoder;
 use anyhow::Result;
 use contracts::IZeroEx;
 use model::order::OrderKind;
+use model::TokenPair;
 use primitive_types::U256;
+use shared::baseline_solver::BaseTokens;
 use shared::zeroex_api::{Order, OrderRecord, OrdersQuery, ZeroExApi};
 use std::sync::Arc;
 
 pub struct ZeroExLiquidity {
     pub api: Arc<dyn ZeroExApi>,
     pub zeroex: IZeroEx,
+    pub base_tokens: Arc<BaseTokens>,
 }
 
 impl ZeroExLiquidity {
-    pub async fn get_liquidity(&self) -> Result<Vec<Liquidity>> {
-        Ok(self
-            .api
-            .get_orders(&OrdersQuery::default())
-            .await?
+    pub async fn get_liquidity(&self, user_orders: &[LimitOrder]) -> Result<Vec<Liquidity>> {
+        let zeroex_orders = self.api.get_orders(&OrdersQuery::default()).await?;
+
+        let user_order_pairs = user_orders
+            .iter()
+            .filter_map(|order| TokenPair::new(order.buy_token, order.sell_token));
+        let relevant_pairs = self.base_tokens.relevant_pairs(user_order_pairs);
+
+        let filtered_zeroex_orders = zeroex_orders
             .into_iter()
+            .filter(|record| {
+                match TokenPair::new(record.order.taker_token, record.order.maker_token) {
+                    Some(pair) => relevant_pairs.contains(&pair),
+                    None => false,
+                }
+            })
             .filter_map(|record| self.record_into_liquidity(record))
-            .collect())
+            .collect();
+
+        Ok(filtered_zeroex_orders)
     }
 
     /// Turns 0x OrderRecord into liquidity which solvers can use.
--- a/crates/solver/src/liquidity_collector.rs
+++ b/crates/solver/src/liquidity_collector.rs
@@ -46,7 +46,7 @@ impl LiquidityCollector {
             amms.extend(stable_orders.into_iter().map(Liquidity::BalancerStable));
         }
         if let Some(zeroex_liquidity) = self.zeroex_liquidity.as_ref() {
-            amms.append(&mut zeroex_liquidity.get_liquidity().await?)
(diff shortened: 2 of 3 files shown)


vkgnosis:
It makes sense to filter orders and it makes sense to do this early in the process. However, I think it is reasonable to say that we are able to hold one page (1000) orders in memory at the same time. If we stay at this optimization level we won't have to touch the more complicated serde internals. Could you also test the memory usage of more naively filtering out orders after collecting a full page?
I calculated the size of the one Order struct and got  392 bytes stack size + 32 bytes heap size (from `hash` which is Vec<u8>). This gives 424 bytes which times 1000 is still only ~500 kB.
Another way to reduce memory usage could be to comment out some fields in the struct that we don't use. That way they will be skipped when deserializing. I haven't gone through all of them but a bunch are probably not actively used.
For example `order_hash` is the biggest field but not really needed because we only use it as a uniqueness test in `fn retain_valid_orders`. I don't understand why we need to test uniqueness in the first place because it seems reasonable to assume that their api won't return duplicate orders. If we do need it we could just use `HashSet<Order>` and `derive Hash` on `Order` or even use the `salt` field which is documented as "A value that can be used to guarantee order uniqueness.".
Not directly related to this PR but `order_hash` seems to always be 32 bytes large so we could store it as `[u8; 32]` or `H256`.

vkgnosis:
It makes sense to filter orders and it makes sense to do this early in the process. However, I think it is reasonable to say that we are able to hold one page (1000) orders in memory at the same time. If we stay at this optimization level we won't have to touch the more complicated serde internals. Could you also test the memory usage of more naively filtering out orders after collecting a full page?
I calculated the size of the one Order struct and got  392 bytes stack size + 32 bytes heap size (from `hash` which is Vec<u8>). This gives 424 bytes which times 1000 is still only ~500 kB.
Another way to reduce memory usage could be to comment out some fields in the struct that we don't use. That way they will be skipped when deserializing. I haven't gone through all of them but a bunch are probably not actively used.
For example `order_hash` is the biggest field but not really needed because we only use it as a uniqueness test in `fn retain_valid_orders`. I don't understand why we need to test uniqueness in the first place because it seems reasonable to assume that their api won't return duplicate orders. If we do need it we could just use `HashSet<Order>` and `derive Hash` on `Order` or even use the `salt` field which is documented as "A value that can be used to guarantee order uniqueness.".
Not directly related to this PR but `order_hash` seems to always be 32 bytes large so we could store it as `[u8; 32]` or `H256`.

vkgnosis:
The manual filtering deserialization implementation is cool research by the way. We should keep that in mind if we run into this problem in the future. It just feels a little overkill for this problem.

MartinquaXD:
> Could you also test the memory usage of more naively filtering out orders after collecting a full page?

I'm 99% sure I did this test and it didn't produce good results. I did the benchmark once more anyway and it actually produces similarly good results while being waaay less complicated.
While researching I tried a bunch of different ideas and maybe got confused at some point so this solution slipped through the cracks. 🤦‍♂️ 

In any case, I will rewrite this PR to do the simple solution. This change is probably small enough to be merged before today's testing session.

vkgnosis:
I can review it quickly and we can merge it with one approve.

vkgnosis (approved):

PR #1768, opened 2022-04-01
Check price deviation between external prices and settlements

josojo:
In order to ensure that the driver only settles solutions from the solvers that are reasonable, we will check that settlement prices do not deviate too much from external prices (AMM spot prices calculated in the /auction endpoint in the orderbook).

Since for some tokens, the price deviation to external prices could be quite huge - due to a big slippage -, this PR allows providing a token list, on which the checks will be performed. E.g. we could start with USDC, DAI, USDT, UST, ETH, WETH, WBTC and an allowed price deviation of 3%.

More background information can be found here: https://cowservices.slack.com/archives/C0375NV72SC/p1648801241277359

### Test Plan
Unit tests

--- 
This is only a quick and dirty solution. Later one, we will likely replace this check by some envy checks for AMMs.

### DIFF
--- a/crates/solver/src/main.rs
+++ b/crates/solver/src/main.rs
@@ -280,6 +280,19 @@ struct Arguments {
     /// but at the same time we don't restrict solutions sizes too much
     #[clap(long, env, default_value = "15000000")]
     simulation_gas_limit: u128,
+
+    /// In order to protect against malicious solvers, the driver will check that settlements prices do not
+    /// exceed a max price deviation compared to the external prices of the driver, if this optional0 value is set.
+    /// The max deviation value should be provided as a percentage value. E.g. for a max price deviation
+    /// of 3%, one should set it to 3u64
+    #[clap(long, env)]
+    max_settlement_price_deviation: Option<u64>,
+
+    /// This variable allows to restrict the set of tokens for which a price deviation check of settlement
+    /// prices and external prices is executed. If the value is set to none, then all tokens included
+    /// in the settlement are checked for price deviation.
+    #[clap(long, env)]
+    token_list_for_price_checks: Option<Vec<H160>>,
 }
 
 #[derive(Copy, Clone, Debug, clap::ArgEnum)]
@@ -649,6 +662,8 @@ async fn main() {
         args.weth_unwrap_factor,
         args.simulation_gas_limit,
         args.fee_objective_scaling_factor,
+        args.max_settlement_price_deviation,
+        args.token_list_for_price_checks,
     );
 
     let maintainer = ServiceMaintenance {
--- a/crates/solver/src/main.rs
+++ b/crates/solver/src/main.rs
@@ -40,12 +40,7 @@ use solver::{
     },
     solver::SolverType,
 };
-use std::{
-    collections::{HashMap, HashSet},
-    str::FromStr,
-    sync::Arc,
-    time::Duration,
-};
+use std::{collections::HashMap, str::FromStr, sync::Arc, time::Duration};
 
 #[derive(Debug, Parser)]
 struct Arguments {
@@ -671,8 +666,7 @@ async fn main() {
         args.fee_objective_scaling_factor,
         args.max_settlement_price_deviation
             .map(|max_price_deviation| Ratio::from_float(max_price_deviation).unwrap()),
-        args.token_list_restriction_for_price_checks
-            .map(HashSet::from_iter),
+        args.token_list_restriction_for_price_checks.into(),
     );
 
     let maintainer = ServiceMaintenance {
--- a/crates/solver/src/main.rs
+++ b/crates/solver/src/main.rs
@@ -282,7 +282,7 @@ struct Arguments {
     simulation_gas_limit: u128,
 
     /// In order to protect against malicious solvers, the driver will check that settlements prices do not
-    /// exceed a max price deviation compared to the external prices of the driver, if this optional0 value is set.
+    /// exceed a max price deviation compared to the external prices of the driver, if this optional value is set.
     /// The max deviation value should be provided as a percentage value. E.g. for a max price deviation
     /// of 3%, one should set it to 3u64
     #[clap(long, env)]
--- a/crates/solver/src/main.rs
+++ b/crates/solver/src/main.rs
@@ -298,7 +298,7 @@ struct Arguments {
     /// prices and external prices is executed. If the value is not set, then all tokens included
     /// in the settlement are checked for price deviation.
     #[clap(long, env)]
-    token_list_for_price_checks: Option<Vec<H160>>,
+    token_list_restriction_for_price_checks: Option<Vec<H160>>,
 }
 
 #[derive(Copy, Clone, Debug, clap::ArgEnum)]
@@ -671,7 +671,8 @@ async fn main() {
         args.fee_objective_scaling_factor,
         args.max_settlement_price_deviation
             .map(|max_price_deviation| Ratio::from_float(max_price_deviation).unwrap()),
-        args.token_list_for_price_checks.map(HashSet::from_iter),
+        args.token_list_restriction_for_price_checks
+            .map(HashSet::from_iter),
     );
 
     let maintainer = ServiceMaintenance {
--- a/crates/solver/src/main.rs
+++ b/crates/solver/src/main.rs
@@ -2,6 +2,7 @@ use anyhow::anyhow;
 use clap::{ArgEnum, Parser};
 use contracts::{BalancerV2Vault, IUniswapLikeRouter, WETH9};
 use ethcontract::{Account, PrivateKey, H160};
+use num::rational::Ratio;
 use reqwest::Url;
 use shared::{
     baseline_solver::BaseTokens,
@@ -39,7 +40,12 @@ use solver::{
     },
     solver::SolverType,
 };
-use std::{collections::HashMap, str::FromStr, sync::Arc, time::Duration};
+use std::{
+    collections::{HashMap, HashSet},
+    str::FromStr,
+    sync::Arc,
+    time::Duration,
+};
 
 #[derive(Debug, Parser)]
 struct Arguments {
@@ -283,13 +289,13 @@ struct Arguments {
 
     /// In order to protect against malicious solvers, the driver will check that settlements prices do not
     /// exceed a max price deviation compared to the external prices of the driver, if this optional value is set.
-    /// The max deviation value should be provided as a percentage value. E.g. for a max price deviation
-    /// of 3%, one should set it to 3u64
+    /// The max deviation value should be provided as a float percentage value. E.g. for a max price deviation
+    /// of 3%, one should set it to 0.03f64
     #[clap(long, env)]
-    max_settlement_price_deviation: Option<u64>,
+    max_settlement_price_deviation: Option<f64>,
 
     /// This variable allows to restrict the set of tokens for which a price deviation check of settlement
-    /// prices and external prices is executed. If the value is set to none, then all tokens included
+    /// prices and external prices is executed. If the value is not set, then all tokens included
     /// in the settlement are checked for price deviation.
     #[clap(long, env)]
     token_list_for_price_checks: Option<Vec<H160>>,
@@ -662,8 +668,9 @@ async fn main() {
         args.weth_unwrap_factor,
         args.simulation_gas_limit,
         args.fee_objective_scaling_factor,
-        args.max_settlement_price_deviation,
-        args.token_list_for_price_checks,
+        args.max_settlement_price_deviation
+            .map(|max_price_deviation| Ratio::from_float(max_price_deviation).unwrap()),
+        args.token_list_for_price_checks.map(HashSet::from_iter),
     );
 
     let maintainer = ServiceMaintenance {
--- a/crates/solver/src/settlement.rs
+++ b/crates/solver/src/settlement.rs
@@ -14,7 +14,7 @@ use num::{rational::Ratio, BigInt, BigRational, One, Signed, Zero};
 use primitive_types::{H160, U256};
 use shared::conversions::U256Ext as _;
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     ops::{Mul, Sub},
 };
 
@@ -309,9 +309,12 @@ impl Settlement {
         auction_id: u64,
         solver_name: &str,
         external_prices: &ExternalPrices,
-        max_settlement_price_deviation: u64,
-        tokens_to_satisfy_price_test: &Option<Vec<H160>>,
+        max_settlement_price_deviation: &Ratio<BigInt>,
+        tokens_to_satisfy_price_test: &Option<HashSet<H160>>,
     ) -> bool {
+        if matches!(tokens_to_satisfy_price_test, Some(token_list) if token_list.is_empty()) {
+            return true;
+        }
         // The following check is quadratic in run-time, although a similar check with linear run-time would also be possible.
         // For the linear implementation, one would have to find a unique scaling factor that scales the external prices into
         // the settlement prices. Though, this scaling factor is not easy to define, if reference tokens like ETH are missing.
@@ -334,11 +337,11 @@ impl Settlement {
                 {
                     return false;
                 }
-                let external_price_sell_token = match external_prices.0.get(sell_token) {
+                let external_price_sell_token = match external_prices.price(sell_token) {
                     Some(price) => price,
                     None => return false,
                 };
-                let external_price_buy_token = match external_prices.0.get(buy_token) {
+                let external_price_buy_token = match external_prices.price(buy_token) {
                     Some(price) => price,
                     None => return false,
                 };
@@ -348,19 +351,16 @@ impl Settlement {
                 // |----------------------------------------------------------------------------------------------------------|
                 // |                     clearing_price_sell_token / clearing_price_buy_token                                  |
                 // is bigger than:
-                // max_settlement_price_deviation/100
+                // max_settlement_price_deviation
                 //
                 // This is equal to: |clearing_price_sell_token*external_price_buy_token-external_price_sell_token*clearing_price_buy_token|>
-                // max_settlement_price_deviation/100*clearing_price_buy_token*external_price_buy_token *clearing_price_sell_token
+                // max_settlement_price_deviation*clearing_price_buy_token*external_price_buy_token *clearing_price_sell_token
 
                 let price_check_result = clearing_price_sell_token
                     .clone()
                     .mul(external_price_buy_token)
                     .sub(&external_price_sell_token.mul(&clearing_price_buy_token)).abs()
-                    .gt(&Ratio::new_raw(
-                        BigInt::from(max_settlement_price_deviation),
-                        BigInt::from(100u64),
-                    )
+                    .gt(&max_settlement_price_deviation
                     .mul(&external_price_buy_token.mul(&clearing_price_sell_token)));
                 if !price_check_result {
                     tracing::debug!(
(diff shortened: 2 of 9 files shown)


crates/solver/src/main.rs
```diff
@@ -280,6 +280,19 @@ struct Arguments {
     /// but at the same time we don't restrict solutions sizes too much
     #[clap(long, env, default_value = "15000000")]
     simulation_gas_limit: u128,
+
+    /// In order to protect against malicious solvers, the driver will check that settlements prices do not
+    /// exceed a max price deviation compared to the external prices of the driver, if this optional value is set.
+    /// The max deviation value should be provided as a percentage value. E.g. for a max price deviation
+    /// of 3%, one should set it to 3u64
```
MartinquaXD, line 287: Not sure if that is fine grained enough.
Alternatively one could try to use `f64` or if you want to keep the properties of integers make it so that `1u64` is equal to `0.01%` or something like that.

vkgnosis:
Have we written more about the motivation for this? It feels like there is some context missing.

crates/solver/src/main.rs
```diff
@@ -280,6 +280,19 @@ struct Arguments {
...
+    #[clap(long, env)]
+    max_settlement_price_deviation: Option<u64>,
```
vkgnosis, line 289: This makes more sense to me as a float. 3% would be `0.03`.

crates/solver/src/main.rs
```diff
@@ -280,6 +280,19 @@ struct Arguments {
...
+
+    /// This variable allows to restrict the set of tokens for which a price deviation check of settlement
+    /// prices and external prices is executed. If the value is set to none, then all tokens included
```
MartinquaXD, line 292: > If the value is set to none

From the perspective of someone using the CLI this wording would be confusing because they can't set the value to `None` expliclitly.

josojo:
Here is my write up in slack: https://cowservices.slack.com/archives/C0375NV72SC/p1648801241277359

crates/solver/src/settlement/external_prices.rs
```diff
@@ -16,7 +16,7 @@ use std::collections::{BTreeMap, HashMap};
 /// A collection of external prices used for converting token amounts to native
 /// assets.
 #[derive(Clone, Debug)]
-pub struct ExternalPrices(HashMap<H160, BigRational>);
+pub struct ExternalPrices(pub HashMap<H160, BigRational>);
```
vkgnosis, line 19: @nlordell made this private intentionally so users of the struct wouldn't have to / be able to see the internals of it. It would be nicer if we added a new public method in the impl below to get the price with a comment explaining in what unit the price is.

crates/solver/src/settlement.rs
```diff
@@ -299,6 +303,79 @@ impl Settlement {
...
+    // Checks whether the settlement prices do not deviate more than max_settlement_price_deviation from the auction prices on certain pairs.
+    pub fn satisfies_price_checks(
+        &self,
+        auction_id: u64,
+        solver_name: &str,
+        external_prices: &ExternalPrices,
+        max_settlement_price_deviation: u64,
+        tokens_to_satisfy_price_test: &Option<Vec<H160>>,
+    ) -> bool {
+        // The following check is quadratic in run-time, although a similar check with linear run-time would also be possible.
```
MartinquaXD, line 315: Since this is `O(n^2)` it might make sense to add an early return at the start of the function if the feature was explicitly disabled (`tokens_to_satisfy_price_test == Some(vec![])`).

crates/solver/src/settlement.rs
```diff
@@ -299,6 +303,78 @@ impl Settlement {
...
     }
 
+    // Checks whether the settlement prices do not deviate more than max_settlement_price_deviation from the auction prices on certain pairs.
+    pub fn satisfies_price_checks(
+        &self,
+        auction_id: u64,
+        solver_name: &str,
+        external_prices: &ExternalPrices,
+        max_settlement_price_deviation: u64,
+        tokens_to_satisfy_price_test: &Option<Vec<H160>>,
```
vkgnosis, line 313: Might be simpler for this function to take this as a HashSet. Driver would convert the command line argument into the hashset in it's `new` function.

crates/solver/src/settlement.rs
```diff
@@ -299,6 +303,78 @@ impl Settlement {
...
+            .clearing_prices()
+            .iter()
+            .combinations(2)
+            .any(|clearing_price_vector_combination| {
+                let sell_token_info = clearing_price_vector_combination.get(0).unwrap();
+                let sell_token = sell_token_info.0;
+                let clearing_price_sell_token = sell_token_info.1.to_big_rational();
+                let buy_token_info = clearing_price_vector_combination.get(1).unwrap();
+                let buy_token = buy_token_info.0;
+                let clearing_price_buy_token = buy_token_info.1.to_big_rational();
```
nlordell, line 334: ubernit: You can use de-structuring to make this look a little cooler:

```suggestion
            .any(|clearing_price_vector_combination| {
                let (sell_token, sell_price) = clearing_price_vector_combination[0];
                let clearing_price_sell_token = sell_price.to_big_rational();
                let (buy_token, buy_price) = clearing_price_vector_combination[1];
                let clearing_price_buy_token = buy_price.to_big_rational();
```

Not blocking merge at all though!

vkgnosis (approved):
Code looks good to me. I haven't spent a long time thinking about the formula used though.

crates/solver/src/settlement.rs
```diff
@@ -299,6 +303,78 @@ impl Settlement {
...
+
+                if matches!(tokens_to_satisfy_price_test, Some(token_list) if (!token_list.contains(sell_token)) || !token_list.contains(buy_token))
+                {
+                    return false;
```
nlordell, line 338: Shouldn't this return true? Otherwise we filter out all settlements that trade a token that is not in the list.

crates/solver/src/settlement.rs
```diff
@@ -299,6 +303,78 @@ impl Settlement {
...
     }
 
+    // Checks whether the settlement prices do not deviate more than max_settlement_price_deviation from the auction prices on certain pairs.
+    pub fn satisfies_price_checks(
+        &self,
+        auction_id: u64,
+        solver_name: &str,
+        external_prices: &ExternalPrices,
+        max_settlement_price_deviation: &Ratio<BigInt>,
+        tokens_to_satisfy_price_test: &Option<HashSet<H160>>,
```
nlordell, line 313: Maybe a way to make this simpler would be to:
```suggestion
        tokens_to_satisfy_price_test: &HashSet<H160>,
```

Since `None` and `Some(HashSet::default())` are equivalent? Then we can always `unwrap_or_default()` when creating it so we don't have to handle `None` values everywhere.

crates/solver/src/settlement.rs
```diff
@@ -469,6 +545,102 @@ pub mod tests {
         }
     }
 
+    #[test]
+    pub fn satisfies_price_checks_sorts_out_invalid_prices() {
```
nlordell, line 549: Maybe a unit test to check some of the other paths like:
- Order where one token is included but not the other
- Order where where no tokens are included

crates/solver/src/settlement.rs
```diff
@@ -299,6 +303,78 @@ impl Settlement {
...
+    // Checks whether the settlement prices do not deviate more than max_settlement_price_deviation from the auction prices on certain pairs.
+    pub fn satisfies_price_checks(
+        &self,
+        auction_id: u64,
+        solver_name: &str,
+        external_prices: &ExternalPrices,
+        max_settlement_price_deviation: &Ratio<BigInt>,
+        tokens_to_satisfy_price_test: &Option<HashSet<H160>>,
+    ) -> bool {
+        if matches!(tokens_to_satisfy_price_test, Some(token_list) if token_list.is_empty()) {
```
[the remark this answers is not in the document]
josojo: No, the plan was to provide an option to test it on all token. If the tokens_to_satisfy_price_test.is_none, then I want the test to run on all tokens

crates/solver/src/settlement.rs
josojo: I wanted to have this optional, such that I have an option to run the test on all tokens.

I think I should name the variable differently. Probably: `restrictive_token_set_for_price_check`. Then it would be much easier to understand!  I will change it.

josojo: For None, I was imaging it that we would test on all tokens, and for an empty set, we would test it on no tokens.

Probably, I could have that made clear by a better variable name: 'restrictive_token_list_for_price_check'? Then none means no restriction on the token set, and empty means that we have a very strong restriction - name no tokens?
Would that make it easier to understand?

crates/solver/src/settlement.rs
vkgnosis: You could remove the outer negation, change `any` to `all`, negate inside the clojure. Double check this but I think it is equivalent.

josojo: But any is more performant, isn't it? Then it can stop early.

MartinquaXD: `all()` can also return early if it finds an item which does NOT satisfy the predicate.

crates/solver/src/settlement.rs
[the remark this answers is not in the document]
nlordell: Ahhhhhhh - I misread the document help. Maybe in this case a
```rust
enum TestTokens {
    All,
    Some(HashSet<H160>),
}
impl From<Option<Vec<H160>>> for TestTokens {
    // ...
}
```

crates/solver/src/settlement.rs
nlordell: Same confusion as ☝️

crates/solver/src/settlement.rs
josojo: You guys are right. Thanks!

crates/solver/src/settlement.rs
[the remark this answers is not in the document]
josojo: Lovely! implemented here: [6b2a5b9](https://github.com/gnosis/gp-v2-services/pull/1768/commits/6b2a5b91a1feefd1f2b7cd8159d5c958260079b8)

nlordell (approved):
Looks good! Thanks for addressing all the nits 😄!

crates/solver/src/main.rs
```diff
@@ -280,6 +281,19 @@ struct Arguments {
     /// but at the same time we don't restrict solutions sizes too much
     #[clap(long, env, default_value = "15000000")]
     simulation_gas_limit: u128,
+
+    /// In order to protect against malicious solvers, the driver will check that settlements prices do not
+    /// exceed a max price deviation compared to the external prices of the driver, if this optional value is set.
+    /// The max deviation value should be provided as a float percentage value. E.g. for a max price deviation
+    /// of 3%, one should set it to 0.03f64
+    #[clap(long, env)]
+    max_settlement_price_deviation: Option<f64>,
```
bh2smith, line 290: nit: why not max_price_deviation?

crates/solver/src/main.rs
```diff
@@ -280,6 +281,19 @@ struct Arguments {
...
+
+    /// This variable allows to restrict the set of tokens for which a price deviation check of settlement
+    /// prices and external prices is executed. If the value is not set, then all tokens included
+    /// in the settlement are checked for price deviation.
+    #[clap(long, env)]
+    token_list_restriction_for_price_checks: Option<Vec<H160>>,
```
bh2smith, line 296: Nit: why not price_check_tokens?

crates/solver/src/settlement.rs
```diff
@@ -201,6 +205,21 @@ pub enum Revertable {
     HighRisk,
 }
 
+pub enum TokensForPriceCheck {
+    All,
+    Tokens(HashSet<H160>),
+}
```
bh2smith, line 211: This could also be PriceCheckTokens

crates/solver/src/settlement.rs
```diff
@@ -299,6 +318,77 @@ impl Settlement {
...
     }
 
+    // Checks whether the settlement prices do not deviate more than max_settlement_price_deviation from the auction prices on certain pairs.
+    pub fn satisfies_price_checks(
+        &self,
+        auction_id: u64,
+        solver_name: &str,
+        external_prices: &ExternalPrices,
+        max_settlement_price_deviation: &Ratio<BigInt>,
+        tokens_to_satisfy_price_test: &TokensForPriceCheck,
```
bh2smith, line 328: These two things are always being passed around together to other functions. I think they could be their own struct. Having a dedicated struct for this might even help make these names shorter. 

Suggestion. 

Struct PriceCheck with fields

Max_deviation (currently your max settlement price deviation)
Tokens (currently your tokens to satisfy price test)

crates/solver/src/settlement.rs
```diff
@@ -299,6 +318,77 @@ impl Settlement {
...
+    ) -> bool {
+        if matches!(tokens_to_satisfy_price_test, TokensForPriceCheck::Tokens(token_list) if token_list.is_empty())
```
bh2smith, line 330: Maybe I'm a bit rusty but I can not tell what this is trying to do. Maybe it has something to do with the long names... or the if statement on the inside without an else... I don't know

crates/solver/src/settlement.rs
```diff
@@ -299,6 +318,77 @@ impl Settlement {
...
+            .clearing_prices()
+            .iter()
+            .combinations(2)
+            .all(|clearing_price_vector_combination| {
+                let (sell_token, sell_price) = clearing_price_vector_combination[0];
+                let clearing_price_sell_token = sell_price.to_big_rational();
+                let (buy_token, buy_price) = clearing_price_vector_combination[1];
+                let clearing_price_buy_token = buy_price.to_big_rational();
+
+                if matches!(tokens_to_satisfy_price_test, TokensForPriceCheck::Tokens(token_list) if (!token_list.contains(sell_token)) || !token_list.contains(buy_token))
```
bh2smith, line 350: Same here. I find this match confusing because of the content on the right. I think if you make a struct as described above, it could implement some method that moves the awkward logic of these match statements into one place.

crates/solver/src/settlement.rs
```diff
@@ -299,6 +318,77 @@ impl Settlement {
...
+                // Condition to check: Deviation of clearing prices is bigger than max_settlement_price deviation
+                //
+                // |clearing_price_sell_token / clearing_price_buy_token - external_price_sell_token/external_price_buy_token)|
+                // |----------------------------------------------------------------------------------------------------------|
+                // |                     clearing_price_sell_token / clearing_price_buy_token                                  |
+                // is bigger than:
+                // max_settlement_price_deviation
+                //
+                // This is equal to: |clearing_price_sell_token*external_price_buy_token-external_price_sell_token*clearing_price_buy_token|>
+                // max_settlement_price_deviation*clearing_price_buy_token*external_price_buy_token *clearing_price_sell_token
```
bh2smith, line 371: It's a shame the names here are so long. Really make the equations hard to read. 
```suggestion
                // This is equal to: |clearing_price_sell_token * external_price_buy_token - external_price_sell_token * clearing_price_buy_token|>
                // max_settlement_price_deviation * clearing_price_buy_token * external_price_buy_token * clearing_price_sell_token
```

crates/solver/src/settlement.rs
```diff
@@ -299,6 +318,77 @@ impl Settlement {
...
+                    Some(price) => price,
+                    None => return true,
+                };
+                let external_price_buy_token = match external_prices.price(buy_token) {
+                    Some(price) => price,
+                    None => return true,
+                };
+                // Condition to check: Deviation of clearing prices is bigger than max_settlement_price deviation
+                //
+                // |clearing_price_sell_token / clearing_price_buy_token - external_price_sell_token/external_price_buy_token)|
```
bh2smith, line 364:
```suggestion
                // |clearing_price_sell_token / clearing_price_buy_token - external_price_sell_token / external_price_buy_token)|
```

crates/solver/src/settlement.rs
```diff
@@ -299,6 +318,77 @@ impl Settlement {
...
+                // |----------------------------------------------------------------------------------------------------------|
+                // |                     clearing_price_sell_token / clearing_price_buy_token                                  |
```
bh2smith, line 366:
```suggestion
                // |                     clearing_price_sell_token / clearing_price_buy_token                                 |
```

crates/solver/src/settlement.rs
```diff
@@ -299,6 +318,77 @@ impl Settlement {
...
+                // max_settlement_price_deviation
+                //
+                // This is equal to: |clearing_price_sell_token*external_price_buy_token-external_price_sell_token*clearing_price_buy_token|>
+                // max_settlement_price_deviation*clearing_price_buy_token*external_price_buy_token *clearing_price_sell_token
+
+                let price_check_result = clearing_price_sell_token
+                    .clone()
+                    .mul(external_price_buy_token)
+                    .sub(&external_price_sell_token.mul(&clearing_price_buy_token)).abs()
+                    .lt(&max_settlement_price_deviation
```
bh2smith, line 377: Sneaking an inequality into a chain of other arithmetic operations is misleading for me, I was not expecting the result to be Boolean on the next line

I think this would be more clear if the inequality was made in the if statement below.

bh2smith (approved):
I would really like to see a dedicated struct for the two fields that always occur together. This could house some of the match logic that I find difficult to read. 

As usual, I think the naming conventions are to long in some places but this could be resolved with the mentioned struct. 

Approving because all my comments are nits, the logic is sound.

crates/solver/src/settlement.rs
josojo: I kind of like the succinctness of it

crates/solver/src/main.rs
josojo: WIth your naming, I the None option would be harder to understand. If one passes in none, for `price_check_tokens` I would guess no tokens are being used.

josojo:
I have not addressed the struct proposal, as it's also not done for other variables of the driver. Though, actually, I think its a good proposal. 
Since there is a little bit time pressure, I would like to merge it and not wait until the discussion was happening.

If someone else agrees that this should be done, please note it in this PR, and then I will do it in a follow-up PR.

nlordell:
> If someone else agrees that this should be done, please note it in this PR, and then I will do it in a follow-up PR.

I really like the struct proposal as well.

> as it's also not done for other variables of the driver

I don't think it needs to be done in the `Arguments` struct, but instead in the `fn main()` you can:
```
let price_checker = PriceCheck::new(
    args.max_settlement_price_deviation
        .map(|max_price_deviation| Ratio::from_float(max_price_deviation).unwrap()),
    args.token_list_restriction_for_price_checks.into(),
);
```

And then pass in `price_checker` to the `Driver` instance instead of passing in both fields.

crates/solver/src/settlement.rs
nlordell: I hadn't realized before, but the `add`, `mul`, etc. operations are using the `std::ops::*` traits which means they are what get called with `+`, `*`, etc. (so no need to use the method calls). This means you could have written this as:

```rust
    let price_check_result =
        (clearing_price_sell_token * &external_price_buy_token - &external_price_sell_token * &clearing_price_buy_token).abs()
        < max_settlement_price_deviation;
```

Note that in this case, I also see the value of writing it in terms of the `add`, `mul`, etc. method calls and am  happy either way.

crates/solver/src/main.rs
nlordell: I prefer `max_settlement_price_deviation` here because for runtime arguments `max_price_deviation` would be super ambiguous (price deviation for what?).

crates/solver/src/settlement.rs
nlordell: `matches` syntax is just sugar for:

```rust
if match tokens_to_satisfy_price_test {
    TokensForPriceCheck::Tokens(token_list) if token_list.is_empty() => true,
    _ => false,
}  {
    // ...
}
```

I think `clippy` even suggests this construct in some cases.

Issue #1739, opened 2022-03-21
Reduce Number of Unsupported Tokens


nlordell:
This _epic_ issue captures the work around trying to reduce the number of unsupported tokens.

### Some tasks:
- [ ] Facilitate identifying unsupported tokens (record metrics, logs, etc.)
    - May require adding new metric collection mechanisms that aren't Prometheus based
- [ ] Once we have more data on what are the common tokens that aren't supported, there might be some low-hanging fruit that for supporting additional tokens with low effort.
- [ ] Long term: trade simulation. We could simulate trades from the price estimators for users with balance and approval. This would allow us to check if one of our solvers would indeed be able to match the token on-chain (even if we can't find liquidity to validate it with our "bad token detector").

### We should also start measuring unsupported token counts:
- [ ] Counter for how often a token is deemed "unsupported" when users are querying quotes.

PR #1771, opened 2022-04-01
Add metrics for bad token detection

MartinquaXD:
Implements first task of #1739 

In order to develop a strategy to handle bad token detection and measure its success we need more information.
This PR covers the following pieces of information:
1. The number of times a bad token detector returns good token, bad token or an error.
2. The error when a bad token detection fails.
3. The reason when a token gets detected as bad.

Because prometheus can't handle string based information very well I decided to simply log that information. When the time comes to evaluate the findings we simply have to query Kibana and look at that data.
To do that I added the `InstrumentedBadTokenDetector` which wraps around a `dyn BadTokenDetecting` and simply takes care of the metrics and logging.
This pattern is already frequently used by us but I also added an extension trait which help us build `BadTokenDetecting` hierarchies with a builder pattern. I think this design is pretty neat and would also help us with the way more complex `dyn PriceEstimating` hierarchies. Let me know what you think.

Note that a good tokens get over reported because requests containing bad tokens return way earlier. This indicates that we are checking tokens more often than we need to.

### Test Plan
Manually tested metrics and logs for error, good quality and bad quality case.

Error:  
```
2022-04-01T11:20:37.776Z  WARN request{id=1}: shared::bad_token::instrumented: bad token detection for 0x134f18a6864efb61ea8636d1bdb0cc3c4c8fd798 returned error:
and some extra context

Caused by:
    0: with context
    1: some test error
```

Bad token:
```
2022-04-01T11:20:46.661Z  WARN request{id=3}: shared::bad_token::instrumented: bad token detection for 0x134f18a6864efb61ea8636d1bdb0cc3c4c8fd797 returned Bad { reason: "deny listed" }
```

Metrics:
```
# HELP gp_v2_api_token_quality_results Tracks how many token detections result in good or bad token quality or an error.
# TYPE gp_v2_api_token_quality_results counter
gp_v2_api_token_quality_results{quality="bad"} 1
gp_v2_api_token_quality_results{quality="error"} 1
gp_v2_api_token_quality_results{quality="good"} 8
```

### DIFF
--- a/crates/shared/src/bad_token/instrumented.rs
+++ b/crates/shared/src/bad_token/instrumented.rs
@@ -1,0 +1,62 @@
+use super::{BadTokenDetecting, TokenQuality};
+use crate::metrics::get_metric_storage_registry;
+use anyhow::Result;
+use prometheus::CounterVec;
+use prometheus_metric_storage::MetricStorage;
+
+pub trait InstrumentedBadTokenDetectorExt {
+    fn instrumented(self) -> InstrumentedBadTokenDetector;
+}
+
+impl<T: BadTokenDetecting + 'static> InstrumentedBadTokenDetectorExt for T {
+    fn instrumented(self) -> InstrumentedBadTokenDetector {
+        InstrumentedBadTokenDetector {
+            inner: Box::new(self),
+        }
+    }
+}
+
+#[derive(MetricStorage, Clone, Debug)]
+#[metric(subsystem = "token_quality")]
+struct Metrics {
+    /// Tracks how many token detections result in good or bad token quality or an error.
+    #[metric(labels("quality"))]
+    results: CounterVec,
+}
+
+pub struct InstrumentedBadTokenDetector {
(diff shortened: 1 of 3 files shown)


crates/shared/src/bad_token/instrumented.rs
```diff
@@ -0,0 +1,62 @@
...
+            // tokens/errors and get the information from Kibana when we need it.
+            Err(err) => {
+                tracing::warn!(
+                    "bad token detection for {:?} returned error:\n{:?}",
+                    token,
+                    err
+                );
+                "error"
+            }
+            Ok(quality @ TokenQuality::Bad { .. }) => {
```
sunce86, line 48: Interesting syntax :)

sunce86 (approved):
Nice PR, seems like a right fit for the purpose it needs to fulfill...

nlordell (approved):

vkgnosis (approved):

PR #1778, opened 2022-04-02
skip "already known transaction" errors

sunce86:
This PR skips node submission errors that are related to already known transactions.
Discussed here: https://cowservices.slack.com/archives/C0361CDD1FZ/p1648897295271429

### DIFF
--- a/crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
+++ b/crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
@@ -17,7 +17,7 @@ use futures::FutureExt;
 use gas_estimation::{EstimatedGasPrice, GasPrice1559};
 use shared::Web3;
 
-const ALREADY_KNOWN_TRANSACTION: [&str] = [
+const ALREADY_KNOWN_TRANSACTION: [&str; 6] = [
     "Transaction gas price supplied is too low", //openethereum
     "Transaction nonce is too low",              //openethereum
     "already known",                             //infura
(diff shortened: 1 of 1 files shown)


crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
```diff
@@ -17,6 +17,15 @@ use futures::FutureExt;
 use gas_estimation::{EstimatedGasPrice, GasPrice1559};
 use shared::Web3;
 
+const ALREADY_KNOWN_TRANSACTION: [&str; 6] = [
```
nlordell, line 20: One thing you can do so you don't need to change the type if you add new messages is:
```suggestion
const ALREADY_KNOWN_TRANSACTION: &[&str] = &[
```

nlordell (approved):

crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
```diff
@@ -64,8 +73,17 @@ impl TransactionSubmitting for CustomNodesApi {
                     });
                 }
                 Err(err) => {
-                    tracing::warn!(?err, ?lable, "single custom node tx failed");
-                    super::track_submission_success(lable.as_str(), false);
+                    // error is not real error if transaction pool responded that received transaction is already in the pool
+                    let real_error = match err.clone() {
```
MartinquaXD, line 77: The error does not need to be cloned. This also works:
```rust
let real_error = match &err {
//...
```

crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
sunce86: Very nice. Thanks.

crates/solver/src/settlement_submission/submitter/custom_nodes_api.rs
sunce86: True, thanks.

MartinquaXD (approved):

PR #1782, opened 2022-04-04
Query 0x liquidity only fillable by GPv2Settlement contract

### LINKED ISSUE #1769: Also query 0x limit orders which can only be settled by us
0x allows people to decide who can settle their orders.
If the `sender` field of an order is `0x000..000` anyone is allowed to fill the order, if it contains a specific address only this address is allowed to fill it.

Right now we only collect the orders which can be settled by anyone but in the future it would make sense that some people (PMMs) submit orders which are specifically intended for us.
In that case the `sender` field would contain `0x9008D19f58AAbD9eD0D60971565AA8510560ab41` (our settlement contract).

AFAIK there is no way to get orders for both addresses with a single query so we should run queries for both in parallel.
Probably the most appropriate place to add this change is in `ZeroExLiquidity::get_liquidity()`.

MartinquaXD:
Fixes #1769 

Before this change we only queried 0x orders as liquidity which can be filled by anybody. But 0x also offers the option to make your limit order only fillable by a specific address.
If people want to become private market makers for cow protocol using 0x they simply have to set the `sender` address of their order to the address of our settlement contract.
Now we only have to additionally query those 0x limit orders and we are done.
At the moment this gives us no additional liquidity because nobody uses this option, yet. But at least we are prepared already.
Since the additional orders get fetched in parallel this PR should add no runtime overhead to the liquidity collection.

Note:
This only other way to become a PMM at the moment is to contact us directly so we can list your address as a market maker which entails the privilege of submitting partially fillable orders. Because the 0x integration goes through the 0x smart contract it's less gas efficient but offers a lower barrier of entry to becoming a market maker. This way also enables people to earn a fee by market making for us.

### Test Plan
CI

### DIFF
--- a/crates/solver/src/liquidity/zeroex.rs
+++ b/crates/solver/src/liquidity/zeroex.rs
@@ -3,7 +3,7 @@ use crate::interactions::ZeroExInteraction;
 use crate::liquidity::{Exchange, LimitOrder, Liquidity};
 use crate::settlement::SettlementEncoder;
 use anyhow::Result;
-use contracts::IZeroEx;
+use contracts::{GPv2Settlement, IZeroEx};
 use model::order::OrderKind;
 use model::TokenPair;
 use primitive_types::U256;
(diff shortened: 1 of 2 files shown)


nlordell (approved):
Easy peasy.

Have you tried placing a 0x limit order to see if it shows up?

vkgnosis (approved):

